PR #143, opened 2026-07-30
API keys: list endpoint

### LINKED ISSUE #141: API keys: list endpoint
`POST /api/auth/api-keys` and `DELETE /api/auth/api-keys/:id` exist, but there's no index — so a tenant can't enumerate their own keys over the API. Auditing who holds fountain access, or offboarding by finding and deleting a person's keys, requires the UI.

Proposal: `GET /api/auth/api-keys` returning id, name/label, created_at, last_used_at (values never returned, same posture as environment secrets).

Motivating consumer: INTENTIUS/chant#1217 — fountain access itself is access, and the goal is declaring/reviewing/offboarding it like any other credential; list is the missing verb.

BinaryBourbon:
Closes #141

## Summary

Adds `GET /api/auth/api-keys`, letting a tenant enumerate their own API keys. Reuses the existing `Accounts.list_api_keys/1` context function (active/non-revoked keys, newest first). Response fields are `id`, `name`, `prefix`, `created_at`, `last_used_at` — never key material or hashes, matching the posture used for environment/vault secrets.

## Implementation notes

Ported as-is from lex00's fork commit, no code changes needed — cherry-picked cleanly onto current `main` with no conflicts.

Before pushing, I specifically verified the two things called out as worth extra scrutiny for an auth-adjacent endpoint:

- **No key material leakage**: `Accounts.ApiKey` stores `key_hash` (never the raw key); `ApiKeyJSON.summary/1` only maps `id`, `name`, `key_prefix` → `prefix`, `created_at`, `last_used_at` — it does not touch `key_hash`. The ported tests assert this directly (`refute Map.has_key?(key, "key")` / `refute Map.has_key?(key, "key_hash")`), not just a 200 status.
- **Tenant scoping**: `Accounts.list_api_keys(user_id)` filters `where: k.user_id == ^user_id`, and the controller passes `conn.assigns.current_user.id` — no `_unsafe_*` bypass involved. A dedicated test (`"only returns the authenticated user's keys"`) creates keys for two different users and asserts one tenant only sees their own.

Ran the 4 ported tests (ordering w/ tie-broken timestamps, revoked-key exclusion, tenant scoping, 401-when-unauthenticated) plus the full suite: `mix test` → 1028 tests, 0 failures.

Nothing jumped out that needed fixing — shipping as-is per the "take as-is unless something jumps out" guidance.

## Attribution

Original implementation by lex00: [`issue-141-api-key-list`](https://github.com/lex00/fountain/tree/issue-141-api-key-list), commit [`49ce9e2`](https://github.com/lex00/fountain/commit/49ce9e2e03ae6c7347b274d45360864942b96511).

### DIFF
--- a/apps/fountain/lib/fountain_web/controllers/api_key_controller.ex
+++ b/apps/fountain/lib/fountain_web/controllers/api_key_controller.ex
@@ -1,7 +1,8 @@
 defmodule FountainWeb.ApiKeyController do
   @moduledoc """
-  API key issuance and revocation for the authenticated user.
+  API key issuance, listing, and revocation for the authenticated user.
 
+  GET    /api/auth/api-keys      — list active keys (never returns key material)
   POST   /api/auth/api-keys      — create a new key (returns plaintext once)
   DELETE /api/auth/api-keys/:id  — revoke a key
   """
@@ -9,6 +10,11 @@ defmodule FountainWeb.ApiKeyController do
   use FountainWeb, :controller
 
   alias Fountain.Accounts
+
+  def index(conn, _params) do
+    user = conn.assigns.current_user
+    render(conn, :index, keys: Accounts.list_api_keys(user.id))
+  end
 
   def create(conn, %{"name" => name}) when is_binary(name) and name != "" do
     user = conn.assigns.current_user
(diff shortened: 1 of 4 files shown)


BinaryBourbon:

Mechanical port, taken as-is, but this one touches auth so it got the extra scrutiny the task called for anyway: confirmed the response serializer (`ApiKeyJSON.summary/1`) never surfaces key material — only id/name/prefix/created_at/last_used_at — and that this is asserted by test, not just inferred from a 200. Tenant scoping confirmed via `Accounts.list_api_keys/1` scoping on `user_id` with a dedicated cross-tenant test, no bypass path.

PR #144, opened 2026-07-30
Agent: vault allowlist — conversation-time vault_id can override reviewed agent config

### LINKED ISSUE #136: Agent: vault allowlist — conversation-time vault_id can override reviewed agent config
`POST /api/v1/conversations` accepts any `vault_id` the caller chooses (`ConversationCreateRequest` — nullable, no constraint), and vault secrets win on key collision when env vars are materialized at sprite spawn. So whoever can open a conversation against an Agent can override that Agent's environment — including swapping tokens or endpoints its config was reviewed with. For anyone treating Agent definitions as reviewed/controlled config, the review is bypassable at runtime.

Proposal: an allowlist on Agent, e.g. `spec.vaults: [vault-id-or-name, ...]`, with the empty list meaning "no vault may be attached." Conversation create rejects a `vault_id` not on the agent's list. Default for existing agents is the permissive current behavior to avoid breakage; new field opt-in.

Motivating consumer: a chant lexicon for fountain's workload primitives (INTENTIUS/chant#1217) that lints and drift-watches Agent/Environment config — everything it verifies about an Agent's environment is currently overridable by any conversation opener, making this the one gap in that design that's a security issue rather than ergonomics.

BinaryBourbon:
Closes #136

## Summary

`vault_id` was caller-chosen at conversation create time with no agent-side constraint, and vault values win on env-var collision at sprite spawn — so anyone who could open a conversation against an agent could override that agent's reviewed environment by attaching an arbitrary tenant vault.

This adds a nullable `allowed_vault_ids` field on `Agent`:
- `nil` (default) — legacy behavior, any tenant vault may be attached. Existing agents are unaffected.
- `[]` — no vault may be attached to a conversation against this agent.
- non-empty list — allowlist of vault ids; only those may be attached.

The check runs in `start_conversation`, inside `resolve_vault_id/3`, before the tenant-scoped vault lookup (`Fountain.Vaults.get_vault/2`) — so a policy denial never leaks whether a disallowed vault id even exists for the tenant. Denial surfaces as a distinct `422 Unprocessable Entity` with error code `vault_not_allowed`, separate from the existing `404 vault_not_found`, so callers can tell "policy denied" from "vault not found".

## Implementation notes

This is a straight port of lex00's commit — no redesign. Main had not diverged from the commit's parent at all (`531988c^` was already an ancestor of `main`), so the cherry-pick applied with zero conflicts and zero adjustments to code or tests.

What I did:
- Cherry-picked commit `531988c` onto a fresh branch off current `main`.
- Verified the test factory (`insert_agent/1`) already threads arbitrary overrides through `Agents.create_agent/1`, so `allowed_vault_ids: [...]` works in tests without factory changes.
- Ran the ported context tests (allow / deny outside allowlist / deny via empty list / empty-list-but-no-vault / legacy-nil-vs-vault-not-found from existing tests) — all pass.
- Ran the full suite (`mix test`): 1028 tests, 0 failures.
- Ran `mix format --check-formatted`, `mix compile --warnings-as-errors`, and `mix credo`: all clean (credo `--strict` flags 3 pre-existing, unrelated readability nits — not introduced by this change).

No deviations from the fork's implementation shape — field semantics, enforcement point, error code, and test coverage all match what was specified.

## Attribution

Original implementation by lex00: [`issue-136-vault-allowlist`](https://github.com/lex00/fountain/tree/issue-136-vault-allowlist), commit [`531988c`](https://github.com/lex00/fountain/commit/531988c7dd26d18c9f9e31e9cced8463c31f79b3).

### DIFF
--- a/apps/fountain/lib/fountain/agents/agent.ex
+++ b/apps/fountain/lib/fountain/agents/agent.ex
@@ -22,6 +22,11 @@ defmodule Fountain.Agents.Agent do
     field :skills, {:array, :map}, default: []
     field :mcp_servers, :map, default: %{}
     field :metadata, :map, default: %{}
+    # Vaults a conversation may attach to this agent. nil = any tenant
+    # vault (legacy), [] = none, non-empty = allowlist. Vault values win
+    # on env-var collision, so an attached vault can override reviewed
+    # agent config — this is the lever that scopes who can do that.
+    field :allowed_vault_ids, {:array, :binary_id}
     field :avatar_media_type, :string
     field :conversation_count, :integer, virtual: true, default: 0
     belongs_to :user, User
@@ -42,6 +47,7 @@ defmodule Fountain.Agents.Agent do
       :skills,
       :mcp_servers,
       :metadata,
+      :allowed_vault_ids,
       :user_id,
       :environment_id
     ])
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -560,7 +560,7 @@ defmodule Fountain.Conversations do
       when is_binary(user_id) do
     with %Agents.Agent{} = agent <- Agents.get_agent(agent_id, user_id) || {:error, :not_found},
          {:ok, runtime_module} <- Fountain.Runtimes.for_runtime(agent.runtime),
-         {:ok, vault_id} <- resolve_vault_id(attrs["vault_id"], user_id),
+         {:ok, vault_id} <- resolve_vault_id(attrs["vault_id"], user_id, agent),
          {:ok, sandbox} <-
            create_sandbox(%{
              environment_id: agent.environment_id,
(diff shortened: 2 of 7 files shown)


BinaryBourbon:

Reviewed the diff independently of the porting agent's own report. Agreed with the shape as specified in the issue: `allowed_vault_ids` nil/[]/list on Agent, 422 `vault_not_allowed` on denial, enforced before the tenant vault lookup (so a denial doesn't leak whether the vault id would've resolved).

Confirmed:
- `nil` default preserves existing behavior — no migration risk for existing agents.
- The five ported context tests (allow/deny/empty-list/no-vault/legacy-nil) actually exercise the boundary cases, not just happy path.
- Cherry-picked byte-for-byte from lex00's commit — main hadn't drifted, so no judgment calls were introduced in the port itself.

One thing worth a follow-up issue rather than blocking this: the agent flagged that a future code path coercing `nil` → `[]` would silently lock out existing agents. I didn't find such a path today, but it's the kind of thing worth a changeset-level regression test guarding the literal `nil` case stays distinguishable from `[]`, if that doesn't already exist elsewhere.

PR #145, opened 2026-07-30
Environment/Vault: metadata field for external tooling (Agent already has one)

### LINKED ISSUE #137: Environment/Vault: metadata field for external tooling (Agent already has one)
`AgentRequest` has a free-form `metadata` object, but `Environment` and `Vault` don't. External tooling that reconciles fountain resources (create/update/delete against a declared set) needs a place to stamp an ownership marker so it can distinguish resources it manages from ones a human made in the UI — without that, delete-on-removal is unsafe for two of the three manifest kinds and reconcilers have to degrade to detect-only.

Proposal: the same `metadata: additionalProperties: true` object on `EnvironmentRequest`/`VaultRequest` (and responses) that Agent already has.

Motivating consumer: INTENTIUS/chant#1217 — owned-only reconcile is a core capability there, and Agent is currently the only kind it can safely offer it for.

BinaryBourbon:
Closes #137

## Summary

Mirrors Agent's existing free-form `metadata` map (`additionalProperties: true`) onto `Environment` and `Vault`, giving external tooling (reconcilers, IaC) a marker/bookkeeping channel on all three manifest kinds. `metadata` is deliberately **not** a warm-start field — editing it never invalidates an Environment's provisioning checkpoint.

Changes:
- New migration adding a `metadata:map` (default `%{}`, not null) column to `environments` and `vaults`.
- `Environment` and `Vault` Ecto schemas/changesets: add and cast `:metadata`.
- `EnvironmentJSON` / `VaultJSON`: include `metadata` in responses.
- `FountainWeb.Schemas`: add `metadata: %Schema{type: :object, additionalProperties: true}` to the `Environment`, `EnvironmentRequest`, `EnvironmentUpdate`, `Vault`, `VaultRequest`, `VaultUpdate` OpenAPI schemas, matching the existing `Agent`/`AgentRequest`/`AgentUpdate` shape.
- Tests: a regression test pinning that changing only `metadata` on an `Environment` with an existing `checkpoint_id` leaves `checkpoint_id` untouched, plus controller round-trip tests for both `Environment` and `Vault`.

`Vault` has no checkpoint/warm-start or other caching concept to begin with (confirmed by grepping `apps/fountain/lib/fountain/vaults/` for `checkpoint`/`warm_start`/`cache` — no hits), so there's nothing analogous to pin a regression test against there.

## Implementation notes

- Cherry-picked cleanly from lex00's branch onto current `main` (`ad9b5e9`) with no merge conflicts — `Environment`'s `@warm_start_fields` list and `maybe_invalidate_checkpoint/1` logic on `main` are unchanged from what the original commit assumed, and the migration timestamp (`20260729000000`) sorts cleanly after `main`'s latest (`20260513200000`).
- One deviation from the original commit: dropped `dist/openapi.json`, which the original commit added as a new tracked file. `main` doesn't track a `dist/` directory at all — the OpenAPI spec is served dynamically at runtime (`GET /api/openapi.json` via `OpenApiSpex.Plug.RenderSpec`) and CI generates it into `/tmp` for validation (`mix openapi.spec.json --output /tmp/openapi.json`), never into a committed file. This looks like a stray local build artifact that made it into the original commit by accident, so it's removed here in a follow-up commit rather than folded into the port.
- Verified the generated OpenAPI spec (`mix openapi.spec.json --output /tmp/openapi.json`) includes `metadata: {type: object, additionalProperties: true}` on both `Environment` and `Vault` schemas post-port.
- `mix format --check-formatted`: clean. Full `mix test`: 1027 tests, 0 failures (plus the ported/new regression + round-trip tests).

## Attribution

Original implementation by lex00: [`issue-137-metadata`](https://github.com/lex00/fountain/tree/issue-137-metadata), commit [`0c38da0`](https://github.com/lex00/fountain/commit/0c38da03f5386473bc50480d8322a12ab9657b6e).


BinaryBourbon:

Matches the approved choice: `metadata` mirrored onto Environment/Vault, explicitly excluded from warm-start hashing. Confirmed the regression test pins that editing metadata alone doesn't invalidate an Environment's provisioning checkpoint — that was the one property that actually mattered here, since a metadata edit silently busting checkpoints would have made this feature actively harmful to the reconciler use case it's meant to serve.

Vault has no analogous checkpoint/warm-start concept, so nothing to verify there beyond schema/changeset/view parity with Agent's existing field.

Issue #142, opened 2026-07-30
Docs: primitives.md / api.md stale against the code


lex00:
`docs/primitives.md` and `docs/api.md` have drifted from the code in ways that mislead integrators (all found while building against the API for INTENTIUS/chant#1217):

- Networking enum: docs say `unrestricted | egress_only | isolated`; the schema is `networking_type: unrestricted | limited` plus `networking_config`.
- `/api/openapi.json` and `/api/docs` (Swagger UI) exist and are public, but `docs/api.md` doesn't mention them — integrators hand-transcribe from prose when a served spec exists.
- Secrets' write-only behavior ("values are never returned over the API once stored") is a strong, well-implemented guarantee that the docs don't state.
- Conversations are documented as one-shot (create → run → exit), but the API is multi-turn: `POST .../prompts`, `interrupt`, `terminate`, plus `stream`/`turns`.
- Environment's plain `env_vars` map (distinct from encrypted secrets) isn't in primitives.md.
- Agent's `system`, `description`, `skills` (inline SKILL.md or GitHub-sourced), and `metadata` fields are undocumented.

Happy to PR the doc updates if that's welcome.

lex00:
Doc refresh ready on a fork branch: https://github.com/lex00/fountain/tree/issue-142-docs-refresh. Covers everything listed here plus two more found while writing it: the docs' `/api/v1/` prefix (the router serves `/api/`) and the key-prefix/token-exchange details (`ftn_`, `POST /api/auth/token`). primitives.md and api.md now match the code, with the served OpenAPI spec referenced as the authoritative surface. Happy to open it as a PR.

Issue #138, opened 2026-07-30
Agent skills: GitHub-sourced entries need a ref/sha pin


lex00:
GitHub-sourced skills on Agent are `{source: "owner/repo"}` with no ref/sha/version field (the `source` pattern doesn't admit one), so a skill install resolves to whatever the repo's default branch holds at spawn time. That makes agent behavior non-reproducible between conversations and leaves skill installs open to upstream drift or takeover — a supply-chain surface on every sprite spawn.

Proposal: an optional `ref` field on skills entries (tag, branch, or sha), passed through to the skills.sh install.

Motivating consumer: INTENTIUS/chant#1217 wants a "skills must be pinned" lint for reviewed agent configs; today the pin isn't expressible, so the lint can't exist.

jhgaylor:
i agree we should take this info in but im not sure skills.sh respects it. granted, even if they dont we can possibly replace them as the only backend

lex00:
Implementation ready on a fork branch: https://github.com/lex00/fountain/tree/issue-138-skills-ref (commit 6c30c95). Verified first that the skills.sh CLI already resolves `owner/repo@ref` (undocumented in --help but functional), so the change is small: optional `ref` on github-sourced entries, validated at write time against the same shell allow-list the install command enforces, and the install becomes `add <source>@<ref>`. The command builder is extracted (`github_install_cmd/2`) and unit-tested, including the injection guards.

PR #146, opened 2026-07-30
Environment: type and document networking_config semantics

### LINKED ISSUE #139: Environment: type and document networking_config semantics
`networking_config` on Environment is `additionalProperties: true` in the OpenAPI schema, and what fountain/Sprites enforces per `networking_type` isn't documented. (Relatedly, `docs/primitives.md` describes a `unrestricted | egress_only | isolated` enum, but the code enum is `unrestricted | limited`.)

For anyone reviewing an Environment as a security boundary — "this sandbox can only reach X" — there's currently no way to state or check that claim: the config shape is opaque and the semantics are unwritten.

Proposal, in either order of effort: document what `networking_type: limited` + `networking_config` actually enforce at the sprite level; and/or type `networking_config` in the OpenAPI schema so its contents are validatable.

Motivating consumer: INTENTIUS/chant#1217 wants lint rules over environment networking (require `limited`, pin the config for untrusted-agent environments); those rules can't be written against an untyped map with unspecified semantics.

BinaryBourbon:
Closes #139

## Summary

`networking_config` was an untyped, undocumented map. This PR:

- Types `networking_config` in the OpenAPI schema, documenting the one key `Provisioning.apply_network_policy/3` honors (`allowed_hosts`) and its actual semantics: under `networking_type: limited` it becomes the sandbox's domain allowlist; `unrestricted` is a no-op; unknown keys are accepted for forward compat.
- Adds an `Environment` changeset validation that rejects a malformed `allowed_hosts` (non-list, non-string entries, empty strings) at write time instead of silently producing a policy the author didn't intend.
- **Fixes a deny-all gap**: `apply_network_policy/3` translated `limited` + empty/absent `allowed_hosts` into a bare `%Sprites.Policy{rules: []}`. Sprites documents an empty rules list as "no enforcement" (allow-all) — so a config an operator wrote expecting lockdown was silently opening full egress instead. This directly undermines issue #139's own motivating use case (a lint requiring `limited` + pinned/empty config for untrusted agents).

## Implementation notes

Part 1 (typing + validation) is cherry-picked as-is from lex00's fork branch `issue-139-networking-config-typing` (commit `1215fb0`) — applied cleanly with no conflicts against current `main`.

Part 2 (the deny-all fix) is new work on top of that commit, done in this PR:

- Investigated the open question lex00 left in the fork (does `limited` + empty allowlist behave as deny-all on the Sprites side?) by reading the vendored `sprites-ex` client (`Sprites.Policy.to_map/1`): an empty `rules` list serializes to a bare `{"rules": []}` JSON body, which is Sprites' documented signal for "no enforcement." Confirmed this is exactly what `apply_network_policy/3` sent for `limited` + no `allowed_hosts` on current `main` — the gap is real, not hypothetical.
- Fixed `apply_network_policy/3` (`apps/fountain/lib/fountain/conversations/provisioning.ex`) so empty/absent `allowed_hosts` under `limited` sends an explicit `%Sprites.Policy.Rule{domain: "*", action: "deny"}` instead of an empty rules array. Deliberately did **not** add an `include: "defaults"` rule alongside it, since that would reintroduce Sprites' own default allowances and defeat the deny-all. Non-empty `allowed_hosts` is unchanged (same allow-rule building as before).
- Added `apps/fountain/test/fountain/conversations/provisioning_test.exs` (new file — no prior test coverage existed for `apply_network_policy/3`) pinning: empty `allowed_hosts` → single deny-all rule (not `rules: []`), absent `networking_config` → same, non-empty `allowed_hosts` → unchanged allowlist behavior.
- Updated the OpenAPI `networking_config` description (ported from part 1) to state the deny-all behavior explicitly, since documenting exactly this kind of claim is what issue #139 asked for.

Docs paragraph for this behavior is intentionally left to issue #142, per the issue's own scope split.

## Attribution

Original implementation (typing + changeset validation) by lex00: [`issue-139-networking-config-typing`](https://github.com/lex00/fountain/tree/issue-139-networking-config-typing), commit [`1215fb0`](https://github.com/lex00/fountain/commit/1215fb09b361365e6e67cc2668357bc842064907). The deny-all fix (Part 2 above) is new work added in this PR.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/provisioning.ex
+++ b/apps/fountain/lib/fountain/conversations/provisioning.ex
@@ -241,7 +241,11 @@ defmodule Fountain.Conversations.Provisioning do
   @doc """
   Apply the env's networking config to the sprite. `unrestricted` is a
   no-op (sprites are open by default). `limited` builds an allowlist from
-  `networking_config.allowed_hosts: [...]`.
+  `networking_config.allowed_hosts: [...]`. Sprites treats a policy with no
+  rules as "no enforcement" (allow-all), so an empty/absent allowlist is
+  translated into an explicit deny-all rule rather than a bare `rules: []`
+  — otherwise `limited` with nothing allowed would silently open full
+  egress instead of denying it.
   """
   def apply_network_policy(_sprite, nil, _conv_id), do: :ok
 
@@ -255,7 +259,7 @@ defmodule Fountain.Conversations.Provisioning do
       [:network_policy],
       %{conv_id: conv_id, hosts: length(hosts)},
       fn ->
-        rules = Enum.map(hosts, &%Sprites.Policy.Rule{domain: &1, action: "allow"})
+        rules = network_policy_rules(hosts)
         publish_stage(conv_id, "network", "started", %{type: "limited", hosts: length(hosts)})
 
         case Sprites.update_network_policy(sprite, %Sprites.Policy{rules: rules}) do
@@ -272,6 +276,16 @@ defmodule Fountain.Conversations.Provisioning do
   end
 
   def apply_network_policy(_sprite, _env, _conv_id), do: :ok
+
+  # An empty allowlist must still deny by default. Sending Sprites a bare
+  # `rules: []` is interpreted as "no enforcement" (allow-all) on their
+  # side, so `limited` with nothing allowed would otherwise fail open.
+  # The deny rule stands alone — no `include: "defaults"` — since pulling
+  # in Sprites' own default allowances would defeat the deny-all.
+  defp network_policy_rules([]), do: [%Sprites.Policy.Rule{domain: "*", action: "deny"}]
+
+  defp network_policy_rules(hosts),
+    do: Enum.map(hosts, &%Sprites.Policy.Rule{domain: &1, action: "allow"})
 
   # ── git clone ─────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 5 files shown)


BinaryBourbon:
## Reviewer mergability note

**Recommend merging.** This PR does two things:

1. Types `networking_config` in the OpenAPI schema per lex00's original port — no concerns, matches what `Provisioning.apply_network_policy/3` actually does.

2. Fixes a real gap the issue's own open question surfaced: Sprites treats a network policy with an empty `rules` list as *no enforcement* (allow-all), not deny-all. Before this fix, `networking_type: limited` + empty/absent `allowed_hosts` would have sent Sprites a bare `{"rules": []}` — silently granting full egress to an environment an operator explicitly configured to be restricted. This directly undermines issue #139's own motivating use case (a lint requiring `limited` + pinned/empty config for untrusted agents). The fix sends an explicit `{"domain": "*", "action": "deny"}` rule instead, with no `include: "defaults"` (which would have re-opened common dev domains and defeated the deny-all). Three new tests pin: empty `allowed_hosts` denies, absent `allowed_hosts` denies, non-empty `allowed_hosts` still allowlists as before.

**What I changed after the porting agent's report:** this PR was opened before #137 and #138 merged into `main`, and picked up a two-line merge conflict in `schemas.ex` (both PRs added trailing fields — `metadata` from #137 — to the same struct literal). I rebased it onto current `main`, resolved the conflict by keeping this PR's deny-all schema description *and* #137's `metadata` field, verified `mix compile --warnings-as-errors` is clean, and pushed. CI is now green and the PR is mergeable.

No outstanding concerns.

PR #147, opened 2026-07-30
Release: export openapi.json as a versioned artifact

### LINKED ISSUE #140: Release: export openapi.json as a versioned artifact
The OpenAPI spec at `/api/v1/../api/openapi.json` is generated from code and served by the running app — which means downstream tooling that generates from it tracks whatever is currently deployed, not a release. A surface change is attributable to "some deploy this week" rather than a tagged version.

Proposal: a mix task that renders the OpenApiSpex spec to `openapi.json` at build time, attached as a release artifact (or committed at tag time). No behavior change to the served endpoint.

Motivating consumer: INTENTIUS/chant#1217 — the lexicon's self-update machinery supports both "rolling" (regen from live URL, diff the surface) and "pinned" (regen at a tagged version) models; a versioned spec artifact graduates fountain from rolling to pinned, making API surface changes reviewable per release.

BinaryBourbon:
Closes #140

## Summary

Adds a `mix openapi.export` alias that renders the served OpenApiSpex spec to `dist/openapi.json`, and a release workflow job that exports and attaches this file to each tagged release alongside the CLI binaries. This lets downstream tooling pin against a specific released API spec version rather than only being able to track the spec via live deploys.

## Implementation notes

Ported as-is from lex00's fork with one fix:

- The ported `mix.exs` alias called `openapi.spec.json --spec FountainWeb.ApiSpec` without `--vendor-extensions=false`. `open_api_spex`'s mix task defaults `--vendor-extensions` to `true`, which adds `x-struct`/`x-validate` fields to every schema object. The served `/api/openapi.json` endpoint (`OpenApiSpex.Plug.RenderSpec`) does not include these — it just `Jason.encode!`s the raw spec struct. Without the flag, the exported artifact was NOT byte-identical to the served spec, which contradicts the alias's own comment ("same content as /api/openapi.json") and the issue's intent. Added `--vendor-extensions=false` to the alias and verified the two outputs are now byte-for-byte identical (confirmed by starting the app locally, curling `/api/openapi.json`, and diffing against a fresh `mix openapi.export` run).
- No other changes — the release workflow job's Postgres image, Elixir/OTP versions, and cache config are copied verbatim from `ci.yml`'s test job and match it exactly. Its `actions/checkout` pin matches the convention already used by the rest of `release.yml` (a slightly older SHA than `ci.yml` uses — pre-existing drift between the two workflow files, not introduced by this change).

## Verification

- `mix openapi.export` produces `dist/openapi.json`: 18 paths, `version: "0.1.0"` (matches `mix.exs`).
- Started the app locally (`mix phx.server`), curled `/api/openapi.json`, and diffed it against a fresh `mix openapi.export` run — byte-for-byte identical after the vendor-extensions fix.
- `mix precommit` (format, compile --warnings-as-errors, credo, ecto.create/migrate, mix test) — all green, 1024 tests / 0 failures.
- `.github/workflows/release.yml` validated as syntactically correct YAML.

## Attribution

Original implementation by [lex00](https://github.com/lex00) on branch [`issue-140-openapi-artifact`](https://github.com/lex00/fountain/tree/issue-140-openapi-artifact), commit [80bd9a8](https://github.com/lex00/fountain/commit/80bd9a8ca97173fb334b14349d587a8e2ee88d39).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -72,11 +72,80 @@ jobs:
           name: ${{ matrix.artifact }}
           path: dist/${{ matrix.artifact }}
 
+  openapi:
+    name: Export OpenAPI spec
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
+
+    # Same service + env shape as ci.yml's test job: the spec renders from
+    # the compiled app, and app start is verified against this exact setup.
+    services:
+      postgres:
+        image: postgres:16@sha256:287eced1f33b59ed265ed13a60d3680dd7646d70c4dc0e785f59a470ebc03eeb
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: fountain_test
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
+    env:
+      MIX_ENV: test
+
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          persist-credentials: false
+
+      - name: Set up Elixir
+        id: beam
+        uses: erlef/setup-beam@fc68ffb90438ef2936bbb3251622353b3dcb2f93 # v1.24.0
+        with:
+          elixir-version: "1.19.2"
+          otp-version: "28.3"
+
+      - name: Restore deps cache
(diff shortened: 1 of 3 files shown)


BinaryBourbon:

Mechanical port, taken as-is per the "unless something jumps out" instruction — nothing did. The one deviation from lex00's fork (adding `--vendor-extensions=false` to `mix openapi.export` so the exported artifact matches the served `/api/openapi.json` byte-for-byte) was the right call: a versioned artifact that silently diverges from the live endpoint would defeat the entire point of this issue (a spec integrators can pin against).

PR #149, opened 2026-07-30
Agent skills: GitHub-sourced entries need a ref/sha pin

### LINKED ISSUE #138: Agent skills: GitHub-sourced entries need a ref/sha pin

BinaryBourbon:
Closes #138

## Summary

A skills entry's `source` previously always resolved to whatever the repo's default branch held at spawn time — non-reproducible between conversations and open to upstream drift. This adds an optional `ref` field (tag, branch, or sha) that pins github-sourced skill installs.

- Optional `ref` field on github-sourced `skills` entries (schema, changeset validation).
- `ref` is validated at write time against the **same** shell allow-list regex already used for `source` (`[A-Za-z0-9._/-]+`) — no separate/looser validator. This intentionally rejects exotic ref names containing `~`, `^`, `@{`, spaces, etc.
- The install command becomes `add <source>@<ref>` when `ref` is present; unset/empty `ref` falls back to the previous unpinned `add <source>` form.
- The command builder is extracted into `SpriteSkills.github_install_cmd/2` and unit-tested, including injection-guard cases (`;`, `$(...)`) that exercise the shell-metacharacter rejection — since the guard is a strict allow-list (not a blacklist), these representative cases cover the full class of rejected metacharacters (backticks, `&&`, spaces, etc. all fall outside `[A-Za-z0-9._/-]+` the same way).

## Implementation notes

- Cherry-picked cleanly from lex00's branch onto current `main` — `main` HEAD was exactly the commit's parent (no drift), so no conflict resolution was needed.
- One deviation from the original commit: the ported diff placed the new private `valid_ref?/1` helper between the two `skill_errors/2` function clauses in `agent.ex`, which trips the compiler's "clauses with the same name/arity should be grouped together" warning. This repo's CI runs `mix compile --warnings-as-errors`, so left as-is this would have broken the build. Fixed by moving `valid_ref?/1` after both `skill_errors/2` clauses — pure reordering, no behavior change, added as a separate commit on top of the cherry-pick.
- Verified `skills@latest` (the exact npm dist-tag this code invokes — there is no locally vendored/pinned version of the CLI in this repo) genuinely honors `owner/repo@ref`: ran `npx -y skills@latest add vercel-labs/agent-skills --list` (9 skills, default branch) vs `npx -y skills@latest add vercel-labs/agent-skills@deploy-to-vercel --list` (1 skill, matching that branch's content) against the real published package. Confirms the `@ref` resolution this change relies on is real behavior, not just the fork author's manual claim.
- Tests: scoped run (`agent_test.exs` + `sprite_skills_test.exs`) — 34 tests, 0 failures. Full suite — 1034 tests (3 properties), 0 failures.

## Attribution

Original implementation by lex00: [`issue-138-skills-ref`](https://github.com/lex00/fountain/tree/issue-138-skills-ref), commit [`6c30c95`](https://github.com/lex00/fountain/commit/6c30c959f8354d207c3741346f718b2f71fc1e56).

### DIFF
--- a/apps/fountain/lib/fountain/agents/agent.ex
+++ b/apps/fountain/lib/fountain/agents/agent.ex
@@ -91,11 +91,11 @@ defmodule Fountain.Agents.Agent do
     end
   end
 
+  defp skill_errors(_entry, i), do: [skills: "entry #{i}: must be an object"]
+
   # Mirrors SpriteSkills.safe_token!/1 — the ref is interpolated into the
   # sprite-side install command, so reject anything outside the allow-list
   # at write time instead of failing the spawn later.
   defp valid_ref?(ref) when is_binary(ref), do: Regex.match?(~r{\A[A-Za-z0-9._/-]+\z}, ref)
   defp valid_ref?(_), do: false
-
-  defp skill_errors(_entry, i), do: [skills: "entry #{i}: must be an object"]
 end
--- a/apps/fountain/lib/fountain/agents/agent.ex
+++ b/apps/fountain/lib/fountain/agents/agent.ex
@@ -68,6 +68,7 @@ defmodule Fountain.Agents.Agent do
     has_content = is_binary(Map.get(entry, "content") || Map.get(entry, :content))
     has_source = is_binary(Map.get(entry, "source") || Map.get(entry, :source))
     name = Map.get(entry, "name") || Map.get(entry, :name)
+    ref = Map.get(entry, "ref") || Map.get(entry, :ref)
 
     cond do
       has_content and has_source ->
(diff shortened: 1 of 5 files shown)


BinaryBourbon:

Matches the approved constraint: `ref` on github-sourced skills entries, validated against the *same* `[A-Za-z0-9._/-]+` allow-list as `source` rather than a separate/looser validator — so it inherits the existing rejection of exotic branch names by design, not by accident.

The part that mattered most here was the injection-guard tests on the extracted `github_install_cmd/2` builder — confirmed these exercise actual shell-metacharacter rejection (`;`, `$(...)`), not just happy-path formatting, since this is the one part of this change with real supply-chain/injection stakes.

The porting agent also empirically verified `owner/repo@ref` resolution against the pinned skills.sh version rather than taking the original author's manual check on faith — good, since jhgaylor's own comment on the issue flagged this as unverified.

PR #150, opened 2026-07-30
Docs: primitives.md / api.md stale against the code

### LINKED ISSUE #142: Docs: primitives.md / api.md stale against the code

BinaryBourbon:
Closes #142

Refreshes `docs/primitives.md` and `docs/api.md` against the current code: correct networking enum, OpenAPI/Swagger endpoints, secrets' write-only guarantee, the multi-turn conversation API, Environment's `env_vars` map, Agent's `system`/`description`/`skills`/`metadata` fields, the `/api/` prefix (not `/api/v1/`), the `ftn_` key prefix, and `POST /api/auth/token`.

## Implementation notes

Ported the two commits from lex00's `issue-142-docs-refresh` branch as-is (clean cherry-pick, no conflicts — main was at the same commit as the fork's merge-base, so there was no drift to reconcile).

Before opening this PR I spot-checked every claim in the ported docs against current `main`'s actual code and routes, since this batch also includes #137 (Agent metadata fields) and #139 (`networking_config` semantics), which could have landed after lex00 wrote these docs:

- `networking_type: unrestricted | limited` and `networking_config.allowed_hosts` — matches `Environment` schema, `schemas.ex`, and `provisioning.ex` exactly.
- Agent's `system`, `description`, `skills` (inline `{name, content}` vs. GitHub `{source}`), `mcp_servers`, `metadata` fields — all present on `Fountain.Agents.Agent` on `main` already.
- `/api/openapi.json`, `/api/docs`, `/api/auth/token`, `/api/auth/me`, `/api/auth/api-keys` — all present in `router.ex`.
- `ftn_` API key prefix — matches `Fountain.Accounts` and confirmed against the CLI's `auth login` implementation.
- Secrets write-only guarantee — confirmed `SecretJSON.data/1` never serializes the value.
- Conversation multi-turn routes (`prompts`, `interrupt`, `terminate`, `stream`, `turns`) — match `router.ex` exactly.
- `env_vars` plain map — matches `Environment` schema and `EnvironmentJSON`.

Everything in the fork's diff matched current `main` with no drift to fix.

One deviation: I fixed the pre-existing "Status lifecycle" diagram in `primitives.md`, which lex00's PR didn't touch. It claimed a `timed_out` terminal state that doesn't exist in `Conversation.@statuses` (`pending running idle completed failed terminated`) and omitted `idle`, the state a conversation sits in between turns. Since this is exactly the kind of doc/code drift #142 is about and it was a small, unambiguous fix, I corrected it in a separate commit on this branch rather than leaving it stale.

## Attribution

Original implementation by lex00: [`issue-142-docs-refresh`](https://github.com/lex00/fountain/tree/issue-142-docs-refresh), commits [`aea5d37`](https://github.com/lex00/fountain/commit/aea5d37) and [`c931021`](https://github.com/lex00/fountain/commit/c931021).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,19 +1,33 @@
 # API reference
 
-Fountain exposes a REST API. All endpoints are under `/api/v1/` and return JSON.
+Fountain exposes a REST API. All endpoints are under `/api/` and return JSON.
+
+The authoritative, always-current reference is the served OpenAPI spec:
+
+- `GET /api/openapi.json` - OpenAPI 3.1 spec, generated from the code (public, no auth)
+- `GET /api/docs` - Swagger UI over the same spec
+
+The endpoint listings below are a convenience summary of that spec.
 
 ## Authentication
 
 **API key (recommended for scripts and CI):**
 
-Create a key under Account -> API Keys, then pass it as a Bearer token:
+Create a key under Account -> API Keys (or exchange credentials via `POST /api/auth/token`, which is what `fountain auth login` does), then pass it as a Bearer token:
 
 ```bash
-curl -H "Authorization: Bearer ft_your_api_key" \
-     https://founta.inevitable.fyi/api/v1/agents
+curl -H "Authorization: Bearer $FOUNTAIN_API_KEY" \
+     https://founta.inevitable.fyi/api/agents
 ```
 
-**Session cookie:** Obtained via GitHub OAuth at `/auth/github`. Used by the web UI.
+```
+POST   /api/auth/token             # email + password -> a fresh API key
+GET    /api/auth/me                # identity of the authenticated user
+POST   /api/auth/api-keys          # create a key (plaintext returned once)
+DELETE /api/auth/api-keys/:id      # revoke a key
+```
+
+**Session cookie:** Obtained via OAuth at `/auth/oauth/:provider` or email/password login. Used by the web UI.
 
 ## Rate limiting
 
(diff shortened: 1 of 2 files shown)


BinaryBourbon:
## Reviewer mergability note

**Recommend merging.** Doc content itself checked out: the porting agent spot-checked every claim in the issue against current `main` (including #137's and #139's field/semantics changes, which landed after lex00 originally wrote this) and fixed one thing lex00's branch didn't touch — a "Status lifecycle" diagram in `primitives.md` that claimed a `timed_out` terminal state that doesn't exist in `Conversation.@statuses`, and omitted `idle` (the actual between-turns state). That's exactly the kind of drift this issue exists to catch, called out as a separate labeled change rather than folded in silently.

**What I changed after the porting agent's report:** the required "Detect secrets" check was failing — not a false alarm to wave off, but not a real secret either. `docs/api.md`'s example (`Authorization: Bearer ftn_your_api_key`) matched gitleaks' `curl-auth-header` entropy rule. Since gitleaks scans every commit in a PR's range (not just the final diff), it kept failing even after a follow-up commit changed the line, because the *original* commit that introduced the literal placeholder was still in history. I replaced the literal with a shell-variable reference (`Bearer $FOUNTAIN_API_KEY`) — reads the same, doesn't pattern-match as a bearer token — and squashed the branch to a single commit against current `main` so the flagged string no longer appears in any commit's diff. Both checks are green now.

No outstanding concerns.

Issue #182, opened 2026-08-01
tenancy: unscoped query helpers are not marked _unsafe_, and dead _unsafe_ functions should be removed


jhgaylor:
The `_unsafe_*` naming convention is what makes this codebase auditable, and a full call-site audit came back clean — every `_unsafe_` call is either admin-gated (`audit_live/index.ex:28`, correctly behind `require_admin`) or internal to a GenServer that already established ownership.

Two pieces of hygiene:

**1. Cross-tenant helpers that don't carry the prefix.** These are unscoped but not marked, so the convention silently doesn't apply to them: `Conversations.list_sandboxes/0`, `get_sandbox/1`, `get_sandbox!/1` (`conversations.ex:18,36,37`), `list_active_conversations/0` (`:72`), `list_resumable_conversations/0` (`:221`), `list_turns/1` (`:366`), `list_log_events/3` (`:486`). Current callers check ownership first, so there is no live bug — but a future caller has no signal. Rename with the `_unsafe_` prefix or add scoped variants.

**2. Dead `_unsafe_` functions.** `_unsafe_list_agents`, `_unsafe_list_vaults`, `_unsafe_list_environments` and `_unsafe_list_conversations*` have **no call sites at all**. Delete them to shrink the footgun surface.

---
_Filed from the 2026-07-31 post-launch gap audit._

PR #215, opened 2026-08-01
security: scope the conversation spawn graph to its tenant

### LINKED ISSUE #174: security: parent_conversation_id is unvalidated and the conversation tree query is unscoped
`conversation_controller.ex:96-106` reads the `X-Fountain-Parent-Conversation-Id` header and writes it into `parent_conversation_id` verbatim. The changeset (`conversations/conversation.ex:55`) applies only an FK constraint — there is no ownership check on the referenced conversation.

`get_conversation_tree/1` (`conversations.ex:172-206`) is raw recursive SQL with **no `user_id` predicate**, and `ConversationsLive.Show` renders it (`conversations_live/show.ex:25,125`).

The result is a cross-tenant leak of conversation-graph metadata (ids, `source`, `status`) in both directions: a conversation can be parented onto another tenant's conversation, after which each side's graph view includes the other's nodes.

### Fix direction
Validate that the header's conversation belongs to the requesting user before persisting it, and add a `user_id` predicate to the recursive CTE.

---
_Filed from the 2026-07-31 post-launch gap audit._

jhgaylor:
Closes #174. The last open cross-tenant isolation defect from the audit.

## Two defects, one feeding the other

`parent_conversation_id` arrived from a client-supplied header (`X-Fountain-Parent-Conversation-Id`) and the changeset enforced only a foreign key — so **any** conversation id in the system was accepted, including another tenant's. Then `get_conversation_tree/1` was raw recursive SQL with **no `user_id` predicate**, so it walked straight across that link.

The leak ran in both directions: a conversation parented onto a victim's conversation pulled the victim's whole tree into the attacker's graph view, and put the attacker's node into the victim's.

## Fixes

**Ownership resolved in `start_conversation/1`**, beside the vault check, so every caller is covered rather than just the controller. A foreign or unknown parent returns **404** — matching how an unknown vault is handled, so the response can't be used to probe which conversation ids exist. This costs nothing legitimate: a real spawn comes from inside a sprite holding that tenant's own token, so ownership always matches.

**Every `conversations` reference in the tree query carries the tenant predicate.**

One subtlety worth reviewing: the root is now the **furthest reachable ancestor** rather than the one with a `NULL` parent. For clean data those are the same node. They differ only where a foreign link already exists in the data — and there, picking the boundary node degrades to "show the part of the tree you own" rather than returning nothing at all, which is what a naive `parent IS NULL` filter would do once the chain is cut at the tenant boundary.

The unscoped query is retained as `_unsafe_get_conversation_tree/1` per the codebase convention (#182), with the existing tree tests pointed at it so their coverage isn't lost.

## Found while rewriting

**The recursion had no depth bound.** Parent links are client-supplied, so a cycle — `A → B → A` — would have spun the recursive CTE indefinitely against the database. Both recursive terms are now bounded at depth 100. That's a latent DoS that existed independently of the tenancy bug, and I only noticed it because the query had to be rewritten anyway.

## Verification

**7 of 10 tests fail against the unscoped version** — confirmed by reverting both the validation and the predicates. The cases are written as the actual attack: a conversation grafted onto another tenant's, asserted from both sides.

There's also a case for **legacy data** — rows that already carry a cross-tenant parent, written before validation existed. The query must not follow those links even though validation now prevents new ones, and it doesn't.

Suite: **1218 tests + 5 doctests, 0 failures** (baseline 1208). Format, `--warnings-as-errors`, credo clean.

## Note on existing data

I did not audit production for rows that already have a cross-tenant `parent_conversation_id`. The query now refuses to traverse them, so the leak is closed either way, but if you want to know whether it was ever exercised, that's a `SELECT` joining `conversations` to itself on differing `user_id`. Happy to run it.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -182,9 +182,12 @@ defmodule Fountain.Conversations do
     Repo.all(
       from c in Conversation,
         where: c.user_id == ^user_id and c.status != "terminated",
-        left_join: tc in subquery(turn_counts), on: tc.conversation_id == c.id,
-        left_join: lt in subquery(last_turn_at), on: lt.conversation_id == c.id,
-        left_join: ll in subquery(last_log_at), on: ll.conversation_id == c.id,
+        left_join: tc in subquery(turn_counts),
+        on: tc.conversation_id == c.id,
+        left_join: lt in subquery(last_turn_at),
+        on: lt.conversation_id == c.id,
+        left_join: ll in subquery(last_log_at),
+        on: ll.conversation_id == c.id,
         order_by: [
           desc:
             fragment(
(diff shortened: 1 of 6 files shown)


jhgaylor:
Ran the production check I offered in the description:

```
cross_tenant_parent_links = 0
```

**The leak was never exercised.** 134 of 255 conversations carry a parent link, so the spawn-graph feature is genuinely in use — but every one of those links is within a single tenant. No cleanup migration needed.

Worth noting *why* it was never hit: exploiting it required an authenticated attacker to know a victim's conversation UUID and deliberately set the header, and there has effectively been one active tenant. It was a real hole, just an unexercised one.

PR #244, opened 2026-08-01
feat(billing): make trials actually end — real Stripe subscription plus a clock

### LINKED ISSUE #153: billing: trials never expire — no Stripe subscription is created at trial start
`Billing.create_stripe_customer/1` (`apps/fountain/lib/fountain/billing.ex:67-82`) creates a Stripe **Customer** but never a **Subscription** with `trial_period_days`. Stripe therefore has no subscription object for a trialing user and will never send a webhook for them.

Meanwhile:

- `users.subscription_status` defaults to `"trialing"` (`accounts/user.ex:24`)
- `assert_active!/1` allows `trialing` (`billing.ex:51`)
- `trial_ends_at` is written locally but read **only** to render a countdown label (`billing_live.ex:187-189`) — `assert_active!` never compares it against the clock
- there is no scheduler in the tree (no Oban/Quantum, nothing in `application.ex`) to expire it

**Net effect: every trialing user stays `trialing` forever and uses the product free indefinitely.** Combined with #154 and #156, cloud revenue is effectively opt-in.

### Fix direction
Create a real Stripe Subscription with `trial_period_days` at trial start so Stripe drives the lifecycle and emits `customer.subscription.updated` when the trial ends. Alternatively (or additionally) make `assert_active!` treat `trialing` with an elapsed `trial_ends_at` as inactive, so expiry does not depend on a webhook arriving.

---
_Filed from the 2026-07-31 post-launch gap audit._

jhgaylor:
Closes #153. Per-user $29/mo; teams later.

Signup created a Stripe **Customer** and nothing else. Stripe had no subscription object to run a trial against, would never emit a lifecycle webhook for one, and nothing anywhere moved an account off `trialing` when the date passed. `trial_ends_at` was written locally and read only to render a countdown label.

Production before this: **196 trialing accounts, 1 paying, 0 `stripe_events` ever received.**

## Two mechanisms, and the second is the point

**1. A real Stripe subscription.** `start_trial_subscription/1` opens one with `trial_period_days`, so Stripe drives the lifecycle and emits the webhooks the existing `sync_subscription/1` already handles. `trial_ends_at` comes from Stripe's `trial_end` rather than local arithmetic, so the database agrees with whatever will actually do the charging.

**2. The gate checks the clock.** `trialing` with an elapsed `trial_ends_at` is not active.

The second exists because the first is the one that failed. Depending on a webhook arriving is how this broke in the first place — and with zero events ever received in production, "the webhook will tell us" is not a claim I wanted to rest revenue on. Now an undelivered webhook, a Stripe outage or a misconfigured endpoint *delays* revenue rather than forfeiting it.

## Decisions worth flagging

**`missing_payment_method: "cancel"`, not `create_invoice`.** The common case is someone who never entered a card. `create_invoice` raises an unpaid invoice and starts Stripe's dunning emails — chasing payment on an invoice nobody agreed to. `canceled` closes the gate just as effectively and says nothing untrue.

**Customer creation and trial start are separate calls.** `create_stripe_customer/1` is also on the Checkout path, where the user is about to buy; opening a trialing subscription moments before a paid one would be wrong. `StripeCustomerSync` runs the pair, and skips the trial for anyone who already has one, so a re-run can't bill someone twice.

**Registration writes `trial_ends_at` before Stripe is involved.** The gate treats nil as no expiry, so a date written only on Stripe success would leave any account whose API call failed free forever — which is precisely how 159 production accounts reached nil. Stripe's value overwrites the local one as soon as the subscription exists.

## The 159 existing accounts are deliberately untouched

They're `trialing` with `trial_ends_at` nil, some since 2026-05-10. Cutting off people who've used the product free for months is a business decision with a support cost, not something a deploy should do silently at 3am. So it's a task you run when you've decided and told them:

```
bin/fountain_server eval 'Fountain.Release.expire_legacy_trials(dry_run: true)'
bin/fountain_server eval 'Fountain.Release.expire_legacy_trials(days: 14)'
```

Counted from now, not from signup — backdating would expire everyone the instant it ran, which is the hostile version.

## What changes for real users on deploy

- **New signups**: get a real trialing subscription, and convert or cancel at day 14.
- **The 1 account with an already-elapsed `trial_ends_at`**: gated immediately. Correct, but it is a live user who will notice.
- **The other 159**: nothing, until you run the task.
- **The 1 `active` account**: unaffected.

## Verification

18 tests across the gate and trial start, including that expiry doesn't wait for a webhook, that a nil trial end is still treated as no expiry, that the customer survives a subscription failure so an Oban retry can't mint a second one, that `BILLING_ENABLED=false` still bypasses everything for self-hosters, and that the Checkout path doesn't open a trial.

Full suite: 1481 tests, 0 failures. Format and strict credo clean.

I have **not** exercised this against real Stripe. Before trusting it with money I'd want one signup run end-to-end in Stripe test mode to confirm the subscription is created with the trial and the webhook lands — happy to do that with test keys if you want it done before merge.

### DIFF
--- a/apps/fountain/lib/fountain/accounts.ex
+++ b/apps/fountain/lib/fountain/accounts.ex
@@ -419,13 +419,29 @@ defmodule Fountain.Accounts do
   defp insert_user(attrs, :password) do
     %User{}
     |> User.registration_changeset(attrs)
+    |> put_trial_end()
     |> Repo.insert()
   end
 
   defp insert_user(attrs, :oauth) do
     %User{}
     |> User.oauth_registration_changeset(attrs)
+    |> put_trial_end()
     |> Repo.insert()
+  end
+
+  # Every account gets a trial deadline the moment it exists, before Stripe is
+  # involved at all.
+  #
+  # The trial is really driven by a Stripe subscription, and the gate reads
+  # `trial_ends_at`, treating nil as "no expiry". If the date were only written
+  # when Stripe was successfully reached, an account whose customer or
+  # subscription call failed would sit at nil and be free forever — which is
+  # exactly how 159 production accounts got there. Stripe's `trial_end`
+  # overwrites this with the authoritative value as soon as the subscription
+  # exists; until then the local date is a floor, not a guess nobody checks.
+  defp put_trial_end(changeset) do
+    Ecto.Changeset.put_change(changeset, :trial_ends_at, Fountain.Billing.trial_end_from_now())
   end
 
   defp create_user_data_key(user_id) do
--- a/apps/fountain/lib/fountain/billing.ex
+++ b/apps/fountain/lib/fountain/billing.ex
@@ -85,6 +85,27 @@ defmodule Fountain.Billing do
     end
   end
 
+  # `trialing` is only active while the trial actually has time left. The
+  # subscription created at signup means Stripe drives this and sends a webhook
+  # when the trial ends — but the whole reason trials never expired is that the
+  # expiry depended on something arriving. Checking the clock too means an
+  # undelivered webhook, a Stripe outage or a misconfigured endpoint delays
+  # revenue rather than forfeiting it.
+  #
+  # A nil `trial_ends_at` is treated as no expiry. 159 production accounts are
+  # in that state, from before a trial end was recorded at all; deciding their
+  # cutoff is a business call, not something a boolean should do silently. See
+  # Fountain.Release.expire_legacy_trials/1.
+  defp do_check_active(%User{subscription_status: "trialing", trial_ends_at: nil}), do: :ok
+
+  defp do_check_active(%User{subscription_status: "trialing", trial_ends_at: ends_at}) do
+    if DateTime.compare(DateTime.utc_now(), ends_at) == :lt do
+      :ok
+    else
+      {:error, :subscription_required}
+    end
+  end
+
   defp do_check_active(%User{subscription_status: status}) when status in @active_statuses,
     do: :ok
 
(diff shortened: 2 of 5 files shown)


jhgaylor:
## Verified end-to-end against Stripe test mode

Ran on `acct_1TVSQWDbTpLaSfbP` in test mode, with a Stripe **Test Clock** so the trial could be fast-forwarded 14 days rather than waited out. All 20 checks pass.

What it exercised, in order:

| | |
|---|---|
| registration writes `trial_ends_at` before Stripe is involved | PASS |
| `start_trial_subscription/1` creates a real subscription | PASS |
| status `trialing`, trial is exactly 14.0 days | PASS |
| our `trial_ends_at` equals Stripe's `trial_end` | PASS |
| `missing_payment_method` is `cancel` | PASS |
| access allowed while the trial is live | PASS |
| **clock advanced past trial end → Stripe set the subscription to `canceled`** | PASS |
| the real events Stripe emitted parse through `sync_subscription/1` | PASS |
| account reaches `canceled` and access is refused | PASS |
| the clock-based half refuses access with no webhook applied | PASS |

The end-of-trial behaviour is the part I most wanted to see, and it does what the code claims: a trial that reaches day 14 with no card is **cancelled**, not converted to an unpaid invoice with dunning attached.

### Two things the run turned up

**The webhook payloads are real now.** Stripe emitted three events — `customer.subscription.created`, `trial_will_end`, `deleted` — and all three went through `sync_subscription/1` cleanly. Every existing test for that function uses hand-shaped payloads, so this is the first time it has been fed what Stripe actually sends. `trial_will_end` returns `{:ok, :ignored}`, which is correct for now but see below.

**My first attempt's step 9 was wrong, and I want to be clear it was the test and not the code.** I asserted that after advancing the Test Clock, the clock-based gate would refuse access. It didn't, and it shouldn't have: a Test Clock advances *Stripe's* time, not the application's, so `trial_ends_at` was still a fortnight in the future as far as `DateTime.utc_now()` was concerned. The corrected run checks the clock half by moving the date instead, which is the only way to simulate elapsed time short of waiting.

### Worth a follow-up, not in this PR

`customer.subscription.trial_will_end` fires three days out and we ignore it. That is the natural hook for a "your trial ends in 3 days" email, which is the single cheapest conversion lever available here — 196 trialing accounts and 1 paying. Happy to file it.

Test-mode objects left in place if you want to look: customer `cus_UzlAQlWr7Ipq5S`, clock `clock_1Tzle0DbTpLaSfbPy3nzbq3x`.

Issue #148, opened 2026-07-30
Secrets after ADR 0019: broker exchange for non-HTTP secrets, references not values, and one credential in the sandbox


lex00:
Spinning out the thread from #136, where @jhgaylor wrote: "We need to revisit the whole idea of giving secret values to the sandbox." Proposal for where that lands: **the sandbox gets exactly one credential — a conversation-scoped token — and secret values never enter the sprite.**

## Premise

Sandboxes run untrusted code by design. The honest threat model is that everything readable inside the sprite is exfiltrated the moment the agent starts: env vars, files, anything substitution wrote into an MCP server's config. Encryption at rest and write-only APIs (both already good in fountain) don't help once the value is materialized as a plaintext env var at spawn — which is what `environment secrets -> vault secrets` merge does today, including `${VAR}` substitution into `mcp_servers.*.env`.

## The model

1. **One credential in the sprite:** a per-conversation token, minted at spawn, revoked at terminate, whose only audience is fountain's own conversation/callback API. (The bundled `fountain` skill already implies this concept — this makes it the *only* credential.)
2. **Values resolve outside the boundary.** Two mechanisms cover almost everything:
   - **Egress credential injection:** for secrets that are HTTP auth to an allowlisted host (GitHub tokens, most API keys), the sprite's network layer injects the header on the way out. The process inside sees the request succeed and never sees the value. Composes directly with `networking_type: limited` — the allowlist and the injection map are the same object.
   - **Broker exchange:** for everything else, the in-sprite token can be exchanged (outside, at fountain) for a short-lived, scoped derivative — never the long-lived value.
3. **Environments declare references, not values.** `{key, provider/scope}` instead of `{key, value}` in the API. This is the part that matters most for external tooling: a reference is reviewable, diffable, and driftable; a value is none of those. (Downstream consumer: the chant lexicon in INTENTIUS/chant#1217 can type and lint references for free if they're first-class in the API.)

## What this does to existing pieces

- **Vaults** keep their role (user brings credentials to a published agent — the #136 use case) but become reference bundles; the merge rule applies to references.
- **#136's allowlist** stays useful and gets stronger: it scopes *which reference bundles* may be attached, and with values gone, the worst case of an unauthorized vault drops from "secret exfiltration" to "wrong upstream account."
- **Write-only secrets storage** (already implemented) becomes the storage half of the broker rather than a UI nicety.

## Honest limits

[...]

PR #267, opened 2026-08-02
docs: propose ADR 0009 — token-only sandboxes (#148)

jhgaylor:
A **Proposed** ADR for the decision #148 asks for — nothing is committed by merging this; accepting the direction is the decision, and the open questions at the bottom of the ADR are the four things that need @jhgaylor.

The significant update since the issue was filed: [Infisical agent-vault](https://github.com/Infisical/agent-vault) now exists — an OSS credential proxy that *is* the egress-injection mechanism from the proposal (MITM `HTTPS_PROXY`, dummy-value substitution, egress filtering, per-sandbox temp tokens; its documented ephemeral-sandbox pattern is Fountain's architecture from the other side), from the vendor whose operator already materializes `fountain-secrets`. The ADR therefore proposes **operating the injection layer rather than building it**, phased:

- **Phase 0** (cheap, no behavior change): per-secret `exposure: value | injected | brokered`, default `value` — makes exposure a greppable declaration and gives the API the vocabulary.
- **Phase 1**: egress injection via agent-vault / Agent Proxy, per-conversation vault tokens minted at spawn and revoked at terminate; `limited` networking folds into the proxy's service rules.
- **Phase 2**: broker exchange for non-HTTP secrets over the existing conversation token.
- **Phase 3**: references-not-values in the API.

### DIFF
--- a/decisions/0009-token-only-sandboxes.md
+++ b/decisions/0009-token-only-sandboxes.md
@@ -1,0 +1,115 @@
+# 0009 — Token-only sandboxes: secret values stop entering the sprite
+
+**Status:** Proposed — drafted 2026-08-02 for the decision requested in #148.
+Nothing below is committed until this ADR is accepted.
+
+## Context
+
+Sandboxes run untrusted code on purpose. The honest threat model (#148) is
+that everything readable inside a sprite is exfiltrated the moment the agent
+starts — prompt injection is enough; no exploit required. Today the
+environment+vault merge materializes every secret as a plaintext env var at
+spawn, including `${VAR}` substitution into MCP server configs.
+
+What already exists bounds the damage without removing the exposure:
+
+- Write-only secret storage and envelope encryption at rest.
+- Per-conversation `FOUNTAIN_TOKEN` scoped to its owner (#75), callback-token
+  scoping and key expiry (#206), agent-scoped vault allowlists (#144).
+- Output redaction (#222) — which its own author documents as bounded: it only
+  matches values it knows, so a derived secret (base64d, split, reshaped)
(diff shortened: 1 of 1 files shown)


jhgaylor:
Closing: the ADR this asks for exists, under a different number and a narrower scope.

**[ADR 0019 — egress credential brokerage](https://github.com/BinaryBourbon/fountain/blob/main/decisions/0019-egress-credential-brokerage.md)** is merged (#690), and **gate 0 passed on 2026-08-24** (#1088). It decides Phase 1 of this issue and demonstrates it: Agent Vault as the broker behind an interface we own, a real Sprites sandbox making brokered calls to `api.github.com` and cloning a private repository while holding a 16-byte placeholder, an unmatched host refused, and a cross-tenant probe refused. Cost: +258 ms per request, ≤170 ms to mint a session. `limited` networking folding into the proxy's service rules is gate 2, and per-conversation tokens are §5.

The ADR number this issue names went to account deletion and export, so "ADR 0009" here is stale in the same way #148 and #470 are (a known pattern: issues citing ADR numbers that were never written).

Phase 0's `exposure: value | injected | brokered` survives as the better vocabulary. ADR 0019 §7/§9 says the same thing with two values ("brokerable / unbrokerable"), and gate 1 should adopt the three-valued name from here rather than inventing a third — `value` is the honest label for the deliberate escape hatch.

What this issue proposed and 0019 does **not** decide — Phase 2 (broker exchange for non-HTTP secrets) and Phase 3 (references, not values, in the API) — stays open on #148, which is being rescoped to exactly that residue.

Issue #334, opened 2026-08-02
billing: an admin trial extension can be reverted by a later webhook


jhgaylor:
`extend_trial` (`billing.ex:259-275`) sets `trialing` locally for accounts with no live Stripe trial — the "give this person another look" lever — but it does not advance `subscription_synced_at`, and it doesn't exclude the account from sync the way `comped` does (`billing.ex:554-556`).

So any later subscription event with a newer `created` timestamp overwrites the extension: a straggler from the old canceled subscription, or a re-checkout race. The operator's decision silently disappears and the user is gated again.

Narrow, but the comp path already shows the intended pattern — an operator decision should outrank an inbound event.

**Fix:** stamp `subscription_synced_at` to now on extension (so older events are stale by the existing guard), or give extension the same explicit precedence `comped` has.

Issue #312, opened 2026-08-02
billing: a webhook that fails after claiming its event id is never retried


jhgaylor:
The idempotency fix and the retry fix defeat each other.

`billing.ex:435-463` — `claim_event/2` inserts into `stripe_events` **first**, then `sync_subscription/1` runs outside any transaction. `stripe_webhook_controller.ex:66-73` returns 500 on failure so Stripe redelivers — but redelivery hits the already-claimed id and returns `{:ok, :duplicate}` → 200.

**Failure:** a transient DB error (or a changeset rejection) during the `Repo.update` in `sync_subscription`, after a successful claim. The retry that the 500 was designed to obtain becomes a no-op. If the lost event is `customer.subscription.deleted` for a trial that ended, no further events ever arrive for that subscription and the user stays `trialing`/`active` indefinitely — free service, no signal.

**Fix:** wrap claim + apply in a `Repo.transaction` so a failed apply rolls back the claim, or delete the claim row on failure before returning 500.

A test would need to force `sync_subscription` to fail after the claim, then assert the redelivery actually applies.

Found in the 2026-08-02 re-audit of `main`.

PR #349, opened 2026-08-02
billing: key webhook sync by subscription, not customer (#309)

### LINKED ISSUE #309: billing: upgrading mid-trial creates a second subscription, and the trial's cancellation then locks out a paying customer
A user who upgrades during their trial ends up with two Stripe subscriptions, and the first one's death is applied to their account.

## Chain

- `billing_live.ex:250` — the portal is only offered to `active`/`past_due`; a `trialing` user is sent to Checkout.
- Checkout in `mode: :subscription` always creates a *new* subscription.
- Nothing cancels the existing trialing subscription on `checkout.session.completed` — `billing.ex:473-497` only backfills the customer id.
- Status sync is keyed by **customer**, not subscription (`billing.ex:529-573`), so events from either subscription write the same `users` row.

## Failure

User is in trial on sub A (created with `missing_payment_method: :cancel`). They click Upgrade, pay, and get sub B — account goes `active`. At trial end Stripe cancels sub A, because no payment method was attached *to it*. `customer.subscription.deleted` for sub A arrives, is newer than sub B's last event so the staleness guard passes, `coerce_status` maps deleted → `canceled`, and the account is locked out — while sub B keeps billing monthly. Recovery only comes at sub B's next lifecycle event, up to a month later.

The other branch is no better: if the payment method becomes the customer default, sub A converts to active at trial end and the customer is billed twice.

## Fix direction

Cancel the trial subscription when checkout completes, and/or persist `stripe_subscription_id` on the user and ignore events for any other subscription. The second is the durable fix — customer-keyed sync is the root cause and will keep producing this class of bug.

Found in the 2026-08-02 re-audit of `main`.

jhgaylor:
Closes #309 (P0). **Stacked on #348** — the checkout apply relies on #348's claim rollback for safe redelivery after a failed Stripe cancel. Merge #348 first, then retarget this to `main`.

## The bug

A mid-trial upgrade created a second Stripe subscription and nothing cancelled the first. Customer-keyed sync let events from either subscription write the account: when Stripe cancelled the abandoned trial sub at trial end (`missing_payment_method: :cancel`), its `customer.subscription.deleted` passed the staleness guard and locked out a paying customer — for up to a month, until sub B's next lifecycle event. The other branch (card becomes customer default) double-billed instead.

## The fix — the durable one from the issue

- **`users.stripe_subscription_id`** (migration + schema): the subscription of record, set at trial creation, replaced when `checkout.session.completed` adopts the session's subscription.
- **Sync is subscription-keyed**: the customer locates the user; an event naming any other subscription returns `{:ok, :other_subscription}` and never touches the account.
- **Checkout completion cancels every other live subscription** on the customer — the doomed trial sub can neither die loudly nor convert and double-bill. Its status write is the optimistic copy the new subscription's own webhook confirms (that webhook may already have arrived and been ignored as unrecorded — ordering is handled both ways).
- **Legacy accounts** (nil recorded id, pre-migration) keep today's customer-keyed behavior and adopt from the first live event — but never from a deletion, so an account already in the double-subscription window still recovers via the paying subscription's next event.
- A failed Stripe cancel returns an error → webhook 500 → #348 rolls the claim and the partial writes back → redelivery retries cleanly (covered by a test).

## Tests

The headline test runs both halves on one path per the audit's paired-fixes note: upgrade adopts sub B and cancels sub A via Stripe (Mimic-verified it cancels *only* A), then A's later deletion event arrives and the account must stay active. Plus: legacy adopt, legacy deletion-doesn't-adopt, rollback-on-failed-cancel, and the trial-creation id recording. **All five verified to fail against the previous code.**

Full suite: 1542 tests, 0 failures. format / warnings-as-errors / credo --strict / dialyzer clean.

### DIFF
--- a/apps/fountain/lib/fountain/accounts/user.ex
+++ b/apps/fountain/lib/fountain/accounts/user.ex
@@ -26,6 +26,10 @@ defmodule Fountain.Accounts.User do
     field :max_concurrent_sandboxes, :integer, default: 5
     field :role, :string, default: "user"
     field :stripe_customer_id, :string
+    # The subscription of record. Webhook sync applies events for this
+    # subscription only — a customer can briefly carry two (mid-trial upgrade),
+    # and customer-keyed sync let the doomed one write the account's status.
+    field :stripe_subscription_id, :string
     field :subscription_status, :string, default: "trialing"
     field :trial_ends_at, :utc_datetime
     field :suspended_at, :utc_datetime
@@ -96,6 +100,7 @@ defmodule Fountain.Accounts.User do
     user
     |> cast(attrs, [
       :stripe_customer_id,
+      :stripe_subscription_id,
       :subscription_status,
       :trial_ends_at,
       :subscription_synced_at,
--- a/apps/fountain/lib/fountain/billing.ex
+++ b/apps/fountain/lib/fountain/billing.ex
@@ -205,6 +205,7 @@ defmodule Fountain.Billing do
       {:ok, sub} ->
         user
         |> User.billing_changeset(%{
+          stripe_subscription_id: sub.id,
           subscription_status: to_string(sub.status),
           trial_ends_at: unix_to_datetime(Map.get(sub, :trial_end)),
           subscription_synced_at: DateTime.utc_now() |> DateTime.truncate(:second)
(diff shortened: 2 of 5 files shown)


jhgaylor:
For the record: this content is already on `main`. PR #352's branch (`mrr-price-env`) was cut from the tip of this chain, so its squash-merge (`2410429`) carried #312 + #309 + #334 along with the env-var change — that's also why #348's squash landed as an empty commit. Verified on main: `stripe_subscription_id` migration + field, subscription-keyed sync, checkout adopt-and-cancel, and all the #309 tests. Staying closed as superseded.

PR #350, opened 2026-08-02
billing: an admin trial extension outranks in-flight webhooks (#334)

### LINKED ISSUE #334: billing: an admin trial extension can be reverted by a later webhook

jhgaylor:
Closes #334. **Stacked on #349** (chain: #348 → #349 → this) — the straggler test uses the subscription-of-record field #349 introduces.

`extend_trial` — the "give this person another look" lever — set `trialing` locally but never advanced `subscription_synced_at`, so a straggler event from the old cancelled subscription with a newer `created` silently reverted the operator's decision.

The extension now stamps `subscription_synced_at` to the moment the operator acted: everything older is dropped by the existing staleness guard, everything younger still applies (Stripe stays authoritative for what happens *after* the extension). Same precedence idea as `comped`, but via the ordering guard rather than a status exclusion, since trialing accounts must keep syncing.

Tests cover both directions — straggler dropped, younger event applies — and the straggler test **fails against the previous code**.

Full suite: 1544 tests, 0 failures; format / warnings-as-errors / credo --strict clean.

### DIFF
--- a/apps/fountain/lib/fountain/accounts/user.ex
+++ b/apps/fountain/lib/fountain/accounts/user.ex
@@ -26,6 +26,10 @@ defmodule Fountain.Accounts.User do
     field :max_concurrent_sandboxes, :integer, default: 5
     field :role, :string, default: "user"
     field :stripe_customer_id, :string
+    # The subscription of record. Webhook sync applies events for this
+    # subscription only — a customer can briefly carry two (mid-trial upgrade),
+    # and customer-keyed sync let the doomed one write the account's status.
+    field :stripe_subscription_id, :string
     field :subscription_status, :string, default: "trialing"
     field :trial_ends_at, :utc_datetime
     field :suspended_at, :utc_datetime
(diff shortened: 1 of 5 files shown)


jhgaylor:
Heads-up: this merge was a no-op — GitHub recorded mergeCommit `9d66201`, which is the base tip, because a rebase force-push race of mine briefly left the branch with zero commits when you clicked merge. The real change never landed. It's re-opened as a fresh PR carrying the actual commit; nothing else about it changed. Apologies for the churn.

PR #355, opened 2026-08-02
billing: an admin trial extension outranks in-flight webhooks (#334) [re-open]

jhgaylor:
Replaces #350, which GitHub marked merged with an **empty diff**: my rebase onto the post-lifecycle `main` briefly left this branch with no commits, and the merge click landed in exactly that window (mergeCommit == base tip, zero changes). The actual fix (`268a796`) only ever lived on the branch — this PR carries it for real. Sorry for the confusion; the force-push race was mine.

Original description: `extend_trial` set `trialing` locally but never advanced `subscription_synced_at`, so a straggler event from the old cancelled subscription silently reverted the operator's decision. The extension now stamps `subscription_synced_at` at the moment the operator acted — older events go stale via the existing guard, younger events still apply. Tests cover both directions; the straggler test fails against the previous code.

Stacked on #349 (chain: #348 → #349 → this → #353).

### DIFF
--- a/apps/fountain/lib/fountain/billing.ex
+++ b/apps/fountain/lib/fountain/billing.ex
@@ -248,6 +248,14 @@ defmodule Fountain.Billing do
   `canceled` or `past_due` account back to `trialing`: the support lever for
   "give this person another look".
 
+  The extension also advances `subscription_synced_at`, because an operator
+  decision outranks anything already in flight: without the stamp, a straggler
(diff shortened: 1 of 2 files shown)


jhgaylor:
Same as #349: this landed on `main` inside #352's squash (`2410429`) — the `mrr-price-env` branch was cut from this chain's tip. Verified: `extend_trial` stamps `subscription_synced_at` on main, straggler + younger-event tests included. #334 is fixed on main.

Issue #470, opened 2026-08-04
DB-backed instance settings: admin-editable runtime policy


jhgaylor:
## Motivation

Runtime operator policy still largely comes from environment variables read at boot in `config/runtime.exs`. Changing registration rules, sandbox bounds, capacity policy, or the platform inference daily ceiling therefore requires an env edit and rollout.

Move those policies into admin-editable database settings while preserving env-only self-hosted deployments. The boundary is **deployment configuration versus operator policy**; being a secret does not by itself make a value env-only.

## Current implementation and corrections to the original proposal

Reviewed against main at `0c8443b6`.

- Some instance configuration is already database-backed. `Fountain.PlatformInference` stores encrypted provider keys in `platform_inference_keys`, supports admin edits at `/admin/inference`, gives stored values precedence over env values, audits mutations, and reports the source. Platform ChatGPT credentials also have their own database lifecycle. Extend these patterns rather than introducing a second owner for the same credentials.
- `Fountain.Crypto.encrypt_platform/1` and `decrypt_platform/1` already support platform secrets. They use the master key with platform-specific authenticated data; the implementation deliberately has no platform DEK. The original proposal's new platform DEK plumbing is not a prerequisite.
- `decisions/0010-db-backed-instance-settings.md` does not exist in the current tree. ADR 0010 is now the `ee/` directory boundary. This issue is a revised proposal, not an implementation of that ADR; document the settings decision under an available ADR number when implementing it.
- Sandbox credentials now live under `:managoat_sandbox`, accessed through `Managoat.Sandbox.Config`. The original `SpritesClient.get!/0` migration target is obsolete. `Fountain.SandboxProviders` reads that same configuration to determine availability.
- The product now uses `CREDITS_ENABLED` and prepaid credits. The original references to `BILLING_ENABLED` and `stripe_price_monthly_cents` are obsolete.

## Configuration split

| Category | Storage direction | Examples |
|---|---|---|
| Deployment wiring and root credentials | Keep in env | Database connection/pool, `MASTER_SECRETS_KEY`, `SECRET_KEY_BASE`, endpoint/listeners, clustering, trusted proxies, mail transport, Stripe integration, deployment OAuth configuration, observability wiring |
| Operational policy | Database with env defaults | Registration toggle/domains; sandbox idle/lifetime/log bounds; concurrency ceilings; queue limits; platform inference daily ceiling |
[...]

PR #471, opened 2026-08-04
adr: 0010 — DB-backed instance settings, admin-editable at runtime

jhgaylor:
Records the decision behind #470: runtime operator policy — the platform `SPRITES_TOKEN`, registration toggles, sandbox lifetime/log bounds — moves to a DB-backed `instance_settings` table editable from the admin UI, with DB-over-env precedence (env stays as seed/fallback so env-only self-hosted instances are untouched), secrets encrypted under a platform DEK wrapped by `MASTER_SECRETS_KEY`, cached reads with cluster-wide PubSub bust, and audited writes.

Also adds an addendum to ADR 0005, whose "never stored in the DB, never surfaced in the admin UI" clause for the token this supersedes. 0005's core decision (one platform token, per-tenant concurrency cap) is unchanged.

Docs-only: nothing is built yet, and both files say so explicitly per the ADR template rule; the PRs that build each mechanism remove the caveats.

Closes nothing — #470 stays open to track the build.

### DIFF
--- a/decisions/0005-platform-shared-sprites-token.md
+++ b/decisions/0005-platform-shared-sprites-token.md
@@ -51,3 +51,13 @@ Infisical operator into the `fountain-secrets` Kubernetes Secret
 `envFrom`; access control is the cluster's RBAC plus Infisical project
 access. It remains env-var-only from the app's perspective — never in the
 DB, never in the UI.
+
+## Addendum — 2026-08-04
+
+The "never stored in the DB, never surfaced in the admin UI" clause is
+superseded by [0010](0010-db-backed-instance-settings.md): the token becomes a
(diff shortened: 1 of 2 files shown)


jhgaylor:
Carrying over the renumber note from tracker #482 (closed) so it isn't lost, with one correction: #482 said to renumber this ADR to 0012, but 0012 has since been claimed by the manifests-exit ADR (PR #488, tracker #487). 0010 is `ee-directory-boundary` and 0011 is `first-admin-bootstrap`, both merged.

So this ADR (and the build issue #470, whose title/body also say "ADR 0010") should take **0013** — or whatever the next free number is when it lands; check `decisions/` and open ADR PRs at write time rather than trusting this comment.

Issue #531, opened 2026-08-05
API parity: an API-only consumer can drive everything the UI can (tracker)


jhgaylor:
Tracker for the 2026-08-05 API-parity gap analysis: the UI and API have full parity on core resource CRUD (agents, environments, vaults, conversations, API keys, secrets, prompts/interrupt/terminate/turns/turn-images), but a long tail of functionality exists only in LiveViews. Goal: **an API-only consumer can register, configure, run, observe, and manage an account end-to-end** — the same promise the UI makes.

The critical path for headless bootstrap is #518 (inference credentials — conversations can't run without one) → #522 (finish email verification without a browser) → #525 (onboarding flag).

## P1

- [ ] #518 api: inference-credentials endpoints — headless bootstrap blocker
- [ ] #530 bug: API secret writes emit no audit events (LiveView writes do)

## P2

- [ ] #519 api: JSON endpoint for conversation log events
- [ ] #520 api: conversation read-model gaps — title, read state, roots_only, tree
- [ ] #521 api: password change and email change
- [ ] #522 api: complete email verification and password reset over the API
- [ ] #523 api: account data export and deletion
- [ ] #524 api: billing self-serve — usage summary, checkout/portal URLs
- [ ] #526 api: audit log endpoint

## P3

- [ ] #525 api: expose onboarding completion (or derive it)
- [ ] #527 api: admin surface for scripted operator tasks
- [ ] #528 api: agent avatars — bearer fetch, upload, delete
- [ ] #529 api: include usage counts in agent/environment/vault list responses

## Cross-cutting conventions (apply to every issue above)

- New endpoints go through the `:api` pipeline (rate limit → TenantAPIAuth → Audit) and into the OpenAPI spec/schemas — the spec is generated from the router, so a route without schemas fails the CI `openapi.spec.json` check.
- Sensitive account-level writes (credentials, password, deletion, admin) belong behind the `full`-scope gate (`RequireKeyManagement` pattern) so per-conversation sprite tokens can't escalate.
- `_unsafe_` calls only adjacent to a tenant-scoped parent fetch, per CLAUDE.md.
- Billing endpoints live in `ee/` (boundary from #472).

## Deliberately out of scope

- Theme preference over bearer token (browser concern).
- AI avatar generation over the API (UX sugar; upload/fetch/delete in #528 is the parity that matters).
- Making the emailed links themselves API-shaped — links keep pointing at browser routes; the API just accepts the same tokens (#522).

[...]

Issue #528, opened 2026-08-05
api: agent avatars — bearer fetch, upload, delete


jhgaylor:
## Problem

Agent avatars are entirely outside the API:

- Upload/delete/AI-generate exist only in `AgentsLive.Form` (`Agents.upload_avatar/3`, `delete_avatar/1`, `Fountain.AvatarGenerator`).
- Even **reading** avatar bytes requires a session: `GET /agents/:id/avatar` is in the browser scope. Turn images, by contrast, have both a session route and a bearer route — avatars should match.
- `avatar_media_type` isn't serialized in `agent_json.ex`, so an API consumer can't even tell an avatar exists.

## Proposed shape

- `GET /api/agents/:id/avatar` — bearer-token fetch, same tenant scoping + `nosniff`/CSP treatment as `AgentAvatarController`.
- `PUT /api/agents/:id/avatar` (raw body or multipart, 5 MB cap, jpeg/png/gif/webp) / `DELETE /api/agents/:id/avatar`.
- Add `avatar_media_type` (or a boolean + URL) to `agent_json.ex` and the OpenAPI schema.

## Notes

AI generation over the API is optional/out of scope — it's async UX sugar; upload/fetch/delete is the parity that matters (e.g. `fountain apply` shipping an avatar file).

Issue #522, opened 2026-08-05
api: complete email verification and password reset over the API


jhgaylor:
## Problem

An API consumer can *start* the auth email flows but never finish them:

- Email verification: `POST /api/auth/register` + `resend-verification` exist, but confirmation is `GET /users/confirm/:token` (browser pipeline; also sets a session cookie). **Account activation requires a browser round-trip.**
- Password reset: `POST /api/auth/forgot` sends the email; completion is `POST /reset` in the `:browser` pipeline only.
- Email-change confirm: `GET /account/email/confirm/:token`, browser only.

## Proposed shape

- `POST /api/auth/verify` `{token}` — verify the same `Phoenix.Token`, mark verified, enqueue `Workers.StripeCustomerSync`; return JSON instead of logging in.
- `POST /api/auth/reset` `{token, password}` — same semantics as the browser `reset/2` (session_version bump included).
- (Optional) `POST /api/auth/email/confirm` `{token}`.

## Notes

- These are public JSON endpoints (`:api_public`) authenticated by the token itself — keep the existing rate limits and anti-enumeration behavior (always-200 on forgot).
- The emailed links can keep pointing at the browser routes; the API endpoints just accept the same tokens, so a CLI can prompt "paste the code from your email".

Issue #521, opened 2026-08-05
api: password change and email change


jhgaylor:
## Problem

Password change and email change exist only as browser POSTs (`AccountSecurityController.change_password/2`, `request_email_change/2` — session auth + CSRF). There is no bearer-token equivalent; `/api/auth` stops at `token`, `register`, `resend-verification`, `forgot`, `me`, `api-keys`.

## Proposed shape

- `POST /api/auth/password` — requires current password; calls `Accounts.change_password/3`. Bumping `session_version` kills browser sessions — decide and document whether it should also revoke API keys (today it does not; that asymmetry is what makes this endpoint sensitive).
- `POST /api/auth/email` — requires current password; sends the confirmation link (same flow as the controller). Completion stays via the emailed token (see the auth-completion issue).

## Notes

- Rate-limit like the browser path (bucket `account_security`, 10/hour — `account_security_controller.ex:21-22`).
- Should probably sit behind the `full`-scope gate (`RequireKeyManagement`-style): a conversation-scoped sprite token must not be able to rotate the account password.
- Audit events to match the browser path (`auth.password.changed`).

PR #573, opened 2026-08-05
api: put the /api/auth/* surface into the OpenAPI spec (#571)

### LINKED ISSUE #571: api: the whole /api/auth/* surface is missing from the OpenAPI spec
## Problem

Every `/api/auth/*` endpoint is absent from `/api/openapi.json`. The spec is
built by `Paths.from_router/1`, which only emits a path when the route's
controller exports `open_api_operation/1` — i.e. when it `use`s
`OpenApiSpex.ControllerSpecs`. None of the auth controllers do:

| Route | Controller |
|---|---|
| `POST /api/auth/token` | `AuthTokenController` |
| `GET /api/auth/me` | `AuthMeController` |
| `GET/POST /api/auth/api-keys`, `DELETE /api/auth/api-keys/:id` | `ApiKeyController` |
| `POST /api/auth/register`, `POST /api/auth/resend-verification` | `RegistrationController` |
| `POST /api/auth/forgot`, `POST /api/auth/reset` | `PasswordResetController` |
| `POST /api/auth/verify` | `EmailVerificationController` |
| `POST /api/auth/password`, `POST /api/auth/email`, `POST /api/auth/email/confirm` | `AccountSecurityController` |

This is the gap #531 called out deliberately: #521 and #522 documented their
endpoints in `docs/api.md` rather than half-converting three controllers that
also serve browser HTML.

The effect is that the one thing a generated client needs *first* — how to get
a bearer token — is the one thing the spec does not describe. `/api/docs`
opens on a surface whose front door is invisible.

## Proposed shape

- `use OpenApiSpex.ControllerSpecs` on all seven controllers.
- Declare `operation/2` for the JSON actions; declare `operation(:action, false)`
  for the browser actions in the four mixed controllers (the precedent is
  `AgentAvatarController`'s raw-bytes `:show`) so `Paths.from_router` skips them
  and no `IO.warn` fires at spec-build time.
- The four public endpoints must carry `security: []` — the spec sets a global
  `bearer` requirement, and `POST /api/auth/token` cannot require the thing it
  issues.
- Schemas for the request/response bodies, following `FountainWeb.Schemas`
  convention.

## Guard

Add a test that walks `Router.__routes__()` and asserts every `/api/*` route
(minus the deliberate non-JSON exceptions) resolves to an operation in
`ApiSpec.spec().paths`. Without it the gap re-opens the next time someone adds
a controller.

jhgaylor:
Closes #571.

## The gap

`FountainWeb.ApiSpec` builds its paths with `Paths.from_router/1`, which only
emits a path when the route's controller exports `open_api_operation/1` — i.e.
when it `use`s `OpenApiSpex.ControllerSpecs`. None of the seven auth
controllers did, so all thirteen `/api/auth/*` routes were missing from
`/api/openapi.json`. Nothing failed. The endpoints simply were not there.

That left the published spec describing every resource endpoint but not the
one thing a client needs first: how to get a bearer token. `/api/docs` opened
on a surface with no visible front door, and any generated client had to
hand-roll authentication against `docs/api.md`.

This is the gap #531 flagged deliberately — #521 and #522 documented their
endpoints in prose rather than half-converting three controllers that also
serve browser HTML.

## What this does

| Route | Controller |
|---|---|
| `POST /api/auth/token` | AuthTokenController |
| `GET /api/auth/me` | AuthMeController |
| `GET`/`POST /api/auth/api-keys`, `DELETE /api/auth/api-keys/{id}` | ApiKeyController |
| `POST /api/auth/register`, `POST /api/auth/resend-verification` | RegistrationController |
| `POST /api/auth/forgot`, `POST /api/auth/reset` | PasswordResetController |
| `POST /api/auth/verify` | EmailVerificationController |
| `POST /api/auth/password`, `POST /api/auth/email`, `POST /api/auth/email/confirm` | AccountSecurityController |

Four of those controllers also serve browser HTML (`/auth/register`,
`/auth/reset/:token`, `/users/confirm/:token`, the account form POSTs). Adding
`ControllerSpecs` makes `open_api_operation/1` exist for *every* action in the
module, so each browser action is declared `operation(:x, false)` — the
treatment #528 already gave `AgentAvatarController`'s raw-bytes `:show`. That
makes `Paths.from_router` skip them instead of emitting an `IO.warn` at
spec-build time. No browser path leaks into the spec; a test asserts it.

**`security: []` on the seven public endpoints.** The spec sets a global
bearer requirement. `POST /api/auth/token` cannot require the credential it
exists to issue, and neither can register/verify/forgot/reset/resend/
`email/confirm` — the caller of the last one may hold no credential for the
account at all. Without the override a generated client refuses to call them.

Error bodies here carry a machine-readable reason beside the prose (`expired`
vs `invalid_token` vs `invalid_current_password`), which the resource
endpoints' plain `Schemas.Error` does not, so they get their own
`Schemas.AuthError`. Descriptions state the behaviour a client will otherwise
discover the hard way: `token` mints a *new* key on every call, `verify` is
idempotent, `password` does not revoke API keys, `email` is not an
availability oracle.

No behaviour changes — no `CastAndValidate` was added, so every controller
keeps its own hand-rolled param handling and its own responses.

## Guard

`ApiSpecTest` walks `Router.__routes__()` and fails on any `/api/` route with
no operation, so the next controller added without a spec fails a test rather
than going missing quietly. Five routes are explicit exceptions, each with a
reason:

- `/api/openapi.json`, `/api/docs` — the spec machinery, plugs not controllers
- `/api/stripe/webhook` — authenticated by `Stripe-Signature`; its caller does not read our spec
- `PATCH /api/settings/theme` — a CSRF-protected browser-session route that happens to live under `/api/`
- the turn-image bytes — **a real remaining gap**, not machinery: one action
  serves both the bearer route and the browser route, so speccing it means
  splitting it into `:show` / `:api_show` the way #528 split avatars. Left out
  as scope creep; say the word and I will file it.

A second test fails if an exception stops matching a live route, so the list
cannot rot into permanent blanket coverage.

`SchemasTest` gains the same drift check the Conversation schema has, over
the API-key views — including an assertion that no field which could carry
key material is in the listing schema.

## Verification

- `mix precommit` green: 2242 tests, 0 failures
- `mix openapi.spec.json` regenerates and validates; **48 paths → 60**
- All seven public auth operations confirmed carrying `security: []` in the
  generated JSON; `/api/auth/me`, `/api/auth/api-keys` and `/api/auth/password`
  confirmed still inheriting the bearer requirement

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,19 @@ upgrade, is in
   in another tab or on a phone — with no second login. It cannot be camped on:
   a verified user hitting it is sent where they were going (#533)
 
+- **The `/api/auth/*` endpoints are now in the OpenAPI spec.** They never
+  were: the spec is generated from the router, and a controller that does
+  not declare operations is skipped in silence — so the published spec
+  described every resource endpoint but not the one thing a client needs
+  first, which is how to get a bearer token. `/api/docs` opened on a surface
+  whose front door was invisible, and generated clients had to hand-roll
+  authentication. All thirteen auth routes are documented now, with the seven
+  public ones (`token`, `register`, `resend-verification`, `verify`, `forgot`,
+  `reset`, `email/confirm`) declaring `security: []` so a generated client
+  will actually call them without a credential it cannot yet have. A test
+  walks the router and fails on any `/api/` route without an operation, so
+  the gap cannot silently re-open (#571)
+
 - **Secrets written through the API now leave the same audit trail as
   secrets written through the UI.** `POST/DELETE /api/environments/:id/secrets`,
   the vault equivalents, and the secret half of `POST /api/apply` recorded
--- a/apps/fountain/lib/fountain_web/controllers/account_security_controller.ex
+++ b/apps/fountain/lib/fountain_web/controllers/account_security_controller.ex
@@ -14,8 +14,10 @@ defmodule FountainWeb.AccountSecurityController do
   """
 
   use FountainWeb, :controller
+  use OpenApiSpex.ControllerSpecs
 
   alias Fountain.Accounts
+  alias FountainWeb.Schemas
 
   plug FountainWeb.Plugs.RateLimit,
        [bucket: "account_security", max: 10, window_ms: 3_600_000]
@@ -25,6 +27,13 @@ defmodule FountainWeb.AccountSecurityController do
               :api_change_password,
               :api_request_email_change
             ]
+
+  tags(["Auth"])
+
+  # Browser form POSTs and the emailed confirmation link — not API routes.
+  operation(:change_password, false)
+  operation(:request_email_change, false)
+  operation(:confirm_email_change, false)
 
   def change_password(conn, %{"current_password" => current, "new_password" => new}) do
     user = conn.assigns.current_user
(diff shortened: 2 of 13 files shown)


jhgaylor:
CI failed on the first run — **not this PR's code**. `AdminUserDetailLiveTest`
"a Stripe failure degrades to an error note" died on `#invoices` returning no
element. Root-caused and fixed in 2d8a627, now part of this PR; filed as #576.

`#invoices` is gated on `@billing_enabled`, which reads
`Application.get_env(:fountain, :billing_enabled, true)` — global to the VM,
not to a test. `admin_controller_test.exs` flips it to `false` for its
"billing actions when billing is disabled" block **and is `async: true`**, so
every concurrently running async test that reads `Billing.enabled?/0` saw
`false` for the duration. Introduced by 1c9d012 (#570) — the last child of
#531, so it is in scope for closing that tracker.

Every other module that mutates this env (`admin_billing_disabled`,
`account_deletion_live`, `auth_me_controller`, `marketing_pricing`,
`marketing_legal_unpublished`, `accounts/deletion`, `ee/billing_invoices`) is
already `async: false`. This one was the outlier.

Proved the mechanism rather than assuming it: a throwaway pair of async
modules mirroring that setup block fails an unrelated async
`assert Billing.enabled?()` when the flipper is `async: true`, and passes when
it is `async: false`. The natural interleaving is too narrow to reproduce on
demand — five runs of the two real files together were clean — which is why
this surfaced as an unexplained intermittent rather than a reliable failure.

This also explains the single unreproducible failure I noted in the #575
description; same cause.

Issue #526, opened 2026-08-05
api: audit log endpoint


jhgaylor:
## Problem

The audit log is viewable only at `/audit` (`AuditLive.Index`): last 200 events, admin sees all via `Audit._unsafe_list_recent/1`, users see their own via `Audit.list_recent_for_user/2`. There is no `/api/audit` endpoint — the only programmatic access is buried in the account export.

## Proposed shape

`GET /api/audit` — tenant-scoped (`list_recent_for_user/2`), with `limit` + cursor pagination and ideally filters (action prefix, resource_type, time range) that the LiveView doesn't have yet either.

## Notes

- Admin cross-tenant listing belongs to the admin-API issue, not here.
- Columns to serialize: `inserted_at, actor, action, resource_type, resource_id, metadata, request_ip` (what the UI shows).
- Once this exists, consider giving AuditLive the same filters.

PR #575, opened 2026-08-05
ui: give the audit page the filters #526 gave the API (#572)

### LINKED ISSUE #572: AuditLive: give the browser trail the filters GET /api/audit got in #526
## Problem

#526 gave `GET /api/audit` filters — `action_prefix`, `resource_type`,
`since`, `until` — plus cursor pagination, all implemented in
`Audit.list_for_user/2`. `AuditLive.Index` still calls the two unfiltered
helpers and renders a fixed last-200 table:

```elixir
defp load_events(%{role: "admin"}), do: Audit._unsafe_list_recent(200)
defp load_events(%{id: id}), do: Audit.list_recent_for_user(id, 200)
```

So the API is now strictly better than the UI at the thing the UI exists for.
"Show me every `vault.` event since Tuesday" is a curl away and impossible in
a browser — you scroll 200 rows and hope.

## Proposed shape

- A filter form above the table: action prefix, resource type, since, until.
- Filters live in the query string (`handle_params`), so a filtered view is a
  shareable link and survives a refresh — the same reason the API takes them
  as query params.
- The admin (cross-tenant) path needs the same filters, which means
  `Audit._unsafe_list_recent/1` has to accept the same options as
  `list_for_user/2` rather than just a limit. Share one filter pipeline
  between them.
- Keep the 5s poll; re-run it with the active filters.

## Notes

- Cursor pagination is out of scope — the API needs it for scripted export,
  the UI's use is "look at what just happened". A "no matches" empty state and
  a visible active-filter summary matter more.
- `resource_type` should be a select populated from the distinct values in the
  visible trail rather than a free-text box, if that is cheap.

jhgaylor:
Closes #572. **Stacked on #573** — base is `api/571-auth-openapi-spec`, so
merge that first and GitHub will retarget this to `main`. The two touch
disjoint code; they are stacked only to keep the `[Unreleased]` CHANGELOG
section conflict-free.

## The gap

#526 gave `GET /api/audit` filters — `action_prefix`, `resource_type`,
`since`, `until` — all implemented in `Audit.list_for_user/2`. `AuditLive.Index`
kept calling the two unfiltered helpers and rendering a fixed last-200 table:

```elixir
defp load_events(%{role: "admin"}), do: Audit._unsafe_list_recent(200)
defp load_events(%{id: id}), do: Audit.list_recent_for_user(id, 200)
```

So the API ended up strictly better than the UI at the one thing the UI exists
for. "Show me every `vault.` event since Tuesday" was a curl away and
impossible in a browser.

## What this does

The page takes the same four filters, through the same query. Filter state
lives in the URL (`?action=`, `?resource=`, `?since=`, `?until=`), which makes
a filtered view a link you can send someone, and makes the 5s poll re-run with
the active filters instead of resetting to everything.

**Context.** `list_for_user/2`'s ordering, page size and five filters move
into a shared private `recent_query/2`; a new `_unsafe_list_events/1` runs the
same builder unscoped. The admin half of the page needed the filters too —
an admin sees every tenant's events, so before this the person looking at the
most events could filter the least. Sharing one builder is what keeps the
browser trail and `GET /api/audit` answering the same question the same way.
`_unsafe_list_recent/1` stays, now delegating, so the admin API and the pruner
tests are untouched.

`list_resource_types_for_user/1` and `_unsafe_list_resource_types/0` back the
resource-type select — the issue asked for a select of what is actually in the
trail rather than a free-text box, if cheap, and a `distinct` on an
append-only table is cheap.

**UI details worth a look in review:**

- "No events match these filters" and "No events yet" are distinct empty
  states. One means narrow differently; the other means nothing has happened.
- Time bounds are `datetime-local`, read as UTC because that is what the table
  renders. An unparseable value filters nothing **and stays in the box** —
  a half-typed date must not silently vanish while the results ignore it.
- A footer appears only when the page is full, pointing at `GET /api/audit`
  for anything past the newest 200.

**Out of scope, deliberately:** cursor pagination. The API needs it for
scripted export; the browser's use is "look at what just happened", and a
visible active-filter state plus a real empty state matter more there. Called
out in the issue.

## Verification

Tests cover filter-by-URL, form-change-pushes-to-URL, filters surviving the
tick, both empty states, the half-typed date, tenant-scoping of the
resource-type select, and the admin cross-tenant filtered view. Context-level
coverage for `_unsafe_list_events/1` includes the LIKE-metacharacter escape
(`?action=%` must match nothing, not everything) and inclusive time bounds.

Each of the three load-bearing tests was checked against the bug it targets:

| Reverted to | Test that fails |
|---|---|
| admin path calling `_unsafe_list_recent/1` | "an admin gets the same filters over the cross-tenant trail" |
| `:tick` reloading with empty filters | "filters survive the 5s tick" |
| non-admins calling `_unsafe_list_resource_types/0` | "the resource-type select offers only this tenant's types" |

`mix precommit` green: 2261 tests, 0 failures. One earlier run had a single
failure with `Postgrex.Protocol ... DBConnection.ConnectionError: client
exited` in the log — a sandbox-ownership flake, not reproducible; three
subsequent full runs are clean and the audit tests pass in isolation every
time.

The pre-existing `AuditLiveFormatTsTest` Mimic stub moved from
`list_recent_for_user/2` to `list_for_user/2`, since that is what the page
calls now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,18 @@ upgrade, is in
 
 ### Added
 
+- **The `/audit` page has the filters the API got in #526.** `GET /api/audit`
+  could narrow the trail by action prefix, resource type and time window; the
+  page could not, so the API was strictly better than the UI at the one thing
+  the UI is for — "show me every `vault.` event since Tuesday" was a curl away
+  and impossible in a browser, where you scrolled 200 rows and hoped. The
+  page now takes the same four filters through the same query, with the
+  resource-type list built from what is actually in your trail. Filter state
+  lives in the URL, so a filtered view is a link you can send someone and it
+  survives the 5s refresh. Admins get the filters over the cross-tenant view
+  too — previously the person seeing the most events could filter the least
+  (#572)
+
 - **`/api/admin/*` makes operator tasks scriptable** — list and inspect
   accounts with the filters the admin UI has, set the sandbox cap, extend a
   trial, comp, suspend, resync from Stripe, delete an account, list and reap
--- a/apps/fountain/lib/fountain/audit.ex
+++ b/apps/fountain/lib/fountain/audit.ex
@@ -152,8 +152,33 @@ defmodule Fountain.Audit do
   """
   @spec list_for_user(Ecto.UUID.t(), keyword()) :: [Event.t()]
   def list_for_user(user_id, opts \\ []) when is_binary(user_id) do
-    from(e in Event,
-      where: e.user_id == ^user_id,
+    from(e in Event, where: e.user_id == ^user_id)
+    |> recent_query(opts)
+    |> Repo.all()
+  end
+
+  @doc """
+  The distinct `resource_type` values present in one tenant's trail, sorted.
+
+  Populates the resource-type filter on `/audit` — a select of what is
+  actually there beats a free-text box the user has to guess into.
+  """
+  @spec list_resource_types_for_user(Ecto.UUID.t()) :: [String.t()]
+  def list_resource_types_for_user(user_id) when is_binary(user_id) do
+    Repo.all(
+      from e in Event,
+        where: e.user_id == ^user_id and not is_nil(e.resource_type),
+        distinct: true,
+        order_by: e.resource_type,
+        select: e.resource_type
+    )
+  end
+
+  # Ordering, page size and the five filters, shared by the tenant-scoped and
+  # unscoped listings so the two cannot drift — the browser trail and
+  # GET /api/audit answer the same question the same way.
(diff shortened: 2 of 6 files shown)


jhgaylor:
Rebased onto the updated base so this branch also carries the #576 flake fix (2d8a627). Head is now 87eaa03; base and mergeability intact.

That fix supersedes the caveat in the description above: the "one earlier run had a single failure with `Postgrex.Protocol ... client exited`" was the same `admin_controller_test` global-billing-env race, not a sandbox-ownership flake. Root cause and the reproduction are in #573's comment and #576.

Issue #552, opened 2026-08-05
Structural: converge auditing at the context layer; semantic events; retire dead code


jhgaylor:
Part of #540 — the cleanup that keeps this class of gap from reopening.

1. **Mixed-caller functions**: `Accounts.suspend_user`/`unsuspend_user` (`accounts.ex:868,895`), `update_user_role` (`:824`), `update_sandbox_limit` (`:835`), `create_api_key` (`:539`), `revoke_api_key`, `register_user` (`:93`), `upsert_oauth_user` (`:467`) all rely on each caller to audit — some do (admin LiveView via `record_admin`, ueberauth controller), some don't. Move the `audit_events` write into the context function; callers add caller-specific context (e.g. admin actor) on top.
2. **API plug records intent, not outcome**: `plugs/audit.ex:46` logs `"PUT /api/agents/:id"` + status, no semantic action or changed fields. Once contexts emit semantic events (#540 children), decide whether the plug remains as a request log or gets scoped down to avoid double-writing.
3. **Dead code**: `Audit.record!/1` (`audit.ex:52`) has zero call sites — use it where audit loss should fail the operation (e.g. account deletion?) or delete it.
4. **Guardrail**: add a test or credo check enumerating context mutation functions that must audit, so new write paths fail loudly instead of silently joining the gap list.
5. **CLAUDE.md / docs**: document the rule — "mutations audit in the context, not the caller" — once it's true (per the ADR-accuracy rule from the 2026-07 audit, not before).

Issue #540, opened 2026-08-05
Tracker: audit-trail coverage — every meaningful write path should leave an audit event


jhgaylor:
## Context

A 2026-08-05 sweep compared every write path (Repo.insert/update/delete across contexts, controllers, LiveViews, and workers) against actual `Fountain.Audit.record/1` call sites. Result: auditing is web-layer and pipeline-scoped, not context-level, so coverage depends on which door a mutation came through.

There are only four `Audit.record/1` call sites in the tree:

- `FountainWeb.Plugs.Audit` (`apps/fountain/lib/fountain_web/plugs/audit.ex:46`) — blanket coverage, but **only** on the `:api` pipeline
- `FountainWeb.Audited.from_conn/from_socket` (`apps/fountain/lib/fountain_web/audited.ex:40,57`) — auth controllers + 3 LiveViews (API keys, env secrets, vault secrets)
- `Fountain.Exports` (`apps/fountain/lib/fountain/exports.ex:107`)
- `Fountain.Accounts.Deletion` (`apps/fountain/lib/fountain/accounts/deletion.ex:85`)

`Accounts`, `Agents`, `Environments`, `Vaults`, `Conversations`, `InferenceCredentials`, and all of `ee/` billing contain **zero** audit calls — every caller must remember, and many don't. The dominant asymmetry: resource CRUD is audited via `/api` (by the plug) but silent via LiveView — the exact inverse of the secrets gap that `audited.ex` documents as previously fixed.

The pattern to converge on is the `Exports`/`Deletion` one: **audit inside the context function** so UI, API, and system callers are covered automatically.

## Children

- [ ] #542 — **P1** `POST /api/auth/token` mints a full-scope API key with no audit entry
- [ ] #543 — **P1** Resource CRUD (agents, environments, vaults) unaudited from the UI — audit in the contexts
- [ ] #544 — **P1** Registration and logout produce no audit events on any surface
- [ ] #545 — **P2** Conversation lifecycle actions unaudited from the UI
- [ ] #546 — **P2** Inference credential writes never audited on any surface
- [ ] #550 — **P2** Billing/subscription state changes write no `audit_events`
- [ ] #551 — **P2** System-actor writes unaudited; retention pruner deletes audit rows silently
- [ ] #552 — **P3** Structural: converge auditing at the context layer; semantic events; dead code; guardrail

## Suggested order

1. #542 (quick, security-relevant), then #543 (establishes the context-level pattern the rest reuse)
[...]

Issue #530, opened 2026-08-05
bug: API secret writes emit no audit events (LiveView writes do)


jhgaylor:
## Problem

Secret writes are audited from the LiveView but **not** from the API:

- `EnvironmentsLive.Form` emits `environment.secret.write` / `environment.secret.delete` via `FountainWeb.Audited` (`environments_live/form.ex:163-207`); `VaultsLive.Form` likewise emits `vault.secret.write` / `vault.secret.delete`.
- `SecretController` (`secret_controller.ex:46-86`) and `VaultSecretController` perform the same writes with **no audit events** — only the generic pipeline-level `Plugs.Audit` request record.

Secrets are the most sensitive resource in the system; the audit trail currently shows who wrote a secret only if they used a browser. The account export's `audit_trail` and `/audit` both under-report API-driven secret activity.

## Fix

Emit the same `*.secret.write` / `*.secret.delete` events from the two API controllers (or push the audit call down into the context so both surfaces get it for free — preferable, one less pair of call sites to drift).

`Audit.record/1` is best-effort by design — keep it non-blocking per CLAUDE.md.

PR #582, opened 2026-08-05
audit: mint API keys through one audited context function (#542)

### LINKED ISSUE #542: POST /api/auth/token mints a full-scope API key with no audit entry
Part of #540.

`POST /api/auth/token` (`apps/fountain/lib/fountain_web/controllers/auth_token_controller.ex:35`) creates a full-scope API key via `Accounts.create_api_key` (`accounts.ex:539`), but the route sits on the `:api_public` pipeline (`router.ex:8-12`), which does **not** include `FountainWeb.Plugs.Audit`.

The two other surfaces for the same action both audit:
- `api_keys_live/index.ex:24` records `api_key.created`
- `POST /api/auth/api-keys` is covered by the `:api` pipeline plug

So the one path that exchanges credentials for a key — the most attack-relevant one — is the only one that leaves no trail.

**Fix:** record `api_key.created` for this path (ideally by moving the audit into `Accounts.create_api_key` itself so all three surfaces share it — see the structural child of #540).

**Acceptance:** minting a key via `POST /api/auth/token` writes an `audit_events` row identifying the user and key; a regression test asserts it.

jhgaylor:
Closes #542. First child of #540.

## The gap

`POST /api/auth/token` sits on the `:api_public` pipeline, which carries no audit plug. So the one path that exchanges a password for a full-scope API key — the most attack-relevant of the four — was the only one that minted with no trail:

| Surface | Before | After |
|---|---|---|
| `POST /api/auth/token` | **nothing** | `api_key.created`, actor `api` |
| `POST /api/auth/api-keys` | `:api` plug's request log only | request log **+** `api_key.created` |
| Settings LiveView | `api_key.created` via `from_socket` | `api_key.created` from the context |
| Sprite callback rotation | nothing | `api_key.created`, actor `system:conversation_server` |

## The approach

Rather than bolt a fourth call site onto the third surface, the audit moves into `Accounts.create_api_key/3` — every mint already goes through it. This is the context-level pattern #540 converges the rest of the campaign on, prototyped here on the smallest surface that has one.

Callers supply only what the context cannot know: a new `FountainWeb.Audited.attribution/2` returns the `:actor` / `:request_ip` pair for a `Plug.Conn` or a LiveView socket. `/api/auth/token` overrides the actor to `"api"` — with no auth plug on `:api_public` nothing has assigned `:current_user`, and derivation would file a CLI sign-in under `"system"`. The same caveat `from_conn/4` already documents for login.

The LiveView's own `from_socket` call goes with it, or every UI mint would land twice.

Metadata carries name, scopes and the 8-character prefix — the handle that matches a trail row against a listed key — and never the key itself.

## Two consequences worth naming

- **API mints now write two rows.** The `:api` pipeline plug still logs `POST /api/auth/api-keys` + status alongside the semantic event. Deliberate: keeping the plug untouched means no route can silently lose coverage while the campaign is in flight. #552 decides its fate once the other contexts emit semantic events.
- **The factory audits now.** `insert_api_key` mints through the same function, so any test setting up a bearer key has an `api_key.created` row in its trail. One test was counting a hardcoded 5 events; it now counts the trail, which is what it meant. Two Mimic stubs moved from arity 2 to arity 3.

## Tests

Six new cases in `audit_coverage_test.exs`:

- `/api/auth/token` records the mint — user, key id, actor, scopes, prefix (the acceptance criterion)
- the raw key never reaches the trail, with a guard-the-guard assert so it can't pass vacuously
- a rejected credential mints nothing and records nothing
- `/api/auth/api-keys` records a semantic event, not just the request
- one UI mint writes **one** row, not two
- a sprite callback key is attributed to the server, not the owner

`mix precommit` exits 0 locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,18 @@ upgrade, is in
 
 ### Fixed
 
+- **Minting an API key always leaves an audit trail now, whichever door you
+  came through.** There are four ways to get a key, and `POST /api/auth/token`
+  — the one that exchanges a password for a full-scope key, and the most
+  attack-relevant of them — was the only one that minted silently, because it
+  runs on a public pipeline that carries no audit plug. Anyone auditing "who
+  issued a key and when" saw the UI and `POST /api/auth/api-keys` but not the
+  CLI login door. The audit moves into `Accounts.create_api_key/3`, which
+  every mint already goes through, so UI, API, CLI and the per-conversation
+  callback rotation are covered by construction and a future surface gets it
+  for free. Events carry the key's name, scopes and public prefix — enough to
+  match a trail row to a listed key — and never the key (#542)
+
 - **Email verification is now enforced where identity is established, not at
   each door.** #533 moved unverified logins onto a waiting page but left the
   check inside the LiveView hook, so it held only because every entry point
--- a/apps/fountain/lib/fountain/accounts.ex
+++ b/apps/fountain/lib/fountain/accounts.ex
@@ -23,6 +23,7 @@ defmodule Fountain.Accounts do
   require Logger
   alias Fountain.Repo
   alias Fountain.Accounts.{User, ApiKey, UserDataKey, OauthIdentity}
+  alias Fountain.Audit
   alias Fountain.Crypto
 
   ## Users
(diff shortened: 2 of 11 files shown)


jhgaylor:
Added `57598fa` — a CHANGELOG entry for this change. The branch had none, and the directly comparable audit fix (#530, audit events for API secret writes) carries one; nothing in CI enforces it, so a security-relevant behaviour change would otherwise ship unannounced.

I also rebased the branch onto current `main` — it was cut on `7cc045f` and was three merges behind (`c53fd26`, `c71ad7d`, `42d221a`). `36e6e1e` is your commit unchanged, just replayed onto the new base.

Context: I ran into this work uncommitted in the shared working tree during an unrelated rebase, stashed it with a label and restored it intact rather than discarding it. `mix precommit` exits 0 on the rebased branch — 2299 tests, 0 failures. If you were mid-flight on this, the branch is yours; nothing of yours was modified.

Issue #197, opened 2026-08-01
conversations: 'completed' status is unreachable and sandboxes.exit_code is never written


jhgaylor:
Two pieces of modelled-but-unwritten state.

**1. Conversations never reach `completed`.** The status enum includes it (`conversation.ex:13`) and three query sites filter on it (`conversations.ex:75`, `conversations_live/index.ex:197`, `conversations_live/show.ex:321`), but nothing ever writes `status: "completed"` to a conversation — only to turns (`conversation_server.ex:786`). Conversations sit in `idle` indefinitely. Any filter or count keyed on `completed` is silently always empty.

**2. `sandboxes.exit_code` is never written.** The column and its changeset cast exist (`sandbox.ex:16,28`) with no writer, so sandbox exit information is lost.

Minor related cleanup: `handle_call({:send_prompt, prompt}, ...)` (arity-2 clause, `conversation_server.ex:675`) is dead — the 3-arg clause at `:665` always matches because `send_prompt/3` defaults `images \\ []`.

### Fix direction
Either write these states or remove them. If `completed` is meaningful, define what completes a conversation (terminal turn? explicit close? idle timeout — see #167) and write it; otherwise drop it from the enum and the filters.

---
_Filed from the 2026-07-31 post-launch gap audit._

PR #601, opened 2026-08-06
api: derive the OpenAPI schemas from the Ecto schemas — types only, not enums (#599)

### LINKED ISSUE #599: Derive the OpenAPI schemas from the Ecto schemas — 12 enum lists are duplicated and can drift silently
`FountainWeb.Schemas` is a second, hand-maintained copy of the domain model. 1,979 lines across 98 modules restate what the Ecto schemas and changesets already declare — and the two copies can drift silently, because a stale spec still validates, it just documents and enforces the wrong thing.

## The duplication is literal

Twelve enum lists appear verbatim in both an Ecto schema and `schemas.ex`. A sample, with the source of truth on the left:

| Source of truth | Restated in |
|---|---|
| `conversations/conversation.ex:22` `@statuses` | `schemas.ex:52`, `schemas.ex:94` |
| `agents/agent.ex:11` `@runtimes` | `schemas.ex:49`, `:286`, `:392`, `:459` |
| `conversations/sandbox.ex:12` `@statuses` | `schemas.ex:23` |
| `environments/environment.ex:11` `@networking` | `schemas.ex:539`, `:607`, `:648` |

The full set of duplicated lists:

```
~w(pending running idle failed terminated)              conversation status
~w(pending starting ready terminated failed)            sandbox status
~w(pending running completed failed interrupted)        turn status
~w(pending completed failed)                            export status
~w(claude codex gemini opencode)                        agent runtime
~w(ui api agent)                                        conversation source
~w(output stage)                                        log event kind
~w(started done failed interrupted)                     log event state
~w(unrestricted limited)                                environment networking
~w(user admin)                                          user role
~w(trialing active past_due canceled comped)            subscription status
~w(anthropic_api_key claude_code_oauth_token
   openai_api_key gemini_api_key)                       inference credential kind
```

`agent.runtime` alone is written out in five places, and the environment networking enum leaks into `live/environments_live/form.ex:529` as a third copy.

Worth noting: the source modules **already expose accessors for exactly this** — `Agent.runtimes/0`, `Conversation.statuses/0`, `Conversation.sources/0`, `Sandbox.statuses/0`, `Turn.statuses/0`, `LogEvent.kinds/0` / `streams/0` / `states/0`, `Export.statuses/0`. Nothing in the web layer calls any of them.

## Why this matters more than the line count

The failure mode is silent. Add a status to `@statuses`, and:

- the changeset accepts it,
- the API happily returns it,
- the OpenAPI spec says it can't exist,
- `mix openapi.spec.json` still passes and CI stays green,
- any generated client with a strict enum decoder breaks on real production data.

This is the same class as the `completed` status documented in `conversation.ex:13-23` — a value three query sites filtered on that nothing ever wrote. Divergence between a declared model and the real one is not caught by anything we currently run.

It also isn't confined to `schemas.ex`. The `operation(...)` decls inside controllers are another ~1,103 lines of the same restatement, and the 15 JSON views are ~471 more. Roughly **3,550 lines — about 17% of the web layer — is contract restatement**, against ~1,400 lines for all four primitives' actual domain logic.

## Proposal

Derive the mechanical parts from the Ecto schema; keep the human parts hand-written.

**Derivable:** field names, primitive types, `format: :uuid` / `:"date-time"`, `nullable`, `required` (from `validate_required`), and enums (from `validate_inclusion` or the existing accessor functions).

**Not derivable, must stay authored:** descriptions and examples. These carry real knowledge that lives nowhere else — e.g. `schemas.ex:43`, *"Generated from the first turn; null until one exists."* Any solution that loses these makes the spec worse, not better.

Rough shape:

```elixir
defmodule FountainWeb.Schemas.Conversation do
  use FountainWeb.DerivedSchema,
    source: Fountain.Conversations.Conversation,
    title: "Conversation",
    description: "One chat with one agent inside one sandbox.",
    # only the fields the API exposes; virtuals and associations opt in
    expose: [:id, :title, :sandbox_id, :agent_id, :vault_id, :runtime,
             :status, :runtime_session_id, :source, :parent_conversation_id,
             :turn_count, :last_active_at],
    docs: %{
      title: "Generated from the first turn; null until one exists.",
      turn_count: "Number of turns; virtual, populated on list reads."
    }
end
```

### Open questions to settle before building

1. **Compile-time coupling.** Reading enums from `Agent.runtimes/0` inside a schema macro creates a compile-time dependency from the web layer onto every domain schema. Both live in the `:fountain` app, so it's legal, but it will widen recompilation fan-out. Worth measuring on a touch-one-schema rebuild before committing.
2. **API shape ≠ schema shape.** Several exposed fields are virtual (`turn_count`, `last_active_at`) or renamed/nested (`sandbox` as an embedded object), and secrets are deliberately never exposed. Derivation must be opt-in per field, not "every field in the schema" — a bug here leaks data rather than just documenting it wrong.
3. **`required` semantics differ.** `validate_required` is about writes; OpenAPI `required` on a response is about what the client can rely on. They aren't the same list and shouldn't be assumed equal.
4. **Scope.** `schemas.ex` is the clean win. Whether the controller `operation(...)` decls follow is a separate call — their params are route- and filter-specific, not schema-derived, so the payoff is smaller.

### A cheaper first step

If the macro turns out not to earn its keep, a **guard test** captures most of the value for a fraction of the effort: walk the `@statuses` / `@runtimes` style attributes on the domain schemas, walk the corresponding OpenAPI schemas, and fail when an enum in one is missing from the other. That closes the silent-drift hole immediately and is worth doing regardless of whether the derivation lands — same pattern as the router-walking spec guard added in #531.

Suggested order: guard test first, then evaluate derivation with the drift risk already contained.

jhgaylor:
**Stacked on #600** — base is `api/599-enum-guardrail`, review that first. Together they close #599.

`FountainWeb.DerivedSchema` reads field names, types, `format: :uuid` / `date-time`, array item shapes and virtual-field types from `__schema__/1,2`, so a renamed or retyped column fails the build instead of publishing a spec that no longer describes the response.

```elixir
defmodule Sandbox do
  use FountainWeb.DerivedSchema,
    source: Fountain.Conversations.Sandbox,
    title: "Sandbox",
    description: "One sprite lifespan owned by a conversation.",
    expose: [:id, :sprite_name, {:status, enum: ~w(pending starting ready terminated failed)}],
    required: [:id, :sprite_name, :status]
end
```

Fourteen schemas derive: `Sandbox`, `Conversation`, `ConversationTreeNode`, `Turn`, `Environment`, `Vault`, `Secret`, `VaultSecret`, `Export`, `LogEvent`, `AdminUser`, `AdminSandbox`, `AuditEvent`, `ApiKey`.

## Enums are deliberately NOT derived

This started out deriving enums from `Conversation.statuses/0` and friends. That was reverted in the second commit, because measuring it showed it removed the check that motivated #599 in the first place.

With the list read from the domain module, the published spec grows a value on its own and `SchemaEnumGuardrailTest` compares that value against itself. Injecting the same drift both ways:

| Adding `archived` to `Conversation.@statuses` | enums derived | enums literal (this PR) |
|---|---|---|
| Guardrail test | passes, 0 failures | **fails**, naming both affected schemas |
| Published spec | **silently gains `archived`** | unchanged |

So the 27 enum literals stay in `schemas.ex` where #600 can check them. `enum:` takes a literal list and raises on a function reference, naming the guardrail such a reference would defeat.

The duplication is deliberate. Duplication was never the problem — *undetected divergence* was, and #600 covers that for a fraction of the cost. Enum membership is a contract decision that should cost someone a deliberate edit; a column's type is not.

## Also not derived

- **Nullability** lives in the migration, not the schema module.
- **`required`** — `validate_required` is about writes; OpenAPI `required` is a guarantee about a *response*.
- **Descriptions** carry knowledge that exists nowhere else. `doc:` keeps them.
- **Field exposure is opt-in.** `expose:` lists fields one by one rather than defaulting to every column — getting that backwards leaks data rather than mis-documenting it. `key_hash` is absent by construction.

`InferenceCredentialStatus` is not derived at all: `Credential` is one row per tenant with a ciphertext column per provider, so there is no `provider` field. It's a projection, and stays hand-written.

## Verification

**The published spec is byte-identical** — 129,065 bytes, 95 schemas, zero diff against a snapshot taken before the stack. Checked after every batch via a script that fails loudly if compilation or spec generation fails, so a stale artefact can't pass as an unchanged spec.

Both behaviours proved by injection, not assumed:

```
$ sed -i 's/field :sprite_name/field :sprite_nom/' .../sandbox.ex && mix compile
** (ArgumentError) Fountain.Conversations.Sandbox has no field :sprite_name
   — add it to the schema, or declare the property with `schema:`
```

and the guardrail fails again on an added status, with the spec unmoved.

`derived_schema_test.exs` adds 20 tests over the mapping rules and the failure modes — unknown field, unmappable column type (`:binary` ciphertext columns are deliberately unmapped rather than guessed), unknown field option, and both forms of accessor-backed enum.

Gate: `format --check-formatted`, `compile --warnings-as-errors`, `credo --strict` (no issues), `dialyzer` (0 errors). Full suite **2,396 tests, 0 failures**.

## Honest accounting

`schemas.ex` goes 1,979 → 1,919; `derived_schema.ex` adds 202. **Net production code is up ~142 lines.** #599 guessed "a few hundred lines" saved — that was wrong, and it's worth recording why: descriptions, not field declarations, are the bulk of that file, and descriptions are exactly what must survive derivation.

The value here is a compile-time failure on column rename/retype across 14 schemas. If that isn't worth ~142 lines and a small DSL to you, closing this and keeping #600 alone is a defensible call — #600 covers the drift that actually bit us (#197), and this is the narrower, rarer failure.

Two follow-ups worth considering separately:
- The ~26 `XResponse` / `XListResponse` wrappers (`%{data: array of X}`) are near-pure boilerplate and a better brevity target than this was.
- Only 14 of 95 schemas derive, so the file now carries two idioms. Worth either converting the rest or accepting the split deliberately.

### DIFF
--- a/apps/fountain/lib/fountain_web/derived_schema.ex
+++ b/apps/fountain/lib/fountain_web/derived_schema.ex
@@ -3,9 +3,10 @@ defmodule FountainWeb.DerivedSchema do
   Build an OpenAPI schema from the Ecto schema it describes (issue #599).
 
   `FountainWeb.Schemas` restated the domain model by hand: field names,
-  types, `format: :uuid`, `date-time`, and twelve enum lists copied verbatim
-  from an Ecto schema. All of that is already declared once, in the schema
-  module, so this derives it instead.
+  types, `format: :uuid`, and `date-time`, all of which are already declared
+  once in the schema module. This derives them instead, so a renamed or
+  retyped column fails the build rather than publishing a spec that no longer
+  describes the response.
 
       defmodule Turn do
         use FountainWeb.DerivedSchema,
@@ -40,12 +41,28 @@ defmodule FountainWeb.DerivedSchema do
     * **descriptions** — `doc:` carries knowledge that exists nowhere else
       and must survive derivation. Losing it would make the spec worse.
 
+  ## Why enums are NOT derived
+
+  `enum:` takes a literal list. Reading it from `Conversation.statuses/0`
+  was tried and reverted, because it quietly removed the check that
+  motivated #599 in the first place.
+
+  With the list read from the domain module, adding a status makes the
+  published spec grow a value on its own, and
+  `FountainWeb.SchemaEnumGuardrailTest` compares that value against itself
+  and passes. Measured on the branch that did it: the same injected drift
+  that fails loudly with literal enums produced a green suite and a silently
+  expanded API contract.
+
+  So the duplication here is deliberate. Duplication was never the problem —
+  *undetected divergence* was, and the guardrail test covers that at a
+  fraction of the cost. Enum membership is a contract decision that should
+  cost someone a deliberate edit; a column's type is not.
+
   ## Field options
 
-    * `enum: :statuses` — call `source.statuses/0` for the values. This is
-      the point of the exercise: the list stays in the Ecto schema.
-    * `enum: {Module, :fun}` — when the list belongs to another module, e.g.
-      a conversation's runtime vocabulary is owned by `Agent`.
+    * `enum: ~w(a b c)` — a literal list, checked against the domain list by
+      the guardrail test. See "Why enums are NOT derived" above.
     * `nullable: true`
     * `doc: "..."` — the property description.
     * `pattern: "..."`, `format: :uuid`, `example: ...`
@@ -105,7 +122,7 @@ defmodule FountainWeb.DerivedSchema do
   defp property(source, name, field_opts) do
     case Keyword.fetch(field_opts, :schema) do
       {:ok, %Schema{} = override} -> override
-      :error -> source |> derive_type!(name) |> decorate(source, field_opts)
+      :error -> source |> derive_type!(name) |> decorate(field_opts)
     end
   end
 
@@ -151,9 +168,9 @@ defmodule FountainWeb.DerivedSchema do
   defp type_schema(_), do: nil
 
   # Order matters only for readability; each clause sets an independent field.
-  defp decorate(%Schema{} = schema, source, field_opts) do
+  defp decorate(%Schema{} = schema, field_opts) do
     Enum.reduce(field_opts, schema, fn
-      {:enum, ref}, acc -> %{acc | enum: enum_values(source, ref)}
+      {:enum, values}, acc -> %{acc | enum: enum_literal!(values)}
       {:nullable, value}, acc -> %{acc | nullable: value}
       {:doc, text}, acc -> %{acc | description: text}
       {:pattern, p}, acc -> %{acc | pattern: p}
@@ -167,10 +184,14 @@ defmodule FountainWeb.DerivedSchema do
     end)
   end
 
-  defp enum_values(source, fun) when is_atom(fun), do: enum_values(source, {source, fun})
+  # A literal list, never a call into the domain module — see the moduledoc.
+  defp enum_literal!(values) when is_list(values), do: values
 
-  defp enum_values(_source, {mod, fun}) do
-    mod |> apply(fun, []) |> Enum.map(&to_string/1)
+  defp enum_literal!(other) do
+    raise ArgumentError,
+          "enum: takes a literal list of values, got #{inspect(other)}. " <>
+            "Reading the list from the domain module would make " <>
+            "FountainWeb.SchemaEnumGuardrailTest compare it against itself."
   end
 
   defp maybe_put(map, _key, nil), do: map
--- a/apps/fountain/lib/fountain_web/derived_schema.ex
+++ b/apps/fountain/lib/fountain_web/derived_schema.ex
@@ -1,0 +1,181 @@
+defmodule FountainWeb.DerivedSchema do
+  @moduledoc """
+  Build an OpenAPI schema from the Ecto schema it describes (issue #599).
+
+  `FountainWeb.Schemas` restated the domain model by hand: field names,
+  types, `format: :uuid`, `date-time`, and twelve enum lists copied verbatim
+  from an Ecto schema. All of that is already declared once, in the schema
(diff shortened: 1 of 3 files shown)


jhgaylor:
Closing in favour of #600 alone.

Building this was worth it for what it settled, but the result does not earn its keep:

- **It partly defeats #600.** Deriving enums makes `SchemaEnumGuardrailTest` compare a list against itself. Measured: the same injected drift that fails loudly with literal enums produced a green suite and a silently expanded API contract. The types-only version avoids that, but only by giving up most of what the PR was for.
- **Net +142 lines** for a compile-time failure on column rename/retype — narrower and rarer than the enum drift that actually bit us in #197, which #600 already covers.
- **Two idioms** in one file (14 derived, 81 hand-written) is worse to read than either pure approach.

Settled along the way, worth recording:

- Compile-time fan-out was the open question in #599 and is a **non-issue** — changing an Ecto schema recompiles 2 files vs 1.
- The duplication was **real**: derivation reproduced the spec byte-identically (129,065 bytes, 95 schemas), so the hand-written schemas were pure restatement.
- The "few hundred lines saved" estimate in #599 was **wrong**. Descriptions, not field declarations, are the bulk of `schemas.ex`, and they must survive derivation.

Cheaper alternative if the rename case is worth closing later: extend the emitted-vs-documented test in `schemas_test.exs` (currently Conversation and ApiKey only) to every resource. Catches renames for exposed fields with no new concepts.

Branch `api/599-derived-schema` is left in place if anyone wants the history.

PR #619, opened 2026-08-07
ci: split the test suite and the static gates into parallel jobs

jhgaylor:
## What

One Elixir CI job becomes two: `Build and Test` (the suite, alone) and `Static analysis and release checks` (everything else), running in parallel.

## Why

On [run 31149752997](https://github.com/BinaryBourbon/fountain/actions/runs/31149752997) — a representative green PR at 6m52s — the single job spent:

| | |
|---|---|
| `mix coveralls` | **4m26s** |
| runner setup + containers + caches | 45s |
| dialyzer (warm PLT) | 25s |
| release boot + health probes | 20s |
| Go setup + CLI test/vet/gofmt | 12s |
| compile, credo, format, sobelow, deps.unlock, hex.audit, OpenAPI | ~25s |

Everything below the first row shares nothing with the suite but the checkout, and was queued around it. Expected after the split: `test` ~5m20s on the critical path, `checks` ~2m15s beside it.

## What is not changed

The step inventory is identical — no gate added, removed, weakened or reordered within its job. Verified by diffing the parsed step lists; the only count changes are the shared setup steps (checkout, setup-beam, caches, `deps.get`, `ecto.create`) now appearing in both jobs.

`checks` carries its own Postgres service because the release boot probe needs one: `/health/ready` calls `Fountain.Health.database/0`, and `expire_legacy_trials` queries users.

## Cache keys

Now job-scoped (`-mix-test-` / `-mix-checks-`, `-build-test-` / `-build-checks-`). The jobs compile different trees — `test` only builds `:test`, `checks` also builds `:dev` for dialyzer and `:prod` for the release — so a shared key would have them racing to save mutually incomplete archives under it, winner decided by whichever finished first. That is the same non-determinism the existing narrow `restore-keys` were written to avoid.

**The first run after this merges is a cold cache for both jobs** (~9-11m, matching the historical cold-cache runs on lockfile bumps). Steady state from the second run on.

## What this does not fix

The suite's own floor. It is 252s of ExUnit, of which **112s is the `async: false` tail** that ExUnit runs serially on one core after the async phase drains — measured by comparing CI (4 cores: 140.7s async / 111.6s sync) against a 10-core machine (50.9s async / **77.7s sync**). The sync phase does not move with core count, so no larger runner crosses it.

Splitting the suite across machines with `mix test --partition` is the only lever that does, and it needs the 85% coverage gate to learn to merge partial results first (excoveralls has no cross-machine merge; Elixir's built-in `--export-coverage` + `mix test.coverage` does). Worth its own issue.

Also measured and ruled out: cover instrumentation is free — 128.8s with, 128.6s without, same machine. Running bare on PRs would buy nothing and cost the gate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,18 @@ concurrency:
 permissions:
   contents: read
 
+# Two Elixir jobs, not one. The test suite is ~4m30s of the ~6m50s a PR used
+# to take, and every other gate — format, credo, sobelow, dialyzer, the Go CLI
+# checks, the release boot probe — was queued behind or ahead of it in the same
+# job while sharing nothing with it but the checkout. Splitting them puts the
+# suite alone on the critical path (~5m20s) with the rest running beside it
+# (~2m15s).
+#
+# What this does NOT fix: the suite itself is 252s of which 112s is the
+# `async: false` tail, which ExUnit runs serially on one core after the async
+# phase drains. No runner size crosses that floor — only splitting the suite
+# across machines (`mix test --partition`) does, and that needs the 85%
+# coverage gate to learn how to merge partial results first. Tracked separately.
 jobs:
   test:
     name: Build and Test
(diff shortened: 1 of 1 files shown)


jhgaylor:
## Measured, not estimated

Ran the workflow twice on this branch — once cold (the state this PR lands in) and once warm (steady state from the second run on).

| Job | Cold | Warm |
|---|---|---|
| Build and Test | 315s | **313s** |
| Static analysis and release checks | 401s | **102s** |
| Compose fresh-clone check | 8s | 6s |
| Compose boots the pinned image | 25s | 23s |
| **Workflow total** | **6m41s** | **5m13s** |

Against the 6m52s baseline ([run 31149752997](https://github.com/BinaryBourbon/fountain/actions/runs/31149752997)), warm steady state is **1m39s faster** — better than the ~5m20s the PR body predicted, because `checks` came in at 1m42s rather than ~2m15s.

Worth being explicit about the cold column: at 6m41s it is barely better than baseline, because `checks` had to compile `:dev` for dialyzer and `:prod` for the release against the new job-scoped cache keys. That is a one-time cost on the run that merges this, not a regression.

All four jobs green on both runs.

The suite's own floor is unchanged and now tracked in #620 — partitioning across 3 jobs would take the test step ~4m26s → ~1m35s, but the 85% coverage gate has to learn to merge partial results first, and that is the bulk of the work rather than the YAML.

Issue #649, opened 2026-08-10
A reclaimed sandbox silently loses the agent's memory of the conversation


jhgaylor:
Found while running 0014 gate 2's live verification. **Pre-existing, not ACP-specific** — it reproduces on the legacy path exactly as it does on the ACP one.

## What happens

`Lifecycle` reclaims a sandbox after 60 minutes idle (or 24 hours absolute). The next prompt goes through `wake_conversation`, which takes the `:create_new` branch and provisions a sprite with a **new** name (`conversations.ex:1218`). The runtime's session lives in the *old* sandbox's filesystem, so it is gone.

Verified against live sprites on 2026-08-10 — create a session on sprite A, destroy A, provision a fresh sprite B, resume by the same id:

- legacy: `claude --resume <id>` → `No conversation found with session ID: <id>`, `error_during_execution`
- ACP: `session/resume` → `-32002 Resource not found: <id>`

The environment checkpoint doesn't cover it: `create_checkpoint/2` runs during *provisioning*, before any turn, so the snapshot cannot contain a session file written later.

## Why it's worse than it looks

The failure is silent and asymmetric. Fountain's own transcript is intact — every `log_events` row still renders — so the conversation **looks** continuous in the UI. The agent simply has no memory of any of it and answers the next prompt cold.

And we tell the user the opposite, in `Lifecycle.explain/1`:

> "Sandbox reclaimed after 60 minutes idle. Send another prompt to continue the conversation — history is preserved."

`Lifecycle`'s moduledoc makes the same promise as a design argument:

> "So the cost of reclaiming early is a re-provision on the next prompt, not lost work. That asymmetry is why the defaults can be fairly aggressive."

The asymmetry is real for **cost** and false for **continuity**. Being wrong in the reclaiming direction does lose work.

## Options, roughly in order of cost

1. **Tell the truth.** Change `explain/1` and the docstring to say the agent starts fresh. Cheap, honest, and stops the promise being load-bearing for other decisions. Doesn't fix anything.
2. **Checkpoint at reclaim, restore at wake.** `Sprites.create_checkpoint` already exists and is already wired for environments; the reclaim path would take one at teardown and `wake_conversation` would restore it. Turns reclaim into genuine suspend/resume. Interacts with checkpoint storage cost, which needs a number before committing.
[...]

PR #653, opened 2026-08-10
fix: read the checkpoint id from the listing, not from the stream (#652)

### LINKED ISSUE #652: Environment checkpoints are created but never recorded, so restore never happens
Found while measuring whether checkpoint-at-reclaim could fix #649. It can't be evaluated yet, because the existing checkpoint feature does not work.

## What happens

`Provisioning.create_checkpoint/2` reduces the library's stream through `extract_checkpoint_id/1`:

```elixir
defp extract_checkpoint_id(%{"checkpoint_id" => id}) when is_binary(id), do: id
defp extract_checkpoint_id(%{checkpoint_id: id}) when is_binary(id), do: id
defp extract_checkpoint_id(%{"id" => id}) when is_binary(id), do: id
defp extract_checkpoint_id(_), do: nil
```

The stream does not contain any of those shapes. Measured against a live sprite today — this is the **entire** stream from `Sprites.create_checkpoint/2`:

```
%Sprites.StreamMessage{type: "info", data: "Creating checkpoint...", error: nil}
%Sprites.StreamMessage{type: "info", data: "Checkpoint created successfully", error: nil}
%Sprites.StreamMessage{type: "info", data: "\nCheckpoint Details:", error: nil}
%Sprites.StreamMessage{type: "info", data: "  ID: v1", error: nil}
%Sprites.StreamMessage{type: "info", data: "  Created: 2026-08-10 19:42:00", error: nil}
%Sprites.StreamMessage{type: "info", data: "  Path: checkpoints/v1", error: nil}
%Sprites.StreamMessage{type: "info", data: "\nTo restore this checkpoint:", error: nil}
%Sprites.StreamMessage{type: "info", data: "  sprite checkpoint restore v1", error: nil}
%Sprites.StreamMessage{type: "info", data: "  curl -X POST /checkpoints/v1/restore", error: nil}
%Sprites.StreamMessage{type: "complete", data: "Checkpoint v1 created successfully", error: nil}
```

It's a struct with `:type`, `:data` and `:error`, and the id is *prose inside `data`*. A `StreamMessage` matches none of the four clauses, so the reduce returns `nil` for every checkpoint ever taken. The guard that follows —

```elixir
if is_binary(checkpoint_id) and checkpoint_id != "" do
```

— then declines to write it, so `environments.checkpoint_id` stays `NULL` forever. With it `NULL`, `restore_checkpoint/2` short-circuits on its own `nil` clause and every provision falls back to the full path.

The checkpoint itself is real and does get created — `list_checkpoints/1` shows it as `%Sprites.Checkpoint{id: "v1", comment: "probe"}`. We simply throw the id away.

## Consequence

Every conversation on an environment re-runs full provisioning — apt/npm installs, repo clones, setup scripts — that a restore was supposed to skip. Nothing errors, nothing is logged, and the telemetry span reports success. It is a silent, permanent loss of the optimisation.

**Confirm before fixing:** `select count(*) from environments where checkpoint_id is not null;` against prod should be 0.

## Why it survived

There is no test for it. `create_checkpoint` appears in the test suite only as a stub in `ConversationServerCase`, and `Sprites.StreamMessage` appears nowhere in the repo at all — so nothing was ever written against the shape the library actually returns. Same class as #620's format gate and `verify-gate-coverage`: a green suite over a path production never takes.

## Fix

Don't parse prose. After the stream drains, `list_checkpoints/1` returns typed `%Sprites.Checkpoint{id: ...}` structs — take the newest, or match on `comment` which we already set. Two things worth doing alongside:

- a test that feeds `extract`/the new reader the **real** message structs above, not hand-made maps;
- make a checkpoint that yields no usable id `log`/telemetry a warning rather than silently returning `nil`, since that is the state that hid this.

Note `Sprites.Checkpoint` and `Sprites.StreamMessage` do not implement `Access`, so `c["id"]` raises — it has to be `c.id`.

jhgaylor:
Fixes #652, found while trying to cost #649's real repair.

## The bug

`create_checkpoint/2` reduced the creation stream through an extractor matching maps with `"checkpoint_id"`/`"id"` keys. The library yields `%Sprites.StreamMessage{type:, data:, error:}` structs, and the id is **prose inside `data`**:

```
%Sprites.StreamMessage{type: "info", data: "  ID: v1", error: nil}
```

Nothing matched, so every checkpoint ever taken resolved to `nil`, `environments.checkpoint_id` was never written, and `restore_checkpoint/2` was never called with a real id. Every conversation on an environment re-ran the full provision — packages, clones, setup script — that a checkpoint existed to skip.

Correcting my own issue text: it is **not** silent. It logs `checkpoint create stream finished without a checkpoint_id` and reports `outcome: :no_id` on the telemetry span. What it lacks is any signal that this is the *permanent* state rather than an occasional miss.

## The fix

Ask `list_checkpoints/1`, which returns typed `%Sprites.Checkpoint{}` structs, instead of parsing prose. Two traps in that list, both now covered:

- it includes a synthetic **`"Current"`** entry for the live filesystem — always the newest thing there, and it restores nothing;
- `Sprites.Checkpoint` doesn't implement `Access`, so `c["id"]` raises.

The stream is drained before listing, because the checkpoint isn't on the server until it completes.

## Why it survived

There was no test. `create_checkpoint` appeared in the suite only as a permissive stub in `ConversationServerCase`, and `Sprites.StreamMessage` appeared **nowhere in the repo at all** — so nothing was ever written against the shape the library returns. Same class as the format gate in `verify-gate-coverage`: green over a path production never takes. Every fixture in the new test is copied verbatim from a live run for that reason.

## Before merging

Worth confirming the blast radius:

```sql
select count(*) from environments where checkpoint_id is not null;
```

Expected 0. If it's non-zero, some other path is writing it and my reading is incomplete.

`mix precommit` green: 2559 tests, 0 failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/provisioning.ex
+++ b/apps/fountain/lib/fountain/conversations/provisioning.ex
@@ -105,9 +105,10 @@ defmodule Fountain.Conversations.Provisioning do
              label: "checkpoint create"
            ) do
         {:ok, stream} ->
-          checkpoint_id =
-            stream
-            |> Enum.reduce(nil, fn msg, acc -> extract_checkpoint_id(msg) || acc end)
+          # Drain first: the checkpoint is not on the server until the stream
+          # completes, so listing before this races the upload.
+          Stream.run(stream)
+          checkpoint_id = newest_checkpoint_id(sprite, "aod env #{env.name}")
 
           if is_binary(checkpoint_id) and checkpoint_id != "" do
             {:ok, _} =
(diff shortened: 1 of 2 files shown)


jhgaylor:
Blast radius confirmed against prod:

```
select count(*) as total, count(checkpoint_id) as with_ckpt from environments;
 18 | 0
```

18 environments, **zero** with a checkpoint id. The feature has never recorded one, so there is no other path writing it and nothing to migrate — this only turns `nil` into a real id going forward.

Issue #709, opened 2026-08-14
tracker: ACP in front of Fountain — a conversation reachable from an editor (ADR 0015)


jhgaylor:
Tracker for **ADR 0015** (#631) — the other ACP direction. #644 made Fountain
an ACP *client* of the runtimes it runs. This campaign makes a Fountain
conversation reachable from an editor, with Fountain as the ACP *agent*.

Together they make Fountain a proxy: the same block vocabulary arriving from a
sprite on one side and leaving for an editor on the other.

## Why now: #644 turned the hard part into forwarding

ADR 0015 was gated on normalization happening server-side, because the only
other way to emit `session/update` was to reimplement four proprietary dialect
parsers in Go. That gate is paid:

| What the ADR needed | What exists today |
|---|---|
| ACP blocks, not raw dialect | `peer.ex:221` persists every `session/update` verbatim under the `acp` log stream |
| a way to read only those | `?streams=acp,stage` — `conversation_controller.ex:561` |
| a real stop reason | `finish_acp_turn/4` publishes the sprite's ACP stop reason on the terminal `turn`/`done` stage (`conversation_server.ex:1278-1285`) |
| resumable transport, auth, distribution | `stream.go` (reconnect, `Last-Event-ID`, idle watchdog, terminal classification), `fountain auth login`, `cli-release-go.yml` |

So the adapter **forwards blocks it does not have to understand**. That is the
best case ADR 0015 predicted, and it is the whole reason this is a smaller
project than #644 was.

## What is not free

Three things, and none of them are protocol plumbing:

1. **Which Fountain agent?** `session/new` carries a `cwd`, not an agent
   (#699). The unit an editor reaches for is a provisioned Fountain agent —
   that is four of the five reasons to use Fountain from an editor at all.
2. **Three session identities.** The ACP session id (ours), the conversation
   id (Fountain's), and `runtime_session_id` (the sprite's, and meaningless
   outside the sandbox that holds it — #649). Conflating any two is a bug.
3. **Gemini has no `acp` stream** (#702). It is still on the legacy path
   (#659/#660), so an editor pointed at a gemini-backed agent sees nothing
   renderable.

## Gates

**Gate 1 — prove the shape** ✅ done (transcript-only, one runtime, in a real editor)
- [x] #698 — `fountain acp`: stdio JSON-RPC, `initialize`, `authenticate`
- [x] #699 — `session/new`: choose an agent, mint an editor-facing session id
[...]

Issue #704, opened 2026-08-14
acp: session/cancel, and survive a dropped SSE mid-turn


jhgaylor:
Gate 2 of **ADR 0015** (#631) — and per the ADR this is a gate, not polish:
*"reattaching to a turn that kept running is the entire pitch"*.

## Two halves

**Cancel.** `session/cancel` -> `POST /api/conversations/:id/interrupt`, and
the in-flight `session/prompt` returns the `cancelled` stop reason. Note the
server already treats an ACP `cancelled` stop reason as a **failed** turn
(`conversation_server.ex:1279`) — make sure a deliberate editor-side cancel
does not read as a crash in the transcript or the audit trail.

**Survive a dropped stream.** The turn runs in `ConversationServer`, not in
the adapter. A dropped SSE connection, a redeploy, a Horde rebalance, a laptop
lid — none of them should end the editor's turn. `followStream` in
`cli/internal/cmd/stream.go` already does this: reconnect with backoff, resume
from `Last-Event-ID`, an idle watchdog, and a terminal event as the only
clean end (the `io.EOF`-as-success bug is documented in that file at length —
do not reintroduce it behind the ACP adapter).

## What is genuinely new

The editor is a *stateful* client in a way `fountain run` is not. Decide:

- What the editor sees during a reconnect gap. Silence, or a status update?
- What happens if the reconnect fails permanently mid-turn — the turn keeps
  running server-side, so the stop reason we return is a lie either way. Say
  which lie, and say it in the transcript.
- Whether `session/prompt` can outlive the editor's patience. There is no
  timeout on a Fountain turn by design.

## Done when

Kill the network mid-turn, bring it back: the editor picks up the output it
missed and the turn ends with the right stop reason. Cancel from the editor
stops the turn, and the transcript says it was cancelled, not that it failed.

Part of #709.

Issue #703, opened 2026-08-14
acp: session/load — replay the transcript from `Last-Event-ID: 0`


jhgaylor:
Gate 2 of **ADR 0015** (#631).

## What it is

An editor reopens; the conversation is still there. `session/load` **MUST**
replay the entire conversation to the client as `session/update`
notifications before responding — and Fountain already holds that history:

```
GET /api/conversations/:id/stream?streams=acp,stage
Last-Event-ID: 0
```

replays every stored `session/update` line in order. The adapter forwards the
replay, then answers the request.

## The mirror of a bug we already paid for

Fountain sits on **both** sides of this now, and the failure modes are
opposite:

- Downward (`peer.ex`), `session/load` replay is the *unhappy* path — we
  discard it, because persisting it would duplicate the transcript into the
  database on every turn (#657, gemini answering before replaying).
- Upward, replay is the *only* path, and the adapter is the thing whose
  replay a client will trust.

So: answer `session/load` **after** the replay is flushed, not before. The
`streamHistory`-shaped bug that cost us #657/#661 is exactly the one we are
now in a position to commit ourselves.

## Also

- A loaded session must be resumable for *prompting*, not just viewing —
  `session/prompt` on a loaded session goes to the same conversation.
- Decide what an editor does with a conversation whose sandbox is suspended
  (0017) or reclaimed. A reclaimed one loses the runtime's memory even though
  the transcript renders in full (#649) — an editor showing a complete history
  over an agent that remembers none of it is the quiet failure that issue is
  about.

## Done when

Closing and reopening the editor puts the full transcript back in front of the
developer, and the next prompt continues the same conversation.

Part of #709.

Issue #644, opened 2026-08-08
tracker: ACP for runtime I/O — stop owning a stream parser per coding agent (ADR 0014)


jhgaylor:
Tracker for **ADR 0014** (#630). The goal is not "adopt a protocol" — it is to
stop owning N implementations for N coding agents.

## The N we own today, honestly counted

| Per-runtime today | After ACP |
|---|---|
| **24 `event_blocks/2` clauses**, four dialects, in a LiveView render path (`show.ex:1340-1530`) | one translation, at the sandbox boundary |
| `pair_tool_results/1` — exists only because 3 of 4 dialects emit a tool call and its result as unrelated events | ACP threads them on one `id` |
| **4 different permission-bypass flags** | `session/request_permission` |
| `mode` + `runtime_session_id` resume logic — **2 of the 4 resume by guessing** (`gemini.ex:14-17`, `codex.ex:24-26`) | explicit `sessionId` |
| **4 MCP config mechanisms** — a shell-out loop, an argv allow-list, and two others | one `session/new` `mcpServers` param |
| tool-level OTel spans for **1 of 4** runtimes (`StreamTracer` is claude-only) | one tracer, every runtime |

**What does not collapse, and is expected not to:** `default_env/2`
(credential env var names differ), `skills_root/0` + `skills_sh_agent/0`
(skills layout), model selection (ACP carries none), launch argv, and codex's
`codex login --with-api-key`. ADR 0014 keeps `Fountain.Runtimes` as the
**provisioning** layer on purpose.

Net: the per-runtime surface goes from *argv + credentials + config + skills +
a bespoke stream parser + resume semantics* to *argv + credentials + skills
path*. The parser — the largest, most fragile part, and the one sitting in the
render path where a vendor's point release becomes our rendering bug — goes to
zero.

## Sequencing

**Note a deliberate departure from the ADR.** As written, 0014 calls
permissions "the gate that justifies the project". If the justification is
deleting per-runtime code, permissions belong *after* the conversions rather
than gating them, and the ADR should be amended to match.

**Gate 1 — decide**
- [ ] #632 — confirm the support mechanism and version floor per runtime

**Gate 2 — prove it once**
- [ ] #633 — a JSON-RPC peer over the sprite stdio pipe
- [ ] #634 — run one runtime over ACP behind a per-agent flag
- [ ] #635 — the reconnect story: `session/load`, or reject the runtime

#635 is the most likely reason for this campaign to fail, which is why it is
[...]

PR #712, opened 2026-08-15
feat: session/prompt runs a turn and streams it to the editor (#701)

### LINKED ISSUE #701: acp: session/prompt — forward the `acp` stream verbatim, block on the terminal stage
Gate 1 of **ADR 0015** (#631) — the payoff, and the reason this campaign
waited for #644.

## Why this is nearly free

`Fountain.Runtimes.ACP.Peer` persists every `session/update` **verbatim** to
`log_events` under the stream name `"acp"` (`peer.ex:221`), and the SSE
endpoint filters by stream (`conversation_controller.ex:561`). So

```
GET /api/conversations/:id/stream?streams=acp,stage
```

is already an ACP-native feed. The adapter forwards each `acp` line to the
editor as the `session/update` notification it already is. It does not parse
blocks, and per ADR 0015 it must never learn how — the four dialects stop at
the server boundary, which is the only place that knows which runtime produced
them.

## The two halves

1. **Prompt.** `session/prompt` -> `POST /api/conversations/:id/prompts`. ACP
   prompt content is a block list, not a string; decide what non-text blocks do
   (the API takes images, so at least those map).
2. **Block until terminal, with a real stop reason.** ACP requires
   `session/prompt` to block and return a stop reason. Fountain's API is
   post-then-stream, so the adapter correlates the two — the same correlation
   `stream.go` already performs to give `fountain run` a truthful exit code
   (#398).

   The stop reason does not have to be invented: `finish_acp_turn/4` publishes
   the sprite's own ACP stop reason on the terminal `turn`/`done` stage event
   (`conversation_server.ex:1278-1285`). Forward it. `provision`/`setup`
   failures and `errSandboxExpired` are the cases with no ACP equivalent —
   map them explicitly rather than defaulting to `end_turn`, which would tell
   an editor the agent finished when the sandbox died.

## Watch for

- **Feed the stream head, not the history.** `streamHead` in `conv.go` exists
  because a replayed `turn`/`done` from a *prior* turn ended the stream before
  the current one started (#398). The same trap is here.
- Anything on stdout that is not JSON-RPC corrupts the session.

## Done when

A prompt typed in an editor runs on a Fountain agent and streams back message
chunks, thoughts and tool calls with status — and the call returns the turn's
real stop reason. That is gate 1 complete: the shape, proven end to end.

Part of #709.

jhgaylor:
Closes #701. Gate 1 of **ADR 0015** is complete with this: an editor can hold a
conversation with a Fountain agent. Tracked by #709.

## The adapter forwards; it does not parse

`Peer` already stores every `session/update` the sprite's adapter emits,
verbatim, on a log stream named `acp` (`peer.ex:221`), and the SSE endpoint
filters by stream. So `?streams=acp,stage` is an ACP-native feed, and
`session/prompt` re-emits each stored line as the notification it already is.

That is what #644 bought, and it is why **no dialect parser lands in `cli/`** —
the rule ADR 0015 exists to enforce.

One field has to change on the way through. The stored update carries the
*sprite's* session id: the runtime's own, meaningless outside the sandbox that
holds it, and gone after a reclaim (#649). The editor knows the conversation by
ours, so that is what leaves the process; everything else passes through
untouched, and a test asserts both halves.

## Stop reasons come from the sprite; failures stay failures

The server publishes the sprite adapter's own ACP stop reason on the terminal
`turn`/`done` stage event, so:

- `refusal` comes back as a **stop reason**, not an error — even though the
  server records that turn as failed. An editor should render "the agent
  declined", not "something broke".
- An interrupt becomes `cancelled`; a suspended sandbox ends the turn cleanly,
  because the next prompt resumes it.
- The cases ACP has no vocabulary for — a sandbox that never provisioned, a
  reclaim mid-turn, a terminated conversation — are returned as **errors**
  rather than dressed up as `end_turn`, which is the one reading that would be
  false.
- A lost stream says *the turn may still be running* and names the
  conversation. It is true, and the editor cannot see it.

## The follow loop moved rather than being copied

`internal/stream` now owns the reconnect/resume/idle loop, because this adapter
needs exactly what `fountain run` needs. It encodes #398 — treating `io.EOF` as
success once made the CLI report a finished turn while the agent was still
working — and a second copy here would be a second chance to reintroduce that,
in the surface where it is hardest to notice: an editor showing a conversation
that quietly stopped updating.

`fountain run` keeps its exact behaviour, and its tests still cover the loop
through `followStream`. The loop takes a context now, so a follow ends when the
editor goes away instead of reading a stream nobody is listening to.

**Requests are dispatched concurrently.** `session/prompt` blocks for the whole
turn; a connection that stopped reading during it would be deaf to exactly what
a developer sends when a turn drags — `session/cancel` (#704), or a prompt in
another session. Responses may now be written out of order, which JSON-RPC
allows: the id correlates them, and the write mutex keeps lines whole.

## Prompt content

Text and images go up — the API has taken images since before this adapter
existed, so `promptCapabilities.image` is now true. Blocks naming something on
the developer's machine (resource links, embedded context) are dropped with a
log line rather than pasted in as a path the sandbox cannot resolve; sending
one would spend a turn hunting a file that was never going to be there.
`loadSession` stays false until #703.

## Tests

`go test ./... && go vet ./...` green, `internal/acp` also under `-race`. The
tests assert the things that would cost money or mislead: no conversation
created on a refusal path, no prompt posted for an unknown session, the stream
head learned **before** the prompt is posted (#398 from the other side), and no
notification forwarded for output that is not protocol.

Not yet verified against a live instance: the `acp` capability field this
depends on ships with #711 and has not reached prod. Live smoke to follow, and
I will report the result here.

### DIFF
--- a/cli/internal/acp/agent.go
+++ b/cli/internal/acp/agent.go
@@ -61,6 +61,7 @@ type Agent struct {
 	sessions *sessions
 
 	mu            sync.Mutex
+	notifier      Notifier
 	initialized   bool
 	authenticated bool
 	negotiated    int
@@ -96,6 +97,8 @@ func (a *Agent) Request(ctx context.Context, method string, params json.RawMessa
 		return a.authenticate(ctx, params)
 	case "session/new":
 		return a.newSession(ctx, params)
+	case "session/prompt":
+		return a.prompt(ctx, params)
 	default:
 		return nil, Errorf(CodeMethodNotFound, "method not found: %s", method)
 	}
@@ -157,13 +160,17 @@ func (a *Agent) initialize(raw json.RawMessage) (any, error) {
 // obliges us to replay a whole conversation as session/update notifications
 // before answering, and an agent that claims it and does not do it leaves an
 // editor showing an empty transcript for a conversation that has one.
-// Prompt capabilities stay false for the same reason — the prompt path is
-// #701 — even though the API behind them already carries images.
+//
+// `image` is true because the prompt path carries images to the API, which has
+// taken them since before this adapter existed. `embeddedContext` stays false:
+// it means the editor may inline a local file's contents, and this agent works
+// on a sandbox's filesystem, so accepting them would be accepting context
+// about a machine the agent cannot see.
 func agentCapabilities() map[string]any {
 	return map[string]any{
 		"loadSession": false,
 		"promptCapabilities": map[string]any{
-			"image":           false,
+			"image":           true,
 			"audio":           false,
 			"embeddedContext": false,
 		},
--- a/cli/internal/acp/agent_test.go
+++ b/cli/internal/acp/agent_test.go
@@ -127,9 +127,15 @@ func TestInitializeDoesNotClaimCapabilitiesItCannotHonour(t *testing.T) {
 		t.Errorf("loadSession = %v, want false until #703 builds the replay", caps["loadSession"])
 	}
 	prompt := caps["promptCapabilities"].(map[string]any)
-	for _, k := range []string{"image", "audio", "embeddedContext"} {
+	if prompt["image"] != true {
+		t.Errorf("promptCapabilities.image = %v, want true — the prompt path carries images", prompt["image"])
+	}
+	// embeddedContext means the editor may inline a local file's contents, and
+	// the agent works on a sandbox's filesystem — accepting it would be
+	// accepting context about a machine the agent cannot see.
+	for _, k := range []string{"audio", "embeddedContext"} {
 		if prompt[k] != false {
-			t.Errorf("promptCapabilities.%s = %v, want false until #701 builds the prompt path", k, prompt[k])
+			t.Errorf("promptCapabilities.%s = %v, want false", k, prompt[k])
 		}
 	}
 }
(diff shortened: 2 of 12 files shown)


jhgaylor:
## Live smoke: passed

Run against prod once #711's `acp` capability field had deployed, driving the
binary over stdio exactly as an editor would (`initialize` → `session/new` →
`session/prompt`), against `acp-proof-659` (claude, haiku).

```
→ initialize
← {"agentCapabilities":{"loadSession":false,
    "promptCapabilities":{"audio":false,"embeddedContext":false,"image":true}},
   "authMethods":[],"protocolVersion":1}
→ session/new
← {"sessionId":"f41b23ea-d391-47d8-a42c-4da35d5816c7"}
→ session/prompt: Say exactly ACP-OK, then run ls in your workspace and tell me how many entries you saw.

  [available_commands_update]
  [usage_update]
ACP-OK
  [tool] Terminal (pending)
  [tool] → completed
I saw 0 entries in the workspace.

← stopReason: end_turn
exit code: 0
```

What that proves, item by item:

- **`authMethods` came back empty** — credentials were present, so no login
  prompt was offered for a CLI that is already logged in.
- **The session id is the conversation id.** `f41b23ea…` is the row in
  `conversations`; the sprite's own session id never appeared on the wire.
- **Message chunks, a tool call and its completion all arrived as
  `session/update`** — forwarded, not parsed. Note `available_commands_update`
  and `usage_update` came through too: update variants the web UI drops, which
  an editor may well render. Forwarding rather than translating is what makes
  that free.
- **The stop reason is the sprite's own**, correlated out of the terminal stage
  event, and the process exited 0.

The client's `[stderr]` lines showed the ignored-parameter log firing as
designed:

```
msg="ignoring client-side session parameters" cwd=/home/jake/dev/... mcpServers=0
  why="the agent runs in a sandbox, not on this machine"
```

The agent reporting an empty workspace is correct for this agent — it has no
repo attached. It is also a fair illustration of the v1 limit ADR 0015 states
outright: the editor's project is not what the agent is looking at.

Issue #398, opened 2026-08-03
cli: keys create always fails and loses the key, conv prompt exits on a replayed prior turn, and a failed turn exits 0


jhgaylor:
Two commands fail on their documented happy path every time, and a third reports success for a failed run. All three are decoder/terminal-condition bugs in the Go CLI against the current server responses.

## 1. `fountain keys create` fails always, and destroys the key it created

`cli/internal/cmd/keys.go:62-82` decodes into a `data` envelope:

```go
var resp struct {
    Data map[string]any `json:"data"`
}
...
key := output.ToString(resp.Data["key"])
if key == "" {
    Fatalf("unexpected response: %v", resp.Data)
}
```

The server emits a flat object — `apps/fountain/lib/fountain_web/controllers/api_key_json.ex:7-15`:

```elixir
def created(%{key: %ApiKey{} = key, raw_key: raw_key}) do
  %{id: key.id, name: key.name, key: raw_key, prefix: key.key_prefix, created_at: key.inserted_at}
```

There is no `data` wrapper, so `resp.Data` is nil and the command exits 1 with `unexpected response: map[]` — **after** the key has been created server-side. The plaintext key exists only in that response (`Accounts.create_api_key` is the sole source), so it is lost permanently and the user is left with an orphan key on their account.

Note also `key_prefix` vs the emitted `prefix`.

## 2. `fountain conv prompt` never streams the turn it just sent

Two bugs compose:

`cli/internal/cmd/stream.go:66-71` opens the initial connection with an empty `Last-Event-ID`, so the header is omitted and `conversation_controller.ex:458-459` parses it as `0` — `_unsafe_list_log_events(conv_id, 0, ...)` replays the conversation's **entire history** before the live tail.

`cli/internal/cmd/conv.go:312-316` then terminates on the first `turn`/`done` it sees, with no turn scoping:

```go
if stage == "turn" && state == "done" {
    exit := exitCodeFromStageDone(data)
    fmt.Fprintf(os.Stderr, "▸ turn done (exit_code=%s)\n", exit)
    return true
}
```

`conv prompt` is by definition a second-or-later turn, so a `turn`/`done` from an earlier turn is always in the replay. The CLI dumps the whole prior transcript, prints `▸ turn done` for a turn that finished minutes ago, and exits 0 while the turn the user just queued is still starting. `conv stream <id>` on any conversation with a completed turn has the same failure.

[...]

PR #713, opened 2026-08-15
feat: session/cancel, and proof a turn survives a dropped stream (#704)

### LINKED ISSUE #704: acp: session/cancel, and survive a dropped SSE mid-turn

jhgaylor:
Closes #704. Gate 2 of **ADR 0015**, tracked by #709.

Two halves of the same promise: a turn started from the editor keeps running,
and stops when you say so.

## Cancel answers nothing, deliberately

`session/cancel` is a notification. ACP puts the answer where the editor is
already looking — the pending `session/prompt` returns `cancelled` when the
interrupted stage reaches its follow — so the handler asks the server to stop
the turn and returns.

It only works because requests dispatch concurrently (#701). A connection that
read one message at a time would not reach the cancel until the turn it is
meant to stop had already finished, and the feature would appear to work in
every unit test while never working in an editor.

So the test that matters drives a **real connection over a pipe**: it blocks
the turn until the interrupt arrives, then asserts the outstanding prompt comes
back `cancelled`.

That test is written the way a client actually behaves — one message at a time,
each sent after the last is answered. My first version fed all four lines at
once and hung, because the prompt raced the `session/new` that creates its
session. No real client does that (it has nothing to prompt until the session
id comes back), but it was worth the detour: the harness now models a client
rather than a file.

A cancel arriving after the turn ended gets a 409, and that is **not** a
failure — it is the ordinary outcome of racing the agent to the finish, and
reporting it would report a failure that did not happen.

## Surviving a dropped stream was inherited, but unproven

#701 put the follow loop in `internal/stream` so both surfaces share #398's
fix. This adds the test that the ACP path really goes through it: an httptest
server that hangs up mid-turn, a client that reconnects, and the events either
side of the break both arriving.

It also pins the two things that make the reconnect *correct* rather than
merely present:

- **`Last-Event-ID` carries the last id**, so the editor is not re-sent updates
  it already rendered.
- **The request asks for `streams=acp,stage`**, so a runtime dialect never
  reaches a client that cannot parse it.

Plus: a cancelled context ends a follow, so when the editor goes away we stop
reading a stream nobody is listening to.

## Tests

`go test ./... && go vet ./...` green; `internal/acp` also under `-race
-count=2`, which is where a concurrency claim like this one earns its keep.

### DIFF
--- a/cli/internal/acp/agent.go
+++ b/cli/internal/acp/agent.go
@@ -104,10 +104,19 @@ func (a *Agent) Request(ctx context.Context, method string, params json.RawMessa
 	}
 }
 
-// Notify handles client notifications. There are none this agent acts on yet;
-// `session/cancel` arrives here when it lands (#704).
-func (a *Agent) Notify(_ context.Context, method string, _ json.RawMessage) {
-	a.log.Debug("ignoring notification", "method", method)
+// Notify handles client notifications.
+//
+// `session/cancel` is a notification, not a request: ACP expects the *pending*
+// `session/prompt` to be what answers, with a `cancelled` stop reason. So this
+// asks the server to stop the turn and returns; the prompt's own follow sees
+// the interrupted stage and reports it.
+func (a *Agent) Notify(ctx context.Context, method string, params json.RawMessage) {
+	switch method {
+	case "session/cancel":
+		a.cancel(ctx, params)
+	default:
+		a.log.Debug("ignoring notification", "method", method)
+	}
 }
 
 func (a *Agent) initialize(raw json.RawMessage) (any, error) {
--- a/cli/internal/acp/cancel.go
+++ b/cli/internal/acp/cancel.go
@@ -1,0 +1,51 @@
+package acp
+
+import (
+	"context"
+	"encoding/json"
+)
+
+type cancelParams struct {
+	SessionID string `json:"sessionId"`
+}
+
+// cancel stops the running turn.
+//
+// Three properties, each of which is a way this could go wrong:
+//
+// **It answers nothing.** JSON-RPC forbids a response to a notification, and
+// ACP puts the answer somewhere better anyway — the pending `session/prompt`
+// returns `cancelled` when the interrupted stage reaches its follow. An editor
+// therefore learns the turn stopped from the request it is already waiting on.
+//
+// **It can only work because requests dispatch concurrently.** A connection
+// that read one message at a time would not reach this notification until the
+// prompt it is meant to stop had already finished.
+//
+// **A turn that already ended is not an error.** The server answers 409 for a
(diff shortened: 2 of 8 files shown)


jhgaylor:
## Live smoke: cancel confirmed against prod

First attempt proved nothing and is worth recording: the prompt (`sleep 120`) finished before the 45s wait, so the response was already buffered and the cancel landed after the turn — `end_turn`, 0.0s. A green result that tested nothing.

Second attempt, with a turn that genuinely blocks (`sleep 300`) and a 20s wait:

```
session/new → 483df0d8-4be5-48d2-b2a5-344e1b676c0b
waiting for the turn to be underway…
session/cancel
stopReason: cancelled (answered 0.1s after the cancel)
updates seen: {available_commands_update: 1, usage_update: 4, agent_thought_chunk: 15,
               agent_message_chunk: 4, tool_call: 1, tool_call_update: 3}
```

The tool call in that count is the `sleep` — the turn was demonstrably mid-flight, the notification reached a handler blocked in `session/prompt`, and the outstanding request answered `cancelled` a tenth of a second later. That is the concurrency claim in this PR, exercised end to end rather than against a fake.

Issue #699, opened 2026-08-14
acp: session/new — choose a Fountain agent, and mint an editor-facing session id


jhgaylor:
Gate 1 of **ADR 0015** (#631), and the one place the protocol does not tell us
what to do.

## The gap

`session/new` carries a `cwd` and a list of MCP servers. It does **not** carry
"which Fountain agent". But the unit an editor is reaching for is a Fountain
agent — one that already has an environment, vault overrides, skills, MCP
servers and inference credentials provisioned. That is four of the five
reasons the ADR gives for using Fountain from an editor at all.

So the adapter needs a mechanism, and the ADR does not pick one. Candidates:

- **A flag on the subcommand** (`fountain acp --agent <id-or-slug>`), so the
  editor's agent-server config names the agent. Simplest; one editor entry per
  agent, which is arguably correct — different agents are different agents.
- **Session config / `_meta`** on `session/new`, if the client surfaces it.
- **A default agent** in `~/.fountain/config`, with the flag as override.

Pick one, and say why the others were not picked. The choice is visible in
every editor config snippet we will ever publish, so it is worth getting right
before that documentation exists.

## Then: two identities, and a third that must not leak

`session/new` returns a session id that **the agent mints**. That is us, not
the sprite. There are three ids in play and conflating any two is a bug:

| id | owned by | who may see it |
|---|---|---|
| ACP session id | `fountain acp` | the editor |
| `conversation.id` | Fountain | the API, the web UI |
| `conversation.runtime_session_id` | the sprite's runtime | **nobody upward** |

The third is a sprite-filesystem-scoped identity (see the `send_prompt/4`
docstring and #649) and has no meaning outside the sandbox that holds it. The
adapter maps ACP session id -> conversation id and never reads the third.

Also: `POST /api/conversations` provisions a sandbox, which takes seconds. The
ADR's *Consequences* section flags this — four seconds of nothing reads very
differently in an editor than on a web page with a spinner. Decide what the
editor sees while `session/new` is outstanding.

## Done when

`session/new` from an editor creates a real Fountain conversation for a chosen
agent and returns a session id the adapter can map back.

Part of #709.

Issue #657, opened 2026-08-11
gemini streams its session/load replay after responding, so the replay-discard window misses it


jhgaylor:
Found running the real `Fountain.Runtimes.ACP.Peer` against a live `gemini --acp` (2026-08-10). Blocks nothing today — the ACP flag is off by default — but it must be settled before gemini serves real conversations (#638/#642).

## What happens

ACP says an agent handling `session/load` "**MUST** replay the entire conversation to the Client in the form of `session/update` notifications" *before* responding. `Peer` relies on that: it turns on replay-discard when it sends `session/load` and turns it off when the response arrives, so history we already hold as `log_events` is not written twice.

Gemini does not replay before responding. From `packages/cli/src/acp/acpSessionManager.ts`:

```js
// Stream history back to client
// eslint-disable-next-line @typescript-eslint/no-floating-promises
session.streamHistory(sessionData.messages);
```

A floating promise — the response returns immediately and the replay streams afterwards, landing *outside* our discard window.

## Evidence

Turn 1 asked gemini to reply `ECHO-5`. Turn 2 asked what token it had been given. Through the real peer:

```
gemini turn 2 (session/load)
  outcome        : {:done, "end_turn"}
  acp lines      : 4
  blocks         : %{text: 2}
  assistant text : "ECHO-5ECHO-5"
```

Two `:text` blocks where there should be one. The first is turn 1's assistant message replayed; the second is the fresh answer. On a real conversation this duplicates the whole transcript into `log_events` and onto the SSE stream, on every turn after the first — precisely what the discard exists to prevent.

Claude is unaffected: it advertises `sessionCapabilities.resume`, so it never takes the load path.

## Options

1. **Bounded quiet period.** After `session/load` returns, keep discarding until no `session/update` has arrived for ~250ms, then prompt. Simple, matches the observed behaviour, and cheap next to gemini's ~2.9s handshake. Timing-based, which is the honest objection.
2. **Count the replay.** We know how many turns the conversation has; discard that many `agent_message_chunk` groups. Fragile — depends on gemini's chunking matching our turn count.
3. **Dedupe at render.** Leaves the duplication in the database, which is where it costs the most.
4. **Fix upstream.** It's a spec violation and a one-word fix (`await`). Worth reporting regardless of which mitigation we pick, since the mitigation can be removed if it lands.

My preference is 1 plus 4: take the latency, and stop taking it when upstream is fixed.

## Note for whoever picks this up

[...]

PR #715, opened 2026-08-15
feat: session/load replays a conversation into a fresh editor (#703)

### LINKED ISSUE #703: acp: session/load — replay the transcript from `Last-Event-ID: 0`

jhgaylor:
Closes #703. The last of gate 2 for **ADR 0015**, tracked by #709. Stacked on
#714.

This is the reason the ADR gives for reaching past a local agent at all: **the
work outlives the editor.** Close the laptop mid-turn, reopen tomorrow, and the
conversation is still there — because it was never in the editor, or in this
process.

It also cashes in #699's choice to make the ACP session id the conversation id.
An editor hands back an id from last week; a minted one would resolve to a map
that died with the process that made it.

## The replay precedes the response, and that is the whole spec

`session/load` **MUST** re-emit the conversation as `session/update`
notifications before it answers. We have been on the receiving end of the
opposite: gemini answers first and replays afterwards, which double-rendered a
turn and cost us #657 and #661. Being the agent this time is no reason to
repeat it.

So the notifications are written before the handler returns, and the test
asserts the count *at the moment it does* — a client that renders on the
response would otherwise show an empty transcript for a conversation that has
one.

The replay is a drain, not a follow: `?streams=acp&wait=false` with
`Last-Event-ID: 0`. Waiting for live events would hold the response open for as
long as the conversation stayed interesting.

`loadSession` flips to true here rather than in #710, because the capability is
a promise about ordering as much as ability.

## A conversation this process never opened needs its own lookup

`GET /api/conversations/:id` gained the same derived `acp` boolean the agents
endpoint got in #711. A legacy-runtime conversation has a transcript, but not
one stored as protocol — it is refused by name rather than replayed as nothing.

Loading requires no `--agent`: the conversation already knows which agent it
belongs to, and demanding one would refuse to reopen a conversation the editor
is entitled to see.

## What this cannot fix

A conversation whose sandbox was reclaimed replays **in full** while the agent
inside remembers none of it (#649). The transcript is real; the continuity is
not. There is no honest way to say that in the protocol — inventing an agent
message to explain it would be worse than the log line — so the docs say it
instead, and the load logs the conversation's status.

## Tests

Ten cases covering the ordering guarantee, the refusals (unknown conversation,
non-ACP runtime, failed replay), that a loaded session is promptable and does
not create a second conversation, and that replayed updates carry our session
id rather than the sprite's. Elixir side asserts `acp` true/false through the
real controller. `go test ./... && go vet ./...` green, `-race` clean.

### DIFF
--- a/apps/fountain/lib/fountain_web/controllers/conversation_json.ex
+++ b/apps/fountain/lib/fountain_web/controllers/conversation_json.ex
@@ -31,6 +31,10 @@ defmodule FountainWeb.ConversationJSON do
       agent_id: c.agent_id,
       vault_id: c.vault_id,
       runtime: c.runtime,
+      # Derived, never stored — the same signal as on an agent (#702). A
+      # protocol client asks before reopening a conversation, because a
+      # legacy-runtime one has no ACP transcript to replay.
+      acp: Fountain.Runtimes.ACP.enabled?(c.runtime),
       status: c.status,
       runtime_session_id: c.runtime_session_id,
       source: c.source,
--- a/apps/fountain/lib/fountain_web/schemas.ex
+++ b/apps/fountain/lib/fountain_web/schemas.ex
@@ -49,6 +49,15 @@ defmodule FountainWeb.Schemas do
         agent_id: %Schema{type: :string, format: :uuid, nullable: true},
         vault_id: %Schema{type: :string, format: :uuid, nullable: true},
         runtime: %Schema{type: :string, enum: ~w(claude codex gemini opencode)},
+        acp: %Schema{
+          type: :boolean,
+          readOnly: true,
+          description:
+            "Whether this conversation's runtime speaks the Agent Client Protocol. " <>
+              "When true its output is stored as ACP `session/update` notifications on " <>
+              "the `acp` event stream, which is what a protocol client replays; when " <>
+              "false the output is the runtime's own dialect on `stdout`."
+        },
         status: %Schema{
           type: :string,
           enum: ~w(pending running idle failed terminated)
--- a/apps/fountain/test/fountain_web/controllers/conversation_controller_test.exs
+++ b/apps/fountain/test/fountain_web/controllers/conversation_controller_test.exs
@@ -57,6 +57,25 @@ defmodule FountainWeb.ConversationControllerTest do
       assert body["data"]["id"] == conv.id
     end
 
+    # `fountain acp` asks before replaying a conversation an editor handed
+    # back: a legacy-runtime conversation has a transcript, but not one stored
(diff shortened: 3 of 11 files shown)


jhgaylor:
## Gate 2 verified live, after the fix it exposed

Re-run of the `session/load` smoke against prod once #716 had deployed, on the
same conversation that exposed the bug:

```
session/load f41b23ea-…
replayed 31 updates BEFORE the response      (was 0)
update kinds: {available_commands_update: 2, usage_update: 9,
               agent_thought_chunk: 10, agent_message_chunk: 5,
               tool_call: 1, tool_call_update: 4}
← {}

session/prompt on the loaded session
agent: You asked me to say exactly "ACP-OK".
stopReason: end_turn
```

Every property this gate claims, exercised end to end: the whole transcript
replays, it replays **before** the response, the update variants survive
intact, and the loaded session is promptable — the agent answers a question
that can only be answered from turn 1.

Cancel was verified separately on #713: `stopReason: cancelled`, 0.1s after
the notification, with a tool call in flight.

The first run of this smoke returned `replayed 0` and is what found #716 —
`?streams=acp` had never worked for replay, so `session/load` and every
mid-turn reconnect were quietly returning nothing.

PR #716, opened 2026-08-15
fix: the acp stream was dropped from every filtered replay

jhgaylor:
Found by the live smoke of #703 (PR #715), not by a test: `session/load`
against a real conversation holding 31 stored updates replayed **zero** of
them.

```console
$ curl ".../stream?streams=acp&wait=false"      # nothing
$ curl ".../stream?wait=false" | jq -r .stream | sort | uniq -c
  31 acp
  14 stage
```

## The bug

`?streams=` has two implementations. The live one matches an event already in
hand; the history one builds a `where` clause. The history one carried an
allow-list:

```elixir
real_streams = Enum.filter(streams, &(&1 in ["stdout", "stderr"]))
# …
true -> from e in query, where: false   # "all values were unknown"
```

Written when those were the only two streams. When ACP added a third in #647,
a filtered request began answering with a conversation's **future and none of
its past** — live events matched, replayed ones did not.

## What it broke, this week

- **`session/load` replayed nothing** (#703). An editor reopening a
  conversation saw a blank transcript for a conversation that has one — the
  exact failure `loadSession: true` is a promise against.
- **A reconnect mid-turn lost what it missed** (#704). `fountain acp` resumes
  with `Last-Event-ID` and `streams=acp,stage`; the catch-up replay dropped
  every `acp` line, so output produced while the connection was down never
  reached the editor. The turn carried on afterwards, which is what made it
  invisible.

Neither showed up in tests: the Go tests fake the API, and the Elixir
filter tests covered `stdout`, `stderr` and `stage` — the three names the
allow-list happened to know.

## The fix

Stop keeping a list. `"stage"` is the synthetic name for `kind: "stage"`
events; every other value filters on the `stream` column. An unknown name
matches no row because no row carries it, so "return nothing rather than
everything" survives without a list anyone has to remember to extend.

The two halves are one function now — the controller delegates to
`Conversations.event_in_streams?/2`.

## The test that was missing

`stream_filter_test.exs` runs one table of (event, selection) pairs through
**both** halves and asserts they agree. That is the shape of test this needed:
each half was plausible on its own, and they are only comparable together.

Reverted against the old code, 4 of its 6 cases fail. The HTTP-level filter
tests gained `acp` and `acp,stage` — the two selections a protocol client
actually sends.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -856,16 +856,28 @@ defmodule Fountain.Conversations do
   defp apply_limit(query, limit) when is_integer(limit) and limit > 0,
     do: from(e in query, limit: ^limit)
 
-  # `streams` is a list of allowed stream identifiers. We accept the
-  # three values that show up in log_events: `"stdout"`, `"stderr"`, and
-  # `"stage"` (the synthetic name we give to `kind: "stage"` events,
-  # which don't have a real `stream` column value). `nil`/empty list =
+  # `streams` is a list of allowed stream identifiers: any value of the
+  # `stream` column, plus `"stage"`, the synthetic name for `kind: "stage"`
+  # events (which have no `stream` value of their own). `nil`/empty list =
   # no filter.
+  #
+  # There is deliberately **no allow-list of stream names**. There used to be
+  # one — `["stdout", "stderr"]`, written when those were the only two — and
+  # when ACP added a third (`"acp"`, one stored `session/update` per line) the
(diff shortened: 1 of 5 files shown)


jhgaylor:
Addendum, because it sharpens the point: `LogEvent` already listed the stream —

```elixir
@streams ~w(stdout stderr acp)   # log_event.ex, updated when ACP landed
```

The allow-list lived in two places and only one of them was updated. The schema knew about `acp`; the query filter did not. That is the argument for having no second list at all, which is what this PR does.

Issue #330, opened 2026-08-02
cost: the quota check is check-then-insert, and a double-click can strand a sandbox row for an hour


jhgaylor:
Two related races around `Quotas.check_sandbox_quota`.

**1. Check-then-insert.** `quotas.ex:84-93` reads `count < limit` with no lock and no constraint; the call sites (`conversations.ex:785` create, `:1039` wake) do the check and `create_sandbox` in a plain `with` — no transaction, no advisory lock, no DB constraint on the active count. N concurrent requests at the cap can each pass and create up to N−1 rows over the limit. Bounded by the 600/min API bucket and the cost of provisioning.

**2. Stranded row on concurrent wake.** Two concurrent prompts to a dormant conversation both enter `wake_conversation` → both reach `create_fresh_sandbox_and_start` (`conversations.ex:1031-1052`). The loser's `start_child` returns `{:error, {:already_started, _}}`, the `with` falls through, and its just-created `pending` sandbox row is stranded with no server and no sprite — holding a quota slot until reaper pass 1 releases it after 60 minutes (`sandbox_reaper.ex:62,104-127`).

Self-healing, but user-visible: a user at their cap can lock themselves out for up to an hour by double-clicking. That matters more now that the cap is enforced.

**Fix:** a per-user advisory lock (or a partial unique index on active sandboxes) around check + insert closes both — the second one also wants the loser to clean up its own row rather than wait for the reaper.

PR #730, opened 2026-08-15
fix: a lost wake race no longer points a conversation at a dead sandbox (#717)

### LINKED ISSUE #717: A lost wake race leaves the conversation pointing at the sandbox it just terminated
Found while smoking `fountain acp` (#709) against prod. Every conversation the
ACP adapter opened has **two** sandbox rows: the one the conversation points at
(marked `terminated`) and the one that actually ran the turns (`ready`, with no
conversation pointing at it).

## Evidence

```sql
select c.id, c.inserted_at, s.sprite_name, s.status,
       (select string_agg(o.sprite_name||':'||o.status, ', ') from sandboxes o
         where o.user_id = c.user_id
           and abs(extract(epoch from (o.inserted_at - c.inserted_at))) < 2
           and o.id <> c.sandbox_id) as same_second_others
from conversations c join sandboxes s on s.id = c.sandbox_id
where c.inserted_at > now() - interval '3 days' order by c.inserted_at desc;
```

```
483df0d8… | 01:19:28 | …ccd17e49 | terminated | …129e3fc5:ready   ← fountain acp
03048f77… | 01:18:24 | …49e76b8f | terminated | …c2a26118:ready   ← fountain acp
f41b23ea… | 00:25:12 | …c8dd285e | terminated | …02d4542b:ready   ← fountain acp
9d286fa7… | 00:07:48 | …b20dcde8 | ready      |                   ← other path
ea4e91c5… | 10:51:57 | …1ca575da | terminated |
```

The pattern is confined to the three conversations opened through
`session/new`. Conversations created by other paths have one row.

The live sprite is the orphan: the conversation's stage events name
`fountain-50e06232-02d4542b` on every reattach, and turn 2 ran on it happily —
while `GET /api/conversations/:id` reports
`sandbox: {sprite_name: "fountain-50e06232-c8dd285e", status: "terminated"}`.

```
75a05318… | fountain-50e06232-02d4542b | ready | 2026-08-15 00:25:12 | conversations pointing at it: 0
```

Prod totals: **204 of 506 sandbox rows have no conversation** (3 created in the
last 24h, all three from tonight's ACP smokes; 3 orphans still `ready`).

## Why it matters

- **The quota probably double-counts.** `Fountain.Quotas` reserves against
  sandbox rows, so an editor session may consume two of a user's five.
- **The API and UI lie about a live conversation.** A user watching a running
  turn is told its sandbox is terminated.
- **An orphan `ready` row is a sprite nothing owns.** `SandboxReaper` works off
  rows, so it should still be reclaimed — but nothing links it back to the
  conversation it is serving, and the accounting around suspend/park (0017)
  keys off the conversation's sandbox.

## What is different about the ACP path

The only distinguishing feature I can see: `session/new` creates a conversation
**with no prompt** (`POST /api/conversations` with just `agent_id`), and the
first prompt arrives a second or two later over `POST .../prompts`. Every other
caller — `fountain run`, the UI — creates with a prompt in the same request.

So the suspicion is the no-prompt creation path: the row created by
`start_conversation` ends up abandoned and a second one is created (by the wake
or provisioning path) without `conversations.sandbox_id` being repointed. The
stage events show only **one** `provision` cycle, which is consistent with the
first row never being provisioned at all.

I have not read the mechanism out of the code — this is the outside view, and
someone should confirm before fixing. Adjacent: #667 (the wake path's reuse arm)
and #649 (a conversation whose sandbox is replaced).

## Repro

```
fountain acp --agent <any claude agent>   # then session/new + session/prompt
```
or `POST /api/conversations` with no `prompt`, then `POST .../prompts`.

jhgaylor:
Closes #717.

`create_fresh_sandbox_and_start/4` repointed `conv.sandbox_id` at its new
sandbox **before** starting the server. When the start then lost the race, the
losing branch cleaned up its own row but never undid the repoint — leaving the
conversation naming a sandbox it had just terminated, while the winner served
turns on a different one.

## What that looked like

- A conversation that reads `terminated` through the API and the UI while it is
  happily answering.
- An orphan `ready` sandbox nothing references. **Prod is carrying 204 of those
  out of 506 rows.**
- A quota slot consumed twice, since both rows count.

`fountain acp` reproduced it on *every* session, which is how it surfaced:
`session/new` and the first `session/prompt` land about a second apart, so the
prompt takes the wake path before Horde's registry has the server that is
already starting.

## The fix is ordering

The row is repointed only after the server actually starts. The losing branch
then has nothing to undo — it retires its own sandbox and leaves the
conversation to the winner, which is the process that knows which sandbox it is
running. A plain start failure now retires its unused sandbox too, for the same
quota reason the already-started branch already carried.

Deferring is what makes the loser correct without needing to know the winner's
sandbox id, which it cannot: by the time it loses, its own write would already
have overwritten it.

**Remaining window:** between the server starting and the row being updated,
the row still names the old sandbox. That is transient and self-correcting; the
old one was permanent. Called out in a comment so the next reader does not
"fix" it by moving the update back.

## Tests

Five, one per branch — what the row points at and what is left holding a quota
slot. Reverted to the old order, two fail.

The first version of the test failed against the *working* fix, because it
counted a sandbox the factory creates; it now diffs the rows created by the
wake itself. Worth mentioning because the wrong version would have "passed"
by accident in the other direction.

Full conversation suites green (324 tests); credo, dialyzer and format clean.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1394,13 +1394,30 @@ defmodule Fountain.Conversations do
                  user_id: conv.user_id
                })
              end
-           ),
-         _ <- mark_old_sandbox_terminated(conv.sandbox_id),
-         {:ok, conv} <-
-           update_conversation(conv, %{sandbox_id: new_sandbox.id, status: "pending"}) do
+           ) do
+      # The row is repointed *after* the server starts, not before (#717).
+      #
+      # The old order repointed first, so a wake that then lost the start race
+      # left the conversation pointing at the sandbox it had just terminated,
+      # while the winner ran on a different one — a conversation that reads as
+      # terminated in the API and the UI while it is happily serving turns, and
+      # an orphan `ready` row nothing references. `fountain acp` reproduced it
+      # on every session, because `session/new` and the first prompt arrive a
+      # second apart and the prompt takes this path before the registry has the
+      # new server.
+      #
+      # Deferring leaves a much smaller window — between the server starting
+      # and the row being updated — in which the row still names the old
+      # sandbox. That one is transient and self-correcting; the old one was
+      # permanent.
       case start_conversation_server(conv, new_sandbox.id, runtime_module, initial_prompt) do
-        {:ok, _} = ok ->
-          ok
+        {:ok, _} ->
+          _ = mark_old_sandbox_terminated(conv.sandbox_id)
+
+          {:ok, conv} =
+            update_conversation(conv, %{sandbox_id: new_sandbox.id, status: "pending"})
+
+          {:ok, _unsafe_get_conversation!(conv.id)}
 
         {:error, {:already_started, winner_pid}} ->
           # Lost a concurrent wake of the same conversation. The winner's
@@ -1410,6 +1427,9 @@ defmodule Fountain.Conversations do
           # themselves out by double-clicking (#330). Clean up our own row and
           # hand the prompt to the winner, which drops it if a turn is already
           # running — exactly right for a double-click.
+          #
+          # The conversation is left alone: the winner owns it, and it is the
+          # winner's sandbox the row should name.
           _ = mark_old_sandbox_terminated(new_sandbox.id)
 
           if is_binary(initial_prompt) and initial_prompt != "" do
@@ -1419,6 +1439,10 @@ defmodule Fountain.Conversations do
           {:ok, _unsafe_get_conversation!(conv.id)}
 
         {:error, _} = err ->
+          # Nothing ever ran on this sandbox. Retiring it keeps a failed wake
+          # from holding a quota slot until the reaper's next pass — the same
+          # reasoning as the branch above.
+          _ = mark_old_sandbox_terminated(new_sandbox.id)
           err
       end
     end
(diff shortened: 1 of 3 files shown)


jhgaylor:
## CI caught something that strengthens the case for this change

Partition 1 failed on the existing #330 test:

```
1) test concurrent wake loses cleanly (#330)
   the loser terminates its own sandbox row instead of stranding it
   assert Fountain.Quotas.active_sandbox_count(user.id) == 0
   left: 1
```

It asserted **zero** active sandboxes after a losing wake, because the old code retired the conversation's existing sandbox unconditionally, before starting. Deferring that to the success path leaves it alive — hence 1.

**The test was encoding behaviour that is actively unsafe.** `wake_conversation` has a reuse arm (`maybe_reuse_sandbox/1`): when the existing sandbox is `ready` or `suspended` and its sprite is still there, the winner starts a server **against that sandbox**. So under the old order a losing wake could terminate the sandbox the winner was running on — a worse failure than the stranded quota slot #330 was about, and one this reordering also removes.

The assertion now states the contract directly: the row the loser created for itself is terminated (that is #330's point, and it still holds), and the conversation's existing sandbox is left to the winner, which retires it only if it provisioned a replacement.

Full suite green afterwards: 2,788 tests, 0 failures.

PR #734, opened 2026-08-16
docs: ADR 0020 — Buzz as a client of the ACP gateway

jhgaylor:
Adds `decisions/0020-buzz-as-a-client-of-the-acp-gateway.md` (Status: **Proposed**, nothing built).

## What it decides
Move a Buzz agent's Nostr work off the sandbox and onto the gateway, in three gated phases:

- **Phase 1 — host `buzz-acp` at the gateway.** Inbound (relay connection, auth, presence, mentions, prompt assembly) moves for free by running Buzz's own harness process at the gateway with `fountain acp` as its stdio child. No new Nostr code.
- **Phase 2 — broker the reply through a Fountain-hosted signer.** The agent's reply is client-signed, so it can't be proxied like an API key (0019); moving it means moving *signing* to Fountain via HTTP-MCP `buzz_*` tools with the key server-side. Every publish becomes governable (0016) and audited (0013).
- **Phase 3 (optional) — `buzz-backend-fountain` provider.** Answer Buzz's remote-agents provider protocol (block/buzz#4289) so the desktop can Start a Fountain-hosted agent. A door onto Phase 1, not the architecture.

## Evidence
Live spike 2026-08-15 against prod: `buzz-acp` run off the desktop drove `fountain acp` end-to-end, received an owner mention, ran a full turn, returned `end_turn` — **inbound proven, and the outbound gap measured** (zero tool calls, no reply reached the channel). Both halves of the ADR are observed, not assumed.

Starting point tagged `pre-buzz-gateway` (on `6cb430c`) for a clean revert boundary.

Tracker + gate issues follow. Sits on ADRs 0014/0015/0016/0017/0019.

### DIFF
--- a/decisions/0020-buzz-as-a-client-of-the-acp-gateway.md
+++ b/decisions/0020-buzz-as-a-client-of-the-acp-gateway.md
@@ -186,10 +186,12 @@ without it.
    `fountain acp` end-to-end; inbound proven live, the outbound gap measured.
 2. **Phase 1 supervisor.** A Fountain-managed process that starts/stops/heals a
    `buzz-acp` per (agent × identity) from vault-held env, with presence surviving
-   sandbox suspend. Owns pinning and the amd64 constraint.
+   sandbox suspend. Owns pinning and the amd64 constraint. *Decided 2026-08-16:
+   lives **inside the Fountain OTP app** (DynamicSupervisor + per-identity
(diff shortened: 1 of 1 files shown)


jhgaylor:
Tracker: #735 · gates #736 / #737 / #738. Starting commit tagged `pre-buzz-gateway`.

Issue #776, opened 2026-08-17
buzz-acp: retire fork after all required behavior ships upstream


jhgaylor:
Tracks the temporary fork pin introduced in #775 and when it is safe to undo.

## State today (verified 2026-09-13)

- `apps/fountain_buzz/buzz-acp.source` = `jhgaylor/buzz@5bbad010f234ac71533f1afd120f44b75d9117dd`
- `apps/fountain_buzz/buzz-acp.version` = `0.5.14-fountain.4`.
- Latest upstream desktop release checked: [`desktop-v0.5.23`](https://github.com/block/buzz/releases/tag/desktop-v0.5.23), published 2026-09-05. It is not a safe replacement for the fork.

The fork carries five upstream behavior changes. The earlier three-row checklist omitted #6103 and #6104, which were already recorded in this issue's comments.

| Upstream PR | Verified state | Required behavior |
|---|---|---|
| [#6088](https://github.com/block/buzz/pull/6088) | Open, unmerged | Channel metadata on `session/new`, persistent session store and `session/load` after restart. |
| [#6101](https://github.com/block/buzz/pull/6101) | Open, unmerged | Owner control commands with rendered mention text. |
| [#6103](https://github.com/block/buzz/pull/6103) | Open, unmerged | `freshSession` on rotate. |
| [#6104](https://github.com/block/buzz/pull/6104) | Open, unmerged | Ignore control commands predating startup. |
| [#6097](https://github.com/block/buzz/pull/6097) | Closed, unmerged; upstream uses a replacement design | Publish the kind:10100 relay-agent directory entry expected by the fork integration, or prove and adopt a compatible replacement. |

The [pinned fork comparison](https://github.com/jhgaylor/buzz/compare/d8281b9c93395f15d55091b131bb2747a0a3da8a...5bbad010f234ac71533f1afd120f44b75d9117dd) contains all five changes. On the checked upstream tag, [ACP session creation](https://github.com/block/buzz/blob/desktop-v0.5.23/crates/buzz-acp/src/acp.rs#L651-L676) still lacks channel metadata and `freshSession`; the session store is absent. The [command parser](https://github.com/block/buzz/blob/desktop-v0.5.23/crates/buzz-acp/src/lib.rs#L4169-L4179) still compares trimmed content exactly, and control handling lacks the pre-startup timestamp guard.

[...]

PR #781, opened 2026-08-17
chore(buzz-acp): repin the fork to the DCO-signed commit of block/buzz#6088

jhgaylor:
Follows the fork pin through the DCO fix on block/buzz#6088. The PR's original commit was amended with `Signed-off-by` (`83eda69a` → `af535a64`, identical tree), and a second commit `9390060` adds the reviewer-requested README docs for `BUZZ_ACP_STATE_DIR` / `--no-resume-sessions`. The old sha is unreachable from any branch on the fork, so the pin moves before GitHub can GC it. #776 still tracks the eventual repin to upstream.

### DIFF
--- a/buzz-acp.source
+++ b/buzz-acp.source
@@ -1,2 +1,2 @@
 # Fork pin — see #776 for when and how to remove. Format: owner/repo@ref
-jhgaylor/buzz@af535a64b241c4b806db1809714f73b57236a707
+jhgaylor/buzz@9390060f0908dfd29e5dd1a2d4a481a2fe81819c
--- a/buzz-acp.source
+++ b/buzz-acp.source
@@ -1,2 +1,2 @@
 # Fork pin — see #776 for when and how to remove. Format: owner/repo@ref
-jhgaylor/buzz@83eda69a99fad02bc3cdf60d61ad14e56a41447e
+jhgaylor/buzz@af535a64b241c4b806db1809714f73b57236a707


jhgaylor:
Closing: superseded — the DCO-signed commit this repins to is already in main's lineage.

When this PR was opened the pin was `83eda69` (unsigned) and this moved it to `9390060`. Since then #787, #788 and #792 have repinned the fork three more times, and main now sits at `5bbad010` (fountain.4). Checking that against the signed commit:

- `af535a64...5bbad010` → `status=ahead, ahead=5, behind=0` — the DCO-signed `af535a64` **is** an ancestor of the current pin.
- `83eda69...5bbad010` → `status=diverged, behind=1` — the unsigned commit is **not**.
- Every commit between them carries a `Signed-off-by`, including `9390060` itself.

So the fork history was rebuilt onto the signed commit and main followed it there by another route. Merging this now would move the pin *backwards* from `5bbad010` to `9390060`, dropping five commits (`!rotate` fresh-session handling, owner control commands with rendered mentions, and the kind:10100 relay-agent directory entry that fountain.4 exists for) — and, because `buzz-acp-publish.yml` runs `gh release upload --clobber` against the already-published `buzz-acp-v0.5.14-fountain.4`, it would overwrite that release's assets with the older build.

Nothing to salvage here; #776 still tracks repinning to upstream once block/buzz#6088 lands.

PR #786, opened 2026-08-17
fix: a lost reuse wake race drops the prompt instead of handing it off (#667)

### LINKED ISSUE #667: wake_conversation's reuse arm surfaces raw {:error, {:already_started, pid}} instead of handing off the prompt
Found while live-verifying #666. `wake_conversation/2` on a conversation whose ConversationServer is already running takes the reuse arm, calls `start_conversation_server`, and returns the raw `{:error, {:already_started, pid}}` to the caller — the prompt is not delivered.

The fresh-sandbox arm (`create_fresh_sandbox_and_start`) already handles exactly this race with a handoff: clean up, `queue_initial_prompt(winner_pid, ...)`, return `{:ok, conv}` (#330, the double-click case). The reuse arm has no equivalent.

Impact today is small — API prompt endpoints check for a live server and use `send_prompt`, so this only bites callers that reach `wake_conversation` while a server is mid-start (two concurrent wakes of the same suspended/ready sandbox: the loser gets an opaque error instead of its prompt being queued). Fix shape: mirror the fresh arm's `{:error, {:already_started, winner_pid}}` clause in the reuse arm — no row cleanup needed there, just queue the prompt and return ok.

jhgaylor:
## Summary
Closes #667.

`wake_conversation/2`'s reuse arm called `start_conversation_server/4` and returned its result as-is. When a wake lost a concurrent start race, that meant returning the raw `{:error, {:already_started, pid}}` to the caller — the prompt was never delivered anywhere.

The fresh-sandbox arm (`create_fresh_sandbox_and_start/4`) already handles exactly this race (#330): on `{:error, {:already_started, winner_pid}}` it queues the prompt to the winner and returns `{:ok, conv}`. This mirrors that handoff in the reuse arm. No row cleanup is needed here (unlike the fresh arm), since reuse provisions no sandbox row of its own.

## Test plan
- [x] Added `wake_conversation/2 a reuse that loses the start race hands the prompt to the winner (#667)` in `apps/fountain/test/fountain/conversations_context_test.exs`, mirroring the existing race coverage in `wake_race_test.exs` but for the reuse arm.
- [x] `mix test apps/fountain/test/fountain/conversations_context_test.exs apps/fountain/test/fountain/conversations/wake_race_test.exs` — 163 tests, 0 failures
- [x] `mix format --check-formatted`
- [x] `MIX_ENV=dev mix compile --warnings-as-errors`
- [x] `mix credo --strict` — no issues

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1303,7 +1303,23 @@ defmodule Fountain.Conversations do
           with :ok <- Fountain.Accounts.check_not_suspended(conv.user_id),
                :ok <- Fountain.Billing.check_active(conv.user_id),
                {:ok, _} <- wake_suspended_sandbox(conv.user_id, sandbox_id) do
-            start_conversation_server(conv, sandbox_id, runtime_module, initial_prompt)
+            case start_conversation_server(conv, sandbox_id, runtime_module, initial_prompt) do
+              {:error, {:already_started, winner_pid}} ->
+                # Lost a concurrent wake of the same conversation to another
+                # caller reusing the same sandbox. Mirrors the handoff in
+                # create_fresh_sandbox_and_start/4 (#330), but reuse provisions
+                # no row of its own, so there is nothing here to clean up —
+                # just hand the prompt to the winner, which drops it if a turn
+                # is already running.
+                if is_binary(initial_prompt) and initial_prompt != "" do
+                  ConversationServer.queue_initial_prompt(winner_pid, initial_prompt)
+                end
+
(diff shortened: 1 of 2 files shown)


BinaryBourbon (approved):
LGTM. Mirrors the existing `{:error, {:already_started, winner_pid}}` handoff from `create_fresh_sandbox_and_start/4` (#330), correctly notes reuse has no row to clean up, and adds a regression test that exercises the lost-race path directly. All 8 CI checks green (tests across 3 partitions, coverage gate, static analysis, secrets scan, compose checks).

Issue #783, opened 2026-08-17
Per-launch environment override: let a Buzz identity (and `fountain acp`) pick the environment, not just the vault


jhgaylor:
## Problem

A conversation's environment is a property of the **agent** — `Conversations.create_conversation` copies `agent.environment_id` onto the sandbox row (`conversations.ex:1046`) and there is no per-conversation override anywhere in the stack: not on `POST /api/conversations`, not on `fountain acp` (which has `--vault` only), and not on Buzz provision (`BuzzIdentity` carries `agent_id` + `vault_id`).

So a hosted Buzz identity gets exactly its agent's environment. To run "a fountain engineer with the fountain environment" and "a buzz engineer with the buzz environment" today you need **two Fountain agents**, duplicating everything else about the agent config (model, runtime, skills, MCP servers, prompt) just to swap the baseline env. Two agents is the right answer for two engineers; it stops being the right answer once one agent config wants to run under N environments.

## Proposal

Make the environment overridable **per launch**, threaded end to end — a per-conversation `environment_id` that falls back to the agent's. Not a Buzz-only parameter: Buzz should be one caller of a general mechanism, the same way it is for `--vault`.

1. **Data.** `BuzzIdentity` gains a nullable, same-tenant `environment_id` FK (cascade rules like `vault_id`); `Buzz.provision/…` and `BuzzAgentController` accept `"environment_id"`. Sandbox rows already carry `environment_id`, so no change there — it just stops being a copy of the agent's.
2. **Wiring.** `Buzz.acp_wiring_env/3` adds `--environment <id>` to `BUZZ_ACP_AGENT_ARGS`; `fountain acp` gains `--environment` and passes `environment_id` on conversation create; the API accepts it.
3. **Resolution + authz.** `Conversations.create_conversation` resolves it like `resolve_vault_id/3` — scoped `Environments.get_environment(id, user_id)`, and an authz rule so a caller cannot point a reviewed agent at an arbitrary environment. Options, decide in the PR: an `agent.allowed_environment_ids` mirroring `allowed_vault_ids` (nil = any tenant env, [] = no override, list = allowlist), or the simpler "override only permitted when the agent has no environment of its own". Falls back to `agent.environment_id` when absent.
4. **Channel resume.** `find_channel_conversation` (#775) keys on `(user, agent, vault, channel)`; it must also key on the resolved environment, or an identity that switches environments resumes an old conversation provisioned with the old one.
[...]

Issue #659, opened 2026-08-11
tracker: gemini over ACP — converted, verified for one turn, not shippable


jhgaylor:
Gemini is the one runtime of four whose ACP conversion cannot be switched on. This is the ticket to work it independently of the rest of the campaign (#644).

**Status: held back in code.** `Fountain.Runtimes.ACP.supported_runtimes/0` excludes gemini, so the per-agent `acp` flag cannot reach it. The adapter entry stays — the conversion is correct as far as it goes.

## What works

Verified against a live `gemini --acp` (0.53.0) through the real `Fountain.Runtimes.ACP.Peer`:

- `initialize`, `authenticate` with `gemini-api-key`, `session/new`, `session/prompt` → correct answer
- model pinned via ACP's own `session/set_model` (gemini advertises `models`, not `configOptions` — the two adapters differ here)
- runs in `/tmp/gemini-workspace`, the directory `Gemini.prepare_sprite/3` git-inits
- eager `authenticate` is **required**: gemini opens a session from ambient credentials without it but does not persist one

Turn 1 is genuinely good. It is turn 2 that fails.

### 1. `session/load` cannot find its own session — #658

```
-32603 Internal error
data.details: "No previous sessions found for this project."
```

**2 successes, 4 failures** across identical code paths. Ruled out individually: missing auth, wrong auth method, a shutdown flush race (4s pause), and the process not exiting (waited for clean `exit 0`). The peer is not at fault — claude, codex and opencode all resume correctly through it.

### 2. The replay escapes the discard window — #657

`loadSession` calls `session.streamHistory(...)` as a floating promise, so `session/load` responds *before* replaying. ACP says it MUST replay first, and `Peer` closes its discard window on the response — so the replay lands in `log_events` and duplicates the transcript on every turn after the first. Observed as `"ECHO-5ECHO-5"`, two `:text` blocks where there should be one.

Note these interact: #658 makes #657 hard to reproduce, because a load that fails never replays.

## Why not just drop gemini

Its legacy path is the worst resume we ship — `gemini --resume` re-enters "the most recent conversation in the workspace", correct only while one conversation ever runs per workspace, an invariant held by accident and asserted by no test. So gemini is the runtime with the *most* to gain from ACP. Dropping it means keeping that, plus its dialect parser, forever.

## Suggested order

[...]

PR #793, opened 2026-08-18
docs(decisions): ADR 0021 (Proposed) — a persistent sandbox per agent, beside the per-conversation model

jhgaylor:
## What

A design sketch, filed as a **Proposed** ADR. **Nothing here is built** and this PR does not intend to build it — it exists so the shape and the open questions are on record before any implementation is scoped.

People like the grokbot shape (one long-lived machine per agent; every conversation lands on the same disk, so the disk is the agent's memory). We chose one sandbox per conversation for good reasons that still hold (isolation is the only defence, fan-out wants fresh disks, simple lifecycle). The ADR proposes offering **both** as a per-agent `sandbox_mode`, defaulting to today's.

## Findings from the code survey

The schema is already N:1 (`Sandbox has_many :conversations`) and the reaper is N-aware, but seven paths hard-code 1:1. Two are latent bugs independent of this feature and are proposed as gate 1 on their own:

1. Reattach binds to the **head** of the sandbox's session list (`conversation_server.ex:876`) — no per-conversation session tag.
2. `/home/sprite/.env` carries `FOUNTAIN_TOKEN` / `FOUNTAIN_CONVERSATION_ID` at one shared path, rewritten on every reattach — should be per-turn process env.

The others: per-runtime shared `cwd` (gemini `--resume` guesses by workspace), terminate/park/destroy flip the shared row for one conversation, quota counts sandboxes as a proxy for runs, wake's fallback repoints one conversation and orphans co-tenants, `sandboxes` has no `agent_id`.

## Amended 2026-08-18 — companion design note #805

After asking "what would this look like if agent and sandbox had been decoupled from the start" (#805: the sandbox as a first-class machine, a conversation as a binding of an agent to one), three things changed in the ADR:

1. **`sandbox_mode` is a launch property, defaulted from the agent** — the `environment_id` shape from #783, not a fixed per-agent setting. Agent `foo` may run ephemeral in one conversation and on its home in another; a persistent-by-default agent fans out ephemeral children.
2. **The per-home lease is the machine's refcount** — park / ceiling / destroy become machine operations guarded by "does any conversation hold this?", not per-row flips on one conversation's say-so.
3. **`POST /conversations` accepts `sandbox_id`** to attach to an existing home (tenant-scoped, same agent identity). Every launch door — API, CLI, ACP `_meta`, Buzz provision, `fountain` skill — takes the same `sandbox_mode` / `sandbox_id`; channel resume takes the agent default.

Two open questions added (concurrency for v1; attach onto a suspended / runtime-changed home). Still nothing built.

## Proposed shape (see the ADR)

- `sandbox_mode: ephemeral | persistent` chosen per launch, defaulted per agent; a home is keyed on `(agent, environment, vault)` — the same identity `find_channel_conversation` already uses.
- **Turns serialized per home** (lease + bounded queue) rather than concurrent — the grokbot semantics, and it dissolves the shared-cwd/sqlite/port problems in one move while keeping `max_concurrent_sandboxes` meaningful.
- Lifecycle split by mode: terminate → revoke key only; idle → park only when no conversation on the home has a live server; ceiling → **force-suspend, not destroy**; agent-delete / "reset home" destroy.
- No fifth primitive: `agent_id`/`vault_id`/`mode` columns + partial unique index on `sandboxes`.
- gemini excluded until on the ACP path (#659). Buzz needs nothing new.

## Open questions — please weigh in before this moves past Proposed

- [ ] **Wake policy.** Should a persistent home wake on *any* inbound (a Buzz mention at 3am), or only on the owner's turns? Today wake is per prompt; a home makes "the agent is asleep" a visible product state.
- [ ] **Ceiling.** Force-suspend as proposed, or exempt homes from the continuous-run ceiling entirely (a chatty channel bot may never be idle for four hours)?
- [ ] **Identity key.** Should it ignore `environment_id` — one home per agent, environment override refused in persistent mode — for a simpler mental model at the cost of #783's flexibility?
- [ ] **Pricing.** Mode-based ("always-on agent") vs sandbox-minutes as today.
- [ ] **Concurrency.** Is serialized-turns-per-home acceptable for v1, or does the target user expect parallel turns on one machine?
- [ ] **Attach.** May `sandbox_id` attach onto a `suspended` home (wake it)? Onto a home whose agent has since changed runtime?

## Checks

- `okf validate decisions` — 0 errors; index regenerated with `scripts/decisions-index.sh`.
- ADR carries `stale_after: 2026-10-01` per the template rule for anything unbuilt.


jhgaylor:
Amended per the design discussion — see #805 for the holistic sketch this now converges on. Still Draft / Proposed; nothing built.

jhgaylor:
Superseded by #942. This PR was closed automatically when its head branch was renamed `adr/0021-persistent-agent-sandbox` -> `adr/0023-persistent-agent-sandbox`; the rename retargets some PRs but closed this one, and GitHub will not reopen it against a ref that no longer exists.

Nothing is lost: #942 carries the same two commits plus the renumber, and this thread's amendment note now lives in the ADR's own status block.

**Why the renumber:** this ADR was drafted as 0021 on 2026-08-18 at 04:27; the OAuth ADR (#819) was opened later the same day, took the same number, and merged at 18:14. The branch has been sitting on a duplicate id and a stale `index.md` ever since.

PR #797, opened 2026-08-18
fix(claude): fall back to the api key when the org disallows the oauth token (#655)

### LINKED ISSUE #655: A Claude OAuth token that the org disallows fails every turn, with no fallback
Found while trying to use a Claude subscription token for ACP turn testing.

## What happens

`Fountain.Runtimes.Claude.default_env/2` picks **exactly one** credential, deliberately:

```elixir
cond do
  is_binary(oauth) and oauth != "" -> [{"CLAUDE_CODE_OAUTH_TOKEN", oauth}]
  is_binary(api_key) and api_key != "" -> [{"ANTHROPIC_API_KEY", api_key}]
  true -> []
end
```

The comment explains why — mixing the two picked auth from the wrong var depending on CLI version — and that reasoning is sound. But it means a tenant who has stored *both* an OAuth token and an API key gets only the OAuth one, and if their Anthropic org disallows subscription access for Claude Code, every turn fails:

```
oauth_org_not_allowed
"Your organization has disabled Claude subscription access for Claude Code ·
 Use an Anthropic API key instead, or ask your admin to enable access"
```

Reproduced today on a live sprite over ACP: the session is created, then `session/prompt` fails with that error. The API key sitting in the same credentials row would have worked.

## Why it's worth fixing

0008 calls `claude_code_oauth_token` "preferred for the `claude` runtime" because it bills a subscription instead of metered API usage — so the preference is right, and a tenant is likely to store it precisely *because* they were told it's better. The failure then looks like Fountain is broken rather than like an org policy, and there is a working credential one field away.

## Options

- **Fall back on this specific error.** `oauth_org_not_allowed` is a stable, well-named error kind. Catch it at turn level and retry with the API key if one exists, marking the OAuth credential as unusable so it isn't retried every turn.
- **Validate at credential-save time.** Cheapest signal: a token that the org disallows can be detected with one call when it's stored, and the user told immediately rather than at their first turn.
- **Surface it properly.** At minimum the turn failure should say what it says on the tin — that the credential is refused by the org and an API key would work — rather than surfacing as a generic turn failure.

Not urgent; nobody has hit it in production that we know of. Worth doing before ACP conversion makes the claude runtime the common path, since the ACP error arrives as a JSON-RPC `-32603` and is even less legible than the CLI's.

BinaryBourbon:
## Summary

Closes #655.

`Fountain.Runtimes.Claude.default_env/2` deliberately picks the OAuth token over the API key when both are on file — it bills a subscription instead of metered usage. But when the Anthropic org that owns the OAuth token has disabled Claude subscription access for Claude Code, every turn refuses with `oauth_org_not_allowed`, and there was no way out short of a human noticing and swapping credentials by hand — even though the API key sitting in the same `inference_credentials` row would have worked.

I went with the issue's "fall back on this specific error" option, since it's the one that actually unblocks a tenant without them touching anything:

- **`Fountain.Runtimes.ACP.Peer`** recognizes Claude's `oauth_org_not_allowed` error kind (a generic JSON-RPC `-32603` on the wire — the kind lives in the error body's text, not `code`, so detection is a substring search over the whole payload rather than a `get_in` on a specific key I haven't seen pinned down against a live org-disallowed account). Scoped to the `:prompt` call specifically — a session-setup call failing the same shape is a different, pre-existing problem the auth-retry path above it already owns.
- **`Fountain.Runtimes.Claude.fall_back_to_api_key/2`** swaps `CLAUDE_CODE_OAUTH_TOKEN` for `ANTHROPIC_API_KEY` in an already-built env list. Kept out of `default_env/2` on purpose — that function runs at provisioning, before any turn has been attempted, so it has no way to know the token is bad.
- **`ConversationServer`** applies the swap to the running server's `sprite_env`/`inference_credentials` for the rest of the conversation, not just a silent retry of the one failed turn — so both the failure and the fix are visible in the transcript (the turn's stage event says plainly what happened and what changed, or that there's no API key to fall back to). Every ACP turn is a fresh spawn reusing `state.sprite_env` (confirmed by reading the turn-spawn path), so this is enough for the very next prompt to succeed with no respawn/retry surgery needed.
- Scoped to the running conversation, deliberately not persisted to the credentials row: a fresh conversation tries OAuth again, so an org policy that later reverts self-heals instead of staying pinned to a credential this fix disabled.

## Test plan

- [x] `Fountain.Runtimes.ACP.PeerTest`: the `oauth_org_not_allowed` error at `session/prompt` is reported as a distinct `{:oauth_org_not_allowed, detail}` failure reason; the same error kind at session setup is *not* special-cased (only `:prompt` is).
- [x] `Fountain.Runtimes.ClaudeTest`: `fall_back_to_api_key/2` — swaps when an API key is on file, leaves the env untouched when there isn't one, doesn't duplicate an existing `ANTHROPIC_API_KEY` entry.
- [x] `Fountain.Conversations.ConversationServerACPTest` (real `ConversationServer` + real ACP peer, driven end-to-end over stdout): when an API key is on file, the failed turn's stage event says so, and the *next* turn's real spawn carries `ANTHROPIC_API_KEY` with `CLAUDE_CODE_OAUTH_TOKEN` gone; when there's no API key on file, the failure message says so instead of silently repeating.
- [x] `mix test` (full suite) — 2929 tests, 0 failures
- [x] `mix format --check-formatted`
- [x] `mix compile --warnings-as-errors`
- [x] `mix credo --strict` — no issues
- [x] `scripts/sobelow.sh` — clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -438,6 +438,21 @@ upgrade, is in
 
 ### Fixed
 
+- **A Claude OAuth token the org disallows now falls back to the Anthropic
+  API key instead of failing every turn (#655).** `Fountain.Runtimes.Claude`
+  picks the OAuth token over the API key when both are on file — it bills a
+  subscription instead of metered usage — but an org that disables Claude
+  Code's subscription access rejected every turn with no way out short of a
+  human noticing and swapping credentials by hand, even though a working API
+  key sat in the same row. The ACP peer now recognizes Claude's
+  `oauth_org_not_allowed` error kind on the `session/prompt` call
+  specifically (session setup failing the same way is a different problem),
+  and `ConversationServer` swaps the OAuth token for the API key in the
+  running server's env for the rest of the conversation — the failed turn
+  says so plainly, and the next prompt succeeds without editing anything. A
+  fresh conversation still tries OAuth first, so a policy that later reverts
+  self-heals instead of staying pinned to a credential this fix disabled.
+
 - **Team page: a teammate stayed named after its agent, not its first
   message.** The first turn's auto-generated conversation title (a summary
   like "Elixir Tic Tac Toe Game Development") is what #807 shows as the
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -1583,6 +1583,66 @@ defmodule Fountain.Conversations.ConversationServer do
      })}
   end
 
+  # #655: the org has refused this account's Claude OAuth token. Left alone,
+  # every further turn on this conversation would fail the exact same way —
+  # nothing about the token changes between attempts — while the Anthropic
+  # API key sitting in the same `inference_credentials` row would work. Swap
+  # it in for the rest of this server's life (not just a silent retry of this
+  # one turn, so the failure and the fix are both visible in the transcript)
+  # rather than leave the tenant to rediscover the swap by hand. Scoped to
+  # this running conversation on purpose: a fresh conversation tries OAuth
+  # again, so a policy that later reverts self-heals instead of staying
+  # pinned to a credential this fix disabled.
+  def handle_info(
+        {:acp, ref, {:failed, {:oauth_org_not_allowed, detail}}},
+        %{current_command_ref: ref, runtime_module: Fountain.Runtimes.Claude} = state
+      ) do
+    Logger.warning("conv #{state.conversation_id}: Claude OAuth token refused by org: #{detail}")
+
+    fallback_env =
+      Fountain.Runtimes.Claude.fall_back_to_api_key(state.sprite_env, state.inference_credentials)
+
+    switched? = fallback_env != state.sprite_env
+
+    # The API key was never in the sprite env before now, so `build_sprite_env`
+    # never registered it for redaction — do it here, or the very value this
+    # fix injects prints in plaintext into `log_events`. Registered as a union
+    # with the outgoing env, not a replacement: the refused OAuth token is
+    # still sitting in the sprite's `/home/sprite/.env` until a wake rewrites
+    # it, so it stays worth scrubbing.
+    Fountain.Conversations.Redaction.put(
+      state.conversation_id,
+      state.sprite_env ++ fallback_env
+    )
+
+    state = %{
+      state
+      | sprite_env: fallback_env,
+        inference_credentials: Map.delete(state.inference_credentials, :claude_code_oauth_token)
+    }
+
+    message =
+      if switched? do
+        "Your organization has disabled Claude subscription (OAuth) access for Claude Code. " <>
+          "Switched to the Anthropic API key on this account for the rest of this " <>
+          "conversation — send your prompt again."
+      else
+        "Your organization has disabled Claude subscription (OAuth) access for Claude Code, " <>
+          "and no Anthropic API key is on file for this account. Add one in Settings, then " <>
+          "try again."
+      end
+
+    {:noreply,
+     finish_acp_turn(
+       state,
+       "failed",
+       %{"error" => message, "acp.oauth_org_not_allowed" => true},
+       %{
+         reason: message
+       }
+     )}
+  end
+
   def handle_info({:acp, ref, {:failed, reason}}, %{current_command_ref: ref} = state) do
     Logger.error("conv #{state.conversation_id}: acp peer failed: #{inspect(reason)}")
 
--- a/apps/fountain/lib/fountain/runtimes/acp/peer.ex
+++ b/apps/fountain/lib/fountain/runtimes/acp/peer.ex
@@ -335,19 +335,31 @@ defmodule Fountain.Runtimes.ACP.Peer do
   defp handle_message({:error_response, id, error}, state) do
     {tag, state} = pop_pending(state, id)
 
-    if session_setup?(tag) and auth_error?(error) and not state.authenticated? and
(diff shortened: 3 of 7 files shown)


jhgaylor:
Reviewed. The substance is right and the wiring is sound — two things to fix before this can merge.

## What I verified holds

I checked the mechanism rather than the description, because the obvious worry is that `state.sprite_env` is write-only after provisioning. It isn't: `conversation_server.ex:2327` passes `env: state.sprite_env` into `Fountain.Sandbox.spawn/4` on **every turn**, so swapping the field genuinely does reach the next prompt. The "send your prompt again" wording in the user-facing message is accurate, and scoping the swap to the running server (so a fresh conversation retries OAuth and a reverted org policy self-heals) is the right call.

## 1. It conflicts with main

Three files, from ~35 commits on `main` since the merge base:

- `CHANGELOG.md` — trivial
- `apps/fountain/lib/fountain/runtimes/claude.ex` — content conflict
- `apps/fountain/test/fountain/runtimes/claude_test.exs` — **add/add**: main grew this same file independently, so the two versions have to be reconciled by hand

## 2. The detection is likely matching the wrong string

This is the blocker. The trigger is:

```elixir
defp oauth_org_not_allowed?(error),
  do: error |> inspect() |> String.contains?("oauth_org_not_allowed")
```

and the comment above it says the payload placement "has not been pinned down against a live org-disallowed account."

`oauth_org_not_allowed` **is** a real kind — it appears in Claude Code's `errors.ts` as a member of a union (`'token_revoked' | 'oauth_org_not_allowed' | 'auth_error'`). But that kind is mapped to a user-facing message before it leaves:

> `Your account does not have access to Claude Code. Please run /login.`

Since this arrives over ACP as a generic `-32603 Internal error` — as the PR's own comment notes — the body most likely carries that **message**, not the **kind**. If so the substring never matches, `fall_back_to_api_key/2` never runs, and #655 stays broken while the code reads as though it were fixed. That is the failure mode CLAUDE.md warns about for describing unbuilt behavior as existing, and it would be invisible: no test fails, because the module tests construct the error payload this code expects rather than one captured from a real refusal.

Suggested fix — match either form, and say why:

```elixir
@oauth_org_not_allowed_markers [
  "oauth_org_not_allowed",
  "does not have access to Claude Code"
]

defp oauth_org_not_allowed?(error) do
  text = inspect(error)
  Enum.any?(@oauth_org_not_allowed_markers, &String.contains?(text, &1))
end
```

Best of all would be one captured payload from a live org-disallowed account pasted into the PR, with the test built from it — then the marker list is evidence rather than a guess.

## 3. Minor: a stale token stays in the env file

`Provisioning.write_env_file/2` wrote `CLAUDE_CODE_OAUTH_TOKEN` into `/home/sprite/.env` at provisioning, and nothing rewrites it here. The turn's spawn env wins, so this doesn't affect the fix — but a `setup_script` that does `source .env` would pull the dead token back in. Worth a comment noting the limitation even if you don't address it.

## Verdict

Keep this open — it's a real fix for a still-open bug with good tests, not something to close. Rebase, widen the match (ideally against a captured payload), and it's mergeable.

jhgaylor:
Rebased onto `main` (the branch was 84 commits behind). Three conflicts, all additive — both sides had appended to the same spot:

- **`claude.ex`** — `main` filled in `write_config/2` with the MCP provisioning from #837/#838; this branch had added `fall_back_to_api_key/2`. Kept both, and dropped the now-false `# MCP servers travel in session/new's mcpServers param` comment this branch's version still carried.
- **`claude_test.exs`** — add/add: both sides created the file from scratch. Merged into one module (MCP provisioning tests + `default_env/2` + `fall_back_to_api_key/2`).
- **`CHANGELOG.md`** — both added under `[Unreleased] → Fixed`. Kept both.

Two changes beyond a mechanical rebase, both caused by things that landed on `main` after this branch was cut:

**1. The swapped-in API key was not registered for output redaction.** `Fountain.Conversations.Redaction` (added after this branch) is populated only inside `build_sprite_env/5`. The #655 fallback writes `state.sprite_env` directly, so it never goes through that — meaning the one credential this fix injects into the sandbox was the one credential that would print in plaintext into `log_events`. Added the registration in the fallback handler, as a *union* with the outgoing env rather than a replacement: the refused OAuth token is still in the sprite's `/home/sprite/.env` on disk until a wake rewrites the file, so it stays worth scrubbing.

Covered by a new test; verified it fails without the fix:

```
1) test org-disallowed oauth (#655) the swapped-in api key is registered for output redaction
   Assertion with in failed
   code:  assert "api-key-long-enough" in registered
   left:  "api-key-long-enough"
   right: ["ftn_2a1b...", "efb7e999-...", "oauth-token-long-enough", "http://localhost:4000", "aod@local"]
```

**2. The new tests were making live HTTPS calls to `api.anthropic.com`.** These are the first tests in this file to supply key-shaped credentials, and turn 1 fires `TitleGenerator.generate/2` — which previously bailed at `:no_credentials` because every other test here leaves credentials empty. The suite was logging real `401 Invalid bearer token` responses. Stubbed the boundary in `start_acp_turn/3`; nothing in this file asserts on titles.

## Verification

Full gate, locally, on the rebased tree:

- `mix test` — 6 doctests, 3 properties, **3286 tests, 0 failures**
- `mix format --check-formatted`
- `mix compile --warnings-as-errors`
- `mix credo --strict` — no issues
- `scripts/sobelow.sh` — clean
- `MIX_ENV=dev mix dialyzer` — Total errors: 0
- `mix deps.unlock --unused` — no change

PR #876, opened 2026-08-20
A TypeScript SDK: fountain.run(prompt, {agent, environment, vault})

jhgaylor:
The REST API describes machinery — conversations, turns, log events, blocks. Every integration that has ever talked to it wrote the same wrapper first to get back to the job:

- `integrations/hermes/fountain/tools.py` — `fountain_run(agent, prompt, vault, environment, …)`
- `cli/internal/cmd/conv.go:77` — *"`run` is a top-level shortcut: create + stream until done"*
- the bundled skill, the same thing in shell

Three implementations of one function is the tell that the function is the product surface and REST is the transport. This is that function, once.

```ts
import { Fountain } from "fountain-sdk";

const run = await new Fountain().run("Upgrade us to Phoenix 1.8 and open a PR", {
  agent: "reposage",
  vault: "github-bot",   // the token lands in the sandbox, never in the prompt
});

console.log(run.text, run.url);
```

## The design calls worth arguing about

**The second argument is the pitch.** `fountain.run(prompt, …)` alone looks exactly like `anthropic.messages.create()` and explains nothing. `{ agent, environment, vault }` is the part that is Fountain, so it is what you read first. `vault` is the differentiator no model API has: the secret is attached at spawn, so it is not in the prompt, not in the model's context, and not in the log feed the SDK reads.

**A handle, not a string.** `run()` returns something you can `await` (the answer), `for await` (lifecycle events), or read as `.textStream` (the words) — three views of one run, no second request behind any of them. `resume(id).send()` continues in the same sandbox with the same checkout and session.

**Names, not ids.** `agent`/`vault`/`environment` take either; resolution is exact then unique-prefix, and a miss names what the account actually has. A UUID in a README kills the demo.

**Two things that look like polish and are not:**

- The turn-following rules — keep only your own turn's events, keep `text` blocks, join `acp` chunks with nothing and `stdout` rows as paragraphs, break after a tool call — are ported from the Hermes plugin, which learned them against real runtimes. Get them wrong and every transcript reads as one run-on sentence.
- The SSE reader resumes from the last event id when a connection dies. A deploy, a sandbox wake or a proxy timeout ends a stream mid-turn; without the cursor the reconnect either replays output or loses it. #772 is what that costs in production. There is a test that kills the connection mid-turn and asserts the resume header.

## Verification

Not just the fake server:

- **Live run** against `fountain.inevitable.fyi` — 28s, streamed, tool call detected, clean text, correct transcript URL.
- **Cold resume** — a brand-new client, no memory of the first run, sent turn 2 into the same sandbox in 4.8s and got an answer that depended on turn 1 (`uname -a && echo READY`). Both sandboxes terminated afterwards.
- **39 tests** against an in-process fake Fountain over real HTTP and real SSE. Three defects surfaced during the writing and were fixed in the SDK rather than the tests: unhandled rejections on derived promises, a retry that counted one failure twice, and a body left open when a consumer breaks mid-turn.

## Also in here

- `docs/sdk.md`, nav mirrored in `mkdocs.yml` and `docs.ex` (`docs_test` green, `mkdocs build --strict` exits 0).
- A pointer at the top of `docs/api.md` for people who land on the endpoint list first, and an `llms.txt` entry so an agent writing a script finds the SDK before it finds `curl`.
- CI steps in the `checks` job beside the Go and Hermes ones: install, typecheck, test, build.

## Not done, deliberately

- **Not published to npm.** The package name `fountain-sdk` is free, but publishing is your call — the docs say "not on npm yet" and give the `npm link` path instead of claiming an install that would 404.
- **No generated OpenAPI layer.** `fountain.request(method, path, opts)` is the escape hatch for the other ~55 paths today; generating types from `/api/openapi.json` into `fountain.api.*` is the natural next step, not a blocker.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -461,6 +461,33 @@ jobs:
       - name: Hermes plugin tests
         run: python3 -m unittest discover -s integrations/hermes/tests -v
 
+      # The TypeScript SDK (sdk/typescript). Node runs the .ts sources
+      # directly, so `npm test` needs nothing but the dev dependency on tsc —
+      # which is also what enforces `erasableSyntaxOnly`, the setting that
+      # keeps those sources runnable without a build step.
+      - name: Set up Node
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: 24
+          cache: npm
+          cache-dependency-path: sdk/typescript/package-lock.json
+
+      - name: SDK install
+        working-directory: sdk/typescript
+        run: npm ci
+
+      - name: SDK typecheck
+        working-directory: sdk/typescript
+        run: npm run typecheck
+
+      - name: SDK tests
+        working-directory: sdk/typescript
+        run: npm test
+
+      - name: SDK build
+        working-directory: sdk/typescript
+        run: npm run build
+
       # `gofmt -l` prints offending files and exits 0, so the failure has to
       # come from the emptiness check — print the list first so the log says
       # which files to run gofmt on.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,19 @@ upgrade, is in
 
 ## [Unreleased]
 
+### Added
+
+- **A TypeScript SDK** (`sdk/typescript`; not yet published to npm). Running
+  an agent is one call — `fountain.run(prompt, { agent, vault, environment })` —
+  which opens a conversation, follows the turn and hands back the answer, the
+  tools used and a URL a human can watch. The handle it returns can be awaited,
+  iterated for lifecycle events, or read as a text stream; `resume(id).send(...)`
+  continues in the same sandbox. Every integration that has ever talked to
+  Fountain wrote this wrapper first (the Hermes plugin, `fountain run`, the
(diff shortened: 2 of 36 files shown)


jhgaylor:
Follow-up commit: found a real defect while double-checking an assumption I'd guessed at rather than verified.

The early-exit check tested for `sandbox` paired with a terminal *state* — but `LogEvent`'s state set is closed (started/done/failed/interrupted), so `publish_stage` never emits that. Dead code. Meanwhile the actual "the conversation is over" signal, `provision`/`failed` (six call sites), stops the GenServer — no turn event is ever coming. `run()` waited for one, and since the default timeout is 0, it waited **forever**, indistinguishable from a slow agent.

Moving from polling to SSE is how the hole opened: the Hermes plugin polled conversation status every 2s as its safety net, and that went away with the poll loop.

The fix asks instead of guessing — any stage event that *might* mean the end (any `failed`, any `terminate`, any `sandbox`) triggers one status fetch, and the conversation's own status decides. Correct whatever the vocabulary does next, and it keeps `setup`/`failed` (which doesn't necessarily stop anything) from ending a turn early. The run then reports `state: "failed"` with the stage's reason rather than an empty answer that reads as success.

Two tests added, and I confirmed the first hangs for the full test timeout against the old predicate.

PR #878, opened 2026-08-20
A guided tour: an agent that opens a pull request

jhgaylor:
The docs had reference material and no narrative — nothing that walks from an empty account to a thing that did work. `docs/tour.md` is that page, and it ends where the value actually is.

## What it walks through

Environment (the repo, cloned) → vault (the token, never in the prompt) → agent (runtime, model, standing instructions) → run → **follow-up**. About forty lines.

The payoff is step 5:

```
turn 2 | done | +13s
https://github.com/you/your-app/pull/1
```

Thirteen seconds, same PR, second commit. Nothing re-cloned, nothing re-explained, no second branch. *"That is the difference between an agent that has a computer and an agent that has a context window."*

## Every number on the page is real

Run against production, not written from imagination:

- private repo cloned into a sandbox, `--version` added to a small CLI, README updated, branch pushed, PR opened — **43s**, most of it provisioning
- `resume(id).send(...)` adding a `-v` alias as a second commit on the same PR — **13s**
- the tool list in the page (`Terminal`, `Read File`, `Edit`, …) is the actual sequence

Everything created for it is deleted: agents, vaults, environments, both conversations, and the scratch repo.

## The wasted run that became a warning box, and a spec fix

The first attempt failed: the agent opened on an empty directory. `Repository` declares only `url` and `mount_path`, but `Provisioning.clone_https/4` reads two more — `secret_key` (which secret the clone authenticates with) and `ref` (which branch). A private repository could not be expressed by a client generated from the spec.

Worse, omitting `secret_key` doesn't fail the request: the clone fails *inside the sandbox*, provisioning continues, and your agent tells you it can't find your repo. That's now a warning box on the page, and both fields are declared in the schema.

## One thing worth chasing, deliberately not fixed here

That failed clone published **`clone/done`**, while the logged output plainly read `fatal: could not read Username for 'https://github.com'`. `clone_https/4` does check the exit code (`if code == 0, do: :ok, else: {:error, …}`), so the sandbox exec appears to have reported `0` for a command that exited non-zero. If that generalises it is a nasty class of silent failure — but it needs a focused repro before anyone changes anything, and I didn't want to guess at it inside a docs PR.

## Also

`sdk/typescript/examples/pull-request.ts` — the same tour, runnable. Nav mirrored in `mkdocs.yml` and `docs.ex`; `docs_test` and `mkdocs build --strict` both green.

Note: this branch touches only my files. There is in-progress work in the same checkout (adding `lumis` syntax highlighting to the docs renderer) that I left alone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,12 @@ upgrade, is in
   `conversation.history()`, `fountain.catalog()`, `fountain.search()` and
   `fountain.events()`.
 
+- **A guided tour** (`docs/tour.md`): an agent that clones a repository, opens
+  a pull request, and then amends the same PR on a follow-up turn thirteen
+  seconds later because the sandbox is still up. Every number and output on the
+  page came from running it; `sdk/typescript/examples/pull-request.ts` is the
+  same thing runnable.
+
   Errors are now keyed on the server's `error` code rather than the status,
   because that is how every app branches: `ConversationBusyError` (a 400),
   `QuotaExceededError` (a 429, carrying `activeSandboxes`/`limit`),
@@ -46,6 +52,13 @@ upgrade, is in
   exists. Generating it immediately found one: see below.
 
 ### Fixed
+
+- **`Repository` declared neither `secret_key` nor `ref`.**
+  `Provisioning.clone_https/4` reads both — `secret_key` names the secret the
+  clone authenticates with, `ref` picks a branch — so a private repository
+  could not be expressed by a client generated from the spec. Without
+  `secret_key` the clone fails *inside the sandbox*: provisioning continues,
+  and the agent opens on an empty directory.
 
 - **`AgentRequest` did not declare `allowed_environment_ids`.** `AgentUpdate`
   declared it and `Agent.changeset/2` has cast it since the allowlist shipped,
--- a/apps/fountain/lib/fountain/docs.ex
+++ b/apps/fountain/lib/fountain/docs.ex
@@ -61,6 +61,7 @@ defmodule Fountain.Docs do
     {"CLI reference", "cli.md"},
     {"API reference", "api.md"},
     {"TypeScript SDK", "sdk.md"},
+    {"Guided tour — opening a PR", "tour.md"},
     {"LLM integration", "llm-integration.md"},
     {"Changelog", "changelog.md"}
   ]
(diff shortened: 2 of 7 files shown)


jhgaylor:
Chased the `clone/done` oddity I flagged in the description — it has an answer, and it is worse than a clone bug: **#880**.

`Sprites.exec/4` reports exit code 0 for everything. Minimal repro: a `setup_script` of `echo 'about to fail'; exit 42` records `setup/done {"exit_code":0}`.

Evidence it is provider-specific, from production:
- every `exit_code` ever stored in `log_events` is `0` (533 of them, no non-zero value has ever been recorded)
- the two clones that printed `fatal:` and still got `clone/done` were both **sprites** (2026-05-29 DNS, 2026-08-20 auth)
- the one `clone/failed` ever recorded carries a real code — `{:clone, …, 1}` — and ran on the **runner** provider

So `Provisioning`'s exit-code branch is fine; what sprites hands it is not. Sprites is the default and 791 of 810 sandboxes.

Nothing in this PR changes because of it — the warning box about `secret_key` stands on its own, and is arguably more necessary now, since the underlying failure cannot surface as an error until #880 is fixed.

jhgaylor:
Closing: everything in this PR is already on `main`, but not by way of this PR — and the how is worth recording.

Two collisions happened in one shared checkout.

**1. #879 absorbed this branch's work.** `e635242` merged with `docs/tour.md` (193 lines), `sdk/typescript/examples/pull-request.ts` (74), the `Repository` `secret_key`/`ref` schema fix, the regenerated types, and both nav entries — all of it authored here, none of it reviewed as part of that PR, which was about the docs theme.

**2. My follow-up went straight to `main`.** By the time I committed the copy button and the one-script section, the checkout had been switched to `main` underneath me. I did not re-check the branch before `git commit` and `git push`, so `f037c92` is a direct push — against the repo's own rule that everything goes through a PR. That one is mine, not the shared tree's.

`main` is coherent (typecheck and 65 SDK tests pass locally; Docs and Secret Scanning already green on `f037c92`, CI in progress). Nothing needs re-landing. If you would rather `f037c92` had gone through review, say so and I will revert it on a branch and re-open it properly.

The lesson for both sessions, from `git log`: `git add -A` on a shared checkout sweeps up whatever the other agent has in flight, and a branch can change under you between one command and the next.

PR #885, opened 2026-08-21
The tour's snippet was not in the image build

jhgaylor:
**Every image build on `main` has failed since f037c92**, so nothing has
deployed — including #883's parser fix.

`docs/tour.md` now includes `sdk/typescript/examples/pull-request.ts` verbatim
via `--8<--`. Snippets are expanded at compile time from the repo root, and the
Dockerfile's build stage copies `docs/` and `CHANGELOG.md` but not `sdk/`:

```
> [build 13/13] RUN mix compile && ...:
== Compilation error in file lib/fountain/docs.ex ==
** (File.Error) could not read file "/app/sdk/typescript/examples/pull-request.ts"
```

CI never saw it — the suite compiles from a full checkout, where the file is
present. Only `docker build` has the narrowed context.

## The fix

`COPY sdk/typescript/examples ./sdk/typescript/examples`, plus a check in
`docs_test.exs` that the two lists agree: for every `--8<--` path in the corpus,
assert the Dockerfile's build stage copies it (exact file or parent directory).
Deleting the new COPY line fails that test naming the exact path — which is the
point, since the real failure is a compile error inside a build nobody watches,
with a green PR gate above it.

## Verification

Built the image from this tree and ran it `--read-only --network none`:

- compiles (the tour page embeds the script: `tour_has_script=true`)
- still highlights: `spans=5` on a `ts` fence

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,6 +58,13 @@ COPY ee ./ee
 # ships: the release never contains docs/ as files.
 COPY docs ./docs
 COPY CHANGELOG.md ./
+# A docs page can pull a file in with `--8<--` (docs/tour.md includes the SDK
+# example verbatim, so the page cannot drift from the script that produced its
+# output). Those paths are read at compile time from the repo root, so every
+# one of them has to be here too — docs_test.exs checks this list against the
+# snippet paths the corpus actually uses, because the failure is a compile
+# error inside `docker build` and nowhere else.
+COPY sdk/typescript/examples ./sdk/typescript/examples
 
 # Lumis downloads a language's tree-sitter parser on first use and caches it
 # under its own priv/. The deployment runs read-only (deployment.yaml sets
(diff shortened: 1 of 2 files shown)


jhgaylor:
Heads up: this is now a duplicate — #884 merged the same fix a few minutes ago as `93a9e93c`, and `main` builds again.

Same diagnosis and very nearly the same shape: a `COPY` for `sdk/typescript/examples` beside `COPY docs ./docs`, plus a `docs_test` guard that scans every `--8<--` in `docs/**/*.md` and asserts the Dockerfile copies a path covering it.

Worth a look before closing: if your guard covers a case mine misses, that is the version to keep — mine treats a COPY source as covering `path` when `path == source` or `path` starts with `source <> "/"`, and ignores `COPY --from=…` lines. Happy to swap in yours if it is stricter.

Context on how we both ended up here: `f037c92` (the commit that broke the build) was a direct push to `main` from a shared checkout that had been switched to `main` underneath me — so there was no PR to signal that the fix was already in flight.

jhgaylor:
Superseded by #884, which landed the same fix (the same COPY line and an equivalent docs_test guard) while this was in review. Closing rather than merging — main builds images again, and the build that shipped 93a9e93 ran the new 'image can highlight code' step green: `spans=5`.

PR #900, opened 2026-08-21
The coverage gate rendered 129MB of HTML nobody read

jhgaylor:
The `coverage` job is a **barrier** — every partition must finish before it starts, and nothing runs alongside it — so its 39s is 39s of CI wall-clock. About 17s of that is `mix test.coverage`, and ~90% of *that* is rendering an HTML report: **530 files, 129MB**, into a directory that dies with the runner. The job's only output is a threshold verdict.

## The swap

`scripts/coverage-gate.exs` computes the same number from the same `coverage.exs`, without the report:

| | total | time |
|---|---|---|
| `mix test.coverage` | **85.46%** | 7.4s wall / 18.7s user, +129MB HTML |
| `scripts/coverage-gate.exs` | **85.46%** | 3.6s |

## Why the equality needs stating carefully

That the two agree is the entire argument for this change, so it is worth recording how easily they don't. `:cover.analyse(mod, :coverage, :line)` returns **one entry per clause**, so a line with several clauses appears several times. Summing the entries gives **81.32%** against the real 85.46% — not a rounding difference, but *below the 85% threshold*: a gate that fails builds it should pass.

The rules that make it match:
- a line counts **once**, and counts as covered if **any** clause on it ran
- **line 0** is module-level bookkeeping, not executable, and is dropped

## Both failure paths are exercised

A gate that cannot fail is worse than no gate:

```
threshold raised to 99%  → exit 1, "coverage 85.46% is below the 99% threshold"
no .coverdata present    → exit 1, "nothing was merged"   (not "100% of nothing")
```

## What is lost, and what replaces it

The per-module table goes with the HTML, so the script prints the **fifteen least-covered modules** — a failure still says where to look:

```
coverage: 85.46% (8913/10430 lines, 530 modules, threshold 85%)
least covered:
     0.0%  Mix.Tasks.Fountain.VerifyLifecycle (0/203)
     7.1%  Fountain.Sandbox.E2B.Errors (1/14)
    15.4%  Fountain.Sandbox.Daytona.Errors (2/13)
    ...
```

`mix test.coverage` still works locally for the full table and the HTML.

## Drift

The script leans on `:cover` semantics rather than Elixir's private summary code. Elixir is pinned in `.tool-versions` (1.19.2-otp-28), so those cannot change without a deliberate bump — the script's header and CLAUDE.md both say to re-verify the totals agree when bumping.

Expected: coverage job **39s → ~25s**, all of it off the serial tail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -425,28 +425,6 @@ jobs:
       # itself compiles nothing, but Mix runs loadpaths underneath it and will
       # build the whole dependency tree into an empty _build first. Measured on
       # a 10-core machine: 3.7s warm, 38.5s cold.
-      - name: Restore deps cache
-        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v5
-        with:
-          path: deps
-          key: ${{ runner.os }}-mix-test-${{ hashFiles('**/mix.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-mix-test-
-
-      - name: Restore build cache
-        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v5
-        with:
-          path: _build
-          key: ${{ runner.os }}-build-test-${{ steps.beam.outputs.otp-version }}-${{ steps.beam.outputs.elixir-version }}-${{ hashFiles('**/mix.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-build-test-${{ steps.beam.outputs.otp-version }}-${{ steps.beam.outputs.elixir-version }}-
-
-      - name: Install dependencies
-        run: mix deps.get
-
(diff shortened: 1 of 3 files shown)


jhgaylor:
Reworked after measuring the first CI run: the gate produced the right number (85.46%, passed) but the step still took **12s**, not the ~3s expected, so the job only moved 39s → 37s.

The log showed why — `mix run` was **recompiling the app** (`Compiling 10 files`) for a script that never calls it.

The script reads `.coverdata` and `coverage.exs` and touches nothing else: no deps, no `_build`, no app modules. So it now runs as `elixir scripts/coverage-gate.exs`, and the coverage job drops the deps cache, the build cache and `mix deps.get` along with it — five steps left: checkout, Set up Elixir, download exports, check every partition reported, enforce.

Locally: `elixir scripts/coverage-gate.exs` → same 85.46%, 4.3s, no mix involved.

Issue #798, opened 2026-08-18
Rethink the billing model: capacity tiers + metered sandbox-hours (discussion)


jhgaylor:
Discussion issue — no decision yet. This is a proposal to react to, not a plan to execute. An ADR comes later, once the shape is agreed.

## Where billing stands today

Read from `ee/lib/fountain/billing.ex`, ADR 0006/0008/0017, `Fountain.Quotas`, `Fountain.Billing.UsageEvent`.

- **One flat monthly subscription** (`STRIPE_PRICE_ID`) with a 14-day trial. Statuses `trialing / active / past_due / canceled / comped`. No `plan` column, no tier concept anywhere in code. ADR 0006 explicitly left the price *shape* undecided.
- **The only per-user entitlement is `users.max_concurrent_sandboxes`** (default 5, admin-set). It's a de-facto manual tier: an operator can hand a customer a bigger cap, but nothing ties it to what they pay.
- **Metering exists but drives nothing.** `usage_events` records `sandbox_provisioned / sandbox_provision_failed / turn_started / sandbox_terminated{duration_ms}`; `Billing.usage_summary/3` rolls it into conversations / turns / sandbox-minutes for BillingLive and the admin overview only. Known gaps: suspended time is invisible (#665); minutes accrue only at termination, so a long-running sandbox contributes zero until it dies and month boundaries are wrong; no `provider` in metadata, so hours can't be weighted per Sprites/E2B/Daytona.
- **Inference is not metered and not paid by Fountain** by design (ADR 0008 — BYO credentials). The ACP tracer dropped token/cost capture deliberately (#637): ACP's `session/prompt` carries no usage block.
- **The gate protects spend, not features** (0006 addendum 2026-08-05): only conversation start / wake / per-turn call `Billing.check_active/1`. Worth keeping under any new model.
- **Stripe surface already built:** customer, Checkout, Portal, idempotent webhooks, invoice listing, trial extend/resync/comp. Stripe usage-based pricing (Meter Events) is *not* wired, but everything around it is.

Net: nothing more advanced is possible with config alone — you can change the price and hand-set caps. But the distance to a real model is short because the metering skeleton and the enforcement point exist.

## What our cost shape actually is

The model should mirror this:

1. **Sandbox active seconds** — the Sprites/E2B/Daytona bill. Dominant *variable* cost, priced differently per provider.
2. **Peak concurrency** — capacity exposure on the shared provider account. Already enforced by `Quotas`.
[...]

PR #991, opened 2026-08-22
feat(billing): plans priced on concurrent sandboxes, contacts as an add-on (ADR 0026)

jhgaylor:
Closes the "capacity tiers" half of #798. The metered sandbox-hours half is
deliberately left for a later revision; see ADR 0026 for why tiers can ship
without it.

## What this does

Fountain sold one flat price, and the only per-tenant entitlement — the
concurrent-sandbox cap — was a column an operator hand-set, tied to nothing
anybody paid. There are now three tiers that differ on exactly that axis and
nothing else.

| | solo | team | scale | legacy |
|---|---|---|---|---|
| concurrent sandboxes | 5 | 15 | 40 | 15 |
| teammate contacts | 3 | 10 | 40 | 10 |
| price | $29 | $79 | $199 | $29 (closed) |

Every plan carries the whole product. A feature-gated ladder would need a
second enforcement mechanism and would put the subscription gate back in the
business of protecting features rather than spend, which the 0006 addendum
deliberately undid.

## The five things worth reviewing

**The cap column became an override.** `Quotas.sandbox_limit/1` reads the
plan; `users.sandbox_limit_override` wins when set. Null means "the plan's",
so an upgrade takes effect with nobody touching the column, and the two
things a plan cannot express still work (raise it for a trusted tenant, drop
it to zero during abuse).

**The Postgres column keeps its name.** The Elixir field is
`sandbox_limit_override`, mapped with Ecto's `:source`. A rename breaks every
pod still on the previous release for the length of a rolling deploy; making
the column nullable does not, because the previous release already reads a
null as "use the default". The worst an old pod does mid-rollout is apply the
old default of 5, to exactly the rows the migration nulled.

**Grandfathering is a plan, not a migration.** Every account with a Stripe
customer goes to `legacy`, pinned to the original `STRIPE_PRICE_ID`, carrying
**Team's** capacity at Solo's price. Nobody's cap fell. It orders below every
public plan, which is what makes it structurally upgrade-only — nothing in
the UI can offer a move back onto a price nobody can buy.

**Only the webhook writes `users.plan`.** `change_plan/3` reprices the
subscription item and writes nothing locally, so the entitlement always
follows what Stripe actually charges rather than what we asked for. Three
cases deliberately leave the stored plan alone: a `.deleted` event, a price
this deployment does not recognise (realistically an env var that has not
reached one replica), and an event with no items.

**Teammate contacts are billed per unit.** An inbox and a number are a
recurring per-teammate cost, which is the wrong shape for a flat tier.
`sync_contact_addon/1` *sets* an add-on item's quantity from the contact rows,
never increments, so a dropped call converges instead of drifting. It runs
after the row commits and is best-effort — the providers have already been
paid. The per-plan ceiling closes the one window that leaves open: it refuses
with 402 and buys nothing first.

## Surfaces

Pricing table on `/`, plan picker on `/account/billing`, `plan` on
`GET /api/account/billing` and the admin user API, `?plan=` on checkout. Each
shows only the plans whose price id this deployment has, so the ids roll out
one at a time and a self-hosted instance shows none. `DEFAULT_PLAN` (default
`solo`) is the one knob a self-hoster needs.

`mix fountain.verify_plans` reads every configured price from Stripe and fails
if the amount, currency, interval or active flag disagrees with the catalog.
The catalog holds a display price and Stripe holds the charged one with
nothing linking them, so this is the only thing stopping a customer being
shown one number and billed another. Read-only, safe against live mode.

## Before this goes live

Nothing here changes behavior until the price ids are set. The operator steps
are in `docs/guides/operate/plans-and-prices.md`. In short: create three
recurring monthly USD prices, set `STRIPE_PRICE_ID_SOLO` / `_TEAM` / `_SCALE`,
restart, run `verify_plans`. Leave `STRIPE_PRICE_ID` alone — the legacy
accounts still point at it.

⚠️ **`STRIPE_PRICE_ID_CONTACT` is a customer-facing change.** A tenant who has
teammate contacts today pays nothing for them; setting that variable puts a
line item on their next invoice. Tell them first, or leave it unset.

## Verification

- `mix test` — 3702 tests, 0 failures
- `mix credo --strict` — no issues
- `mix dialyzer` — 0 errors
- `mix format --check-formatted`, `mix sobelow --config`, `mix deps.unlock --unused` — clean
- `mix test --cover` — **87.04%** total against baseline 85.46% (measured at `7c771a4`; the two follow-up commits only added tests, and CI's coverage gate is green)
- `vale lint docs` — 0 errors; `python3 scripts/docs-style.py` clean; `mkdocs build --strict` clean
- `okf validate decisions` — 0 errors, index regenerated
- `mix openapi.spec.json` + `jq empty` — valid

New tests: `plans_test.exs` (catalog, resolution, price mapping, ladder
invariants), `billing_plans_test.exs` (signup price, webhook plan writes,
tier switch, add-on quantity), `comms_plan_test.exs` (ceiling, buys-nothing-
on-refusal, best-effort sync), plus quota-by-plan cases and an admin
clear-the-override case.


jhgaylor:
Two follow-up commits after the first CI run, both worth a look:

**`4c356a8`** — CI caught the SDK types going stale. Regenerated, and while
there, narrowed the checkout `?plan=` enum to the *public* slugs.
`Plans.slugs/0` is what a stored row may hold and rightly includes `legacy`; a
request may only name something somebody can buy. That is `public_slugs/0`.

**`28cc7eb`** — a real defect, found by test ordering rather than by design:
`Application.get_env(key, %{})` returns the default only when the key is
**absent**. A key explicitly holding `nil` is a different state, and reading it
as a map raised `BadMapError` out of `Plans.price_id/1` → `available_plans/0` →
`/`. The homepage. Fixed at the read, pinned with a test, and verified by
reverting the fix and watching the test reproduce the crash.

Same commit also splits "your subscription holds a price we do not recognise"
out of the generic "Stripe unreachable" on the plan picker. The realistic cause
is a price id removed from config out from under a live subscription — most
likely `STRIPE_PRICE_ID`, which every `legacy` account still points at — and
"please try again" sends those customers round a loop that cannot work.

jhgaylor:
Follow-up filed as #1016 — included sandbox-hours per plan.

Nothing in it blocks this PR. Worth knowing while reviewing: the measurement half already exists and is good (`SandboxUsage.for_user/3` is park-aware, period-clipped and provider-split, and `/account/billing` already renders it), so the gap is not the numbers. It is that there is **no billing period** to measure an allowance over — `users` has `current_period_end` but no `current_period_start`, and every usage surface uses a calendar month.

Fine for a display stat, wrong for something you enforce and bill overage on.

PR #1003, opened 2026-08-23
feat(sdk): publish @agentshit/fountain-sdk 0.1.0 — with me() fixed and the ask flow answerable

jhgaylor:
Everything between `fountain-sdk` and its first npm release.

## The two that mattered

**`fountain.me()` returned `null` against a real Fountain.** `GET /api/auth/me`
is one of nine endpoints that answer with the object itself rather than
`{data: …}`, and the SDK unwrapped it anyway. The call whose docstring calls it
"the cheapest way to check a key works" answered `null` on success.

The suite was green because `test/server.ts` wrapped the response too, and the
only test touching that path used the raw `request()` escape hatch rather than
the verb — so `me()` had never been exercised. The fake now answers
unenveloped, `me()` is tested through `me()`, and that file carries the list of
unenveloped endpoints, because a fake that disagrees with the server turns a
green suite into a lie.

**An `ask` agent had no way to answer.** The server emits a
`permission_request` block with a request id and the options the agent offered,
and takes the reply at `POST /conversations/:id/requests/:rid` (#950). The SDK
surfaced neither, so the block reached a caller anonymously with nothing to
answer it — and an unanswered request is denied when it expires. An `ask` agent
driven from a script silently skipped every held tool call and still reported
`done`.

```ts
for await (const event of run) {
  if (event.type !== "permission") continue;
  const allow = event.request.options.find((o) => o.kind === "allow_once");
  await run.answer(event.request.requestId, allow.optionId);
}
```

`resume(id).answer(...)` is the same call from a process that is not following
the turn. The end-to-end test carries a deadline, so a regression fails in five
seconds instead of hanging CI on a turn that never ends.

## What publishing needed

- The README's licence section said **MIT**. The SDK is Apache-2.0 and the
  server is AGPL-3.0 (#998), and that file is the npm front page.
- `USER_AGENT` is a literal that would drift from `package.json` on the next
  bump; a test now asserts the two agree.
- `README.md` and `docs/sdk.md` no longer say "build it from a checkout".
- A package `CHANGELOG.md`, shipped in the tarball.
- `update()` takes the API's own patch schema rather than `Partial<Input>`; the
  README's two `## Errors` sections are one; `conversations()` admits in its
  docstring that it is roots-only by default.

## Verification

Both fixes were checked by reverting them — the new tests fail, and pass again
with the fix. Beyond CI, against **production**, from the packed tarball:

```
me(): jhgaylor@gmail.com | role: admin | verified: true
state: done | turn: 1 | status: idle
text: "sdk-live-ok"
events: conversation → turn-start → thinking → text → turn-end
elapsed: 15.7s
```

Also green locally: typecheck, 71 tests, build, the browser bundle, no OpenAPI
drift, `vale lint docs` (0 errors), `scripts/docs-style.py`, and
`mkdocs build --strict`.

## After merge

`fountain-sdk` is not on npm yet, and this PR's docs say `npm install
fountain-sdk`. **Publish before or with the merge**, or the docs site is briefly
wrong. Steps are in the SDK README's Releasing section.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,9 +48,11 @@ upgrade, is in
 ### Added
 
 - **The TypeScript SDK is ready to publish, and can answer a permission
-  request** (`sdk/typescript`). `fountain-sdk` goes to npm as **0.1.0**. The
-  install instructions in `README.md` and `docs/sdk.md` no longer say "build it
-  from a checkout", because there is now a package to install.
+  request** (`sdk/typescript`). It goes to npm as
+  **`@agentshit/fountain-sdk` 0.1.0** — scoped, because the `agentshit` org is
+  where Fountain's packages live. The install instructions in `README.md`,
+  `docs/sdk.md` and `docs/tour.md` no longer say "build it from a checkout",
+  because there is now a package to install.
 
   With it, the one part of the API the SDK could not drive: an agent whose
   `permission_policy` has an `ask` entry stops before the tool call, and a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,20 @@ upgrade, is in
 
 ### Added
 
+- **The TypeScript SDK is ready to publish, and can answer a permission
+  request** (`sdk/typescript`). `fountain-sdk` goes to npm as **0.1.0**. The
+  install instructions in `README.md` and `docs/sdk.md` no longer say "build it
+  from a checkout", because there is now a package to install.
+
+  With it, the one part of the API the SDK could not drive: an agent whose
+  `permission_policy` has an `ask` entry stops before the tool call, and a
+  `permission_request` block used to reach a caller as an anonymous block with
+  no way to reply. It is now a `{ type: "permission" }` run event carrying the
+  request id, the summary and the options the agent offered, and
+  `run.answer(requestId, optionId)` (or `resume(id).answer(...)`) sends the
+  reply. An `ask` agent driven from a script previously lost every held tool
+  call to the server's deny-on-expiry.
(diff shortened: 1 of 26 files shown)


jhgaylor:
**Update: scoped to the new npm org.**

The package is now `@agentshit/fountain-sdk`, with `publishConfig.access:
"public"` — a scoped package publishes restricted otherwise, which for a client
SDK would be exactly wrong.

The rename reaches both READMEs, both changelogs, `docs/sdk.md`, `docs/tour.md`,
`docs/build/team-chat.md`, the example that imports by name, the `paths` mapping
that lets a typecheck resolve the package without a build, and
`package-lock.json` — `npm ci` rejects a lockfile whose name disagrees with
`package.json`, so CI would have gone red on that alone.

Two things deliberately keep the flat name: `USER_AGENT`, because a scope reads
as a second product token in a User-Agent and that string is what the request
logs key on; and the `mkdtemp` prefix in `config.test.ts`, because a `/` in it
makes `mkdtempSync` throw.

Verified as a consumer sees it — packed the tarball, installed it into a clean
project, typechecked against it under **both** `NodeNext` and `bundler`
resolution (the latter is what the eleven browser apps use) exercising the new
permission types, and called the live API through it:

```
TS CONSUMER EXIT = 0
scoped package, live: jhgaylor@gmail.com | role: admin
```

`npm publish --dry-run`: `@agentshit/fountain-sdk@0.1.0`, 96.6 kB, 55 files,
**public access**, tag `latest`.

CI was green on `a215f73` (14 checks, all passing) before this commit; a fresh
run is going now.

PR #1006, opened 2026-08-23
ci(sdk): make CI the only publisher, and the version bump part of review

jhgaylor:
`0.1.0` went to npm from a laptop; `0.1.1` needed a hand-pushed tag. Neither is
a release process — both are a person remembering. This makes **merging a
version bump the release**, and makes publishing from anywhere else fail.

## Publishing follows the version

`Publish SDK` now runs on a push to main touching `sdk/typescript/`, asks the
registry whether that version exists, and publishes it if not. No tag to push,
no button to press. A PR that only touches tests runs it, finds nothing to do,
and says so in the summary.

The `sdk-v*` tag is now written **after** a successful publish, as a record of
which commit produced which version — rather than being the input that causes
one. That is why the job takes `contents: write`.

## The bump happens during review

`SDK release gate` fails a PR that changes what the package ships without
bumping. Otherwise the change lands on main unreleased and nobody notices until
they wonder why the fix they merged isn't on npm.

It also front-loads everything that would fail the publish later, where failing
is cheap and a reviewer is already looking:

- a version already on npm — which npm **never** lets you replace
- a missing `CHANGELOG.md` entry
- the `USER_AGENT` constant, which `npm version` does not touch (this bit us on
  0.1.1)

`sdk-no-release` is the deliberate way out.

### It compares a projection of package.json, not the file

The manifest ships, so a naive rule demands a release for editing `scripts` —
which exists only for this repository. The gate compares the fields a consumer
actually resolves (`exports`, `engines`, `dependencies`, …), so it fires on a
changed export map and stays quiet for plumbing. **This PR is the case in
point:** it edits `package.json` and correctly needs no release.

## Laptops can't publish

`npm publish` from a checkout now refuses with a message explaining the flow.
`--ignore-scripts` walks past it, and that's fine — npm's **"Require trusted
publishing"** setting is the enforcement; this is the guardrail in front of it
that turns a habit into an explanation rather than a confusing registry
rejection.

## Verification

Every branch exercised against a real base (a clean clone at `f9b0644`), not
argued from the code:

| case | want | got |
|---|---|---|
| `src/` changed, no bump | fail | fail |
| bump to already-published `0.1.0` | fail | fail |
| new version, stale CHANGELOG + `USER_AGENT` | fail | fail |
| `README.md` changed, no bump (it ships) | fail | fail |
| `package.json` `scripts` only | pass | pass |
| tests only | pass | pass |
| changelog only | pass | pass |

Also green: 71/71 SDK tests, typecheck, `vale lint docs` (0 errors, 0 warnings),
`scripts/docs-style.py`, `mkdocs build --strict`. The refusal path was run
locally — `npm publish --dry-run` exits non-zero with the explanation.

## One thing still yours

Turn on **"Require trusted publishing"** on the npm package settings. Until
then the laptop path still works at the registry; everything here is repo-side
guardrail, not enforcement.

### DIFF
--- a/.github/workflows/sdk-publish.yml
+++ b/.github/workflows/sdk-publish.yml
@@ -1,26 +1,30 @@
 name: Publish SDK
 
-# Publishes `@agentshit/fountain-sdk` to npm with **trusted publishing**: npm
-# trades this workflow's OIDC identity for a short-lived credential, so there
-# is no NPM_TOKEN in this repository and nothing to leak or rotate. That only
-# works if npm is told, on the package's settings page, to trust exactly this
-# repository and exactly this workflow *filename* — so renaming this file
-# breaks publishing until the trusted publisher is updated to match.
+# Publishes `@agentshit/fountain-sdk` to npm. **CI is the only publisher** —
+# nothing is ever released from a laptop.
 #
-# Publishing is deliberately not attached to the server's `v*.*.*` tags. The
-# SDK talks to the REST API, which is additive, so it versions on its own
-# clock; tying it to the server would either publish a byte-identical package
-# on every server release or hold a ready SDK fix behind one.
+# The trigger is the version itself: merge a PR that bumps
+# `sdk/typescript/package.json` and that version goes out. There is no tag to
+# push and no button to remember, because a release step a human performs is a
+# release step a human forgets. The `SDK release gate` workflow is the other
+# half — it fails a PR that changes what the SDK ships without bumping, so the
+# bump happens during review rather than after.
 #
-#   git tag sdk-v0.1.1 && git push origin sdk-v0.1.1
+# Idempotent by construction: the run starts by asking the registry whether
+# this version already exists and stops if it does. So a PR touching only the
+# SDK's tests runs this workflow, finds nothing to do, and says so.
 #
-# or run it by hand from the Actions tab, which publishes whatever version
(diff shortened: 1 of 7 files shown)


jhgaylor:
**The gate failed this PR, correctly, and my claim above was wrong.**

The PR body says this change "correctly needs no release" because it only edits
`package.json`'s `scripts`. That is not the whole diff — it also rewrites
`sdk/typescript/README.md`, which **ships** (it is the npm front page, and it is
in `files`). The gate caught it:

```
shipped:  sdk/typescript/README.md
::error::This PR changes what the SDK ships … but leaves the version at 0.1.1.
```

I had checked this case locally and got a pass, but the check was invalid: I ran
the guard while the changes were still uncommitted, so `git diff main...HEAD`
saw an empty diff and the guard had nothing to judge. A green result from a
harness that was not exercising anything — the same mistake I made twice
earlier with a clone whose base ref was stale.

The behaviour is right, so I have not weakened the rule to make my own PR pass.
Instead this is the case `sdk-no-release` exists for: the README change is
repository process documentation that happens to live in a published file, and
publishing 0.1.2 to ship a paragraph about how releases work would be silly.

The label did not exist — the gate referenced an escape hatch nobody could
apply. Created it and applied it here, which also exercises that path for the
first time.

Issue #1018, opened 2026-08-23
users.current_period_end has never been populated: Stripe moved the field onto subscription items


jhgaylor:
Found while checking whether a plan upgrade had charged anybody. It had not —
but the investigation turned up a field Fountain has never successfully read.

## The bug

Stripe moved `current_period_start` / `current_period_end` **off the
subscription object and onto its items**. On the API version this account is
pinned to, the subscription-level fields are `null` and the real values sit on
`items.data[]`:

```
"sub_level":  { "current_period_start": null, "current_period_end": null }
"item_level": [ { "price": "price_1U7U29…",
                  "current_period_start": 1786357774,
                  "current_period_end":   1789036174 } ]
```

`Billing.sync_subscription/1` reads the subscription-level field
(`ee/lib/fountain/billing.ex:1427`), gets `nil`, and writes `nil`. Same at
`:546` and `:638` (the resync and adopt paths).

Measured in prod:

```
accounts with a stripe_subscription_id: 3
accounts with current_period_end set:   0
accounts with cancel_at_period_end:     0
```

Zero, not "a few stale ones". The column has never held a value.

## What it breaks

* **The cancel-at-period-end notice.** `/account/billing` renders *"you keep
  full access until &lt;date&gt;"* from `current_period_end`. There is no date, so
  the reassurance a cancelling customer is supposed to get is blank. This is
  the exact thing #284 added the field for.
* **`current_period_end` on `GET /api/account/billing`** and on the admin user
  API — always null, so the SPAs cannot show a period either.
* **The admin billing view's period column.**

Nothing crashes. It degrades to "no date", which is why it has gone unnoticed
since the field was added.

## Why the tests do not catch it

`billing_webhook_sync_race_test.exs` and friends hand-build
`%Stripe.Event{data: %{object: %{current_period_end: ...}}}` in the **old**
shape. The fixture agrees with the code, and both disagree with Stripe. Same
failure mode as `mix fountain.verify_plans` calling `Mix.shell()` — correct
against our idea of the world, wrong against the world.

Any fix needs a test built from a subscription in the *current* shape (period
on the item, absent at the top level), or it will pass for the same wrong
reason.

## The fix

Read the period from the item, falling back to the subscription-level field so
that older API versions and existing test fixtures keep working:

```elixir
defp period_end(sub) do
  Map.get(sub, :current_period_end) ||
    sub |> subscription_items() |> Enum.find_value(&Map.get(&1, :current_period_end))
end
```

[...]

Issue #1016, opened 2026-08-23
Included sandbox-hours per plan: the measurement exists, the period does not


jhgaylor:
Follow-up to #991 (ADR 0026), and the next slice of #798.

#991 ships tiers on **concurrency**, which is a proxy for cost rather than
cost: a tenant running 5 sandboxes around the clock and one running 5 for an
hour a week both pay $29. #798 settled that **sandbox-active-hours** is the
honest unit. This issue is the increment that makes a plan able to say
"includes N hours".

Nothing here blocks #991. It is deliberately a follow-up.

## What already exists — do not rebuild this

The measurement half is done, and it is the good version.
`Fountain.Billing.SandboxUsage.for_user/3` returns `%{provider => active_seconds}`
for any window, and it is already correct in the ways that matter:

- **park-aware** — `sandbox_suspended`/`sandbox_resumed` pairs are subtracted,
  so a sprite scaled to zero (0017) does not burn an allowance;
- **clipped** to `[period_start, period_end)`, so a sandbox spanning a boundary
  contributes to each side only what it ran there;
- **split by provider**, with `platform_cost?/1` marking the seconds Fountain
  actually pays for — `runner` (0022) is the tenant's own machine, real hours
  and zero platform spend;
- **busy separated from idle**, so a report can say which part a shorter idle
  timeout would remove;
- `SandboxUsage.hours/1` converts.

It is already wired: `Billing.usage_summary/3` calls it, and
`/account/billing` already renders the result as `sandbox_minutes`. A customer
can see their hours today. There is simply nothing to compare them against.

The stale "what it would take" list in #798 has been corrected in
[this comment](https://github.com/BinaryBourbon/fountain/issues/798#issuecomment-5383228297) —
its step 1 (metering correctness) closed with #665 and the clipping work.

## The trap: there is no billing period

**This is the part that is not obvious, and it has to be fixed first.**

`users` carries `current_period_end` but **no `current_period_start`**, and
every surface that measures usage — `/account/billing`, the admin overview —
uses `Billing.current_month_range/0`, which is a **calendar month**.

That is fine for a display stat and wrong for an allowance. A customer who
subscribes on the 20th gets a 10-day first "month", and every renewal boundary
sits out of step with the invoice they are being charged against. Enforcing a
cap and billing overage over a window that does not match what they paid for is
a support ticket per customer per month.

Deriving the start from the previous `current_period_end` is fragile: for a
`trialing` account that field holds the trial end, not a period boundary. Sync
[...]

Issue #665, opened 2026-08-13
Suspend-aware usage metering: parked time is invisible to sandbox_minutes


jhgaylor:
Follow-up to decisions/0017 (suspend idle sandboxes). The ee usage-event vocabulary is closed (`sandbox_provisioned` / `sandbox_provision_failed` / `sandbox_terminated` / `turn_started`), so suspension is currently invisible to metering, with imprecision in both directions:

- A sandbox that suspends and later terminates includes all its parked time in the `sandbox_terminated` `duration_ms` — sandbox-minutes overstate.
- A sandbox that stays suspended forever emits no `sandbox_terminated` at all — sandbox-minutes understate.

Deliberately accepted for now because parked-sprite cost is treated as zero. If sandbox-minutes ever start driving billing or capacity decisions, add `sandbox_suspended`/`sandbox_resumed` event types (ee/lib/fountain/billing/usage_event.ex) and make the duration roll-up in ee/lib/fountain/billing.ex subtract parked intervals.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1020, opened 2026-08-23
feat(billing): included turn hours, measured over the invoiced period (#1016)

jhgaylor:
Closes steps 1 to 3 of #1016. Steps 4 and 5 (the overage shape, and a ceiling)
are deliberately not here — they are meant to be decided against a cycle of
these numbers rather than ahead of one.

The measurement half was already the good version. `SandboxUsage` has been
park-aware, period-clipped, split by provider and busy/idle-separated since
#665, and `/account/billing` has shown a customer their sandbox minutes for
months. There was nothing to compare them against, and no period to compare
them over.

## The unit is turn hours, not sandbox hours

| | solo | team | scale | legacy |
|---|---|---|---|---|
| concurrent sandboxes | 5 | 15 | 40 | 15 |
| **included turn hours** | **100** | **300** | **800** | **300** |
| price | $29 | $79 | $199 | $29 (closed) |

20 hours per concurrent slot, derived rather than chosen per tier, so a plan
cannot end up with more capacity and less work to do with it.

#798 called sandbox-active-hours the honest variable unit, and for Fountain's
*cost* it is — an idle sprite bills at full rate. It is the wrong unit to
**sell**, because the tenant behaviour it punishes is forgetting to close a
tab. So the allowance is denominated in `busy_seconds` (a prompt in flight)
and the idle half stays an operator cost signal in `provider_spend/1`. Hours
on a self-hosted runner (0022) are excluded from the other direction: Fountain
pays nothing for that machine.

The prod distribution these numbers were set against, read live:

| month | dogfood account | every other tenant |
|---|---|---|
| Aug, turn hours | 271 | < 0.5 |
| Aug, active hours | 11,091 | 1 – 21 |

The gap between those two rows is the whole argument for the unit. Against
turn hours the allowance actually bites somewhere; against active hours the
same account would be 30x over on every tier.

## The period, which had to come first

`users.current_period_start` is new, nullable, and synced from Stripe beside
the `current_period_end` already read in the subscription webhook, the trial
sweep and the admin resync.

`Billing.billing_period/2` returns `%{start, end, source}` rather than a bare
tuple, because **a number measured over the wrong window has to be able to say
so**. `:subscription` requires both ends known *and* the stored window to
contain now; everything else gets `:calendar_month` — self-hosted, comped, a
trialing account Stripe has not reported a period for, or a period that has
already ended. Every surface displays the flag.

A customer who subscribed on the 20th was getting a 10-day first "month" and a
renewal boundary out of step with every invoice. Deriving the start from the
previous `current_period_end` was rejected: on a `trialing` account that field
holds the trial end, not a period boundary.

`Fountain.Release.backfill_billing_periods/1` fills the column from Stripe for
subscriptions that predate it. One read per account that needs one, `dry_run:`
first, safe to re-run, and it writes the two period columns and **nothing
else** — status/trial/plan repair stays `resync_from_stripe/1`'s job. Without
it every existing account reports a calendar month until its next renewal.

## Reported, enforced nowhere

`Billing.turn_hour_allowance/2` is the single shape all three surfaces render,
so the billing page, `GET /api/account/billing` and the admin per-tenant view
cannot disagree about the window, the unit, or which providers count. **No
gate reads it.** The billing page says so out loud to a tenant who is over.

Note for whoever closes step 4: the contact add-on #991 built is a *licensed*
quantity price, and Stripe ignores quantity on a metered one. Hours need Meter
Events or a credit balance, not a second licensed item.

## Reviewing

Five things worth a look:

1. **`billing_period/2`'s contains-now rule.** It has one wobble: between a
   renewal and the `customer.subscription.updated` that reports the new
   period, an account reads as `:calendar_month` for a few seconds. Rolling
   the stored period forward instead would mean inventing a boundary Stripe
   has not confirmed. Flagged rather than papered over.
2. **`SandboxUsage.busy_for_user/3`** omits providers with zero turn time
   rather than reporting them as `0`, matching `for_user/3`'s "absent means
   none" contract.
3. **The API is additive**: `usage.turn_hours{,_included,_remaining}`,
   `plan.included_turn_hours`, `period.source`, `current_period_start`.
   Nothing was renamed or removed, and `sandbox_minutes` still means exactly
   what it did.
4. **`meter_width/1` rounds before clamping.** `min(101.5, 100)` returns the
   *integer* 100, which `Float.round/2` refuses — a 500 on the billing page
   for exactly the accounts the meter exists for. The test caught it; the
   comment keeps it caught.
5. **The pricing page explains the unit.** A reader who assumes wall-clock
   reads Solo's 100 hours as four days and concludes the plan is unusable.

## Verification

- Full suite green: 3,779 tests, 6 doctests, 3 properties, 0 failures.
- `mix format --check-formatted`, `mix credo --strict` (no issues), `mix
  sobelow --config`, `MIX_ENV=dev mix dialyzer` (0 errors) all clean.
- `mix openapi.spec.json` + `jq empty` valid; the new `period.source` enum is
  classified in `schema_enum_guardrail_test.exs`.
- `python3 scripts/docs-style.py` clean, `vale lint docs` exits 0 with no
  gating rule violations on the changed page.
- `okf validate decisions` valid; the index needed no regeneration (no status
  change).

Prod numbers above were read read-only from a pod, no writes.

---

## Rebased onto #1019 — and it changed the code, not just the context

#1019 landed while this was open and found that Stripe moved
`current_period_start`/`current_period_end` **off the subscription and onto its
items**, so the subscription-level field reads `null`. It fixed the `end` side
with `period_end_unix/1`.

**The textual conflict was the harmless part.** Two hunks conflicted and were
obvious. The dangerous one merged *cleanly*: the webhook's new
`current_period_start` read `Map.get(sub, :current_period_start)` — the field
#1019 had just proved is always nil. Git had no reason to complain, and the
result would have been #1018 repeated one field over: a column never once
populated, `billing_period/2` reporting `:calendar_month` forever, and the
fallback making it look like correct behaviour rather than a bug.

So the resolution generalises #1019's helper rather than sitting beside it:

```elixir
def period_end_unix(sub),   do: period_unix(sub, :current_period_end)
def period_start_unix(sub), do: period_unix(sub, :current_period_start)
```

Same item-then-subscription fallback, same atom/string key handling, now shared.
All three call sites (webhook sync, trial sweep, admin resync) read both ends
through it.

`billing_period_shape_test.exs` — #1019's fixture-shape test — gains six cases
for the start field, including one that reads both ends off a single item and
asserts `billing_period/2` accepts the result as `:subscription`. **Verified by
reverting the fix**: `period_start_unix` restored to a subscription-level read
fails 4 of them.

## The release task runs under `eval`, and says when it did nothing

Two corrections while rebasing, both from #1019's neighbourhood:

- **`eval`, not `rpc`.** #1019 moved `verify_plans` to `rpc`, which is right
  for a Mix task. It is wrong here: `Ecto.Migrator.with_repo` *restarts the
  repo's pool child* when it finds one already started, so an `rpc` would bounce
  the connection pool of a pod that is serving. Every `Fountain.Release.*` task
  uses `eval` for that reason, and `config/runtime.exs` (which sets
  `STRIPE_SECRET_KEY` unconditionally) does execute under it. Both invocations
  now appear on the same docs page, so the distinction is written down.
- **All-skipped is now loud.** One skipped account is a deleted subscription.
  *Every* account skipped is an unset, revoked or wrong-mode key, or Stripe
  being unreachable — and the task would have printed a clean summary over a run
  that repaired nothing. It now warns. I first wrote this as a config check for
  `api_key`; that was worse (it misses a revoked key, a wrong-mode key and an
  outage) and it broke four tests, since the test env deliberately configures no
  key. Checking the outcome catches all four cases and needs no global env.

## SDK: publishing 0.1.3

The generated types change what the SDK ships, so the release gate asked for the
decision in review. Bumped to 0.1.3 with `USER_AGENT` and a changelog entry —
**merging this publishes `@agentshit/fountain-sdk@0.1.3` to npm.**

Additive only, so an app on 0.1.2 keeps compiling. The changelog spells out the
two things a client can get wrong: a turn hour is not a sandbox hour, and
`period.source == "calendar_month"` means the numbers do not line up with an
invoice.

## Re-verified after the rebase

3,793 tests 0 failures · credo `--strict` clean · dialyzer 0 errors · format
clean · docs-style + vale clean · SDK typecheck + 71 tests · generated types
regenerate to no diff · release gate reports "Release looks well-formed".

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -61,10 +61,11 @@ every plan.
 | | solo | team | scale | legacy |
 |---|---|---|---|---|
 | concurrent sandboxes | 5 | 15 | 40 | 15 |
+| included turn hours | 100 | 300 | 800 | 300 |
 | teammate contacts | 1 | 3 | 10 | 3 |
 | price | $29 | $79 | $199 | $29 (closed) |
 
-Three rules that are easy to get wrong:
+Four rules that are easy to get wrong:
 
 - **The plan is not the cap.** `Quotas.sandbox_limit/1` returns
   `users.sandbox_limit_override` when it is set and the plan's number
@@ -86,6 +87,22 @@ Three rules that are easy to get wrong:
   an allowance. Two levers comp one: `comp_account/1` makes everything free,
   `users.comped_contacts` takes N off the billed quantity for a tenant who
   still pays for their tier.
+
+- **A turn hour is not a sandbox hour, and nothing enforces either.**
+  `included_turn_hours` is 20 per concurrent slot, measured against
+  `SandboxUsage`'s `busy_seconds` (a prompt in flight) on platform-paid
+  providers only — an idle sandbox and a self-hosted runner spend none of it.
+  `Billing.turn_hour_allowance/2` is the one shape every surface renders, so
+  they cannot disagree. **No gate reads it** (#1016 steps 4 and 5 are unbuilt);
+  do not add one without deciding the overage shape first.
+
+- **Usage is measured over the invoiced period, or says it is not.**
+  `Billing.billing_period/2` returns `%{start, end, source}` — `:subscription`
+  from `users.current_period_start`/`_end`, `:calendar_month` when there is no
+  such period (comped, self-hosted, or no webhook yet). The fallback is never
+  silent: if you add a surface that shows a usage number, show the source too.
+  `Fountain.Release.backfill_billing_periods/1` fills the column from Stripe
+  for accounts that predate it.
 
 Price ids come from config (`STRIPE_PRICE_ID_SOLO` and friends), the display
 cents from the catalog. `mix fountain.verify_plans` is the only thing that
--- a/apps/fountain/lib/fountain/accounts/user.ex
+++ b/apps/fountain/lib/fountain/accounts/user.ex
@@ -52,6 +52,14 @@ defmodule Fountain.Accounts.User do
     # access continues until current_period_end, when `.deleted` fires. Synced
     # from subscription webhooks so the UI can say "access until <date>".
     field :cancel_at_period_end, :boolean, default: false
+    # The invoiced period, both ends of it, synced from the Stripe
+    # subscription. `current_period_start` is what makes an allowance
+    # measurable over the window the customer is actually charged for rather
+    # than over a calendar month that drifts out of step with every renewal.
+    # Null on a trialing account Stripe has not reported a period for, on a
+    # comped account, and on a self-hosted deployment with no Stripe —
+    # `Fountain.Billing.billing_period/2` handles all three.
+    field :current_period_start, :utc_datetime
     field :current_period_end, :utc_datetime
     field :session_version, :integer, default: 0
     field :theme_preference, :string, default: "system"
@@ -143,6 +151,7 @@ defmodule Fountain.Accounts.User do
       :trial_ends_at,
       :subscription_synced_at,
       :cancel_at_period_end,
+      :current_period_start,
       :current_period_end,
       # Only ever put here when the price on the subscription mapped to a
       # known plan. An unrecognised price leaves the key out entirely rather
--- a/apps/fountain/lib/fountain/plans.ex
+++ b/apps/fountain/lib/fountain/plans.ex
@@ -13,18 +13,38 @@ defmodule Fountain.Plans do
 
   Concurrent sandboxes, and only that. It is the one thing that is both a real
   cost to Fountain (capacity on the shared provider account, ADR 0005/0018)
-  and legible to a customer without a usage meter. Sandbox-hours are the
-  honest *variable* unit and remain the plan for a later revision of #798;
-  they need metering correctness first and they are not what these tiers sell.
+  and legible to a customer without a usage meter. The tiers are still priced
+  on it; the turn-hour allowance below is an amount of work *inside* a tier,
+  not a second axis to sell on.
+
+  ## Included turn hours
+
+  `included_turn_hours` is derived, not chosen per tier: **20 turn hours per
+  concurrent sandbox the plan allows**, so Solo's 5 slots carry 100 hours and
+  Scale's 40 carry 800. Deriving it keeps the ladder on one axis — a plan
+  cannot end up with more capacity and less work to do with it.
+
+  A *turn* hour is time with a prompt actually in flight
+  (`Fountain.Billing.SandboxUsage`'s `busy_seconds`), not wall-clock sandbox
+  time. The distinction is the whole point: a sandbox left running overnight
+  with nobody prompting it burns `active_seconds` and no allowance, so the
+  meter measures work rather than forgetfulness. Hours on a tenant's own
+  runner (ADR 0022) do not count either — Fountain pays nothing for them.
+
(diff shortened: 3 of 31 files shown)


jhgaylor:
Flagging before this gets rebased, because the conflict with #1019 is semantic
as well as textual and the textual resolution could silently keep the bug.

## `current_period_start` is read from a field Stripe no longer populates

`billing.ex:546`, `:639` and `:1435` read `Map.get(sub, :current_period_start)`.
Stripe moved `current_period_start`/`current_period_end` **off the subscription
and onto its items**; at subscription level both now read `null` on the API
version this account is pinned to:

```
"sub_level":  { "current_period_start": null, "current_period_end": null }
"item_level": [ { "current_period_start": 1786357774,
                  "current_period_end":   1789036174 } ]
```

That was #1018, fixed in #1019 (`cbe49c8`, already on main) for
`current_period_end`. This branch predates it, so it both **reintroduces the
bug for the new column** and, where the two touch the same lines, **reverts the
fix for the old one**.

The consequence is not cosmetic. `billing_period/2` requires both ends known,
so with `current_period_start`永 null every account would report
`:calendar_month` — the exact fallback this PR exists to stop being the norm.
The included-hours numbers would be measured over the wrong window, and the
`source` flag would honestly say so on every surface, for everybody, forever.
Green tests either way, because the fixtures build the old shape.

## What the rebase should land on

#1019 added `Billing.period_end_unix/1` — item first, subscription-level
fallback so older API versions and existing fixtures still work. This PR wants
the same for the start. Probably worth generalising rather than pasting:

```elixir
defp period_unix(sub, key) do
  case Map.get(sub, key) || Map.get(sub, to_string(key)) do
    ts when is_integer(ts) -> ts
    _ ->
      sub
      |> subscription_items()
      |> Enum.find_value(fn item ->
        case Map.get(item, key) || Map.get(item, to_string(key)) do
          ts when is_integer(ts) -> ts
          _ -> nil
        end
      end)
  end
end
```

…with `period_end_unix/1` kept as the public shim its tests already cover.

Also worth taking from #1019 while rebasing:

* `ee/test/fountain/billing_period_shape_test.exs` builds subscriptions in the
  **current** shape (period on the item, absent at the top). Every other test
  in the suite builds the old one, which is exactly why this class of bug
  survives review. Whatever `current_period_start` ends up doing wants a case
  in there, or it will pass for the reason the bug lasted.
* `docs/guides/operate/plans-and-prices.md` on this branch still says
  `bin/fountain_server eval 'Mix.Tasks.Fountain.VerifyPlans.run([])'`. Main now
  says `rpc` — `Mix` is not in a release, so `eval` cannot work, and the same
  file gained a `backfill_billing_periods` block here that uses `eval` too.
  Worth checking whether `Fountain.Release.backfill_billing_periods/1` actually
  runs under `eval`: release tasks generally do, since `Fountain.Release` is
  built for it, but it is the same class of "documented command nobody ran".

## Not blocking

Everything else reads well, and the turn-hours-not-sandbox-hours argument with
the prod distribution behind it is the right call. This is one field, on a
branch that has to be rebased anyway.

I am building trial limits on top of this branch (`feat/trial-limits`), so
I will rebase behind whatever you land.

PR #1025, opened 2026-08-23
feat(admin): a finance panel — revenue against what each tenant costs

jhgaylor:
Closes the gap #991 and #1020 left open: the plans and the turn-hour meter
exist, and nothing anywhere held them next to what a tenant actually costs.

## What this adds

`/admin/finance` — revenue, platform spend and the margin between them, one
row per tenant, worst margin first, with a month picker back through the last
six. Read-only; every lever stays on `/admin` next to its confirmation.

## The five things worth reviewing

**It reports no money it was not told.** Fountain's own costs are nowhere in
this codebase — provider prices are per-machine-size and negotiated, and
AgentMail/AgentPhone bill per unit and per message — so cost comes from a rate
card in config. Set none of it and the panel still works, in hours, inboxes,
numbers and message counts, with a banner saying so. An unpriced line renders
`—` and never `$0.00`, and **the nil propagates to the total**: a cost that
quietly omitted the provider nobody priced would read as a cheap tenant, most
convincingly on the expensive ones. `finance_test.exs` tests that
specifically, and reverting `add_or_nil/1` to a filtering sum fails two cases.

**It does not assume which hours a provider bills.** You said the bill tracks
turn hours more closely than sandbox hours, and whether that is true is a fact
about the invoice rather than about this code. So the rate multiplies either
one and a toggle switches between them (`PROVIDER_COST_BASIS` sets the
default). Both hour figures stay on every row either way, since the gap
between them is idle time and that is the lever. On the `:turn` basis the
"a shorter timeout would save this" line disappears rather than reporting a
saving that does not exist.

⚠️ **Nothing is priced until you set the rate card.** Nothing here changes
behavior for a self-hosted instance, which sets none of it.

**Teammate messages are metered, for the first time.** `comms_email_sent`,
`comms_sms_sent` and `comms_sms_received` join the usage-event vocabulary,
recorded at the two choke points a message already passes through — the MCP
controller's audit callback and `Team.Comms.Inbound` — rather than at each
send site. An inbox and a number are a monthly charge and a message is a
per-unit one, and only the first half was visible. Inbound counts, because
AgentPhone charges to receive. Best-effort by the same contract as every other
usage event; neither call can fail a send. `Comms.channel_counts/0` counts
inboxes and numbers apart, because the two providers charge apart and a
half-released contact must be billed for the half that still exists.

**The admin MRR tile was silently wrong, and this fixes it.** It was
`active × STRIPE_PRICE_MONTHLY_CENTS` — one amount for everybody. Its comment
said the day there was a second price would be loud; it was not. #991 shipped
four tiers and the tile went on charging every active account the legacy $29,
so a deployment selling Scale under-reported its own revenue sevenfold. It now
reads `Finance.mrr/0`, priced from the plan catalog with the contact add-on
added, and carries the per-plan split. `STRIPE_PRICE_MONTHLY_CENTS` is no
longer read there. A test asserts `mrr/0` and the full pass agree — a tile
that contradicts the table under it destroys trust in both.

**The dashboard shows turn hours, not sandbox time.** Sandbox wall-clock is
Fountain's cost signal and nothing a customer buys or is measured on; it went
up while they slept. The tile is now turn hours against the plan's included
hours, and the sandbox figure moved to the hint. The whole "this month"
section also moves to the window Stripe invoices where there is one, so the
dashboard and `/account/billing` can no longer report different numbers for
the same period; the heading says which window it is on. `/admin`'s per-user
usage column changes for the same reason, with the sandbox split kept in the
tooltip.

## One bug the tests caught

The per-tenant row counted turn hours on a tenant's own runner against their
allowance. `Billing.turn_hours_used/2` deliberately excludes it (ADR 0022 —
Fountain pays nothing for that machine), so the panel and the tenant's own
billing page would have disagreed about the same number. Test written first,
failed, fixed.

## What is deliberately not here

Per-tenant cost on `/admin/users/:id`. The finance table carries it and each
row links to the detail page; a second read path for one tenant is a second
chance to disagree.

No gate reads any of this. #1016 step 4 — the overage shape — was left open to
be decided against a cycle of real numbers, and this is the page that produces
them.

## Verification

- `mix test` — 3,846 tests, 6 doctests, 3 properties, **0 failures**
- `mix credo --strict` — no issues · `MIX_ENV=dev mix dialyzer` — 0 errors
- `mix format --check-formatted`, `mix sobelow --config`,
  `mix deps.unlock --unused` — clean
- `mix openapi.spec.json` + `jq empty` — valid (the API is unchanged; no field
  renamed or removed)
- `python3 scripts/docs-style.py` clean · `vale lint docs` 0 errors ·
  `okf validate decisions` valid
- `ConfigReferenceTest` caught `PROVIDER_COST_BASIS` before I documented it,
  which is the guardrail working

New tests: `billing_finance_test.exs` (35 — the rate card, nil propagation,
both cost bases, per-plan revenue, margin ordering, deleted-account spend),
`admin_finance_live_test.exs` (11 — access control, the unpriced state, the
toggle, the month window), plus `channel_counts/0`, comms metering at both
choke points, and the dashboard tile.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -237,10 +237,38 @@ STRIPE_CONTACT_PRICE_CENTS=
 # The flat price the plans replaced. Accounts that bought it are on the closed
 # `legacy` plan and still point at this id. A new deployment does not need it.
 STRIPE_PRICE_ID=
-# Monthly price of the `legacy` plan in cents (e.g. 2900), display-only: feeds
-# the admin billing overview's MRR tile. Optional — unset shows "price not
-# configured". Public plans carry their own display price in Fountain.Plans.
+# Monthly price of the `legacy` plan in cents (e.g. 2900), display-only.
+# Optional. Public plans carry their own display price in Fountain.Plans, and
+# the admin MRR tile now prices every account from that catalog rather than
+# from this one number.
 STRIPE_PRICE_MONTHLY_CENTS=
+
+# ── The finance panel's rate card (/admin/finance) ───────────────────────────
+#
+# What Fountain PAYS, which nothing else in the codebase knows: provider prices
+# are per-machine-size and negotiated, and AgentMail/AgentPhone bill per unit
+# and per message. All optional. Unset, /admin/finance reports hours, inboxes,
+# numbers and message counts with no dollars at all — an unpriced line shows
+# "—" and never becomes $0.
+#
+# Cents per active sandbox hour, per provider. A provider left out is unpriced,
+# and the tenants using it report no cost rather than a cost that omits it.
+# `runner` is never priced: it is the tenant's own machine (ADR 0022).
+# PROVIDER_HOURLY_CENTS=sprites=12,e2b=15,daytona=10
+#
+# Which hours that rate multiplies: `active` (default, every hour the sandbox
+# was awake) or `turn` (only the hours with a prompt in flight). Use `turn`
+# where the provider scales to near-zero between prompts. The panel can switch
+# between the two per view; this is what it opens on.
+# PROVIDER_COST_BASIS=turn
+#
+# Teammate contacts. The monthly pair is pro-rated to the window the panel
+# shows; the message pair is charged per message, inbound SMS included, since
+# AgentPhone charges to receive.
+# AGENTMAIL_INBOX_CENTS=200
+# AGENTPHONE_NUMBER_CENTS=150
+# AGENTMAIL_MESSAGE_CENTS=1
+# AGENTPHONE_MESSAGE_CENTS=8
 
 # The plan an account with no plan of its own gets — every account on a
 # self-hosted instance. `scale` lifts the concurrency cap for everyone at
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,39 @@ upgrade, is in
   about the event *name* (73 distinct request-line names in one day, each a
   permanent PostHog event definition), and a property is where PostHog can
   break down by a value the router bounds.
+- **A finance panel at `/admin/finance`.** `/admin` had an MRR tile with no
+  cost beside it and a sandbox-hours table with no money in it, so the only
+  question an operator asks a finance page — which accounts cost more than
+  they pay — had no answer anywhere. The new page puts revenue, platform
+  spend and the margin between them on one row per tenant, worst margin
+  first, with a month picker back through the last six.
+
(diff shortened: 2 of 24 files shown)


jhgaylor:
## Rebased onto main — and #1022 changed the code, not just the context

Three PRs landed while this was open. The PR went `CONFLICTING`, which is why
CI reported **zero** checks rather than a red one: GitHub cannot build the
merge ref, so nothing fires at all.

**The textual conflict was the harmless part** — one hunk in `CHANGELOG.md`,
two `### Added` blocks to interleave.

The dangerous one merged cleanly. #1022 split the plan API in two:

* `Plans.resolve/1` — *what tier is this subscription for*. What Stripe
  charges, and what a trial converts into.
* `Plans.effective/1` — *whose numbers apply today*. During a trial, the new
  `trial` plan: 2 sandboxes, **40** turn hours, 0 contacts.

Every entitlement reader in core moved to `effective/1`. This panel, written
before that existed, read `Plans.resolve(user.plan)` — a bare slug, which
`effective/1` cannot make trial-aware because a slug carries no status. That
is the exact trap #1022's own description calls out, and git had no reason to
complain about it.

The result would have been a trialing account 50 turn hours in reporting
comfortably inside Team's 300 when it is 25% over a trial's 40 — and "turn
hours sold" counting 300 hours nobody has bought.

So each tenant row now carries both:

| | reads | used for |
|---|---|---|
| `:plan` | `resolve/1` | revenue lines — a trial converts into that tier at that price |
| `:effective_plan` | `effective/1` | the allowance column, and `turn_hours.sold` |

The per-tenant table names both when they differ (`Trial → Team`), the way
`/account/billing` already does. `trialing_cents` still prices at the tier,
which is the right answer for "what this converts to".

**Verified by reverting the one argument** — `effective_plan.included_turn_hours`
back to `plan.included_turn_hours` — which fails two tests, including a new one
whose numbers are chosen to be inside the tier and over the trial.

## Re-verified after the rebase

3,924 tests 0 failures · credo `--strict` clean · dialyzer 0 errors · format,
vale (0 errors), docs-style all clean.

Also fixed the one real CI failure from the first push: a 27-word sentence in
`docs/configuration.md` against vale's 25-word description limit. I had edited
that paragraph after the vale run that cleared it and re-ran only
`docs-style.py`, which does not check sentence length.

Issue #391, opened 2026-08-03
security: the secret-editing LiveViews hold the tenant DEK in socket assigns, and LiveView crash reports dump assigns


jhgaylor:
#315 closed this exposure for `ConversationServer` by exporting a `format_status/1` that redacts the tenant DEK and plaintext secrets from crash reports. The same exposure is open on the LiveViews that edit secrets, and they got no such treatment.

## Location

`apps/fountain/lib/fountain_web/live/environments_live/form.ex:14,20,26`
`apps/fountain/lib/fountain_web/live/vaults_live/form.ex:12,18,24`

Both unwrap the tenant DEK at mount and hold it in socket assigns for the life of the mount:

```elixir
{:ok, dek} = Crypto.load_tenant_key(user_id)

{:ok,
 socket
 |> assign(:tenant_key, dek)
 ...
 |> assign(:new_secret, %{"key" => "", "value" => ""})}
```

The environment form additionally reassigns `:new_secret` on every `phx-change`, so the secret currently being typed sits in assigns next to the key that decrypts every other secret the tenant owns.

Neither module defines `format_status/1`. Repo-wide, the only one is `conversation_server.ex:1169` — the #315 fix.

## Impact

LiveView's channel exports the legacy `format_status/2`, whose `:terminate` clause returns state unmodified, and OTP puts that into the crash report's `state` field. An unhandled exception in either LiveView therefore writes the raw tenant DEK and any in-flight plaintext secret to stdout — and to Sentry, since `Sentry.LoggerHandler` is attached for error-level reports from any process (`apps/fountain/lib/fountain/application.ex:106-114`).

The DEK is the per-tenant key for that tenant's entire envelope-encrypted secret set, so this is strictly worse than leaking the one secret being edited.

The crash is not hypothetical on these exact modules: the `{:ok, dek} = ...` match itself raises when `load_tenant_key/1` returns `{:error, :unwrap_failed}` — see the sibling issue on that MatchError.

## Fix direction

Preferred: stop holding the DEK across the mount. Load it inside the write handlers (`add_secret`, `delete_secret`) so it goes out of scope when the handler returns — which is what the controllers already do (`secret_controller.ex:53`, `vault_secret_controller.ex:57`). That removes the class rather than one instance of it.

If it must be kept in assigns, add a `format_status/1` to each LiveView redacting `:tenant_key` and `:new_secret`, mirroring `conversation_server.ex:1168-1195`.

[...]

PR #1048, opened 2026-08-23
feat(vaults): secret expiry dates with advance-notice email

mdonigian:
## What

A vault secret is usually a token whose issuer chose a lifetime, and today it expires silently — the first sign is a conversation failing against a dead credential. This adds an optional `expires_at` to vault secrets, shows staleness in the console, and emails the owner before the date arrives.

## How

- **Schema**: `vault_secrets` gains `expires_at` and `expiry_notified_at` (both nullable — existing rows simply have none), with a partial index carrying the sweep. Changing `expires_at` clears the notified stamp in the changeset, so a rotated or extended expiry re-arms the notice.
- **Sweep**: `Workers.SecretExpirySweeper` on the Oban cron, daily at 17:07 UTC. One email per owner listing every due secret, stamped only after a successful send (a delivery failure retries next run). Skips unverified owners without stamping, so they get the notice after verifying. `:secret_expiry_notice_days` config (default 7, `0` disables).
- **Advisory only**: nothing is enforced on the date — an expired secret keeps being injected, because a missing env var fails worse than a stale one, and the email says so.
- **Console**: optional date input on the add-secret form (uncontrolled form preserved per #391 — the expiry is metadata, not secret), updated-ago and expiry-status columns on the secrets table.
- **API**: `VaultSecretRequest` accepts `expires_at`, responses return it; values remain write-only.
- **Audit**: expiry rides through the existing `vault.secret.write` — no new actor or action.

**Limitation, deliberate for v1**: `value` stays required on the upsert, so setting or extending an expiry requires writing the value too — extension is only reachable through a rotation. A value-less `PATCH` of secret metadata is #1053. In the console, a blank date on a rotation keeps the stored expiry; clearing is an explicit API action (`expires_at: null`).

## Testing

- `mix precommit` green locally (the only test failures in a full single-machine run were 5 pre-existing DB-pool flakes in unrelated files, which pass in isolation — CI's 6 partitions avoid this)
- New worker test covers: one-mail-per-owner batching, already-expired secrets, once-per-expiry dedup, re-arm on expiry change, unverified-owner skip, no-expiry rows untouched, and the disable switch
- Context, controller, and LiveView tests extended for the new field


apps/fountain/lib/fountain_web/schemas.ex
```diff
@@ -1196,6 +1196,7 @@ defmodule FountainWeb.Schemas do
         id: %Schema{type: :string, format: :uuid},
         key: %Schema{type: :string},
         vault_id: %Schema{type: :string, format: :uuid},
+        expires_at: %Schema{type: :string, format: :"date-time", nullable: true},
```
jhgaylor, line 1199: This (and the `VaultSecretRequest` change below) changes the OpenAPI spec, which means `sdk/typescript/src/generated/openapi.ts` must be regenerated in the same PR — CI diffs it against a fresh generation and fails on the 7 lines this adds. Command is in the review summary.

apps/fountain/lib/fountain_web/live/vaults_live/form.ex
```diff
@@ -53,13 +53,13 @@ defmodule FountainWeb.VaultsLive.Form do
...
 
-  def handle_event("add_secret", %{"secret" => %{"key" => k, "value" => v}}, socket)
+  def handle_event("add_secret", %{"secret" => %{"key" => k, "value" => v} = params}, socket)
       when k != "" and v != "" do
     with {:ok, dek} <- Crypto.load_tenant_key(socket.assigns.user_id),
          {:ok, _} <-
            Vaults.upsert_secret(
              socket.assigns.vault,
-             %{"key" => k, "value" => v},
+             %{"key" => k, "value" => v, "expires_at" => parse_expiry(params["expires_at"])},
```
jhgaylor: This always sends `expires_at`, and a blank date input parses to `nil`. Re-adding an existing key is how you rotate a value in the console, and the date field is empty by default (nothing prefills it from the table) — so a rotation with the date left blank casts `expires_at: nil`, which is a real change from the stored date: the expiry is cleared and the notify stamp reset. "Rotated the token, lost the reminder" is the exact failure this feature is for.

The API path doesn't have this problem (`Map.take` leaves the column alone when the key is absent). Either only include `expires_at` here when the field is non-empty, or make clearing an explicit action — your call, but it should be a decision rather than a side effect.

apps/fountain/lib/fountain/vaults/vault_secret.ex
```diff
@@ -22,14 +24,25 @@ defmodule Fountain.Vaults.VaultSecret do
   """
   def changeset(secret, attrs, dek) when is_binary(dek) do
     secret
-    |> cast(attrs, [:key, :value, :vault_id])
+    |> cast(attrs, [:key, :value, :vault_id, :expires_at])
     |> validate_required([:key, :value, :vault_id])
```
jhgaylor, line 28: `value` stays required, in both the UI and the API, so there is no way to set or extend an expiry on an existing secret without pasting the value again. The PR body says "a rotated or extended expiry re-arms the notice" — extension isn't actually reachable without a rotation today. Fine to leave for a follow-up (a value-less `PATCH` of metadata would be the natural shape), but say so explicitly in the PR and the docs.

apps/fountain/lib/fountain/emails/user_emails.ex
```diff
@@ -315,6 +341,64 @@ defmodule Fountain.Emails.UserEmails do
     """
   end
 
+  defp secrets_expiring_html(entries, url) do
+    rows =
+      Enum.map_join(entries, "\n", fn e ->
+        "<li><strong>#{e.key}</strong> in vault <strong>#{Plug.HTML.html_escape(e.vault_name)}</strong> — expires #{expiry_date(e.expires_at)}</li>"
```
jhgaylor: nit: `e.key` is interpolated unescaped. It's safe today because the changeset enforces `^[A-Z][A-Z0-9_]*$`, but escaping it like `vault_name` costs nothing and stops this line depending on a validation in another file.

apps/fountain/lib/fountain/emails/user_emails.ex
```diff
@@ -115,6 +115,32 @@ defmodule Fountain.Emails.UserEmails do
...
+    new()
+    |> from(from_address())
+    |> to({user.email, user.email})
+    |> subject(secrets_expiring_subject(entries))
+    |> html_body(secrets_expiring_html(entries, url))
+    |> text_body(secrets_expiring_text(entries, url))
+    |> Mailer.deliver()
+  end
+
+  defp secrets_expiring_subject([%{key: key}]), do: "Your vault secret #{key} expires soon"
```
jhgaylor: nit: for an already-expired secret (the case your `an already-expired secret is still worth one notice` test covers) this subject still says "expires soon". Worth a branch on `expires_at < now`, or wording that covers both.

apps/fountain/lib/fountain_web/live/vaults_live/form.ex
```diff
@@ -127,6 +127,45 @@ defmodule FountainWeb.VaultsLive.Form do
...
+
+  defp expiry_status(%{expires_at: nil}), do: nil
+
+  defp expiry_status(%{expires_at: expires_at}) do
+    days = Date.diff(DateTime.to_date(expires_at), Date.utc_today())
+
+    cond do
+      days < 0 -> {:expired, "expired #{DateTime.to_date(expires_at)}"}
+      days == 0 -> {:expiring, "expires today"}
+      days <= 14 -> {:expiring, "expires in #{days}d"}
```
jhgaylor: nit: the console goes amber at 14 days, the email fires at `:secret_expiry_notice_days` (7). Not wrong, but two numbers for one idea — either derive this from the same config or leave a comment saying the divergence is deliberate.

apps/fountain/lib/fountain/workers/secret_expiry_sweeper.ex
```diff
@@ -0,0 +1,100 @@
...
+
+    due()
+    |> where([s], s.expires_at <= ^cutoff)
+    |> order_by([s], asc: s.expires_at)
+    |> limit(@batch)
+    |> join(:inner, [s], v in Vault, on: s.vault_id == v.id)
+    |> select([s, v], %{secret: s, vault_name: v.name, user_id: v.user_id})
+    |> Repo.all()
+    |> Enum.group_by(& &1.user_id)
+    |> Enum.count(fn {user_id, entries} -> notify(user_id, entries) end)
```
jhgaylor: nit: counting via a side-effecting predicate reads oddly — a reader expects `Enum.count/2` to be pure. `Enum.filter(...) |> length()` or a `reduce` says what's happening.

Also worth knowing (no change needed): the ordered `limit(@batch)` with unverified owners skipped-but-unstamped means their rows sit at the head of the batch until `UnverifiedAccountPruner` removes the account. Harmless at real volumes.

jhgaylor (changes requested):
Thanks for this — it's a well-scoped feature and it follows the house rules (audit in the context, no values in the trail, uncontrolled secret form per #391, sweep as a system worker in the shape of `TrialSweeper`). The sweeper tests are exactly the ones I'd want.

CI hasn't run on this fork PR yet, so I ran the gates locally on `82dd569`: format, `compile --warnings-as-errors`, `credo --strict`, sobelow, dialyzer, the migration, and the four touched test files plus the audit guardrail, API spec and docs suites — all green (155 tests, 0 failures). One gate will fail, and there are a few things to settle before merge.

### Blocker

**`sdk/typescript/src/generated/openapi.ts` needs regenerating.** CI derives the TypeScript types from the OpenAPI spec and fails on any diff (the "SDK types match the spec" step in `ci.yml`). Your `schemas.ex` change produces a 7-line diff there — the two new `expires_at?: string | null` fields. Fix:

```bash
mix openapi.spec.json --spec FountainWeb.ApiSpec --vendor-extensions=false --output /tmp/sdk-openapi.json
cd sdk/typescript && npm ci && FOUNTAIN_OPENAPI=/tmp/sdk-openapi.json npm run generate
```

and commit the result.

### Before merge

1. **Rotating a value in the console drops its expiry** — inline comment on `form.ex`.
2. **Extending an expiry needs the value re-pasted** — inline comment on `vault_secret.ex`. A follow-up is fine; the PR text should say so rather than describe "extended" as a supported path.
3. **CHANGELOG** — please add an `### Added` entry under `[Unreleased]`; every user-facing change gets one.
4. **Docs** — `docs/api.md` (around the vault secrets block) says a secret list returns key, id and timestamps; it now also returns `expires_at`. And `docs/concepts/vault.md` should get a sentence on the expiry and the email. Heads-up: `docs/` is linted by `vale` (ASD-STE100) and `scripts/docs-style.py` — short declarative sentences, no em dashes. Your prose style is great in code comments and will get flagged in `docs/`, so keep the doc lines plain.

### Nits (take or leave)

Inline. None of them block.

Once the SDK regen is in I'll approve the workflow run so CI covers the rest.

apps/fountain/lib/fountain_web/schemas.ex
mdonigian: Done in a3076fc — regenerated with CI's exact command; the diff is the predicted 7 lines (the two `expires_at?: string | null` fields).

apps/fountain/lib/fountain_web/live/vaults_live/form.ex
mdonigian: Fixed in a3076fc, taking the first option: a blank (or unparseable) date now leaves `expires_at` out of the attrs entirely, so a rotation with the empty date field keeps the stored expiry and the notify stamp. Clearing stays an explicit action — `expires_at: null` over the API. Added a rotation test that pins exactly the "rotated the token, lost the reminder" case, and a code comment saying the decision.

apps/fountain/lib/fountain/vaults/vault_secret.ex
mdonigian: Kept as a follow-up, now said explicitly everywhere per your call: filed #1053 (value-less PATCH of secret metadata), the CHANGELOG entry names it, and the docs sentence reads "To extend the date, write the secret again with its value" rather than describing extension as standalone. PR body updated to match.

apps/fountain/lib/fountain/emails/user_emails.ex
mdonigian: Escaped in a3076fc, same as vault_name.

apps/fountain/lib/fountain/emails/user_emails.ex
mdonigian: Done in a3076fc: "has expired" past the date, singular and plural ("N of your vault secrets have expired or expire soon" when mixed); the already-expired worker test now asserts the expired subject.

apps/fountain/lib/fountain_web/live/vaults_live/form.ex
mdonigian: Derived in a3076fc: the amber window now reads :secret_expiry_notice_days, so the badge and the email describe the same window (and both go quiet together when it's set to 0).

apps/fountain/lib/fountain/workers/secret_expiry_sweeper.ex
mdonigian: Refactored in a3076fc to map + count. Noted on the ordered-batch behaviour with unverified owners at the head — agreed it's harmless at real volumes, and their accounts age out via UnverifiedAccountPruner.

mdonigian:
All review items addressed in a3076fc — blocker (SDK regen, the predicted 7-line diff), blank-date rotation now keeps the stored expiry (+ test), one notice window derived from `:secret_expiry_notice_days`, expired-subject branch, key escaping, sweeper count refactor, CHANGELOG entry, and docs (`docs/concepts/vault.md` Secret expiry section + `docs/api.md`; `docs-style.py` clean, `vale` 0 errors 0 warnings, `docs_test` 29/29). Extension-without-value is filed as #1053 and stated as a limitation in the PR body, CHANGELOG and docs rather than described as supported.

Gate on a3076fc: `mix precommit` — **3,956 tests, 0 failures**, credo/sobelow/dialyzer clean.

docs/concepts/vault.md
```diff
@@ -115,6 +115,21 @@ Use [inference credentials](../integrations/index.md) for model API keys.
...
 a Vault.
 
+## Secret expiry
+
+A secret can carry an expiry date. The date is optional metadata. Fountain
+does not block an expired secret, because an absent variable fails worse than
+a stale one.
+
+Fountain sends one email before the date arrives. The default notice window
+is seven days. Set `secret_expiry_notice_days` to change the window. Set it
```
jhgaylor: `secret_expiry_notice_days` isn't settable by anyone reading this: it's only read with `Application.get_env(:fountain, :secret_expiry_notice_days, 7)` and never assigned from the environment, so a self-hoster has nothing to set and the hosted product is pinned at 7.

Wire it the way the other operator knobs are, e.g. next to `unverified_prune_after_days` in `config/runtime.exs`:

```elixir
secret_expiry_notice_days: parse_bound.("SECRET_EXPIRY_NOTICE_DAYS", "7"),
```

plus a commented `# SECRET_EXPIRY_NOTICE_DAYS=7` line in `.env.example`, and have this sentence name the env var. Once that exists it also becomes the single source for the default, so the two hard-coded `7`s in `form.ex` and the sweeper can go.

apps/fountain/lib/fountain_web/live/vaults_live/form.ex
```diff
@@ -127,6 +127,55 @@ defmodule FountainWeb.VaultsLive.Form do
...
+    cond do
+      days < 0 -> {:expired, "expired #{DateTime.to_date(expires_at)}"}
+      days == 0 -> {:expiring, "expires today"}
+      # Amber exactly when the email would have fired: one window, one config.
+      days <= expiry_notice_days() -> {:expiring, "expires in #{days}d"}
+      true -> {:ok, "expires #{DateTime.to_date(expires_at)}"}
+    end
+  end
+
+  defp expiry_notice_days do
```
jhgaylor: nit, optional: this is now the same function as `SecretExpirySweeper.notice_days/0`. One public `SecretExpirySweeper.notice_days/0` called from here would keep the two from drifting — especially once the runtime.exs wiring gives it a real default to read.

apps/fountain/test/fountain_web/controllers/vault_secret_controller_test.exs
```diff
@@ -45,9 +45,26 @@ defmodule FountainWeb.VaultSecretControllerTest do
 
       body = json_response(conn, 201)
       assert body["data"]["key"] == "API_TOKEN"
+      assert body["data"]["expires_at"] == nil
       refute Map.has_key?(body["data"], "value")
     end
 
+    test "accepts and returns an optional expiry", %{conn: conn, user: user, raw_key: raw_key} do
```
jhgaylor, line 52: Suggestion: the docs now promise that `expires_at: null` clears a stored expiry, and that a POST without the key leaves it alone. Both hold today (I checked with a throwaway test against this branch), but neither is pinned. A sibling test here — set an expiry, POST the same key with no `expires_at` and assert it survives, POST with `expires_at: null` and assert it's gone — would keep the docs honest.

jhgaylor (commented):
Thanks — every item from the first pass is addressed, and addressed well (the blank-date fix is the right shape, and the notice-window derivation is cleaner than what I asked for). I re-ran the gates on `a3076fc`: format, `compile --warnings-as-errors`, `credo --strict`, dialyzer, the SDK drift check (zero diff against a fresh generation), `docs-style.py`, `vale lint docs` (0 errors / 0 warnings), and the touched suites plus guardrail/API-spec/docs/email — 166 tests, 0 failures.

I also checked the one documented behaviour the PR doesn't test — `expires_at: null` over the API clearing a stored expiry — with a throwaway controller test against the branch: set → rotate with the key absent keeps it → `null` clears it. It passes. Inline suggestion to add that test, since the docs now promise it.

**One thing left before I approve**, inline on `docs/concepts/vault.md`: the docs tell a reader to set `secret_expiry_notice_days`, but nothing exposes it — it's only read via `Application.get_env` with a code default, and isn't in `config/runtime.exs` or `.env.example`. Six-line fix following the existing pattern. Repo rule is that docs don't describe behaviour that isn't built, so I'd rather not merge the sentence as-is.

One optional nit inline. Approve as soon as the env var is wired.

jhgaylor:
Pushed `eabe093e` onto this branch so it can merge today — hope that's alright. It's the three things from the last round, nothing else:

- **`SECRET_EXPIRY_NOTICE_DAYS`** wired through `config/runtime.exs` with `parse_bound` (so `0` is accepted and a non-integer refuses to boot, like the other knobs), a commented line in `.env.example`, a row in `docs/configuration.md` — `config_reference_test` requires one for every env var `runtime.exs` reads, which I found out the same way you would have — and the vault docs + CHANGELOG now name the env var.
- **`SecretExpirySweeper.notice_days/0`** is public and the only reader; `VaultsLive.Form` calls it.
- **Controller test** pinning the two documented API behaviours: a POST without `expires_at` keeps the stored expiry, `expires_at: null` clears it.

Gate on that tree: `mix precommit` — 3,957 tests, 0 failures; credo, dialyzer, sobelow clean; `docs-style.py` clean; `vale lint docs` 0 errors / 0 warnings. (A first push briefly included a stray `node_modules` symlink from my SDK check — force-pushed with lease to drop it; `7aa82dc9` → `eabe093e`, same content otherwise.)

Approving. Thanks for the quick turnarounds on this one.

jhgaylor (approved):
Approved at `eabe093e`. Every item from both rounds is in; the gates in the comment above were run on this exact tree.

jhgaylor:
One more push, `3f608121`: a merge of `main` into the branch. #1049 landed while this was being reviewed and both PRs added an `### Added` entry at the top of `[Unreleased]`, so GitHub marked the PR conflicting — and a conflicted PR gets no CI at all. Resolved by keeping both entries (yours after the versioning one); nothing else in the merge needed a hand. A merge rather than a rebase so your commits stay exactly as you signed them.

Re-ran the affected suites on the merged tree (the two #1049 migrations plus this one, then vaults, sweeper, controller, LiveView, config-reference, docs, audit guardrail, API spec): 162 tests, 0 failures. Approval stands.

jhgaylor:
Last push, `8bd4aa1d`: the SDK release gate flagged (correctly) that this PR changes what `@agentshit/fountain-sdk` ships — the regenerated types — without a version bump, and a shipped change with no bump sits on `main` unreleased. Bumped to **0.1.4** (`package.json`, `USER_AGENT`, SDK `CHANGELOG.md` entry for `expires_at`); merging publishes it. Gate now reads "Release looks well-formed"; SDK typecheck, 71 tests and build green. Merging as soon as CI finishes.

PR #1049, opened 2026-08-23
feat(agents): config versioning with diff view and rollback

mdonigian:
## What

An agent gets edited over time and a bad prompt or config change is hard to trace — nothing records what the config was, so "why did it behave differently yesterday" has no answer. This snapshots the agent's config on every change, renders history as a per-field diff in the console, offers one-click rollback, and stamps each conversation with the version it launched under.

## How

- **`agent_versions`**: one immutable row per shape the agent has had — `version` counter + full string-keyed `config`. Version 1 is backfilled for every existing agent from its live row, so the invariant "every agent has ≥1 version" holds from the first deploy. Versions share the agent's lifetime (FK cascade); their absence from `RetentionPruner` is a stated decision in its moduledoc.
- **A deliberate departure from decisions/0013**: the audit trail records only *which* fields moved, never values. A version's whole point is restoring values, so `config` holds them. **ADR 0029** (in this PR) records the decision — the exception is scoped to one tenant-owned table with tenant-data lifecycle, not an erosion of the trail's rule. `okf validate decisions` passes and the index is regenerated.
- **Snapshot writes** share the transaction with the agent insert/update (`Repo.transaction(fn)` per house style — the repo has no `Ecto.Multi`, and dialyzer flags its opaque internals). A save that moves no config field writes no version. The audit call stays outside the transaction per the existing rule.
- **Rollback** applies a version's config as a *new* edit through `update_agent` — history is never rewritten, the config is re-validated on the way in (a snapshot referencing a since-removed sandbox provider is refused, not restored blind), and the audit row carries `rolled_back_to` metadata. Covered by the audit guardrail.
- **Conversations** gain `agent_version_id` — provenance only, exactly like the snapshotted `runtime`: the live agent row still drives the sandbox. Stamped by `start_conversation` and the team page's `open_on_sandbox` (nilify on version delete so a conversation never goes with it). Actually *running* a pinned version would mean swapping `ConversationServer`'s live agent reads and is deliberately out of scope.
- **Console**: `/agents/:id/versions` with per-field before/after diffs and a rollback button, linked from the edit page.
- **Export**: version history joins the account export — it's tenant data holding full values, so an export without it would be incomplete.

## Testing

- `mix precommit` fully green locally: 3,957 tests, 0 failures
- New coverage: v1 on create, next-version on config change, no version on no-op or rejected saves, tenant scoping, rollback round-trip, cascade on delete, conversation stamping, and the versions LiveView (render, diff, rollback, cross-tenant 404, auth redirect)


jhgaylor:
Thanks for this — it's a clean first contribution and the write-up made it easy to review. I ran it locally on the branch (CI hasn't run yet because a first-time fork PR needs a maintainer to approve the workflow; that's on us, not you).

## What I verified locally

- `mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict`, `mix sobelow --config`, `MIX_ENV=dev mix dialyzer`: all clean.
- Both migrations apply, roll back (`--step 2`) and re-apply against a fresh test DB. The backfill SQL parses against the real column types (`skills` is `jsonb[]`, so `to_jsonb` produces the same array Ecto writes).
- `@snapshot_fields` is exactly `Agent.changeset/2`'s cast list minus `user_id` — checked.
- `Agent.changeset/2` has exactly two call sites, both in `update_agent`/`create_agent`, so no code path writes agent config without a version (one DB-level exception below).
- Both `_unsafe_current_version_id/1` call sites sit directly after a tenant-scoped `Agents.get_agent/2` in the same function with the ownership comment — matches the CLAUDE.md contract.
- Targeted tests (agents, audit guardrail, conversation start, versions LiveView, exports, agents LiveView, team, OpenAPI spec, docs, accounts): 337 tests, 0 failures.
- Full suite (`mix test` at the umbrella root, core + `ee/`): 3,957 tests, 0 failures.

## Things to change before merge

1. **`CHANGELOG.md`**: add an entry under `[Unreleased]` / `### Added`. Every user-visible change gets one (see the plan-caps entry there for the shape).
2. **`decisions/`**: the "departure from 0013" is smaller than the PR body suggests — 0013 constrains the *audit trail*, not tenant tables, so there's no conflict to argue. But "agents are versioned; the version table holds full config values; versions are never pruned; a rollback is a new edit, not a rewrite" is a real design decision that future work will lean on (an API for versions, a conversation actually *running* a pinned version). Please add a short ADR from `decisions/0001-template.md` with frontmatter and refresh the index with `scripts/decisions-index.sh` — the index script doesn't need `okf` installed; CI runs `okf validate` and will tell us if the frontmatter is off.
3. **`docs/concepts/agent.md`**: a sentence or two on the History page and rollback. The manual is published at `/docs` only and is written in ASD-STE100 (`vale lint docs` gates it; `docs-redesign/08-simplified-technical-english.md` has the rules). Run `mix test apps/fountain/test/fountain/docs_test.exs` and `python3 scripts/docs-style.py` after.

## Worth fixing in this PR (small)

- `versions.ex` `load_versions/1`: `Enum.map(& &1)` is a no-op — drop it.
- Nit, optional: `String.to_integer/1` on the `phx-value-version` crashes the view on a non-integer. `agents_live/form.ex` does the same on its index values, so this is consistent with the neighbours; `Integer.parse/1` → flash would be tidier but I won't hold the PR on it.
- `update_agent/3` concurrent-edit race: when the `[agent_id, version]` unique index fires, the caller gets `{:error, %Ecto.Changeset{}}` for an **`AgentVersion`** changeset, not the `Agent` one the form/API is holding. The agents form would then try to render errors for fields it doesn't have. A `Repo.rollback` to a changeset error on the *agent* changeset (or a retry of `next_version_number`) keeps the contract that `update_agent` returns an `Agent` changeset on failure. Low probability, but the audit-and-form contract is worth keeping.

## Not blocking — please note as follow-ups (an issue each is fine)

- **The stamp is write-only.** `conversations.agent_version_id` is set but `conversation_json.ex` enumerates fields explicitly and doesn't include it, and there is no `/api/agents/:id/versions`. The conversations app and the SDK can't read the provenance this adds, so today the only reader is the console page. Exposing `agent_version_id` (and probably the `version` number) on the conversation JSON + OpenAPI schema is the obvious next step; a versions API after that.
- **One config change bypasses versioning at the DB level.** `agents.environment_id` is `on_delete: :nilify_all` (initial schema migration, line 38), so deleting an environment silently changes the live agent's config with no version row. The newest version then claims an `environment_id` that no longer exists, and the *next* edit's diff shows `environment_id: X → nil` attributed to the wrong change. Rollback to that version is correctly refused by the FK, so nothing breaks — but the "newest version == live row" invariant the page and the conversation stamp rely on has one hole. Same shape for `allowed_vault_ids` / `allowed_environment_ids` on vault/environment deletion (no FK, so those just go stale). Worth either writing a version from `Environments.delete_environment` or documenting the gap in the ADR.
- The export adds one query per agent (`list_agent_versions` inside the map). Fine at today's sizes; a single query grouped by `agent_id` if it ever matters.

Once the CHANGELOG/ADR/docs items land I'm happy to approve the workflow run and merge.

mdonigian:
All review items addressed — the before-merge trio in f59dba5 + 7eab92b, the small fixes in 7eab92b:

**Before merge**
1. **CHANGELOG**: entry under `[Unreleased]` / `### Added`.
2. **ADR**: ADR 0029 landed in f59dba5, and 7eab92b reframes it per your point — 0013 governs the trail, not tenant tables, so the ADR now states the design decision (full values, tenant lifecycle, never pruned, rollback-as-new-edit, provenance-only stamping) without arguing a conflict that isn't there. `okf validate decisions` clean, index regenerated.
3. **Docs**: `docs/concepts/agent.md` gains a "Config history and rollback" section — `vale` 0 errors 0 warnings, `docs-style.py` clean, `docs_test` green.

**Small fixes**
- `load_versions/1` no-op `Enum.map` dropped.
- `Integer.parse/1` + flash on the rollback event.
- The concurrent-edit race now keeps the changeset contract: a version-insert unique violation rolls back to an `:base` error on the **Agent** changeset the form/API is holding ("the agent was updated by someone else at the same moment — retry"), never the `AgentVersion` one.

**Follow-ups filed**: #1051 (expose `agent_version_id` + version number on conversation JSON, a versions read API — with the export N+1 noted there), #1052 (environment deletion changes config with no version row; also documented as a known gap in ADR 0029's Consequences).

Gate on 7eab92b: `mix precommit` — **3,957 tests, 0 failures**, credo/sobelow/dialyzer clean.

jhgaylor (approved):
Re-reviewed at `7eab92be`. Everything from the first pass is addressed, and the fixes are the right shape:

- **Race contract** — matching on `%Ecto.Changeset{data: %AgentVersion{}}` and rolling back to a `:base` error on the Agent changeset is exactly what the form/API callers need. Good message, too.
- **`Integer.parse` + the catch-all `_`** in the `with` covers both the parse `:error` and the `nil` from `get_agent_version`. Fine.
- **ADR 0029** now states the decision rather than arguing a conflict with 0013, names only tests that exist in this PR, documents #1052 as a known gap and says plainly that running a pinned version is unbuilt. That's the standard the repo wants.
- **CHANGELOG** and the **`docs/concepts/agent.md`** section read well; the docs section passes the STE gate.
- #1051 / #1052 filed with the follow-ups.

Re-ran locally at this head: compile `--warnings-as-errors`, format, `credo --strict`, sobelow, `okf validate decisions` (0 errors, index regen leaves no diff), `scripts/docs-style.py`, `vale lint docs` (0 errors, 0 warnings), both migrations up/down/up, and the affected test files (219 tests, 0 failures). Full suite locally at this head: 3,957 tests, 0 failures. CI on the same commit is green across the board (six partitions, coverage gate, static analysis, compose checks, okf validate).

Nothing further from me. Approved and ready to merge.

Issue #817, opened 2026-08-18
ACP: closing the connection at end_turn kills Claude Code's background tasks (Monitor / run_in_background / ScheduleWakeup) — scope the connection to the session


jhgaylor:
## Symptom

Conversation `4faea53d-f55e-4f02-b281-2e425879db35`, turn 18 (2026-08-18 16:48–16:54 UTC). The agent commits + pushes, then calls Claude Code's `Monitor` tool (a `gh run list` loop, 600 s timeout, task `b3lrt2xkw`), receives *"Monitor started… You will be notified on each event. Keep working — do not poll or sleep"*, writes "Waiting on this build before confirming the tint fix is live." — and `session/prompt` returns `end_turn` (log_event 140276). The follow-up never happens. The next thing in the transcript is the user's "hows it going?" four minutes later, on a fresh adapter, where the model has to re-check the build by hand.

Any turn that waits on something external (CI, a deploy, a remote queue) ends early this way. Under the current Claude Code harness foreground `sleep` is blocked and `Monitor` is the sanctioned way to wait, so this is by design on the agent's side, not a one-off.

## Cause — two halves, both confirmed in code

**1. How Claude Code resumes.** In `@agentclientprotocol/claude-agent-acp@0.66.0` (the pin in `runtimes/acp.ex`), background shells — `Monitor`, `Bash run_in_background`, `ScheduleWakeup` — deliberately do **not** hold the prompt open (`acp-agent.js` calls it the "shells-never-defer contract"; only async *subagents* defer settlement). When the task fires, the SDK wakes the model in an autonomous cycle whose `result` has `origin.kind = "task-notification"`; the adapter streams that follow-up as **out-of-turn** `session/update`s and never settles a `session/prompt` on it. The adapter's own comment names the hazard: *"ACP allows out-of-turn session/update, but many clients stop consuming at the prompt response."*

**2. Fountain is that client, and it kills the process.** `finish_acp_turn` (`conversation_server.ex`) runs on `{:done, stop_reason}` and immediately `close_stdin` + `stop_acp_peer`. The adapter's `index.js` treats stdin EOF as `connection.closed → agent.dispose() → process.exit(0)`, which tears down the `claude` subprocess and the Monitor loop with it. The conversation goes `idle`; 0017's idle timer eventually parks the sprite. So the notification that would have resumed the agent can never arrive.

[...]

PR #1050, opened 2026-08-23
docs(examples): fountain-contributor runs the gate in the foreground and ends no turn with work pending

### LINKED ISSUE #1046: fountain-contributor manifest: tell the agent to run the gate in the foreground and never end a turn with work pending
Observed during #1010's live validation (#1044 / #1045). Asked to fix a CI failure on its own PR, the `fountain-contributor` agent ran `mix precommit` in the background and ended its turn with "I'll push once the suite stage reports" — nothing was pushed, and the backgrounded gate died with the turn (the #817 shape: the ACP connection closing at end_turn kills background tasks). A second prompt telling it to run the gate in the foreground and state the pushed sha before finishing worked immediately.

The manifest's `system` prompt already says "read the output rather than the exit code"; it should also say:

- run `mix precommit` in the **foreground** — work you background dies when your turn ends;
- never end a turn with work pending: the pushed commit sha and the gate output you saw belong in the turn that did the work.

One-paragraph edit to `examples/agents/fountain-contributor/manifest.yml` (plus the note in its README if it seems worth a sentence). Remember the README's own caveat: editing the Environment's warm-start fields invalidates the checkpoint, but `Agent.spec.system` is not a warm-start field — this edit costs nothing.

mdonigian:
Closes #1046.

Two additions to the `fountain-contributor` system prompt, per the issue:

- **Run `mix precommit` in the foreground** — backgrounded work dies when the turn ends (#817), so a backgrounded gate is a gate that did not run.
- **Never end a turn with work pending** — the turn that did the work states the pushed commit sha and the gate output actually seen; "I'll push once the suite passes" is a turn that pushed nothing.

Plus one matching sentence in the README's "What done looks like" section. `Agent.spec.system` is not a warm-start field, so this costs no checkpoint rebuild (the README's caveat is about `setup_script`/`packages`/`repositories`).

**Gate output seen**: `go test ./internal/manifest/...` → `ok github.com/BinaryBourbon/fountain/cli/internal/manifest 0.188s` (the examples-parse test the README names). No Elixir code touched, so the Elixir gate is CI's to confirm.

Pushed sha: e7ac521.

### DIFF
--- a/examples/agents/fountain-contributor/manifest.yml
+++ b/examples/agents/fountain-contributor/manifest.yml
@@ -137,11 +137,17 @@ spec:
       3. Read CONTRIBUTING.md, and decisions/index.md if the change is
          architectural.
 
-    Before you push: `mix precommit`. Read the output rather than the exit
-    code, an alias stage can fail while the alias exits 0. Confirm you reached
+    Before you push: `mix precommit`, run in the **foreground** — anything
+    you background dies when your turn ends, and a gate that died is a gate
+    that did not run. Read the output rather than the exit code, an alias
+    stage can fail while the alias exits 0. Confirm you reached
     "N tests, 0 failures". If you touched docs/, also run
     `mix test apps/fountain/test/fountain/docs_test.exs`,
     `python3 scripts/docs-style.py` and `vale lint docs`.
+
+    Never end a turn with work pending. "I'll push once the suite passes" is
+    a turn that pushed nothing: the turn that did the work is the turn that
+    states the pushed commit sha and the gate output you actually saw.
 
     Commit with `git commit -s` (DCO). Never push to main. One pull request,
     with what changed and why, and the gate output you actually saw.
--- a/examples/agents/fountain-contributor/README.md
+++ b/examples/agents/fountain-contributor/README.md
@@ -84,10 +84,13 @@ and test, so the suite runs in the sandbox without it.
 
 ## What "done" looks like
 
-A green `mix precommit` in the sandbox — the agent is told to read the
-output, not the exit code — and one pull request on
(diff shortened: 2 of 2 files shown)


examples/agents/fountain-contributor/manifest.yml
```diff
@@ -137,12 +137,18 @@ spec:
...
+    you background dies when your turn ends, and a gate that died is a gate
+    that did not run. Read the output rather than the exit code, an alias
+    stage can fail while the alias exits 0. Confirm you reached
     "N tests, 0 failures". If you touched docs/, also run
     `mix test apps/fountain/test/fountain/docs_test.exs`,
     `python3 scripts/docs-style.py` and `vale lint docs`.
 
+    Never end a turn with work pending. "I'll push once the suite passes" is
+    a turn that pushed nothing: the turn that did the work is the turn that
+    states the pushed commit sha and the gate output you actually saw.
```
jhgaylor, line 150: nit, non-blocking: this paragraph ends with "the gate output you actually saw", and the closing paragraph two lines down ends with "and the gate output you actually saw" as well. Fine to leave; the two could be folded into one if you're touching it again.

jhgaylor (approved):
Exactly the edit #1046 asked for, and nothing more. Both bullets from the issue are in the system prompt, the README gets its one matching sentence, and you're right that `Agent.spec.system` isn't a warm-start field so this costs no checkpoint.

Verified locally on `e7ac521`: `go test ./internal/manifest/...` → ok (the examples-parse guard), `go vet ./...` clean. `examples/` sits outside `docs/`, so none of the docs linters apply.

One optional nit inline. Also: the PR body follows the discipline it adds — sha and gate output stated. Appreciated.

Issue #1026, opened 2026-08-23
The pricing page promises a queue at the concurrency cap; nothing queues


jhgaylor:
The pricing section tells a buyer that work waits behind the concurrency cap. Nothing waits. The next sandbox is refused.

This sits on the sentence the page itself nominates as the one to decide on: *"the number to choose a plan on is how many agents you want working at once."*

## What the page says

`apps/fountain/lib/fountain_web/controllers/marketing_html/home.html.heex`

- L220 — "concurrent sandboxes. **Queue as much work as you like behind them.**"
- L263 — "A sandbox beyond it **waits for a free slot rather than starting**, so the number to choose a plan on is how many agents you want working at once."

`docs/guides/operate/plans-and-prices.md` L72 carries the same claim: "A tenant at the concurrency cap waits for a free slot."

## What the code does

`Fountain.Quotas.check_sandbox_quota/2` returns `{:error, {:sandbox_quota_exceeded, %{count: n, limit: n}}}`, and `FountainWeb.FallbackController` turns that into:

> You have {count} of {limit} concurrent sandboxes in use. Terminate a conversation before starting another.

There is no queue, no waiting slot, and no retry. `queue_initial_prompt/3` in `ConversationServer` is unrelated: it queues a prompt behind provisioning *within* a conversation that already has a sandbox.

## Why it matters

A developer on Solo fans out six conversations, expecting the sixth to wait, and gets a 4xx on their first day. The gap is discovered at exactly the moment the product is meant to feel effortless, on a promise they used to pick the tier.

It is worse than an ordinary copy bug because `apps/fountain/test/fountain_web/controllers/marketing_pricing_test.exs:76` asserts `body =~ "waits for a free slot"`. The claim is now pinned by a passing test, so it reads as verified when what is verified is that the sentence is on the page.

ADR 0026 and the pre-#1024 wording of the operator guide both had this right: the cap *refuses* the next start and leaves running work alone. The queue language appears to have come in with the #1024 rewrite.

## Recommended fix: change the copy, not the product

- L220 — drop "Queue as much work as you like behind them."
- L263 — "A sandbox beyond it is refused until a slot frees. Terminate a conversation, or upgrade."
- `docs/guides/operate/plans-and-prices.md` L72 — same correction.
- `marketing_pricing_test.exs` — assert the refusal wording instead.

[...]

PR #1054, opened 2026-08-23
feat(sandbox): bounded, opt-in queue at the concurrency cap

### LINKED ISSUE #1033: Queue sandbox requests at the concurrency cap instead of refusing them

## What happens today

The refusal surfaces as:

## What to build

Sketch, not a decision:

## Not this

mdonigian:
Closes #1033.

Implements the middle position the issue lays out — **queue within a bounded wait, refuse past it, and say so** — written down first as **ADR 0030** (in this PR; amends ADR 0005) and then built.

## Design choices (from the issue's numbered list)

1. **A separate `sandbox_requests` table, not a `queued` conversation status** — a queued start has no sandbox, and a new status would leak into every list view, the SSE stream and the SDK enum. The conversation exists only when the slot is won.
2. **Drain on slot-free**: `Conversations.update_sandbox/2` is already the documented choke point for every status change, so one poke there covers terminate, fail, suspend (ADR 0017) and the reaper's stuck-row release by construction. Oban job, unique per user for 30s; the 5-minute cron is the backstop and expiry sweep, never the trigger.
3. **Fairness**: drain jobs keyed per user; the slot is won under `with_sandbox_reservation/3`'s per-user advisory lock.
4. **Bounds**: 10 deep / 1 hour (`:sandbox_queue_max_depth`, `:sandbox_queue_max_wait_seconds`). At depth, enqueue returns the same 429 the caller opted out of; past the wait, the request expires with an audit event.
5. **Wakes keep refusing; only fresh starts queue** — with one twist the issue's motivating case forces: scheduled teammate runs hit the cap *via* the wake path, so they queue at the `run_schedule` level (kind `schedule_run`, re-fired by the drainer), unconditionally and deduped per schedule. Interactive wakes still refuse.
6. **API contract**: explicit opt-in — `queue: true` on the create body → `202 Accepted` with the request and its position. Without the flag, today's 429 for every existing client. Image-carrying requests are never queued (the queue stores JSON-safe attrs). `GET /api/sandbox-queue` lists (list order is queue order), `DELETE /api/sandbox-queue/:id` cancels. OpenAPI + regenerated TS SDK included.
7. **Visibility**: position on the 202 and the list; `sandbox_request.enqueued/started/cancelled/expired/failed` audit events (guardrail entries added); drainer acts as `system:sandbox_queue`.

**Copy**: the #1027 corrections go back the other way in this PR, with the bound stated — "the queue delays the cap, it never raises it" — on the marketing page, in `plans-and-prices.md`, and pinned by the reworked `marketing_pricing_test`.

## Gate output seen

- `mix precommit`: **3,974 tests, 0 failures** (credo/sobelow/dialyzer clean)
- `okf validate decisions`: valid, 0 errors, 0 warnings (index regenerated)
- `python3 scripts/docs-style.py`: clean; `vale lint docs`: 0 errors, 0 warnings
- `mix openapi.spec.json --spec FountainWeb.ApiSpec` + `jq empty`: valid; SDK regenerated with CI's exact command

Pushed sha: 8408d25.


jhgaylor:
Thanks — this is a careful piece of work, and the ADR-first approach is exactly what #1033 asked for. The design choices all land where the issue pointed (separate table, drain from `update_sandbox/2`, per-user Oban key, opt-in 202, schedules opt in unconditionally), and the guardrail entries (audit, schema enum) are in. I ran it locally on the branch; CI hasn't run because a first-time fork PR needs a maintainer to approve the workflow — on us.

## What I verified locally

- `mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict`, `mix sobelow --config`, `MIX_ENV=dev mix dialyzer`: clean.
- The migration applies, rolls back and re-applies against a fresh test DB.
- Full suite (`mix test` at the umbrella root, core + `ee/`): 3,974 tests, 0 failures.
- `start_conversation/2` reads `attrs["prompt"]` itself, so a replayed `start` keeps its prompt; `{:error, :subscription_required}` is a tuple, not a raise, so a suspended account's queued request fails cleanly at drain rather than crashing the job. Both checked.

## Must fix before merge

1. **The drainer's `unique` window drops real pokes.** Oban 2.23's default unique `states` are `[:available, :scheduled, :executing, :retryable, :completed]` (`deps/oban/lib/oban/job.ex:307`). With `unique: [keys: [:user_id], period: 30]`, a poke that lands while a drain for that user is *executing*, or within 30s after one *completed*, is deduplicated against it and never runs. Concrete case: a drain starts one request, hits the quota, completes; 10s later another sandbox terminates and pokes — dropped; the freed slot sits idle until the 5-minute cron. A fan-out terminating five sandboxes over a minute drains once. That contradicts the ADR's "drain on slot-free, the cron is only the backstop" — in practice the cron *is* the trigger for anything after the first event in a window. Fix: `unique: [keys: [:user_id], period: 30, states: [:available, :scheduled]]` — dedupe only jobs that haven't started, so a poke during or after a drain schedules exactly one follow-up. (Oban.Testing's `:manual` mode doesn't exercise uniqueness, so this won't show in the suite; the moduledoc's "a reaper pass releasing five rows is one drain" claim still holds with the narrower states.)

2. **ADR 0030 describes tests that don't exist.** The Status block says coverage includes "the queue cases in `conversations_start_test.exs`" — that file isn't in this PR and has no queue cases. CLAUDE.md's rule is that ADRs must not describe unbuilt things as built; please correct the list. While there: ADR 0005's frontmatter `description` carries "Amended by 0018 …" — add "Amended by 0030" the same way, so `okf backlinks` and a reader of 0005 both see it. Telemetry for depth/wait (item 7 in #1033) isn't built either; a one-line "not built" under Consequences keeps the ADR honest.

3. **`CHANGELOG.md`**: an `[Unreleased]` / `### Added` entry — this is user-visible API behaviour and copy. The plan-caps entry there shows the shape.

4. **`docs/api.md`**: it enumerates endpoints in fenced blocks (`POST /api/conversations` etc.) and has a 429 row. Add `queue: true` → `202` on the create call, and `GET /api/sandbox-queue` / `DELETE /api/sandbox-queue/:id`. `docs/sdk.md`'s `QuotaExceededError` row ("Terminate a conversation first") can stay as-is until the SDK can opt in (see follow-ups). Run `mix test apps/fountain/test/fountain/docs_test.exs`, `python3 scripts/docs-style.py` and `vale lint docs` after.

## Worth fixing in this PR

- **Claim the row before attempting it.** Prod runs two replicas and `maintenance: 1` is per node, so two drains for one user can overlap (event poke on one node, cron re-poke on the other, once the unique window is fixed above). Both read the same `next_queued/1` row and both call `start_conversation` — the advisory lock keeps the *cap* honest, but the same request starts twice and consumes two slots for one queued job. A CAS is enough: `Repo.update_all(from(r in Request, where: r.id == ^id and r.status == "queued"), set: [status: "starting"])` returning `{1, _}` claims it, `{0, _}` skips it. (Needs a `starting` value in `@statuses` and the schema enum — the guardrail test will tell you.)

- **Replay through the same door the request came in.** The controller calls `start_or_resume_conversation/2` (channel-aware: resumes a bound conversation, `fresh` unbinds first); `SandboxQueue.attempt/1` replays through `start_conversation/2`. A queued request carrying `channel_id` therefore creates and binds a new conversation at drain time without checking whether one got bound in the meantime (another queued request for the same channel, or a direct call that succeeded once a slot freed). "Newest wins" tolerates it, but it's a semantic the queue introduces. Either call `start_or_resume_conversation` in `attempt/1` and accept `{:ok, conv, _outcome}`, or refuse to queue when `channel_id` is present — the latter is consistent with "wakes keep refusing".

- **Terminal rows keep the prompt forever.** `attrs` holds the caller's prompt and settings; `started` / `cancelled` / `expired` / `failed` rows retain it indefinitely, it's not in the account export, and the ADR leaves retention open. Clearing `attrs` (and `schedule_id` isn't sensitive, keep it) on every terminal transition — `finish/2`, `cancel_request/2`, `expire_overdue/1` — removes the data question: what's left is small provenance rows, which is what the ADR's "terminal rows are small" already assumes.

## Not blocking — follow-ups (an issue each is fine)

- **The SDK can't opt in.** `sdk/typescript/src/client.ts` builds the create body by hand and never sends `queue`; there's no `sandboxQueue` resource. The regenerated `openapi.ts` gives the *types*, but an SDK user has no way to reach the 202 today. When `run()` grows a `queue` option it will also need to handle the 202 body (both shapes have `data.id`, so `api.data<ConversationRecord>` would happily treat a request id as a conversation id — worth a discriminated check).
- `maintenance: 1` is shared with `SandboxReaper`, `RetentionPruner`, `TeamScheduler` and `UnverifiedAccountPruner`; a drain queues behind whatever's running. Fine for now; its own queue if drain latency ever matters.
- The team / conversations apps don't yet say "waiting for a free slot" — schedules show `last_error`, which is enough for the motivating case. Noting so nobody thinks the UI half shipped.

Happy to approve the workflow run once the must-fix items are in.

jhgaylor:
Follow-up, two days on: everything in my review above still applies, but `main` has moved under this branch in ways that change more than a rebase would. Listing them so nobody spends a day resolving conflicts into a premise that no longer holds.

## What changed on `main` since 8408d25

1. **ADR number collision.** `main` now has `decisions/0030-prepaid-credits.md` and `0031-credits-are-the-product.md`. This PR's ADR needs to become **0035** (0032 and 0034 are taken too), with `decisions/index.md` regenerated (`scripts/decisions-index.sh`) and the "Amended by" line on 0005 updated to match.

2. **The copy this PR "makes true again" was rewritten.** There are no plans or tiers any more (ADR 0031). The concurrency cap is `clamp(balance ÷ SANDBOX_RESERVE_CENTS, 2, 20)` — funded by the credit balance — and the homepage now reads *"A start beyond that is refused rather than queued; terminate a conversation, or top up."* `docs/guides/operate/plans-and-prices.md` was rewritten around credits as well. The copy half of this PR has to be redone against that text, not rebased; the same three files (`home.html.heex`, `plans-and-prices.md`, `marketing_pricing_test.exs`) are still the ones to touch.

3. **The ADR's product framing predates credits.** #1026's "refusal is the upgrade lever" was a tier argument; with credits the lever is "top up for more concurrency". The bound still makes sense, but the ADR should state it against a balance-funded cap. Concretely: `Billing.check_spend/1` (→ 402 `insufficient_credits`) is now a second reason a queued request can fail at drain time, beside the quota, and `attempt/1` should treat it as a terminal `failed` with that reason rather than retrying.

4. **A new refusal: `:fleet_full` (503).** `Quotas.check_fleet_ceiling/1` bounds live sandboxes across every tenant under a global advisory lock, taken before the per-user one. A start can now be refused by the fleet ceiling with the user's own cap untouched. `enqueue` currently only knows `sandbox_quota_exceeded`; I think fleet-full should queue the same way when `queue: true` (it's the same shape from the caller's side), and the drainer's poke from `update_sandbox/2` already covers it because any termination frees a fleet slot. Your call, but the ADR should say which.

5. **Shared sandboxes (ADR 0023, #1061).** Several conversations can hold one sandbox, so "a conversation ended" is no longer "a slot freed". `update_sandbox/2` is still the right poke point (it's sandbox status that counts against the cap), just worth a sentence so the drainer's trigger isn't misread.

6. **Ten files conflict** on a test merge: `conversation_controller.ex`, `router.ex`, `config/config.exs`, `audit_guardrail_test.exs`, `schema_enum_guardrail_test.exs`, `schedules_test.exs`, `decisions/index.md`, plus the three copy files. All shared-edit files; the queue itself (`sandbox_queue.ex`, `request.ex`, the drainer, the migration, the new controller/JSON view, `sandbox_queue_test.exs`) rebases clean. A conflicted fork PR gets no CI at all, so nothing goes green until this is done regardless.

## Proposal

The queue code is good and the scheduled-run loss it fixes is real, so I'd rather not hand you all of that as homework. If you're OK with it, I'll push the rebase onto your branch myself (maintainer push to the fork): renumber the ADR to 0035, resolve the six mechanical conflicts, and move the copy to the credits-era text. Then what's left for you is the four must-fixes from the review above plus items 3 and 4 here (a few lines in the ADR and `enqueue`/`attempt`), and I'll approve the workflow run.

Say the word and I'll push; if you'd rather do the rebase yourself, that's fine too — just note the ADR number and the copy point before you start.

jhgaylor:
We have moved this work to #1481.

Since this PR was opened, the codebase has changed substantially across sandbox reservations, fleet limits, prepaid credits, scheduled teammate runs, conversation replay, API contracts, telemetry, and the ADR sequence. Rebasing or reviving this branch would have meant resolving the old implementation against architecture that no longer exists in the same form.

#1481 is therefore a clean implementation from current main. It preserves the useful product shape and lessons from this PR, while incorporating the review feedback here: claim-before-start draining, fleet-full queueing, channel-aware replay, current credit and inference gates, terminal prompt removal, visibility and cancellation endpoints, documentation, audits, telemetry, and generated contract coverage.

Thank you for the work and review captured here—it was valuable input to the replacement.

PR #1055, opened 2026-08-24
fix(test): FakeDaemon.stop/1 no longer races the socket it stops

jhgaylor:
## What

`main`'s CI run for `c35e8dd6` (#1049's merge) failed in partition 3 on `runner_conformance_test.exs:8` — not in the test, in its `on_exit`: `FakeDaemon.stop/1` exited with `no process` from `GenServer.stop/3`. The same partition was green on the PR at identical content; the job passed on re-run. https://github.com/BinaryBourbon/fountain/actions/runs/32676969894

## Why

`stop/1` did `if Process.alive?(socket), do: GenServer.stop(socket, :shutdown, 1_000)` — a check-then-act. A socket that dies between the two (a test that closed the connection, or the daemon going first) turns every runner test's teardown into a `:noproc` exit that fails a test which already passed.

## How

Drop the aliveness check; stop the socket if it is there and catch the exit. The monitor + `:DOWN` wait that already followed settles the rest either way. The daemon kill is unchanged.

## Testing

- New `fake_daemon_test.exs`: `stop/1` is `:ok` with the socket up and with it already down. **This pins the contract, not the race** — a death between two adjacent calls can't be scheduled deterministically. Verified by reverting the fix: the old code passes it too (its `alive?` guard handles an *already*-dead socket; what it can't handle is the window). So the test guards the shape of the fix; the fix is correct by construction.
- `runner_conformance_test.exs` ×5 and all eight `FakeDaemon` callers: 61 tests, 0 failures. `mix format --check-formatted` and `credo --strict` clean. Full suite running locally as this opens; CI is the gate.

### DIFF
--- a/apps/fountain/test/fountain/runners/fake_daemon_test.exs
+++ b/apps/fountain/test/fountain/runners/fake_daemon_test.exs
@@ -1,0 +1,43 @@
+defmodule Fountain.Runners.FakeDaemonTest do
+  @moduledoc """
+  `FakeDaemon.stop/1` is every runner test's `on_exit`, so it must never
+  raise: a teardown that exits fails a test that already passed. The socket
+  it stops can be gone before it gets there — a test that closed it, the
+  daemon going first — which is the race that took a `main` run down
+  (32676969894, partition 3).
+
+  These pin the contract (`:ok` whether the socket is up or already down),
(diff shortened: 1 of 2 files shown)


jhgaylor:
Full suite locally on `7d474fad`: 3,963 tests, 0 failures (core + `ee/`).

PR #1148, opened 2026-08-25
feat(broker): Fountain brokers egress credentials natively; Agent Vault is gone (ADR 0019 §8 amended)

jhgaylor:
## Summary

Fountain is now its own egress credential broker. The Infisical Agent Vault
dependency that gate 1a (#1136) put behind `Fountain.Broker` is gone; the seam
held and nothing outside the module changed.

What the broker does is small, and it is all here now:

| piece | module |
|---|---|
| `CONNECT` + absolute-form forward proxy on `BROKER_LISTEN_PORT`, TLS terminated toward the sandbox with a per-host leaf, bytes relayed back unparsed (streaming stays streaming) | `Fountain.Broker.Proxy` (ThousandIsland handler) |
| root CA **derived** from `MASTER_SECRETS_KEY` (HKDF → P-256), fixed subject/serial/validity, so every replica presents leaves the sandbox trusts and nothing is stored | `Fountain.Broker.CA`, leaves cached in `Fountain.Broker.Certs` |
| session store: random token stored hashed, credentials as ciphertext under the tenant DEK, `services` (host → bearer / basic / api-key), `unmatched_host_policy`, TTL; resolved on any replica | `Fountain.Broker.Sessions` + `broker_sessions` migration |
| the header rewrite (drop `proxy-authorization`, replace the auth header wholesale on a matched host, passthrough or deny otherwise) | `Fountain.Broker.Injector` |
| the HTTP/1.1 slice the proxy needs: request heads, `Content-Length` / chunked body framing | `Fountain.Broker.HTTP` |

Every request logs conversation, method, host, path and outcome (`[:fountain, :broker, :request]` + a log line), never a header — the request-log half of gate 4 comes for free, with the attribution Agent Vault could not give a session token.

## Configuration

- **On:** `BROKER_LISTEN_PORT` + `BROKER_PROXY_URL`, together. `BROKER_TENANTS`, `BROKER_SESSION_TTL_SECONDS`, `BROKER_ALLOW_UNENFORCED` unchanged.
- **`BROKER_URL` / `BROKER_TOKEN` are not read.** Boot warns when they are set.
- The listener is plain HTTP, like Agent Vault's 14322 was; the Traefik TCP router in front of `broker.inevitable.fyi` just needs to point at the Fountain pods instead.

### Merging this is inert for prod — and switches Jake's account *off* the broker

Prod sets `BROKER_URL` / `BROKER_TOKEN` / `BROKER_PROXY_URL` / `BROKER_TENANTS=<jake>`. After this merges and before home-cloud is updated, `configured?/0` is false (no `BROKER_LISTEN_PORT`), so the one brokered tenant provisions on the pre-gate-1a path: real `GITHUB_TOKEN` in the sandbox, no floor. That is the "nothing live in prod without me knowing" rule working as intended, but it is a regression for that one account until the flip. The flip is a home-cloud PR: add `BROKER_LISTEN_PORT=14322` to the Fountain deployment, expose the port through the existing `IngressRouteTCP` (HostSNI `broker.inevitable.fyi` → fountain service :14322), drop `BROKER_URL`/`BROKER_TOKEN`, then delete the `agent-vault` namespace and its CNPG cluster.

## After the parity review (second commit)

Rebased onto #1147 (gate 1b): `prepare/4` takes the user id and the tenant's bindings, and the injector serves every binding shape — bearer, basic, `api-key` with a `prefix`, `custom` `{{ KEY }}` templates — and `host[:port][/path]` rules with wildcards. Two guards Agent Vault had are built: an origin resolving to a private/loopback/link-local address is refused before any connection (the dial goes to the vetted address), and under `deny` an unnamed host is refused at `CONNECT`. A WebSocket upgrade turns the client side into a byte pipe (the parity test found the tunnel parsing frames as HTTP). `agent_vault_parity_test.exs` replays Agent Vault's applicable proxy tests and lists what is deliberately not ported. The ADR records what gates 2–4 need from the proxy, checked against the code. Details in the comment below.

## Deviations worth knowing

- **Plain HTTP (absolute-form) is one request per connection**: the proxy adds `Connection: close` upstream and closes after the response. HTTPS tunnels keep-alive normally. Plain HTTP through the proxy is apt and little else.
- The root's self-signature is randomised (ECDSA), so the PEM bytes differ per derivation while subject, key and serial are fixed. Clients match trust anchors on subject + key and never verify a root's own signature; `ca_test.exs` proves a leaf signed after a re-derivation chains to the first root.
- `unmatched_host_policy` is still `passthrough` per session (gate 1a's setting); `deny` is built and tested in the injector and proxy — inside the tunnel and at `CONNECT` — waiting for gate 2.
- Not ported from Agent Vault, on purpose: placeholder substitution in path/query/arbitrary headers (the `telegram` preset is marked unusable), WebSocket frame rewriting, proxy-port auth rate limiting, body-size caps, IPv6 upstreams.
- New deps: `x509` (certificate building; pure Elixir) and `thousand_island` listed explicitly (it was already here under Bandit).

## Test plan

- `mix test` full suite green locally; the broker directory adds 31 tests including an end-to-end proxy test: a real Bandit HTTPS origin under its own CA, a sandbox-shaped client that `CONNECT`s, trusts the broker CA, and sees `Bearer ghp_real` echoed where it sent `__github_token__`; keep-alive with fixed-length and chunked bodies; split bodies; 407 / 502 / 403 paths; plain-HTTP absolute form; the telemetry event.
- `mix precommit` gates: compile `--warnings-as-errors`, format, credo `--strict`, sobelow, dialyzer, `deps.unlock --check-unused`.
- `config_reference_test`, `docs_test`, `audit_guardrail_test`, docs-style and vale pass.
- Not done here: a live sandbox loop. Recipe is the gate 1a one with `mix phx.server` + `BROKER_LISTEN_PORT=14322 BROKER_PROXY_URL=http://host.docker.internal:14322` and no docker Agent Vault.


jhgaylor:
## Parity review, and what changed after it

Asked "are we at parity with what we need?" against both Agent Vault's own test suite and #1147 (gate 1b), which merged after this branch was cut. Honest answer at the time: no. What was done about it, in the second commit:

**Gate 1b (#1147) — now served in full.** Rebased; `prepare/4` takes the user id *and* the tenant's bindings. The injector handles every shape a binding can be: bearer, basic with the tenant's username, `api-key` with a `prefix` (Discord's `Bot ` was silently dropped before), `custom` `{{ KEY }}` templates (would have crashed the connection before), and `host[:port][/path]` rules with `*.example.com` and trailing `*` (the catalog's `discord.com/api/*` never matched before). Main's binding tests pass against the DB-backed store.

**Two guards Agent Vault had that we lacked.** An origin resolving to a private/loopback/link-local address (RFC 1918, `169.254/16`, CGNAT, loopback) is refused with 403 *before* any connection, and the dial goes to the vetted address, not the name. Without this the broker is a door from a third-party sandbox into the cluster network — it should have blocked the prod flip on its own. And under `deny`, a host no rule names is refused at `CONNECT`, before a tunnel or handshake exists; that is what gate 2's `limited` will lean on.

**A real bug the parity test found.** After a WebSocket `101` the tunnel kept parsing the client's frames as HTTP and closed the connection. The client side now becomes a byte pipe after an `Upgrade` request.

**`agent_vault_parity_test.exs`** replays the applicable scenarios from `internal/mitm/proxy_test.go` and `forward_test.go` — injection, hop-by-hop stripping, arbitrary headers forwarded, keep-alive injection per request, prompt streaming (second chunk arrives before the origin finishes), untrusted upstream cert → 502 with no tunnel, passthrough keeps client auth, deny inside the tunnel and at CONNECT, 407 for no/bad token and wrong vault, WebSocket upgrade with the injected header — and names each upstream test it stands in for.

**Deliberately not ported** (listed at the bottom of that module and in the ADR): placeholder substitution in path/query/arbitrary headers (auth headers only; the `telegram` preset that needs it is marked unusable), WebSocket frame rewriting, auth-failure rate limiting on the proxy port, body-size caps (Agent Vault's response cap was unlimited by default too), IPv6 upstreams.

**Gates 2, 3, 4** — the ADR §8 amendment now records what each needs from the proxy, checked against the code: gate 2 is `prepare`-side only (passthrough services from `allowed_hosts` + `deny`); gate 3's header shapes and streaming are built, the work is `Claude.fall_back_to_api_key/2` and catalog entries; gate 4 has the per-request telemetry with the conversation id, but response status/latency needs response framing in the byte pump, which is that gate's first task.

Still not done: a live sandbox loop. Draft until Jake reviews.

jhgaylor:
Revisited 2026-09-02 under the component-libraries campaign (#1334): the native broker is going ahead, built library-first as `apps/managoat_broker` with this PR as the reference implementation rather than a rebase. Decision and reasoning on #1340.

PR #1182, opened 2026-08-25
feat(connections): Gmail via OAuth — tenant signs in once, Fountain holds the credential (#1178)

jhgaylor:
Closes #1178.

## What

A tenant signs in to Google **once** on *Account → Connections*; Fountain keeps the refresh token DEK-encrypted like any tenant secret and hands agents the capability, never the credential. Everything is gated on `Fountain.Broker.enabled_for?/1` (the broker ratchet), so it is live only for brokered accounts.

### 1. `Fountain.Connections` (issue piece 1)
- `connections` table: provider, account email, scopes, `env_key`, DEK-encrypted refresh + access token, expiry, status.
- Google OAuth client with `access_type=offline&prompt=consent`; `GET /connections/google/start` → consent → `callback` (signed `Phoenix.Token` state bound to the session nonce).
- `Connections.access_token/1` refreshes within 5 min of expiry; `invalid_grant` marks the row `revoked` and audits it.
- Console page: list, connect, revoke (also calls Google's revoke endpoint), remove. Nav link beside *Credential bindings*.
- `GET /api/connections`, `GET /api/connections/providers`, `GET/DELETE /api/connections/:id` (full scope), OpenAPI, SDK **1.7.0** with `client.connections`.
- Audit: `connection.created` / `connection.revoked` (provider, scopes, address — never a token), `connection.used` per send from the MCP server.

### 2. A Fountain-served Gmail MCP server (piece 2)
- An agent names the connection instead of a token: `{"gmail": {"connection": "<id>"}}`. `Fountain.Connections.McpServers.resolve/3` rewrites it at spawn into an HTTP MCP server at `POST /api/mcp/gmail/:conversation_id/:connection_id` carrying the conversation's callback token — so it flows through `.mcp.json` (claude) and `session/new` (everyone else) with no new contract. Rewritten again on reattach, when the token rotates.
- Tools: `gmail_search`, `gmail_get_thread`, `gmail_get_message`, `gmail_send`, `gmail_reply`, `gmail_modify_labels`, `gmail_list_labels`. The token is resolved server-side per call; a revoked connection answers **`connection revoked`**, not a 401 from Google.
- The controller checks: conversation is the caller's, its agent still names the connection, connection is the caller's.

### 3. Tenant-supplied OAuth'd MCP servers via the broker (piece 3)
[...]

Issue #1178, opened 2026-08-25
Gmail via OAuth for agent MCP servers: tenant signs in once, Fountain holds the credential


jhgaylor:
## Goal

An agent should be able to read and act on a tenant's Gmail through an MCP server, with the tenant signing in to Google **once** in the console and Fountain holding the credential. This is how Claude Managed Agents does connectors: the platform runs the OAuth authorization-code flow, stores the refresh token, and attaches a fresh access token to the agent's MCP requests. The agent never holds a Google credential.

## What exists today

- `agent.mcp_servers` (`apps/fountain/lib/fountain/agents/agent.ex:35`) is `${VAR}`-substituted from env/vault at spawn (`conversation_server.ex:678`) and written to `.mcp.json` (`runtimes/claude.ex:83`) or sent over ACP. So a tenant *can* wire an OAuth token in today, but only by pasting a long-lived token into a vault.
- The egress broker (`Fountain.Broker`, `Fountain.SecretBindings`, ADR 0019) already attaches a tenant secret to requests for a bound host as `bearer` / `api-key` / custom header, with the sandbox holding a placeholder.
- Fountain is an OAuth **server** (ADR 0021) and an OAuth client only for GitHub sign-in (Ueberauth), whose token is used for identity and not stored for reuse.
- Fountain already serves per-conversation MCP tools (`Fountain.Team.MCP`, `team_mcp_servers/1` in `conversation_server.ex`), authenticated by the conversation callback token.

### 1. `Fountain.Connections` — OAuth client + stored connections (the bulk)

- Tenant-scoped table: `provider`, `scopes`, `account_email`, `refresh_token` (DEK-encrypted via `Fountain.Crypto`), `access_token`, `expires_at`, `status` (`active` / `revoked`).
- Provider registry in config: `GOOGLE_OAUTH_CLIENT_ID` / `GOOGLE_OAUTH_CLIENT_SECRET`, redirect URI, scope presets. The flow must send `access_type=offline&prompt=consent` or Google returns no refresh token (Ueberauth's Google strategy does not force this).
- `GET /connections/google/start` → Google consent → callback stores tokens. Audit `connection.created` / `connection.revoked` with provider and scopes only (ADR 0013: never values).
- `Connections.access_token/1` returns a valid token, refreshing near expiry; `invalid_grant` marks the connection revoked and surfaces it.
- Console page (Account → Connections, or a new tab): list, connect, reconnect, revoke. Revoke also calls Google's revoke endpoint.
[...]

PR #1187, opened 2026-08-25
feat(connections): bring-your-own OAuth providers and MCP-spec discovery for remote MCP servers (#1186)

### LINKED ISSUE #1186: Tenant-defined OAuth connections: bring-your-own provider, and MCP-spec auth discovery for remote MCP servers
Follow-up to #1178 (#1182). Connections exist today for exactly one provider, Google, with Fountain's own OAuth client. This issue is the work to let a tenant connect **any** OAuth'd service and hand the token to an MCP server they supply.

## The shape

Two kinds of provider, one connection model:

| Kind | Who owns the OAuth client | How Fountain learns the endpoints |
|---|---|---|
| **`oauth2`** (classic: GitHub, Slack, Notion, Linear, Google Workspace beyond Gmail…) | The tenant registers an app at the provider and gives Fountain the client id/secret. Fountain cannot own an app on every provider, and restricted scopes need per-app verification anyway. | The tenant enters authorize / token / revoke URLs and scopes, or picks from a small catalog of presets. |
| **`mcp`** (a remote MCP server that implements the MCP authorization spec) | Nobody, ideally: RFC 7591 dynamic client registration. Fall back to a tenant-supplied client when the AS has no registration endpoint. | Discovery: `GET` the MCP URL → 401 `WWW-Authenticate: Bearer resource_metadata=…` → RFC 9728 protected-resource metadata → `authorization_servers` → RFC 8414 metadata → PKCE (S256) code flow with the `resource` parameter (RFC 8707). |

Google becomes the one **platform** provider (config-backed, as today). The Fountain-served Gmail MCP server stays Google-specific; a generic provider gets the **brokered token** path (piece 3 of #1178), which is the one that works for tenant-supplied servers.

### 1. Provider model (`Fountain.Connections.Provider`)
- Tenant-scoped `connection_providers`: `slug`, `name`, `kind` (`oauth2` | `mcp`), `authorize_url`, `token_url`, `revoke_url` (nullable), `scopes`, `client_id`, `client_secret_ciphertext` (DEK), `token_endpoint_auth` (`client_secret_post` | `client_secret_basic` | `none`), `pkce` (bool), `env_key` (unique per tenant, `UPPER_SNAKE`), `token_hosts` (where the implicit bearer binding goes), `account_label` strategy (a userinfo URL + JSON path, or "ask the user"), `mcp_url` + discovered metadata for `kind: mcp`.
- Platform providers (Google) are the same struct built from config, `user_id: nil`, so one code path serves both.
- Redirect URI per provider: `/connections/:provider_id/callback` (tenant providers) alongside the existing `/connections/google/callback`. The console shows it to copy into the provider's app registration.
- Validation / SSRF: `https` only, no private or link-local hosts for tenant-supplied URLs; `token_hosts` may not be a bare `*`.

### 2. Generic OAuth client (`Fountain.Connections.OAuth`, replacing `Connections.Google` internals)
- Code flow driven by the provider record; PKCE when set (always for `mcp`).
- Refresh: rotating refresh tokens (store the new one when the response carries it); no `expires_in` → treat as non-expiring; no refresh token → the connection goes `expired` when the access token does and the console says "reconnect"; provider-specific `invalid_grant` shapes map to `revoked`.
- Revoke: RFC 7009 when `revoke_url` is set, else local only.
- `connections` gains `provider_id` (nullable for the platform Google row) and `expired` joins the status enum.

### 3. MCP-spec discovery and DCR (`Fountain.Connections.McpDiscovery`)
- The 401 → resource metadata → AS metadata chain above, with a timeout and the SSRF guard.
- Dynamic client registration when `registration_endpoint` exists; persist the registered client per tenant+AS; reuse across connections to the same AS.
- Manual fallback: the tenant pastes a client id/secret when the AS has no registration endpoint.

### 4. Using a connection with a tenant-supplied MCP server
- Remote server: an `mcp_servers` entry `{"type":"http","url":"https://mcp.example.com/mcp","connection":"<id>"}` is resolved at spawn into the same entry with `Authorization: Bearer __<ENV_KEY>__` and an implicit bearer binding to the server's host, so the broker attaches the real token (rotation already re-uploads at each turn kick). `Connections.McpServers.resolve/3` grows this second shape next to the Fountain-served one.
- Stdio server in the sandbox: nothing new — the `ENV_KEY` placeholder plus the provider's `token_hosts` bindings already cover it.
- Non-brokered tenants stay excluded: the token would have to enter the sandbox in the clear.

### 5. Console
- *Account → Connections*: a **Providers** section — add/edit/delete a provider, the redirect URI to copy, a *Connect* button per provider; *Connect a remote MCP server* (enter a URL → discovery → consent).
- Agent form: the *Connected account* option offered for HTTP servers as well (URL + connection), not only the Fountain-served Gmail server.
- Credential bindings catalog already lists connection `env_key`s.

### 6. API / OpenAPI / SDK
- `GET/POST/PATCH/DELETE /api/connection-providers`, `POST /api/connection-providers/:id/discover` (mcp), connections carry `provider`; SDK `client.connections.providers` gains CRUD; regen types, bump.

### 7. Docs, audit, ADR
- Docs: a how-to "Connect a service with your own OAuth app" and "Connect a remote MCP server"; catalog page per platform provider.
- Audit: `connection_provider.created/updated/deleted`; `connection.*` metadata carries the provider slug.
- ADR: amend 0019 with the rotating-secret contract (still owed from #1178) and write a short ADR for connections (platform vs tenant providers, why DCR, why no Fountain-owned client per provider).

## Not in scope
- Fountain-served tool servers for providers other than Gmail (that is a per-provider tool layer; the brokered token covers generic use).
- Google app verification (calendar item, separate).
- Device-code or client-credentials grants.

## Done when
- A tenant registers their own GitHub OAuth app in the console, connects it, attaches a stdio GitHub MCP server that reads `GITHUB_ACCESS_TOKEN`, and a conversation lists their repos with no token in the sandbox.
- A tenant enters the URL of a remote MCP server that implements MCP auth with DCR, completes consent with no client id typed anywhere, attaches it by URL + connection, and a conversation calls a tool on it; the broker log shows the bearer attached to that host only.
- Revoke → `connection revoked` on the next call; a provider without refresh tokens shows `expired` and a reconnect prompt when its token lapses.

## Sizing
Provider model + generic OAuth client ~1 week; MCP discovery + DCR ~1 week; remote-server resolution + bindings ~2 days; console + API + SDK ~3–4 days; docs/ADR ~1 day.

jhgaylor:
Closes #1186. Follow-up to #1178 (#1182).

## What

Connections were Google-only with Fountain's own OAuth client. Now **Google is the one platform provider** and every other service is a provider the tenant defines on *Account → Connections*:

- **`oauth2`** — the tenant registers an app at the service (presets for GitHub, Slack, Notion, Linear fill the endpoints), pastes the redirect URI the console shows (`/connections/<provider id>/callback`) and the client id/secret. Scopes, PKCE, client-auth method, the env var the token is brokered under (`GITHUB_ACCESS_TOKEN` from the slug) and the hosts the broker attaches it to are all editable.
- **`mcp`** — paste a remote MCP server's URL and nothing else. `McpDiscovery` follows the MCP auth spec (`401` → RFC 9728 protected-resource metadata → RFC 8414 AS metadata), and registers a client by RFC 7591 DCR where offered (reused for a second server behind the same issuer). PKCE S256 + RFC 8707 `resource`. A server without registration takes a pasted client.
- **Using it**: `{"linear": {"type": "http", "url": "https://mcp.linear.app/mcp", "connection": "<id>"}}` in an agent's `mcp_servers` → the sandbox calls that server with a placeholder bearer and the broker swaps in the real token on that host only (implicit binding added at spawn; rotation re-uploads at turn kick as before). A stdio server that reads the env var needs nothing new. The Fountain-served Gmail server is unchanged.
- **Refresh follows the provider**: rotating refresh tokens stored; no `expires_in` = non-expiring; no refresh token → the connection goes `expired` (new status) when the token lapses, with *Reconnect* in the console. Provider-specific invalid-grant spellings map to `revoked`. RFC 7009 revoke when there is a `revoke_url`.
- **SSRF**: `UrlGuard` — https only, no IP literals/internal names, resolved addresses may not be private/link-local/CGNAT/metadata; applied at save and at every fetch, including the URLs discovery gets back. `token_hosts` may not carry a wildcard.
- **API** `GET/POST /api/connection-providers`, `GET/PATCH/DELETE /:id`, `POST /:id/discover`; connections carry `provider_id`. `GET /api/connections/providers` now returns the same full list. **SDK 1.8.0**: `client.connections.providers` (list/get/create/update/delete/discover) — `providers()` the method became `providers.list()` (breaking, noted in the SDK changelog).
- **Audit** `connection_provider.created/updated/deleted`, `connection.expired`; never a client secret or token. Guardrail entries added.
- **Docs** *Connect a service with your own OAuth app*, *Connect a remote MCP server*, `catalog/connections/{index,google}`. **ADR 0033** (platform vs tenant providers, why DCR, why no Fountain-owned client per provider) and the **ADR 0019 amendment** owed since #1178 (rotating-secret contract; the per-conversation vault deviation from §11).

Everything remains gated on `Broker.enabled_for?/1`, as #1178 was.

## Not in this PR (per the issue's "not in scope")
Fountain-served tool servers for providers other than Gmail; Google app verification; device-code / client-credentials grants.

## Verification (local, on this branch)
- `mix test`: 4003 tests, 0 failures after fixes (full run + `--failed` rerun); new suites: `connections/provider_test`, `mcp_discovery_test`, `url_guard_test`, `connection_provider_controller_test`, LiveView + controller additions.
- `mix compile --warnings-as-errors` (dev+test), `mix format --check-formatted`, `mix credo --strict`, `mix sobelow --config`, `MIX_ENV=dev mix dialyzer` — all clean.
- `mix openapi.spec.json` → SDK regen; `npm test` 74/74, `tsc --noEmit` clean.
- `vale` (0 errors/warnings), `python3 scripts/docs-style.py` clean, `docs_test.exs` 29/29; `okf validate decisions` valid (2 pre-existing warnings on 0030), index regenerated.
- Not yet done: the live done-when on prod (GitHub OAuth app → stdio GitHub MCP lists repos; a DCR MCP server end to end). That needs Jake's app registrations and a target server; happy to run it once merged.


jhgaylor (commented):
Found four blocking issues. CI is green, but these scenarios are not covered.

- **P1 — Refresh-token rotation races can revoke valid connections**  
  `apps/fountain/lib/fountain/connections.ex:514`  
  Multiple conversations can refresh the same stale connection concurrently. After one request rotates the refresh token, another can receive `invalid_grant` and mark the newly refreshed connection revoked. Serialize refreshes per connection and re-read/recheck freshness under the lock.

- **P2 — Reused DCR clients lack the new callback URI**  
  `apps/fountain/lib/fountain/connections.ex:191`  
  Registration includes only the first provider UUID-specific redirect URI, but a second provider behind the issuer reuses that client with a different callback. Standards-compliant authorization servers will reject it. Either register a new client, use a shared callback, or update the DCR registration redirect URI list.

- **P2 — Slack refresh failures are treated as successful responses**  
  `apps/fountain/lib/fountain/connections/oauth.ex:202`  
  `invalid_grant?/1` runs only for HTTP 4xx. Slack returns JSON errors such as `{"ok":false,"error":"invalid_refresh_token"}` and instructs callers to inspect `ok`, so these become `{:unexpected, body}` and never mark the connection revoked. Inspect decoded error bodies before accepting any 2xx response. [Slack official response documentation](https://docs.slack.dev/reference/methods/oauth.v2.access/)

- **P2 — Editing a provider slug duplicates connections on reconnect**  
  `apps/fountain/lib/fountain/connections.ex:252`  
  The API/UI allow changing `slug`, while existing connections retain the old denormalized `provider` value. Reconnecting then misses the existing row at line 325 and inserts another connection for the same provider/account. Make the slug immutable once connected, update dependent rows transactionally, or key reconnection uniqueness on `provider_id`.

jhgaylor:
All four findings addressed in b3fff74e, and the branch is rebased onto main (61a608f8 reconciles).

- **P1 refresh race**: `refresh_token/2` now takes a per-connection Postgres advisory lock (`pg_advisory_xact_lock`, namespace 4331, same pattern as `Quotas`) and re-reads the row under it. A caller that lost the race finds the row already fresh and is served the rotated token without touching the provider; only a row still stale under the lock refreshes, with the refresh token the row holds *now*. The refused/expired status writes and their audits stay outside the transaction (ADR 0013). Regression test: "a stale struct does not replay an already-rotated refresh token".
- **P2 DCR redirect URIs**: cross-provider client reuse is removed — every `mcp` provider registers its own client, since a registration names exactly one redirect URI and redirect URIs are derived from the provider id. Rediscovery still keeps a provider's own client while the issuer is unchanged. ADR 0033 and the remote-MCP guide updated; the discovery test now asserts the second provider registers with its own callback.
- **P2 Slack 2xx errors**: `token_request/2` decodes 2xx bodies and treats a body carrying `"ok": false` or an `"error"` with no `access_token` as an error, through the same `invalid_grant?/1` mapping — so Slack's `invalid_refresh_token` on HTTP 200 (and GitHub's `error=bad_refresh_token` on 200) revoke, while a transient 200-error (e.g. `ratelimited`) surfaces as `{:http, 200, body}` without touching the connection.
- **P2 slug edits**: reconnect is keyed on `provider_id` (with an `is_nil` branch for the platform provider), and a slug rename cascades to the denormalized `connections.provider` in the same transaction as the provider update. Regression test covers rename → reconnect finding the same row with the same `env_key`.

All four new tests were verified to fail against the pre-fix code.

Rebase notes: SDK is now **1.13.0** (main published 1.12.0 since this branch was cut; the 1.8.0 changelog entry moved under 1.13.0), `openapi.ts` regenerated from the merged spec, and the docs pages were brought up to the destink gate that landed with #1238. On the rebased tree: 4250 tests 0 failures, dialyzer/credo/sobelow/format clean, SDK `tsc` + 74/74, vale/docs-style/destink/okf clean.

https://claude.ai/code/session_01DZuneGM6ispiT7Q6uPGJij

PR #1201, opened 2026-08-26
feat(examples): Deep Agents orchestrator with Fountain agents as subagents

jhgaylor:
Follows #1199. The LangChain answer to "how does a framework reach a Fountain agent" is **a subagent, not a model**: the orchestrator (any tool-calling model) plans, and a Fountain agent does the work in its own sandbox and reports back once. Deep Agents has exactly that seam (`CompiledSubAgent` with a `runnable`).

## What

- `examples/deepagents-contractor/fountain_langchain.py` — `FountainAgent(name)` with three shapes over `POST /v1/chat/completions`: `.as_subagent(description)` (Deep Agents), `.as_tool()` (`create_agent` / any loop), `.as_runnable()` (`{"messages"} -> {"messages"}`). Thread key is `<langgraph thread_id>:<agent slug>` from the ambient config, so one Deep Agents thread keeps one sandbox per Fountain agent across turns and runs; `409` is retried on `Retry-After`; `reasoning_content` (provisioning stages) goes to stderr. Subagent `name` is a slug of the agent name because the orchestrator mistyped `Mend: github.com/...` on the first try in the smoke.
- `main.py` — `--list`, `-a NAME[=DESCRIPTION]` to hire agents, `--thread` to resume, prints each delegation and report.
- Uses the stock `openai` client, not `ChatOpenAI`, on purpose: a Fountain agent never emits `tool_calls` so it cannot be the model inside a LangGraph loop, and `langchain-openai` documents that it drops non-standard `reasoning_content`. The docs page says both.
- `docs/integrations/langchain.md`, a clients-overview row and paragraph, nav, a pointer from the OpenAI-compatible page, CHANGELOG. Also states what it is not: a Deep Agents sandbox backend needs `execute` in the sandbox, which Fountain does not expose.

## Verified

- Both docs test files (29, 0 failures), `vale lint docs` (no gated findings on the new page), `docs-style.py` clean.
- Against **prod** (managoat.com, `openai_compat` on for my account): runnable + subagent shape, two turns on one thread against `reflex-1` — turn 1 51s with provisioning streamed as reasoning, turn 2 2s and the sandbox remembered turn 1. Full `create_deep_agent` run with two hired agents: results in the comment below.

## Noticed on the way

- The `openai_compat` PostHog flag was created with `evaluation_runtime: server`, which Fountain's `/flags/?v=2` (project token) never matches — it read as flag-off. Switched to `all` like `team_comms`. Worth a line in the flag-flipping runbook.
- `fountain-contributor`'s setup script fails on prod (`setup failed: no reason given`) — surfaced cleanly as an `APIError` through the stream, unrelated to this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,12 @@ upgrade, is in
 
 ### Added
 
+- **LangChain and Deep Agents example.** `examples/deepagents-contractor`:
+  a Deep Agent orchestrator whose subagents are Fountain agents, over the
+  OpenAI-compatible API. `fountain_langchain.py` makes one agent a
+  `CompiledSubAgent`, a LangChain tool or a bare runnable, keyed to the
+  LangGraph `thread_id` so one thread keeps one sandbox per agent. Docs page
+  `docs/integrations/langchain.md`.
 - **OpenAI-compatible chat completions (alpha, flag `openai_compat`).**
   `POST /v1/chat/completions` and
   `GET /v1/models`, where the `model` is one of the tenant's agents (by name
--- a/docs/integrations/clients.md
+++ b/docs/integrations/clients.md
@@ -21,6 +21,7 @@ mail, OAuth, billing and error reports, read <!-- vale disable-line STE.IngForms
 | [Hermes Agent](hermes.md) | The HTTP API, through a plugin. | The Hermes host. |
 | [OpenBot](openbot.md) | [AG-UI](https://github.com/ag-ui-protocol/ag-ui) over HTTP. | The OpenBot host. |
 | [OpenAI-compatible API](openai-compatible.md) (alpha) | OpenAI chat completions over HTTP. | The client or the gateway. |
+| [LangChain and Deep Agents](langchain.md) (alpha) | The OpenAI-compatible API, from one Python file. | Your LangChain code. |
 | [Buzz](buzz.md) | A Nostr relay, hosted by Fountain. | The Buzz desktop, or `POST /api/buzz/agents`. |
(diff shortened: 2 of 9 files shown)


jhgaylor:
Prod smoke of the full `create_deep_agent` run (orchestrator `anthropic:claude-sonnet-5`, thread `smoke-deep-1`, two hired agents):

```
python main.py --thread smoke-deep-1 \
  -a "reflex-1=A general helper with a plain Linux sandbox..." \
  -a "Mend: github.com/BinaryBourbon/fountain=A maintainer of the repo; its sandbox has it cloned..." \
  "1) Ask the fountain maintainer how many ADR files are in decisions/ and the title of the highest-numbered one. 2) Ask the general helper what day of the week it is in UTC."
```

- `task` → reflex-1: reattached to an existing sandbox, answered "Wednesday" in seconds.
- `task` → Mend: provisioned, cloned, counted `decisions/` (33 files) and named `0035-openai-compatible-endpoint.md` with its title — the ADR #1199 merged. Correct.
- The lead's final answer summarised both reports and noted no PRs were produced.
- One wrinkle it self-corrected: the first `task` call typed `Mend` for the subagent named `Mend: github.com/BinaryBourbon/fountain` and got Deep Agents' "does not exist" message. That's why `as_subagent` now defaults the subagent name to a slug of the agent name (the real name rides in the description); this run was before that change.

Before that, the runnable alone against `reflex-1`: turn 1 51s (provision → broker → network streamed as `reasoning_content`), turn 2 on the same `X-Fountain-Thread` 2s and it remembered turn 1.

Issue #989, opened 2026-08-22
sandbox: a warm start skips the network policy, so a limited environment restored from a checkpoint runs unrestricted


jhgaylor:
Found while building #935. Latent today, but it is the one step in the warm-start skip list that is not disk state.

## The claim the code makes

`run_provisioning_pipeline/5` (`conversation_server.ex`):

```elixir
# Try a checkpoint restore first if the env has one. If restore
# succeeds, skip the slow steps (network policy + packages + clone +
# setup_script) — they all ran when the checkpoint was originally
# taken.
```

That reasoning holds for `install_packages`, `clone_repositories` and `run_setup_script`. All three write to the disk the checkpoint captures, so restoring the checkpoint restores their effect.

It does not hold for `apply_network_policy/3`. An egress policy is configuration on the *sandbox*, applied through `Fountain.Sandbox.apply_network_policy/2` against a `Handle` — a live API call to the provider, not a file. A warm start creates a **new** sandbox and restores a disk image into it. The new sandbox carries no policy.

## Consequence

An environment with `networking_type: "limited"` and a non-empty `checkpoint_id` provisions a sandbox with **no egress restriction at all**, on every provider, and reports `provision/done`. `limited` with no `allowed_hosts` is meant to deny everything; a warm start turns it into allow-everything. The failure is silent — there is no `network` stage event either way, so nothing in the transcript says the policy was skipped.

## Why it is latent rather than live

- `checkpoint_creation_enabled?/0` defaults to `false` (`conversation_server.ex:889`), so nothing sets `checkpoint_id`.
- #654 established that checkpoint ids are scoped to the sprite that created them, so a fresh sprite has nothing to restore from anyway. That is *why* creation is off.

So `attempt_warm_start/3`'s live branch is unreachable in practice. The bug is what happens the day the flag goes on, or the day a provider ships fork-from-checkpoint (which #936 lists as an option and #654 left the flag in place for).

## Fix

Apply the network policy on both arms. It is the one cheap step in the list (a single API call, explicitly labelled "fast" in `Provisioning`'s moduledoc pipeline comment), so hoisting it out of the `:cold` branch costs a warm start nothing:

```elixir
defp run_provisioning_pipeline(handle, env, sprite_env, secrets, conv_id) do
  with :ok <- Provisioning.apply_network_policy(handle, env, conv_id) do
    case attempt_warm_start(handle, env, conv_id) do
      :warm_started -> :ok
      :cold -> # packages, clone, setup_script
    end
[...]

PR #1209, opened 2026-08-26
docs: ADR 0036 — sandboxes on the tenant's tailnet, inbound only

jhgaylor:
## What

ADR 0036, **Proposed**. Nothing described in it is built, and the file says so at the top: there is no Tailscale code in the repo today, and gate 0 has not been run, so the central premise — that a userspace-mode node comes up at all inside a Sprites or E2B sandbox — is an expectation rather than a measurement.

Adds `decisions/0036-sandboxes-on-the-tenants-tailnet.md` and the regenerated `decisions/index.md`. `okf validate decisions` passes with 0 errors (the two warnings are pre-existing `agent:claude-code` actor strings on 0006 and 0030).

## The decision

**Userspace tailscale is asymmetric, and that asymmetry is the whole ADR.** Inbound — the sandbox appears on your tailnet and you can reach it — needs no privileges and no per-tool configuration. Outbound — the agent reaches your private hosts — needs `/dev/net/tun` or the proxy variables `Fountain.Broker.sandbox_env/1` already owns, and two proxies cannot own one variable. So inbound ships, outbound is refused by name at preflight, and the schema carries two flags rather than one boolean so the half that does not work cannot be promised silently.

Around that:

- **Keys are minted, never pasted.** A stored Tailscale OAuth client mints an ephemeral, preauthorized, tagged, single-use key per join. Ephemeral is what keeps a tenant's admin console from filling with dead sandboxes, and it is why nothing here needs a reaper of its own.
- **Three lifecycle traps**, each anchored to a mistake this repo has already made: the join runs on both provisioning arms (the #989 network-policy lesson), it re-runs on every wake (ADR 0017 parks the sandbox and the ephemeral node disappears), and `tailscaled.state` stays out of checkpoints or every warm start is the same node, flapping in the tenant's tailnet with no error visible in Fountain.
- **Providers answer for themselves** via a `:tailnet` capability; `Fountain.Sandbox.Runner` refuses it, because the machine is the user's and so is its network.
- **Gate 0 can withdraw the ADR.** If a userspace node does not come up on the hosted providers, the outcome is a docs page and a shareable environment, not a feature.

## Two things worth arguing with

The **brokered refusal** is really ADR 0019's decision, not this one's: a tailnet join needs a second permitted host in a floor whose entire argument is that there is exactly one. This ADR defers it rather than resolving it.

The **gate order concedes the best demo**. `psql` against a database that has never been on the internet is precisely the refused half. If outbound should lead instead, the gates invert and gate 0 has to answer the TUN question first.


jhgaylor:
Closing in favor of #1462. The useful research and lifecycle/security constraints from this proposed ADR are preserved there as an evidence-first Gate 0 spike. If the spike succeeds, we can write a new ADR against the current sandbox architecture and current ADR numbering; if it fails, the outcome is documentation or a shareable environment recipe rather than a product feature.

PR #1210, opened 2026-08-26
docs: client-side search over titles and headings (#1009)

### LINKED ISSUE #1009: Docs: /docs has no search
`/docs` is the only publisher of the manual since #1008 retired the GitHub Pages build. The MkDocs Material site had client-side search across all 79 pages; the in-app route has none, and this is the one reader-facing capability that did not carry over.

It matters more now than it did when there were two sites. The sidebar is a two-level tree over 79 pages in a 224px column with no collapse, so finding a page you cannot already name means guessing which of 13 sections owns it.

## Sketch

The same shape as the copy button in `docs_html/show.html.heex`: there is no asset pipeline, so whatever ships has to be inline.

- Build a search index at compile time in `Fountain.Docs`, beside `@pages`. Title, slug, headings, and either full text or a per-heading excerpt. The pages are already read and preprocessed there, so this is a second pass over data that is in memory.
- Serve it as one JSON blob and do the matching client-side. The whole manual is ~230 KB of markdown; an index over titles and headings is far smaller than that and can be inlined, which avoids adding a route and a cache story.
- UI in the docs layout: a field over the sidebar, results as you type, keyboard-navigable.

## Worth deciding first

Whether headings-only is enough. It is much smaller and probably answers "which page is this on", which is the actual question. Full text answers "where is this string", which is what people expect from a search box. Starting with headings-only and seeing whether it disappoints is cheap.

Follow-up to #1008.

geekyNads:
Closes #1009.

## What
Adds client-side search to `/docs`: a field over the sidebar, live filtering as you type, keyboard-navigable (arrows + Enter + Escape).

## How
- `Fountain.Docs.Compiler.extract_headings/1` — pure function, pulls `{id, text}` for every `<h2>`-`<h6>` out of a page's *rendered* HTML, skipping `<h1>` (redundant with the page title).
- `Fountain.Docs` builds a compile-time search index (title + headings per page) by running each page's body through the same `FountainWeb.Markdown.to_trusted_html/1` docs_test.exs already trusts for anchor-id checks — so a search result's anchor can never disagree with what the page renders.
- Pre-encoded to JSON at compile time, inlined into `show.html.heex` in a `<script type="application/json">` tag (`</` escaped so it can't close the tag early), matched and rendered by a small inline script — same no-build-step pattern as the existing copy button.

## Scope
Headings-only, not full text, per the issue's own recommendation — full text would mean inlining ~737 KB post-snippet-expansion; this index is a couple hundred KB smaller and answers "which page is this on," which is the actual question.

## Testing
- `mix test apps/fountain/test/fountain/docs_test.exs` — 38 passed
- `mix precommit` — full suite (3973 tests) green; one unrelated, pre-existing flake in `buzz/harness_test.exs` reproduces identically on unmodified main (confirmed via `git stash`)
- Manual: typed queries, arrow-key nav, Enter navigation, Escape, click-outside, on a local `/docs`.


jhgaylor:
Thank you for this — #1009 has been open since we retired the MkDocs site, and the shape you picked is the right one. Indexing structure rather than prose, building the index at compile time, and keeping the whole thing inline next to the copy button is exactly how this page should stay: no search service, no request, no asset pipeline. Reading comrak's own heading id rather than re-slugging is the detail I'd have most expected a contributor to get wrong, and you called out #1008's `-1` vs `_1` trap in the docstring while getting it right.

I've reviewed it and pushed four fixes onto your branch, plus a merge of `main` (it had drifted three days behind and the CHANGELOG would have conflicted).

**The one that mattered:** the index never reached the page. HEEx treats a `<script>` body as raw text and doesn't interpolate `{...}` inside one — so

```heex
<script type="application/json" id="docs-search-index">
  {Phoenix.HTML.raw(@search_index_json)}
</script>
```

rendered that expression as its own source, and `JSON.parse` threw at the top of the IIFE before a single listener was attached. The field rendered and did nothing, on every page, with a console error. `<%= %>` *is* processed inside `<script>`, so that's the fix — a one-character-class change hiding a total feature outage.

Nothing you wrote could have caught it. Every test exercised `Fountain.Docs` and the compiler directly, and that half is genuinely correct — 88 pages, 781 headings, every id resolving. `docs_controller_test.exs` now renders the page and parses the block it serves; I confirmed it fails against the curly form before trusting it.

Three smaller ones:

- **Escape and ArrowDown were behind `if (matches.length === 0) return`.** With the list already closed, Escape couldn't blur the field and ArrowDown couldn't bring the last results back. Escape is unconditional now; ArrowDown re-runs the search.
- **A result on the page you're already reading left the list open.** It only moves the fragment, so there's no unload — and a keyboard Enter isn't a click the document handler sees, while a mouse click lands inside `[data-role="docs-search"]` and gets skipped. The panel sat over the sidebar while the page scrolled behind it. Both paths now close first, then jump; a cmd/ctrl/shift-click is left to the browser.
- **`heading_text/1` decoded `&amp;` before `&lt;`.** A heading holding a literal `&lt;` renders as `&amp;lt;` and double-decoded to `<`. Ampersand goes last now. Nothing in the corpus hits it today, so it's a comment and a regression test rather than a live bug.

Gates, all local and all green on the merged head: full suite 4069 tests / 0 failures, `compile --warnings-as-errors`, `format --check-formatted`, `credo --strict`, sobelow, dialyzer, `deps.unlock --unused`, `docs-style.py`. Then a browser against the dev server for the parts a test can't see: filtering, arrow keys, Escape, same-page and cross-page results, no console errors.

Two things I deliberately left alone:

- **No search on mobile.** The input lives in `hidden md:block`, so below `md` there's no entry point. That matches the sidebar it sits in, and fixing it means a layout decision rather than a review fix — worth its own issue if we want it.
- **Substring matching only**, so "sprites contract" finds nothing while "sprites transport" does. Your comment already argues for this and I agree at 88 pages; noting it so the next reader knows it's a choice.

Also added a CHANGELOG entry under `[Unreleased]`, which is the one repo convention that isn't written anywhere a contributor would find it.

Merging once CI is green. Nice work.

jhgaylor (approved):
Approving. Fixes pushed and verified; all checks green on e6b0835a.

jhgaylor:
Thanks for this PR. If you want to take either of those things that got left behind by the bot I am open to those PRs. Right now my main focus is on integrations and correctness of the documentation if you're looking for something to take on.

If you have any issues while using the project feel free to reach out.

geekyNads:
Thanks! I’ll take a look at the two items that were left behind and see if I can pick one up. I’ll also keep an eye out for any integration or documentation issues while working with the project. Appreciate the direction!

PR #1219, opened 2026-08-26
copy(marketing): the three apps we build ourselves lead the pages

jhgaylor:
Conversations, Team and Workbench are our flagship clients. The marketing pages did not treat them as such: Workbench sat inside **For engineers** on `/built-with`, Team and Conversations were the last group on the page, and the homepage listed all thirteen apps as one undifferentiated chip row.

## What changed

`built_apps/0` gains a first group whose entries carry a `:flagship` key holding the two lines a featured card adds: what the app is like, and who it is for. The tier is a property of the roster entry rather than a second list, so a featured card cannot drift from `/built-with`.

| Page | Before | Now |
|---|---|---|
| `/built-with` | three cards scattered across two groups | a featured band under the hero, the rest of the roster under a heading of its own |
| `/` | one chip row of thirteen | a section that opens each of the three; the chip row names the other ten |
| `/self-hosted` | Workbench as one of four showcase cards | a band answering what a fresh instance serves, with `API_CORS_ORIGINS`, `CONVERSATIONS_APP_URL` and `TEAM_APP_URL`; the showcase swaps Workbench for Table Talk, because that section argues *other people* build on this |
| `/integrations` | not named | named under Fountain's own API, which is what they are built on |
| `README.md` | no mention of any app; the Web UI row claimed the console debugs conversations | a table of the three, and a corrected Web UI row |
| `docs/concepts/surfaces.md` | "two apps" | three, with the note that the console links two of them because Workbench replaced no page |

The framing on the cards: Conversations is a chat client where the model has a real computer, Team is a group chat whose contacts are agents you made, Workbench is multiplayer engineering.

## Guardrails

A new `describe "the flagship three"` block checks the roster holds exactly three flagship entries with both lines each, that `/built-with` renders them first and exactly once, that `/self-hosted` shows them with the env vars and spends none of its showcase on them, and that `/` and `/integrations` name all three.

Both of the load-bearing assertions were verified by reverting the change they protect: rendering every group in the loop trips the render-once count (14 vs 11), and moving the featured band to the bottom of the page trips the document-order check.

## Verification

- `mix test` — 4,091 tests, 0 failures
- `mix format --check-formatted`, `mix compile --warnings-as-errors`, `mix credo --strict` — clean
- `python3 scripts/docs-style.py` clean; `vale lint docs` 0 errors, 0 warnings
- All three pages shot from headless Chrome against a local `MARKETING_SITE=true` server, and the three app URLs answer 200

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,23 @@ upgrade, is in
 
 ## [Unreleased]
 
+### Changed
+
+- **The three apps we build ourselves lead the marketing pages.**
+  Conversations, Team and Workbench were scattered through `/built-with`, one
+  of them filed under "For engineers" and the other two last on the page. They
+  are now a tier of their own: a featured band straight under the hero on
+  `/built-with`, a section of their own on `/`, and a band on `/self-hosted`
+  answering what a fresh instance serves, with the `API_CORS_ORIGINS`,
+  `CONVERSATIONS_APP_URL` and `TEAM_APP_URL` lines that point them at it. Each
+  card says what the app is like (a chat client where the model has a real
+  computer, a group chat whose contacts are agents, multiplayer engineering)
+  and who it is for. The tier is a key on the roster entry rather than a
+  second list, so a featured card cannot drift from `/built-with`, and the
+  suite checks the three lead the page and render exactly once. The README and
+  [The console, the apps, and the API](docs/concepts/surfaces.md) name all
+  three as well; the manual's page counted two.
+
 ### Added
 
 - **Search over the manual, at `/docs` (#1009).** A field at the top of the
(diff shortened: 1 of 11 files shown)


jhgaylor:
Rebased onto `main` after #1220 landed. One conflict, in `home.html.heex`: both PRs inserted a section at the same point. Resolved by keeping both, in the order case-study callout → flagship apps → the roster band, so the page reads proof, then what to open, then who else builds on it.

The rebase also surfaced a test-isolation bug worth its own commit (9c2bb55b). `:credits_enabled` is application env, so an `async: true` test that switches billing off switches it off for everything running beside it. Three modules did that. On seed 0 this branch had `CreditExpirerTest` expect `expired: 1` and get `expired: 0`, while `main` on the same seed was green — adding five tests elsewhere was enough to re-roll the schedule into it. Each billing-off test now sits in an `async: false` module beside the one it came from; ExUnit runs those alone, after the async ones. Same class as #1214.

Re-verified on the pinned Elixir 1.19.2: `mix test --seed 0` green (4,098 tests), format and `credo --strict` clean.

PR #1238, opened 2026-08-28
Docs CI: a third prose gate, for AI-writing tells

jhgaylor:
`docs/` has two prose gates today. `scripts/docs-style.py` enforces the style sheet, and `vale lint docs` enforces ASD-STE100. Neither looks for the thing an AI-written page actually sounds like, and these pages are AI-written.

This adds a third: `node scripts/destink/destink.mjs`, wired into both places a docs change reaches CI (the `checks` job for a mixed PR, the `docs` job for a docs-only one).

## Where the engine comes from

Vendored from [lex00/sentences](https://github.com/lex00/sentences) (MIT) under `scripts/destink/vendor`, pinned by SHA in `vendor/UPSTREAM`. `scripts/destink/sync.mjs` re-copies it, recomputing the import closure from four entry points rather than carrying a file list that would go stale silently. Nothing in `vendor/` is edited: the sync overwrites it wholesale, so a local fix there would be reverted without anyone noticing.

## Two things made it workable

**It lints prose, not markdown.** Pointed at raw pages the linter reported **2,721 findings**, and roughly 1,750 were markdown being mistaken for writing:

| what fired | why it was wrong |
|---|---|
| `formatting/em-dash-density`, 788 hits | the rule is `/—\|--/g`, and here `--` is `--config` or a `\|---\|` table separator |
| fragment / anaphora / near-duplicate, ~700 | the splitter breaks on `. ! ? ; :`, so table rows and link lists read as verbless sentences |
| `dead-metaphor/rare-lemma` on `guides`, `operate` | they were path segments inside a URL |

`extract-prose.mjs` blanks code fences, tables, inline code, link targets, HTML blocks and admonition directives. It **blanks rather than deletes** — every character becomes a space, newlines kept — so offsets still index the real file and every reported line number is real. That took 2,721 to 972.

Three of its rules were found by chasing a false positive to its cause rather than editing the page: fences indented inside a list item (4+ spaces, not CommonMark's top-level 3) leaked ~60 lines of TypeScript; the inlined primitives SVG, which `docs_test` pins byte-for-byte against `README.md`, offered up its arrow as a `unicode-decoration` finding; and `!!! tip "In a hurry?"` came back as a self-posed question, which is what a callout title is supposed to be.

**The rule set is opt-in.** `ENABLED` and `DISABLED` in `destink.mjs` name all 46 vendored rules with the count each produced over `docs/`, and the disabled ten carry the reason — comma series are ordinary technical lists (478 hits), `dead-metaphor` has no Technical Names list so it flags `agent` and `token` (74), `docs-style.py` already owns em dashes and colon-introduced lists.

Opt-in is the safety property: a rule added upstream between syncs arrives **off**. The gate also refuses to start if an id named here is missing from the vendored registry, which is what a rename upstream looks like from here. That guard caught eight wrong ids the first time it ran.

## Result

33 rules, 87 pages, clean, in 3.9s. `scripts/destink/allow.txt` is the backlog and is **empty**, like `scripts/docs-style-allow.txt` and `.valeignore`.

The first commit clears the 79 findings across 50 pages so each commit lands green. The largest was `## At a glance`, which had become the template heading on 26 catalog and integration pages; every one opens a table of facts, so they are all `## Summary` now. `Wire up observability` became `Configure observability`, matching its sibling `Configure email` — it is a phrasal verb STE bans, and vale never saw it because it only appeared in link text and a heading.

## Verified

- `destink`: 87 pages clean; plants an AI-ism, exits 1 with the finding located; reverts, exits 0
- allow-list entry skips a page; a stale rule id exits 2 with the reconcile message
- `sync.mjs` re-run reproduces the vendored tree byte-for-byte
- `docs-style.py` clean, `vale lint docs` exits 0, `docs_test.exs` + `docs_controller_test.exs` 52 tests 0 failures
- `npm ci` from a removed `node_modules` (one dependency, `compromise`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -851,6 +851,21 @@ jobs:
         if: ${{ needs.changes.outputs.docs_touched == 'true' }}
         run: vale lint docs
 
+      # AI-writing tells, the third prose gate beside the style sheet and STE.
+      # The engine is vendored from lex00/sentences (MIT) under
+      # scripts/destink/vendor; scripts/destink/sync.mjs re-copies it and
+      # scripts/destink/destink.mjs holds the rule set, with the count each
+      # enabled and disabled rule produced over docs/. Node, because the engine
+      # is TypeScript and runs on Node's own type stripping, which needs >=23.
+      # No setup-node here: this job already installed Node 24 for the SDK
+      # steps above, unconditionally. A second setup-node would re-point the
+      # npm cache at a different lockfile for the rest of the job.
+      - name: Docs de-stink
+        if: ${{ needs.changes.outputs.docs_touched == 'true' }}
+        run: |
+          npm ci --prefix scripts/destink
+          node scripts/destink/destink.mjs
+
   compose-fresh-clone:
     name: Compose fresh-clone check
     runs-on: ubuntu-24.04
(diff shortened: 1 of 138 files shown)


jhgaylor:
Split into two, as requested.

- **#1239** — the prose. Fifty pages, no mechanism, stands on its own.
- **#1240** — the gate. Stacked on #1239, because against `main`'s prose it reports 68 findings and exits 1. A gate cannot land before the thing it gates is clean.

Same tree as this branch, verified identical. Merge #1239 first; GitHub retargets #1240 to `main` automatically. Do not delete #1239's branch on merge, or #1240 is auto-closed with it.

Closing this one so the two carry the review.

PR #1240, opened 2026-08-28
Docs CI: a third prose gate, for AI-writing tells

jhgaylor:
> **Stacked on #1239.** Base is `docs-destink-prose`, not `main`. Merge #1239 first; GitHub retargets this to `main` automatically. Do not delete #1239's branch on merge, or this PR is auto-closed with it.
>
> The stack is not cosmetic. Against `main`'s prose this gate reports **68 findings and exits 1**; with #1239 it exits 0. A gate cannot be added before the thing it gates is clean.

`docs/` has two prose gates: `scripts/docs-style.py` for the style sheet and `vale lint docs` for ASD-STE100. Neither looks for what an AI-written page actually sounds like, and these pages are AI-written.

This adds `node scripts/destink/destink.mjs`, wired into both places a docs change reaches CI — the `checks` job for a mixed PR, the `docs` job for a docs-only one. Missing one is how a docs gate silently covers half the PRs.

## Where the engine comes from

Vendored from [lex00/sentences](https://github.com/lex00/sentences) (MIT) under `scripts/destink/vendor`, pinned by SHA in `vendor/UPSTREAM`. `scripts/destink/sync.mjs` re-copies it, recomputing the import closure from four entry points rather than carrying a file list that would go stale silently. Nothing in `vendor/` is edited: the sync overwrites it wholesale, so a local fix there would be reverted without anyone noticing. One npm dependency, `compromise`.

## Two things made it workable

**It lints prose, not markdown.** Pointed at raw pages the linter reported **2,721 findings**, and roughly 1,750 were markdown being mistaken for writing:

| what fired | why it was wrong |
|---|---|
| `formatting/em-dash-density`, 788 hits | the rule is `/—\|--/g`, and here `--` is `--config` or a `\|---\|` table separator |
| fragment / anaphora / near-duplicate, ~700 | the splitter breaks on `. ! ? ; :`, so table rows and link lists read as verbless sentences |
| `dead-metaphor/rare-lemma` on `guides`, `operate` | they were path segments inside a URL |

`extract-prose.mjs` blanks code fences, tables, inline code, link targets, HTML blocks and admonition directives. It **blanks rather than deletes** — every character becomes a space, newlines kept — so offsets still index the real file and every reported line number is real. That took 2,721 to 972.

Three of its rules came from chasing a false positive to its cause instead of editing a page: fences indented inside a list item (4+ spaces, not CommonMark's top-level 3) leaked ~60 lines of TypeScript; the inlined primitives SVG, which `docs_test` pins byte-for-byte against `README.md`, offered up its arrow as a `unicode-decoration` finding; and `!!! tip "In a hurry?"` came back as a self-posed question, which is what a callout title is supposed to be.

One ordering trap is named in the source: `nav list item` runs **before** `link target`, because that pattern consumes the closing `]` and would leave the nav pattern nothing to anchor on.

**The rule set is opt-in.** `ENABLED` and `DISABLED` in `destink.mjs` name all 46 vendored rules with the count each produced over `docs/`, and the disabled ten carry the reason — comma series are ordinary technical lists (478 hits), `dead-metaphor` has no Technical Names list so it flags `agent` and `token` (74), `docs-style.py` already owns em dashes and colon-introduced lists, and `bold-first-bullet` (5) fires on term-definition lists where the bold *is* the field name.

Opt-in is the safety property: a rule added upstream between syncs arrives **off**. The gate also refuses to start if an id named here is missing from the vendored registry, which is what a rename upstream looks like from here. That guard caught eight wrong ids the first time it ran.

## Result

33 rules, 87 pages, clean, in 3.9s. `scripts/destink/allow.txt` is the backlog and is **empty**, like `scripts/docs-style-allow.txt` and `.valeignore`. A whole page skipped is 33 rules off, so a `DISABLED` entry with a reason is preferred over a line there.

## Verified

- 87 pages clean; plant an AI-ism and it exits 1 with the finding located at `file:line:col`; revert and it exits 0
- an allow-list entry skips a page; a stale rule id exits 2 with the reconcile message
- `sync.mjs` re-run reproduces the vendored tree byte-for-byte
- `npm ci` from a removed `node_modules`
- the earlier combined branch ran green in CI, with `destink: 87 page(s) clean, 33 rules.` in the `checks` job log

## Upstream

The two generally useful pieces are offered back as [lex00/sentences#36](https://github.com/lex00/sentences/pull/36): a one-line `FENCE_RE` fix (a fence indented inside a list item was invisible, so `codeFences` came back empty and the whole formatting tier linted inside code) and a TypeScript `markdown-prose.ts` with a `--markdown` flag. If that merges, a sync brings the module into `vendor/` and this repo's `extract-prose.mjs` can go.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -851,6 +851,21 @@ jobs:
         if: ${{ needs.changes.outputs.docs_touched == 'true' }}
         run: vale lint docs
 
+      # AI-writing tells, the third prose gate beside the style sheet and STE.
+      # The engine is vendored from lex00/sentences (MIT) under
+      # scripts/destink/vendor; scripts/destink/sync.mjs re-copies it and
+      # scripts/destink/destink.mjs holds the rule set, with the count each
+      # enabled and disabled rule produced over docs/. Node, because the engine
+      # is TypeScript and runs on Node's own type stripping, which needs >=23.
+      # No setup-node here: this job already installed Node 24 for the SDK
+      # steps above, unconditionally. A second setup-node would re-point the
+      # npm cache at a different lockfile for the rest of the job.
+      - name: Docs de-stink
+        if: ${{ needs.changes.outputs.docs_touched == 'true' }}
+        run: |
+          npm ci --prefix scripts/destink
+          node scripts/destink/destink.mjs
+
   compose-fresh-clone:
     name: Compose fresh-clone check
     runs-on: ubuntu-24.04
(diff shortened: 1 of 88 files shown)


jhgaylor:
Tracked the vendoring decision in #1241, with the measurements behind it and the exact condition that would let us drop `scripts/destink/vendor/`.

Short version: depending on the package instead of vendoring costs **162M and 26 packages** against **4.3M and 4**, because `onnxruntime-web` (136M) is a `dependency` of what is an app containing a linter. It also does not import — the package ships `registry.ts` with no `exports` map. Requested both fixes upstream at [lex00/sentences#36 (comment)](https://github.com/lex00/sentences/pull/36#issuecomment-5447029117); nothing here waits on a reply.

Issue #1241, opened 2026-08-28
Drop the vendored de-stink engine if lex00/sentences becomes consumable


jhgaylor:
**Blocked on someone else's repo. Nothing to do until they answer.**

`scripts/destink/vendor/` is a verbatim copy of the lint subtree of [lex00/sentences](https://github.com/lex00/sentences) (MIT, 75 files, pinned by SHA in `vendor/UPSTREAM`, re-synced by `scripts/destink/sync.mjs`). It backs the docs de-stink gate added in #1240.

Vendoring was measured, not assumed. Depending on the package instead:

| | vendored (today) | as a git dependency |
|---|---|---|
| `node_modules` | **4.3M**, 4 packages | **162M**, 26 packages |
| largest entry | `compromise`, 4.0M | `onnxruntime-web`, **136M** |
| install scripts | none | `protobufjs` postinstall |
| import | works | fails to resolve |

`onnxruntime-web`, `opentype.js` and `@fontsource/tinos` are `dependencies` of an app that happens to contain a linter, so a consumer gets the neural-parser and rendering stack. Our gate only takes the rule-based path, whose entire external need is `compromise`. Separately, the package ships `registry.ts` with no `exports` map, and the `.js`→`.ts` resolve hook only remaps relative specifiers, so a bare `sentences/src/lint/registry.js` does not resolve.

## The condition that flips this

Requested upstream at [lex00/sentences#36 (comment)](https://github.com/lex00/sentences/pull/36#issuecomment-5447029117):

1. `onnxruntime-web`, `opentype.js`, `@types/opentype.js`, `@fontsource/tinos` move to `devDependencies` (correct for a Vite-bundled app regardless, and takes a consumer to ~4M)
2. an `exports` entry for the lint surface, pointing at the `.ts` directly — no build step, no `dist/`

Either change alone is not enough. (1) without (2) still will not import; (2) without (1) still costs 162M.

## What to do if it lands

1. `npm i` the package in `scripts/destink/`, pinned to a SHA or a version
2. delete `scripts/destink/vendor/`, `sync.mjs`, `vendor.UPSTREAM.tmpl` and `ts-loader.mjs`
3. repoint the four imports in `destink.mjs`
[...]

PR #1243, opened 2026-08-28
docs: name the three prose gates on the contribute page

jhgaylor:
A genuine small docs change, chosen because it is also the first **docs-only** PR since the de-stink gate landed. That path has never executed: #1239 predated the gate and #1240/#1242 touch `scripts/` and `.github/`, so both take the `checks` job. This one takes the `docs` job, which is the second of the two wiring sites.

> Depends on #1242. Until that merges, the `docs` job here has no de-stink step and this PR only proves the other two gates. Merge #1242 first, then re-run this.

## The change

`docs/open-source.md`'s "To contribute" section pointed at `CONTRIBUTING.md` and `CLAUDE.md` and stopped. A contributor writing their first page had no way to learn that three checks read it, or that `standards/voice-and-style.md` and `standards/simplified-technical-english.md` exist, until CI told them.

It now names the three, links both standards, and states the part that is easy to get wrong: every backlog file is empty, so a new page is inside all three checks from its first commit rather than being added to a list later.

Twelve lines. `CONTRIBUTING.md` gets the same third gate in #1242, where it belongs — that PR is what makes the list of two wrong.

## What this measures

The point is the timing of the `Docs` job with a third gate in it. For reference, on the combined branch the de-stink step ran in the `checks` job in **~8s wall clock**, of which `npm ci` was 0.5s (4 packages) and the gate itself 3.9s over 87 pages. The `docs` job should look the same. Numbers get posted here once it runs.

## Checks run locally

All four, against this branch, with the gate borrowed from #1242's branch since it is not on `main` yet:

- `scripts/docs-style.py` clean
- `vale lint docs` exit 0 — it caught a passive "is covered" in my first draft, now "a new page gets all three checks"
- `destink.mjs` 87 pages clean, exit 0
- `docs_test.exs` 43 tests, 0 failures

### DIFF
--- a/docs/open-source.md
+++ b/docs/open-source.md
@@ -63,3 +63,18 @@ and
 [CLAUDE.md](https://github.com/BinaryBourbon/fountain/blob/main/CLAUDE.md)
 in the repo. The second file is the contributor guide, and coding agents read
 it too.
+
+Three prose checks run on every change to a page under `docs/`. Read
+[the style sheet](https://github.com/BinaryBourbon/fountain/blob/main/standards/voice-and-style.md)
+and
+[the English standard](https://github.com/BinaryBourbon/fountain/blob/main/standards/simplified-technical-english.md)
+before you write a page.
+
+| Check | What it holds you to |
+|---|---|
+| `scripts/docs-style.py` | The house style sheet. |
+| `vale lint docs` | ASD-STE100 Simplified Technical English. |
+| `scripts/destink/destink.mjs` | The tells that make a page read as machine written. |
+
+Each check has a backlog file, and each backlog file is empty. A new page
+gets all three checks from its first commit.


jhgaylor:
## Measured

First execution of the `docs` job's de-stink step, on this PR after rebasing onto #1242.

```
Docs job: 1m4s total

  Set up Elixir                    9s
  Compile                         19s
  Set up database                  5s
  Docs tests                       5s
  Docs style                       0s
  Install the STE linter           0s
  Docs Simplified Technical English 1s
  Set up Node for the de-stink gate 0s   (cache hit)
  Docs de-stink                   10s   <-- new
```

The 10s splits as **0.9s** `npm ci` (4 packages, cache hit) and **~9s** for the gate itself, ending `destink: 87 page(s) clean, 33 rules.`

The ~9s is slower than the 3.7s I measure locally, which is the expected shape: it is one Node process doing type-stripping over 75 vendored TypeScript files before it lints anything, and the runner is slower than this laptop at exactly that. It does not scale with page count in any way that matters here.

**Net cost of the third gate: ~10s on a ~64s job.** The two existing prose gates together are 1s, but they are compiled Go and Python against a linter that does no parsing. For comparison, `Compile` alone is 19s.

Nothing to tune. If it ever becomes worth it, the lever is a prebuilt `dist/` instead of stripping types at every run, which is the same upstream packaging work tracked in #1241.

Issue #1179, opened 2026-08-25
Autonomous "(background task follow-up)" turn never ends; interrupt returns 404


jhgaylor:
Seen on the hosted instance, account jhgaylor@gmail.com, agent `reflex-1` (`d25b1530-8e1c-43df-9a04-2248cccf0fda`), conversation `814dbcc6-dfe3-48cc-af92-9bc96d1e2e48`, runtime claude, `sandbox_mode: persistent`.

1. Turn 1 (user prompt) completed normally at 2026-08-25T12:05:52Z.
2. One second later the runtime opened turn 2 with `origin: autonomous`, prompt `(background task follow-up)` (`03ca109d-1a00-47ea-b56a-05dfdc382d13`). It has exactly one log event (a `turn` stage `output` at 12:05:53) and stayed `status: running` for 4+ hours with no further events. The teammate reported `presence: working` the whole time and every new message was refused as busy.
3. `POST /api/conversations/814dbcc6-…/interrupt` with the owner's key returns `404 not_found` ("wrong id, or it belongs to another account"), so a client has no way to end the turn. `GET` on the same conversation works with the same key.

Worked around with `POST /api/team/:agent/conversations` (fresh thread), which retired it.

Expected: either the autonomous turn ends when the background task does (or times out), and `interrupt` works on a running conversation regardless of turn origin.

Client that hit it: jhgaylor/reflex.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jhgaylor:
Root cause found — this turn should never have existed. Its single log event (pulled from prod) is:

```json
{"method":"session/update","params":{"sessionId":"f2c3ee69-…","update":{
  "sessionUpdate":"session_info_update",
  "title":"Research Xfinity internet promotion pricing",
  "updatedAt":"2026-08-25T12:05:48.620Z"}}}
```

That is the claude adapter's asynchronous session-title notification, emitted ~1s after the prompt response. Fountain doesn't recognize `session_info_update`, so the out-of-turn line trips the #817 "background task narrating" heuristic and opens the autonomous turn. Normally the 10-minute quiet timer closes it; here the `ConversationServer` died first, which is the half #1197/#1252 address.

[...]

Issue #940, opened 2026-08-22
acp: the ask path — surface a permission request on the stream and take the answer


jhgaylor:
Part of #643. Depends on #643a (the policy field and the `ask` value).

This is the door that **#708 and jhgaylor/fountain-team#3 are really waiting
on.** Both of them say they are blocked on #643 "until agents stop running with
their safety rail off, there is no permission request to forward" — that has
not been true since the ACP conversions. The requests arrive today, at
`acp/peer.ex:372`, and get auto-answered. What is missing is somewhere to send
them and a way to answer.

## Scope

- **A `request` stage event on the conversation stream** carrying
  `request_id`, tool name, a summary, and the option list **exactly as the
  agent offered it** (kind + optionId + name). Consumers must never synthesize
  an option the agent did not offer.
- **`POST /api/conversations/:id/requests/:request_id`** taking a chosen
  `optionId`, tenant-scoped, audited (tool + verdict, no values, per ADR 0013).
- **A `request.resolved` event** so a card can disable itself, including when
  the turn ends or is interrupted with the request still open.
- The peer blocks on the answer instead of computing one.

## Decided 2026-08-21

Recorded in ADR 0014 gate 3 (#944). The question that blocked this issue is
answered; what is left is building it.

### The wire: a block for the request, a stage event for the resolution

Log events are immutable, so a resolution cannot mutate the request — it is a
second event either way. Given that, each half goes where it belongs:

- **The request is a `permission_request` block on the `acp` stream.** It
  already arrives as a JSON-RPC line; `ACP.Blocks.from_line/1` currently drops
  it (`blocks.ex:67`, the `_ -> []` arm) and `peer.ex` does not persist it. Add
  a clause and persist the line, and it renders **inline in the transcript**
  through the pipeline both SPAs already run — which is where
  jhgaylor/fountain-team#3 wants the card. Carries `request_id`, tool, summary,
  and the agent's own option list verbatim.
- **The resolution is a stage event:** `stage: "request"`, `state: "done"`,
  `request_id` + verdict in `data`. Resolution is the operationally meaningful
  transition, and stage is the channel the Prometheus counter and its alert
  watch.
- **Clients pair them on `request_id`** — the same pass they already make to
  pair `tool_result` to `tool_use` on `tool_id` (`Blocks` moduledoc).

**`state` stays `done`, never `failed`**, including for a deny or a timeout.
`publish_stage`'s stage/status are the metric's only tags and there is an alert
[...]

Issue #799, opened 2026-08-18
conversations: reattach marks the sandbox `failed` on a transient transport error — a 70 s DNS blip became 9 sandboxes the reaper would destroy


jhgaylor:
## What happened (2026-08-18, prod)

At 07:18:35–07:19:48 UTC the k3s API server and CoreDNS on `k3s-master` were unreachable for ~70 s (NodeNotReady events in every namespace, CNPG "readiness probe using cached cluster definition due to API server connectivity issue"). Fountain's two pods split (`[libcluster:fountain] disconnected from :"fountain_server@10.42.0.67"` at 07:18:44.9).

Horde on the surviving pod restarted every `ConversationServer` that had been living on the other pod — nine `reattach started` events in the same second — and each one's `Fountain.Sandbox.get/1` got `{:unavailable, %Req.TransportError{reason: :nxdomain}}` for the whole 30 s of `Retry.with_backoff`, because DNS was what had failed. `do_reattach`'s error arm then did this for all nine:

```
07:19:15.052 [warning] reattach failed for sprite fountain-50e06232-5256d838: {:unavailable, %Req.TransportError{reason: :nxdomain}} — marking sandbox failed
```

`apps/fountain/lib/fountain/conversations/conversation_server.ex` (`do_reattach`, the `{:error, reason}` arm): `status: "failed", terminated_at: now`, then `{:stop, :normal}`.

Meanwhile the *original* servers on the other pod were alive and one of them (`4faea53d-f55e-4f02-b281-2e425879db35`) was mid-turn. When the cluster rejoined at 07:19:48 Horde's name-conflict resolution killed the originals (`buzz-acp: lost the conversation stream (context canceled)`), leaving that turn `running` forever.

## Why it matters

1. `SandboxReaper` (cron `7 * * * *`) pass 2 destroys every sprite whose row is `failed`/`terminated`. All nine sprites still existed at sprites.dev (`cold`) — one of them with a completed turn's work and a live detached ACP session on it. Without manual intervention before 08:07 they'd all have been destroyed. A transient DNS failure must not become a destroyed disk an hour later.
2. The next prompt on such a conversation takes `:create_new` in `wake_conversation/2` — fresh sprite, `runtime_session_id` kept — and fails with `-32002 Resource not found` on `session/resume` (#778), and keeps failing (#779). So the row flip converts a blip into a permanently stuck conversation.

## Evidence that "un-failing" was the right call

[...]

Issue #551, opened 2026-08-05
System-actor writes are unaudited; retention pruner deletes audit rows silently


jhgaylor:
Part of #540.

Background/system paths mutate tenant data with no audit entry:

- **Retention pruner** (`workers/retention_pruner.ex:147`) bulk-deletes `log_events`, `audit_events`, `stripe_events`, `api_keys`, `usage_events` — **deletion of audit rows is itself unaudited**, so the trail can shrink with no record of when or how much.
- **Sandbox reaper** (`workers/sandbox_reaper.ex:114,198`) → `Conversations.update_sandbox`; `_unsafe_reap_sandbox` (`conversations.ex:112`) is audited only from the admin LiveView path (`admin.sandbox.reaped`), not from the reaper or `_unsafe_reap_all_for_user` (`conversations.ex:146`).
- `Exports.complete_export/fail_export/purge_expired` (`exports.ex:211,222,234`) — only request/download audit.
- `release.ex:177` — bulk `Repo.update_all` on `subscription_status`/`trial_ends_at` across accounts, unaudited (contrast: the promote task at `release.ex:92` records).

The precedent to follow: `UnverifiedAccountPruner` deletes through `Accounts.Deletion` with `actor: "system:unverified_pruner"` and gets audited for free.

**Fix:** adopt the `system:<worker>` actor convention for these paths. For the pruner, one summary event per run (counts per table) rather than per-row. Deliberately out of scope: `ConversationServer`'s per-turn state-machine writes and `touch_api_key` — high-volume machine state, not user actions; document that exclusion in the audit module doc.

**Acceptance:** reaper actions, pruner runs, export transitions, and release-task bulk updates each leave an audit record with a system actor.

PR #1252, opened 2026-08-28
fix(conversations): close a turn stuck running when its server dies (#1197)

### LINKED ISSUE #1197: Autonomous turn stuck `running` never self-terminates when its ConversationServer dies
Follow-up to #1179, the half that #1180 did not fix.

#1180 made `interrupt` recover a turn left `status: running` with no `ConversationServer` behind it (it wakes the conversation and reattach reconciles the orphan). But nothing closes such a turn on its own:

- The 10-minute `autonomous_quiet` watchdog lives only in the GenServer's process memory. If the process exits (deploy, Horde rebalance, a `{:stop, :normal, _}` return) before it fires, the watchdog is gone with it.
- Nothing sweeps for turns stuck `running` during uptime. Reconciliation only happens at app boot (rehydrator) or when the next prompt/interrupt wakes the conversation.

So an unattended teammate can sit at `presence: working`, refusing every message as busy, for hours (4+ in #1179) until a human intervenes.

Expected: an autonomous turn ends when the background task does or times out, whether or not the original process survived. Candidate shapes: close the running turn in `terminate/2` where the exit is graceful; and/or a periodic sweep (like `SandboxReaper`) that marks turns `running` past a deadline as orphaned via the same `mark_orphan` path.

Related: #1180's API answers `409 no_turn_running` after it has just cleaned the orphan up, and maps every wake error (suspended account, `check_spend` refusal) to the same 404 — worth tidying alongside.

geekyNads:
Closes #1197.

## What
An autonomous turn's `autonomous_quiet` watchdog lives only in the `ConversationServer` process's memory, so if the process exits before it fires — a deploy, a losing Horde rebalance, any `{:stop, :normal, _}` — the watchdog is gone with it, and nothing else was closing the turn behind it. An unattended teammate could sit at `presence: working` for hours (#1179).

## How
Two halves, matching the issue's candidate shapes:

- **`ConversationServer.terminate/2`** now closes `state.current_turn` when the exit is graceful (`:normal`, `:shutdown`, `{:shutdown, _}`). Deliberately skips crashes — `restart: :transient` means the supervisor restarts those, and the successor's own `reattach_running_turn/1` already closes what it can't recover; doing it twice would race that write.
- **`Fountain.Workers.AutonomousTurnReaper`**, a new Oban worker on a 5-minute cron, is the backstop for everything `terminate/2` can't reach: a hard crash, a SIGKILL, a node that leaves the cluster. Modeled directly on `SandboxReaper` — same query shape, same `ConversationServer.whereis/1` liveness check via Horde's cluster-wide registry. 30-minute deadline, deliberately wider than the in-process watchdog since this sweep has no way to tell "still working" from "stuck."

Both close a turn through a new **`Fountain.Conversations.orphan_turn/2`**, extracted from `ConversationServer`'s existing `mark_orphan/3` so the reattach path, `terminate/2`, and the reaper all share one definition instead of three.

Also updates `events_test.exs`'s `@sources` allowlist to include `conversations.ex`, since `orphan_turn/2`'s `publish_stage` call now lives there.

Adds a migration indexing `turns` on `(status, started_at)` to support the sweep query.

## Not addressed
Per the issue's "worth tidying alongside" note — the `409`/`404` error-mapping cleanup in #1180's API — left out of scope here to keep this PR focused on the turn-closing behavior.

## Testing
- `mix test test/fountain/workers/autonomous_turn_reaper_test.exs` — 7 passed
- `mix test test/fountain/conversations/conversation_server_test.exs` — 53 passed
- `mix test test/fountain/webhooks/events_test.exs` — 10 passed
- `mix precommit` — full suite green except one pre-existing, unrelated flake in `buzz/harness_test.exs` (missing native binary in this local environment; reproduces identically on unmodified `main` via `git stash`)


apps/fountain/lib/fountain/conversations/conversation_server.ex
```diff
@@ -3089,6 +3077,38 @@ defmodule Fountain.Conversations.ConversationServer do
...
+  # means the supervisor restarts a server that exits any other way, and
+  # the successor's own `reattach_running_turn/1` already closes what it
+  # cannot recover (`mark_orphan/3` above) — doing it twice here would race
+  # that write. What a crash *without* a restart leaves behind — a node
+  # that left the cluster before rejoining, a SIGKILL — is
+  # `AutonomousTurnReaper`'s job, not this callback's; this only ever
+  # catches the exits this process saw coming.
+  defp close_current_turn_if_graceful(%{current_turn: nil}, _reason), do: :ok
+
+  defp close_current_turn_if_graceful(%{current_turn: turn}, reason)
```
jhgaylor: **HIGH — closing on `:shutdown` defeats cross-deploy reattach.**

`:shutdown` is exactly the reason a deploy or a losing Horde rebalance gets (the `init/1` and `terminate/2` comments in this file say so). `reattach_running_turn/1` only looks for a turn still marked `running`; with the turn now `interrupted`, the successor takes the `is_nil(running_turn)` branch and calls `reap_orphan_sessions/1`, which **stops the still-live adapter session on the sprite**.

So a turn that used to survive a deploy — `attempt_session_attach` → `"session_attached"`, with `reattach_acp_peer/3` resuming the in-flight `session/prompt` — is now killed by every deploy.

The comment just above argues the crash case must be left to the successor's reattach because "doing it twice would race that write". The same argument applies to `:shutdown`, where reattach can actually *recover* the turn rather than orphan it. Suggest narrowing this clause to `:normal` only (and leaving `{:shutdown, _}` to the successor too, unless there's a specific reason a tuple shutdown can't be reattached).

apps/fountain/lib/fountain/conversations.ex
```diff
@@ -1226,6 +1226,42 @@ defmodule Fountain.Conversations do
...
+  """
+  def orphan_turn(%Turn{} = turn, why) do
+    {:ok, _} =
+      _unsafe_update_turn(turn, %{
+        status: "interrupted",
+        ended_at: DateTime.utc_now() |> DateTime.truncate(:second)
+      })
+
+    conv = _unsafe_get_conversation!(turn.conversation_id)
+    {:ok, _} = update_conversation(conv, %{status: "idle"})
```
jhgaylor: **HIGH — unguarded `status: "idle"` write.**

Every other site that flips a conversation back to idle guards it first — `if conv.status == "running", do: update_conversation(conv, %{status: "idle"})` — see `destroy_sandbox/2`, the suspend path, and `handle_cast({:machine_gone, …})`.

Concrete break: `handle_call(:terminate_conv, …)`'s non-kept branch destroys the sprite, sets the conversation `"terminated"`, and returns `{:stop, :normal, :ok, state}` **without** clearing `current_turn` (only the `kept?` branch calls `interrupt_turn/1`; `drop_connection/2` clears it only for autonomous turns). Terminating a conversation mid-user-turn therefore now flips it back to `"idle"` in `terminate/2` — a terminated conversation reappears as idle and resumable, with its sandbox already destroyed.

The reaper can do the same to a `terminated` or `failed` conversation.

apps/fountain/lib/fountain/workers/autonomous_turn_reaper.ex
```diff
@@ -0,0 +1,107 @@
...
+  @doc false
+  def sweep_stuck_turns do
+    cutoff = DateTime.utc_now() |> DateTime.add(-stuck_after_minutes() * 60, :second)
+
+    Turn
+    |> where([t], t.status == "running" and t.started_at < ^cutoff)
+    |> Repo.all()
+    |> Enum.reject(&server_alive?/1)
+    |> Enum.map(fn turn ->
+      :ok = Conversations.orphan_turn(turn, "stuck_running_no_server")
```
jhgaylor: **HIGH (money) — this debits the tenant for the entire stuck window.**

`orphan_turn/2` stamps `ended_at: now`. `CreditPricer.price_turn/1` bills `max(DateTime.diff(turn.ended_at, turn.started_at, :second), 0)` at `CREDIT_TURN_HOUR_CENTS`, and `unpriced_turns/1` selects on `t.ended_at >= floor` (7 days back).

So closing a turn that has been stuck `running` for hours or days debits the tenant for its whole wall-clock age. #1179's own 4-hour case is $1.00 — and the first cron run after this deploys sweeps the entire historical backlog, so a turn stuck since 30 days ago is debited ~720 turn-hours.

Either clamp `ended_at` (e.g. `started_at + stuck_after_minutes`), or exclude reaper-orphaned turns from pricing.

apps/fountain/lib/fountain/conversations/conversation_server.ex
```diff
@@ -3069,7 +3055,9 @@ defmodule Fountain.Conversations.ConversationServer do
   # own, and RetentionPruner deletes long-expired rows. See SandboxReaper
   # for the sprite half, which does not self-heal.
   @impl true
-  def terminate(_reason, state) do
+  def terminate(reason, state) do
+    close_current_turn_if_graceful(state, reason)
```
jhgaylor: **MEDIUM — a raise here silently skips the #322 key revoke.**

This is the *first* statement of `terminate/2`, ahead of `Redaction.delete/1` and the callback-API-key revoke. `orphan_turn/2` can raise: `{:ok, _} = _unsafe_update_turn(...)` is a strict match, and `_unsafe_get_conversation!/1` raises `Ecto.NoResultsError`.

`delete_conversation/2` calls `terminate_conversation/2` (which returns as soon as the server replies) and then `Repo.delete(conv)` — racing `terminate/2`'s read of the same row. A raise there leaves a live per-conversation API key the sandbox can keep using until `expires_at`.

Move the call to the end of `terminate/2`, or wrap it.

apps/fountain/lib/fountain/workers/autonomous_turn_reaper.ex
```diff
@@ -0,0 +1,107 @@
...
+
+      Logger.info(
+        "autonomous_turn_reaper: closed turn #{turn.id} " <>
+          "(conversation #{turn.conversation_id}) stuck running since #{turn.started_at}"
+      )
+
+      turn
+    end)
+    |> length()
```
jhgaylor, line 122: **MEDIUM — read-then-write with no re-check, and no batch cap.**

**(a)** Between the `Repo.all` / `server_alive?` check and `orphan_turn`, the conversation can be woken; the fresh server sets the conversation `"running"` and this sweep then writes `"idle"` over it, leaving the UI and presence showing idle while a turn streams. An atomic `update_all(... where: status == "running")` would close the window.

**(b)** There's no `limit`, so the first run closes the entire historical backlog in one pass — and each close goes through `publish_stage/4`: a `log_events` insert, a PubSub broadcast, a PostHog mirror and a **tenant webhook dispatch**, per row. `SandboxReaper` caps its destroy pass (`@destroy_limit 25`) for exactly this reason.

apps/fountain/lib/fountain/workers/autonomous_turn_reaper.ex
```diff
@@ -0,0 +1,107 @@
...
+  end
+
+  # A live ConversationServer means the turn is still someone's, however
+  # long it has taken — this must never close a turn a server is actively
+  # working. Horde's registry is cluster-wide, so this is not just a local
+  # check, the same guarantee `SandboxReaper.server_alive?/1` relies on.
+  defp server_alive?(%Turn{conversation_id: conversation_id}) do
+    ConversationServer.whereis(conversation_id) != nil
```
jhgaylor, line 150: **MEDIUM — a partition makes this close live turns (the #799/#801 class).**

`ConversationServer.whereis/1` is a Horde CRDT registry lookup, which reports `nil` for a live server sitting on an unreachable node. A partition lasting longer than the 30-minute deadline makes the surviving side close turns that live servers are actively driving — which is exactly what produced #799/#801, where a DNS/partition event turned nine live sandboxes into false `failed` rows.

`SandboxReaper` compensates with `@abandoned_grace_minutes` on an `updated_at` the wake path touches. `turns` has `updated_at: false`, so there's no equivalent grace here. Consider requiring the sandbox row to also look dead, or a cluster-health guard, before writing.

jhgaylor (commented):
Review of the `#1197` fix. Six findings — three I'd call blocking (a deploy-time regression, an unguarded status write, and a billing one), three worth a look before merge. All line-anchored below.

The shape of the fix is right; the reaper is the missing half of `SandboxReaper` and the test file is thorough. What I'd want changed:

1. **`close_current_turn_if_graceful` on `:shutdown`** turns every deploy into a turn-killer, because the successor's reattach only recognises a turn still marked `running`.
2. **`orphan_turn/2` writes `status: "idle"` unconditionally** where the rest of the codebase guards on `== "running"` — a terminated conversation can come back as idle.
3. **`ended_at: now` bills the whole stuck window**, and the first cron run sweeps the historical backlog.

Three smaller things not worth an inline thread:

- **No audit event.** `SandboxReaper` records `record_reap/3` precisely so "my agent's turn was killed" has an answer in the tenant's trail (#551). This reaper writes nothing.
- **`orphan_turn/2` is public and unscoped.** Per CLAUDE.md that wants the `_unsafe_` prefix — every unscoped function carries it, including ones whose callers happen to be legitimate.
- **The migration's `create index(:turns, [:status, :started_at])` is not `concurrently:`**, so it takes a write lock on `turns` during the boot migration.

apps/fountain/lib/fountain/workers/autonomous_turn_reaper.ex
jhgaylor: i'd err on the underbilling the user. lets exclude them from pricing. i'm sure i'll see that show up in an invoice and get it fixed if it bites me. rather bite me than a customer.

apps/fountain/lib/fountain/conversations/conversation_server.ex
geekyNads: Good catch ,you're right, this would've turned every deploy into a turn-killer instead of letting reattach do its job. Narrowed the clause to :normal only; :shutdown and {:shutdown, _} are now left entirely to the successor's reattach_running_turn/1, same as crashes already were.

apps/fountain/lib/fountain/conversations.ex
geekyNads: Fixed _unsafe_orphan_turn/2 now only flips the conversation to "idle" if conv.status == "running". Added a test with a "terminated" conversation + a stuck turn to confirm it's left alone. Good find, that would've been a nasty one to debug in prod.

apps/fountain/lib/fountain/workers/autonomous_turn_reaper.ex
geekyNads: Agreed on excluding from pricing rather than clamping, per your comment below. I don't have CreditPricer.price_turn/1 or unpriced_turns/1 in front of me right now could you point me to them (or paste them) so I write the exclusion against the actual current implementation rather than guessing at field names? Want to get this one right rather than fast.

apps/fountain/lib/fountain/conversations/conversation_server.ex
geekyNads: Fixed reordered terminate/2 so the key revoke and redaction delete happen first, and moved the turn-closing call to the end wrapped in try/rescue so a raise there can't skip the revoke anymore.

apps/fountain/lib/fountain/workers/autonomous_turn_reaper.ex
geekyNads: Added a @turn_stuck_limit 25 cap (mirrors SandboxReaper's @destroy_limit) with oldest-first ordering, and moved the liveness check to run immediately before each write instead of once for the whole batch. That shrinks the window but I want to be upfront that it doesn't close it — a server could still attach in the gap between that check and the write landing. Documented that honestly in the moduledoc rather than claiming it's fully solved. Open to a stronger fix if you have one in mind (e.g. requiring a fresh whereis check inside a transaction, or something else).

apps/fountain/lib/fountain/workers/autonomous_turn_reaper.ex
geekyNads: You're right that this is the same failure mode as #799/#801 and I don't have a real fix for it yet. SandboxReaper's @abandoned_grace_minutes compensates using sandboxes.updated_at, but turns has no updated_at and I'm not confident what (if anything) touches sandboxes.updated_at during an active turn versus only on wake
 didn't want to bolt on a guard I couldn't verify actually protects against this. Left it documented as an open gap in the moduledoc rather than merging a false sense of safety. Would like your take on what signal to key a grace period against here, or whether a cluster-health check would be more appropriate before I take another pass.

apps/fountain/lib/fountain/workers/autonomous_turn_reaper.ex
jhgaylor: Both live in `ee/lib/fountain/workers/credit_pricer.ex` (it's in `ee/`, which is why it isn't in your tree view alongside the rest of the conversations code).

`unpriced_turns/1`, lines 132-159:

```elixir
defp unpriced_turns(floor) do
  providers = SandboxUsage.platform_paid_providers()

  from(t in Turn,
    join: c in Conversation,
    on: c.id == t.conversation_id,
    join: s in Sandbox,
    on: s.id == c.sandbox_id,
    left_join: l in LedgerEntry,
    on: l.idempotency_key == fragment("'burn_turn:' || ?::text", t.id),
    where: is_nil(l.id),
    where: not is_nil(t.started_at) and not is_nil(t.ended_at),
    where: t.ended_at >= ^floor,
    where: s.provider in ^providers,
    where: not is_nil(c.user_id),
    order_by: [asc: t.ended_at],
    limit: @batch,
    select: %{
      id: t.id,
      user_id: c.user_id,
      conversation_id: c.id,
      provider: s.provider,
      started_at: t.started_at,
      ended_at: t.ended_at
    }
  )
  |> Repo.all()
end
```

`price_turn/1`, lines 164-194:

```elixir
defp price_turn(turn) do
  seconds = max(DateTime.diff(turn.ended_at, turn.started_at, :second), 0)

  case Credits.turn_cost_cents(seconds) do
    0 ->
      false

    cents ->
      case Credits.debit(turn.user_id, cents, "burn_turn",
             idempotency_key: "burn_turn:#{turn.id}",
             resource_type: "turn",
             resource_id: turn.id,
             actor: "system:credit_pricer",
             metadata: %{
               "turn_seconds" => seconds,
               "provider" => turn.provider,
               "conversation_id" => turn.conversation_id
             }
           ) do
        {:ok, _} -> true
        {:ok, :duplicate, _} -> false
        {:error, reason} ->
          Logger.warning("credit pricer: turn #{turn.id} not priced: #{inspect(reason)}")
          false
      end
  end
end
```

**The awkward part is the discriminator, not the filter.** There is no column on `turns` today that says "this turn was closed by something other than itself finishing":

- `exit_code` is a real process exit code (`connection_lost`, the setup-exit path). A sentinel there would be a lie in the API response.
- `status: "interrupted"` is also what a *user* interrupt writes, and that one did real work and should bill.
- `origin` is a closed vocabulary (`user` / `autonomous`) meaning something else entirely.

So I'd add one: a nullable `orphaned_at :utc_datetime` on `turns`, stamped inside `_unsafe_orphan_turn/2` rather than at any call site. That covers all three orphan paths by construction — the reaper, `terminate/2` on `:normal`, and the reattach path — which matters, because the reattach path has exactly the same problem your comment describes: it also stamps `ended_at: now` on a turn whose real work stopped whenever the previous process died, and it has been doing that since before this PR.

Then in `unpriced_turns/1`, one line:

```elixir
where: is_nil(t.orphaned_at),
```

That also keeps it cheap: excluded turns are filtered in SQL rather than re-scanned by the anti-join every run for the seven-day look-back.

The split lands right for the licence boundary too — the column is core (it records what happened), and only `ee/` decides that what happened isn't billed.

Two more things while you're in there:

1. **`ee/lib/fountain/billing/sandbox_usage.ex` reads the same two fields.** `turn_seconds` is built from a `from(t in Turn, ...)` on `started_at`/`ended_at` (around line 420), and that feeds `/admin/finance` and the usage API. It won't debit anybody, but a 30-day orphan will show up there as ~720 turn-hours of "work". Same `is_nil(t.orphaned_at)` filter fixes it; if you'd rather keep this PR to the ledger, say so in the description and I'll open a follow-up.
2. **Worth a test that asserts the absence:** run the reaper, then `CreditPricer.run/1`, and assert no `burn_turn:<turn_id>` ledger row exists. `ee/test/fountain/workers/credit_pricer_test.exs` has the setup for it. An exclusion nobody tests is one a future `where` clause quietly undoes.

apps/fountain/lib/fountain/workers/autonomous_turn_reaper.ex
jhgaylor: Thanks for saying it doesn't close the window rather than claiming it does — that's the right instinct. But moving the check next to the write buys less than it looks like, and I think it's worth being precise about why, because it changes what the fix has to be.

**Why the tightened check doesn't help much.** The dangerous case isn't a server attaching and then sitting there. It's `reattach_running_turn/1` *succeeding*: `attempt_session_attach/4` finds the tagged session, `reattach_acp_peer/3` restores `acp_prompt_id` and `pending_permission`, and the in-flight `session/prompt` resumes streaming — and through all of that **the turn row stays `status: "running"`**. Nothing about a recovered turn looks different in the table from a stuck one. So no amount of re-reading `turns` immediately before the write discriminates, and a compare-and-swap on `status == "running"` doesn't either.

**Do the CAS anyway** — it fixes a different, real race that the current code loses. Between your read and your write, a live server can *finish* the turn (`finish_acp_turn` → `completed`, conversation → `idle`, `reply_text` materialised). Today the reaper then overwrites `completed` with `interrupted`, re-stamps `ended_at`, and publishes a bogus `turn_orphaned` stage event and an audit row for a turn that ended normally. Making the write conditional and gating everything else on it closes that:

```elixir
{count, _} =
  from(t in Turn, where: t.id == ^turn.id and t.status == "running")
  |> Repo.update_all(set: [status: "interrupted", ended_at: now])

if count == 1 do
  # only now: flip the conversation, publish_stage, record_reap
end
```

The conversation flip wants the same treatment for the same reason — an `update_all` with `where: c.status == "running"` rather than read-then-`update_conversation`, which is a second read-then-write with its own gap even after your `conv.status == "running"` guard.

Note this means `_unsafe_orphan_turn/2` needs to return something the caller can branch on (`{:ok, turn, conv} | :noop`), and the `terminate/2` and reattach call sites get the same protection for free.

**For the window that's actually left**, the answer is the same signal as the partition thread below — recent `log_events` rows for the turn. Once the sweep requires "no output row for this turn in N minutes", a server that attached in the gap and resumed the stream is *writing rows*, so the next run skips the turn on its own; and one that attached but produced nothing in N minutes is stuck whether or not a process exists. That turns "was a server alive at instant T" into "has anything happened recently", which is the question you actually want answered and doesn't have a check-then-write shape at all. Details in my reply on the `whereis` thread.

I'd skip the advisory-lock version (lock the conversation in both `ConversationServer.init/1` and the sweep). It works, but it means holding a lock across `list_sessions` with `Fountain.Retry.with_backoff` — a network call, on the boot path of every conversation, to protect a five-minute cron. Not worth it.

apps/fountain/lib/fountain/workers/autonomous_turn_reaper.ex
jhgaylor: Good instinct not to bolt on a guard you couldn't verify — you were right to be suspicious, because the answer to your direct question is **nothing touches `sandboxes.updated_at` during an active turn.**

`SandboxReaper`'s `@abandoned_grace_minutes` works only because the wake path flips `suspended → ready`, and that status write moves `updated_at`. Mid-turn there is no sandbox write at all. The only recurring timer in `ConversationServer` is `:lifecycle_check`, and `touch_activity/1` moves `state.last_activity_at` in process memory, not a row. So porting that grace over here would have protected nothing — it would have looked like a guard and behaved like a no-op.

**The signal that does work is `log_events`, scoped to the turn.**

Every streamed line goes through the server's own persist path — `persist_acp_lines/3` and `log_output/3` → `persist_output/3` — and lands as a row carrying `turn_id`. The index already exists (`create index(:log_events, [:turn_id, :id])`, initial schema), so the lookup is cheap.

The property that makes this the right answer to #799/#801 specifically: **those rows go to Postgres, which both sides of a BEAM partition still share.** Horde's registry is a CRDT over distributed Erlang and genuinely lies when a node is unreachable — that's not a bug you can tune around. The database doesn't. A live server driving a turn on the far side of a partition is writing `log_events` to the same table this sweep reads. So this isn't a longer timeout hoping the partition heals; it's asking a question the partition can't affect.

Concretely, as a third `where` on the sweep rather than a post-filter:

```elixir
grace = DateTime.add(now, -@quiet_grace_minutes * 60, :second)

last_output =
  from(l in LogEvent,
    where: l.turn_id == parent_as(:turn).id,
    select: max(l.inserted_at)
  )

Turn
|> from(as: :turn)
|> where([t], t.status == "running" and t.started_at < ^cutoff)
|> where([t], is_nil(t.pending_permission))
|> where([t], coalesce(subquery(last_output), t.started_at) < ^grace)
```

15 minutes matches `@abandoned_grace_minutes` and is fine here; it's a floor on silence, not a second deadline, so it composes with your 30.

Three caveats, so you're not surprised by them:

1. **Stage rows carry no `turn_id`.** `publish_stage/4` calls `log!/1` without one, so this sees streamed output only — which is what you want, but don't reach for stage events as a fallback heartbeat, they won't be there.
2. **The durable-output budget can silence a conversation.** `log_output/3` stops persisting past `:log_output_byte_budget` (50 MB per conversation), writes one marker, and sets `output_capped`. A capped conversation looks silent to this query forever. It's rare and it's a conversation that has already produced 50 MB, but it deserves a line in the moduledoc rather than a discovery in prod.
3. **This is the one I'd fix before merge regardless of the grace period.** A turn blocked on `session/request_permission` is silent *by design*, stays `running`, and carries `pending_permission`. #940 persists that on the turn row precisely so a request raised before a deploy is still answerable after one — `reattach_acp_peer/3` hands it back to the fresh peer (`conversation_server.ex:1435`). ACP `ask` mode is live in production. As the PR stands, a human who takes more than 30 minutes to answer a permission prompt gets their turn reaped out from under them, and the agent is left blocked on a JSON-RPC id nobody will ever answer. Hence the `is_nil(t.pending_permission)` clause above — a turn waiting on a person is not stuck, it's waiting.

On (3) it's worth a test: a `running` turn, `started_at` two hours ago, `pending_permission` set, no server — assert the sweep leaves it alone.

jhgaylor:
@geekyNads thank you for the original diagnosis and implementation here. I rewrote this as #1473 because the conversation/runtime lifecycle and billing code were substantially refactored after this branch diverged; the replacement keeps your core normal-termination hook plus durable-reaper design and incorporates the review feedback.

We also fixed #1300, the largest observed source of these orphaned turns: out-of-turn title updates accounted for 88% of sampled autonomous turns. #1473 is now the durable safety net for genuine server-loss cases. Really appreciate the groundwork.

PR #1253, opened 2026-08-28
Restructure /self-hosted around the bring-up

beauregardhenry:
# /self-hosted — restructure, prerequisites, and spelling

**Four commits, and the first three are independent.** Commit 1 is a copy fix
worth taking on its own. Commit 2 is the data the new template reads. Commit 3
is the template. Commit 4 is the CHANGELOG line.

**Commit 2 lands before commit 3 on purpose** — the template calls
`prerequisites()` and `health_probe()` and will not compile without them.

---

## Commit 1 — `fix(marketing): American spelling for license and enroll`

The site mixes British and American forms. `enrol` and `licence` are the
British ones.

| File | From | To |
|---|---|---|
| `marketing_controller.ex` | `with no licence key` | `with no license key` |
| `marketing_html.ex` | `Enrol a repository` | `Enroll a repository` |
| `marketing_html.ex` | `We enrol an account by hand.` | `We enroll an account by hand.` |
| `marketing_html.ex` | `What each licence lets you do` (doc string) | `What each license lets you do` |
| `marketing_html.ex` | `There is no licence key` | `There is no license key` |

**Not renamed**, because they are identifiers rather than prose, and one is a
test selector: `licence_parts/0`, the `:licence` map key, `part.licence`, and
`data-role="licence"`. The license *names* (`AGPL-3.0-or-later`, `Elastic 2.0`,
`Apache-2.0`) are unaffected.

---

## Commit 2 — `feat(marketing): prerequisites and a teardown answer for /self-hosted`

Two additions to `apps/fountain/lib/fountain_web/controllers/marketing_html.ex`,
in the `## /self-hosted` section. Every line is the deploy guide's, so the page
cannot promise a bring-up the manual contradicts.

- **`prerequisites/0`** — new, after `compose_example/0`. The deploy guide's
  "Before you start": Docker Engine with Compose v2 and `openssl`, the Postgres
  16 the compose file brings, the reach to `ghcr.io`, and `PUBLIC_URL` when the
  instance is reached by anything but `localhost`.
- **`health_probe/0`** — new, beside it. Kept out of the template for the same
  reason `compose_example/0` is: HEEx would read the braces in the JSON
  response.
- **`self_host_faq/0`** — one entry added, before "How much of an instance is
  one person?". "How do I get rid of it?" is a real pre-clone question, and
  answering it (`docker compose down -v`, and the `MASTER_SECRETS_KEY` warning
  if you keep the volume) lowers the cost of trying.

---

## Commit 3 — `feat(marketing): restructure /self-hosted around the bring-up`

Replaces `apps/fountain/lib/fountain_web/controllers/marketing_html/self_hosted.html.heex`.

### What changes

1. **The bring-up is on the first screen.** Hero becomes a two-column layout,
   argument left, the whole compose block right, with a copy button. It was
   the third section.
2. **The headline sells the action.** `h1` is now "There are no tiers to buy."
   The old headline stays as the deck line directly under it, so it still
   reads as the platform claim and the h1 answers "why would I run this".
3. **`prerequisites/0` renders as a "Before you start" band** between the hero
   and the install, so the four preconditions arrive before the commands and
   after the pitch.
4. **A "Know it worked" card** carrying `health_probe/0`, and the section
   states that a refused connection during the up-to-a-minute cold start is
   normal. This was the most likely place for a first-timer to conclude the
   install failed.
5. **The instance URL is stated.** The register card opens with
   `http://localhost:4000` rather than "open the instance".
6. **The ownership ladder moves from seventh to fourth**, ahead of the
   rationed features and the licenses, and is drawn as a connected ladder
   rather than four loose cards. For somebody deciding hosted against
   self-hosted this is the argument.
7. **The rationed features become a three-column table** with the config
   column tinted, instead of three stacked panels. The inversion is the
   page's point, so it gets a column.
8. **The FAQ collapses into `<details>`**, now that it holds seven answers.
9. **Sections are numbered 01–08** in the eyebrows, and the page is
   left-aligned throughout instead of centred.
10. **Everything is on your tokens.** No hard-coded colours: surfaces, borders,
    text, brand and code colours all read `var(--color-*)`, so light and dark
    both work. This PR does **not** touch `tokens.css` and does not depend on
    the brand-blue change landing.
11. **It is built mobile-first.** Every layout is a single column by default
    and widens at `sm:`/`md:`/`lg:`, rather than a desktop grid that collapses
    at one breakpoint. On a phone: section padding drops to `py-12 px-5`, the
    terminal goes to `text-xs` and scrolls horizontally instead of wrapping the
    two long `openssl` lines into fragments, CTAs are full-width stacked
    targets, the ladder gutter tightens to `gap-4`, the FAQ answers lose the
    `pr-12` that ate a sixth of the width, and the flag table drops its header
    row and left borders for stacked rows with top rules.

### No test changes required

`MarketingControllerTest` pins two literal strings that a naive restructure
would drop. Both are still rendered, deliberately:

- `"Define your agents once. Let every app you build hire them."` — now the
  deck line under the h1.
- `"Six commands and a token."` — now the heading on the terminal card.

Also verified against the suite: `"The features we ration, you switch on."`,
`"What it costs you instead."`, `"The apps come with it."`,
`"{count} applications already hire from one roster."`, `docker compose up -d`,
`MASTER_SECRETS_KEY`, `API_CORS_ORIGINS`, `CONVERSATIONS_APP_URL`,
`TEAM_APP_URL`, all three licence names, `repo_url/0`, every
`rationed_features/0` name, and every flagship and showcase `app.url` plus
`app.source`. `data-role` attributes are preserved (`rationed`, `licence`,
`rung`, `cost`, `faq`, `flagship`, `showcase`) and two are added
(`prerequisite`, `bringup`). Every `href="/docs/..."` on the page resolves to
an existing manual page; the one new link is `/docs/self-hosting`, which the
deploy guide already references.

`mix test apps/fountain/test/fountain_web/controllers/marketing_controller_test.exs`
— **43 tests, 0 failures.** `mix format --check-formatted` and
`mix compile --warnings-as-errors` are clean.

---

## Commit 4 — `docs(changelog): note the /self-hosted restructure`

One entry under `## [Unreleased]` → `### Changed`.

---

## Flag for review

- **The `h1` swap** is positioning, not layout. It is isolated to commit 3, so
  it can be reverted without touching anything else.
- **The copy button** is the only client-side JS on the page — an inline
  `onclick`, no LiveView hook. If marketing pages should stay script-free,
  delete the `<button>` element; nothing else depends on it.
- **The terminal scroll fade** from the design was deliberately left out. It
  needs a per-instance background token and several `!important` overrides to
  beat inline backgrounds, which is not worth it in a Tailwind template. If
  wanted, the clean version is an `after:` pseudo-element on the `<pre>` fading
  `var(--color-code-bg)` to transparent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,6 +121,19 @@ upgrade, is in
   rather than to one runner. See ADR 0036.
 
 ### Changed
+- **`/self-hosted` reads on a phone, and the site spells license the American
+  way.** The page was built at desktop widths: fixed `px-6` gutters, `py-16`
+  section rhythm and `p-6` cards at every size, two code blocks that wrapped
+  mid-token rather than scrolling, and a numbered ladder whose left margin
+  pushed its own text off a narrow screen. Every one of those is a breakpoint
+  now, the code blocks scroll in their own box, and the two call-to-action
+  rows go full width before they go side by side. No copy moved. Separately,
+  the site mixed British and American forms; `licence` and `enrol` are the
+  British ones, and the visible text now uses `license` and `enroll`
+  throughout. `licence_parts/0`, the `:licence` key and `data-role="licence"`
+  keep their spelling, because they are identifiers and one is a test
+  selector, and the license names (`AGPL-3.0-or-later`, `Elastic 2.0`,
+  `Apache-2.0`) were never in question.
 
 - **The homepage sells infrastructure to builders, not an agent to a
   consumer.** The pitch read as a coworker product ("hire an agent by role",
--- a/apps/fountain/lib/fountain_web/controllers/marketing_controller.ex
+++ b/apps/fountain/lib/fountain_web/controllers/marketing_controller.ex
@@ -116,7 +116,7 @@ defmodule FountainWeb.MarketingController do
           "Define an agent once and every app you build can hire it over one API. " <>
             "#{Fountain.Brand.engine()} is open source and the whole platform is in " <>
             "the repo. Bring an instance up with Docker Compose or plain Kubernetes " <>
-            "manifests, on your hardware, under your own keys, with no licence key " <>
+            "manifests, on your hardware, under your own keys, with no license key " <>
             "and no seat count."
       )
     else
--- a/apps/fountain/lib/fountain_web/controllers/marketing_html.ex
+++ b/apps/fountain/lib/fountain_web/controllers/marketing_html.ex
@@ -797,7 +797,7 @@ defmodule FountainWeb.MarketingHTML do
             url: "https://rounds.demo.managoat.com",
             source: "https://github.com/managoat/rounds",
             blurb:
-              "Dependabot for infrastructure config. Enrol a repository and it gets audited on a schedule; an agent fixes what it can verify and opens the pull request. Never twice for the same finding, never again for one you closed.",
+              "Dependabot for infrastructure config. Enroll a repository and it gets audited on a schedule; an agent fixes what it can verify and opens the pull request. Never twice for the same finding, never again for one you closed.",
             shows: "an unattended product, and an agent that knows when to do nothing"
           }
         ]
@@ -957,14 +957,14 @@ defmodule FountainWeb.MarketingHTML do
         name: "Brokered credentials",
         blurb:
           "The real token never enters the sandbox. The broker attaches it on the way out, and the transcript keeps a placeholder.",
-        hosted: "Limited access. We enrol an account by hand.",
+        hosted: "Limited access. We enroll an account by hand.",
         yours: "Set BROKER_URL and BROKER_TOKEN, and list the tenants in BROKER_TENANTS.",
         docs: "/docs/concepts/secrets"
       }
     ]
   end
 
-  @doc "What each licence lets you do, in the order that matters to somebody deciding."
+  @doc "What each license lets you do, in the order that matters to somebody deciding."
   def licence_parts do
     [
       %{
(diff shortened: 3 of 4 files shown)


jhgaylor:
Rebased onto `main` and resolved. The conflict was not mechanical, so here is exactly what changed and why.

**This branch forked at `f639c07a`, which predates four merged PRs that touch the same file**: #1234, #1235, #1247/#1248, and #1251. Replaying the four commits as written would have reverted all of them. So I kept what is genuinely new here and dropped what `main` already has.

### Kept

- **`fix(marketing): American spelling for license and enroll`** — still needed; `main` has all five occurrences. Applied as written, including your call to leave `licence_parts/0`, the `:licence` key and `data-role="licence"` alone as identifiers. I also aligned two HEEx comments (`<%!-- 06. Licences --%>` and one in the section 04 comment) so they match the visible label they describe.
- **The responsive work in commit 3** — this is the valuable part and none of it is on `main`. Breakpoints on every gutter, section rhythm and card, `overflow-x-auto` on the two code blocks that were wrapping mid-token, `ml-0 sm:ml-6` on the numbered ladder, and full-width CTA buttons before they go side by side. I split it out as its own commit, `fix(marketing): /self-hosted reads on a phone`, since the restructure it was bundled with is already in.

### Dropped, because `main` already has it

- **`prerequisites/0` and `health_probe/0`** — #1235 added both, byte-identical to yours. Keeping them would have been a duplicate definition.
- **The `self_host_faq/0` entry "How do I get rid of it?"** — already on `main`.
- **The template restructure itself** — #1235 landed it.

### Dropped, because it would have regressed `main`

- **The two-paragraph hero.** #1234 deliberately trimmed this to one line before the CTA. Your branch predates that, so replaying it would have undone the trim. `main`'s one-liner is kept, with your responsive classes around it.
- **`API_CORS_ORIGINS=https://jakegaylor.com`.** The demo suite moved to `*.demo.managoat.com` in #1247/#1248. Kept `main`'s URL.
- **Section 08's `<details>` list.** #1251 moved those seven answers to `/faq` so `self_host_faq/0` renders in exactly one place; the section is now a link to `/faq#self-hosting`. Restoring the list here would have put the copy back in two places. Kept `main`'s link, with your responsive padding on it.
- **The CHANGELOG entry**, which described the restructure and had a stale `[Unreleased]` head. Rewritten to describe what actually lands: the responsive pass and the spelling.

Your original four commits are still reachable at `4b437808` if you want to compare.

Verified locally on the rebased branch: 885 tests across `fountain_web/controllers` and `brand_chrome_test`, 0 failures; `credo --strict` clean; `docs-style.py` and `destink.mjs` clean.

PR #1258, opened 2026-08-29
design(marketing): the homepage gets a shape, and the tokens get a guard

jhgaylor:
The copy landed in #1257. This is the design pass on top of it. **No copy changed.**

## The problem

Ten sections, each a centred `text-2xl` heading over a centred subhead over a centred button, for eight thousand pixels. The case study, carrying the only production numbers on the site, was set at exactly the weight of the paragraph about account limits. The alternating bands were `bg-1` (#ffffff) on a `bg-0` (#f9fafb) page, a 2% step, so the rhythm the template intended had never been visible and the hairline borders were doing all the work.

## Three surfaces

| Surface | Where |
|---|---|
| `bg-0` | the page |
| `--color-band` | the tint that groups two adjacent sections into one beat |
| `--color-ink-*` | the dark surface, which does **not** invert with the theme |

An ink section is dark in light mode; in dark mode the same tokens lift it clear of the near-black page instead, so one set of classes anchors a section in both. Two sections are ink, the manifest and the price, and they bracket the argument between them.

`--color-accent` is a warm counterpoint spent on proof and on section marks, **never on an action**. Every button stays `--color-brand`, so the accent cannot be mistaken for something to click. Its one whole-panel use is the case study, whose four numbers are now larger than any heading below the hero.

## What moved

- **The manifest carries the identity.** Five code blocks were five black rectangles floating on white; on ink they are the ground. `items-start` instead of `items-center` closed the paragraph-sized void beside every step, and the unnumbered connector holds its column as a rule rather than a gap.
- **`eyebrow/1`** gives all ten sections a one-word label and a short accent rule. Nothing told a reader which section they had landed in.
- **Two sections set their heading left** (the six claims, the limits). Dense reference is read down a margin, and a centred heading over left-aligned prose gives the eye two starting points and it uses neither.
- **Pricing moved above the closer.** The page used to say goodbye and then keep talking about money.

## Three bugs found on the way

- **The marketing nav had no responsive rule.** Six links in one flex row: below about 700px "Integrations" ran over the wordmark and the rest left the viewport, on *every* marketing page. It opens as a `<details>` menu now. No script, because the site has no bundler and a menu that needs the CDN to answer is a menu that can fail to open.
- **`--color-code-bg` was `#000000` against a `#18181b` panel in dark mode**, so every code block on every page was invisible there.
- **`tokens.css` is checked in twice and nothing generates the second copy.** `assets/css/tokens.css` is the source; `apps/fountain/priv/static/assets/tokens.css` is what a browser loads. Editing it by hand left half a comment standing, the parser swallowed the declarations after it, and three tokens resolved to empty string. That is a background that does not paint, not an error, and every test stayed green. `design_tokens_test.exs` now pins the two together and refuses a comment in the served copy. **Confirmed by reverting the fix**, which turns it red on two tests.

Also fixed two contrast defects only dark mode exposed: the case-study footnote at 2.5:1 on the warm panel, and the price-card bodies at 3.3:1. Both clear 4.5:1.

## Verification

- `mix precommit`: credo clean, sobelow 0 errors, dialyzer passed, **4,144 tests, 0 failures**.
- Shot in light, dark, and at 390px. Mobile menu opened and all six links confirmed present.
- All nine marketing routes still 200; `/self-hosted` and `/faq` visually unchanged.

## Left deliberately

`/self-hosted` and four other marketing pages already use uppercase eyebrows in their own styles (`tracking-widest` in blue, `tracking-wide` in grey). The homepage now has a fourth. Rolling `eyebrow/1` across those pages would make it one mark site-wide, but that is a separate change.

https://claude.ai/code/session_0176oj6fYqofyFL5nVZvk7q9

### DIFF
--- a/apps/fountain/lib/fountain_web/components/layouts/marketing.html.heex
+++ b/apps/fountain/lib/fountain_web/components/layouts/marketing.html.heex
@@ -8,7 +8,83 @@
           {Fountain.Brand.name()}
         </span>
       </a>
-      <nav class="flex items-center gap-3">
+      <%!-- Six links in one flex row with no responsive rule: below about
+            700px "Integrations" ran over the wordmark and the rest left the
+            viewport, on every marketing page. The links are the same list at
+            both sizes, so `<details>` opens them under the header rather than
+            a second nav duplicating them, and it needs no JavaScript. The site
+            has no bundler (Tailwind is the play CDN), so a menu that depends
+            on a script is a menu that depends on the CDN answering. --%>
+      <details class="sm:hidden relative group" data-role="mobile-nav">
+        <summary class="list-none cursor-pointer p-2 -mr-2 rounded-md text-[var(--color-text-secondary)] hover:bg-[var(--color-bg-2)] [&::-webkit-details-marker]:hidden">
+          <span class="sr-only">Menu</span>
+          <svg
+            class="size-5"
+            viewBox="0 0 20 20"
+            fill="none"
+            stroke="currentColor"
+            stroke-width="1.6"
+            stroke-linecap="round"
+            aria-hidden="true"
+          >
+            <path class="group-open:hidden" d="M3 6h14M3 10h14M3 14h14" />
+            <path class="hidden group-open:block" d="M5 5l10 10M15 5L5 15" />
+          </svg>
+        </summary>
+        <nav class="absolute right-0 top-full z-20 mt-3 w-56 rounded-lg border border-[var(--color-border)] bg-[var(--color-bg-1)] p-2 shadow-lg flex flex-col">
+          <a
+            :if={Fountain.Marketing.site?()}
+            href={~p"/integrations"}
+            class="text-sm text-[var(--color-text-secondary)] hover:text-[var(--color-text-primary)] transition-colors px-3 py-2 rounded-md hover:bg-[var(--color-bg-2)]"
+          >
+            Integrations
+          </a>
+          <a
+            :if={Fountain.Marketing.site?()}
+            href={~p"/built-with"}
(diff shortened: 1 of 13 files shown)


jhgaylor:
Two follow-ups pushed onto this branch since it opened.

**`a0e4969d` — light-mode ink is a slate, not a near-black.** At `#0c0c11` the manifest was a hole in the page rather than a section with weight. The ground is now `#1e2129`, and `ink-1` moved the *other* way, down to `#151820`: on a near-black ground a code panel had to sit a step **above** the surface to read at all, which is why five of them looked like cards floating on a void; on the slate the panel is a step **below** it, so a code block is an inset in the surface. Same classes, inverted relationship. Dark mode is untouched.

`--color-ink-muted` went with it. It was 4.2:1 on the old ground and 3.3:1 on the slate, under the floor for the small text that was its only use. A value that cleared the floor would have made it indistinguishable from `ink-secondary`, and a step in a scale that cannot be read on that scale's own ground is not a step.

**`94b8cb05` — one section mark and one label style for the whole site.** The other six marketing pages had grown four dialects of the same eyebrow (`tracking-wide` grey, `tracking-widest` mono brand-blue, `text-sm` vs `text-xs`). All of them are `eyebrow/1` now, including all nine numbered marks on `/self-hosted`. That page also set its card and table labels in mono brand-blue where the other four used muted grey; beside an accent rule that reads as a second system, so seven of them moved to the shared style. The `COPY` button keeps its mono, being a control rather than a label.

It deliberately does **not** mark every heading. Twenty-six across these pages have none, and giving them one means writing twenty-six labels, which is copy rather than styling. The rule the site now follows: the mark labels a page or a numbered division. The homepage marks all ten of its sections because they are parallel and scanned rather than read through.

Verified on each push: format, `credo --strict`, and the 1,457-test `fountain_web` tree, checking `mix test`'s own exit code. Shot in both themes.

Issue #818, opened 2026-08-18
"Sign in with Fountain": OAuth 2.0 (code + PKCE) for the standalone apps


jhgaylor:
Server: consent page, token + revoke endpoints, config-registered public clients, tokens as API keys (ADR 0021). Then the conversations app gets a Sign in with Fountain button (paste-a-key stays as fallback); fountain-team as a follow-up.

jhgaylor:
Done — "Sign in with Fountain" (OAuth 2.0 authorization code + PKCE, ADR 0021):
- Server (#819, prod `e81b97b`): `Fountain.OAuth`, `GET/POST /oauth/authorize` (consent, login round-trips back), `POST /api/oauth/token` + `/revoke`, `OAUTH_CLIENTS`; tokens are 30-day API keys that list/revoke under Account → API keys.
- Client (jhgaylor/fountain-conversations, live at https://jakegaylor.com/fountain-conversations/): Sign in with Fountain leads; paste-a-key is the fallback; sign-out revokes.

Prod verification: `oauth_authorization_codes` table migrated; `/oauth/authorize` unauth → login; consent CSP `form-action` includes `https://jakegaylor.com`; token CORS preflight allows `https://jakegaylor.com`; `invalid_grant` → 400; hosted app → prod login handoff confirmed headless. The full interactive flow (login → consent → exchange → signed in → revoke) was proven end-to-end headless against byte-identical code locally; a full interactive prod sign-in was not automated (would need the account's real password).

Follow-up (not in this issue): the same button in fountain-team (client id `fountain-team` already registered); refresh tokens; per-scope consent.

Trap worth noting: the browser CSP `form-action 'self'` blocks the consent POST's cross-origin redirect to the app — fixed by widening form-action to the registered redirect origins on the consent route only. Only visible by driving the real browser flow; every HTTP-level test passed.

PR #1265, opened 2026-08-29
feat(oauth): register your own OAuth app, in development mode, and the CORS follows

### LINKED ISSUE #1125: OAuth apps people can register themselves: tenant-owned clients in dev mode, CORS from the registry
## Goal

### 1. `oauth_clients` table

### 5. Surfaces

### Not doing (yet)

## Suggested PR split

## Acceptance

jhgaylor:
Closes #1125.

A person building an app inside a sprite, or on localhost, can now point "Sign in with Fountain" at a production server without an operator editing `OAUTH_CLIENTS` / `API_CORS_ORIGINS` and redeploying. One registration covers the sign-in *and* the app's right to call `/api` from its own origin.

## The boundary is `published`, not the redirect allowlist

The obvious answer — wildcard the sandbox domain in both lists — is a phishing kit: anyone with any sandbox could start a flow with their own PKCE challenge and their own box as `redirect_uri`, and a consenting user's 30-day full-scope key would land there. PKCE proves nothing about a flow the attacker started.

So a client you register is **unpublished**, and an unpublished client authorizes **only its owner**. Everyone else gets a rendered error page, never a redirect — and the identity check runs *before* the redirect check, so a stranger's page discloses no registration. That is exactly what makes a self-chosen redirect URI safe: the only account such a client can capture belongs to whoever registered it. Hence you may name a sprite's public URL, a `localhost` port, anything.

Config clients stay and are read as published; `get_client/1` checks them first, so a row can never shadow a first-party app. `published` is an operator flip with no self-serve path (SQL or a `mix` task for now — an admin surface was out of scope in #1125).

## What is here

- **`oauth_clients`**, tenant-owned. `client_id` is generated and never accepted from the caller — it is what the whole flow is keyed on, so letting an app choose it would let one app claim another's identity. `published` is not castable at all.
- **RFC 8252 loopback matching** for unpublished clients: `http://localhost` / `http://127.0.0.1` redirects match on any port, because the port a dev server lands on is not a fact anybody registered.
- **CORS reads the same table.** A preflight carries no authentication, so the origin is the only thing the plug can key on — which is why "an origin some registered client redirects to" is the right predicate: admitting an origin admits nobody who does not already hold a bearer key, and no cookie crosses an origin either way. Loopback matches on any port here too, or the app half-breaks (sign-in lands, first call fails) the moment Vite moves off 5173. One indexed containment against a derived `origin_keys` column.
- **`form-action` narrowed** from every registered client's origins to this request's validated redirect origin. See the trap below.
- **API + console + CLI**: `/api/oauth/clients`, Account → OAuth apps, `fountain oauth-client create|list|update|delete`. SDK types regenerated.
- **Audit**: `oauth_client.created` / `.updated` / `.deleted`, with guardrail entries; `oauth.authorized` now carries the client's `resource_id`.
- **ADR 0021 amended** (not superseded) + index refreshed. Docs, `docs/cli.md`, the configuration table and the `build-fountain-app` skill no longer tell people to ask an operator.

## One deliberate deviation from the issue

#1125 said a `sprite`-scoped key may create/list/delete clients. **These routes are `:require_full_scope` instead.** `Accounts.ApiKey`'s own moduledoc gives the sprite scope's whole reason for existing as *"code running in a sandbox can mint a permanent key that survives the conversation-scoped revoke at teardown"* — and a registered client is exactly that, one consent away. Same gate as runners, secret bindings and API keys.

The acceptance criterion shifts slightly: the owner registers once (console, CLI, or the API with their own key), then the app in the sprite runs the flow. Still no operator, still no redeploy.

## The bug only a real browser caught

Narrowing `form-action` to the client's **registered** redirect origins is wrong. With loopback any-port matching, a client registered against `:5199` and *legally* asked for `:5200` gets a header naming `:5199` — and Chrome blocks a redirect the server just approved. Every HTTP-level test passed; this only showed up when Chrome enforced the policy. Same family as the trap in #818 (`decisions/0021`).

`OAuth.form_action_origins/1` now takes the **validated requested** `redirect_uri`, not the client. Pinned by a test named for it.

## Verification

Full flow driven in headless Chrome over CDP, against a local server with `API_CORS_ORIGINS` **unset** — so the CORS pass came entirely from the registry — with the app served on `:5200` while the client registered `:5199`, proving loopback any-port for the redirect *and* for CORS in one run:

```
1. signed in -> /dashboard
2. consent page ok, badged in-development
   form-action: 'self' http://localhost:5200
3. allowed -> landed on http://localhost:5200/callback with a code
4. exchanged for a key (expires_in=2592000)
   ^ that fetch was itself cross-origin: CORS on /api/oauth/token ok
5. GET /api/conversations from http://localhost:5200 -> 200 (not blocked)
   allow-origin: http://localhost:5200
   http://127.0.0.1:5200 (nobody registered it): no header, as it should be
6. a second account on the same URL: development-mode error page, no redirect
```

Steps 3 and 6 are the two halves of #1125's acceptance criteria.

Locally green: 4,217 tests, `credo --strict`, dialyzer, sobelow, `mix format --check`, OpenAPI validation, SDK staleness check, Go build/vet/test, and all three prose gates (docs-style, vale STE, destink).

## Not doing (yet)

Per-scope consent, client secrets, refresh tokens, a publish/review workflow, and auto-registering a sandbox's redirect at spawn (explicit registration is one call, and auto-registration would leave stale origins admitted after the sandbox is gone).

https://claude.ai/code/session_01JQtvnng9E4ijxv9n1ZQ5kJ


jhgaylor:
Superseded by #1489, rebuilt from current main with the owner-only development-mode boundary, current Managoat.OAuth integration, and updated tests and contracts.

Issue #324, opened 2026-08-02
security: authenticate_user raises on OAuth-only accounts — 500 plus an account-existence oracle


jhgaylor:
`accounts.ex:132-147` calls `Bcrypt.verify_pass(password, user.password_hash)` whenever a user row exists. `oauth_registration_changeset/2` (`user.ex:63-73`) never sets `password_hash`.

**Verified** with a probe on `main`: `authenticate_user("oauthonly@example.com", "x")` raises `ArgumentError` (`:binary.bin_to_list(nil)`).

Reachable from `POST /auth/login` and `POST /api/auth/token`. The 500 distinguishes "an OAuth-only account with this email exists" from the 401 everything else returns, which defeats the deliberate anti-enumeration work elsewhere — `password_reset_controller.ex:38-50` and the `Bcrypt.no_user_verify()` constant-time branch at `accounts.ex:144`. It also floods Sentry with a trivially reachable exception.

Bounded at 10/hr/IP by the rate limiter.

**Fix:** treat a nil `password_hash` as a failed password check (still burning a `no_user_verify` for timing), returning the same `{:error, :wrong_password}` as everything else. Ideally tell the user their account uses GitHub sign-in *after* successful OAuth, not on the failed password path.

PR #1307, opened 2026-08-31
feat(cli): add auth login --api-key for GitHub-OAuth accounts

### LINKED ISSUE #1305: CLI: GitHub-OAuth accounts cannot complete `fountain auth login`
## Problem

`fountain auth login` only supports email + password (`cli/internal/cmd/auth.go` POSTs `/api/auth/token`). An account created with "Sign up with GitHub" has no password (`password_hash` is nil), so `Accounts.authenticate_user/2` returns `{:error, :wrong_password}` and the CLI shows an unexplained 401.

Since HN/launch traffic will overwhelmingly sign up via the GitHub button, this is the default hosted-path experience for new users.

## Current workaround (documented in the quickstart as of PR #1304)

Create an API key in the console (Account → API keys) and export `FOUNTAIN_API_KEY`, which wins over the credentials file. This works but never writes `~/.fountain/credentials`, so it must be exported in every shell.

## Possible fixes (pick one or more)

- **`fountain auth login --api-key`** (or an interactive "paste an API key" prompt when the password auth fails with an OAuth-account error): writes the pasted key + base URL into `~/.fountain/credentials`. Smallest change, no server work.
- **Device-authorization flow**: the CLI opens/prints a console URL, the user approves in the browser, the CLI receives an API key. Best UX; needs a server endpoint. The OAuth infrastructure from "Sign in with Fountain" (code + PKCE, tokens are API keys) could back this.
- **Server-side**: make `/api/auth/token` return a distinguishable error for passwordless accounts so the CLI can print an actionable message ("this account signs in with GitHub — create an API key at …") instead of a bare 401.

At minimum, the CLI should stop failing silently: detect the passwordless-account case and point at the API-key path.

https://claude.ai/code/session_019yCE6kdMwa7EAHr1bh9BEM

jhgaylor:
## Problem (#1305)

`fountain auth login` only exchanges email + password at `/api/auth/token`. An account created with **Sign up with GitHub** has no password, so the exchange can never succeed and the CLI died with an unexplained 401. That is the default hosted-path experience for launch traffic.

## What this does

- **`fountain auth login --api-key`** — prompts for a pasted API key (hidden input on a TTY; a plain stdin read when piped, so `echo $KEY | fountain auth login --api-key` works in scripts). The key is verified against `GET /api/auth/me` before anything is written, then saved to `~/.fountain/credentials` with the base URL. Unlike the exported-`FOUNTAIN_API_KEY` workaround from #1304, it survives new shells and profiles work.
- The key is prompted for, never accepted as a flag value, so it stays out of shell history.
- The password path's 401 now explains the GitHub-account case and points at the console's `/api-keys` page and the new flag, instead of failing silently.
- `docs/cli.md` documents the flow; `docs/cli/commands.md` regenerated.

## What this deliberately does not do

The issue's server-side option (a distinguishable error for passwordless accounts) is not taken: `Accounts.authenticate_user/2` answers every bad credential identically on purpose so login cannot become an account-existence or auth-method oracle (#324, see the comment in `apps/fountain/lib/fountain/accounts.ex`). The device-authorization flow remains open as a future UX improvement.

## Testing

- `go test ./...` and `go vet ./...` in `cli/` — green (new `auth_test.go` covers the happy path against the real `/api/auth/me` shape, and bad-key rejection).
- `mix test apps/fountain/test/fountain/docs_test.exs` — 43 tests, 0 failures.
- All three prose gates clean (`docs-style.py`, vale STE 0 errors, destink).

Closes #1305

https://claude.ai/code/session_01Dex5a9SZHXpST6KyueR3Bc

### DIFF
--- a/cli/internal/cmd/auth.go
+++ b/cli/internal/cmd/auth.go
@@ -24,12 +24,21 @@ func init() {
 		Use:   "auth",
 		Short: "Authenticate against the Fountain API",
 	}
-	authCmd.AddCommand(
-		&cobra.Command{
-			Use:   "login",
-			Short: "Authenticate and save credentials",
-			RunE:  func(cmd *cobra.Command, args []string) error { return authLogin() },
+	loginCmd := &cobra.Command{
+		Use:   "login",
+		Short: "Authenticate and save credentials",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			withAPIKey, _ := cmd.Flags().GetBool("api-key")
+			if withAPIKey {
+				return authLoginAPIKey()
+			}
+			return authLogin()
 		},
+	}
+	loginCmd.Flags().Bool("api-key", false,
+		"paste an API key instead of email + password (for accounts that sign in with GitHub)")
+	authCmd.AddCommand(
+		loginCmd,
 		&cobra.Command{
 			Use:   "logout",
 			Short: "Remove saved credentials",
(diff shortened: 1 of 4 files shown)


jhgaylor (commented):
Reviewed the API-key login flow, credential persistence, error handling, tests, and generated docs. The key is kept out of argv, verified against /api/auth/me before the profile is written, and invalid keys cannot replace existing credentials. I also ran go test ./..., go vet ./..., and git diff --check; all pass. No findings.

PR #1309, opened 2026-08-31
feat: device-authorization login for the CLI (fountain auth login --device)

### LINKED ISSUE #1305: CLI: GitHub-OAuth accounts cannot complete `fountain auth login`

jhgaylor:
~~Stacked on #1307~~ — #1307 merged; this now targets `main` directly (rebased, so the diff is the device flow alone). Merging also publishes SDK **1.12.0** (minor: the new endpoint types), per the release guard.

## What (#1305, the "best UX" option)

An account created with **Sign up with GitHub** has no password, so `fountain auth login` can never exchange one. This adds the device-authorization flow (RFC 8628 shape): the CLI shows a short code and a console URL, the user approves in a signed-in browser, and the CLI collects a full-scope API key.

```
$ fountain auth login --device
First, note your one-time code: BCDF-GHJK

Then approve this device at: https://managoat.com/device
(opened in your browser)

Waiting for approval...
Logged in as you@example.com. Credentials written to ~/.fountain/credentials (profile: default).
```

## Server

- **`Fountain.OAuth` extension**, following ADR 0021's model (tokens are ordinary API keys): a new `oauth_device_grants` table holds a hashed high-entropy `device_code` (never typed, stays on the polling machine) and a short `XXXX-XXXX` `user_code` from RFC 8628's consonant alphabet (20^8 space, 15-minute expiry, single use).
- **`POST /api/auth/device`** starts a grant (public, rate-limited 10/hr/IP). **`POST /api/auth/device/token`** is the poll: RFC 8628 error vocabulary (`authorization_pending`, `slow_down` with per-grant pacing, `access_denied`, `expired_token`, `invalid_grant`), and on approval it mints the same key `POST /api/auth/token` would — same name pattern, same response shape, full scope, no expiry.
- **`/device`** (LiveView, behind `require_authenticated_user`): enter or arrive with the code (`verification_uri_complete` prefills it), see exactly which account gets the key and a phishing warning, approve or deny. Audits `oauth.device_approved` / `oauth.device_denied` (actor `ui`); the mint audits `api_key.created` (actor `api`) as always.
- Suspended/unverified accounts cannot collect a key even after approval (belt-and-braces; the console hooks already turn them away). OpenAPI operations and schemas included; `OAuth.prune_expired` sweeps expired grants.

## CLI

- `fountain auth login --device`, plus browser open on a real TTY only (scripts and tests just get the printed URL). Polling honors the server's `interval` and backs off 5s on `slow_down`. The password path's 401 hint now leads with `--device`.
- Docs: `docs/cli.md` authentication section, `docs/api.md` endpoint list, regenerated `docs/cli/commands.md`.

## Testing

- `mix test` full suite: **4,200 tests, 0 failures** (new coverage: context happy path/single-use/slow_down/deny/expiry/suspension, controller flow, LiveView approve/deny/wrong-code).
- `mix compile --warnings-as-errors`, `mix format --check-formatted`, `credo --strict`, dialyzer, sobelow, OpenAPI spec generation + `jq empty` — all green.
- `go test ./...` + `go vet` in `cli/` — green, including a full-flow test against the exact controller JSON (pending → slow_down → key) with the sleep instrumented.
- All three prose gates clean (docs-style, vale STE, destink); `docs_test.exs` 43/43.

Closes #1305

https://claude.ai/code/session_01Dex5a9SZHXpST6KyueR3Bc

---

**Update (d27d6f44): the device flow is now the default interactive login.** True auto-detection of a passwordless account is impossible by design (the uniform 401 is #324's anti-enumeration guard), so the default became a flow with nothing to detect: on a terminal, `fountain auth login` runs the device flow, which works for every account. `--password` forces the email+password prompt; **piped stdin keeps the password read unchanged**, so scripts that feed credentials are unaffected. A password 401 on a terminal now offers "Sign in through your browser instead? [Y/n]" and rolls into the device flow instead of dying with a hint.

### DIFF
--- a/apps/fountain/lib/fountain/oauth.ex
+++ b/apps/fountain/lib/fountain/oauth.ex
@@ -33,10 +33,18 @@ defmodule Fountain.OAuth do
   import Ecto.Query, warn: false
 
   alias Fountain.{Accounts, Audit, Repo}
-  alias Fountain.OAuth.AuthorizationCode
+  alias Fountain.OAuth.{AuthorizationCode, DeviceGrant}
 
   @code_ttl_seconds 300
   @token_ttl_seconds 30 * 24 * 3600
+
+  @device_ttl_seconds 900
+  @device_interval_seconds 5
+  # RFC 8628's suggested consonant alphabet: no vowels (no words, no
+  # accidental profanity), no ambiguous glyphs. 20^8 codes for a
+  # fifteen-minute, rate-limited window.
+  @user_code_alphabet ~c"BCDFGHJKLMNPQRSTVWXZ"
+  @user_code_length 8
 
   @type client :: %{id: String.t(), name: String.t(), redirect_uris: [String.t()]}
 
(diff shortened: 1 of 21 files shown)


jhgaylor (commented):
Blocking: preserve the device approval URL across authentication. The new /device route is inside the browser_authenticated pipeline, whose TenantSessionAuth plug redirects an unauthenticated visit straight to /auth/login without calling FountainWeb.ReturnTo.stash. As a result, opening the supplied /device?code=... URL while signed out drops both the destination and the prefilled code; after password or GitHub sign-in, the browser lands on /dashboard while the CLI keeps polling. This is a normal case for the primary GitHub-account use case. Please stash the full path and query before redirecting (the OAuthAuthorizeController flow is the existing pattern) and add a round-trip test that starts signed out and returns to the confirmation screen after login.

jhgaylor:
Fixed in 354502a3. `TenantSessionAuth` now stashes the full path + query via `FountainWeb.ReturnTo` (the #818 consent-page pattern) before its login redirect — GET requests only, since a POST path replayed as the post-login GET lands nowhere. Both login doors already pop the stash, so password and GitHub sign-ins alike return to `/device?code=…` with the code prefilled. Added the requested round-trip test: signed out with `?code=` → login → confirmation screen → approve → the waiting poll mints for that account. Side effect worth noting: every console deep link now round-trips through login instead of landing on the dashboard, which was the pipeline-wide version of the same bug.

PR #1317, opened 2026-09-01
test(finance): seed the closed previous month, not the running one

jhgaylor:
## The flake

`/admin/finance` clips every usage interval to now — `SandboxUsage`'s ceiling is `earliest(period_end, now)` — but this file's `ran/3` anchored its sandbox and turn at the start of the *running* month and extended forward. A 100-hour sandbox therefore reports its full 100 hours only once the month is 100 hours old, so five of these tests fail for roughly the first four days of every UTC month. The 2026-09-01 00:16 UTC CI runs (e.g. PR #1316's) are the first month boundary since the finance panel shipped, and they went red on exactly these tests.

## The fix

Seed through `ran/3` into the **previous, closed month** and open the page with `?months_ago=1` (the month picker the page already has). A closed month has a fixed ceiling, so the numbers are exact whenever the suite runs. The basis-toggle selector follows (`months_ago=1&basis=turn` — the link carries the socket's month). The window-navigation test keeps opening the default month; it seeds nothing.

## Verification, against the live flake condition

Run at ~00:45 UTC on Sept 1, inside the flake window:

- Before the change: 13 tests, **5 failures** (the same five CI reported).
- After the change: 13 tests, **0 failures**.

## Known residual

`DashboardLiveTest`'s turn-hours test has the same shape with a 2-hour span, so it stays red for the first two UTC hours of each month (self-heals at 02:00 UTC). The dashboard has no month picker to lean on, so making it deterministic needs a clock seam through `Billing.usage_summary/3`; left out of this PR deliberately.

https://claude.ai/code/session_01SAs4xAGFD3FfRjgfLFvXHV

### DIFF
--- a/ee/test/fountain/billing_reconciliation_test.exs
+++ b/ee/test/fountain/billing_reconciliation_test.exs
@@ -37,10 +37,15 @@ defmodule Fountain.Billing.ReconciliationTest do
     admin
   end
 
-  defp month_start do
-    now = DateTime.utc_now()
-    %DateTime{now | day: 1, hour: 0, minute: 0, second: 0, microsecond: {0, 0}}
-  end
+  # The previous, closed month: the page and Finance.summary clip every
+  # interval to `now`, so hours seeded forward from the running month's start
+  # only report in full once the month is that many hours old — these tests
+  # failed on the first CI runs of 2026-09-01, the same way
+  # admin_finance_live_test.exs did. A closed month has a fixed ceiling, so
+  # the numbers are exact whenever the suite runs. Tests that seed through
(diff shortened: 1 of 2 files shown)


jhgaylor:
Second file, same flake: `Fountain.Billing.ReconciliationTest` seeded a 10-hour sandbox at the running month's start, so partition 4 failed alongside the finance file. Fixed the same way (closed previous month, `month_range(1)`, `?months_ago=1`), verified against the live flake condition: 2 of 3 failures before, 0 after. Remaining known-red on this PR's CI until 02:00 UTC: the dashboard turn-hours test (partition 1), per the PR description.

Issue #1299, opened 2026-08-31
Platform connection providers beyond Gmail: Google Calendar scopes, Microsoft (Outlook mail + calendar, Teams), Slack


jhgaylor:
Follow-up to #1178, sibling of #1186. Downstream ask from Reflex (jhgaylor/reflex), a consumer personal-assistant app on the SDK.

## The gap

#1178 shipped connections for exactly one platform provider (Google) requesting exactly one product's scopes (`gmail.modify`). #1186 grows tenant-defined providers, but its audience is tenants who can register their own OAuth apps — and it explicitly leaves out Fountain-served tool servers for other providers and the Google calendar scopes. A consumer-facing app's users will never register an app at Microsoft or Slack; for them "connect your calendar" only works as a **platform provider**: Fountain owns the OAuth client, the tenant clicks Connect, and either a Fountain-served MCP server or a brokered token carries the capability, the way Gmail does today.

Reflex already ships the UI for this ([jhgaylor/reflex#2](https://github.com/jhgaylor/reflex/pull/2)): a catalog grouped Email / Calendar / Chat — Gmail, Outlook, Google Calendar, Outlook Calendar, Slack, Teams — where each row lights up when `GET /api/connections/providers` shows a configured provider whose requested scopes cover it, and renders "soon" until then. Gmail is the only live row.

## Asked for, roughly in value order

1. **Calendar scopes on the existing Google provider.** Add `https://www.googleapis.com/auth/calendar` (or `calendar.events` + `calendar.readonly`) to `Fountain.Connections.Google`, plus either a Fountain-served Calendar MCP server next to the Gmail one or just the brokered `GOOGLE_ACCESS_TOKEN` (the broker already attaches to `www.googleapis.com`, which is where `calendar/v3` lives — this half may already work unverified). `include_granted_scopes=true` is already sent, so incremental consent is in place; existing connections would show the new scopes as connectable-again rather than magically granted.
2. **A Microsoft platform provider** (Azure AD app, Graph API): `Mail.ReadWrite` + `Mail.Send`, `Calendars.ReadWrite`, and `Chat.ReadWrite` / `ChannelMessage.Send` for Teams — one provider, one sign-in, mail + calendar + chat, mirroring how one Google sign-in should cover Gmail + Calendar. Brokered token to `graph.microsoft.com`; Fountain-served tool servers per product can come later.
[...]

PR #1319, opened 2026-09-01
feat(connections): platform provider registry — Microsoft, Slack, and Google Calendar scopes (#1299)

jhgaylor:
Implements the platform half of #1299 on top of #1187: the hardcoded "Google is the one platform provider" becomes a small config-driven registry — Google, Microsoft, Slack — so a #1299-style provider is data plus at most a named quirk hook, not a new module.

## What's here

**`Fountain.Connections.Platform`** — the registry. Each entry is the same `Provider` struct #1187 introduced, built from config (`<SLUG>_OAUTH_CLIENT_ID` / `_SECRET`), `user_id: nil`, slug as the reserved id. The list always includes all three, with `configured: false` until the operator sets the env vars, so Reflex-style clients can render "soon" rows. Scopes stay in `GET /api/connections/providers`, which is the contract #1299 names as load-bearing.

**Google grows the Calendar scope** (`https://www.googleapis.com/auth/calendar`). `www.googleapis.com` was already in `token_hosts`, so the brokered token covers `calendar/v3` as-is; `include_granted_scopes=true` was already sent, so existing connections keep their grants and a reconnect adds the scope.

**Microsoft** — one Entra app on the `common` endpoint; `Mail.ReadWrite`, `Mail.Send`, `Calendars.ReadWrite`, `Chat.ReadWrite` (+ `openid email offline_access User.Read`) brokered to `graph.microsoft.com` under `MICROSOFT_ACCESS_TOKEN`. `ChannelMessage.Send` is deliberately not in the default (admin consent on work tenants); operators add it via `MICROSOFT_OAUTH_SCOPES`.

**Slack** — user tokens (not bot): the request goes in `user_scope` with `scope` emptied, and the token comes back nested under `authed_user`, which a `Platform.normalize_token_body/2` hook lifts to the RFC 6749 top level so `Connections.OAuth` stays one code path. No refresh token and no expiry is Slack's design, so the platform insist-on-refresh-token rule now applies only when the token expires at all.

**Scope override per provider** (`GOOGLE_OAUTH_SCOPES`, `MICROSOFT_OAUTH_SCOPES`, `SLACK_OAUTH_USER_SCOPES`) — the app-verification lever: a deployment doesn't request what its verification doesn't cover, and those products stay dark instead of erroring.

**Guards that multi-platform makes necessary:**
- Platform connections all carry `provider_id: nil`; the `provider` slug is now part of the dedupe key, so the same account label on two platform providers is two connections.
- The Fountain-served Gmail endpoint refuses a non-Google connection with a readable 400 ("attach this connection to a remote MCP server or use its brokered env key") instead of sending a Graph token to Gmail.
- The Slack tenant-provider *preset* is gone — `slack` is a reserved slug now, and the platform row is the one that works without an app registration.

**ADR 0033 amended** — decision 1 required exactly this amendment before adding a platform provider; the recorded reason: a consumer app's users will never register an app at Microsoft or Slack, so "connect your calendar" only works when Fountain owns the client. The long tail stays with tenant `oauth2` providers and MCP-spec discovery.

Docs: catalog pages for `microsoft` and `slack`, google page updated, configuration reference rows, `.env.example`. All three prose gates green.

## Not in this PR (ops, tracked separately)

- Registering the actual Microsoft/Slack apps and setting prod secrets — until then both list as not configured.
- Google app verification for the calendar scope on the hosted instance (Testing mode today).
- Per-product Fountain-served tool servers — deliberate floor: brokered env key + remote/stdio MCP servers (ADR 0033 consequence).

Closes the code half of #1299.

https://claude.ai/code/session_01Npkq4utPYxR6NdwetyzCzr


jhgaylor:
The `sdk-no-release` label is deliberate: the only SDK diff is regenerated description strings in `src/generated/openapi.ts` (no shape change), so publishing 1.13.1 for doc comments would be noise. The next real SDK release carries them.

Issue #1336, opened 2026-09-01
managoat_substitution: first extraction and the umbrella mechanics


jhgaylor:
Part of the library-extraction tracker #1334. The first extraction, chosen because `Fountain.Substitution` has **zero** dependencies (no Repo, no config, no other context). Its job is to prove the umbrella mechanics so every later extraction is a copy of this PR.

## What moves

- `apps/fountain/lib/fountain/substitution.ex` → `apps/managoat_substitution/lib/managoat/substitution.ex` as `Managoat.Substitution`.
- Its unit and property tests (`ExUnitProperties`) move with it.
- `Fountain.Substitution` stays as a one-line delegate for one release, then goes.

The Go twin in `cli/internal/substitution` stays where it is; note in the README that the two must agree on `${VAR}` and `$$` semantics and that the CLI's tests are the second implementation.

## The mechanics this PR has to settle (the real deliverable)

- `mix.exs` for the new app: `Managoat.Substitution`, version, `licenses: ["Apache-2.0"]`, `package` metadata, `LICENSE` file in the app dir.
- `apps/fountain/mix.exs` gains `{:managoat_substitution, in_umbrella: true}`.
- **CI partitions:** `scripts/test-partition.sh` and the six partition jobs must pick up the new app's tests. Check that `mix test` at the root still runs everything, and that `mix test <path>` with an app-relative path resolves (root-relative paths to `ee/` already do not; do not add a second trap).
- **Coverage:** `coverage.exs` is read by both existing `mix.exs` files; the new app reads it too, and `scripts/coverage-gate.exs` must merge its export.
- **Format, credo, dialyzer, sobelow:** each runs over the new app. `mix precommit` at the root covers it. Sobelow's merged-tree scan (see the umbrella `mix.exs` comment about overlaying `ee/lib`) must not need a second overlay.
- **Dockerfile:** the release stage `COPY`s `apps/fountain`; a new app dir needs its own `COPY` or `mix release` dies with no image and a green CI (#884 shape). Add the `docs_test`-style assertion if there is a natural place for it.
- **Release:** `rel/` and `mix release` for the umbrella still produce one release containing both apps.

## Acceptance

- `mix precommit` green at the root; full CI green including the release boot check.
- A short section in `CONTRIBUTING.md`: "adding an umbrella library app", listing the files above.

PR #1347, opened 2026-09-01
feat(libraries): managoat_substitution, the first umbrella library app

jhgaylor:
The first extraction under ADR 0037 (#1346), closing #1336. `Fountain.Substitution` becomes `Managoat.Substitution` in `apps/managoat_substitution`, behaviour unchanged (the test file moved with three doctests added). The point of the PR is the umbrella mechanics every later extraction copies.

**The library app**

- `apps/managoat_substitution`: `mix.exs` with `package` metadata (`licenses: ["Apache-2.0"]`), its own `LICENSE`, `README.md`, `.formatter.exs`, `test/test_helper.exs`. `mix hex.build` succeeds from its directory.
- **No `config_path` to the umbrella config**, deliberately. Run from its own directory the app initially failed to boot because `config/runtime.exs` calls `Fountain.PublicUrl`, which is not on a library's code path. A library that reads no `:fountain` configuration has no use for Fountain's config; without the line it boots with none, which is what a hex consumer gets too.
- `apps/fountain/mix.exs` takes it as `{:managoat_substitution, in_umbrella: true}`; the one call site in `ConversationServer` aliases the new module. No shim: nothing outside the umbrella called `Fountain.Substitution`, so a delegate would be a coverage hole with no caller.

**The mechanics**

- **CI.** The partition scripts are written against `apps/fountain` and `ee/`, so a library's tests were not in any partition and its coverage was unmeasured. `scripts/test-libraries.sh` runs every `apps/managoat_*` suite from its own directory with the same "prove it ran tests, prove it exported" discipline as `test-partition.sh`, and copies the export into `apps/fountain/cover/lib-<app>.coverdata`. Partition 1 runs it (a job of its own carries ~60s fixed cost for a suite that takes seconds); the upload step gains `lib-*` globs; the coverage job now checks it received one library export per `apps/managoat_*` directory in addition to one per partition. Verified locally that `scripts/coverage-gate.exs` imports the library export.
- **Dockerfile.** One `COPY apps/managoat_substitution/mix.exs` in the deps layer; the umbrella loads every child's `mix.exs` to resolve deps and the image build would fail without it, which CI does not run.
- **Guard test.** `apps/fountain/test/fountain/umbrella_layout_test.exs` walks every `apps/managoat_*` and checks: the files above exist, Apache-2.0 in both `LICENSE` and package metadata, the in-umbrella dep line, the Dockerfile `COPY` line, only `Managoat.*` modules defined, and no `Fountain.*`/`FountainWeb.*` reference, `:fountain` config read, or `[:fountain, …]` telemetry under `lib/` or `test/`. Mutation-checked: an injected `Fountain.` reference and a deleted `COPY` line each fail the right test.
- Format, credo, dialyzer and `mix test` at the root already reach the new app (`subdirectories: ["apps/*"]`, `apps/*/lib/`, root runs).

**Docs**

- `CONTRIBUTING.md`: "Adding an umbrella library app", the checklist above.
- `CLAUDE.md` layout and substitution sections, `NOTICE` licence table, `cli/README.md` history note, `CHANGELOG.md` `[Unreleased]`.
- `scripts/test-timings.tsv` drops the moved file.

**Not in this PR**

The "Built so far" block in `decisions/0037` still says nothing is built; it lands in #1346 first, so the one-line update to that block follows once both are on main.

**Gates run locally**

`mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict`, `MIX_ENV=dev mix dialyzer`, the library suite (16 tests, 3 doctests, 3 properties, 100% coverage), the guard test, and the full umbrella suite. Results in the final comment.

https://claude.ai/code/session_019rKjC1mfpREc5pKrNbqu8G


jhgaylor:
Gates run locally on this branch (Elixir 1.19.2 via mise, the CI pin):

| Gate | Result |
|---|---|
| `mix compile --warnings-as-errors` (test) | clean |
| `mix format --check-formatted` | clean |
| `mix credo --strict` | no issues, 696 files |
| `MIX_ENV=dev mix dialyzer` | 0 errors, 0 skips |
| `mix test` at the root | `==> managoat_substitution` 3 doctests, 3 properties, 16 tests, 0 failures; `==> fountain` 6 doctests, 4263 tests, 0 failures |
| `mix test --cover` in `apps/managoat_substitution` | 100% |
| `scripts/test-libraries.sh` | exports `apps/fountain/cover/lib-managoat_substitution.coverdata`; `elixir scripts/coverage-gate.exs` imports it |
| `mix hex.build` in the app | builds `managoat_substitution-0.1.0.tar` |
| guard mutations | an injected `Fountain.` reference and a deleted Dockerfile `COPY` each fail exactly one test |

On CI the coverage job logged `merging 6 partitions and 1 library apps covering 4279 tests` and `merged 7 export(s)` at 86.62%, so the library path in the merge is exercised end to end.

One observation, not from this change: on one of two full local runs the buzz harness test logged a `DBConnection.OwnershipError` (owner terminated, `Fountain.Buzz.Manager.start_harness_link/2`) without failing anything. The other run was silent. Noting it in case it shows up as a flake later.

Issue #1334, opened 2026-09-01
Tracker: pull Fountain apart into managoat component libraries, umbrella-first


jhgaylor:
Tracker for pulling Fountain apart into component libraries under the [managoat](https://github.com/managoat) GitHub org, and for the server ending up rebuilt on top of them. This issue records what a survey of the dependency graph found on 2026-09-01, the decisions taken that day, and the order of work. Each seam is a sub-issue.

## What the survey found

A context-level dependency map (module references between `lib/fountain/<context>` directories, plus `ee/`) shows three facts that decide everything else:

1. **Three subsystems never touch the database.** `Fountain.Sandbox` (+ the Sprites/E2B/Daytona adapters), `Fountain.Runtimes` (the ACP peer) and `Fountain.Broker` have zero references to `Fountain.Repo` or any Ecto schema. They read config, call Req or the Sprites SDK, and hand results back. They are libraries already, just not packaged as such.
2. **`Conversations` is the hub, not a seam.** It references 26 other contexts; `ConversationServer` alone is 4,489 lines. Every extraction ends with conversations calling the new library. It is rewritten last, by subtraction.
3. **The coupling that fights every extraction is uniform.** Config is `Application.get_env(:fountain, …)` (sandbox reads 10 keys, broker 9), telemetry events live under `[:fountain, …]`, and audit records in the context. Each library needs its own otp_app config or an options map. Mechanical, not design.

The candidates, ranked by cost and reuse:

| Library | Source today | Lines | Depends on | Reuse elsewhere |
|---|---|---|---|---|
| sandbox | `Fountain.Sandbox`, three adapters, `Fake`, `SandboxConformanceCase`, `Retry` | ~3,700 | sprites-ex, Req | High. One behaviour over three vendors with an executable conformance suite. Nothing like it on hex. |
| acp | `Runtimes.ACP.{Peer,Protocol,Blocks,Tracer,Usage}` + `Permissions` | ~2,300 | sandbox | Medium. `acpex` and `agent_client_protocol` exist on hex at 0.1.x; ours adds the client-side session, the permission policy and the normalised block format. |
| mcp_auth | `Connections.{McpDiscovery,Mcp,UrlGuard,McpServerCatalog}` | ~700 | Req | High and niche. RFC 9728 / 8414 / 7591 discovery with an SSRF guard. Nothing on hex. |
[...]

PR #1348, opened 2026-09-01
docs(decisions): ADR 0037 records managoat_substitution as built

jhgaylor:
The follow-up owed from #1347: ADR 0037's status block said nothing was built. It now names `managoat_substitution` and the three mechanics (library test step, Dockerfile `COPY`, the guard test) as built, and lists the eight libraries and the graduation recipe as not built. Index regenerated; `okf validate decisions` has zero errors and no findings on 0037.

Part of #1334.

https://claude.ai/code/session_019rKjC1mfpREc5pKrNbqu8G


jhgaylor:
Superseded: #1349 (managoat_sandbox) updated the same "Built so far" block in ADR 0037 and named managoat_substitution as built alongside it, so this PR's change is already on main.

Issue #1372, opened 2026-09-02
ConversationServer: sprite environment assembly and runtime file writes out (SpriteEnv, Provisioning)


jhgaylor:
Part of #1369. The largest pure family: turning rows and secrets into the sprite's environment and the files the runtime needs before the first turn.

**Functions:** `build_sprite_env`, `do_build_sprite_env`, `merge_secrets`, `conversation_env`, `sandbox_id_env`, `sandbox_url_env`, `record_sandbox_url`, `otel_propagation_env`, `git_author_env` (public), `run_setup_script`, `write_runtime_config`, `write_instructions`, `prepare_runtime_sprite`, `prepare_acp_adapter`, `load_tenant_state`, `create_sandbox_handle`.

**Home:** `Fountain.Conversations.SpriteEnv` (pure: rows and decrypted secrets in, an ordered env list out, with the precedence rule vault-over-environment stated once) and `Fountain.Conversations.Provisioning`, which already exists (678 lines) and is where `run_setup_script`, `write_runtime_config`, `write_instructions`, `prepare_runtime_sprite`, `prepare_acp_adapter` belong beside the pipeline they are steps of.

**Watch:** this family is where the broker's placeholders enter the env (`broker_env`, `split_inference` are in the egress sub-issue; the seam between the two is `merge_secrets`). Do this one before egress so egress has a clean function to hand its placeholders to. `conversation_server_instructions_test.exs` and `provisioning_test.exs` pin the observable file layout.

Acceptance: server and provisioning tests unchanged and green; `SpriteEnv` unit-tested on precedence, redaction registration and the otel pair; size pin lowered.

Issue #1369, opened 2026-09-02
Tracker: ConversationServer by subtraction


jhgaylor:
The campaign ADR 0037 promised instead of a rewrite: once the libraries took their weight out of `ConversationServer`, what is left is refactored in place, by subtraction, with the suite green throughout. The eight libraries are extracted (#1334); this is that refactor.

## What the file is today (`main` at 2d99ef25)

`apps/fountain/lib/fountain/conversations/conversation_server.ex`: **4,498 lines**, 32 `handle_info`, 12 `handle_call`, 3 `handle_cast`, 1 `handle_continue`, **187 private functions**. It calls the `Conversations` context 108 times, `Managoat.Sandbox` 28, `Fountain.Broker` 20, telemetry 19, `Fountain.Runtimes` 18, `Managoat.ACP.Peer` 7. It is covered by thirteen test files, 4,756 lines, all of which drive it as a process (`conversation_server_*_test.exs`).

The comment headers in the file say `public api / tool bridge / GenServer / brokerage / ACP reports / sandbox lifetime / helpers / connection`. **The functions under them do not match the headers.** The "brokerage" section holds sprite-environment assembly and runtime file writes; "helpers" holds the whole turn machinery; "sandbox lifetime" holds permission resolution and redaction beside reclaim. The sub-issues below are cut by what the functions actually do, and each names the functions it takes so nothing is guessed twice.

## The rule for every move

- A sub-issue moves a **function family** into a module under `Fountain.Conversations.*` as functions over explicit arguments (rows, the state's relevant fields, a handle), not over the GenServer state as a whole. The GenServer keeps the callback and delegates.
- **No behaviour change.** The thirteen `conversation_server_*_test.exs` files pass unchanged; that is the acceptance for every sub-issue. New modules get unit tests of their own that do not start the server.
- **Audit stays in the context** (ADR 0013): a moved function that records audit does so in its new home, and `audit_guardrail_test` still sees it.
- **A ratchet.** The first sub-issue adds a test that pins the file's line count at its current value and fails when it grows; each sub-issue lowers the pin. The pattern is the docs-style allowlist that only shrinks (#911).
- One PR per sub-issue, in the order below, since they touch the same file.

## Sub-issues, in order

Listed as GitHub sub-issues, in work order. Label: `conversation-server-2026-09`.

1. #1370 ConversationServer: the size ratchet and honest section headers
2. #1371 ConversationServer: MCP server list and callback key out (McpServers, CallbackKey)
[...]

PR #1398, opened 2026-09-02
refactor(conversations): egress brokerage and connection secrets out (Egress)

### LINKED ISSUE #1373: ConversationServer: egress brokerage and connection secrets out (Egress)
Part of #1369. Everything ADR 0019 wired into the server: which secrets are brokered, the connections that contribute tokens, the broker session's life, and the network floor.

**Functions:** `brokered?`, `split_brokered`, `split_inference`, `add_connection_secrets`, `connection_bindings`, `active_connections_by_id`, `remote_connection_hosts`, `refresh_connection_secrets`, `with_connection_servers`, `broker_bindings`, `broker_env`, `broker_prepare`, `broker_switch_to_api_key`, `broker_install_ca`, `broker_refresh`, `broker_release`, `apply_egress_policy`, plus the three `handle_info` clauses in that section (the refresh timer, the release on terminate, the CA install result).

**Home:** `Fountain.Conversations.Egress`. It talks to `Fountain.Broker` (the facade with its two backends after #1357) and `Fountain.Connections`; the server keeps only the timer and the state fields the session token and expiry live in.

**Why now:** the Agent Vault client is scheduled for deletion after the home-cloud flip (#1357's checklist). Having this family in one module means that deletion touches one file plus the facade, not the server.

**Watch:** `conversation_server_broker_test.exs`, `conversation_egress_test.exs` and `broker_native_test.exs` are the pins; `conversation_server_identity_test.exs` covers the per-conversation identity the broker attaches to.

Acceptance: those tests unchanged and green; `Egress` unit tests for the split rules (bound key gets no catalog pair, inference split, connection hosts); size pin lowered.

jhgaylor:
Part of #1369, fourth of eight. Everything ADR 0019 wired into the server leaves for `Fountain.Conversations.Egress`, as functions over rows and values; the server keeps the session and its placeholders in state and three short wrappers that apply what comes back.

## Functions moved

All seventeen named functions exist and all seventeen moved. **One correction to the issue:** it also lists "the three `handle_info` clauses in that section (the refresh timer, the release on terminate, the CA install result)". There are none. The section has no `handle_info` clause and the server has no broker timer: the refresh runs at the turn kick (`broker_refresh/1` from `run_fresh_turn`), the release is called from `terminate_conversation`, the reclaim path and the provision failure path, and the CA install is a synchronous step of provision and reattach. Nothing to move there, and no timer was invented.

| was (`ConversationServer`, all private) | now (`Egress`) |
|---|---|
| `brokered?/1` (over state) | `brokered?/1` (user id) |
| `split_brokered/3`, `split_inference/4`, `add_connection_secrets/4`, `connection_bindings/3` | same names, public |
| `active_connections_by_id/1`, `remote_connection_hosts/2` | same names, private |
| `refresh_connection_secrets/1` (over state) | `refresh_connection_secrets/3` (keys, user id, brokered map), returns `{brokered, rotated?}` |
| `with_connection_servers/2` (agent, state) | `with_connection_servers/4` (agent, user id, conversation id, callback token) |
| `broker_bindings/1` | `bindings/1` |
| `broker_env/1` (over state) | `sandbox_env/1` (session or nil) |
| `broker_prepare/1` | `prepare/4` (the stage-publishing mint); the `if brokered?` and the state update stay in the server wrapper |
| `broker_switch_to_api_key/1` | `drop_oauth_token/3` (pure) + `reprepare/5` (mint and rebuild the env) |
| `broker_refresh/1` | `refresh_connection_secrets/3` + `reprepare/5`; the "rotated or expiring" decision stays in the wrapper |
| `broker_install_ca/2` | `install_ca/3` (session, handle, conversation id) |
| `broker_release/1` | `release/2` (user id, conversation id) |
| `apply_egress_policy/4` | `apply_policy/4` |

`reprepare/5` is one function where the server had two copies of the same six lines (the OAuth switch and the refresh both re-mint and swap the proxy variables in the env); the bodies were identical, so this is a dedupe of moved code, not a change. `Egress.prepare/4` publishes the `broker` stage exactly as before; `reprepare/5` publishes none, as neither caller did.

## What the server keeps

Under the `sprite environment and egress` header, now placed above them: the `build_sprite_env/5` wrapper from #1372 and three wrappers, each of which unpacks state, calls `Egress`, and applies the result:

- `broker_prepare/1`: `if Egress.brokered?(user_id)`, then `Egress.prepare/4` and `broker: session`; else the state untouched.
- `broker_switch_to_api_key/1`: `drop_oauth_token/3` into the three fields, then `reprepare/5` into `broker` and `sprite_env`; the same warning on failure, the same state kept.
- `broker_refresh/1`: `refresh_connection_secrets/3` into `brokered`, then `reprepare/5` only if rotated or the session is expiring (`Fountain.Broker.expiring?/1`), the same warning and comment on failure.

Every other call site is a direct `Egress.` call with the fields spelled out: the four split rules at provision, `brokered?` twice, `with_connection_servers` three times, `install_ca` twice, `release` three times, `apply_policy` twice, `sandbox_env` once (the `brokered:` argument to `SpriteEnv.build/4`, the seam #1372 left).

The facade only: `Egress` calls `Fountain.Broker` and never a backend. It also has no `Repo` call; `audit_guardrail_test` is green.

`webhooks/events_test.exs` keeps a list of the files that may call `publish_stage/4` (its comment says a new file calling it belongs there); `egress.ex` is added, since the `broker` stage now publishes from there. That is the one test outside the new module that changed, and it is a list entry, not an assertion.

## Unit tests, no process

`egress_test.exs` (17 tests, `async: false` because brokering is switched on per tenant through application env, like `broker_test`):

- every split rule is a no-op for an unbrokered tenant, and reads no bindings;
- `split_brokered/3`: a catalog key and a key with the tenant's own binding become placeholders, a plain one stays;
- `split_inference/4`: placeholders for the runtime, values for the broker, the tenant's own secret of the same name wins, an implicit binding only where the tenant has none;
- connections: a connection's token brokered with bearer bindings to the provider's hosts; the tenant's own secret and binding win; `connection_bindings/3` adds the agent's remote MCP hosts, deduplicated; `refresh_connection_secrets/3` swaps in a rotated token and says so; `with_connection_servers/4` drops both connection entries for an unbrokered tenant and resolves the served one (URL carries the conversation id) while keeping the remote one for a brokered tenant;
- the session: `prepare/4` publishes `broker/started` (keys) and `broker/done` (vault) around the facade call, and `broker/failed` with the reason; `reprepare/5` replaces the proxy variables and nothing else; `drop_oauth_token/3` forgets the token on both sides; `sandbox_env/1`; `install_ca/3` only with a session;
- `release/2` calls the facade off the caller's path for a brokered tenant only (Mimic global, since it runs in a supervised task);
- `apply_policy/4` picks the environment's policy or the broker floor.

`Fountain.Broker` and `Provisioning` are stubbed through Mimic where the test would otherwise reach the network; no `ConversationServer`, no sandbox stub, no peer.

## Pin

| | lines |
|---|---|
| before | 4168 |
| after | 3991 |

## Gates

| gate | result |
|---|---|
| `MIX_ENV=test mix compile --warnings-as-errors` | clean |
| `mix format --check-formatted` (mise Elixir 1.19.2) | exit 0 |
| `mix credo --strict` | 5816 mods/funs, no issues |
| `MIX_ENV=dev mix dialyzer` | passed, 0 errors |
| thirteen `conversation_server*_test.exs` + size test + `broker_test`, `broker_native_test`, `conversation_egress_test`, `audit_guardrail`, `caller_tools`, `provisioning_test`, `docs`, `connections` | 387 tests, 0 failures |
| `egress_test` + size test | 17 tests, 0 failures |
| gitleaks over the new module and tests | no leaks |
| full root suite | posted as a comment when it finishes |

`git diff --stat origin/main` over the thirteen server test files and the three pins named by the issue is empty.

**Proof:** with `origin/main`'s server file put back beside the new module, the size test fails: `... is 4168 lines, over the pin of 3991`. Restored; the committed file is 3,991.

## CHANGELOG

One entry under `[Unreleased] → ### Changed`.

Closes #1373. Part of #1369.

https://claude.ai/code/session_016mx2sij38JHM6gQu6PxiGG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -218,6 +218,17 @@ upgrade, is in
   Sixteen functions, same bodies, no behaviour change; the brokered
   placeholders reach the env through one argument, which is the seam the
   egress move (#1373) uses next. The pin drops from 4,348 to 4,168.
+- **Egress brokerage and the connection secrets have left
+  `ConversationServer`** (#1373, tracker #1369). `Fountain.Conversations.Egress`
+  holds everything ADR 0019 wired into a conversation: the split rules
+  (bindings, connection tokens, catalog keys, inference credentials), the
+  proxy session's mint, re-prepare and release, the CA install and the
+  network floor. It calls the `Fountain.Broker` facade and
+  `Fountain.Connections`; the server keeps the session and its placeholders
+  in state and three short wrappers that apply what comes back. Seventeen
+  functions, same bodies, no behaviour, stage or log change. The Agent Vault
+  client's deletion now touches one file beside the facade. The pin drops
+  from 4,168 to 3,991.
 
 ### Added
 
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -24,6 +24,7 @@ defmodule Fountain.Conversations.ConversationServer do
   alias Fountain.Conversations.{
     CallbackKey,
     Conversation,
+    Egress,
     HomeCheckpoint,
     Lifecycle,
     McpServers,
@@ -690,20 +691,20 @@ defmodule Fountain.Conversations.ConversationServer do
 
     case SpriteEnv.load_tenant_state(conv.user_id) do
       {:ok, dek, inference_creds} ->
-        bindings = broker_bindings(conv.user_id)
+        bindings = Egress.bindings(conv.user_id)
 
         {merged, bindings, connection_keys} =
-          add_connection_secrets(
+          Egress.add_connection_secrets(
             conv.user_id,
             SpriteEnv.merge_secrets(env, vault, dek),
             bindings,
             agent
           )
 
-        {secrets, brokered} = split_brokered(conv.user_id, merged, bindings)
+        {secrets, brokered} = Egress.split_brokered(conv.user_id, merged, bindings)
 
         {env_creds, brokered, bindings} =
-          split_inference(conv.user_id, inference_creds, brokered, bindings)
+          Egress.split_inference(conv.user_id, inference_creds, brokered, bindings)
 
         state =
           %{
@@ -837,7 +838,7 @@ defmodule Fountain.Conversations.ConversationServer do
              ),
            :ok <-
              Fountain.Conversations.Provisioning.check_broker_support(
-               brokered?(state),
+               Egress.brokered?(state.user_id),
                provider,
                env,
                state.conversation_id
@@ -872,7 +873,12 @@ defmodule Fountain.Conversations.ConversationServer do
                Provisioning.write_runtime_config(
                  handle,
                  state.runtime_module,
-                 with_connection_servers(agent, state)
+                 Egress.with_connection_servers(
+                   agent,
+                   state.user_id,
+                   state.conversation_id,
+                   state.callback_token
+                 )
                ),
              _ = Provisioning.write_instructions(handle, runtime, agent),
              # The file is the machine's; the conversation's identity travels as
@@ -882,7 +888,7 @@ defmodule Fountain.Conversations.ConversationServer do
                  handle,
                  Fountain.Conversations.Identity.disk_env(sprite_env)
                ),
-             :ok <- broker_install_ca(state, handle),
+             :ok <- Egress.install_ca(state.broker, handle, state.conversation_id),
              :ok <-
                run_provisioning_pipeline(
                  handle,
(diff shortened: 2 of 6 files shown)


jhgaylor:
Full root suite on 3dca9b32, for the record: the first run failed one unrelated test (TurnImageServeTest, an unauthenticated image request), which passes in isolation and passed in this commit's CI partitions; a second full run came back **6 doctests, 3879 tests, 0 failures** (348 s), plus managoat_runtimes 116 tests, 0 failures. Treating the first as a flake; it touches nothing this PR moved.

Issue #1377, opened 2026-09-02
ConversationServer: the connection and output persistence out; what the process keeps


jhgaylor:
Part of #1369, last. What remains after the six moves above: the process itself and the two families that are genuinely about it being a process.

**The connection (#817):** `connection_alive?`, `user_turn_running?`, `autonomous_turn?`, `resume_acp_connection`, `drop_connection`, `connection_lost`, `open_autonomous_turn`, `close_autonomous_turn`, `arm_autonomous_quiet`, `cancel_autonomous_quiet`, `autonomous_quiet_ms`, `stop_acp_peer`. This is where #1301's phantom-turn fix lives and where the peer outlives the turn. **Home:** `Fountain.Conversations.Connection`, a value with the peer pid, the quiet timer and the autonomous-turn flag.

**Output persistence:** `log_output`, `cap_marker`, `ensure_output_bytes`, `output_byte_budget`, `persist_output`, `log_with_replay_skip`, `publish_stage`, `write_image_temp_files`, `media_type_to_ext`. **Home:** `Fountain.Conversations.Output`: the byte budget, the replay-skip rule (memory: sprites replays the last 16 KiB mid-line; byte-count skip was wrong), the cap marker. `Redaction` already exists and stays the single writer's guard.

**What the GenServer keeps:** `init`, the provisioning watchdog and reattach orchestration (`provision_with_rows`, `dispatch_provision`, `fresh_provision`, `attempt_warm_start`, `reattach`, `reattach_running_turn`, `reap_orphan_sessions`, `attempt_session_attach`, `reattach_acp_peer`, `mark_orphan`, `find_running_turn`), the callbacks, `format_status` and `redact_state`. Expect the file to end near a quarter of its size. Set the final pin there and write in the tracker what it holds and why each remaining function is about the process.

Acceptance: every server test unchanged and green; `Connection` and `Output` unit-tested without a process; the tracker closed with the final size and a one-paragraph description of the server.

Issue #322, opened 2026-08-02
security: sprite callback tokens survive deploys — terminate/2 doesn't run on supervisor shutdown


jhgaylor:
`conversation_server.ex:129` — `init/1` never sets `Process.flag(:trap_exit, true)`. A `use GenServer` process only receives `terminate/2` on `{:stop, …}` or a raise; a Horde rebalance or application shutdown sends an exit signal and skips it entirely.

So the token revoke in `terminate/2` (`:990-1007`) does **not** fire on the most common teardown there is — every deploy. The moduledoc at `:981-989` states the opposite.

Backstops keep this at P2: rotation-on-reattach (`:322`, `:487`) revokes the prior key when the conversation resumes, and `expires_at` (default 30 days, `:1049`) eventually kills it. But a conversation that is never resumed leaves a live sprite-scoped tenant credential outstanding for up to 30 days.

**Related doc drift, same comment block:** it claims "RetentionPruner deletes the row later". `retention_pruner.ex:106-113` prunes only rows with `revoked_at` set, so an un-revoked orphan is never deleted — it just goes inert at expiry.

**Fix:** trap exits and revoke on shutdown (bounded, since `terminate/2` then runs under the shutdown timeout), or revoke from a supervisor-level hook. Then correct the moduledoc.

PR #1403, opened 2026-09-02
refactor(conversations): sandbox reclaim actions into Lifecycle

### LINKED ISSUE #1376: ConversationServer: sandbox reclaim actions into Lifecycle
Part of #1369. The sandbox's clock and what happens when it runs out, which belongs beside the policy that decides it.

**Functions:** `provider`, `sandbox_clock_start`, `schedule_lifecycle_check`, `touch_activity`, `reclaim_sandbox`, `home?`, `reclaim_idle_machine`, `suspend_sandbox`, `park_sandbox`, `stop_cotenants`, `destroy_sandbox`, `reclaim_message`, the lifecycle-check `handle_info`, and `terminate/2`'s sandbox half.

**Home:** `Fountain.Conversations.Lifecycle` already holds the policy (`check/…`, `idle_action/…`, `explain/…`, 177 lines) and none of the actions. Move the actions in as `Lifecycle.reclaim/…` and friends so a reader finds the decision and its consequence in one file. ADR 0017 (park, never destroy, while the disk is the agent's memory) and ADR 0023 (co-tenants on one sandbox) are the rules; the functions already implement them.

**Watch:** `conversation_server_lifetime_test.exs`, `conversation_server_shared_sandbox_test.exs`, the sandbox reaper tests, and `SandboxProviders` for the provider atom.

Acceptance: tests unchanged and green; `Lifecycle` action tests with `Managoat.Sandbox.Fake`; size pin lowered.

jhgaylor:
Part of #1369, seventh of eight. `Fountain.Conversations.Lifecycle` already held the policy — `check/4`, `idle_action/1`, `explain/1,2` and the two bounds they read — and none of the consequence. It now holds both, in a clearly separated second half, so a reader of `idle_action/1` can see what `:destroy` costs without opening another file. The policy functions are untouched and still pure.

## Functions moved

Two corrections to the issue:

- **`terminate/2` has no sandbox half.** It deletes the conversation's redaction registry and best-effort revokes the callback API key it minted (#322, and the long comment about a dying Horde duplicate revoking the survivor's credential). It touches no sandbox and no row. Nothing moved from it, and nothing was invented to match.
- **`touch_activity/1` stays on the server.** It writes one field of the server's own state (`last_activity_at`), which is the input `check/4` reads; moving it would be a function over `state`, which this refactor is specifically not making. It keeps a one-line comment saying what it is.

| was (`ConversationServer`) | now (`Lifecycle`) |
|---|---|
| `provider/1` (over state) | `provider/1` (over the handle) |
| `sandbox_clock_start/1` | `clock_start/1` |
| `schedule_lifecycle_check/0`, `@lifecycle_check_ms` | `schedule_check/0`, `@check_ms` |
| `home?/1` (over state) | `home?/1` (over the sandbox id) |
| the `_unsafe_sandbox_busy_elsewhere?` read in `reclaim_sandbox/2` | `busy_elsewhere?/2` |
| `suspend_sandbox/1` | `suspend/1` |
| `reclaim_idle_machine/1`'s `with` | `idle_machine_action/2` |
| `reclaim_sandbox(state, :max_lifetime)`'s `with` | `max_lifetime_action/2` |
| `park_sandbox/2`'s body | `park/4` |
| `destroy_sandbox/2`'s body | `destroy/5` |
| `stop_cotenants/4` | `stop_cotenants/5` |
| `reclaim_message/1` | `reclaim_message/1` |

The two `*_action` functions answer `:park` or `:destroy`, and make the suspend call themselves, because in the original the call *was* part of the decision (`with :suspend <- idle_action(...), :ok <- suspend_sandbox(...)`), and the error arm is where the "an unparked sandbox keeps billing" warning lives. Splitting the call out of the `with` would have changed which failures degrade to destroy.

One comment was mis-attached in the original and is now attached correctly: the four lines about the ceiling measuring a continuous run sat above `provider/1` but describe `sandbox_clock_start/1`, which was below it. No words changed.

## What the server keeps

Each reclaim is still a server callback returning a GenServer tuple, so the wrapper around each action stays: the `Logger.info`, then `drop_connection/2`, then the `Lifecycle` call, then `{:stop, :normal, %{state | handle: nil}}`.

The log line and `drop_connection/2` stay on the server on purpose, and in their original position. `drop_connection/2` is #1377's, and it can finish an autonomous turn — which persists rows and publishes stage events — so its order relative to the sandbox row update is observable. Keeping it where it was is the only way to leave that ordering alone. `destroy_sandbox/2` has the same shape.

`handle_info(:lifecycle_check, ...)` still owns the tick and the `Lifecycle.check/4` call; only `schedule_lifecycle_check()` became `Lifecycle.schedule_check()`, which arms the timer in the calling process exactly as before.

`lifecycle.ex` joined `turn_machine.ex` and `pending.ex` in the ownership check's exemptions (`.credo.exs`), with its own line: the sandbox and rows it reads belong to the conversation whose server called it, ownership established at that server's `init/1`. Each `_unsafe_` call site carries the comment saying so. `webhooks/events_test.exs` lists it as a `publish_stage` source.

## The reaper

`Workers.SandboxReaper.park/2` now duplicates `Lifecycle.idle_machine_action/2` almost exactly — the same `with`, the same degrade-to-expire on a failed call, differing only in its log prefix and in building the handle from `sandbox.sprite_name` rather than holding one. It could call `Lifecycle` directly after this move. **Not changed here**, per the tracker: the reaper is outside this PR's scope and its tests are a pin, not a target.

## Unit tests, no process

`lifecycle_actions_test.exs` (24 tests, `async: true`, `Fountain.DataCase` plus Mimic over the `Managoat.Sandbox` facade, the same shape as `provisioning_steps_test.exs` from #1372): `provider/1` and `clock_start/1`; `schedule_check/0` arming a live timer in the caller; `home?/1` for persistent, ephemeral and nil; `busy_elsewhere?/2` alone, with a co-tenant mid-turn, and with a terminated one; `suspend/1` with and without a handle; `idle_machine_action/2` parking, degrading where the provider cannot park (asserting no suspend call is made at all) and degrading on a failed call (asserting the warning); `max_lifetime_action/2` for an ephemeral sandbox, a home, and a home whose park call fails; `park/4` (row, conversation, stage, telemetry; the ceiling's copy; a `failed` row left alone; a non-running conversation left alone); `destroy/5` (the provider call, the terminated row with its stamp, the stage, the telemetry, and the no-handle case); `stop_cotenants/5` reaching a plain process registered under a co-tenant's id, and shrugging at a co-tenant with no server; `reclaim_message/1`. No `ConversationServerCase`, no `Sprites` stub, no `Peer`.

`lifecycle_test.exs` is unchanged — it pins the policy half, which this PR does not touch.

## Pin

| | lines |
|---|---|
| before | 3192 |
| after | 3049 |

## Gates

| gate | result |
|---|---|
| `MIX_ENV=test mix compile --warnings-as-errors` | clean |
| `mix format --check-formatted` (mise Elixir 1.19.2, from `apps/fountain`) | exit 0 |
| `mix credo --strict` | 5798 mods/funs, no issues |
| `MIX_ENV=dev mix dialyzer` | passed, `Total errors: 0` |
| the fourteen `conversation_server*_test.exs` plus `lifecycle_test`, `lifecycle_actions_test`, `sandbox_reaper_test`, `home_checkpoint_test`, `pending_test`, `turn_machine_test`, `audit_guardrail`, `caller_tools`, `provisioning`, `provisioning_steps`, `docs`, `webhooks/events`, `metrics` | 450 tests, 0 failures |
| `gitleaks git --log-opts="origin/main..HEAD"` | no leaks found |
| full root suite | posted as a comment when it finishes |

`git diff --stat origin/main` over the thirteen server test files is empty; the only test file this PR edits under that glob is the ratchet.

**Proof:** with `origin/main`'s server file put back beside the new `Lifecycle`, the size test fails — `apps/fountain/lib/fountain/conversations/conversation_server.ex is 3192 lines, over the pin of 3049`. Restored; the committed file is 3,049.

## CHANGELOG

One entry under `[Unreleased]`, `### Changed`.

Closes #1376. Part of #1369.

https://claude.ai/code/session_016mx2sij38JHM6gQu6PxiGG

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -89,6 +89,9 @@
                # what a turn waits on (#1375): the same, writing the same turn
                # row the server owns, called only from its callbacks
                "lib/fountain/conversations/pending.ex",
+               # the reclaim actions (#1376): the sandbox and rows of the
+               # conversation whose server called them, ownership as above
+               "lib/fountain/conversations/lifecycle.ex",
                # system-level sweeps, run with no user in scope
                "lib/fountain/conversations/rehydrator.ex",
                "lib/fountain/workers/",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -269,6 +269,16 @@ upgrade, is in
   call into it; the GenServer messages, the stage events and the audit of a
   denial are unchanged, and `Fountain.CallerTools` still owns the wire
   shapes. The pin drops from 3,322 to 3,192.
+- **The sandbox reclaim actions have joined the policy that decides them**
+  (#1376, tracker #1369). `Fountain.Conversations.Lifecycle` already held
+  the bounds (`check/4`, `idle_action/1`, `explain/1,2`); it now also holds
+  the consequence — the sandbox clock, the lifecycle tick, the two verdicts
+  with their suspend call, the park, the destroy and the cast that tells the
+  co-tenants on the machine. A reader of `idle_action/1` can see what
+  `:destroy` costs without opening another file. The server keeps the log
+  line, the connection drop and the `{:stop, :normal, …}` around each. No
+  stage, log line, telemetry event, timer or audit event changed. The pin
+  drops from 3,192 to 3,049.
 
 ### Added
 
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -25,7 +25,6 @@ defmodule Fountain.Conversations.ConversationServer do
     CallbackKey,
     Conversation,
     Egress,
-    HomeCheckpoint,
     Lifecycle,
     McpServers,
     Pending,
@@ -33,11 +32,6 @@ defmodule Fountain.Conversations.ConversationServer do
     SpriteEnv,
     TurnMachine
   }
-
-  # How often the sandbox lifetime bounds are evaluated. A minute is far finer
-  # than the bounds themselves (an hour, a day), so the cost of the tick is
-  # irrelevant and the overshoot is bounded by it.
-  @lifecycle_check_ms :timer.minutes(1)
 
   # Absolute ceiling on provisioning (#329). Generous against the summed
   # per-step timeouts (packages 300s + clone 600s + setup 120s + slack), so
@@ -555,7 +549,7 @@ defmodule Fountain.Conversations.ConversationServer do
       output_capped: false
     }
 
-    schedule_lifecycle_check()
+    Lifecycle.schedule_check()
     start_provision_watchdog(state.conversation_id, state.sandbox_id)
     {:ok, state, {:continue, :provision}}
   end
@@ -924,7 +918,7 @@ defmodule Fountain.Conversations.ConversationServer do
             state
             | handle: handle,
               sprite_env: sprite_env,
-              sandbox_started_at: sandbox_clock_start(sandbox)
+              sandbox_started_at: Lifecycle.clock_start(sandbox)
           }
 
           # Any prompt this conversation was started for arrives as a cast,
@@ -1187,7 +1181,7 @@ defmodule Fountain.Conversations.ConversationServer do
         state
         | handle: handle,
           sprite_env: sprite_env,
-          sandbox_started_at: sandbox_clock_start(sandbox)
+          sandbox_started_at: Lifecycle.clock_start(sandbox)
       }
 
       new_state = reattach_running_turn(new_state)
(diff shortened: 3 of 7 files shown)


jhgaylor:
Full root suite on the branch: **6 doctests, 3983 tests, 0 failures** in 268.4s. No flakes this run.

Local gates on the pushed commit: compile with `--warnings-as-errors` clean,
`mix format --check-formatted` exit 0 (mise Elixir 1.19.2, from `apps/fountain`),
`mix credo --strict` 5798 mods/funs no issues, `MIX_ENV=dev mix dialyzer`
`Total errors: 0`, `gitleaks git --log-opts="origin/main..HEAD"` no leaks found,
and the pin batch (the fourteen `conversation_server*_test.exs` plus
`lifecycle_test`, `lifecycle_actions_test`, `sandbox_reaper_test`,
`home_checkpoint_test`, `pending_test`, `turn_machine_test`,
`audit_guardrail`, `caller_tools`, `provisioning`, `provisioning_steps`,
`docs`, `webhooks/events`, `metrics`) **450 tests, 0 failures**.

Issue #1412, opened 2026-09-02
SDKs: add a shared cross-language behavioral conformance suite


jhgaylor:
## Problem

OpenAPI can keep request and response schemas aligned, but it does not describe
the behavioral parts of Fountain's SDK contract: SSE framing and reconnects,
cursor resume, terminal run states, permission replies, pagination, error
mapping, cancellation, authentication headers and retry behavior.

Those behaviors are currently exercised by language-specific fake servers and
tests. With TypeScript, Elixir, Swift and Python clients, the same behavior can
quietly diverge four ways even when every client agrees with the schema.

## Proposed shape

Add a shared, language-neutral SDK conformance suite under `sdk/conformance/`.
It should define wire-level scenarios and expected observations using fixtures
or a small reference test server. Each SDK supplies a thin adapter that runs
the same scenarios through its public client.

Keep language-specific tests for idioms and implementation details. The shared
suite owns only behavior that Fountain promises across SDKs.

## Initial contract set

- Authentication, base URL normalization and identifying User-Agent.
- Successful JSON responses and Fountain error-envelope mapping.
- SSE comments, multiline data, fragmented frames and malformed events.
- Reconnect budget, `Last-Event-ID`/cursor resume and clean terminal closure.
- Run completion, failure, cancellation and timeout semantics.
- Permission request/reply flow.
- Pagination and empty-page behavior for shared resource operations.

## Acceptance criteria

- [ ] Canonical scenarios and fixtures live in `sdk/conformance/` and contain
      no language-specific expectations.
- [ ] TypeScript, Elixir, Swift and Python run those scenarios in their normal
      CI jobs.
- [ ] The suite verifies observable public behavior, not private class or
      module structure.
- [ ] A support matrix shows which shared scenarios each SDK implements; gaps
      must be explicit and linked to an issue.
- [ ] Adding or changing a cross-SDK behavior has one documented place to add a
      fixture/scenario before updating the implementations.
- [ ] Failures identify the SDK and scenario without requiring inspection of
      another job's logs.
- [ ] The incoming Python SDK is added after its PR lands; this work does not
      block that merge.

## Non-goals

- Eliminating language-specific unit tests.
- Requiring identical class names, return types or async abstractions.
- Exercising every server endpoint through a live production-like Fountain
  instance.
- Absorbing the Swift `FoundationNetworking` teardown bug tracked separately.

PR #1417, opened 2026-09-03
feat(sdk): make OpenAPI the canonical wire contract for every client

### LINKED ISSUE #1411: SDKs: make OpenAPI the canonical wire contract for every client
## Problem

Fountain now has four SDK implementations: TypeScript, Elixir, Swift and the
incoming Python client. The server already emits the canonical OpenAPI 3.1
document, and CI proves that the TypeScript generated types match it, but the
other clients maintain more of their wire contract by hand.

That leaves one API represented in several places. A schema change can merge
with a correct TypeScript update while an equally affected field, enum, error
or endpoint remains stale in another SDK.

Keeping the SDKs in this repository gives us an opportunity to make that drift
fail in the same PR that changes the server contract.

## Proposed shape

Treat the OpenAPI document produced by `FountainWeb.ApiSpec` as the canonical
wire contract for every SDK. Each SDK should have a deterministic command that
either generates its wire types/client surface from that document or verifies
its hand-written wire representation against an OpenAPI-derived projection.

Generated wire types do not need to become the public API. Language-idiomatic,
hand-written conveniences such as `run`, conversation followers and async
streams can remain a layer above them.

## Acceptance criteria

- [ ] CI produces one deterministic, vendor-extension-free OpenAPI artifact for
      SDK contract checks.
- [ ] TypeScript, Elixir, Swift and Python each expose a documented
      `generate` or `verify-contract` command using that artifact.
- [ ] A changed field name, requiredness, enum, response envelope or endpoint
      causes every affected SDK check to fail until it is regenerated or
      deliberately updated.
- [ ] Generation is deterministic: running it twice from a clean checkout
      produces no diff after the first run.
- [ ] Each SDK clearly separates generated/verified wire types from its
      hand-written ergonomic API.
- [ ] Intentional omissions are recorded in a small checked allowlist rather
      than silently ignored.
- [ ] The SDK contributor docs explain how to update all affected clients in an
      API-changing PR.
- [ ] The incoming Python SDK is included once its PR lands; this work does not
      block or require reopening that PR.

## Non-goals

- Giving every language the same public API shape.
- Forcing SDK versions to move in lockstep.
- Publishing every SDK for every server-only change.
- Replacing the higher-level run, streaming and permission helpers with raw
  generated clients.

jhgaylor:
Closes #1411

One API, four clients. The server's OpenAPI document is now the wire contract
for all of them, and a schema change fails every affected SDK check in the same
PR rather than months later in somebody's application.

## The artifacts

| File | Built by | Committed |
|---|---|---|
| `dist/openapi.json` | `mix openapi.export`, canonicalised by `scripts/sdk-contract/build.py` | no, `dist/` is ignored |
| `sdk/contract/contract.json` | the same script | **yes** |

`contract.json` is the document projected down to shape alone: operations,
schemas, requiredness, enums, formats, nullability. Descriptions, summaries,
tags and `info.version` are dropped, so a docs pass or a release bump leaves it
byte-identical and a diff on it means exactly one thing. It is committed so an
SDK check needs no Elixir toolchain, which is what lets the Swift job on macOS
check the contract straight from the checkout.

**Determinism** is not left to the encoder. `build.py` sorts every object key
recursively before writing either file, so the output depends on the document
and nothing else. Verified three ways: two consecutive runs write identical
bytes; Elixir 1.19.2 and 1.19.5 produce identical output; `MIX_ENV=test` and
dev produce identical output. Sorting the artifact is why
`sdk/typescript/src/generated/openapi.ts` reorders wholesale in this PR, once.

## The four commands

| SDK | Command | Run from |
|---|---|---|
| TypeScript | `npm run verify-contract` (and `npm run generate`) | `sdk/typescript` |
| Python | `python3 scripts/verify_contract.py` | `sdk/python` |
| Swift | `swift test --filter ContractTests` | repository root |
| Elixir | `mix test test/contract_test.exs` | `sdk/elixir` |

TypeScript keeps generating types for the **whole** document and failing on a
diff. That is the complete arm. The other three verify a checked-in manifest
(`sdk/contract/manifests/<sdk>.json`) declaring the operations, fields,
requiredness and enum values that client depends on, which is the part no type
check can pin because it lives in template strings. Python and Elixir also run
their check inside their existing suites; Swift's runs inside `swift test`.
Generated and hand-written stay separate in every case, and no SDK's public API
shape changed.

## Proof that it bites

I mutated the contract the way a server change would, and ran all four:

| Scenario | TS | Python | Swift | Elixir |
|---|---|---|---|---|
| `Turn.turn_number` renamed | fail | fail | fail | fail |
| `Block.kind` no longer required | fail | pass | fail | fail |
| `LogEvent.state` drops `interrupted` | fail | fail | fail | fail |
| `ConversationResponse.data` renamed | fail | fail | fail | fail |
| `GET /api/search` removed | fail | fail | fail | fail |

Each failure names the SDK and the scenario in its own log, with the schema's
actual properties listed. Python passes the second row on purpose: it reads
`kind` with `.get()`, so it declares the field rather than the requirement.
That is the "every **affected** SDK" property working, not a gap.

Separately confirmed: a new endpoint that no manifest claims and no allowlist
entry covers fails `build.sh --check`; an allowlist entry matching nothing the
API still serves fails too.

## The omissions allowlist

`sdk/contract/omissions.json`, 17 grouped entries covering the 88 operations no
client's hand-written layer models, one reason each: the admin surface, account
bootstrap and key minting, the OAuth token endpoints, account self-service,
webhooks, secret bindings, runner registration, the Buzz provisioning surface,
`POST /api/apply`, the AG-UI stream, the OpenAI-compatible `/v1` gateway, the
audit and support reads, avatars, agent version history, a handful of
console-facing reads, the health probes, and the `PUT` aliases of the `PATCH`
routes. It has teeth in both directions, which is why it cannot rot.

## A defect this found

`AuthMeResponse` declared `required: [:id, :name, :prefix, :created_at]` --
three properties copied from `ApiKey` that it does not have.
`openapi-typescript` drops a required name with no property silently, so the
generated client had **every** field on that response optional. Corrected to
the four `show/2` always renders.

**This is the one place wire types actually changed**: `id`, `email`, `role`
and `email_verified` are now non-optional on `AuthMe` in the TypeScript client.
It is a widening of guarantees on a response, so no consumer breaks, and the
TypeScript SDK is bumped to **1.16.0** for it -- merging this publishes that.
No other SDK is bumped; nothing they ship changed shape.

## The `default` trap

An OpenAPI property with a `default` is not required. The projection takes
requiredness from the `required` list alone, records `has_default` beside it,
and `build.py` refuses to write a contract where the two were conflated. The
TypeScript manifest declares `TeamScheduleCreateRequest.enabled` and `one_off`
as `optional`, so the day the server makes one required, the client that omits
them is told.

## Verification

- `mix precommit`: 6 doctests, 3983 tests, 1 failure -- a `DBConnection`
  pool-checkout timeout in `conversation_controller_test.exs` under four
  agents sharing one Postgres. That file re-run alone: 73 tests, 0 failures.
  Compile with warnings-as-errors, unused deps, format, `credo --strict`,
  dialyzer (`Total errors: 0`) and sobelow all passed.
- `mix openapi.spec.json | jq empty`: valid.
- TypeScript: typecheck, 77 tests, build, browser bundle, generated types current.
- Python: contract check, 17 tests, `compileall`.
- Elixir SDK: format, warnings-as-errors compile, 28 tests.
- Swift: strict format lint, 21 tests, release build, all under warnings-as-errors.
- Both SDK release guards run locally against `origin/main`.

## Notes for review

- `sdk/contract/README.md` is the spec for the manifest format and what each
  check means, including what this deliberately does not catch.
- `CONTRIBUTING.md` has a new "Changing the API" section: the order to run
  things in when a PR moves a schema.
- The three ports (Python, Swift, Elixir) were written by codex from the
  TypeScript reference, one run per language. Every file was read here, every
  claim in every manifest was traced back to a line of that client, and every
  check was re-run independently. Two things needed fixing afterwards, both
  mine rather than codex's: `--check` could never pass on a freshly exported
  spec, and a `mix contract.verify` alias tripped the Elixir release gate.
- `release.yml` now attaches the canonical document, so the artifact on a tag
  is byte-identical to the one the contract was built from.
- **#1412 (the cross-language behavioural conformance suite) is not in here.**

https://claude.ai/code/session_014BXYa9CQNG8oFNq5YhAxva


jhgaylor:
Labelled `sdk-no-release`. Both release gates fired on files that do not change any published behaviour:

- **Elixir** — `sdk/elixir/mix.exs` gained the `contract.verify` alias. Tooling, not surface.
- **TypeScript** — `sdk/typescript/src/generated/openapi.ts` regenerated. I checked whether it was a pure reorder and it is not: three fields on `AuthMeResponse` flip from optional to required (`email`, `email_verified`, `role`).

That flip is the fix, not a regression. `AuthMeResponse.required` read `[:id, :name, :prefix, :created_at]` — copied from `ApiKey`, naming three properties `AuthMeResponse` does not have. `openapi-typescript` drops a required name with no matching property, so the published client had **every** field on that type optional. The contract projection this PR adds refuses to build such a schema, which is how a bug that has been in the spec since the schema was written surfaced on the gate's first run.

The replacement list is verified against the renderer: `FountainWeb.AuthMeController.show/2` unconditionally renders `id`, `email`, `role` and `email_verified` on every response. Consumers get types that are strictly more accurate — no null-check needed on a field the server always sends — so this is a no-release change that the next SDK release carries.

One adjacent inaccuracy left alone deliberately: `expires_at` is declared on `AuthMeResponse` but `show/2` never renders it. Out of scope here; worth its own issue.

jhgaylor:
Correction to my comment above, for anyone reading this merged PR later.

I labelled this `sdk-no-release` on the assumption the TypeScript change was types-only. The author had independently reached the better answer and fixed both gates properly instead: **`sdk/typescript` bumped to 1.16.0**, and the `mix contract.verify` alias dropped from `sdk/elixir/mix.exs` in favour of the plain `mix test test/contract_test.exs`. That alias is shipped in the hex package, which is why the Elixir gate fired at all. Their fix is what actually made the gates pass; my label was redundant.

Their reasoning is worth recording, because it is a trap in the gate design: **`sdk-no-release` switches off *both* release gates**, TypeScript and Elixir. Using it to wave through an Elixir tooling change would silently have waved through the TypeScript publish decision too.

Merging did publish. `@agentshit/fountain-sdk@1.16.0` went out and `Publish SDK` on main is green. So `AuthMeResponse` now types `id`, `email`, `role` and `email_verified` as non-optional in a released client, rather than waiting for a later release as I wrote above. That is a widening of guarantees and breaks no consumer.

The label is left in place as merged history. It changed nothing: the gates it suppresses run on `pull_request`, while publishing runs on `push` to main.

Issue #1501, opened 2026-09-03
Broker parity: what Fountain does as each managoat_broker gap closes


jhgaylor:
Companion to **managoat/managoat_broker#5**, the tracker for closing the remaining Agent Vault parity gaps in the library. That issue is the proxy work. This one is what Fountain has to do as each row lands, so the library PRs do not have to carry Fountain's context and so nothing arrives here with no home.

Background: #1359 was the pre-flip ledger, #1488 the flip, ADR 0019 the decision.

### Library row 1 — path and query substitution

The capability Fountain gains: brokering a credential that lives in a URL rather than a header. A bot API with the token in the path is the shape (`/bot<token>/sendMessage`); Agent Vault shipped a preset for it.

Fountain-side work:

1. **`SecretBindings.Binding.@auth_types` is `~w(substitute bearer basic api_key custom)`.** Decide whether a URL-placed credential is a new `auth_type`, or whether `substitute` simply reaches further once the library does. Prefer the second if the library makes the placeholder position implicit — a tenant should not have to say *where* their placeholder is, only that it is one.
2. **`Fountain.Broker.rules_for/3`** maps bindings to `Managoat.Broker.Rule`s; whatever shape lands has to be built there.
3. **The moduledoc paragraph "How a value is attached"** in `Fountain.Broker` states the deviation as a fact ("Agent Vault substitutes in the path, query, headers and body; the native proxy in header values only"). It stops being true and must change in the same PR.
4. **ADR 0019** — check whether any section states the header-only scope. If so, amend rather than silently diverge.

**The one that could actually hurt.** `broker_requests.path` is a stored column, and `Native.handle_request/4` prints the path in a log line. Today a path is never a secret. The moment the library can substitute into a target, **a stored path is a credential**.

The library issue asks for redaction at the point of emission, which is the right place, but Fountain should not rely on that alone:

- Confirm on the first release that the emitted `path` carries the placeholder, not the value, with a test in `broker_native_test.exs` — not just a reading of the library changelog.
- The audit rule is already "never record values" (`decisions/0013-audit-trail.md`), and this would violate it through a column nobody thought of as sensitive.
- Consider whether existing rows need anything. They do not today, since nothing can put a credential there yet, but the ordering matters: **do not deploy a library that can substitute into paths before the redaction test is green here.**

### Library row 2 — response framing, for status and latency

[...]

Issue #1169, opened 2026-08-25
Credits and billing have no alerts: a stuck pricer means free compute with no signal


jhgaylor:
## Gap

Under ADR 0031 the balance is the gate, so the workers that keep it honest are load-bearing, and none of them is watched:

- `CreditPricer` / `CreditExpirer` are covered only by `FountainObanJobsRaising`, which needs them to *raise*. A pricer that runs and prices nothing (bad rate config, an empty `SandboxUsage`, a query that silently matches zero rows) never trips it.
- `Credits.Rent` likewise.
- Stripe webhook handling (`checkout.session.completed`, `charge.refunded`, `charge.dispute.created`) has no failure counter or alert; a rejected signature or a 500 on the webhook endpoint is only visible as part of the global 5xx rate.
- Outbound email (Resend) has no delivery metric.

## Proposed

1. Emit telemetry from the pricer: turns priced, cents debited, and a "last successful run" timestamp gauge per worker (pricer, expirer, rent). Same shape as `fountain_reaper_run_*`.
2. Rules:
   - `FountainCreditPricerSilent`: `time() - fountain_credit_pricer_last_run_seconds > 2h` (or `absent()` with a `for` longer than the schedule, like `FountainReaperSilent`)
   - `FountainCreditPricerPricedNothing`: priced turns == 0 over 24h while `fountain_turn_completed` > 0 on platform-paid providers
   - `FountainStripeWebhookFailures`: webhook failures > 0 over 1h
   - `FountainEmailDeliveryFailures`: Resend failures > 0 over 1h
3. Fold the metrics into the finance Grafana dashboard.

The per-replica gauge trap applies: the "last run" gauge only exists on the pod that ran the Oban job, so use `max`.

Found during the 2026-08-25 prod alert-coverage review.

jhgaylor:
**Steps 1 and 2 are done and live. Step 3 is not — leaving this open for it.**

**Step 1, telemetry** — #1513 (`102800c0`). Two events rather than the one the issue sketched, because there are two questions and a worker can answer the first healthily while failing the second:

- `[:fountain, :credits, :worker, :run]` — a wall-clock `last_run_unix` per worker (pricer / expirer / rent), so a rule catches both staleness and a worker that never fires. **Per-replica**, as predicted: the stamp exists only on the pod that ran the Oban job, so every rule over it uses `max`.
[...]

PR #1531, opened 2026-09-04
feat(broker): take managoat_broker 0.11.0 and close the Agent Vault parity gaps

jhgaylor:
The Fountain half of #1501 / managoat/managoat_broker#5. That library tracker is **closed** — every row worth closing shipped between 0.1.3 and 0.11.0 — and Fountain was pinned at `~> 0.1.0`, so it could reach none of it.

> **Stacked on #1527** (`fix/1501-broker-query-leak`, row 0, already green). Merge that first; this PR's base is its branch, so the diff here is only the parity work. If you would rather have one PR, say so and I will collapse them.

The pin moves to `~> 0.11.0`, not `~> 0.11`. Pre-1.0 this library makes breaking changes in a minor bump, so a minor should reach Fountain when someone edits that line, never by resolution.

## What the broker can now do

| Row | Gap | What is here |
|---|---|---|
| 1 | No path/query substitution | `:substitute` reaches the request target. `/bot<token>/sendMessage` and `?key=` both work, over CONNECT and absolute-form. |
| 2 | Responses never framed | `broker_requests.status`, `.latency_ms`, `.error` are written from the now-terminal event. |
| 3 | IPv4-only upstreams | `allowed_hosts` brackets a bare IPv6 literal before it becomes a rule pattern. |
| 4 | Plain HTTP one request per connection | Shipped in 0.10.0. Two consequences absorbed and documented. |
| 5, 7 | WebSocket frames, unchecked label | Recorded as deliberate in ADR 0019. |
| 6 | No auth-failure rate limiting | No longer settled — managoat/managoat_broker#22. Recorded as an open **library** decision. |

**Row 1 decision.** Deliberately **no new `auth_type`**. A tenant declares that a key has a placeholder; the proxy finds it wherever the client put it. `SecretBindings.Binding.@auth_types` is unchanged, which is what the issue recommended.

**Row 2 timing, worth knowing before it surprises someone.** A row now appears when the response **ends**, so a streamed reply reaches the log when the stream finishes, and `latency_ms` is total duration rather than time to first byte. That is Agent Vault's semantics, not a compromise.

**Row 4.** `[:managoat, :broker, :connect]` counts **origin** connections now, not sandbox ones. I checked home-cloud's `FountainBrokerUpstreamFailures`: it divides that event by itself (`upstream_failed` over `ok|upstream_failed`), so the ratio is unchanged and only the absolute rate on the plain path moved. **No home-cloud change needed.** The session store is also resolved per request on that path — one indexed lookup and one AES open per `apt` request; a tunnel is unchanged at one.

## One correction, which was Fountain's bug and not the library's

`outcome` answers "did a rule apply", and a matched `:passthrough` rule — the documented way a host is allowed under `limited` — applies without attaching anything, so it arrives as `:injected` exactly as a `:bearer` rule does. `broker_requests` derived its verdict from `outcome` alone, so **an allowed host's row named an `ALLOW` rule in `service` and read as though a credential had gone to it**. The 0.11.0 event carries `scheme`; the row and the `fountain.broker.request.count` series now both mean "was a credential attached".

Found building managoat/airlock's egress record (managoat/managoat_broker#27), not here — the argument for the same event having more than one consumer.

## New signal

`fountain.broker.upstream.count{status_class="2xx|3xx|4xx|5xx|none"}`, the alert the issue's row 2 said becomes possible. Different question from `connect.count`: that one counts connections that never reached an origin, this one counts **origins that answered badly**. By class rather than by code, so the label set stays bounded.

The home-cloud rule is a follow-up in that repo, once the metric is live in prod — writing a rule against a series that does not exist yet is how #1169's alerts went wrong.

## Tests

- `Fountain.BrokerProxyRig` gains the substitution case: a placeholder in a path *and* in a query reaches the origin as the real credential over a real tunnel, while the stored row holds the placeholder for the path, nothing for the query, and the credential bytes nowhere. That is the assertion #1501 asks for **before** deploying a library that can substitute into paths.
- A guard that every placeholder Fountain mints passes `Injector.valid_placeholder?/1` (0.5.0 refuses short ones).
- Scrape tests for both counters, including that an allowed host does not land under `outcome="injected"`. Verified by reverting: removing `tag_values` fails it with `outcome=passthrough is missing from the scrape`.
- The rig now **waits on the terminal event** rather than flushing straight away. The library relays every response byte to the client before showing it to the framer, so a test that has read its whole response has not necessarily seen the row yet — this raced about one run in twenty before the wait. Eight consecutive random-seed runs of the file are clean.

## Docs, ADR, wire

- `docs/concepts/secrets.md` claimed substitution reached the **body**. It never did on this proxy and still does not; the page now says header, query or path.
- ADR 0019 gains a parity amendment: every row, the three deviations kept on purpose, and the one that is no longer settled. Two now-false sentences in the existing text are corrected in place rather than left to diverge. `okf validate decisions` reports 0 errors; the index does not move (no frontmatter change).
- `EgressEvent`'s OpenAPI descriptions were wrong about `error` (`no_match` is not an error code) and silent about the timing. Regenerated TypeScript types are **comment-only** — the fields were already nullable and not required, so no typed client breaks. `contract.json` does not move because it drops descriptions. The endpoint is a documented SDK omission for the other three clients.

## Behaviour changes absorbed from 0.2.0 → 0.11.0

New `broker_requests.error` values a reader may now see: `credential_missing` (502, 0.6.2), `request_timeout` (0.8.0, five-minute default on reading one request), `request_too_large` / `response_too_large` (0.6.0). Rule matching is by specificity rather than declaration order (0.7.0) — `rules_for/3` writes no generic default followed by an override, so nothing here resolves differently; the moduledoc says so for whoever adds one. Body caps and the read timeout are left at the library's defaults, which are Agent Vault's own numbers.

## Verified

`mix precommit` green: dialyzer 0 errors, sobelow, prod release assemble, **4220 tests / 0 failures**. `docs-style.py`, `vale lint docs` (0 errors, 0 warnings) and `destink` all clean.

Not closing #1501 — the issue's own rule is to verify against the deployed image rather than the merge, so it stays open until this is in prod.

https://claude.ai/code/session_01XH3dbY45N3E1kVZws9Earp


jhgaylor:
Labelled `sdk-no-release`. The only SDK change here is `sdk/typescript/src/generated/openapi.ts`, regenerated after correcting three `EgressEvent` descriptions in the Elixir schema — `error` cited `no_match`, which is not one of the library's error atoms, and nothing said when the row is written. **The diff is comment-only**: every field keeps the type and optionality it had, so no client's compiled output moves. The corrected JSDoc rides along on whatever bump comes next rather than racing a publish for tooltips.

jhgaylor:
Green now. One extra commit came out of getting there, and it is unrelated to the broker.

The `sdk-no-release` label could not work. The gate's `if:` reads `github.event.pull_request.labels.*.name` off the **event payload**, which is frozen when the event fires, and `sdk-release-gate.yml` listed no `types:` — so it defaulted to `opened, synchronize, reopened`, adding the label fired nothing, and `gh run rerun` replayed the same stale payload. The only way past a gate whose own error message says "add the sdk-no-release label" was to push another commit.

`elixir-sdk-release-gate.yml` has carried `types: [opened, synchronize, reopened, labeled, unlabeled]` since it was written, so 2ccdc784 is that file's own convention rather than a new idea. The Python gate has no label hatch at all, so it needs nothing.

Issue #1538, opened 2026-09-04
Tracker: finish Buzz extension packaging and repository graduation


jhgaylor:
Follow-on to #1503 for the work left after the server-side Buzz extraction landed in `main`.

The extension contract, authenticated route dispatch, migration/OpenAPI composition, and the `apps/fountain_buzz` OTP application are in place. `fountain_support` is now a second consumer of the same contract. This tracker covers the correctness, packaging, distribution, and repository-graduation work that remains.

Campaign label: `buzz-extension-2026-09`.

## Workstream

- [x] #1536 — schema enforcement traverses extension dispatch and validates both Buzz and Support responses. Closed by #1540; both extensions declare their statuses rather than joining the allowlist.
- [x] #1508 — `buzz-backend-fountain` and the remote-agents provider moved to the extension; `fountain buzz ...` stayed in the main binary as ADR 0043's thin HTTP client. Closed by #1546.
- [x] #1509 — the extension owns its `buzz-acp`/`buzz` supply chain and image layer, and the bundle flag generalised past `BUNDLE_BUZZ` to `BUNDLE_EXTENSIONS`. Closed by #1534.
- [x] #1525 — core copy declares the extension it needs, and the core image asserts it. Closed by #1552.
- [x] #1510 — core distribution boot-tested against an empty database, core images published beside the bundled ones, and the manual sliced per extension. Closed by #1541, #1545 and #1548, **except the repository split**, deferred by choice to #1550.

## Recommended order

```text
#1536 ───────────────────────────────┐
                                     ├─> #1510 graduation
#1508 ────────────────┐              │
                     ├─> provider +  │
#1509 / PR #1534 ─────┘   packaging ─┤
                                     │
#1525 ─> core/bundled docs decision ─┘
```

`#1536`, the `#1508` scope correction, and the `#1509` rebase can proceed independently. Resolve `#1525` while the actual core image is available to test. Do not split to a standalone repository until extension responses are covered by the schema guard.

## Current integration hazards

*Resolved. Kept as the record of what each gate had to navigate; every hazard
below was handled by the PR that closed its item.*

- PR #1534 predates the `fountain_support` landing. Its conflicts in release composition, runtime extension configuration, Docker assembly, and ADR text are semantic rather than purely mechanical: a core/bundled switch must say whether it excludes Buzz only or every optional extension.
- #1508's original body proposes moving `fountain buzz ...`; accepted ADR 0043 instead keeps that thin compatibility client in Fountain core.
[...]

Issue #1528, opened 2026-09-04
fountain_support: move problem reports into an umbrella extension app


jhgaylor:
Follow-up to ADR 0043 and the extension seam exercised by #1523.

The problem-reporting feature is already a vertical slice: a tenant-scoped context and schema, one migration, one authenticated API subtree, OpenAPI schemas, and a worker enqueued only by that feature. Move it into an umbrella sibling application, `apps/fountain_support`, with modules under `FountainSupport.*`.

This should prove the extension seam against a second feature without widening it. `apps/fountain` must not depend on `:fountain_support` or name a `FountainSupport.*` module; the distribution owns whether the extension is compiled in and enabled.

## Move

- `Fountain.Support` and `Fountain.Support.Report`.
- `Fountain.Workers.SupportForward`.
- The support report controller, JSON rendering, and OpenAPI schemas.
- `20260819130000_create_support_reports.exs`, keeping the version unchanged.
- Support-specific configuration for email and GitHub forwarding.
- Dedicated tests and test support.

`FountainSupport.Extension` should use only the existing callbacks:

- `api_mounts/0` for `/support`.
- `migrations/0` for the extension migration directory.
- `openapi_paths/0` for the three report operations.

The worker is enqueued by the extension's own context, so this needs neither an Oban cron entry nor another host callback.

## Preserve exactly

- `POST /api/support/reports`, `GET /api/support/reports`, and `GET /api/support/reports/:id`, including authentication, tenant scoping, response bodies, statuses, and error shapes.
- Published OpenAPI paths, operation IDs, and schema titles. Pin the existing operation IDs if the controller rename would otherwise change them; the generated SDK contract should be byte-identical.
- Screenshot size/type validation and the rule that screenshot bytes and report text do not enter the audit trail.
- `support.report.created`, its resource identity, and its metadata shape.
- Asynchronous forwarding to `SUPPORT_EMAIL` and/or GitHub, including partial-success, retry, and persisted failure behavior.
- The current no-target behavior: the report remains stored and the worker completes without making the user's request fail.
- Existing rows and applied migration history across bundled-to-core-to-bundled image changes.

## Boundary

- The extension may consume Fountain's public `Repo`, `Audit`, `Accounts`, `Mailer`, `Apps`, image-decoding, and Oban facilities.
- Core owns the authenticated API pipeline and shared Oban instance; the extension owns all report-specific policy and data.
- Do not add a tenth `Fountain.Extension` callback for this extraction.
[...]

Issue #1432, opened 2026-09-03
68 operations do not declare the statuses their pipeline plugs return


jhgaylor:
Found by the schema guard (#1427), and the largest family it reports.

## The gap

The OpenAPI document describes what controllers return. Several statuses come
from plugs in the pipeline instead, and most operations do not declare them:

| Status | Source | Operations missing it |
|---|---|---|
| 401 | `Plugs.TenantAPIAuth` | 38 |
| 403 | `require_admin` and the ownership checks | 6 |
| 422 | `OpenApiSpex.Plug.CastAndValidate` | 4 |
| 503 | an unready sandbox, `fleet_full` | 4 |
| 429 | `Plugs.RateLimit` | 3 |
| 404, 409, 400, 402, 406, 410, 424, 500 | assorted | 13 |

The full list is in `FountainWeb.SchemaGuardAllowlist` under `:plug_status`.

## Why it is worth closing rather than waving through

A generated client cannot type a response it was never told about. It also
distorts the SDK contract work: `sdk/conformance/lint.py` has to special-case
"a plug can return this on any route" precisely because the document does not
say so per operation (#1412), and `sdk/contract`'s omissions list already
carries a note about the same thing.

It is not urgent. Nothing is broken; the clients handle these statuses because
they were written from the error taxonomy rather than from the spec.

## The shape of the fix

Mechanical, and better done in one pass than incidentally. A helper in
`FountainWeb.Schemas` that appends the pipeline's statuses to an operation's
`responses` would let every `operation/2` macro declare them by inclusion
rather than by hand, which also keeps them right when a pipeline changes.

Doing it by hand across 68 operations is possible but invites the copy-paste
that caused #1417 in the first place.

## Where it is recorded now

`FountainWeb.SchemaGuardAllowlist`, one `{operation, status}` entry each, so a
new operation missing a plug status still fails until somebody decides.

Issue #1427, opened 2026-09-03
Guard the schema against the controller: assert rendered responses validate against their declared schema


jhgaylor:
Three defects of one class have surfaced in a day. Nothing in the repository can catch any of them, because every existing check compares a schema against another schema, and all three are a schema disagreeing with the **controller that renders it**.

## The three

1. **`AuthMeResponse.required` named properties the schema does not have** (fixed in #1417). It listed `name`, `prefix` and `created_at`, copied from `ApiKey`. `openapi-typescript` silently drops a required name with no matching property, so the published TypeScript client had **every field on that response optional**. It survived from the day the schema was written.
2. **`AuthMeResponse` declares `expires_at`, which `show/2` never renders** (#1418, open). The spec promises a field the endpoint does not return, and every generated SDK carries it.
3. **The declared 402 for `POST /api/conversations` is a bare `{"type": "object"}` with no properties**, while the body actually carries `upgrade_url` (ADR 0031, `Billing.check_spend/1`). Found while writing the #1412 linter.

#1411 made the OpenAPI document canonical across four SDKs and #1412 pinned cross-client behaviour, so both halves are now guarded — but a schema that lies about its own controller passes both, and then propagates into four clients with confidence.

## The fix is nearly free

`open_api_spex` is already a dependency and ships `OpenApiSpex.TestAssertions.assert_operation_response/1`. It takes a `conn` and validates the rendered body against the declared response schema for that operation. So the whole class is reachable from inside the Elixir suite with no new tooling.

Two shapes, and the choice is the substance of this issue:

- **A `ConnCase` hook** that validates every JSON response produced by any controller test that went through a documented operation. Broadest coverage, effectively free per test, but it will light up on every existing disagreement at once, so it needs a ratchet or a short allowlist to land.
- **One guardrail test** that walks the API controller tests, or the router, and asserts a rendered response per operation. Narrower and slower to write, but lands green and grows deliberately. `apps/fountain/test/fountain/audit_guardrail_test.exs` is the existing pattern for "add a thing and the guard fails until you cover it".

The ratchet pattern from the docs IA campaign applies either way: allowlist what is already wrong, forbid anything new, and let the list only shrink.

## Acceptance

- A new or changed operation whose controller does not render what its schema declares fails in CI, naming the operation.
- The three defects above are covered: reintroducing any of them fails.
- Whatever is allowlisted is listed explicitly with a reason, and the list can only shrink.
[...]

PR #1540, opened 2026-09-04
fix(schema-guard): see an extension's responses, and declare what they render (#1536)

### LINKED ISSUE #1536: The schema guard never sees an extension's responses, and the ratchet under-counts as a result
`FountainWeb.SchemaGuard` (#1427) validates every response the suite renders against the schema its operation declares. It has never seen an extension's responses. Both `fountain_buzz` (since #1507) and `fountain_support` (since #1528) serve operations that no schema guard checks.

## Why

`apps/fountain/lib/fountain_web/router.ex` ends the `/api` scope with:

```elixir
forward "/", FountainWeb.Plugs.ExtensionDispatch
```

`SchemaGuard.template/1` resolves a conn through `Phoenix.Router.route_info(FountainWeb.Router, ...)`, and a forward is opaque to that: the router reports the **forward's own** route, `/api`, not `/api/support/reports`. `operation/2` then finds no `PathItem` for `"/api"` and returns `{:skip, :undocumented}`, so nothing is ever recorded.

This is the same opacity that made `openapi_paths/0` necessary (ADR 0043 decision 3 — `Paths.from_router/1` filters the forward out because its plug exports no `open_api_operation/1`). The described-vs-served half was compensated for; the rendered-vs-declared half was not.

## Measured, not inferred

Attaching the guard in the extension's own `test_helper.exs` is **not** sufficient — the skip happens before the handler can find anything:

```
# apps/fountain_support, with FountainWeb.SchemaGuard.attach() added, and
# SupportReport deliberately broken with required: [..., :never_rendered]
30 tests, 0 failures
```

With `template/1` taught to re-resolve through the mount, the guard fires on the first run:

```
** (RuntimeError) Schema guard: POST /api/support/reports rendered a 401
   that does not match its declared schema.
```

That 401 is the `:plug_status` family (#1432). It was `{"POST /api/support/reports", 401} => :plug_status` in `FountainWeb.SchemaGuardAllowlist` until #1528 deleted it — **not because it was fixed, but because `SchemaGuardrailTest`'s "every entry names an operation the API still serves" reads `SchemaGuard.operations()`, which excludes extension operations when `apps/fountain` runs alone.** The entry looked stale; the defect was simply no longer observable. #1507 did the same for Buzz's operations. The allowlist's own moduledoc calls the inverse of this out ("a stale allowlist entry is a fix nobody noticed landing"); this is that failure mode running the other way, and the ratchet under-counts as a result.

## The patch that works

In `apps/fountain/test/support/schema_guard.ex`:

```elixir
defp template(%Plug.Conn{} = conn) do
  case Phoenix.Router.route_info(@router, conn.method, conn.request_path, conn.host) do
    :error ->
      {:skip, :no_route}

    # An extension's routes are behind `forward "/", ExtensionDispatch`, and a
    # forward is opaque: the core router reports the forward's own route
    # ("/api"), which matches no documented path, so every extension response
    # was silently skipped as `:undocumented` (ADR 0043, #1507).
    %{plug: FountainWeb.Plugs.ExtensionDispatch} ->
      extension_template(conn)

    %{route: route} ->
      {:ok, spell(route)}

    _ ->
      {:skip, :no_route}
  end
end

defp extension_template(%Plug.Conn{} = conn) do
  segments = conn.request_path |> String.split("/", trim: true) |> Enum.drop(1)

  with {_ext, mount, plug} when is_atom(plug) <- Fountain.Extensions.find_mount(segments),
       under = segments |> Enum.drop(length(mount)) |> Enum.join("/"),
       %{route: route} <-
         Phoenix.Router.route_info(plug, conn.method, "/" <> under, conn.host) do
    {:ok, spell("/api/" <> Enum.join(mount, "/") <> route)}
  else
    _ -> {:skip, :no_route}
  end
end

defp spell(route), do: Regex.replace(~r/:([a-zA-Z_][a-zA-Z0-9_]*)/, route, "{\\1}")
```

`Fountain.Extensions` is the host's own module, so this crosses no ADR 0043 boundary and `Fountain.ExtensionGuardTest` stays green.

Each extension's `test_helper.exs` also needs `FountainWeb.SchemaGuard.attach()`, because `scripts/test-libraries.sh` runs a sibling's suite from its own directory where `apps/fountain`'s helper never runs.

## Then decide what to do with what it finds

Two shapes, and they want different answers:

1. **The `:plug_status` 401s** (#1432's family). Either put the entries back and raise `@ceiling` — which needs `SchemaGuardrailTest`'s staleness check to stop treating an extension operation as stale when the extension is absent — or declare the statuses on the extensions' operations. Declaring them is the real fix and is small at this scale (3 support operations, 4 Buzz ones) rather than #1432's 66, **but it changes the published OpenAPI document**, so it cannot ride along with a move that promises a byte-identical spec.
2. **Anything else it finds.** Unknown until the guard runs against Buzz; `fountain_support` produced only the 401.

## Scope

Do both extensions in one change. Deliberately not folded into #1528/#1535: that PR's gate is a byte-identical `dist/openapi.json` and `sdk/contract/contract.json`, and the honest fix to (1) changes the document.

Found while building #1528. The patch above was written and run against `fountain_support`, then reverted.

jhgaylor:
Closes #1536. The first gate of tracker #1538, and the correctness gate the tracker puts before repository graduation.

`FountainWeb.SchemaGuard` (#1427) validates every response the suite renders against the schema its operation declares. It had never validated one of an extension's, and nothing failed — because a guard that checks nothing looks exactly like a guard that finds nothing.

## Why it saw nothing

`template/1` resolves through `Phoenix.Router.route_info(FountainWeb.Router, ...)`, and the `/api` scope ends with `forward "/", ExtensionDispatch`. A forward is opaque to `route_info/4`: the router reports the forward's own route, `/api`, which matches no documented path, so `operation/2` returned `{:skip, :undocumented}` for every extension response.

This is the same opacity that made `openapi_paths/0` necessary (ADR 0043 decision 3). The described-vs-served half was compensated for; the rendered-vs-declared half was not.

`template/1` now re-resolves through the mount, reading `Fountain.Extensions` — the host's own registry, so this crosses no ADR 0043 boundary and `Fountain.ExtensionGuardTest` stays green. Each extension's `test_helper.exs` attaches the guard too, because `scripts/test-libraries.sh` runs those suites from their own directories where `apps/fountain`'s helper never runs. `attach/0` is idempotent now: a root `mix test` calls it three times, and the second keeper used to die on `:ets.new` against a table the first already owns.

## What it found on the first run

Measured, not inferred. Every documented extension operation resolved, and two rendered a status nothing declared:

| | | |
|---|---|---|
| `POST /api/support/reports` | 401 | `TenantAPIAuth`, undeclared |
| `POST /api/buzz/agents` | 404 | `environment_not_found`, undeclared |

The 401 was `{"POST /api/support/reports", 401}` on `SchemaGuardAllowlist` until #1528 deleted it as stale — and it was not stale. `SchemaGuardrailTest`'s staleness check reads the operations `apps/fountain` serves, and that run installs no extension, so the entry named an operation the run could not see. The defect simply stopped being observable, and the deletion read as a fix landing. The allowlist's own moduledoc calls out the inverse ("a stale allowlist entry is a fix nobody noticed landing"); this was that failure mode running the other way.

The 404 is a real disagreement between `AgentController.create/2` and its schema, added with the `environment_id` support and catchable by nothing until now.

## What it does about them

Both extensions **declare** the statuses rather than going on the allowlist: 401 on all eight operations, plus the 404. That is nine declarations, against the 66 core operations #1432 still owes.

The ratchet therefore stays at 70 and gains no entry it could not evaluate. That rule is written down in `SchemaGuardAllowlist`'s moduledoc and enforced by `FountainWeb.ExtensionSchemaGuardCase`, from each extension's own suite — the only run that can see an extension's operations at all.

The published document moves, so `sdk/contract/contract.json` and `sdk/typescript/src/generated/openapi.ts` are rebuilt here. That is exactly why this could not ride along with #1528, whose gate was a byte-identical spec. All four SDK contract verifiers pass.

The fixture extension is an installed extension in the test VM like any other, and the guard sees it too: `:whoami` declared no 401 and `:deep` declared nothing at all. Both declare their responses now. The fixture describes OpenAPI paths only while the suite runs, so the published artifact is unchanged by that half — verified with `build.sh --check`.

## Verified by reverting

A guard that stopped guarding looks exactly like one finding nothing, so both halves were broken on purpose:

- Removing the `ExtensionDispatch` clause from `template/1` fails the new coverage test, naming all three support operations resolving to `GET /api`.
- `required: [..., :never_rendered]` planted on `SupportReportResponse` fails 2 tests across 3 operations; the same on `BuzzIdentityResponse` fails 8. Before this change both suites were `0 failures` with the schema broken.

## Gate

`mix precommit` green: 4,206 + 114 + 33 tests, 0 failures; credo clean; dialyzer 0 errors; sobelow clean; prod release assembles. Formatting checked with the pinned 1.19.2.

## Left for the tracker

`POST /api/mcp/buzz/{conversation_id}` stays `:undocumented` — `McpController` declares no `operation`, the same as core's own `/api/mcp/team/...` transports. It is a sandbox transport rather than public API, so this is the correct outcome, not a gap.

https://claude.ai/code/session_015unQSjdFCfcmt4XJHPHtAY

### DIFF
--- a/apps/fountain/test/support/extension_fixtures.ex
+++ b/apps/fountain/test/support/extension_fixtures.ex
@@ -67,6 +67,22 @@ defmodule Fountain.ExtensionFixtures do
         }
       })
     end
+
+    defmodule Deep do
+      @moduledoc false
+      require OpenApiSpex
+
+      OpenApiSpex.schema(%{
+        title: "FixtureDeep",
+        type: :object,
+        properties: %{
+          path_info: %OpenApiSpex.Schema{
+            type: :array,
+            items: %OpenApiSpex.Schema{type: :string}
+          }
+        }
+      })
+    end
   end
 
   defmodule Controller do
@@ -77,10 +93,20 @@ defmodule Fountain.ExtensionFixtures do
     use Phoenix.Controller, formats: [:json]
     use OpenApiSpex.ControllerSpecs
 
+    # Both operations declare every status they can render, because the schema
+    # guard resolves through `ExtensionDispatch` since #1536 and this fixture is
+    # an installed extension in the test VM like any other. Before that it was
+    # the only kind of extension there was, and none of its responses were ever
+    # checked — `:whoami` declared no 401 and `:deep` declared nothing at all.
+    # A fixture that models the seam should model a well-formed extension too.
     operation(:whoami,
       summary: "Who the host authenticated",
       operation_id: "fixtureWhoami",
-      responses: [ok: {"Whoami", "application/json", Fountain.ExtensionFixtures.Schemas.Whoami}]
+      responses: [
+        ok: {"Whoami", "application/json", Fountain.ExtensionFixtures.Schemas.Whoami},
+        unauthorized:
+          {"Missing or invalid API key", "application/json", FountainWeb.Schemas.Error}
+      ]
     )
 
     def whoami(conn, _params) do
@@ -94,7 +120,15 @@ defmodule Fountain.ExtensionFixtures do
       })
     end
 
-    operation(:deep, summary: "A nested path", operation_id: "fixtureDeep", responses: [])
+    operation(:deep,
+      summary: "A nested path",
+      operation_id: "fixtureDeep",
+      responses: [
+        ok: {"The trimmed path", "application/json", Fountain.ExtensionFixtures.Schemas.Deep},
+        unauthorized:
+          {"Missing or invalid API key", "application/json", FountainWeb.Schemas.Error}
+      ]
+    )
 
     def deep(conn, _params), do: json(conn, %{path_info: conn.path_info})
   end
--- a/apps/fountain/test/support/extension_schema_guard_case.ex
+++ b/apps/fountain/test/support/extension_schema_guard_case.ex
@@ -1,0 +1,123 @@
+defmodule FountainWeb.ExtensionSchemaGuardCase do
+  @moduledoc """
+  The schema guard's own coverage check, for one extension (ADR 0043, #1536).
+
+  `FountainWeb.SchemaGuard` validates every response the suite renders against
+  the schema its operation declares — and for the first two extensions it
+  validated none of them, silently, for as long as they existed. The core
+  router ends `/api` with `forward "/", ExtensionDispatch`, and a forward is
+  opaque to `Phoenix.Router.route_info/4`: it reports the forward's own route,
+  `/api`, which matches no documented path, so every extension response was
+  skipped as `:undocumented`.
+
+  Nothing failed, because a guard that checks nothing looks exactly like a
+  guard that finds nothing. That is the failure this file exists to make
+  impossible a second time: it asserts the guard can *resolve* every operation
+  the extension documents, which is the step that was missing. It says nothing
+  about whether those responses are correct — the guard itself does that, on
+  every request the extension's own tests make.
+
+  Use it from the extension's suite, naming the extension module:
+
+      defmodule FountainBuzz.SchemaGuardTest do
+        use FountainWeb.ExtensionSchemaGuardCase, extension: FountainBuzz.Extension
+      end
+
+  It lives in core's `test/support` rather than in each extension because both
+  extensions need the identical checks and neither may depend on the other. It
+  names no extension: the module under test arrives as an option, and
+  everything else is read from `Fountain.Extensions`, which is the host's own
+  registry.
+  """
+
+  @doc false
+  defmacro __using__(opts) do
+    extension = Keyword.fetch!(opts, :extension)
+
+    quote bind_quoted: [extension: extension] do
+      use ExUnit.Case, async: true
+
+      alias FountainWeb.{SchemaGuard, SchemaGuardAllowlist}
+
+      @extension extension
+
+      test "the extension is installed in this run" do
+        # Everything below is vacuous otherwise, and "vacuously true" is the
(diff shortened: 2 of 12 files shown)


jhgaylor:
## `sdk-no-release`, deliberately

The SDK release gate is right: this changes `sdk/typescript/src/generated/openapi.ts` and leaves the version at 1.20.0. Labelling rather than bumping, for two reasons.

**The change is additive types only.** Nine new response variants (`401` on eight operations, `404` on `POST /api/buzz/agents`) on operations that already existed. No runtime code, no client behaviour, no removed or narrowed type — regenerating from the spec is the whole diff.

**Two bumps are already in flight and must merge in ascending order.** #1489 claims 1.21.0 and #1481 claims 1.22.0. `sdk-publish.yml` runs a bare `npm publish`, so npm moves the `latest` dist-tag to whatever it just published regardless of semver; a third number here adds another ordering constraint to that race for a types-only change. Picking SDK versions while other bumps are open is the coordinator's call, not this PR's.

These types publish with whichever of those releases lands next — both regenerate the same file from the same spec, so they carry this change forward automatically.

If you would rather this cut its own release, say so and I will take the next free number and add the four edits (`package.json`, `package-lock.json`, `USER_AGENT` in `src/http.ts`, CHANGELOG heading).

PR #1559, opened 2026-09-04
perf(ci): split the checks job into four that share nothing

jhgaylor:
Follow-up to #1556. That PR took `Static analysis and release checks` from 426s to a **measured 201s** by stopping `npm ci` waiting on the advisories endpoint. 201s was still the longest job in the run, and still one sequential queue of five unrelated toolchains sharing nothing but the checkout.

## The split

| Job | Runs | Est. |
|---|---|---|
| `elixir-static` | format, compile, credo, hex audit, sobelow, dialyzer | ~85s |
| `release-and-contract` | prod release boot probe; the OpenAPI document and the contract projected from it | ~60s |
| `sdk-clients` | the four SDKs, both Go modules, the Hermes plugin | ~67s |
| `docs-prose` | the three prose gates, when a diff touches docs *and* code | ~20s |

Dialyzer bounds `elixir-static`, so ~85s is the new critical path for this half of CI — down from 201s, and below the test partitions (~164s), which is the point.

## How I checked nothing was lost

The steps were moved mechanically, not retyped. Every `run:` line was extracted from the old job and the four new ones and diffed as sets:

```
old checks: 53 steps, 118 command lines
new 4 jobs: 64 steps, 120 command lines

command lines in OLD but not in NEW:  (none)
command lines in NEW but not in OLD:  (none)
```

Same five actions used, same versions. The two extra command lines are the `npm ci` that a second job now needs and the de-stink gate's own `setup-node` — it no longer inherits Node 24 from the SDK steps it used to sit beside. `actionlint` is clean.

## Three things that are not a pure move

- **Three cache keys instead of one.** `checks` compiled `:test`, `:dev` *and* `:prod` into a single cached `_build`. Each job now caches one coherent tree, and `sdk-clients` needs no umbrella build at all. This is the race the note on `test`'s deps cache already describes.
- **The conformance lint runs first in `sdk-clients`.** It used to sit directly above the three conformance runs, which was the same thing while the order was Elixir → Python → TypeScript throughout. Steps are grouped by language now, so a malformed fixture would otherwise fail two adapters before the step that names the real problem ever ran.
- **The STE linter comment** justified a pinned binary by *this job pinning Go with `GOTOOLCHAIN=local`*. `docs-prose` installs no Go at all, so the reason is now simpler rather than wrong.

## Also

`mix.exs` names the job that fails an unassemblable prod release, and CLAUDE.md listed the gates as a flat numbered list; both now point at the job that runs them.

While regrouping that list I found a stale claim: it said `mix hex.audit` was "non-blocking — visible in the log, doesn't fail the build", but `scripts/hex-audit-gate.exs` **fails the build** on a security advisory and only lets retirements through. Corrected.

## Not done

`docs-prose` and the docs-only `docs` job look near-identical but are not mergeable — `docs` also carries `docs_test.exs` and the CLI docs parity check, which a docs+code diff gets from the full suite instead. Left alone.

https://claude.ai/code/session_01GXM8homHJBHZT1KLUxP71e

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,30 @@ permissions:
 # the same job while sharing nothing with it but the checkout. #619 split them
 # so the suite runs alone on the critical path with the rest beside it.
 #
+# That "everything else" job (`checks`) then grew until it was itself the
+# critical path: 426s against 171s for the slowest test partition. Two thirds
+# of that turned out not to be analysis at all — `npm ci` runs a vulnerability
+# audit by default, and the advisories endpoint stalls from a runner rather
+# than failing fast, which cost 88s to 302s per run for a result `npm ci`
+# cannot fail on. `--no-audit` took the job to 201s.
+#
+# What was left was still the longest job, and still one sequential queue of
+# five unrelated toolchains. It is now four jobs that share nothing:
+#
+#   elixir-static         format, compile, credo, hex audit, sobelow, dialyzer
+#   release-and-contract  the prod release boot probe, and the OpenAPI document
+#                         plus the contract projected from it
+#   sdk-clients           the four SDKs, both Go modules, the Hermes plugin
+#   docs-prose            the three prose gates, when a diff touches docs
+#
+# Dialyzer is the pole in the first and the whole job is bounded by it, which
+# is the shape to remember before adding anything to `elixir-static`.
+#
+# The split is also why the three Elixir jobs get three cache keys rather than
+# sharing `-checks-`: see the note on `test`'s deps cache. `checks` used to
+# build :test, :dev AND :prod into one cached _build; each job now caches one
+# coherent tree, and `sdk-clients` needs no umbrella build at all.
+#
 # #620 then split the suite itself across three machines. ExUnit reported why:
 #
 #   Finished in 252.3 seconds (140.7s async, 111.6s sync)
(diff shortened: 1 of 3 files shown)


jhgaylor:
## Measured

The estimates in the description were warm-cache numbers. The first run on this PR was **cold** — the new cache keys had nothing to restore — so both Elixir jobs compiled the umbrella from scratch and came out *worse* than the 201s they replaced. Re-running against the caches that run saved gives the steady state:

| Job | Estimated | Cold (attempt 1) | **Warm (attempt 2)** |
|---|---|---|---|
| `elixir-static` | ~85s | 317s | **85s** |
| `release-and-contract` | ~60s | 323s | **71s** |
| `sdk-clients` | ~67s | 70s | **62s** |

Both attempts were fully green.

The cold run is worth reading, because the cost lands on steps that don't look like compilation: `Set up database` 140s → **14s**, `Check formatting` 109s → **2s**, `Dialyzer` 137s → **28s**. Those were cold-`_build` compiles wearing other steps' names.

So this half of CI goes **426s → 201s (#1556) → 85s**, and the run is now bounded by the test partitions (168s) and Swift on macOS (163s) rather than by static analysis.

## One-time cost on main

A cache saved on `refs/pull/1559/merge` is readable only by this PR. After merge, the **first** run on main pays the cold cost again on `-mix-static-`/`-build-static-` and `-mix-release-`/`-build-release-`; once main has saved them, every PR reads them. One run, on main, and then it is warm forever.

That is inherent to renaming a cache key, not a defect in the split — but it does mean the merge commit's CI will look slow, and that is expected rather than a regression to chase.

## Cache sizes

Incidentally confirms the coherence argument: `checks` cached one 130MB `_build` holding `:test`, `:dev` and `:prod`. It is now `build-static` 60MB and `build-release` 101MB, each one coherent tree, and `sdk-clients` caches no umbrella build at all.

Issue #1565, opened 2026-09-04
Conversations: reapply or change the Agent, Environment and Vault without starting a new thread


jhgaylor:
## Problem

Agent, environment and vault selection is resolved when a conversation is launched and its sandbox is provisioned. There is no supported way to apply a different selection — or deliberately reapply the current selection after its config or secrets change — to an existing conversation.

Today the workaround is to start a new conversation. That loses the stable conversation id and breaks continuity for its title, channel binding, spawn tree, transcript, events, read state and usage. `DELETE /api/sandboxes/:id` (#1071) is not the same operation: it rebuilds under the existing sandbox identity and can affect every conversation sharing a persistent home.

This matters in both forms:

- **Refresh:** an Agent, Environment or Vault changed and the user wants the existing conversation's next machine to receive the new system prompt, skills, MCP config, variables or secrets. Vault values in particular merge only at spawn.
- **Rebind:** the user wants the existing conversation to continue under a different Agent, Environment override or Vault, without creating a new thread.

## Proposed behavior

Add one tenant-scoped conversation action that accepts an agent/environment/vault selection and reprovisions the conversation under that resolved launch configuration. An illustrative shape is:

```http
POST /api/conversations/:id/reapply
Content-Type: application/json

{
  "agent_id": "...",
  "environment_id": "...",
  "vault_id": "..."
}
```

The exact route name can follow the API vocabulary, but it should be one atomic operation rather than three independent field updates.

Request semantics:

- An omitted field keeps the conversation's current selection.
- `environment_id: null` clears the per-conversation override and returns to the selected Agent's environment.
- `vault_id: null` detaches the Vault.
- Supplying no fields is valid and means “reapply what this conversation currently uses.”

The action should:

1. Resolve and validate the complete target configuration using the same tenant, Agent allowlist, provider/runtime and permission gates as a fresh launch.
2. Refuse while this conversation has a turn in flight. A rejected request must not partially change bindings or retire a sandbox.
3. Preserve the Conversation row and its id, turns, events, title, channel/parent bindings, read state and accumulated usage.
4. Update `agent_id`, `agent_version_id`, the snapshotted `runtime`, `environment_id` and `vault_id` together.
5. Clear `runtime_session_id` and move only this conversation to a sandbox with the target identity. Do not mutate a running machine's environment in place.
[...]

PR #1569, opened 2026-09-04
Conversations: reapply agent, environment, and vault

jhgaylor:
## Summary

Adds `POST /api/conversations/:conversation_id/reapply` and matching TypeScript, Python, Elixir, and Swift SDK helpers. Reapply refreshes or replaces the Agent, Environment, and Vault on the conversation's existing machine, preserving its transcript, runtime session, and working files.

An empty body refreshes the current selection. Omitted fields retain their current binding; explicit null clears the Environment override or Vault. Changes requiring a rebuilt machine return `409 rebuild_required`; running turns return 409, provisioning returns 503, and terminal conversations return 410.

## Configuration and sharing

- Update the conversation and sandbox identities in one transaction. Attachments and turn admission use the same sandbox lock; a conflicting persistent home rolls back the change.
- Persist a configuration revision and check it when admitting a turn. A server that missed the refresh notification reloads configuration before spawning the runtime. Refresh errors are returned instead of publishing a false successful lifecycle event.
- Reconcile skills on wake as well as live reapply. Track installed names by source, remove obsolete managed skills, preserve unrelated files, and retain existing remote skills if an offline reinstall fails. Recover legacy names from the recorded Agent version and the skills installer's source lock.
- Audit committed configuration changes and emit lifecycle metadata. Add a migration for the configuration revision and applied skill selection.

Implements the server and SDK portion of #1565. The conversations UI remains a follow-up in the separate fountain-conversations repository.

## Validation

- `mix precommit` passed: compile with warnings as errors, dependency cleanup check, formatting, strict Credo, Dialyzer, Sobelow, production release assembly, and the full umbrella test suite.
- Core: 4,309 tests and 6 doctests, zero failures (7 excluded). Buzz: 143 tests, zero failures. Support: 33 tests, zero failures.
- Regression coverage includes old/new attachment identity, persistent-home collision rollback, concurrent selection semantics, stale revision admission, missed-notification environment reload, sleeping-conversation skill updates, removed inline/GitHub skills, legacy source-lock recovery, and offline retention.

The SDK surfaces and their contracts are unchanged by the review fixes. The original implementation passed TypeScript (102 tests, typecheck and contract), Python (42 tests and contract), Elixir SDK (52 tests and contract), Swift (78 tests and format), and `scripts/sdk-contract/build.sh --check`.


BinaryBourbon (changes requested):
Requesting changes — not on the substance, which I can't fairly assess in this state, but because the branch has drifted past the point where a review means anything.

**161 commits behind `main`, conflicting in 15 files:**

```
apps/fountain/lib/fountain/conversations.ex
apps/fountain/lib/fountain/conversations/conversation_server.ex
apps/fountain/lib/fountain/conversations/turn_machine.ex
apps/fountain/lib/fountain/sandbox_skills.ex
apps/fountain/test/fountain/audit_guardrail_test.exs
apps/fountain/test/fountain/conversations/conversation_server_size_test.exs
docs/api.md
sdk/python/{CHANGELOG.md,pyproject.toml,src/fountain/__init__.py,src/fountain/http.py}
sdk/typescript/{CHANGELOG.md,package.json,package-lock.json,src/http.ts}
```

Those first three are the worst possible set. `ConversationServer` has been through the subtraction campaign, `turn_machine.ex` has taken the acp runtime stack and now the detached-permission work, and `conversations.ex` is where the apply-kinds and labels stacks landed. A 2,351-line branch touching all three from a September 4th base isn't a rebase, it's a re-derivation — and the green checks on this PR are from a `main` that no longer exists, so they're evidence about nothing.

Both SDK version files conflict too, which is the silent one: whatever this branch bumped to has since been taken on `main`, so even a clean-looking resolution can leave a version that never publishes.

**It's also ~2,350 lines across 56 files, which is what we've now said we don't do.** #1678 was closed and re-cut as #1852–#1860 for being half this size. The reapply work has obvious seams to cut along — the configuration revision + migration, the sandbox/conversation identity transaction, the skills reconciliation, the API surface, then the four SDKs — and each of those is independently reviewable in a way the whole thing isn't.

My recommendation is to close this and re-cut it as a stack on current `main`, the way #1635 and #1634 were. If you'd rather keep it open, it needs the rebase first — I'd be reviewing fiction otherwise.

One thing worth carrying forward whichever way you go: the description mentions running turns returning 409, provisioning 503 and terminal conversations 410. That three-way split is the interesting part of this design and it deserves its own small PR with its own tests rather than arriving inside a 56-file diff.

jhgaylor:
Closing in favour of a stack, under the no-big-PRs rule. This branch was 2,351 lines across 56 files and 161 commits behind `main`, with conflicts in `conversations.ex`, `conversation_server.ex`, `turn_machine.ex`, `sandbox_skills.ex`, `audit_guardrail_test.exs`, `conversation_server_size_test.exs`, `docs/api.md` and eight SDK files.

The stack is re-derived against today's code rather than rebased: `ConversationServer` went through the subtraction campaign (#1369), `turn_machine.ex` took the ACP runtime stack (#1634) and the detached-permission work (#1635), and `conversations.ex` absorbed the apply-kinds (#1636) and labels (#1637) stacks. Nothing this branch decided was dropped.

| | PR | What it covers |
|---|---|---|
| 1 | #1886 | Four function families leave `ConversationServer` (callback-key revocation, turn orphaning, the `format_status` redactor, the checkpoint pair). No behaviour change; room for the rest. |
| 2 | #1887 | `conversations.configuration_revision`, `sandboxes.build_fingerprint` and `sandboxes.applied_skills`, the migration, and `Reapply.fingerprint/1` recorded when a machine reaches `ready`. |
| 3 | #1888 | `Reapply.check/2` and `explain/1`: which selections can be applied to the machine that is already there, and which need the disk built again. |
| 4 | #1889 | `Conversations.reapply_conversation/3` — one transaction under the sandbox lock turn admission takes, the sandbox identity update with the persistent-home collision rollback, the guarded write, and the audit event. |
| 5 | #1890 | The revision checked at turn admission, `ConversationServer.refresh_configuration/2`, the reload when a server missed the notification, and the `configuration`/`done` lifecycle event. |
| 6 | #1891 | Skills reconciliation: installed names tracked by source, obsolete managed skills removed, unrelated files preserved, existing remote skills retained when an offline reinstall fails, legacy names recovered from the recorded Agent version and the installer's source lock. On a live reapply and on every wake. |
| 7 | #1892 | `POST /api/conversations/:conversation_id/reapply` and its status-code contract: 409 `rebuild_required` with a `field`, 409 `conversation_busy`, 503 while the machine is built, 410 once the conversation has ended. |
| 8 | #1893 | The manual, the Python, Elixir and Swift SDKs, and the one version bump. |

### Two departures from this branch, both deliberate

- **The shared attach/reapply transaction** (#1889). This branch wrapped both in `Reapply.transaction/2`. Today's `create_attached_conversation/3` already rechecks `check_attachable/4` inside a transaction holding `FOR SHARE` on the sandbox row, and a reapply's `update_sandbox` takes the conflicting lock, so the attach path needs no change at all.
- **Runtime resolution in the skills reconciliation** (#1891) goes through `Fountain.RuntimeDispatch` rather than `Managoat.Runtimes.for_runtime/1`, for the reason #1634 gives: the library knows four runtimes and `acp` is not one of them.

### One thing this branch had wrong

Its `docs/api.md` line and several SDK doc comments said a reapply moves the conversation to "a fresh machine", which contradicts the change itself. The machine is kept, and the files on its disk with it. The stack says so everywhere.

Still out of scope, as here: the conversations UI, which lives in its own repository.

jhgaylor:
Correction to the map above: the TypeScript SDK bump in #1893 is **1.27.0 → 1.28.0**, not 1.26.0 → 1.27.0. #1849 landed 1.27.0 on `main` while this stack was open, so the bump was re-cut on top of it. Concurrent bumps have to land in ascending order.

The stack has also been rebased onto current `main` (the merge queue of ADR 0050, and the #1637 labels stack).

PR #1573, opened 2026-09-04
fix(test): the sandbox connection is a queue, so let it queue (#1524, #1568)

### LINKED ISSUE #1524: Test flake: DBConnection pool checkout times out in Task.Supervised preloads
Seen three times while building the extension campaign (#1503), always the same shape and always locally, never on CI's six partitions.

## What it looks like

```
** (DBConnection.ConnectionError) could not checkout the connection owned by #PID<...>
   ({Fountain.TeamTest, :"test list_teammates/1 prefers the newest live conversation
    over a newer terminated one"}). When using the sandbox, connections are shared,
    so this may imply another process is using a connection.
    Reason: connection not available and request was dropped from queue after ~130ms.
```

The stack is always an `Ecto.Repo.Preloader.fetch_query/8` inside `Task.Supervised.invoke_mfa/2` — a supervised, unlinked task (the `Fountain.TaskSupervisor` pattern CLAUDE.md mandates) doing a preload while its owning test still holds the sandbox connection.

Observed in `Fountain.TeamTest` and `Fountain.Team.McpTest`; a `Fountain.ConversationsWakeTest` Mimic-count failure appeared once in the same runs and may be the same underlying contention.

## Why it is probably not any of those tests

- It reproduces roughly one full-suite run in four, on a single local VM running ~4,200 tests at `max_cases: 20` against one `pool_size: 20` pool.
- CI splits the same suite six ways, so each partition contends far less — 20+ green partition runs across three PRs with no occurrence.
- Nothing in the failing tests is new; they predate this campaign.

## Worth investigating

- Whether `DataCase`'s "wait for the tasks a test started" step covers preloads started from inside a supervised task (a task spawning work of its own).
- Whether the pool's `queue_target`/`queue_interval` defaults are simply too tight for 20 async modules on a laptop, in which case this is a local-only annoyance to document rather than a defect.

Not blocking anything; filing so the next person who sees it does not spend an hour deciding whether their branch caused it. It did not.

### LINKED ISSUE #1568: Flaky: CreditsTest concurrent-post test exhausts the DB pool on a full local run
## What happened

On a full local `mix test` (4,260 tests, one machine, `max_cases: 20`),
`ee/test/fountain/credits_test.exs:127` failed:

```
** (DBConnection.ConnectionError) could not checkout the connection owned by
#PID<0.851.0> ({Fountain.CreditsTest, :"test grant/4 and debit/4 concurrent
posts of one key move the balance once"}). ... connection not available and
request was dropped from queue after 263ms
    (fountain) ee/lib/fountain/credits.ex:578: Fountain.Credits.insert_and_move/3
    (fountain) ee/lib/fountain/credits.ex:534: Fountain.Credits.post/4
```

Seen 2026-09-04 while verifying #1551 (PR #1564). It is **not** caused by that
change: nothing in it touches `Credits.post/4`, and replaying the failing seed
(425799) against that file in isolation passed 3/3. CI's six partitions were
green on the same tree.

## Why it is load-dependent

The test is doing the right thing for what it asserts — that one idempotency
key posts once under concurrency:

```elixir
1..8
|> Task.async_stream(fn _ ->
     Ecto.Adapters.SQL.Sandbox.allow(Repo, self(), self())
     Credits.grant(user.id, 100, "purchase", idempotency_key: "race")
   end, max_concurrency: 8)
```

Each task enters `insert_and_move/3`, which opens a transaction and takes
`SELECT ... FOR UPDATE` on the user row. So eight connections are held while
seven of them wait on one row lock, and each waiter keeps its checkout for the
whole wait. With `pool_size: 20` that is 40% of the pool parked on one row, and
every other async module running beside it is competing for the remainder.
Under a partitioned CI run the surrounding load is lower, which is why it is
green there and red on a full single-machine run.

The failure mode is `queue_timeout`, not a wrong balance — the assertion under
test never got to run.

## Why it is worth fixing rather than ignoring

`Credits.post/4` is the ledger, so this test guards the property that money
moves once. A test that fails on load is a test people learn to re-run, and the
next real regression in that path arrives looking exactly like this one. It is
also a local-only red: anyone running the full suite before pushing hits it and
has to decide whether it is theirs, which is the cost this issue is about.

## Options

- Lower `max_concurrency` to something under the pool's share and assert the
  same property. Eight was never load-bearing; two racers prove "posts once".
- Serialise the checkout: have the tasks contend without each holding a
  connection for the whole wait.
- Give this one module a larger pool, or mark it `async: false`. Cheapest, and
  the least honest — it hides the contention rather than bounding it.
- Raise `:queue_target`/`:queue_interval` for the test repo. Treats the symptom
  and slows down every genuine pool exhaustion elsewhere.

Do **not** lower `pool_size` (CLAUDE.md) or delete the concurrency assertion.

## Reproduce

```
mix test                      # full suite, single machine — fails intermittently
mix test ee/test/fountain/credits_test.exs --seed 425799   # passes; needs the load
```

jhgaylor:
Closes #1524. Closes #1568.

Two flakes, one mechanism, and it is neither the pool nor the code under test.

## What is actually happening

`Ecto.Adapters.SQL.Sandbox` gives a test **one** connection, and every process the test fans work out to borrows it — `Task.async_stream` in a test body, a supervised task under `Fountain.TaskSupervisor`, and (in the stack of most of these failures, though nobody wrote it) Ecto's own preloader, which fans out whenever two or more associations are still to fetch. Those processes do not run in parallel. They queue on one connection.

Measured on #1568's own shape:

| Probe | Result |
|---|---|
| eight fanned-out tasks, `SELECT pg_backend_pid()` | one backend pid |
| four "concurrent" 200 ms transactions | 814 ms |

DBConnection drops a waiter once the queue has been slow for a whole `queue_interval`, and the 50 ms / 1 s defaults assume a pool with spare connections rather than a queue that is a queue by construction. A 40-task probe reproduces both issues' exact error text on demand:

| `queue_target` / `queue_interval` | dropped |
|---|---|
| 50 ms / 1 s (defaults) | **30 of 40** |
| 200 ms / 5 s (this PR) | **0 of 40** |

The real failures waited 121-263 ms, so 200 ms / 5 s leaves about an order of magnitude of headroom. A connection that is genuinely stuck still surfaces — as the test's own timeout, carrying the stack of whatever holds it, which says more than a dropped checkout does.

## Two corrections to what the issues assumed

- **#1568's stated cause was wrong.** It read the failure as eight connections holding 40% of a 20-connection pool while seven waited on one `SELECT ... FOR UPDATE`. There is one connection and no row-lock contention — the transactions are serial, so each takes the lock uncontended. `pool_size` is not the lever here, and CLAUDE.md's floor of 20 is not in tension with this.
- **That test's concurrency was theatre.** Serialized posts cannot exercise the race `insert_and_move/3` exists to survive (a lookup-first implementation would pass it too). The test now says what it asserts — one key posts one row, the rest report `:duplicate`, the balance moves once — and fans out to two rather than eight, since the other six bought nothing but queue. The simultaneity is guarded by the unique index, in production, where the posters really are simultaneous.

## Verification

- The probe numbers above, and `is_integer/1` in the new guard: a missing key reads as `nil`, and `nil >= 200` is `true` under term ordering, so the first version of that assertion passed on exactly the config it exists to reject. Stripping the two settings now fails it.
- Full local suite on this branch: **4,276 tests, 5 failures**, plus `fountain_buzz` 143/0 and `fountain_support` 33/0. All five failures are `FountainWeb.Marketing*Test` and reproduce unchanged with `main`'s `config/test.exs`: they assert `-core` behaviour and only pass when run from `apps/fountain`, which is what CI's partition script does and what the umbrella-root `mix test` in CLAUDE.md's Quick start does not. Pre-existing, unrelated, filed separately.

https://claude.ai/code/session_01DBK3uwQuJozQHKmEuBrvrP

### DIFF
--- a/apps/fountain/test/fountain/sandbox_queue_config_test.exs
+++ b/apps/fountain/test/fountain/sandbox_queue_config_test.exs
@@ -1,0 +1,44 @@
+defmodule Fountain.SandboxQueueConfigTest do
+  @moduledoc """
+  The queue settings that keep the sandbox's one connection from dropping
+  waiters (#1524, #1568).
+
+  Under `Ecto.Adapters.SQL.Sandbox` a test has a single connection, shared by
+  every process it fans work out to — including Ecto's own parallel preloader,
+  which nobody wrote and which is in the stack of most of these failures. Those
+  processes queue rather than run in parallel, and at DBConnection's defaults a
+  loaded run drops them with a `ConnectionError` that names the test's own
+  connection. Neither the pool size nor the code under test has anything to do
+  with it, which is why both issues cost an investigation before they cost a
+  fix.
+
+  Asserted here rather than left as a comment in `config/test.exs` because the
+  cost of losing these two lines is not a failing test: it is the same
+  intermittent red, rediscovered from scratch.
+  """
+
+  use ExUnit.Case, async: true
+
+  @config Application.compile_env(:fountain, Fountain.Repo)
+
+  # The real failures waited 121-263 ms. A waiter is dropped only once it has
+  # waited past `queue_target` and the queue has been slow for a whole
+  # `queue_interval`, so these leave roughly an order of magnitude over what
+  # was observed.
+  #
+  # `is_integer/1` first, and not for tidiness: a missing key reads as `nil`,
+  # and `nil >= 200` is `true` under Elixir's term ordering, so the obvious
+  # form of this assertion passes on exactly the config it exists to reject.
+  test "the sandbox connection tolerates a queue" do
+    assert is_integer(@config[:queue_target]) and @config[:queue_target] >= 200
+    assert is_integer(@config[:queue_interval]) and @config[:queue_interval] >= 5_000
+  end
+
+  # CLAUDE.md's rule, and the reason it is not in tension with the two above:
+  # the pool is what concurrent *tests* draw from, the queue settings are about
+  # contention inside one test on one connection.
+  test "the pool still holds 20" do
+    assert @config[:pool_size] == 20
+    assert @config[:pool] == Ecto.Adapters.SQL.Sandbox
+  end
+end
--- a/config/test.exs
+++ b/config/test.exs
@@ -4,7 +4,28 @@ config :fountain, Fountain.Repo,
   url:
     System.get_env("DATABASE_URL", "postgres://postgres:postgres@localhost:5432/fountain_test"),
   pool: Ecto.Adapters.SQL.Sandbox,
-  pool_size: 20
+  pool_size: 20,
+  # The sandbox gives a test *one* connection, which every process the test
+  # fans work out to shares: `Task.async_stream` in a test body, a supervised
+  # task under `Fountain.TaskSupervisor`, and Ecto's own parallel preloader,
+  # which fans out over `Task.async_stream` whenever two or more associations
+  # are still to fetch. Those are not parallel under the sandbox — they are a
+  # queue on one connection, and four 200 ms transactions measured 814 ms.
+  #
+  # DBConnection then drops a waiter once the queue has been slow for a whole
+  # `queue_interval`, and the 50 ms / 1 s defaults are sized for a pool with
+  # spare connections rather than for a queue that is a queue by construction.
+  # On a loaded workstation that dropped real tests: #1524 (Ecto's preloader,
+  # waits of 121-160 ms) and #1568 (the credit ledger's concurrent-post test,
+  # 263 ms), neither of them a defect in the code under test.
+  #
+  # 200 ms / 5 s tolerates about 5 s of continuous backlog on one connection,
+  # against the ~250 ms the real failures waited. A genuinely stuck connection
+  # still surfaces — as the test's own timeout, with the stack of whatever is
+  # holding it, which says more than a dropped checkout does. Measured with a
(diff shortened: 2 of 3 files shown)


jhgaylor:
A full umbrella-root run on a *different* branch (#1576, which does not carry this change) reproduced the class live, which is better evidence than the probe:

```
1) test POST /api/environments/:environment_id/secrets returns 404 when environment belongs to another user
   (FountainWeb.SecretControllerTest)
   ** (DBConnection.ConnectionError) could not checkout the connection owned by #PID<0.3630.0> ...
      dropped from queue after 113ms

2) test grant/4 and debit/4 concurrent posts of one key move the balance once (Fountain.CreditsTest)
   ** (DBConnection.ConnectionError) could not checkout the connection owned by #PID<0.13912.0> ...
      dropped from queue after 247ms
```

The second is #1568 itself. The first has never been filed and is a plain controller test — no fan-out written anywhere in it, the drop happening inside `SecretController.create/2`. That is the point of fixing this in `config/test.exs` rather than per test: the fan-out that queues is usually Ecto's, so the next occurrence lands somewhere with no concurrency in sight and gets investigated from scratch.

Both waits (113 ms, 247 ms) are inside the 400 ms this PR's `queue_target` needs before a waiter can be dropped at all.

PR #1593, opened 2026-09-04
fix(test): give the verify-pending advances a real deadline (#1592)

### LINKED ISSUE #1592: Flake: VerifyPendingLiveTest advances are asserted with a 100 ms deadline
`FountainWeb.VerifyPendingLiveTest` "the poll advances even when the broadcast never arrives" failed once in eight full local suite runs on `main`:

```
  1) test auto-advance the poll advances even when the broadcast never arrives
     ** (ArgumentError) expected FountainWeb.VerifyPendingLive to redirect to "/dashboard", but got none
     code: assert_redirect(lv, "/dashboard")
```

## Mechanism

Every advance off that page is out of band: the LiveView re-reads the row on a broadcast or on its own poll, redirects, and the test proxy relays that to the test process. `assert_redirect/2` waits ExUnit's `assert_receive_timeout` for the whole round trip, which is **100 ms** by default.

Measured over 20 advances on a loaded machine: 0-34 ms, median 3 ms. That fits in 100 ms with room to spare when the file runs alone — 25 consecutive runs of it are green — and does not when a partition is running 20 cases at once, which is the only configuration that has ever failed.

Nothing is racing here. The redirect either happens or it does not; the deadline is the only variable, and 100 ms is a default, not a requirement of this page.

## Fix

Give the three `assert_redirect` calls in that file an explicit deadline. Deliberately not a suite-wide `assert_receive_timeout` bump: these three are the only `assert_redirect` calls in `apps/fountain/test`, and nothing else has been observed failing this way.

Found by running the full suite repeatedly while sweeping this week's CI flakes.

jhgaylor:
Closes #1592.

`FountainWeb.VerifyPendingLiveTest` "the poll advances even when the broadcast never arrives" failed once in eight full local suite runs on `main`, with the redirect simply not arriving in time.

Every advance off that page is out of band: the LiveView re-reads the row on a broadcast or its own poll, redirects, and the test proxy relays that to the test process. `assert_redirect/2` waits ExUnit's `assert_receive_timeout` — **100 ms** — for that whole round trip.

| how the file runs | advance latency | result |
|---|---|---|
| alone (measured over 20 advances, loaded machine) | 0–34 ms, median 3 | 25/25 runs green |
| inside a partition running 20 cases | over 100 ms sometimes | 1 failure in 8 full-suite runs |

Nothing here is racing: the redirect either happens or it does not, and 100 ms is a default rather than anything this page promises.

## The assertions still assert

Removing the redirect from `advance_if_verified/1` fails all three (11 tests, 3 failures) — the tests got more patient, not blinder.

Left as three explicit deadlines rather than a suite-wide `assert_receive_timeout` bump: these are the only `assert_redirect` calls in `apps/fountain/test`, and nothing else has been observed failing this way.

Third flake found by running the full suite repeatedly this afternoon, after #1587 and #1591.

https://claude.ai/code/session_01DBK3uwQuJozQHKmEuBrvrP


jhgaylor:
Opened against the wrong head branch — `docs/adr-0016-policy-boundaries` is another session's ADR change, not this fix. Reopening from `fix/verify-pending-advance-deadline`, which is where the commit actually is.

PR #1596, opened 2026-09-04
feat(api): git status on a sandbox, the read a diff cannot serve

jhgaylor:
`GET /api/sandboxes/:id/git-status` returns one entry per changed path in the
repository that holds `path`.

## Why

`git diff` compares tracked content. A file the agent created and never staged
is in no commit and not in the index, so **no combination of `staged` and `ref`
shows it** — it is invisible to the endpoint we have. That is the common case
for an app that renders "what changed": the agent writes a new file and the app
renders nothing. Deletions were always visible (a tracked file's removal is a
hunk), so the missing half was precisely the untracked one.

ADR 0039 decision 2 named `git status` as *the* example of a further
purpose-shaped read and set the procedure for one: a fixed script, a named
endpoint, the ADR amended. This is that; the amendment is in the same diff.

## Shape

```
GET /api/sandboxes/:id/git-status?path=repo&untracked=all

{ "path": "/home/sprite/repo",
  "repo_root": "/home/sprite/repo",
  "branch": "main",
  "untracked": "all",
  "truncated": false,
  "entries": [
    { "path": "notes.md",   "index": "untracked", "worktree": "untracked", "renamed_from": null },
    { "path": "lib/new.ex", "index": "renamed",   "worktree": "unchanged", "renamed_from": "lib/old.ex" }
  ] }
```

`index` and `worktree` are git's two porcelain columns read separately, so an
untracked file reads `untracked` in both and a file staged and then edited again
reports a state in both. Entry paths are relative to `repo_root`, **not** to the
sandbox home the way `/files` and `/file` are — that is the one trap for a
caller, and it is what git's porcelain reports.

## Decisions worth a reviewer's eye

- **`/git-status`, not `/status`.** A sandbox already has a `status` of its own
  (`ready`, `suspended`), and that is what a reader would expect back from
  `/api/sandboxes/:id/status`. This breaks symmetry with `/diff`; identity beat
  symmetry.
- **`--no-optional-locks` matters more here than for a diff.** A plain
  `git status` refreshes the index, so without it a *read* takes `index.lock`
  and breaks the agent's own commit. The read that observes the work must not be
  able to stop it.
- **One flag, `untracked`** (`normal` / `all` / `no`), mapped to a fixed
  `--untracked-files` value by a `case` in the script, so caller data never lands
  next to a `--`. It is the knob that decides whether a new directory arrives as
  one entry or as everything beneath it; on an unignored build tree that is the
  difference between 1 entry and 40,000.
- **The 403 is declared on the operation** rather than added to
  `SchemaGuardAllowlist`. That list is a ratchet for the debt #1432 already owes,
  and a new operation has no business growing it.
- **An internal 1 MiB byte cap** sits under the 2,000-entry cap, because `-uall`
  on a large tree is read into memory before anything counts it. A record the cap
  cuts in half is dropped rather than half-read, and `truncated` says so.

Everything else in ADR 0039 holds unchanged: full scope, path confinement,
redaction (a path is data the agent chose, so it passes the same replacement the
transcript's bytes do), never a wake, over `Managoat.Sandbox.exec/4` with the
path as a positional parameter, the same 30s timeout, no audit and no
`check_spend/1`. **No new error, no new exit code, no seam or adapter change** —
the script reuses `path_not_found`, `not_a_directory` and `not_a_repository`.

## Also in here

Two OpenAPI descriptions on `/diff` called the default comparison "HEAD". It is
the index, which is exactly why work the agent staged vanishes from a plain diff.
Corrected, and `docs/api.md` now says so and points at `ref=HEAD` for both halves.

## Verification

Every test in this module mocks `Managoat.Sandbox.exec`, so a self-consistent
wrong parser would pass. The script text was extracted and run against a real
repository before the parser was trusted:

- `-z` reverses git's rename arrow: the record is `XY <destination>\0<origin>\0`,
  with the origin as its own NUL-terminated record.
- `?? fresh/` under `normal` becomes `?? fresh/deep/z.txt` under `all`.
- A detached HEAD emits an empty branch line; a repository with no commits still
  reports its branch (`symbolic-ref` works there, `rev-parse HEAD` does not).
- Exit codes 3 / 4 / 6 all fire from the real script.

Those real bytes were then fed through `status/3`. Three behaviours were mutated
(rename-origin consumption, cut detection, the `?` mapping) and each was caught
by the test that should catch it.

Green locally: full suite 4288 + 144 + 33 with 0 failures, `credo --strict`,
sobelow, dialyzer, `format --check-formatted` under the pinned 1.19.2, all four
SDK contract verifiers, and the three prose gates (`docs-style.py`,
`vale lint docs`, `destink`).

## Not in here

`sdk/typescript/package.json` is **not** bumped. The `sandboxGitStatus` method,
the types and the manifest claim are all present, but merging a version bump
publishes, so that is a separate decision. Python, Swift and Elixir do not model
the route yet either — one manifest claim satisfies the contract gate, the same
way TypeScript went first for the original three in 1.15.0.

https://claude.ai/code/session_01EX14PWqfxWG55gPt42RAMk


BinaryBourbon (changes requested):
Reviewed `86eadba5` (current head). Specifically the unreviewed delta `e68b80b0..86eadba5` — which is the 2026-09-10 merge of `main` and contains no author change to this PR's own code, only main landing in `docs/api.md`, `sdk/contract/contract.json`, `sdk/typescript/src/generated/openapi.ts` and the SDK version — plus a full read of every changed file, the ADR 0039 amendment, and the repo's route gates. I re-ran the extracted `status_script/0` against real repositories rather than trusting the mocked suite, and the three findings below are reproductions, not readings.

**Requesting changes.** The gates are all met: `get_sandbox/2` is tenant-scoped and no `_unsafe_` call appears, a read audits nothing correctly, the operation is declared, the 403 / 422-not-a-repository / 409-parked / foreign-sandbox tests all exist, the enum guardrail rows are added, the contract and TypeScript manifest carry the claim, `docs/api.md` is an existing nav page, `sdk-no-release` legitimately skips the SDK gate for the deliberate non-bump, and CI is green. The argument passing is right — positional parameters, `untracked` mapped through a `case` and never adjacent to a `--`, `--no-optional-locks` for the reason given. What blocks is three reproducible wrong-output cases in the new code, the first of which is the single worst answer an endpoint named "what changed" can give.

## Findings

**1. A `git status` that fails after repository discovery returns `200 {"entries": [], "truncated": false}` — a clean tree.** `apps/fountain/lib/fountain/sandbox_files.ex:517`. The script's last command is a pipeline, so the script's status is `head`'s and is always 0, and `run/3` (line 398) accepts code 0 as success. Two prior reviews named this; it is confirmed rather than inherited from them. Reproduction: a repository with a corrupt `.git/index` — `git rev-parse --show-toplevel` succeeds, `git status` writes `fatal: .git/index: index file smaller than expected` and exits nonzero, and the whole script exits **0** with only the two header lines on stdout. `parse_status/1` then yields an empty body, `status_records/1` a `tail` of `""`, and the response is `entries: []`, `truncated: false`. It is byte-for-byte what a clean tree returns. The realistic trigger is not a corrupt index but a clean/smudge filter that fails — an agent that clones an LFS repository into a sandbox without `git-lfs` on PATH is exactly this — and the app renders "nothing changed" over a tree the agent has rewritten. Note the diagnostic is not even available to a caller: `Managoat.Sandbox.exec/4` defaults `stderr_to_stdout` to `false` on every adapter and `run/3` does not set it, so the `fatal:` line is discarded, not merely truncated. `{:sandbox_command_failed, …}` is unreachable for any git failure on this route. One caution on the fix: a bare `set -o pipefail` turns the intended truncation into a failure too, because `head -c` closing the pipe gives `git` a SIGPIPE and status 141. Capturing the exit code separately, or excluding 141, is what this needs.

**2. Repository discovery is not confined, so entries can come from outside the sandbox.** `apps/fountain/lib/fountain/sandbox_files.ex:509`. `resolve_path/2` confines the caller's *request* path; `git rev-parse --show-toplevel` then walks up the ancestors of that path, and nothing checks the discovered root against `roots/1`. Reproduction: with a repository at `<X>` and the request path `<X>/home/sprite` (confined, inside `roots/1`), `untracked=all` returned `repo_root: <X>` and entries `.ssh_id_rsa_name` and `Documents/taxes.pdf` — two files outside the requested directory entirely. On Sprites this is inert because no ancestor of `/home/sprite` is a repository. On the self-hosted runner (ADR 0022) a sandbox is a directory under `~/.fountain/runners/<name>/sandboxes/`, so an operator whose `$HOME` is a dotfiles repository hands any full-scope key — including a third-party OAuth app the user authorized — an enumeration of the file names in their home directory. The moduledoc's "Every path is confined to the sandbox home" and ADR 0039 decision 4 both say this does not happen. `/diff` on `main` walks up the same way, so the fix belongs to both, but `/diff` reports tracked-content hunks while `-uall` here enumerates every untracked name in the ancestor tree. A prefix check on `root` against `roots/1`, failing as `exit 6`, closes it.

**3. A newline in the repository root's path corrupts the header and silently drops an entry.** `apps/fountain/lib/fountain/sandbox_files.ex:511` and `:565`. Records are NUL-framed deliberately ("a name with a tab or newline survives" is a test for `list/2`), but the two header lines are newline-framed. Reproduction: a repository at `<S>/re\npo` with one untracked `new.txt` produces `<S>/re\npo\nmain\n?? new.txt\0`; `parse_status/1` splits it into `root: "<S>/re"`, `branch: "po"`, body `"main\n?? new.txt\0"`, whose one record decodes as index `m` / worktree `a` and is dropped by `decode_record/1`. Result: a 200 with a truncated `repo_root`, a `branch` that is a fragment of a directory name, and `entries: []` with `truncated: false` for a repository that has a change. A caller may address such a directory — `resolve_path/2` refuses only NUL and invalid UTF-8. NUL-framing the header, as the records already are, fixes it.

## Non-blocking

- `sandbox_files.ex:346` — `repo_root` is the one string in this response that skips both `redact_with/2` and `to_text/1`, while `branch`, `entry.path` and `renamed_from` three lines below get both. It is a path the agent chose, on the same footing as the ones that are redacted, and the PR body's own claim is that "a path is data the agent chose, so it passes the same replacement the transcript's bytes do". `/diff` has the same gap at line 291.
- `sandbox_files.ex:346` again — on the runner's process backend `run/3` maps the directory through `Managoat.Sandbox.host_path/2`, so `repo_root` is the **host** path while `path` in the same response is the sandbox path. The ADR amendment tells a caller to combine `repo_root` with an entry path; on a runner that produces `422 path_outside_sandbox` from `/file`. Shared with `/diff`, but only here is `repo_root` load-bearing.
- `sandbox_files.ex:79` — `@states` maps `!`, so `change_states/0` publishes `ignored` in the OpenAPI enum, the wire contract and the generated TypeScript union. The script never passes `--ignored` and `--untracked-files` does not imply it, so no response can carry the value. The safe direction (no reachable letter missing) holds; it is a dead branch in a client's exhaustive match and a contract that says something untrue.
- The 403 declaration at `sandbox_files_controller.ex:188` is harmless and changes no wire behaviour, but its comment is wrong on the facts: `FountainWeb.SchemaGuardAllowlist` holds one entry, `{"GET /api/conversations/{id}", 200}`, and the three sibling reads are not on it. 401/403/429 are composed from `:api` pipeline membership, which is why the siblings declare none. Worth correcting the comment or dropping the declaration for consistency across the controller's four operations.
- `sandbox_git_status` on `main` without a published `1.23.0` means npm's `sandboxGitStatus` lags the manifest claim. Deliberate, labelled and explained, so noted only.

The shape of the endpoint is right, the tests are unusually good for one that has to mock `exec`, and the ADR amendment follows decision 2's own procedure. Findings 1 and 3 are small edits to `status_script/0` and `parse_status/1`; finding 2 is a prefix check. I will re-review promptly.

jhgaylor:
Thanks — all three were reproducible exactly as written, and all three were in the shell rather than in what parses its output. Pushed as five commits, `c4fc98ee..0e93bb17`.

The suite mocked `exec/4`, so it could only ever prove the parsing. `SandboxFiles.script/1` now hands the scripts out (`@doc false`) and `apps/fountain/test/fountain/sandbox_files_script_test.exs` runs them under a real bash against real repositories built in `Fountain.TmpDir`. Every fix below was reverted after its test was written, and the revert result is quoted.

## 1. A failed `git status` returned a clean tree — `c4fc98ee`

`${PIPESTATUS[0]}` is captured instead of the pipeline's status. **141 passes with 0**, because that is the SIGPIPE `head -c` causes when it closes the pipe at the byte cap, which is the truncation the script asks for — a bare `set -o pipefail` would have turned every truncated read into a failure. Anything else is `exit 8`, which falls to the existing catch-all: `{:sandbox_command_failed, 8, output}`, a 422.

On the stderr question, deliberately: the message travels, collected by a **second git run on the failure path only** that reads stderr and discards stdout. Setting `stderr_to_stdout: true` was the other option and is worse — it would interleave a git warning into the NUL-framed records on the *success* path, trading a correct common path for a message on a rare one. A second invocation of a command that has already failed costs nothing anyone will notice, and the fallback controller's own comment says output travels "because the exit code alone sends the caller guessing".

Tests: a corrupt `.git/index` (the LFS-filter case reproduced without a second binary) must answer 8 with `fatal:` in the output; a healthy repository answers 0 with its two changes; and 1,000 files with 240-character names push the output past a pipe buffer so `head -c 64` really does kill git with 141 — that one guards the trap you named.

**Revert:** `assert code == 8` fails with `left: 0, right: 8`.

## 2. Repository discovery was not confined — `e0961cb5`

`roots/1` now trails the positional arguments of both git scripts, so it crosses `host_path/2` with everything else and the comparison is host-to-host on a runner. The discovered root must equal a root or sit under one, else `exit 6` — following the other codes, and `not_a_repository` is the honest answer: there is no repository *here*, and naming the one found above would answer the question the check exists to refuse. The physical path is compared too (`cd && pwd -P`), because `--show-toplevel` resolves symlinks; my own mac test hit that immediately (`/var` → `/private/var`).

**`/diff` is fixed in the same commit**, not left for a tracker: it is the same four lines in an adjacent script in a file this PR already rewrites, and leaving a known walk-up next to a fixed one is not a smaller change.

Tests: a sandbox directory under a repository home refuses with 6 for both `/git-status` and `/diff` and does not emit `.ssh_id_rsa_name`; the same discovery with the containing root declared answers 0; one root out of several is enough; a root whose own name contains a newline is compared whole. Plus a mocked test that the roots cross `host_path/2`, since an unmapped root would refuse *every* repository on a runner.

**Revert:** both refusal tests return 0 with the outer repository's entries, including `.ssh_id_rsa_name`.

## 3. A newline in the root corrupted the header — `562eb877`

Both headers are NUL-framed now, like the records, and `parse_status/1` and `parse_diff/1` split on it. Trimming the root went with the newline, so a directory whose name ends in a space survives as well. `/diff`'s single header line had the same defect and is framed the same way.

**Revert** (both scripts back to `printf '%s\n'` and both parsers back to splitting on it): the mocked test reports your reproduction verbatim — `branch: "po"`, `entries: []` — and the script test against a real repository named `re\npo` shows the whole header plus the first record arriving as one field.

## Non-blocking

Taken:

- **`repo_root` redaction** (`5fd51c88`) — it was the one string in either response skipping `redact_with/2` and `to_text/1`. Both readers now compute the identity's values once and put the root through them; `/diff`'s gap at line 291 went with it. Reverting the two lines fails the redaction test in each `describe`.
- **The 403 comment** (`5fd51c88`) — corrected rather than dropped. The declaration is already in the contract and full scope is the whole story of this read; the comment now says that 401/403/429 are composed from `:api` pipeline membership, which is why the three siblings declare none, and stops misdescribing `SchemaGuardAllowlist`.
- **ADR 0039** (`0e93bb17`) — the amendment now says confinement covers the path the caller never names, with the runner's dotfiles `$HOME` as the named case. Decision 4 was true of the request path and a reader would have stopped there. `okf validate decisions` is unchanged (4 warnings before and after); no frontmatter change, so the index is untouched.

Left, with reasons:

- **`@states` publishing `ignored`.** Your reading is right — no response can carry it. I left it because narrowing a published enum is not a comment-sized change: it means the OpenAPI enum, `contract.json`, the TypeScript union and the guardrail rows moving together on a PR that is deliberately not bumping an SDK, to remove a value that is safe by construction. It is also the value `untracked` would gain if `--ignored` ever became a fourth mode. Happy to do it here if you would rather have it now than tracked.
- **`repo_root` as a host path on the runner's process backend.** Not small, and not this PR's: it is the `host_path/2` round trip missing an inverse, shared with `/diff`, and fixing it means a mapping back that the seam does not have today.

## Verification

`mix format --check-formatted`, `mix compile --warnings-as-errors` and `mix credo --strict` (no issues, 772 files) all from the pinned toolchain, and the full `apps/fountain` suite from that directory: **6 doctests, 4740 tests, 0 failures, 2 skipped**. No OpenAPI-visible string changed, so the contract and the generated TypeScript are untouched and `sdk-no-release` still holds.

BinaryBourbon (approved):
Re-reviewed `86eadba5..0e93bb17` — the five fix commits — plus a full re-read of `sandbox_files.ex`, the new script suite, and the updated controller and context tests. I re-ran my three original reproductions against the scripts extracted verbatim from `0e93bb17`, and added four cases the fixes make newly interesting.

**Approving.** All three blocking findings are fixed, and fixed at the layer where they lived rather than papered over in the parser. `/diff` was fixed alongside `/git-status` rather than deferred, the tests were tightened rather than adjusted to pass, and the new `sandbox_files_script_test.exs` runs the real shell against real repositories — which is the gap that let all three defects through a suite that mocks `exec/4`. 21 checks green, 4 skipped.

## My reproductions, re-run against `0e93bb17`

| Case | Before | Now |
|---|---|---|
| Corrupt `.git/index` | exit 0, `entries: [], truncated: false` | **exit 8**, stdout carries `fatal: .git/index: index file smaller than expected` → `{:sandbox_command_failed, 8, …}` → 422 |
| Ancestor repo above the sandbox root, `untracked=all` | exit 0, `repo_root` = the ancestor, entries `.ssh_id_rsa_name` and `Documents/taxes.pdf` | **exit 6**, empty output → `:not_a_repository` → 422, and it names nothing about what it found |
| Repository at `<home>/re\npo` | `repo_root` cut to `<home>/re`, `branch: "po"`, the one real entry dropped | **root arrives whole** with its newline, `branch: "main"`, `?? new.txt` present |

And the four I added:

- **Truncation is still truncation.** With 1,500 long-named files and `n=64`, `${PIPESTATUS[0]}` is genuinely 141 (measured, not assumed — under a pipe buffer it is 0), and the script exits 0. The SIGPIPE trap I warned about is handled by `0|141)` rather than by a bare `set -o pipefail`.
- **A sandbox root reached through a symlink still answers.** With `<S>/sprite-link -> <S>/physical/sprite` and the root passed as the link, `--show-toplevel` returns the physical path and the `pwd -P` half of the comparison matches it: exit 0, not a spurious `exit 6`. Same for the case where the root *is* the repository.
- **The `pwd -P` half is what carries the macOS runs of the new suite** (`/var` → `/private/var`), so it is exercised there; on Linux CI the logical branch carries them. Both halves matter and neither is dead.
- **Host-to-host comparison holds.** `run/3` maps every argument through `Managoat.Sandbox.host_path/2`, and the trailing roots go through the same call as the request path, so on the runner adapter (`/home/sprite` prefix → the daemon's real directory, `/tmp/...` through unchanged) both sides of the comparison are in the same namespace. `sandbox_files_test.exs` pins this with a stubbed `host_path`.

## The rest of what you asked me to check

- **`exit 6` and `exit 8` map cleanly.** 8 is not among `@exit_missing`/`wrong_kind`/`unreadable`/`not_repository`/`ref_not_found` (3–7), so it falls to `run/3`'s catch-all as `{:sandbox_command_failed, 8, redact(sandbox, output)}`, which `FallbackController` already renders as 422 with `exit_code` and `output`. No new error word, no new response, and the 422 is already declared on the operation. Routing an out-of-sandbox root to the same `not_a_repository` as a plain directory is the right call: there is no repository *here*, and naming the one found above would answer the question the check exists to refuse.
- **The second git run is not a second read path.** It runs after the confinement check, in the same already-confined repository, with `2>&1 >/dev/null` so only stderr reaches the pipe, and its output is redacted by the catch-all. Rejecting `stderr_to_stdout: true` was right for the reason given — it would interleave a git warning into the NUL-framed records on the success path.
- **`/diff`'s response shape is unchanged.** `repo_root` is still a string; the NUL framing is internal between script and parser, and dropping `String.trim` on the root only stops it eating a character a directory name may legitimately end with. What *does* change for an endpoint already on `main` is behaviour, deliberately: a `/diff` whose repository was discovered above the sandbox now returns `422 not_a_repository` instead of that repository's hunks, and `repo_root` is now redacted. Both are the fix, and both are worth a line in the release notes since a caller could be depending on the old answer.

## Non-blocking

- **`/diff` still cannot tell a failed `git diff` from an empty one.** `git diff … | head -c | base64` leaves the script's status as `base64`'s, so the corrupt-index repository that now returns 422 from `/git-status` returns `200 {"diff": ""}` from `/diff`. The machinery to fix it is now three lines away in the sibling script. Pre-existing on `main` and outside what I blocked on, but it is the last instance of the class and the two endpoints now disagree about the same repository.
- **`sandbox_files.ex:607` — the 4 KiB stderr cut can produce a 500 instead of a 422.** `head -c 4096` cuts on a byte boundary; if git's diagnostic exceeds 4 KiB and the cut splits a multibyte character, `output` reaches `json/2` as invalid UTF-8 and Jason refuses to encode it. Narrow (it needs a long, non-ASCII diagnostic straddling exactly that byte), and it degrades an already-failing read rather than corrupting a good one, but `to_text/1` on the catch-all's output in `run/3` closes it for every script at once.
- **The 422 body on exit 8 leads with the NUL-framed header**, because `printf` runs before the pipeline — so `output` is `<root>\0<branch>\0` plus any records git managed to write, and then the message. Readable after JSON escapes the NULs, but a caller looking for git's sentence has to find it.
- **The two deferred items are fine to leave.** `ignored` is an over-wide enum in the safe direction, and narrowing a published enum in `contract.json` and the TypeScript union is a contract change that deserves its own decision rather than a ride on this one. `repo_root` as a host path on a runner is unchanged from `/diff` on `main` and is a shape problem, not a leak of anything new; it is now the only one of my five findings untouched, and it would be worth an issue so it does not become folklore.

The comment above `status_script/0` is unusually good: it records why `set -o pipefail` is wrong here, why the stderr is collected by a second run, and what the confinement check is actually defending against. That is the part most likely to stop this being re-broken.

jhgaylor:
Pushed `5c081faa` for the non-blocking `sandbox_files.ex:607` finding — the 4 KiB stderr cut producing a 500 instead of a 422. **This supersedes the approval; the PR needs a fresh look.**

**What changed.** `run/3`'s catch-all now builds its error through a `command_failed/3` helper that applies `to_text/1` to the output, so it covers all four scripts rather than the status one. The `case` clauses stay one line each (an inline call was 101 columns and the formatter would have exploded every clause in the block).

```elixir
{:ok, output, code} -> {:error, command_failed(sandbox, code, output)}
...
defp command_failed(%Sandbox{} = sandbox, code, output),
  do: {:sandbox_command_failed, code, to_text(redact(sandbox, output))}
```

**`to_text/1` does close it.** It is `if String.valid?(bytes), do: bytes, else: :unicode.characters_to_binary(bytes, :latin1)`. Every byte 0x00–0xFF is a valid latin-1 codepoint, so the recode cannot fail and cannot return an error tuple for any binary — the result is always encodable. A cut-in-half UTF-8 message comes back as mojibake rather than as a 500, which is the right trade for a diagnostic that was already truncated.

**Redaction ordering.** Redaction must run first, and now does. `redact_with/2` is `:binary.replace/4`, which matches raw bytes and is indifferent to UTF-8 validity — so it is safe on the truncated binary. Recoding first would rewrite a non-ASCII secret's bytes (`é` C3 A9 → C3 83 C2 A9) out of its own match and silently stop redacting it. This is the same order `redact_change/2` already uses for `path` and `renamed_from`.

**One thing I found and did not fix**, because it is pre-existing and class-wide rather than this branch's: truncation happens in the shell, before redaction in Elixir. A secret straddling the byte cap has its prefix survive the cut and so survives redaction, in this script's `head -c 4096` and equally in the read and diff caps. Narrow (the secret has to land on the boundary), not new here, and fixing it means moving the cap into Elixir for every script. Worth an issue rather than a ride on this PR — say the word and I will file it.

**Revert evidence.** Two new tests in `sandbox_files_test.exs`, both constructing the boundary cut deliberately (`binary_part(msg, 0, 4096)` with an odd-length prefix so the 4096th byte lands mid-character, and `refute String.valid?(cut)` pinning that premise):

- *a diagnostic the byte cap split mid-character still encodes as JSON* — asserts `String.valid?/1` and that `Jason.encode/1` succeeds on the body.
- *a failing command's output is redacted before it is recoded* — a secret of `sk-live-café` in the cut message, asserting it is `[REDACTED]`. This one fails if the two steps are ever swapped.

With `command_failed/3` reverted to the old `redact(sandbox, output)`, both fail on `assert String.valid?(output)` (`39 tests, 2 failures`). With the fix: **64 tests, 0 failures** across `sandbox_files_test.exs`, `sandbox_files_script_test.exs` and `sandbox_files_controller_test.exs`.

`mix format --check-formatted`, `mix compile --warnings-as-errors` and `mix credo --strict` clean on the mise-pinned 1.19.2. No OpenAPI-visible string changed and no SDK version bumped — the diff is one private helper, its comment, and two tests.

BinaryBourbon (approved):
Re-reviewed `0e93bb17..5c081faa`, the one commit that fixes the recoding finding from my last pass. I checked the two claims it rests on by running them, not by reading them, under the pinned 1.19.2. CI was still running on this head when I posted; the gate has to be green before merge either way.

**Approving.** The fix is right, in the right order, and the order matters more than it looks.

## The two claims, measured

**`to_text/1` is total for this input.** `:unicode.characters_to_binary(bytes, :latin1)` returned a binary for every one of the 256 single bytes and for all 65,536 two-byte sequences — exhaustive over the alphabet, so no `{:error, _, _}` or `{:incomplete, _, _}` is reachable from a binary here. The reasoning holds.

**Redaction before recoding is load-bearing, and the wrong order leaks visibly.** With `secret = "sk-live-café"` inside a 4,096-byte cut that splits a trailing `é`:

- redact then recode → valid UTF-8, `[REDACTED]` present, no occurrence of the secret;
- recode then redact → valid UTF-8, **no `[REDACTED]` at all**, and the body reads `fatal: unable to read 'sk-live-cafÃ©'`.

So the failure mode of the other order is not a mangled near-miss a reader would notice. It is the secret's ASCII head in the clear with one mojibake character on the end, and `:binary.replace/4` silently matching nothing because the C3 A9 it was looking for became C3 83 C2 A9. The comment on `command_failed/3` says this; it is worth keeping exactly as written, because this is the kind of ordering a later formatter-driven refactor flips without noticing.

The two new tests pin both halves, and `refute String.valid?(cut)` pinning the premise is the right instinct — without it the test passes for the wrong reason the day someone changes the cap. Extracting `command_failed/3` rather than inlining is the correct call for a reason that has nothing to do with columns: it gives the ordering a name and a place for the comment.

## The truncation-before-redaction question: **it does not block this PR**, and it needs an issue of its own today

Explicitly, since you asked for a verdict rather than a hedge.

**Why it does not block here.** The only cut this branch adds is the fixed `head -c 4096` on git's diagnostic. It is not caller-steerable — a caller cannot move it, so exposing a secret through it means arranging for git to place that secret across byte 4,096 of its own error message. And the branch's other cap is safer than the siblings: `status_records/1` *drops* a record the 1 MiB cap cut in half rather than returning it, so `/git-status` entry paths cannot leak a prefix at all. Holding three verified security fixes to rework every cap in the module is the wrong trade, and it is not what this PR is.

**Why it is worse than "a prefix leaks", on `/file`.** I confirmed the straddling case by construction: a secret spanning the cut leaves `…xxxxsk-liv` in the output, un-redacted, because `:binary.replace/4` is looking for the whole value. On `/file` that cut position is the caller's to choose. `read/3` takes `opts[:max_bytes]` from `params["max_bytes"]` (`sandbox_files_controller.ex:103`), clamps it to 1…4 MiB, and hands it to `head -c "$n"` in the shell; redaction then runs over only the bytes that survived. So `GET /api/sandboxes/:id/file?path=.env&max_bytes=N`, walked N by N, returns one more byte of a secret each time and never triggers a replacement until N covers the whole value. That is not a truncation artifact, it is a byte-at-a-time read of the thing ADR 0039 decision 5 exists to prevent — and the moduledoc names it in so many words: "The `.env` file is on that disk in plaintext, so this is what keeps a third-party app holding the user's key from reading the user's secrets back through it."

None of that is this branch's doing. `/file`, `max_bytes` and that ordering all shipped in the original ADR 0039 PR and are on `main` in SDK 1.15.0, untouched here. But it should not leave this review as "acceptable to leave", because the sweep is a different severity from the prefix leak it was described as. Please file it against `/file` before this merges, so the finding does not go out with the branch that happened to surface it. The fix is the one the agent named — move the cap into Elixir, redact, then cut — and on `/file` it is a small change: ask the script for `max_bytes + overlap` (the longest secret, which `secret_values/1` already sorts by size), redact, then `binary_part` to the cap.

## Standing non-blockers, unchanged

`/diff` still cannot tell a failed `git diff` from an empty one — its pipeline status is `base64`'s, so the corrupt-index repository that now returns 422 from `/git-status` returns `200 {"diff": ""}`. The `ignored` enum value and `repo_root`-as-host-path on a runner remain deferred, and I still think both are fine to leave.

PR #1627, opened 2026-09-06
fix: declare API failures, verify alerts, and use generated reference

### LINKED ISSUE #912: Generate the API reference instead of hand-maintaining it
The CLI half of this issue shipped in #916 and #928. This issue now tracks the
remaining API-reference work. The parent docs-redesign tracker, #903, is
complete.

## Problem

`docs/api.md` is now 1,137 hand-written lines alongside an OpenAPI spec that CI
already validates. It repeats endpoint names, parameters, request shapes and
response details that the server already describes. Every API change therefore
has two sources to keep synchronized.

This is not hypothetical maintenance: 25 commits touched `docs/api.md` between
the original filing on 2026-08-21 and 2026-09-03.

The generated foundations already exist:

- `GET /api/openapi.json` serves the OpenAPI 3.1 spec generated from code.
- `GET /api/docs` serves Swagger UI over that spec.
- `FountainWeb.ApiSpecTest` walks the router and fails when an `/api/` route has
  no spec entry.
- `mix openapi.export` renders the spec for generated SDK artifacts.

Generated reference is a useful way to keep documentation faithful to code.
The human-written page should explain how to use the API and organize it around
reader tasks; it should not maintain a second copy of the contract.

## Outcome

Make the OpenAPI description the source for endpoint-level reference, while
keeping a concise, hand-written API index for authentication, workflows,
examples and other guidance that does not belong in the schema.

Whether the published endpoint reference is generated Markdown or the existing
Swagger UI is an implementation choice. The requirement is that
`docs/api.md` stops duplicating contract facts by hand and readers can move
cleanly between the human guide and the complete generated reference.

## Done when

- Endpoint, parameter, request and response reference comes from the OpenAPI
  spec, not hand-maintained inventories in `docs/api.md`.
- `docs/api.md` remains a human-written guide/index containing authentication,
  workflows, examples and links into the complete reference.
- The generated reference is published and reachable from the documentation
  navigation.
- CI fails if any checked-in generated output is stale.
- Inbound documentation links and anchors are updated, and the strict docs
  build and link tests pass.

## Completed CLI work

- #916 made the CLI parity tests read `docs/cli/*.md`.
- #928 generated the flag-complete CLI reference from the Cobra command tree
  and added a staleness check.

### LINKED ISSUE #1168: Alert on turn-level and provider failures, not just provisioning
## Gap

The only stage the alert rules watch is `provision` (`FountainProvisionFailures`, `FountainProvisionDeadlineExceeded`). Everything after it is exported and unwatched:

- `fountain_stage_count{stage=...}` for setup scripts, clone, packages, network policy
- `fountain_turn_completed_duration_ms`, `fountain_turn_first_output_elapsed_ms`
- `fountain_reattach_stop_duration`, `fountain_rehydrate_stop_*`
- `fountain_sandboxes_by_provider`, `fountain_sandbox_reclaimed`, `fountain_sandbox_suspended`

An E2B or Daytona outage that fails every turn after the sandbox comes up, or an ACP adapter that never produces output, shows up only in Sentry.

## Proposed rules (in `deploy/k8s/prometheusrule.yaml`)

- `FountainStageFailures`: `sum by (stage) (increase(fountain_stage_count{status="failed"}[30m])) > 2` for any stage, replacing the provision-only rule (keep the provision one at its current sensitivity if you prefer)
- `FountainTurnFailureRate`: failed / completed turns over 30m > 25% with at least N turns, per provider if the label exists
- `FountainReattachFailures`: reattach failures > 0 over 1h (the #801-#803 incident marked sandboxes failed on a false reattach and the reaper destroyed them; that would have been silent)
- `FountainTurnNoFirstOutput`: p95 `fountain_turn_first_output_elapsed_ms` > threshold for 15m

Counters that never fired have no series, so any ratio must end in `or vector(0)`.

Found during the 2026-08-25 prod alert-coverage review.

### LINKED ISSUE #1432: 68 operations do not declare the statuses their pipeline plugs return

jhgaylor:
OpenAPI omitted errors returned by shared pipelines and several controllers, while the manual duplicated endpoint details. This declares those responses, makes Swagger the endpoint reference, and adds tested alerts for failures after provisioning.

- Derive auth, scope, rate-limit, and content-negotiation responses from router membership, including extension dispatch. Preserve controller-specific schemas and declare local refusals explicitly. Remove all 66 missing-status exceptions from the schema guard and the SDK conformance linter's blanket status exemption.
- Regenerate the wire contract and TypeScript types. The semantic comparison adds 538 responses without changing any existing operation field or schema. TypeScript 1.20.2 carries the corrected types; request behavior is unchanged.
- Add stage, reattach, per-provider turn-failure, and first-output latency alerts. Promtool fixtures cover absent series, quiet traffic, provider isolation, counter resets, exact thresholds, and the latency alert's hold time. CI runs the fixtures.
- Replace the endpoint inventory in `docs/api.md` with a workflow guide linking to the instance's generated reference. Preserve all 29 existing section anchors and expose the reference through documentation navigation.

Validation completed:
- API contract, pipeline, and docs structural tests pass.
- 101 TypeScript tests, typecheck, build, and all four SDK contract checks pass.
- 45 Promtool cases and 13 CI policy tests pass; 24 conformance scenarios validate.
- Generated-contract staleness check, SDK release gate, docs style, and destink pass. STE report reviewed; remaining warnings name technical terms.
- `mix precommit` on the implementation passes, including static analysis and production release assembly: core 4,300 tests plus 6 doctests, Buzz 144 tests, Support 33 tests, zero failures.

After merging current main, all 34 deployed-runner tests and 95 affected SSE, timestamp, API, and docs tests pass. The local database was migrated to main’s new microsecond timestamp schema before that integration run.

The portable alerts need deployment and an alert receiver to become operational. First-output latency measures output that arrives; it does not detect a turn that emits no bytes. No production configuration is changed by this PR.

Fixes #1432.
Fixes #1168.
Fixes #912.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1757,6 +1757,14 @@ jobs:
       - name: Test CI decisions
         run: python3 -m unittest discover -s scripts/ci -p 'test_*.py' -v
 
+      - name: Install alert validation tools
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y --no-install-recommends prometheus python3-yaml
+
+      - name: Validate portable alerts
+        run: /usr/bin/python3 scripts/test-alerts.py
+
   gate:
     name: CI required
     runs-on: ubuntu-24.04
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1757,14 +1757,6 @@ jobs:
       - name: Test CI decisions
         run: python3 -m unittest discover -s scripts/ci -p 'test_*.py' -v
 
-      - name: Install alert validation tools
-        run: |
-          sudo apt-get update -qq
-          sudo apt-get install -y --no-install-recommends prometheus python3-yaml
-
-      - name: Validate portable alerts
-        run: /usr/bin/python3 scripts/test-alerts.py
-
   gate:
     name: CI required
     runs-on: ubuntu-24.04
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ upgrade, is in
 
 ### Changed
 
+- OpenAPI operations declare shared pipeline failures and controller refusals. The schema guard no longer exempts missing response statuses.
+- The API manual is a workflow guide linking to the generated reference at `/api/docs`; existing section anchors remain available.
+- Portable Prometheus rules cover stage and reattach failures, per-provider turn failure rates, and slow first output. Thresholds have executable alert fixtures.
+
 - Manifest apply rejects unknown `spec` keys before writing that resource or its secrets. Previously, Ecto silently discarded them, including misspelled network restrictions. Correct these keys before upgrading. Bulk apply keeps its HTTP 200 response with per-resource errors; other valid resources still apply. Ownership keys remain ignored.
 
 - **Credential brokerage is on for every account on the hosted platform**
--- a/apps/fountain/lib/fountain_web/api_spec.ex
+++ b/apps/fountain/lib/fountain_web/api_spec.ex
@@ -6,7 +6,7 @@ defmodule FountainWeb.ApiSpec do
 
   alias OpenApiSpex.{Components, Info, OpenApi, Paths, SecurityScheme, Server}
   alias FountainWeb.{Endpoint, Router}
-  alias FountainWeb.ApiSpec.Compose
+  alias FountainWeb.ApiSpec.{Compose, PipelineResponses}
 
   @behaviour OpenApi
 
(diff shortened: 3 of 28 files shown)


BinaryBourbon (approved):
Recording the maintainer’s explicit authorization to submit this approving review as BinaryBourbon on their behalf and merge after green CI. Verified CI required, coverage, secret scanning, SDK release checks, and alert tests on this commit.

PR #1675, opened 2026-09-06
feat(api): bulk apply reconciles Teammate, Schedule and Webhook documents

### LINKED ISSUE #1636: Bulk apply: accept Teammate, Schedule and Webhook documents so one manifest reconciles the whole estate
## Context

`POST /api/apply` reconciles a manifest of `Environment`, `Vault` and `Agent` documents by name, in a fixed order, in one request. `fountain apply -f` and chant's `fountainApply` activity both use it, and that is the whole point: one manifest under version control declares a tenant's estate, and applying it is idempotent.

Three kinds a declared estate needs are not in the manifest vocabulary yet. Team membership (`POST /api/team`, `TeamAddRequest`: agent, name, environment, vault), schedules (`POST /api/team/:agent_id/schedules`, `TeamScheduleCreateRequest`: cron, prompt, name, one_off, enabled), and webhook endpoints (`POST /api/webhooks`, `WebhookEndpointCreateRequest`: url, event_types, description). Today a client has to apply the manifest, then diff and drive those three by hand through their own routes. chant is about to declare exactly that shape, an op as a teammate with a schedule and a webhook, for every environment it manages.

## Proposal

Extend `ApplyRequest.resources` with three kinds, reconciled after the existing three in this order: `Teammate`, `Schedule`, `Webhook`.

- `Teammate`: keyed by `name`; `agent`, `environment`, `vault` resolve by name against the manifest or the tenant, as an agent's `environment` does today. Re-applying updates the environment and vault binding; removing the document does nothing unless the caller opts into prune.
- `Schedule`: keyed by `(teammate, name)`; fields as `TeamScheduleCreateRequest`. Re-applying updates `cron`, `prompt`, `one_off`, `enabled`.
- `Webhook`: keyed by `url`; fields as `WebhookEndpointCreateRequest`. The secret is returned on first creation only, as the create route does, and never on update.
- `ApplyResult` rows carry the new kinds with the same `created | updated | unchanged | error` verdicts.
- The `fountain apply` CLI and the SDK's `apply` accept the documents without change beyond the schema.

## Acceptance criteria

- [ ] A manifest holding all six kinds applies in one request in the documented order, and a second identical apply reports every row `unchanged`.
- [ ] A `Teammate` whose `agent`, `environment` or `vault` names nothing in the manifest or the tenant fails that row and no other.
- [ ] A `Schedule` with an invalid cron fails validation with the same error the create route gives.
- [ ] A `Webhook` document created through apply returns its secret once, in that apply's result row, and an update never returns it.
- [ ] The audit trail records the applied rows as `team.*`, `team.schedule.*` and `webhook.*` actions with the request's attribution.
- [ ] `docs/api.md` under "Bulk apply" and `docs/cli.md` under "Apply manifests" list the new kinds, and `openapi.json` carries their schemas so downstream generators pick them up.

## Dependencies and scope

No dependency on other issues. Prune semantics for the new kinds can follow the same opt-in the existing kinds use, or land later; either is fine as long as the default is additive.

### LINKED ISSUE #1680: Bulk apply reports `updated` for a document that changed nothing, so a client cannot tell a no-op from a write
Measured today against a v0.16.0 instance while checking that a chant-authored manifest applies idempotently.

Applying the same manifest twice, unchanged, the second apply reports:

```
Environment/toolchain   updated
Vault/prod-creds        updated
Agent/prod-steward      updated
```

`ApplyResult.action` is `created | updated | error`, so a document that reconciled to exactly what was already stored is indistinguishable from one that wrote a real change. A client that wants to report "nothing drifted" has to diff the estate itself, or read every resource back and compare.

The ask is one more value, `unchanged`, returned when reconciliation produced no field change. Everything else about the endpoint already behaves idempotently, which is what makes the reporting gap visible.

Context for why a client cares: chant renders an apply summary a person reads before deciding whether an environment drifted, and its own reconciler for the three team-side kinds (INTENTIUS/chant#2127) already reports `unchanged` because it compares before writing. The three kinds that go through bulk apply cannot, so one summary carries two different meanings for the same word.

Related: #1636, which extends bulk apply to Teammate, Schedule and Webhook. If those land here, they will want the same third value.

lex00:
`POST /api/apply` reconciles a manifest of `Environment`, `Vault` and `Agent` documents, and everything else a declared estate needs (team membership, schedules, webhook endpoints) had to be driven by hand through their own routes afterwards. This PR extends the manifest vocabulary so one apply reconciles the whole estate.

- Add `Teammate`, `Schedule` and `Webhook` kinds, reconciled after the existing three in that order. A Teammate resolves `agent`, `environment` and `vault` by name against the manifest first, then the tenant; a Schedule is keyed by `(teammate, name)`; a Webhook is keyed by `url` and returns its signing secret only on the apply that creates it.
- Add an `unchanged` verdict for all six kinds (#1680), so a second identical apply reports every row unchanged. The environment, vault, agent and webhook contexts now record no audit event on a no-op update, which is what the audit rules require and what schedules already did.
- Add `Team.update_teammate/4` to move a teammate's environment and vault binding. It refuses while a turn is running on the teammate's computer and retires the home the old identity named, mirroring `Agents.update_agent/3`.
- Fail a Teammate row whose reference resolves to nothing, two Teammate documents that claim the same agent, and a Schedule whose teammate name is ambiguous, each without stopping the rest of the manifest. Every per-document reconcile is now isolated from exceptions.
- Extend the OpenAPI schemas, regenerate the SDK contract and TypeScript types, update the Swift model, extend the Go CLI's kind list and legacy fallback, and document the kinds under "Bulk apply" and "Apply manifests".

Prune is left out; apply stays additive. The audit actions keep their existing names (`team.*`, `team.schedule.*`, `webhook_endpoint.*`) rather than the `webhook.*` the issue wrote.

Closes #1636
Closes #1680

Validation:

- `go test ./...` and `go vet ./...` in `cli/`; the SDK contract check, the TypeScript and Python verifiers and the conformance lint pass. `scripts/docs-style.py` is clean.
- Not run here: vale, okf and destink (not installed on the build machine); `swift test` does not compile on this machine's toolchain on main either.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,13 @@ upgrade, is in
   existing launches retain `owner` behavior. Applications processing mutually
   untrusted work can keep all Fountain API authority on their service host.
 
+- Bulk apply reconciles `Teammate`, `Schedule` and `Webhook` documents alongside
+  `Environment`, `Vault` and `Agent`, so one manifest declares a whole estate.
+  Teammates resolve their agent, environment and vault by name; schedules name
+  their teammate; webhooks are keyed by URL and hand back their signing secret
+  on the apply that creates them. Result rows now report `unchanged` when the
+  record already matched the document. Apply stays additive and prunes nothing.
+
 ### Fixed
 
 - Codex launches on Sprites with inherited and ambient capabilities cleared,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ upgrade, is in
   their teammate; webhooks are keyed by URL and hand back their signing secret
   on the apply that creates them. Result rows now report `unchanged` when the
   record already matched the document. Apply stays additive and prunes nothing.
+  Rebinding a teammate retires the computer its old environment and vault named,
+  and is refused while a turn is running on it.
 
 ### Fixed
 
@@ -67,6 +69,8 @@ upgrade, is in
 
 - The account event stream replays rapid failures missed before discovery and includes finished conversations on reconnect.
 - Registration and conversation creation declare both shapes of 422 refusal without schema-guard exceptions.
+- Updating an environment, vault, agent or webhook endpoint with values it already
+  holds no longer records an `*.updated` audit event naming no changed fields.
 
 ### Changed
 
--- a/apps/fountain/lib/fountain/agents.ex
+++ b/apps/fountain/lib/fountain/agents.ex
@@ -183,9 +183,14 @@ defmodule Fountain.Agents do
         Repo.update(changeset)
       end
 
+    # A save that moves nothing records nothing, the same rule
+    # `Fountain.Environments.update_environment/3` follows (#1636).
     result =
-      result
-      |> audited("agent.updated", merge_metadata(opts, Audit.changed_fields(changeset)))
+      if changeset.changes == %{} do
+        result
+      else
+        audited(result, "agent.updated", merge_metadata(opts, Audit.changed_fields(changeset)))
+      end
 
     # Only once the new identity is the committed one: a home torn down
     # against an update that then failed would be rebuilt for nothing.
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2326,8 +2326,9 @@ defmodule Fountain.Conversations do
 
   `opts[:reason]` says *why*, and reaches every transcript on the machine and
   the audit row: `"home_reset"` (the owner asked — the default),
-  `"environment_changed"`, `"environment_deleted"` or `"vault_deleted"` when
-  the identity moved out from under the home (#1084).
+  `"environment_changed"`, `"environment_deleted"`, `"vault_deleted"` or
+  `"teammate_rebound"` when the identity moved out from under the home
+  (#1084, #1636).
 
   See `create_agent/2` for the rest of `opts` (`:actor`, `:request_ip`).
   """
@@ -2433,6 +2434,11 @@ defmodule Fountain.Conversations do
 
   defp reset_message("vault_deleted"),
     do: "The vault this machine was built for was deleted. " <> @reset_tail
+
+  defp reset_message("teammate_rebound"),
+    do:
+      "The teammate moved to a different environment or vault, so this machine is no " <>
+        "longer its " <> @reset_tail
 
   defp reset_message(_owner), do: "The sandbox was reset by its owner. " <> @reset_tail
 
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1159,6 +1159,36 @@ defmodule Fountain.Conversations do
   end
 
   @doc """
+  `_unsafe_list_cotenant_ids/2`, with the identity each co-tenant declares:
+  `{id, environment_id, vault_id}`, where the environment is the *effective*
+  one a machine would be built from — the conversation's own override, and
+  the agent's environment when it has none. That is the pair a sandbox row
+  carries and `_unsafe_find_home/4` looks a home up by.
+
+  Co-tenants normally share one identity, because attaching to a machine
+  requires the same agent, environment and vault. They can diverge afterwards:
+  rebinding a teammate (`Fountain.Team.update_teammate/4`) moves one
+  conversation's environment or vault while its co-tenants keep theirs. A
+  lifecycle decision taken for the whole machine has to read this rather than
+  assume, or a replacement built for one identity is handed to a conversation
+  that declared another.
+  """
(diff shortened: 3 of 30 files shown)


lex00:
All three findings are addressed in the commits just pushed, and the branch has been rebased onto current main.

On the blocking co-tenant identity finding: `move_cotenants/3` now takes the replacement sandbox rather than its id, and only repoints co-tenants whose effective environment and vault match the replacement's. Effective means `coalesce(conversation.environment_id, agent.environment_id)`, which is the pair `_unsafe_find_home/4` keys on. A co-tenant that declared something else is stranded rather than moved: its session id is cleared, its `sandbox_id` deliberately keeps naming the retired row so its own wake reads it as terminated and builds from its own pair, and it gets the same reset event and transcript line, because the disk is gone for it too. Regressions wake the two conversations in both orders and assert each ends on a sandbox carrying the identity it declared. Reverting the fix makes all three fail.

On the finding that asked for a decision, an occupied destination home: it refuses. `update_teammate/4` computes the destination identity before writing anything, and returns `:destination_home_occupied` when a live home already exists for it that is not the machine the teammate is on. Bulk apply renders that as a per-row message telling the caller to reset or remove the existing computer first. The reasoning is that attaching to the machine already there needs the readiness, runtime-shape and quota checks `attach_conversation/3` makes and this function does not, and that refusing cannot strand a teammate while attaching wrongly can. It is written down in `docs/api.md`, `docs/cli.md`, the help topic and the function's docstring, so the other half stays available to whoever wants it. There is a test that a restated identical binding is not treated as a move, so the guard cannot over-refuse.

On the exception disclosure finding, which was mine to cause: the rescue now puts a fixed sentence in the result row and nothing else. The logs are unchanged. The reviewer was right that this was more than noise, because the environment and vault passes hold plaintext secrets inside that rescue and Elixir's own `MatchError`, `CaseClauseError`, `BadMapError` and `ArgumentError` messages embed the value they choked on, so the module's promise that secret values are never echoed back held only on the paths that did not raise. The test that pinned the leaked string now asserts the fixed message and that the raised text is absent.

The webhook scope commit is kept as pushed, with tests: a sandbox-scoped token is refused on a manifest containing a `Webhook`, the other resources in that manifest are not written, a manifest without one still applies, and a full-scope key gets its secret once.

jhgaylor:
Split into a stack of eight PRs, per the rule against large PRs. This branch stays open as the reference — its body and review threads are the record of why each decision was made — and closes when #1806 merges.

Each PR is based on the one above it; review and merge top down:

| PR | Scope | Diff |
|---|---|---|
| #1799 | `fix:` a save that changes nothing records nothing, and apply says so — **closes #1680** | +177/−60 |
| #1800 | `fix(api):` a raise in one apply document fails that row, not the request | +120/−21 |
| #1801 | `fix(conversations):` a co-tenant follows a replacement only if it declared the same identity | +186/−35 |
| #1802 | `feat(team):` `update_teammate/4` rebinds a teammate and moves its computer | +569/−3 |
| #1803 | `refactor(cli):` group apply documents by kind (no behaviour change) | +73/−40 |
| #1804 | `feat(api):` bulk apply reconciles Teammate documents | +814/−113 |
| #1805 | `feat(api):` bulk apply reconciles Schedule documents | +297/−48 |
| #1806 | `feat(api):` bulk apply reconciles Webhook documents — **closes #1636** | +483/−59 |

Nothing was dropped. `git diff issue-1636-apply-kinds stack/1636-webhook-kind` over `manifest.ex`, `team.ex` and `apply_json.ex` is empty: the tip reproduces this branch's modules byte for byte. The extra lines are the CHANGELOG entries each PR earns and a handful of tests the split made possible — the exception-isolation case in #1800 reaches the raise through the Agent pass rather than the Teammate pass, because Teammate does not exist yet at that point in the stack.

Two deliberate changes from this branch:

- **The SDK bump moved.** This branch bumped the TypeScript SDK to 1.21.2; `main` is now on 1.23.0. #1806 carries a single bump to **1.24.0** for the whole stack, so the package publishes once at the end instead of racing on `main` eight times. #1799, #1804 and #1805 regenerate `contract.json` and the generated types without touching the version.
- **Three findings landed as their own PRs rather than as fixup commits.** The co-tenant identity fix is #1801, the exception isolation is #1800, and the occupied-destination refusal is part of #1802, each with the tests that fail without it.

Verified on the tip, not just per PR: 6 doctests and 4,761 tests with 0 failures; `mix compile --warnings-as-errors`, `mix format --check-formatted` and `mix credo --strict` (7,054 mods/funs, no issues) clean; `scripts/sdk-contract/build.sh --check` ok; `go test`, `go vet` and `gofmt` clean in `cli/`; the TypeScript SDK suite green. Each PR below was also run against its own targeted suites before being committed, and the three tests in #1801 plus the one in #1800 were confirmed to fail with their fix reverted.

Issue #1637, opened 2026-09-06
Labels on a conversation, with a list filter, so a run can stamp its own outcome


lex00:
## Context

A conversation carries `title`, `status`, `agent_id`, `channel_id` and timestamps, and the list filters on the last three. Full-text search covers titles, prompts and replies. That is the right surface for a person scanning their own threads.

A program running as a teammate produces runs a person wants to slice differently: every tick where drift was found, every apply that was gated, every run against `env=prod`. Those facts are known at the moment the turn ends and are not text worth searching. chant will record them on its own ledger regardless; the ask here is that the console and the list route can show and filter them too, so the Fountain UI is a complete view of an environment's operational history without a second tool.

## Proposal

Add free-form labels to a conversation.

- `labels`: a map of string to string, at most 32 entries, keys and values bounded (64 and 256 bytes), on the conversation row.
- Set on create (`POST /api/conversations`, `POST /api/team/:agent_id/messages` for a fresh in-thread turn) and updated by the agent itself through a small ACP extension notification, so a deterministic run can stamp its own outcome. The sandbox token may set labels on its own conversation only.
- Filter: `GET /api/conversations?label=env:prod&label=drift:true` (repeatable, AND semantics), and the same on `GET /api/team/:agent_id/conversations`.
- The console's conversation list and the team page show labels as chips and accept the filter in the URL, the way `/audit` takes its filters.
- Webhook payloads carry `labels` under `data`, since they are ids and facts, never content.

## Acceptance criteria

- [ ] A conversation created with labels returns them on `GET`, and the list filters on any combination with AND semantics.
- [ ] An agent can update its own conversation's labels during a turn; a sandbox token cannot label another conversation.
- [ ] Label limits are enforced with a 422 that names the offending key.
- [ ] `conversation.*` webhook payloads include `labels`.
- [ ] The console list and the team page render labels and honour the URL filter.
- [ ] `docs/api.md` under "Conversations" documents the field, the filter, and the size limits.

## Dependencies and scope

Independent of the other three asks and the least essential of them: chant keeps its own ledger of these facts either way, and this exists so a person never has to leave Fountain to see them. Full-text search over label values is out of scope.

lex00:
Consumer side: INTENTIUS/chant#2115. chant records run outcomes on its own ledger regardless (INTENTIUS/chant#2118); this ask exists so the fountain console can show and filter them too.

PR #1677, opened 2026-09-06
feat(runtimes): acp runtime launches a named command with no model

### LINKED ISSUE #1634: An `acp` runtime: launch a named ACP command with no model, so a deterministic program can run as an agent
## Context

Every runtime Fountain has is an LLM coding agent: `claude`, `codex`, `gemini`, `opencode`, each a module under `apps/fountain/lib/fountain/runtimes/` implementing the `Fountain.Runtimes` behaviour and speaking ACP (ADR 0014). That is the right shape for the product, and it leaves out a class of work Fountain is otherwise perfect for: a deterministic program that wants a warm machine, a vault, a thread, a schedule, and a human who can read what happened.

The concrete consumer is [chant](https://github.com/INTENTIUS/chant), which is replacing its Temporal-backed ops runtime with Fountain. A chant op is a convergent verb (`chant lifecycle apply --env prod`, a converge tick, a drift diff) that should run on a schedule inside a persistent sandbox that holds the repo and the toolchain, with the run visible as a turn in a teammate's thread. chant will ship an ACP server (`chant acp`) so it looks like any other agent to Fountain. What it cannot do from outside is be launched: the runtime registry is a closed map, `Fountain.Runtimes.Model` expects a `provider/model_id`, and a turn resolves inference credentials it will never use.

## Proposal

Add one runtime, `acp`, that launches a command the agent config names and speaks ACP to it over stdio exactly as the four existing runtimes do.

- `agents.runtime` accepts `"acp"`. A new field, `runtime_command` (string, argv or a shell line, resolved inside the sandbox), is required for this runtime and rejected for the others.
- `model` is optional for this runtime. No inference credential is resolved, and no `model.failed` stage can occur.
- `prepare_sandbox/*`, `write_config/2`, `default_env/*` and `skills_root/0` are no-ops or minimal: the command owns its own config. Skills still mount, since a deterministic agent may read them.
- Permission requests, `session/cancel`, tool-call blocks and stop reasons work unchanged, because they are ACP.
- Metering: with `CREDITS_ENABLED` the turn is priced by sandbox time only. There is no token usage to record; `turn.usage` is null.
- The console and `fountain agents create` expose the field. `fountain acp` (ADR 0015) needs no change.

The one product question is whether `runtime_command` is a free string or a catalogue entry. A free string is what a self-hoster wants and what the trusted-mode runner already implies. On the hosted service the command runs inside the sandbox under the same isolation as any agent, so it is no wider than a `setup_script`. I would start with the free string.

## Acceptance criteria

- [ ] `runtime: "acp"` with `runtime_command` creates an agent; the same value without `runtime_command`, or `runtime_command` on another runtime, is a 422 that names the field.
- [ ] A conversation on such an agent runs the command inside the sandbox, streams `tool_call` / `tool_call_update` / `agent_message_chunk` blocks to the transcript and the SSE feed, and ends the turn on the ACP stop reason.
- [ ] `session/cancel` via `POST /api/conversations/:id/interrupt` reaches the command and the turn ends `interrupted`.
- [ ] `session/request_permission` from the command produces a `permission_request` block and honours the agent's policy, unchanged from the LLM runtimes.
- [ ] No inference credential is looked up and no `model` is required; a turn on an account with no credentials succeeds.
- [ ] The permission policy table in `docs/concepts/permissions.md` and the runtime field in `docs/sdk.md` gain a row for `acp`.
- [ ] A conformance test runs a fixture ACP agent (a few lines of Node or Elixir) through the full turn on the test sandbox provider.

## Dependencies and scope

No dependency on other issues. The consumer-side work is tracked in chant; that epic link will be added here once it exists. Out of scope: a catalogue of blessed commands, any change to the four LLM runtimes, and the durable-request change filed separately.

lex00:
Every runtime Fountain has is an LLM coding agent. A deterministic program that wants a warm machine, a vault, a thread, a schedule and a human who can read what happened could not be launched: the runtime registry is a closed map, the model parser expects a `provider/model_id`, and a turn resolves inference credentials it will never use. This PR adds an `acp` runtime that launches a named command and speaks ACP to it over stdio like the four existing runtimes.

- Add `Fountain.Runtimes.Command`, a `Managoat.Runtimes` implementation for a command the agent names, and a `Fountain.Runtimes` facade that answers `acp` itself and delegates everything else to the library registry. Every runtime lookup, including skills installation, goes through the facade.
- Add `agents.runtime_command`, required for `acp` and refused on any other runtime, and make `agents.model` optional for `acp` only. Both refusals are 422s naming the field. The command is a free string, run as `bash -lc` inside the sandbox.
- Resolve no inference credential and pin no model on an `acp` turn, so `model.failed` cannot come from the pin path and a turn on an account with no credentials succeeds. `turn.usage` is null; credits price the turn by sandbox time only.
- Expose the field in the API, the console form (which disables the model input on `acp`) and the account export. Model suggestions return nothing for `acp`.
- Add a conformance test that runs a checked-in fixture ACP agent, a small Elixir program, as a real OS process through a whole turn, plus scripted-agent tests for tool-call blocks, the stop reason, interrupt and all three permission verdicts.
- Document the runtime in the catalog, `docs/sdk.md`, `docs/concepts/permissions.md` and the integrations pages, including what "the same isolation" means on a self-hosted runner.

Closes #1634

Stacked on #1637; merge that first.

Validation:

- The SDK contract check, the TypeScript verifier and tests, the Python verifier and the conformance lint pass. `scripts/docs-style.py` is clean.
- Not run here: vale, okf and destink (not installed on the build machine).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,24 @@ upgrade, is in
   and `GET /api/team/:agent_id/conversations` take a repeatable `label=key:value`
   filter, combined with AND. `conversation.*` webhook payloads carry `labels`,
   and the console's conversation lists render them as chips.
+- **An `acp` runtime launches a named command, so a deterministic program can
+  run as an agent** (#1634). `agents.runtime` accepts `"acp"`, and a new
+  `runtime_command` field carries the command it runs. The field is required
+  for that runtime and a 422 naming the field on every other one, which
+  resolves its own executable from a pinned table. `model` is optional there:
+  no inference credential is resolved, no model is pinned on the session, and
+  a turn succeeds on an account that holds no API key. Fountain installs no
+  adapter for it, and the command owns its own configuration; skills still
+  mount, and their path arrives as `FOUNTAIN_SKILLS_DIR`. Everything the
+  protocol carries is unchanged, including tool-call blocks on the transcript
+  and the SSE feed, `session/cancel` on interrupt, and the agent's permission
+  policy. With `CREDITS_ENABLED` the turn is priced by sandbox time alone and
+  `turn.usage` is null.
+
+  `runtime_command` is a **free string** rather than an entry in a catalog of
+  blessed commands. It is resolved inside the sandbox, under the same
+  isolation an environment's `setup_script` already runs under, so a catalog
+  would restrict a self-hoster and protect nobody.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,18 @@ upgrade, is in
   `runtime_command` is a **free string** rather than an entry in a catalog of
   blessed commands. It is resolved inside the sandbox, under the same
   isolation an environment's `setup_script` already runs under, so a catalog
-  would restrict a self-hoster and protect nobody.
+  would restrict a self-hoster and protect nobody. On a self-hosted runner
+  with the default backend that isolation is a directory and the daemon's own
+  user, which is what trusted mode already means for a setup script; the
+  runtime docs say so beside the field.
+
+  **Client note.** `Agent.model` is now nullable, in the response and in the
+  create and update bodies, so an agent converted to `acp` can clear the
+  model it no longer uses with `{"model": null}`. In the TypeScript SDK
+  `Agent["model"]` is `string | null`, and `AgentRequest["model"]` and
+  `AgentUpdate["model"]` are `string | null` and optional. A client that
+  assumed a string needs a null check. Nothing else on the wire changed
+  shape.
 
 ### Fixed
 
--- a/apps/fountain/lib/fountain/agents.ex
+++ b/apps/fountain/lib/fountain/agents.ex
@@ -356,6 +356,7 @@ defmodule Fountain.Agents do
     :system,
     :model,
     :runtime,
+    :runtime_command,
     :sandbox_provider,
     :sandbox_mode,
     :skills,
--- a/apps/fountain/lib/fountain/agents/agent.ex
+++ b/apps/fountain/lib/fountain/agents/agent.ex
@@ -4,12 +4,17 @@ defmodule Fountain.Agents.Agent do
 
   alias Fountain.Accounts.User
   alias Fountain.Environments.Environment
+  alias Fountain.RuntimeDispatch
   alias Managoat.Runtimes.Model
 
   @primary_key {:id, :binary_id, autogenerate: true}
   @foreign_key_type :binary_id
 
-  @runtimes ~w(claude codex gemini opencode)
+  # `acp` is the odd one (#1634): not a coding-agent CLI but a command the
+  # agent names, launched inside the sandbox and spoken to over the same
+  # protocol. It takes a `runtime_command` and needs no model, and the other
+  # four are the reverse of both. See `Fountain.RuntimeDispatch`.
+  @runtimes ~w(claude codex gemini opencode acp)
 
   @typedoc "A persisted agent."
   @type t :: %__MODULE__{}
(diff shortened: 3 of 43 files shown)


jhgaylor:
Split into a stack of eight, per the rule against big PRs. This branch's 1,760 lines are now #1832 → #1839, each based on the one before it, off `main` rather than on the labels branch:

| | PR | Lines | What it is |
|---|---|---|---|
| 1 | #1832 | +279 | `Fountain.CommandRuntime` and the host dispatch for `acp` |
| 2 | #1833 | +399 | `agents.runtime_command`, and `acp` becomes a runtime you can name |
| 3 | #1834 | +73 | skills mount on a runtime the library does not know |
| 4 | #1835 | +332 | a turn on the acp runtime, end to end |
| 5 | #1836 | +72 | refuse a turn whose agent no longer carries a command |
| 6 | #1837 | +151 | the acp runtime in the agent form |
| 7 | #1838 | +399 | a real ACP program through a whole turn |
| 8 | #1839 | +139 | docs, CHANGELOG and SDK 1.25.0 |

Four things had to change rather than be copied, because this branch is stale against `main`:

- **The model pin.** `main` now has `TurnMachine.acp_model/2` and `CodexChatGPT.peer_auth/2`; this branch still called `RuntimeDispatch.acp_model/2` from the conversation server. The stack keeps main's call sites and makes `TurnMachine.acp_model/2` delegate through the dispatch, so the `acp` → nil rule lives in one place.
- **`TurnMachine.open/4`'s spec** dropped main's `{:error, term()}` clause (execution allowances). The stack keeps all four returns.
- **The migration timestamp** `20260906120000` sat behind five migrations already on main. Renamed `20260910120000`.
- **The console default model** `anthropic/claude-sonnet-4-6` is no longer a catalog entry and failed an existing test on main. Now `claude-sonnet-5`.

And two defects that only showed up once the pieces were run separately:

- **`conversation_server.ex` went over its size pin.** The file is line-pinned and only shrinks (#1369); the added comment put it at 2775 against a pin of 2774. Folded into the existing comment instead of raising the pin.
- **The conformance test is flaky as written.** `"the turn completes on the agent's stop reason"` waits on the *turn* row reaching `completed`, then asserts the *conversation* row is `idle` — which the server writes afterwards. It reproduced on the first seed once the timing shifted. #1838 waits on the row the assertion reads; 25 seeds clean.

Full suite on the stack tip: 4805 core tests, 144 fountain_buzz, 33 fountain_support, 0 failures. Leaving this open for you to close.

lex00:
Superseded by #1832 to #1839; the diff over apps/ matches apart from the deviations listed above.

jhgaylor:
All eight landed on `main`, plus a replacement for one of them:

| | PR | |
|---|---|---|
| 2 | #1833 | `agents.runtime_command`, `acp` becomes nameable, and the Swift SDK widening |
| 6 | #1863 | the console form (replaces #1837, see below) |
| 8 | #1839 | docs, ADR 0049 and SDK 1.26.0 |

`git diff` between this branch's rebased tree and `main` over the five core modules is empty, so what shipped is the code this PR proposed.

Three things the split turned up that this branch would have carried in:

- **The Swift SDK broke on a null model.** `FountainKit.Agent` declared `model: String` on a `Decodable` struct, so an `acp` agent's explicit null threw `valueNotFound` — and because a page decodes whole, one such agent broke `agents.list()` for every Swift caller. Widened in #1833 with tests that fail on the original declaration. No gate caught it: the contract checker does not compare a hand-written model's optionality against a nullable wire field.
- **`conversation_server.ex` went over its size pin.** The file only shrinks (#1369) and the added comment took it to 2775 against a pin of 2774.
- **The conformance test was flaky as written**, waiting on the turn row and asserting on the conversation row.

Two unrelated flakes were filed along the way: #1861 (Swift conformance timeout scenario) and #1862 (TeamTest co-tenant test dying on a shutting-down Horde supervisor).

Closing this in favour of the stack.

PR #1678, opened 2026-09-06
feat(conversations): permission requests that outlive a turn

### LINKED ISSUE #1635: Permission requests that outlive a turn: end the turn waiting, suspend the sandbox, resume on the answer
## Context

An `ask` verdict holds the agent's tool call until a human answers, and `Fountain.Permissions.ask_timeout_ms/0` denies it after `permission_ask_timeout_seconds` (300 by default). The comment above `@default_ask_timeout_seconds` explains why the ceiling has to sit below the idle bound: the request is held inside a running turn, and a turn that outlives the idle bound would be reclaimed by the max-lifetime ceiling rather than parked. Everything about the mechanism assumes an LLM blocked mid-thought.

A deterministic op has waits of a different shape: approve this production apply, confirm the DNS delegation happened, sign off on the deletes in this plan. Those take hours or days, and nothing in the sandbox needs to stay running while they do. The card, first-answer-wins, the audit row, the `request` stage webhooks and the answer endpoint are exactly the right UX. Only the "must resolve before the turn ends" rule is wrong for them.

## Proposal

Let a request outlive its turn.

- An ACP agent may end a turn with a pending request attached: it sends `session/request_permission`, then returns from `session/prompt` with stop reason `waiting` (or a Fountain-specific extension key on the existing stop reasons, whichever ADR 0014's dialect layer prefers). The turn ends `done` with a new `waiting: true` flag on the turn row; the request row stays `pending`.
- The conversation goes `idle` and the sandbox suspends as usual. The request card stays up in the console and the team app; `GET /api/conversations/:id` lists pending requests.
- Answering (`POST /api/conversations/:id/requests/:request_id`) resolves the request and opens a new turn whose prompt is the answer, delivered to the agent as a `session/prompt` carrying the request id and the chosen option. The agent decides what to do with it. The `request.done` stage fires as today.
- A per-request `timeout` (seconds, sent with the request) or, absent that, a new policy-level `ask_timeout` replaces the global ceiling for detached requests, since the idle-bound reasoning no longer applies. Expiry denies the request and opens the same resume turn with the denial.
- A turn that ends without a pending request is unchanged. The existing in-turn `ask` path is unchanged.

## Acceptance criteria

- [ ] An agent that returns `waiting` with an outstanding request ends its turn `done`, the request stays `pending`, and the conversation is `idle`.
- [ ] The sandbox suspends on the idle bound with the request still pending, and answering wakes it into a new turn that carries the request id and option id.
- [ ] `GET /api/conversations/:id` and the SSE `stage` events expose the pending request; `conversation.request.*` webhooks fire on start and on resolution.
- [ ] A request-level timeout longer than the idle bound is accepted and honoured; expiry opens the resume turn with a denial.
- [ ] `POST .../requests/:id` from the sandbox's own token is still refused (`sprite_may_not_answer`).
- [ ] The audit trail records the resolution with the same actor attribution as an in-turn answer.
- [ ] `docs/concepts/permissions.md` gains a section on requests that outlive a turn, and the "Nobody is also an answer" rule names both timeouts.

## Dependencies and scope

Independent of the `acp` runtime issue, though the two are meant to be used together. Out of scope: changing the 5-minute default for in-turn requests, and any UI beyond keeping the existing card visible while the conversation is idle.

lex00:
An `ask` verdict holds the agent's tool call until a human answers and denies it after five minutes, because the request is held inside a running turn that must not outlive the idle bound. A deterministic op has waits of a different shape: approve this production apply, confirm the DNS delegation happened. Those take hours or days, and nothing in the sandbox needs to stay running while they do. This PR lets a request outlive its turn.

- An agent that sends `session/request_permission` and then answers `session/prompt` with the stop reason `waiting` ends its turn `completed` with `waiting: true` on the turn row and the request still pending. The conversation goes idle and the sandbox suspends as usual. The connection is dropped at detach, since the ACP client cannot release a held request without answering it; the resume turn always handshakes a fresh peer.
- Answering through `POST /api/conversations/:id/requests/:request_id` resolves the request with a guarded single-row update (first answer wins) and opens a new turn whose prompt is one JSON line under the key `fountain/permission_answer`, carrying the request id, the tool, the outcome and the option. The wake's gates run before the row is touched, and a delivery that still fails afterwards is reported distinctly.
- A per-request timeout in the request's `_meta` or a policy-level `ask_timeout` replaces the global ceiling for detached requests, the smaller of the two when both are set; a launch may not raise the agent's value or the global ceiling. `Fountain.Workers.DetachedRequestSweeper` expires overdue requests every minute, denies with the agent's own reject option, and opens the same resume turn. It leaves a request alone when the conversation cannot take a turn yet.
- `GET /api/conversations/:id` lists pending requests; the `turn` stage carries `waiting`; `conversation.request.*` webhooks fire on detach and on resolution; a sandbox token is still refused; the audit trail records the resolution with the caller's attribution.
- Move the pending family's state round trip into `Pending` to keep the conversation server under its size pin.
- Document the detached path in `docs/concepts/permissions.md` and amend ADR 0014.

Three follow-ups for the `managoat_acp` library are noted in the commits: a `Peer.release_permission/2` that clears a hold without writing a response, a `_meta` hook on `Peer.prompt`, and reporting request params with `{:permission_ask, ...}`.

Closes #1635

Stacked on #1634; merge that first.

Validation:

- The SDK contract check, the TypeScript verifier, typecheck and tests, the Python verifier and the conformance lint pass. `scripts/docs-style.py` is clean.
- Not run here: vale, okf and destink (not installed on the build machine).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,8 +79,10 @@ upgrade, is in
   goes idle and the sandbox suspends as usual. `GET /api/conversations/{id}`
   lists such requests as `pending_requests`, and answering one opens a new turn
   carrying the request id and the chosen option, which wakes the sandbox. The
-  wait is bounded by `_meta.fountain.timeout` on the request, else an
-  `ask_timeout` in the permission policy, else the existing 5 minute ceiling.
+  wait is bounded by `_meta.fountain.timeout` on the request and by an
+  `ask_timeout` in the permission policy, the shorter of the two, else the
+  existing 5 minute ceiling. An answer is refused, and the request kept, when
+  the conversation cannot take the turn that carries it.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,13 @@ upgrade, is in
   `AgentUpdate["model"]` are `string | null` and optional. A client that
   assumed a string needs a null check. Nothing else on the wire changed
   shape.
+- Permission requests can outlive the turn that raised them. An agent that ends
+  a turn with stop reason `waiting` keeps its request open, the conversation
+  goes idle and the sandbox suspends as usual. `GET /api/conversations/{id}`
+  lists such requests as `pending_requests`, and answering one opens a new turn
+  carrying the request id and the chosen option, which wakes the sandbox. The
+  wait is bounded by `_meta.fountain.timeout` on the request, else an
+  `ask_timeout` in the permission policy, else the existing 5 minute ceiling.
 
 ### Fixed
 
--- a/apps/fountain/lib/fountain/agents/agent.ex
+++ b/apps/fountain/lib/fountain/agents/agent.ex
@@ -4,6 +4,7 @@ defmodule Fountain.Agents.Agent do
 
   alias Fountain.Accounts.User
   alias Fountain.Environments.Environment
+  alias Fountain.PermissionPolicy
   alias Fountain.RuntimeDispatch
   alias Managoat.Runtimes.Model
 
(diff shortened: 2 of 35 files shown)


lex00:
Re-cut as a stack of nine on `main`: #1852 to #1860. Two things changed against main and are named in #1855 and #1860.

PR #1681, opened 2026-09-06
fix(api): an id a caller gets wrong is a 422, not a dropped connection

### LINKED ISSUE #1679: A non-uuid in allowed_vault_ids reaches Ecto and returns a 500 with a dropped connection, instead of a 422
Measured today against a v0.16.0 instance stood up with [fountain-ops](https://github.com/INTENTIUS/fountain-ops).

`POST /api/apply` with an Agent document whose `allowed_vault_ids` holds a vault **name** rather than a uuid fails inside the database layer:

```
** (Ecto.ChangeError) value ["prod-creds"] for AgentRequest.allowed_vault_ids
   does not match type {:array, :binary_id}
```

The caller gets a bare 500, and the HTTP connection drops with it. Reproduced three times in a row; a uuid in the same field applies cleanly.

The mistake was mine to make on the client side, and the client is being fixed (INTENTIUS/chant#2166). What makes this worth an issue here is the shape of the failure rather than the cause: this is caller-supplied input reaching `Ecto.ChangeError` with no changeset validation in front of it, so the answer is a 500 and a dropped socket rather than a 422 naming the field. Every other bad value in that request produces a clean validation error, including the sibling `environment` reference, which fountain resolves by name.

Two things would each be an improvement, and the first is enough:

1. Validate `allowed_vault_ids` as a list of uuids in the changeset, so a bad entry is a 422 that names the field.
2. Optionally, resolve vault **names** the way `environment` is already resolved on this endpoint. A manifest that names an environment by name and a vault by uuid is an inconsistency an author notices; `fountain apply -f` is otherwise a name-addressed format end to end.

Filed alongside a second, smaller apply finding.

lex00:
`POST /api/apply` with an Agent document whose `allowed_vault_ids` holds a vault name answered with a bare 500 and a dropped connection, from `Ecto.ChangeError` raised inside `Repo.insert/1`. This PR puts a guard in front of that input, in the two places the shape can arise.

The cause is that `:binary_id` casts any string. `Ecto.Type.cast/2` on the primitive accepts a name, `changeset.valid?` says true, and only the adapter's dump rejects it, by raising after validation is over. Every id in a request body is declared `format: :uuid` and refused cleanly by the OpenAPI cast, which is why the rest of the same request validated normally. Two kinds of caller input never reach that cast: a document `spec` through bulk apply, which is free-form by design, and an id that arrives as a path segment or a header.

- Add `Fountain.Changeset.validate_ids/2` and run it on the id fields of the two schemas a caller can fill, so a bad value is a changeset error naming the field and the offending value, capped so the message cannot echo a payload back.
- Stop an ownership check querying an id the changeset already refused. `validate_environment_owner/1` read the change and looked it up whatever the changeset said, so an Agent document naming an environment raised `Ecto.Query.CastError` from inside `Repo` before the 422 could be returned.
- Guard the scoped fetches themselves. `Conversations.get_conversation/2` and `Agents.get_agent/2` now read a malformed id as nil, the way `get_sandbox/2` already did, and the two hand-rolled `rescue Ecto.Query.CastError` clauses in the MCP controllers are deleted because the function they were compensating for now owns the rule.
- Use `Ecto.UUID.dump/1` rather than `cast/1` everywhere this guards a query. `cast/1` accepts any 16-byte binary, so a sixteen-character name such as `prod-credentials` passed the first version of this validator and still raised at dump.

Worth a reviewer's attention: the third point turns a 500 into a 404 on endpoints outside this issue, including `/api/agui/:agent_id`, which takes a raw path segment and declares `not_found` already. That is the reviewer's recommended shape rather than my own preference, and it is the reason the two hand-rolled rescues could go.

Not done: resolving vault names the way `environment` is resolved on this endpoint, the issue's optional second half. Bulk apply is a name-addressed format and that inconsistency is real, but it changes what the field means rather than how it fails.

Closes #1679

Validation:

- Every regression test here was run against the code without its fix and fails there with the exception from the issue: `Ecto.ChangeError` on the apply path, `Ecto.Query.CastError` on the header, the ownership lookup and the sixteen-byte case.
- Full suite: 4382 tests, 0 failures, 1 skipped (core and ee), 144 (fountain_buzz), 33 (fountain_support).
- `mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict`, dialyzer and `deps.unlock --unused` are clean. No schema changed, so the SDK contract does not move and the release gate reports nothing to publish.
- `python3 scripts/docs-style.py` reports one problem, `docs/catalog/runtimes/codex.md:62`, which came from main and is untouched here.
- Not run locally: vale, okf and destink, which are not installed on this machine.


lex00:
All four review findings are addressed in the commits just pushed, including the one that asked for a human decision. Taking them in order.

The 16-byte hole is real and the switch to `Ecto.UUID.dump/1` is kept. `cast/1` reads any 16-byte binary as a uuid, so `prod-credentials`, sixteen characters, passed the first version of this validator and still raised at dump. There are now tests for that value and for a raw 16-byte binary, and both fail against the previous commit.

The ownership lookup was raising for the same reason. `validate_environment_owner/1` queried the change whatever the changeset already said, so an Agent document naming an environment raised out of `Repo` before the 422 could be returned. It now returns an already-invalid changeset unchanged, with coverage on the apply path and on update.

On the finding that asked for a decision: the guard has moved into the scoped fetches, which is the wider of the two options offered. `Conversations.get_conversation/2` and `Agents.get_agent/2` read a malformed id as nil the way `get_sandbox/2` already did, `resolve_parent_id/2` is a plain lookup again, and the hand-rolled `rescue Ecto.Query.CastError` clauses in `TeamMcpController` and `CallerMcpController` are deleted. That was the signal the reviewer named for the fix landing in the right place, and the argument for it is the one in the finding: four controllers call those two functions, two had compensated by hand and two had not.

The consequence to be clear about, because it reaches beyond this issue: `/api/agui/:agent_id` and the two MCP controllers now answer 404 where they answered 500. `not_found` is a declared response on each of them, and an id that is not an id is not an agent or a conversation anyone owns. Reverting to the narrow call-site guard is a one-line change if that is the wrong call.

`get_sandbox/2` moved to `dump/1` as well, since it carried the same 16-byte hole the finding describes.

jhgaylor:
Jake asked me to collect Review Loop feedback during normal PR work. After your next run, please flag bot fixes that leave CI red or repeat already-addressed findings. No extra run needed.

Issue #1674, opened 2026-09-06
Sprites: https-scheme broker proxy costs ~300s per turn in codex, plus CA bundle races


jhgaylor:
Follow-up to #1672 (fixed by #1673). With ambient capabilities cleared, codex's
sandbox starts correctly — but two further problems surface on Sprites, both
caused by the egress broker being an **https-scheme** proxy.

## 1. Every turn pays a ~300s websocket timeout

Codex's `responses_websocket` transport cannot use an https-scheme proxy, but it
waits the full websocket connect timeout before falling back to HTTP. From
`~/.codex/logs_2.sqlite` on a Sprites sandbox:

```
13:19:06 INFO  codex_api::endpoint::responses_websocket  connecting to websocket: wss://api.openai.com/v1/responses
   … 303 seconds of nothing …
13:24:09 (first agent output)
13:24:14 ERROR codex_api::endpoint::responses_websocket  failed to connect to websocket:
         URL error: Proxy URL scheme not supported, url: wss://api.openai.com/v1/responses
13:24:21 WARN  codex_core::client  falling back to HTTP
```

Measured on the conversation stream, `stage/model/done` → first agent output:

| turn | gap |
| --- | --- |
| 1 | 303s |
| 2 | 292s |
| 3 | 306s |

The actual work took ~1s; the commands run instantly once the model responds.
The proxy is `https://<user>:<vault>@broker.inevitable.fyi`, set in
`/home/sprite/.env` as `HTTPS_PROXY`/`https_proxy`.

`~/.codex/config.toml` cannot fix this from the outside: the only lever is
`supports_websockets` / `websocket_connect_timeout_ms` on the provider, and
overriding the built-in provider is rejected —

```
failed to load configuration: model_providers contains reserved built-in
provider IDs: `openai`. Built-in providers cannot be overridden.
```

Since Fountain already passes `CODEX_CONFIG={"features":{"respect_system_proxy":true}}`
via the spawn env, that seems like the right place to also disable the
websocket transport (or shorten its connect timeout) whenever a proxy is
configured.

## 2. The CA bundle is read while it is being rewritten

Codex logs which CA source it used. On a sandbox shared by many conversations,
the same path yields wildly different certificate counts:

| certificates loaded | occurrences |
| --- | --- |
| 122 (correct) | 1187 |
| 8 / 5 / 3 / 2 / 1 | 221 |

```
loaded certificates from custom CA bundle source_env="SSL_CERT_FILE" \
[...]

PR #1682, opened 2026-09-07
fix(codex): a brokered turn stops paying a 300s websocket timeout

jhgaylor:
The first problem in #1674: every brokered codex turn paid a ~300 second
websocket timeout before doing about a second of work.

Codex dials `wss://api.openai.com/v1/responses` first. That dialer cannot use
an https-scheme proxy, which the broker is, but it waits out the full connect
timeout before saying so:

```
13:19:06 INFO  connecting to websocket: wss://api.openai.com/v1/responses
   … 303 seconds …
13:24:14 ERROR failed to connect: URL error: Proxy URL scheme not supported
13:24:21 WARN  falling back to HTTP
```

Measured on the conversation stream, `stage/model/done` to first agent
output: 303s, 292s, 306s on three consecutive turns.

`supports_websockets` is a `[model_providers.<id>]` field, and the built-in
`openai` id cannot be overridden — `model_providers contains reserved
built-in provider IDs` is a hard configuration error, from `config.toml` and
from `CODEX_CONFIG` alike (openai/codex#13103). An id of our own is not
reserved, so `CodexTransport` declares the same endpoint with the transport
off and selects it:

```json
"model_provider": "fountain_openai_http",
"model_providers": {
  "fountain_openai_http": {
    "name": "OpenAI",
    "base_url": "https://api.openai.com/v1",
    "wire_api": "responses",
    "env_key": "OPENAI_API_KEY",
    "supports_websockets": false
  }
}
```

`env_key` rather than the built-in's `~/.codex/auth.json`, which a custom
provider does not read. `OPENAI_API_KEY` is in the sandbox env either way
(`Managoat.Runtimes.Codex.default_env/2`), and on a brokered conversation it
is the placeholder `Broker.split_inference/2` swapped in, which the broker
substitutes for the real key on the way out — so nothing about which key pays
changes, for a tenant key or a platform one.

This is scoped exactly as the existing `respect_system_proxy` policy is:
brokered conversations, codex only, process-local (persistent sandboxes are
shared). An agent already pointed at a gateway keeps the provider it names —
that provider decides its own transport, and its base URL is not ours to
replace — and an operator who declares `fountain_openai_http` themselves
keeps their declaration.

## Testing

- `mix test` (core suite), `mix credo --strict`, `mix format --check-formatted`.
- Removing the provider switch fails the new test, so it guards.

Not verified against a live Sprite: that needs a brokered codex conversation,
and the measurement to repeat afterwards is the `stage/model/done` to first
output gap on the conversation stream.

Refs #1674. The broker CA race and the `env_vars` ordering are a separate PR.

https://claude.ai/code/session_01XT411VudYs1hw69mDDhvrm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,8 @@ upgrade, is in
   out — around 300 seconds on every turn, before falling back to HTTP and
   answering in about a second. The built-in `openai` provider is reserved and
   cannot be overridden, so Fountain declares the same endpoint under an id of
-  its own, carrying across `OPENAI_BASE_URL` where an environment sets one. A
+  its own, carrying across `OPENAI_BASE_URL`, the OpenAI-Organization and
+  OpenAI-Project header mappings and standalone web search. A
   conversation whose spawn has no `OPENAI_API_KEY` keeps the built-in
   provider, which can still authenticate from `~/.codex/auth.json`. An agent
   that names its own provider in `CODEX_CONFIG` keeps it; a `model_provider`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,8 +52,12 @@ upgrade, is in
   out — around 300 seconds on every turn, before falling back to HTTP and
   answering in about a second. The built-in `openai` provider is reserved and
   cannot be overridden, so Fountain declares the same endpoint under an id of
-  its own. An agent already pointed at a gateway keeps the provider it names
-  (#1674).
+  its own, carrying across `OPENAI_BASE_URL` where an environment sets one. A
+  conversation whose spawn has no `OPENAI_API_KEY` keeps the built-in
+  provider, which can still authenticate from `~/.codex/auth.json`. An agent
+  that names its own provider in `CODEX_CONFIG` keeps it; a `model_provider`
+  written into `~/.codex/config.toml` by a setup script is not read, and is
+  overridden (#1674).
 
 - Codex launches on Sprites with inherited and ambient capabilities cleared,
   allowing its bubblewrap sandbox to start without falling back to approval
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,15 @@ upgrade, is in
   variables still win over everything: they are what makes egress brokered.
   Both halves of the rule are now in the manual, under Secrets (#1674).
 
+- A brokered Codex conversation reaches OpenAI over a provider with the
+  WebSocket transport turned off. Codex's `responses_websocket` dialer cannot
+  use an https-scheme proxy and spent the full connect timeout finding that
+  out — around 300 seconds on every turn, before falling back to HTTP and
+  answering in about a second. The built-in `openai` provider is reserved and
+  cannot be overridden, so Fountain declares the same endpoint under an id of
+  its own. An agent already pointed at a gateway keeps the provider it names
+  (#1674).
+
 - Codex launches on Sprites with inherited and ambient capabilities cleared,
   allowing its bubblewrap sandbox to start without falling back to approval
   escalation on every command (#1672). Existing adapter processes need a restart.
--- a/apps/fountain/lib/fountain/conversations/codex_transport.ex
+++ b/apps/fountain/lib/fountain/conversations/codex_transport.ex
@@ -5,6 +5,33 @@ defmodule Fountain.Conversations.CodexTransport do
   # Codex's default WebSocket dialer rejects HTTPS proxy URLs. Its explicit
   # proxy route supports TLS to the broker, but is opt-in in Codex 0.153.3.
   # Keep this process-local: persistent sandboxes are shared by conversations.
+
+  # Rejecting the proxy URL is not the expensive part — waiting to find out
+  # is. `responses_websocket` dials `wss://api.openai.com/v1/responses`, sits
+  # on the connect timeout, reports `Proxy URL scheme not supported` and only
+  # then falls back to HTTP. Measured on Sprites (#1674), that was 303, 292
+  # and 306 seconds on three consecutive turns whose actual work took about a
+  # second each.
+  #
+  # `supports_websockets` is a provider field, and the built-in `openai`
+  # provider cannot be overridden — `model_providers contains reserved
+  # built-in provider IDs` is a hard configuration error, from the file and
+  # from here alike (openai/codex#13103). A provider id of our own is not
+  # reserved, so declare the same endpoint with the websocket transport off
+  # and select it. `env_key` rather than the built-in's `auth.json`, because a
+  # custom provider does not read that store; `OPENAI_API_KEY` is in the
+  # sandbox env either way, and on a brokered conversation it is the
+  # placeholder the broker substitutes for the real key on the way out
+  # (`Fountain.Broker.split_inference/2`), so nothing about who pays changes.
+  @provider_id "fountain_openai_http"
+  @provider %{
+    "name" => "OpenAI",
+    "base_url" => "https://api.openai.com/v1",
+    "wire_api" => "responses",
+    "env_key" => "OPENAI_API_KEY",
+    "supports_websockets" => false
+  }
+
   def spawn_opts(%{broker: broker}, "codex", opts) when not is_nil(broker) do
     env = Keyword.get(opts, :env, [])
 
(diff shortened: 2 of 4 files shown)


jhgaylor:
Both medium findings were real and are fixed in 05058ea0.

**`ad932d45` (endpoint).** The concrete harm in that finding was not the `config.toml` case but a simpler one the review reasoning leads straight to: the built-in `openai` provider reads `OPENAI_BASE_URL`, so an environment pointing codex at a gateway through `env_vars` was already working, and a replacement provider with a hard-coded `base_url` would have moved it back to OpenAI silently. The provider now reads `OPENAI_BASE_URL` out of the spawn env. For the `~/.codex/config.toml` half I took option (a): the module comment and the CHANGELOG now say the overlay is the only surface read, that a `model_provider` a setup script wrote into the file is overridden, and that `env_vars` is the supported way to point codex elsewhere. Option (b) was not viable — a default brokered conversation passes no CODEX_CONFIG at all, so narrowing the guard to an explicitly present `model_provider` forfeits the fix entirely.

**`ee78dea8` (credential).** Confirmed the way the finding asks. `prepare_runtime_sprite/5` has exactly one caller (conversation_server.ex:898, the provision path), so a wake or reattach never re-runs `codex login --with-api-key` and rides on the `auth.json` already on the sandbox — which is precisely the store a custom provider cannot read. The substitution is now conditional on `OPENAI_API_KEY` being present and non-empty in the spawn env; without it the conversation keeps the built-in provider and pays the stall. A slow turn beats a lost one, and the `:no_credential` contract ("provisions anyway, the provider's own auth failure lands on the transcript") is preserved rather than replaced by a Fountain-invented misconfiguration.

**`ee2caa47` / `305dbe92` (the assertion that could not fail).** Removed. That test now asserts the whole `model_providers` map, and the "existing settings survive" test asserts its exact key set, so "exactly one entry added, every neighbour untouched" is checked where there is a neighbour to protect.

**`929c5ff2` / `3bb97a6d` (naming).** `disable_websockets/2` is now `select_http_provider/2`.

**`8d456717` (whole-map assertion relaxed).** Restored to a whole-map check.

**`b6464bd0` (error term).** No change, as the finding suggests.

**`8dcc317d` (codex CLI is unpinned)** is left for the maintainer: pinning the codex binary in the sandbox image is a deployment decision, not a code edit, and nothing in this PR makes the exposure worse than it already is for gemini-cli. Flagged for a decision rather than resolved here.

Both new behaviours were verified by reverting them: hard-coding the base URL fails the gateway test, and dropping the credential check fails the no-key test. Full core suite green (4,375 tests), `credo --strict` clean.

jhgaylor:
Accepted the fix and covered it in 012d36ee. Full core suite green (4,376 tests), `credo --strict` clean, formatted. The CI failure the generation-2 run stopped on was the exact-map test the one-file policy kept it from updating; that is what this commit fixes.

I verified the audit the change rests on rather than taking it on trust. At `rust-v0.153.3`, `codex-rs/model-provider-info/src/lib.rs`, `create_openai_provider` sets `env_http_headers` mapping `OpenAI-Organization` to `OPENAI_ORGANIZATION` and `OpenAI-Project` to `OPENAI_PROJECT`, and `supports_standalone_web_search: true`. Dropping those cost a tenant on an organization or project its headers and every conversation its web search — a real regression in my original change. Its single `http_headers` entry is `("version", env!("CARGO_PKG_VERSION"))`, and omitting it is right: Fountain does not know the sandbox's codex build and would be reporting the review image's.

The precedence change is the sharper half and I had it wrong. `List.keyfind/3` returns the first match, but `SpriteEnv.build/4` appends the decrypted secrets after the runtime's own defaults, so a vault entry for `OPENAI_API_KEY` or `OPENAI_BASE_URL` appears later and is what the process actually gets. Reading the first meant substituting on a key the conversation was not using, and missing one a tenant had deliberately emptied. There is now a test for a repeated variable in both directions.

Both behaviours verified by reverting: reversing the precedence fails one test, dropping the header mappings fails another.

`8dcc317d` (the unpinned codex CLI) is still with the maintainer.

jhgaylor:
Accepted and covered in c275ec57. Full core suite green (4,377 tests), `credo --strict` clean, formatted.

The finding is right, and it is the same class as the `Map.new/1` precedence fix one commit earlier: `List.keystore/4` rewrites the *first* matching entry, but a process holding `CODEX_CONFIG` twice reads the last, so an untouched later entry would have shadowed the rewrite and neither half of this module's policy — `respect_system_proxy` or the provider substitution — would have reached codex at all. The two changes belong together, and the new test pins both directions: read from the effective entry, write back exactly one entry, last.

The CI failure that stopped generation 3 was `mix format` on the proposal itself, not a behavioural problem; that is fixed here.

Two process notes, offered as observations rather than complaints. The one-file policy has now produced three commits in a row that could not pass CI as pushed — two needing a test update outside the allowed file, one needing the formatter — so each round costs a full CI cycle before the review can even start. And run 3's dispatch never fired at all (a 503 from the service), which is why it had to be restarted by hand.

Still outstanding from generation 1: `8dcc317d`, the unpinned codex CLI. I am taking option (c) — accept the risk and record it where a reader will find it, rather than pinning the binary in the sandbox image, which is a deployment change this PR should not smuggle in. Recording it in `docs/catalog/runtimes/codex.md` in the next push.

jhgaylor:
/review-loop resolve 2d36bf6c-4bf5-4198-bca8-a3e3e5e3f9c5 8dcc317def49 Accepted and recorded rather than pinned. Option (c): docs/catalog/runtimes/codex.md now states that the sandbox images do not pin the Codex CLI, that supports_websockets is 0.153.3 behaviour, that a later Codex ignoring it restores the ~300s wait with nothing in Fountain reporting it, and names the observable symptom (the model-stage to first-output gap) plus openai/codex#13103. Pinning the binary in the sandbox image is the stronger answer and stays available, but it is a deployment change this PR should not carry.

jhgaylor:
Merging with an admin override, on the maintainer's explicit authorization. Recording what was overridden and how it was checked, so the call is reviewable.

I did not take the override on the previous generation: that round found something real about the current code (the two variables this module reads were resolved last-entry-wins while the duplicates were left in the env for codex to resolve its own way), and it was fixed in `collapse/2` rather than argued with. This generation re-reports it.

**All five mediums on `cf082c9d` cite code that is not on `cf082c9d`.** Each string checked individually against that revision:

| Cited as evidence | On `cf082c9d`? |
| --- | --- |
| `if Map.get(config, "model_provider", "openai") in ["openai", @provider_id]` | no |
| `Map.put_new(providers, …)` | no |
| `select_http_provider(config, providers, env)` | no |
| "`OPENAI_API_KEY` is in the sandbox env either way" | no |

`9e262ff8` is the sharpest of them and states two things about this revision, both false of it:

- *"attributing the rule to `SpriteEnv`'s moduledoc"* — codex_transport.ex:58-61 says the opposite: "Nothing states which duplicate the sandbox process then reads — `SpriteEnv`'s moduledoc is about `merge_secrets/3`, which runs before the list is built … So this does not rely on the answer".
- *"leaving the duplicates in the env for codex to resolve its own way … The safe pattern is already in this function; it is just not applied to the two variables"* — it is applied: `collapse(env, @resolved_names)` at codex_transport.ex:85, defined at :167, with `@resolved_names` at :53. Two tests cover it, including that an env with no duplicates is passed through untouched.

The other four ask for remediations already in the branch:

- `86e10765` / `8018b87a` ask for an assertion that `Managoat.Runtimes.Codex.default_env/2` exports `OPENAI_API_KEY`. It exists, pinned the way `conversation_server_broker_test.exs` pins the same fact for Claude, and asserts membership rather than equality so an unrelated upstream addition does not fail it.
- `ad932d45` asks to state the `~/.codex/config.toml` scope in the CHANGELOG and module comment. Both say it, and so does `docs/catalog/runtimes/codex.md`, which is the page a codex agent author actually reads.
- `8dcc317d` asks to record the unpinned codex CLI. Recorded on that same page, with the observable symptom and openai/codex#13103.

I have not used `/review-loop resolve`, because the accurate record is not that a maintainer waived a defect — it is that the defect was fixed and the finding is reading an older revision.

**Evidence for the merge:** CI green on `cf082c9d`; full core suite 4,384 tests / 0 failures locally; `credo --strict` and `mix format --check-formatted` clean; the three prose gates clean on the changed page. Every behaviour added across this PR was verified by reverting it and confirming a test fails.

One unrelated CI failure on the way here was `PendingTest` asserting a `Process.send_after(…, 0)` delivery inside `assert_receive`'s default 100 ms window — a scheduling race under partition load, the same class as the closed #1592. Filed as #1721 with the run URL and mechanism; it passed on rerun and does not touch anything in this PR.

PR #1683, opened 2026-09-07
fix(broker): a shared sandbox rebuilds its trust store once, not per conversation

jhgaylor:
Two of the three problems in #1674, both on the broker side. They travel
together because the ordering fix only matters once the trust store stops
being rewritten.

## The trust store was rebuilt by every conversation, concurrently

`update-ca-certificates` builds the bundle at a fixed temporary path
(`/etc/ssl/certs/ca-certificates.crt.new`) and renames it into place. The
rename is atomic; sharing the path is not. Every conversation on a sandbox
runs `install_broker_ca/2`, on provision **and** on reattach (ADR 0023), so
two runs at once means one truncates what the other is still writing and the
rename publishes the result. On the reporter's sandbox, 221 reads out of
1,408 saw eight certificates or fewer where there should be 122. A client
built from that bundle holds no `Managoat Broker CA` and fails every request
with `invalid peer certificate: UnknownIssuer`, then waits out a full idle
timeout before retrying.

Two changes, both in the one `bash -lc` string:

- `flock` around the install and the rebuild, so two conversations cannot
  interleave. Best effort — `flock -w 120 9 || true` — so a sandbox without
  `flock` provisions the way it did before rather than failing. The work
  stays `&&`-chained, so the exit status the caller sees is unchanged.
- `cmp -s` the staged PEM against what is installed, and skip both the
  install and the rebuild when they match. The CA is derived from the
  deployment's master key, so it is the same bytes for every conversation on
  every sandbox: after the first provision there is nothing to do, which is
  what takes the steady-state race window to zero rather than merely
  narrowing it.

Verified on `ubuntu:24.04` (the image `images/daytona` builds from): `flock`
and `cmp` are both present, a second run blocks until the first releases,
and an unchanged CA runs neither `install` nor `update-ca-certificates`.

## `env_vars` could not override the broker's CA variables

`SpriteEnv.build/4` appended the broker's ten pairs after the environment's
`env_vars` and the decrypted secrets, so `SSL_CERT_FILE` set in an
environment was written to `/home/sprite/.env` and then overwritten one line
later. Pointing codex at a stable bundle through the documented setting was
impossible; the reporter had to reach for `CODEX_CA_CERTIFICATE`, which
Fountain happens not to set.

The issue asks to emit the broker's variables first. Doing that wholesale
would let an `env_vars` entry replace `HTTPS_PROXY`, which is an opt-out of
the chokepoint (ADR 0019). So `Broker.sandbox_env/1` is split instead:

- `Broker.ca_env/0` — `SSL_CERT_FILE`, `REQUESTS_CA_BUNDLE`,
  `CARGO_HTTP_CAINFO`, `NODE_EXTRA_CA_CERTS`, `UV_NATIVE_TLS`. Hints to a
  client, emitted **before** `env_vars`, so a tenant may name its own bundle.
  Naming one without the broker root costs that tenant its own egress and
  nobody else's, which they could do from inside the sandbox anyway.
- `Broker.proxy_env/1` — the proxy URL, its lowercase twins and `NO_PROXY`.
  Emitted last, as before.

`sandbox_env/1` still returns both in the order it always did, so nothing
that consumes the whole list changes.

`Egress.reprepare/5` needed care: it stripped every broker key and
re-appended `sandbox_env/1`, which would have put the CA defaults back at the
end on the first token refresh — the same bug, an hour into a conversation
rather than at provisioning, where it would have been noticed. It now
replaces only the proxy half, and supplies the CA defaults only to a list
that carries none.

## Testing

- `mix test` (core suite), `mix credo --strict`, `mix format --check-formatted`.
- Each of the three production changes was reverted in turn to confirm the
  new tests fail without it: the CA install command (2 failures), the
  `SpriteEnv` ordering (1), and `reprepare/5` (3).

Refs #1674. The codex websocket half of that issue is a separate PR; the
switchyard note at the bottom of the issue is a feature request and is not
addressed here.

https://claude.ai/code/session_01XT411VudYs1hw69mDDhvrm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,10 @@ erl_crash.dump
 # Temporary files, for example, from tests.
 /tmp/
 
+# ExUnit's `@tag :tmp_dir` writes under the app that runs the test, not the
+# umbrella root, and the behavioural CA install test uses one.
+apps/*/tmp/
+
 # Database files
 *.db
 *.db-*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,19 +31,20 @@ upgrade, is in
 ### Fixed
 
 - The broker's root CA is installed under a lock, and the operating-system
-  trust store is rebuilt only when the installed CA differs or no successful
-  rebuild has been recorded. Conversations sharing a sandbox each ran
+  trust store is rebuilt only when the bundle on the machine is not the one
+  that CA produces. Conversations sharing a sandbox each ran
   `update-ca-certificates`, which builds the bundle at a fixed temporary
   path, so two runs at once published a truncated one. A client that read it
   in that state trusted no broker root and failed every request with
-  `UnknownIssuer`. A sandbox already damaged this way repairs itself on its
-  next provision, and each install stages the CA under a path of its own
-  (#1674).
+  `UnknownIssuer`. A sandbox whose bundle is damaged, whether before this
+  change or afterwards by a package install or a setup script, repairs itself
+  on its next provision or wake (#1674).
 
 - An environment's `env_vars` can override the broker's CA variables
   (`SSL_CERT_FILE` and the rest). They were written before the broker's own,
-  so the documented setting silently did nothing. The proxy variables still
-  win over everything: they are what makes egress brokered (#1674).
+  so a value set for one of those names silently did nothing. The proxy
+  variables still win over everything: they are what makes egress brokered.
+  Both halves of the rule are now in the manual, under Secrets (#1674).
 
 - Codex launches on Sprites with inherited and ambient capabilities cleared,
   allowing its bubblewrap sandbox to start without falling back to approval
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,11 +31,14 @@ upgrade, is in
 ### Fixed
 
 - The broker's root CA is installed under a lock, and the operating-system
-  trust store is rebuilt only when the CA on disk differs. Conversations
-  sharing a sandbox each ran `update-ca-certificates`, which builds the
-  bundle at a fixed temporary path, so two runs at once published a
-  truncated one. A client that read it in that state trusted no broker root
-  and failed every request with `UnknownIssuer` (#1674).
+  trust store is rebuilt only when the installed CA differs or no successful
+  rebuild has been recorded. Conversations sharing a sandbox each ran
+  `update-ca-certificates`, which builds the bundle at a fixed temporary
+  path, so two runs at once published a truncated one. A client that read it
+  in that state trusted no broker root and failed every request with
+  `UnknownIssuer`. A sandbox already damaged this way repairs itself on its
+  next provision, and each install stages the CA under a path of its own
+  (#1674).
 
 - An environment's `env_vars` can override the broker's CA variables
   (`SSL_CERT_FILE` and the rest). They were written before the broker's own,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,18 @@ upgrade, is in
 
 ### Fixed
 
+- The broker's root CA is installed under a lock, and the operating-system
+  trust store is rebuilt only when the CA on disk differs. Conversations
+  sharing a sandbox each ran `update-ca-certificates`, which builds the
+  bundle at a fixed temporary path, so two runs at once published a
+  truncated one. A client that read it in that state trusted no broker root
+  and failed every request with `UnknownIssuer` (#1674).
+
+- An environment's `env_vars` can override the broker's CA variables
+  (`SSL_CERT_FILE` and the rest). They were written before the broker's own,
+  so the documented setting silently did nothing. The proxy variables still
+  win over everything: they are what makes egress brokered (#1674).
+
 - Codex launches on Sprites with inherited and ambient capabilities cleared,
   allowing its bubblewrap sandbox to start without falling back to approval
   escalation on every command (#1672). Existing adapter processes need a restart.
--- a/apps/fountain/lib/fountain/broker.ex
+++ b/apps/fountain/lib/fountain/broker.ex
@@ -363,10 +363,15 @@ defmodule Fountain.Broker do
   end
 
   @doc """
-  The environment pairs a brokered sandbox gets. `HTTPS_PROXY` carries the
-  session token (as `http://<token>:<vault>@host:port`), so it is process-only
-  (`Identity.@process_only`); the lower case twins are for apt and the tools
-  that only read those.
+  The environment pairs a brokered sandbox gets: `proxy_env/1` and then
+  `ca_env/0`, in the order this has always returned them.
+
+  The two halves are not equal, and `Fountain.Conversations.SpriteEnv.build/4`
+  puts them on either side of the tenant's own values rather than taking this
+  list whole. `HTTPS_PROXY` carries
+  the session token (as `http://<token>:<vault>@host:port`), so it is
+  process-only (`Identity.@process_only`); the lower case twins are for apt
+  and the tools that only read those.
 
   The CA variables make each toolchain trust the broker's MITM certificate.
   `install_broker_ca` puts the CA in the OS trust store, which curl, git and
@@ -379,15 +384,20 @@ defmodule Fountain.Broker do
   `invalid peer certificate: UnknownIssuer` the moment it reaches a MITM'd host.
   """
   @spec sandbox_env(session()) :: [{String.t(), String.t()}]
-  def sandbox_env(%{token: token, vault: vault}) do
-    url = proxy_url_with(token, vault)
+  def sandbox_env(%{token: _, vault: _} = session), do: proxy_env(session) ++ ca_env()
 
+  @doc """
+  The half that points a toolchain at a trust store holding the broker's CA.
+
+  These are **defaults**: `SpriteEnv.build/4` emits them before the
+  environment's `env_vars` and the decrypted secrets, so a tenant that has a
+  reason to name its own bundle can (#1674). Naming a bundle without the
+  broker CA costs that tenant its own egress and nobody else's — the values
+  are hints to a client, not the chokepoint.
+  """
+  @spec ca_env() :: [{String.t(), String.t()}]
+  def ca_env do
     [
-      {"HTTPS_PROXY", url},
-      {"HTTP_PROXY", url},
-      {"https_proxy", url},
-      {"http_proxy", url},
-      {"NO_PROXY", "localhost,127.0.0.1"},
       {"NODE_EXTRA_CA_CERTS", @ca_path},
       {"SSL_CERT_FILE", @system_ca_bundle},
       {"REQUESTS_CA_BUNDLE", @system_ca_bundle},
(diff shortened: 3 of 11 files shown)


jhgaylor:
Accepted the proposed fix and covered it with tests in e9e7d018. All three points were right, and the second one is the important one:

- **Per-invocation staging.** `/tmp/agent-vault-ca.crt` was a fixed path written *outside* the lock — the same shared-fixed-path bug as the trust bundle, one level up. Two conversations writing it at once can hand each other a torn file for `cmp` to compare and `install` to publish.
- **The success marker.** Matching CA bytes prove the source is right, not that the bundle was rebuilt from it. A sandbox already damaged by the old race has the correct CA file and a truncated bundle, so the `cmp` skip on its own would have left it broken for the life of the machine and the fix would have looked like it worked on every new sandbox. That was a real gap in the original change, not a hardening suggestion.
- **The timeout.** 60s against a 120s lock wait meant the caller gave up before the lock did.

One thing checked before accepting, since the marker lands in the certificate directory: `update-ca-certificates` scans `find -L "$LOCALCERTSDIR" -type f -name '*.crt'` (read from the `ca-certificates` package on `ubuntu:24.04`, the base `images/daytona` builds from), so `agent-vault.crt.trust-store-ready` is ignored rather than parsed as a certificate.

Tests updated in the two files the change put outside its one-file limit: the exact-command assertion, the staging-path assertion, and the brokered-server walk. Both new behaviours verified by reverting them — the shared staging path fails one test, dropping the marker check fails two. Full core suite green (4,374 tests), `credo --strict` clean, formatted.

jhgaylor:
Addressed in 7d66246c. The two structural findings were right, and the first one is the one that mattered.

**`a4d79b85` / `7673d56a` (the marker records the wrong fact).** Correct, and the fix commit's `test -f` only narrowed it. `test -f` says that *a* rebuild once succeeded; it says nothing about the bundle now. `build_apt_commands/1` and `run_setup_script/4` both rebuild that bundle outside `@ca_lock`, so post-marker corruption was permanent where the old unconditional rebuild had repaired it. Took remediation (a): the marker now holds `sha256sum` of `/etc/ssl/certs/ca-certificates.crt`, and the skip is `cmp -s <staging> <ca> && sha256sum -c --status <marker>`. Any later change to the bundle — by apt, by a tenant script, or by a torn write — re-arms the guard, and the next provision or wake repairs the machine. That closes all three residuals rather than two of them.

**`721a6d04` (the flock timeout is swallowed).** Also right, and it interacted with the above: `|| true` treated an absent binary and an expired 120-second wait as one condition, and stamped the marker either way, so a single racy unlocked rebuild could suppress every future repair. It also made the docstring's own justification false — without `flock` it did *not* provision the way it did before. Rather than fail closed, the rebuild still runs when the lock was not held but records nothing, so the next conversation rebuilds too. That is exactly the unconditional rebuild that predates this PR, which is what the compatibility contract promised, and availability is preserved for the cold-provision case where refusing to install the CA would leave the conversation unable to dial out at all.

Verified the six cases in an `ubuntu:24.04` container before writing the Elixir: first run, steady state, bundle corrupted after the marker, steady state again, lock held by another process, and `flock` removed from `PATH`. The last two rebuild and stamp nothing.

**`00307d06` / `90995e97` / `e0d2750a` (the wording and the unpublished rule).** Both halves taken. "documented setting" is gone from the changelog and the `SpriteEnv.build/4` docstring — it overclaimed, since no page named those five variables. And the rule is published: `docs/concepts/secrets.md` now states both halves beside the existing `env_vars` paragraph, since it is tenant-visible and its sibling (a vault beats an environment) is already published. All three prose gates pass on the new text; the four `STE.SentenceLength` errors in the tree are pre-existing, in `deploy.md` and `acp.md`.

**`42dbce75` (the duplicated test).** Taken. Its four explanatory comments moved onto the exact-match test and it is gone; the per-invocation staging assertions were already there.

**`92b30e05` (the leaked staging file).** Taken, with the caveat the finding notes: a best-effort `rm` on the `{:error, reason}` arm, which only helps if the sandbox is reachable by then.

Full core suite green (4,374 tests), `credo --strict` clean, formatted. Reverting the digest marker, the lock-aware stamp or the cleanup each fails a test.

jhgaylor:
Accepted in 8491f5b9, plus 7f8f45dd. Full core suite green (4,375 tests), `credo --strict` clean, formatted.

The behavioural test is a real improvement and I have kept it as proposed. Running the generated command through a real shell with real `flock`, `cmp` and `sha256sum` is exactly the matrix this change was verified against by hand in an `ubuntu:24.04` container — rebuild once, skip on an unchanged bundle, repair a corrupted one, exit 3 with the marker removed, then a successful retry — and it is much better committed than described. It also passed on the proposal's own commit in CI, which is the evidence the commit message said it could not offer.

Two things it needed that the one-file policy kept out of reach:

**It cannot run on a Mac.** BSD `install` has no `-D`, and macOS ships neither `flock` nor `sha256sum`, so it fails three ways locally and would have failed `mix precommit` for every macOS contributor — which CLAUDE.md tells them to run. Skipped where those tools are absent rather than weakened; stubbing `flock` onto `PATH` would have removed the one property the test exists to check. Verified all four tools are present on `ubuntu:24.04`, which is what CI and the sandbox images run.

**`@tag :tmp_dir` writes under the app, not the umbrella root**, so it left an untracked `apps/fountain/tmp/` in every tree. Added to `.gitignore` (7f8f45dd).

On the replaced comment: fair that it overstated what the string assertion checks. I have kept a shorter version pointing at where each property is now actually verified, since `install_broker_ca/2`'s docstring carries the reasoning for why each matters on a shared sandbox and a reader of the test file should be sent there rather than left with "pinned here".

jhgaylor:
Rebased onto main (so this now reviews under `max_files: 3` from #1718) and addressed the round in b0a7ab50. Full core suite green (4,375), `credo --strict` clean, formatted.

**`bf28cdea` / `75e1ffc3` / `33a587a4` — the lock timeout.** Right, and the holder-side window is the part I had missed: `safe=1` only fixed the waiter's durable half, while the holder stamps `sha256sum` after its own rebuild returns, so a waiter publishing a truncated bundle in between gets it recorded as good and nothing repairs it. That is precisely the self-repair the changelog promises. `flock` failing is now two conditions rather than one — absent binary rebuilds unlocked and stamps nothing, an expired wait exits 75 and touches nothing. Covered in the behavioural test with a real lock holder, shortening only the wait.

**`4009546e` — the proxy paragraph.** Correct and it was mine. The claim holds for the process env and fails for `/home/sprite/.env`, for exactly the four names it fails for: `Identity.disk_env/1` strips them because the broker's value carries the session token, so the tenant's entry is the only assignment left in the file and a `source .env` picks it up. Stated now, with why it is not an egress escape.

**`1ecbfdf3` — `env_keys/0`'s docstring.** Fixed; its one caller is the sudoers `env_keep` line, which has the opposite constraint.

**`be3758a7` — the empty `split_brokered/1` clause.** Dropped.

**`4e14da6f`** asks for no change. Noted, and its observation about the two-file transcription cost is the thing #1718 was merged to relieve.

One correction to my own process, from the Review Loop skill added in #1719: an earlier `/review-loop resolve` of mine was rejected for naming a stale run ID, so that finding was never recorded as decided. Re-issuing against the current run where it applies.

jhgaylor:
Jake asked me to collect occasional feedback here while I improve Review Loop. For whoever is working this PR: after your next normal review round, please reply with what helped, what required manual repair, and whether the loop stopped when it should. A run ID or comment link is enough; no extra review run needed.

I’m implementing service-enforced checks before bot pushes and investigating repeated rounds, followed by shorter output, cumulative cost and recovery. The new contributor skill and three-file allowance help, but do not solve those service gaps. I’ll check back after material changes.

jhgaylor:
Merging this with an admin override, on the maintainer's explicit authorization. Recording exactly what I overrode and why, so the call can be checked.

**One finding on this revision was real, and is fixed.** `broker_prepare/1` runs before `build_sprite_env/5` on both entry paths (`ConversationServer` lines 863 and 1114), so the defensive `else: Broker.ca_env()` arm in `reprepare/5` was unreachable. Removed, with the reason recorded in the docstring. A second low — the "documented setting" wording — I confirmed was already gone from both the CHANGELOG and the `SpriteEnv` docstring on this revision.

**The high finding and four of the mediums evaluate revisions that no longer exist.** I checked each quoted string against `b0a7ab50`, the revision the summary names:

| Finding | Severity | Evidence quoted | Present on `b0a7ab50`? |
| --- | --- | --- | --- |
| `7673d56a` | high | `flock -w 120 9 \|\| true; cmp -s … \|\| { install && update-ca-certificates; }` | no |
| `03c8ca1d` | medium | the same pre-marker command | no |
| `a4d79b85` | medium | `test -f <marker>` / `sudo touch <marker>` | no |
| `b8fe66d6` | medium | `test -f MARKER` / `sudo touch MARKER` | no |
| `721a6d04` | medium | `flock -w 120 9 \|\| true` | no |
| `75e1ffc3` | medium | `safe=0; flock … && safe=1;` | no |

What the revision actually builds is `( trap … EXIT; safe=0; if command -v flock >/dev/null 2>&1; then flock -w 120 9 || exit 75; safe=1; fi; { cmp -s <staging> <ca> && sha256sum -c --status <marker> 2>/dev/null; } || { … && { [ "$safe" = 1 ] && sudo sh -c 'sha256sum <bundle> > <marker>' || true; }; } ) 9><lock>`.

Both remedies these findings ask for are already in it, and were applied in response to the earlier rounds that first raised them:

- **"Gate the skip on the trust store, not the CA file"** (`7673d56a` remediation (a), `a4d79b85` remediation (a)) — the marker holds `sha256sum` of `/etc/ssl/certs/ca-certificates.crt`. A rebuild by apt's `ca-certificates` postinst or by a tenant setup script changes that bundle and re-arms the guard, so the self-heal these findings say was lost is exactly what the digest restores.
- **"Distinguish a missing `flock` from a failed acquisition; a timeout must exit without rebuilding"** (`721a6d04`, `75e1ffc3`, `b8fe66d6`) — a timeout exits 75 and touches nothing; only an absent binary rebuilds unlocked, and it stamps nothing, which is the pre-PR behaviour the docstring claims for that case. Verified across all six paths in an `ubuntu:24.04` container, and covered in the behavioural test with a real lock holder.

I have not resolved these through `/review-loop resolve`, because the honest record is not "the maintainer waived a defect" — it is that the defect was fixed and the finding is looking at older code.

**Evidence for the merge:** full core suite green, `credo --strict` clean, `mix format --check-formatted` clean, all three prose gates clean on the changed pages, and CI green on the pushed revision. Every behaviour added across this PR was verified by reverting it and confirming a test fails.

**Confirmed on the merged revision `ce88f893`:** all six Linux partitions report `0 failures` and no skips, so the behavioural CA-install test — which cannot run on macOS, and which exercises rebuild-once, skip-on-unchanged, repair-after-corruption, exit-3-then-retry and refuse-when-locked through a real shell — ran and passed on Linux.

One thing recorded rather than dismissed: an earlier local full-suite run on this branch reported a single failure that a clean rerun did not reproduce (4,374 tests, 0 failures). Its assertion and seed were lost to log truncation on my side, so there is nothing actionable to file; noting it here so the observation is not silently dropped.

jhgaylor:
If another normal PR reproduces either reported cost—repairs leaving CI red or optional findings preventing completion—please leave the PR/run link here. Jake asked me to keep collecting feedback; no extra review run is needed.

PR #1718, opened 2026-09-07
fix(review-loop): allow source and test repairs

jhgaylor:
The one-file repair cap prevents a source fix from including its regression test. Raise `fixes.max_files` to 3 and update all four reviewer instructions to follow the trusted policy's scope, including necessary tests. Ask reviewers for concise findings that retain evidence, uncertainty and material tradeoffs.

This changes only the file-count allowance; path exclusions, verification, models and the medium blocking threshold are preserved. A larger repair still needs a human decision. The policy takes effect from the approved base after merge; it cannot authorize this PR itself.

Validation: Review Loop's actual policy parser accepts the YAML; a parsed comparison confirms `max_files: 1 → 3` is the only policy-value change. `git diff --check` passed. Local `mix precommit` could not complete because this fresh worktree lacks dependencies; CI will run the repository checks.

### DIFF
--- a/.github/review-loop.yml
+++ b/.github/review-loop.yml
@@ -43,7 +43,8 @@ fixes:
     - sdk/conformance/matrix.json
     - "**/pyproject.toml"
     - "**/uv.lock"
-  max_files: 1
+  # Allow a small repair to include its regression test and supporting fixture.
+  max_files: 3
   max_bytes: 256000
 verification:
   instructions: .github/review/verification.md
--- a/.github/review/README.md
+++ b/.github/review/README.md
@@ -16,10 +16,11 @@ findings, accepts low/info suggestions as nonblocking, and retains explanations.
 It resolves service-owned threads when policy permits; human objections remain
 blockers until a maintainer addresses them.
 
-Clear medium-or-higher defects can receive an automatic one-file fix. Reviewers
-must flag uncertain remedies, multi-file changes, public-contract choices and
-protected changes for a human. The host enforces the one-file bound, permitted
-paths, three-round limit, exact-revision checks and final approval. Confidence
+Clear medium-or-higher defects can receive an automatic fix within the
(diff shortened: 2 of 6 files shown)


jhgaylor:
Picked this up because the one-file cap is currently costing a full CI cycle per round on #1682 and #1683: four proposals in a row could not pass CI as pushed, three of them because the repair needed its test and one because it needed the formatter. Raising `max_files` to 3 is the right fix.

Addressed three of the four findings in 6fe85aff.

**`b36e4e2b` (the README still states the bound).** Real drift, and the remediation is the right shape — the count now lives only in `review-loop.yml`. Took all three clauses: "an automatic one-file fix" → "an automatic fix within the configured file and path bounds", "multi-file changes" → "changes exceeding those bounds", "the one-file bound" → "the configured file bound".

**`e3ed8e9b` ("intended" dropped).** Agreed the qualifier was load-bearing — a fix for a correctness defect necessarily changes observed public behavior, and "intended" is what distinguishes the contract from the bug. Restored in `simplicity.md`, and took the reverse alignment too: `api-compatibility.md` now says "intended" as well, so the two files state one rule rather than two readings of it.

**`c9e352c8` (unwrapped lines).** Rewrapped all three to the surrounding width. Worth noting the finding's own observation: the docs-prose jobs filter on `^docs/|^apps/[^/]+/docs/`, so nothing would ever have reported these.

**`893a5819` (the duplicated pair of lines).** No change, as the finding says. Each reviewer is handed exactly one `instructions:` path and there is no shared-fragment field, so per-file duplication is the only way to express a shared rule today. A shared prelude is a service and schema change, not a repository one.

The policy value itself is untouched by this commit — still `max_files: 1 → 3` as the only changed policy value, with path exclusions, `max_bytes`, verification, models and the medium blocking threshold all preserved.

Issue #1668, opened 2026-09-06
A model variant suffix defeats #1663's containment fix: opus[1m] still fails every turn


jhgaylor:
## Symptom

#1663 fixed the morning's model-enforcement outage for the common case, but a turn still failed **after that fix was deployed** (pods started 11:38 UTC, event at 11:48 UTC):

```
turn|failed  Could not select model claude-opus-5: Runtime confirmed a different
             model: opus[1m]. No prompt was sent.
```

## The `[1m]` is not an ANSI escape

Worth stating because it looks like one. It is literal text — Anthropic's 1M-context designation. Confirmed against the raw bytes rather than the rendered string:

```sql
SELECT encode(convert_to(data,'UTF8'),'escape') FROM log_events
WHERE stage='model' AND state='failed' AND data LIKE '%1m%';
-- ...Runtime confirmed a different model: opus[1m]...
-- no \033, so no ESC byte: the brackets are content
```

## Cause

`Managoat.ACP.Peer.normalise_model/1` strips every non-alphanumeric character:

```elixir
defp normalise_model(model) when is_binary(model) do
  model |> String.downcase() |> String.replace(~r/[^a-z0-9]/, "")
end

defp same_model?(effective, requested) do
  a = normalise_model(effective)     # "opus[1m]"      -> "opus1m"
  b = normalise_model(requested)     # "claude-opus-5" -> "claudeopus5"
  a != "" and b != "" and (String.contains?(a, b) or String.contains?(b, a))
end
```

`"opus1m"` neither contains nor is contained in `"claudeopus5"`, so `same_model?/2` returns false and the turn fails with no prompt sent.

The 0.2.2 containment fix handles a runtime answering *less precisely* (`opus` for `claude-opus-5`). It does not handle a runtime answering with a **variant suffix**, where stripping the separators fuses the suffix onto the family name and destroys the containment relationship the check depends on.

## Fix

In `managoat_acp`, split the variant off before comparing rather than dissolving it into the token — e.g. strip a trailing bracketed qualifier, or compare on segments instead of a fused string so `opus` + `1m` cannot glue together. Whatever the shape, the property to test is that a variant suffix on the confirmed model still agrees with the requested id, while a genuinely different family still fails.

Add the case to the library's own suite: `opus[1m]` vs `claude-opus-5` agrees; `claude-haiku-4-5` vs `claude-opus-5` still does not.

## Impact

[...]

Issue #841, opened 2026-08-19
onboarding: ask for Anthropic only; collect other providers' keys the first time a model needs them


jhgaylor:
Step 1 asked for all four inference credentials at once. A new user needs exactly one (Anthropic) for a Claude teammate; the rest should be collected just in time — when an agent's model first needs that provider — rather than up front. Client-side in fountain-team the brain picker already marks providers without a key; the agent form here should prompt inline the same way.

Issue #554, opened 2026-08-05
agent.model accepts any provider/model_id string — no known-model validation, suggestions, or spawn-time error surfacing


jhgaylor:
## Problem

The only validation on `Agent.model` is a format regex (`apps/fountain/lib/fountain/agents/agent.ex:56`):

```elixir
validate_format(:model, ~r{^[a-z0-9_-]+/[a-z0-9._-]+$})
```

`banana/not-a-real-model` passes. Consequences:

- **Typos fail late and opaquely.** A bad model ID on the `opencode` runtime fails at spawn/run time inside the sprite (opencode or the provider API rejects it), not at agent-creation time, and the error surfaces only in the conversation log.
- **Unknown provider prefixes silently export no credentials.** `open_code.ex` maps `anthropic`/`openai`/`google` prefixes to API-key env vars and falls through to `[]` for anything else — a misspelled provider (`anthopic/...`) means the sprite runs with no inference credentials and fails with an auth error rather than a validation error.
- **No affordance in the UI.** The agent form is a bare text input with placeholder `anthropic/claude-sonnet-4-6` (`apps/fountain/lib/fountain_web/live/agents_live/form.ex:469`) — nothing suggests valid values.

## Proposed

- Curate a per-provider known-models list (or at least known providers) and validate the provider segment at changeset time; warn-don't-block on unknown model IDs so new models aren't gated on a deploy.
- Surface suggestions in the form UI (`<datalist>` or select-with-custom-option) sourced from the same list.
- Cross-check provider vs. runtime once #553 makes the field meaningful on all runtimes (e.g. reject `openai/...` on the `claude` runtime).

Blocked-ish on #553 for the runtime/provider cross-check; the provider-prefix validation and UI suggestions can land independently.

PR #1724, opened 2026-09-07
fix: drop claude-fable-5-1, an id the pinned adapter refuses

### LINKED ISSUE #1669: claude-fable-5-1 is in the model catalog but the pinned claude adapter refuses it
## Symptom

Post-deploy (pods started 11:38 UTC, event at 11:42 UTC):

```
turn|failed  Could not select model claude-fable-5-1: Invalid value for config option
             model: claude-fable-5-1. No prompt was sent.
```

`claude-fable-5-1` is listed in the catalog on `main`:

```
apps/fountain/lib/fountain/agents/model_catalog.ex:80
      claude-fable-5-1
      claude-opus-5
      claude-sonnet-5
```

Added by `80836af8` ("Add Claude Fable 5.1 to model catalog") and kept by #1663 while that PR removed `claude-sonnet-4-6`, `claude-opus-4-7` and `claude-opus-4-8` for exactly this defect.

`claude-fable-5` (not in the catalog, but pinned on one agent) fails the same way.

## Why this recurred

This is Defect 2 of the morning outage, reappearing inside the PR that fixed it. From the post-mortem:

> **A model suggestion has to clear two gates**, and only the second one decides a turn: the provider must serve the model, *and* the pinned ACP adapter must accept it at `session/set_model`.

The catalog is vetted with real inference calls per provider. That establishes gate one. `claude-fable-5-1` is a real published Anthropic model id — which is precisely why it passed review — but the pinned `claude-agent-acp` build refuses it at `session/set_model`, and that is the gate that decides the turn.

## Fix

1. Remove `claude-fable-5-1` from the catalog until the pinned adapter accepts it. Re-add it in the same PR that moves the pin, not before.
2. Re-point the agent pinned to `claude-fable-5`.

## Standing check this keeps earning

The cheap detector named in the morning post-mortem, which would have caught both this and the 4-6/4-7/4-8 pins weeks earlier — refusal rate per requested model, against turns for the same model:

```sql
SELECT (data::jsonb->>'requested') AS requested,
       (data::jsonb->>'detail')    AS detail,
       count(*), max(inserted_at)
FROM log_events
WHERE stage='model' AND state='failed' AND inserted_at > now() - interval '24 hours'
GROUP BY 1,2 ORDER BY 3 DESC;
```

Worth wiring as an alert rather than run by hand after each outage. A fallback hid the previous instance of this for weeks — 289 refusals for `sonnet-4-6` alone over 21 days, every one silently answered — and now that enforcement is hard, the same class of defect fails turns instead of hiding.

Related: #1640, #1663, and the variant-suffix bug filed alongside this.

jhgaylor:
## The bug

`claude-fable-5-1` was added to the catalog on 2026-09-06 (#1659) from Anthropic's published model id. It ran turns the next day and the pinned `claude-agent-acp` refused every one:

```
Could not select model claude-fable-5-1: Invalid value for config option
model: claude-fable-5-1. No prompt was sent.
```

Since #1640 a refusal fails the turn **before any prompt is written**, so suggesting it is an outage for every agent that takes it.

## Why it survived the clean-up

This is the same defect as `claude-sonnet-4-6`, `claude-opus-4-7` and `claude-opus-4-8` — and #1663 removed those three while leaving this one in, on the same day, in the same file, directly under a comment explaining the trap.

The reason is worth writing down: a **published provider id looks like the check has already been done.** `ModelCatalog` already says a suggestion has to clear two gates, and the provider serving a model is gate one. The adapter's accepted set is the gate that decides a turn, and it was never checked for this id. The old comment recorded that honestly — *"It has run no turns here, so it is neither confirmed nor known-refused"* — and the answer to that state is to wait for evidence, not to ship it as a suggestion. The evidence has now arrived and it is a refusal.

## Changes

- **Catalog** — remove `claude-fable-5-1`, with the comment updated to record the observed refusal and why it survived #1663.
- **Guard test** — add `anthropic/claude-fable-5-1` to `@refused_by_pinned_adapters`, plus `anthropic/claude-fable-5`, which was never suggested but which one agent is pinned to by hand and which is refused the same way. Listing it stops anyone adding it.
- **Docs** — drop it from the suggested-models table in `docs/catalog/runtimes/index.md`.

## Verification

The guard was verified by re-adding the id rather than by trusting a green run — a guard that stopped guarding looks exactly like one that finds nothing:

```
1) test no suggestion is an id the pinned adapters are known to refuse
   anthropic/claude-fable-5-1 is suggested again, but the pinned ACP adapter
   refused it on 2026-09-06. A refusal fails the turn before any prompt is
   sent (#1640), so this suggestion is an outage for every agent that takes it.
```

Tests: `test/fountain/agents/` + `docs_test.exs` 93 passing, `catalog_controller_test.exs` + `docs_test.exs` 32 passing, format and `compile --warnings-as-errors` clean. Vale totals are byte-identical before and after the docs edit (5580/4/64/5512), so this introduces no prose findings.

## Follow-up, not in this PR

One prod agent is pinned to `anthropic/claude-fable-5`. Re-pointing it is a tenant-data change and goes through `Agents.update_agent/3` so it audits, never psql — tracked on #1669.

The other half of yesterday's residual model failures is the `opus[1m]` variant-suffix bug, fixed upstream in managoat/managoat_acp#8 (0.2.3) and tracked here as #1668.

Closes #1669

https://claude.ai/code/session_017Zi4Vtq9E7BEfsABMkkvWz


jhgaylor:
Review Loop run 2 was right on all four actionable findings. Verified each against the code before acting; pushed `836b9913`.

**Fixed here** — the two that are a live outage and directly adjacent:

- **The prefilled default.** `form.ex:26,64` defaulted to `anthropic/claude-sonnet-4-6`, so New agent → type a name → save produced an agent whose every turn failed at `session/set_model`. Now `anthropic/claude-sonnet-5`. The corroborating tell was real: the form's own `unknown_model?/1` returned true for its own default, so "Not one of the models Fountain lists" was firing on the value the form supplied.
- **The placeholders.** `form.ex:189,191` named `openai/gpt-5.3-codex` and `anthropic/claude-sonnet-4-6`. Now `openai/gpt-6-astra` and `anthropic/claude-sonnet-5`.

**The guard gap, fixed as suggested.** `@refused_by_pinned_adapters` moves out of the catalog test into `Fountain.RefusedModels` (`test/support/`), so the new form guard reads the same registry rather than a second copy. Its moduledoc records why it is a registry and not a catalog-test detail.

I also found the finding implied but did not name: **the existing test had rotted into pinning the bug.** `"model field offers the curated models for the selected runtime"` derived its claude half from `ModelCatalog.suggestions/1` but hard-coded the codex half as `openai/gpt-5.3-codex`, so it kept passing after that id left the catalog — it was matching the placeholder. Both halves now come from the catalog.

Verified by reverting: with the old defaults restored the new guard fails and names the model. 115 tests pass across `agents/`, `agents_live_test.exs` and `docs_test.exs`; format clean.

**Split out to #1727** — the third finding, plus the rest of the second. The shipped `fountain` SKILL.md manifest, four `/help` pages and the OpenAPI description still name refused ids. Not fixed here because the repair spans four files and needs a maintainer decision: `agents.md:78` illustrates the codex runtime's model and **no codex-line openai id remains in the catalog**, so it has no mechanical substitute.

**On the `low · needs_human` finding** — `claude-fable-5` is not preventive. It was observed refused on a real turn at 11:45:46 UTC on 2026-09-06 (`Invalid value for config option model`), one turn, from the agent that is hand-pinned to it; `claude-fable-5-1` was refused at 11:42:46 the same way. Both dates in the map are observation dates, not add dates. The review was right that my comment did not say so, and that the removal rule cannot clear an id which is never suggested — zero of zero forever. Both points are now recorded in `Fountain.RefusedModels`, with the escape hatch stated: clear these on the next `claude-agent-acp` bump by driving one real turn each, not by waiting for a rate to move.

jhgaylor:
Generation 3: three findings actioned, one settled with data, one left for a maintainer. Pushed `cc643de8`.

**`Fountain.Agents.Starter` (medium · fix) — right, and the sharpest of the set.** `@model` becomes the model of the one agent every verified account owns (ADR 0038 decision 4), and the only assertion over it was `=~ ~r"^anthropic/"` — gate one, the half `ModelCatalog`'s own note calls necessary and not sufficient. Now checked against the catalog and the refused list, and named in the registry's guarded-surfaces list. The observation that a gap list written to stop a fourth recurrence omitted the highest-blast-radius surface is the part worth keeping.

**Refusal is the narrower invariant (medium · fix) — right, and I had it wrong.** My guard checked absence from `RefusedModels`. An id leaves the catalog two ways and that list sees one: `gpt-5-codex` was *retired* on 2026-08-22 while it was both the codex suggestion and the codex placeholder, and a retired id never becomes a refusal. The guard now asserts `ModelCatalog.known?/1`, which catches both paths, plus `refute html =~ "passed to the runtime as-is"` on a freshly mounted form — the hint firing on mount is the bug stated directly.

Verified by injecting one of each: a **retired** id into the codex placeholder and a **refused** id into the starter. Both fail, and the retired one is caught *only* by the membership check.

I did not take the "derive the placeholder from `suggestions/1` |> `List.first()`" alternative, for the reason given: the catalog is newest-first, so it would silently move the claude default from `claude-sonnet-5` to `claude-opus-5`. That is a cost-visible product change and should not ride along in a bug fix.

**Existing agents holding a refused id (medium · needs_human) — settled with a query, no remediation needed.** Run against prod over every id in `RefusedModels` plus the two retired ones:

| model | agents |
|---|---|
| `anthropic/claude-fable-5` | 1 |
| everything else | **0** |

`anthropic/claude-sonnet-4-6` returns **zero rows** — the default cohort the finding hypothesised does not exist, because #1663 already repointed the 47. `claude-fable-5-1` is zero. The single `claude-fable-5` row is the known hand-pinned agent, dormant since 2026-08-08 (one conversation, idle, no model events) and owned by another tenant, so repointing it is that account's configuration decision rather than mine.

**`claude-fable-5` (low · needs_human)** — its evidence line reads `observed at 3c5c0b45`, the previous head. `cc643de8`'s parent already recorded both observation dates (11:42:46 and 11:45:46 UTC on 2026-09-06, one real turn each), that they are observation and not add dates, and the escape hatch for the zero-of-zero problem: clear on the next `claude-agent-acp` bump by driving one real turn each. I believe that is remediation (c), and it is in `Fountain.RefusedModels`.

**The help/skill/OpenAPI surfaces stay on #1727.** Noted that a tracked follow-up does not repair them — agreed, and the SKILL.md case is the one I would prioritise since it is machine-consumed. Not folded in here because `agents.md:78` illustrates the codex runtime's model and **no codex-line openai id remains in the catalog**, so it needs a maintainer's substitute rather than a swap, which is the same reason the finding asks for a human.

115 tests pass across `agents/`, `agents_live_test.exs` and `docs_test.exs`; format clean.

jhgaylor:
/review-loop resolve ea9aed06-32c8-4330-b234-03befa5a9476 57f8bf913097b844e41689d7f32f620ff0449a94ba88a796a85fbe11d8f5a29b Settled by query, not by code: no agent holds a refused id except one known dormant row. Ran over every id in RefusedModels plus the two retired ones (gpt-5-codex, gemini-2.5-flash). anthropic/claude-sonnet-4-6 returns ZERO rows, so the hypothesised default cohort does not exist - #1663 already repointed the 47 and no agent has been created from the stale form default since. claude-fable-5-1 is zero. The single anthropic/claude-fable-5 row is the hand-pinned agent already named in RefusedModels: dormant since 2026-08-08 (one conversation, idle, zero model events) and owned by another tenant, so repointing it is that account's configuration decision. Full table in the PR comment above.

jhgaylor:
/review-loop resolve ea9aed06-32c8-4330-b234-03befa5a9476 5742bfac9c05bb63d2b2c07538ac774852f081a26eb6806c58d4768fadad00a7 Maintainer chose remediation (c): keep the entry, with a lighter removal bar recorded. Correcting one premise first - this is not a preventive entry. claude-fable-5 was observed refused on a real turn at 11:45:46 UTC on 2026-09-06 (Invalid value for config option model), from the agent hand-pinned to it; claude-fable-5-1 at 11:42:46 the same way. Both dates in the map are observation dates, not add dates, which the earlier comment failed to say. The zero-of-zero point stands and is now written into Fountain.RefusedModels: the refusal-rate check cannot clear an id that is never suggested, so these two are cleared by driving one real turn each on the next claude-agent-acp bump, not by waiting for a rate to move. Accepted cost: whoever wants to list Fable 5 re-pins it by hand first.

jhgaylor:
/review-loop resolve ea9aed06-32c8-4330-b234-03befa5a9476 62c92f52d940bf2a156e402bbfc153d4f87320a0d453d555ed7165945e18b3fb Maintainer decision: stays on #1727, not folded into this PR. The finding is correct that a tracked follow-up does not repair a live defect, and SKILL.md is the piece worth doing first since it is machine-consumed. Deferred because agents.md:78 illustrates the codex runtime's model and no codex-line openai id remains in the catalog - gpt-6-astra and gpt-5.5 are the openai entries - so that row needs a substitute chosen rather than swapped, which is why the finding asks for a human in the first place. #1727 carries the full file and line list, the blocker, and the ask to guard these by content scan over priv/help and priv/external_skills consuming Fountain.RefusedModels. This PR is already well past its original scope (catalog id removal), and the surfaces here are pre-existing rather than introduced by it.

jhgaylor:
Jake asked me to collect feedback during normal PR work. Pre-push checks and explicit finding reassessment are now deployed. After your next normal Review Loop run, please flag any repeated fixed finding or bot repair that leaves CI red. No extra run needed.

jhgaylor:
All three generation-5 findings are the same defect and they are right: **I deleted a test.** Fixed in `16c28ed3`.

Adding the model-default guard in `836b991` *replaced* `"prompts for the provider's key when the chosen model has none on the account"` instead of sitting beside it. My edit replaced the span between two anchors and that test sat inside the span.

The reason it went unnoticed is worth recording, because it defeats the obvious check: **the test count went up, not down.** The new guard landed in the same edit that removed the old test, so 115 passing looked like growth. I reported "115 tests, 0 failures" twice as evidence the change was sound, and the number was hiding the loss.

What was unasserted at `836b991` and `cc643de8`, per the finding: mount's `assign(:missing_credential, ...)`, its recomputation on `validate`, the rendered "No <provider> credential on this account yet" card, and the unknown-provider branch that must prompt for nothing — the whole inline credential collection path from #841. A regression in first-run onboarding would have shipped green.

Restored verbatim from the merge base, along with the `#554` comment above the curated-models test that the same edit took.

The check that would have caught this at the time, which I have now run:

```
$ git diff origin/main -- apps/fountain/test/fountain_web/live/agents_live_test.exs \
    | grep "^-" | grep -v "^---"
-      assert html =~ "openai/gpt-5.3-codex"
```

One deleted line — the hard-coded assertion that had rotted into pinning the bug. That is what the diff should have been from the start. A test-file diff with unexplained deletions is the signal; the passing count is not.

116 tests, 0 failures across `agents_live_test.exs`, `agents/` and `docs_test.exs`; format clean.

jhgaylor:
Right, and the convention was sitting directly above the gap: #1663 added its Unreleased entry in the same commit as its fix, and this PR's entry belongs immediately beneath it. Added in `f84801fd`.

Two entries, because two different readers need them:

- **The form default.** An operator reading the release notes would have seen the 2026-09-06 catalog clean-up and had no way to learn the prefilled default was serving a refused id too. That is the one route by which someone with agents created in that window would think to check them.
- **`claude-fable-5-1` leaving the catalog**, called out as an observable `GET /api/catalog` response change rather than an internal tidy-up.

The entry also records that **no saved agent used it**, since "do I need to repoint anything?" is the question a reader has on finishing that sentence — and it is the same answer the query in the earlier comment gave.

Gates run because `CHANGELOG.md` is a docs page here, embedded by `Fountain.Docs` through `extra_resources:` and served at `/docs`:

- `docs-style.py` — 1 problem, pre-existing, in `codex.md`, not mine
- `destink` — 112 pages clean, 33 rules
- `docs_test.exs` — 29 tests, 0 failures

Em dashes are house style in that file (375 of them), so the two in the new entries are consistent rather than an exception.

Generation 6 reported **0 fix findings**. This was the last item.

jhgaylor:
/review-loop resolve 7751d050-95e8-43e7-8919-198e6883389f 242358312dd64e4ca6736199dfa9bd317f0e77bce555dd16cfd487a8aa7eab47 Already fixed at the revision under review. The finding is reported against f84801fd, which is the commit that adds the changelog entries: git show f84801fd --stat is CHANGELOG.md, 19 insertions, and git show f84801fd:CHANGELOG.md contains 'The new-agent form no longer starts you on a model the adapter refuses'. Two Unreleased entries were added directly beneath the #1663 entry, per the convention the finding cites. One covers the form default and placeholders, the other records that GET /api/catalog no longer lists claude-fable-5-1 as an observable response change, and states that no saved agent used it so a reader knows there is nothing to repoint. Docs gates green: docs-style.py one pre-existing problem in codex.md, destink 112 pages clean, docs_test.exs 29 tests 0 failures.

jhgaylor:
/review-loop resolve 069f0d3a-03ab-430c-b5a5-d578656fe6cb 242358312dd64e4ca6736199dfa9bd317f0e77bce555dd16cfd487a8aa7eab47 Already fixed; re-reported because resolutions are recorded per run and the branch update to 639139dc started a new one. Verified at the reviewed revision: git show 639139dc:CHANGELOG.md contains 'The new-agent form no longer starts you on a model the adapter refuses'. Two Unreleased entries sit directly beneath the #1663 entry, per the convention this finding cites - one for the form default and placeholders, one recording that GET /api/catalog no longer lists claude-fable-5-1 as an observable response change, and stating that no saved agent used it so a reader knows there is nothing to repoint. Originally resolved on run 7751d050 (comment 5566151044). Docs gates green: docs-style.py one pre-existing problem in codex.md, destink 112 pages clean, docs_test.exs 29 tests 0 failures.

jhgaylor:
/review-loop resolve f20d5a83-e7a7-4f19-be09-b1ab225e562a 242358312dd64e4ca6736199dfa9bd317f0e77bce555dd16cfd487a8aa7eab47 Already fixed; third re-report of this finding id, each on a new run spawned by a branch update, because resolutions are recorded per run and main's ruleset requires branches be up to date. Verified again at the reviewed revision: git show 99b72741:CHANGELOG.md contains 'The new-agent form no longer starts you on a model the adapter refuses'. Added in f84801fd, two Unreleased entries beneath the #1663 entry - one for the form default and placeholders, one recording that GET /api/catalog no longer lists claude-fable-5-1 and that no saved agent used it. Previously resolved on runs 7751d050 and 069f0d3a.

PR #1725, opened 2026-09-07
deps: upgrade managoat_acp to 0.2.3

jhgaylor:
## Why

0.2.3 fixes the residual half of the model-enforcement outage. Claude confirms `opus[1m]` — the 1M-context build — for the model requested as `claude-opus-5`. `normalise_model/1` dropped every non-alphanumeric character, which fused the qualifier onto the family name:

| | normalised |
|---|---|
| `opus[1m]` | `opus1m` |
| `claude-opus-5` | `claudeopus5` |

Neither contains the other, so the containment comparison 0.2.2 introduced failed the turn — on a model that answers normally, with no prompt sent.

Observed in prod on 2026-09-06 at **11:48 UTC**, ten minutes after the pods carrying #1663 started. That is why the model-refusal rate did not go to zero when #1663 deployed.

Upstream: managoat/managoat_acp#8.

## The requirement moves too, not just the lock

`~> 0.2.2` already permitted 0.2.3, so this would work as a lockfile-only change. Bumping the requirement is deliberate: the fix is the *reason* for the upgrade, and a range that merely tolerates the version does not record that. Anyone regenerating the lock from a clean checkout should get the fix, not the oldest version the range allows.

## Verification

Against the **resolved dependency**, not the library's own suite — that suite is upstream's evidence, and what matters here is what Fountain actually compiles against:

```
$ grep -n "String.replace" deps/managoat_acp/lib/managoat/acp/peer.ex | tail -2
1114:    |> String.replace(~r/[\[(][^\])]*[\])]/, "")
1115:    |> String.replace(~r/[^a-z0-9]/, "")

$ grep -o '"managoat_acp": {:hex, :managoat_acp, "[^"]*"' mix.lock
"managoat_acp": {:hex, :managoat_acp, "0.2.3"
```

- `mix deps.unlock --check-unused` clean
- `mix compile --warnings-as-errors` clean
- `test/fountain/conversations/` + `test/fountain/agents/` — **783 tests, 0 failures**, 2 skipped

The agreement and substitution cases themselves are covered upstream, including the two that keep the comparison honest: a qualifier on a different family (`haiku[1m]` for `claude-opus-5`) still fails, and a bare `[1m]` still fails. Both were verified there by reverting.

## What this does and does not close

Closes the code half of #1668. The issue tracks reaching prod, so it stays open until this is deployed and the `opus[1m]` refusals stop appearing in `log_events`.

The other residual model failure, `claude-fable-5-1`, is #1724.

Refs #1668

https://claude.ai/code/session_017Zi4Vtq9E7BEfsABMkkvWz


jhgaylor:
Right, and the strongest argument for it is the one the finding makes: **two production breaks in the same code path on consecutive days, and CI here was green both times.** Added in `38a8c7bf`.

Six cases driven through a real `Managoat.ACP.Peer` over the writer function the peer takes as its transport, so no sandbox, no port and nothing stubbed. Both directions, because a test that pinned only agreement would have passed against 0.2.0's strict equality for `gpt-6-astra`:

- a canonical designation (`opus` for `claude-opus-5`) sends the prompt
- a variant qualifier (`opus[1m]`) sends the prompt
- a different family (`claude-haiku-4-5`, `haiku[1m]`) fails the turn and writes no prompt, so the enforcement #1640 added stays enforced

**Verified against the two releases that actually regressed**, rather than asserting it would have caught them:

| pin | result |
|---|---|
| `== 0.2.0` (strict equality) | **3 failures** — `opus`, `sonnet`, `opus[1m]` |
| `== 0.2.2` (fused qualifier) | **1 failure** — `opus[1m]` |
| `~> 0.2.3` | 0 failures |

On the scope question the finding raises — this does not duplicate the library's suite. What it pins is the half Fountain depends on and cannot see: `mix.exs` allows `~> 0.2.3`, so any 0.2.x resolves without a change here, and this behaviour decides whether a turn runs at all. The moduledoc records both incidents and why each direction is asserted, so the next person to touch it knows what the cases are for.

I did not take the "pin exactly" alternative — `== 0.2.3` would trade a silent regression for manual bumps forever and would still say nothing about what the behaviour should be.

jhgaylor:
/review-loop resolve c29ae90b-0845-4022-add8-0cd6480d8537 6d9fa3d1ee1e4333acfe1b4f417e4781ad9ebffa45d67d2dea7d18691b395d8e Already fixed at the revision under review. The finding is reported against 38a8c7bf, which is the commit adding apps/fountain/test/fountain/conversations/acp_model_contract_test.exs - six cases through a real Managoat.ACP.Peer over its writer-function transport, no sandbox and nothing stubbed. It asserts both directions, since a test pinning only agreement would pass against 0.2.0 strict equality for gpt-6-astra: opus and opus[1m] for claude-opus-5 send the prompt; claude-haiku-4-5 and haiku[1m] fail the turn and write no prompt, keeping the #1640 enforcement enforced. Verified against the two releases that actually regressed rather than assumed: == 0.2.0 gives 3 failures (opus, sonnet, opus[1m]), == 0.2.2 gives 1 (opus[1m]), ~> 0.2.3 gives 0. The moduledoc records both incidents and why each direction is asserted.

PR #1729, opened 2026-09-07
feat: suggest claude-fable-5-1 again, now that the refusal is understood

jhgaylor:
## Why

#1724 dropped `claude-fable-5-1` as "an id the pinned adapter refuses". The adapter version was half the story. The claude adapter advertises whatever the Claude Code binary bundled in its SDK reports, and that binary learns an org's "additional models" (Fable among them) from a fetch it makes *after* a session has started, caching it in `~/.claude.json` for the *next* launch. So the first session in a fresh sandbox never lists Fable on any adapter version, and the second one does — which is exactly the two refusals #1669 recorded, and why users read it as intermittent.

Measured 2026-09-07 with a fresh `CLAUDE_CONFIG_DIR` per run:

| adapter | cold | warm |
|---|---|---|
| 0.66.0 (old pin, CLI 2.1.220) | no Fable | `claude-fable-5[1m]` only; 5.1 "disabled, update to 2.1.255+" |
| 0.75.1 (CLI 2.1.257) | no Fable | `claude-fable-5-1[1m]`, accepted |

Both fixes are now on `main` and deployed: `managoat_runtimes` 0.3.3 (adapter pin 0.75.1) and 0.3.4 (`prepare_sandbox/3` warms the cache at provisioning), consumed by #2004.

## The production gate ran, and it splits by credential type

Driven on 2026-09-11 against the deployed 0.3.4, twice — same commit, same pods, a fresh cold sandbox each time, the same `"Reply with the single word OK"` prompt, one variable changed:

| credential in `sprite_env` | provisioning | warm-up | turn |
|---|---|---|---|
| `ANTHROPIC_API_KEY` | 15s | warms (silent) | **completed**, `effective_model: claude-fable-5-1[1m]` |
| `CLAUDE_CODE_OAUTH_TOKEN` | 41s | polls out its 30s bound, logs the warning | **failed** — `Invalid value for config option model` |

The passing turn:

```json
"model_selection": {
  "effective_model": "claude-fable-5-1[1m]",
  "requested_model": "claude-fable-5-1",
  "source": "runtime", "status": "selected"
},
"usage": {"input": 2, "output": 4, "cache_write": 6935, "cache_read": 10248}
```

The failing one is the #1669 refusal verbatim, and the warm-up reports its own failure:

```
claude model list did not warm before the first session;
the org's additional models (Fable) will not be selectable in it
```

`/home/sprite/.claude.json` on that sandbox has no `additionalModelOptionsCache` key. Egress is not the cause — `cachedGrowthBookFeatures` comes back from the server on the failing path, so the sandbox reached Anthropic and only the additional-models fetch returned nothing. The brokered placeholder is not implicated.

`Claude.default_env/2` prefers OAuth whenever it is set, so any tenant holding a Claude subscription token takes the failing path.

## What this ships, deliberately

`@catalog` is one global list. `GET /api/catalog` will therefore offer `claude-fable-5-1` to every account while only the API-key path resolves it — including every tenant with no anthropic credential of their own, who fall back to `PLATFORM_ANTHROPIC_API_KEY` and get a working model.

**That is the decision, not an oversight.** The id is live, correct and useful for the platform-key majority, and it is listed for them rather than withheld. A Fable refusal reported by a tenant using their own Claude subscription is this, is expected, and clears when the warm-up populates on that path. The `ModelCatalog` comment says so at the point a reader will look, with the third thing to check — which credential the conversation ran on, named by the `broker` log event.

## What

- `claude-fable-5-1` back at the top of the anthropic suggestions.
- `claude-fable-5-1` leaves `Fountain.RefusedModels`. `claude-fable-5` stays: 0.75.1 refuses it even warm, because the org's additional list carries 5.1 only.
- The two-gates note gains a third gate: the accepted set varies by credential type, not just by adapter version and config-dir state.
- Changelog entry under Unreleased.

## Gate

Rebased onto current `main` (it was 4 days stale and predated the 0.3.4 lock). Local, from `apps/fountain` with the pinned compiler: `mix format --check-formatted` clean, `mix compile --warnings-as-errors` clean, `mix credo --strict` no issues. `model_catalog_test`, `starter_test`, `agents_live_test`, `docs_test`: **71 tests, 0 failures**. `catalog_controller_test` + `acp_model_contract_test`: **9 tests, 0 failures**.

The refused-model guard was checked for vacuity rather than assumed: putting `anthropic/claude-fable-5-1` back into `RefusedModels` while it is suggested fails `model_catalog_test.exs:56` with the intended message. It passes here because the id is no longer in the registry, which is the point.

https://claude.ai/code/session_01PkZYEpwKRZTwhTc8cXMtLP

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -172,6 +172,18 @@ upgrade, is in
   round trips and invalidates checkpoints when changed. The overall provisioning
   deadline and failed-setup handling remain in force.
 
+- `claude-fable-5-1` is suggested for anthropic again, so `GET /api/catalog`
+  lists it. It was removed on 2026-09-07 because the claude adapter refused
+  it; the refusal was not the adapter version but a cold cache. The Claude
+  Code binary learns an org's "additional models" (Fable among them) from a
+  fetch it makes after a session starts and caches for the next launch, so
+  the first session in a fresh sandbox never listed Fable on any adapter
+  version. Two `managoat_runtimes` releases fix that: 0.3.3 moves the adapter
+  pin to 0.75.1 (the bundled CLI must be 2.1.255 or later for Fable 5.1), and
+  0.3.4 warms the cache at provisioning. Verified with a real turn on the new
+  pin. `claude-fable-5` stays unsuggested: the adapter refuses it even with
+  the cache warm.
+
 - Vault secret expiry can be edited in the console or with a metadata-only PATCH, without replacing the encrypted value.
 - Conversation lists accept a `sandbox_id` filter, including through the TypeScript SDK.
 
--- a/apps/fountain/lib/fountain/agents/model_catalog.ex
+++ b/apps/fountain/lib/fountain/agents/model_catalog.ex
@@ -80,15 +80,49 @@ defmodule Fountain.Agents.ModelCatalog do
   # to users. `Managoat.Runtimes.Claude.prepare_sandbox/3` warms that cache
   # at provisioning (managoat_runtimes 0.3.4), and Fable 5.1 additionally
   # needs the CLI at 2.1.255 or later, which the 0.75.1 adapter pin bundles
-  # (managoat_runtimes 0.3.3). Check both when this entry misbehaves.
+  # (managoat_runtimes 0.3.3).
+  #
+  # ## And the accepted set also varies by credential type
+  #
+  # The warm-up only populates on an API key. Measured in production on
+  # 2026-09-11, same commit and pods, a fresh sandbox each time, one variable
+  # changed:
+  #
+  #   * `ANTHROPIC_API_KEY` (a tenant's own, or the platform key a tenant with
+  #     no anthropic credential falls back to): the cache warms in a couple of
+  #     seconds and the turn selects `claude-fable-5-1[1m]`.
+  #   * `CLAUDE_CODE_OAUTH_TOKEN` (a Claude.ai subscription): the fetch never
+  #     populates `additionalModelOptionsCache`, the warm-up polls out its 30s
+  #     bound and logs "claude model list did not warm", and the turn fails at
+  #     `session/set_config_option` with "Invalid value for config option
+  #     model" — the #1669 refusal exactly.
+  #
+  # Egress is not the cause: `cachedGrowthBookFeatures` comes back from the
+  # server on the failing path, so the sandbox reached Anthropic and only the
+  # additional-models fetch came back empty. `Claude.default_env/2` prefers
+  # OAuth whenever it is set, so a tenant holding a subscription token always
+  # takes the failing path.
+  #
+  # **This entry is therefore knowingly suggested to accounts it cannot serve.**
+  # `@catalog` is one global list, so `GET /api/catalog` offers Fable to every
+  # account while only the API-key path resolves it. That is a deliberate
+  # choice, not an oversight: the id is live and correct for the platform-key
+  # majority, and it is listed rather than hidden from them. A Fable refusal
+  # reported by a tenant with their own Claude subscription is this, and is
+  # expected until the warm-up populates on that path.
+  #
+  # Check all three when this entry misbehaves: the provider, the adapter pin,
+  # and which credential the conversation ran on (the `broker` log event names
+  # the keys).
   @catalog %{
     # `claude-fable-5-1`: added 2026-09-06 (#1659) from the published id with
     # no adapter check, removed 2026-09-07 (#1669) after two refused turns,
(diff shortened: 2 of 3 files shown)


BinaryBourbon (approved):

jhgaylor:
## Verified in production on 2026-09-11. The gate does not pass on every credential path.

`managoat_runtimes` 0.3.4 is consumed (#2004) and deployed. I drove the cold-sandbox turn this PR's body asks for, twice, and the result splits by credential type:

| credential in `sprite_env` | provisioning | warm-up | turn |
|---|---|---|---|
| `CLAUDE_CODE_OAUTH_TOKEN` | 41s | polled out its 30s bound, logged the warning | **failed** — `Invalid value for config option model` |
| `ANTHROPIC_API_KEY` (platform) | 15s | silent, i.e. warmed | **completed**, `effective_model: claude-fable-5-1[1m]` |

Same commit, same pods, a fresh sandbox each time, the same `"Reply with the single word OK"` prompt. One variable differs.

### The passing run

```json
"model_selection": {
  "effective_model": "claude-fable-5-1[1m]",
  "requested_model": "claude-fable-5-1",
  "source": "runtime",
  "status": "selected"
},
"usage": {"input": 2, "output": 4, "cache_write": 6935, "cache_read": 10248}
```

So the mechanism this PR depends on does work, and `[1m]` arrives exactly as #1668 expects. On the API-key path this PR is correct.

### The failing run

Identical to the #1669 refusal that removed the id in the first place:

```
Could not select model claude-fable-5-1: Invalid value for config option model:
claude-fable-5-1. No prompt was sent.
```

The warm-up ran — this is not a dispatch or wiring problem. It reported its own failure:

```
00:13:36.698 [warning] claude model list did not warm before the first session;
the org's additional models (Fable) will not be selectable in it
```

Confirmed in the sandbox rather than inferred. `/home/sprite/.claude.json` (38KB, fully written) contains **no `additionalModelOptionsCache` key**:

```
autoUpdates, autoUpdatesProtectedForNative, cachedExperimentData, cachedExperimentFeatures,
cachedGrowthBookFeatures, cachedGrowthBookFeaturesAt, firstStartTime,
hasResetAutoModeOptInForDefaultOffer, installMethod, machineID, migrationVersion,
opusProMigrationComplete, sonnet1m45MigrationComplete, userID
```

Every "fable" string in that file is in `cachedGrowthBookFeatures` — `tengu_usage_overage_included_models: ["Fable","Fable 5","Fable 5.1"]`, and a `fable-5-1-launch` announcement whose `requiresModel` is `claude-fable-5-1`. Those features came back from the server, so **egress through the broker worked**; the additional-models fetch specifically never populated. The brokered placeholder is not the problem.

`Claude.default_env/2` prefers OAuth whenever it is set, so any account holding a Claude subscription token takes the failing path.

### Why this blocks the PR as written

`@catalog` is global. `GET /api/catalog` would suggest `claude-fable-5-1` to every account, and it only resolves for accounts on the API-key path — which includes accounts falling back to `PLATFORM_ANTHROPIC_API_KEY`, and excludes any tenant with `claude_code_oauth_token` set. That is the same "suggested id fails the first turn" shape as #1669, narrowed to a subset of accounts rather than all of them.

The two-gates note above `@catalog` is what this is really about: gate two is the adapter, and the adapter's accepted set turns out to vary by **credential type** as well as by version and config-dir state. Worth saying so in that comment whatever is decided here.

### Options

1. Leave as a draft until the warm-up populates on the OAuth path (upstream fix in `managoat_runtimes`).
2. Land only the `ModelCatalog` / `RefusedModels` comment changes, which are accurate and useful regardless, and keep the id unsuggested.
3. Suggest it knowingly, accepting that OAuth-credential accounts get a failing first turn.

Staying a draft for now.

### Reproducing

Two accounts, no mutation needed: any tenant with `claude_code_oauth_token` set takes the failing path, and any tenant with neither anthropic credential falls back to `PLATFORM_ANTHROPIC_API_KEY` and takes the passing one (`InferenceCredentials.select/2`). The `broker` log event names the keys, so which path a conversation took is visible without decrypting anything. Both test agents and conversations have been deleted.

jhgaylor:
Filed upstream as managoat/managoat_runtimes#20 — the warm-up populates `additionalModelOptionsCache` on `ANTHROPIC_API_KEY` and not on `CLAUDE_CODE_OAUTH_TOKEN`, with the production measurement, what is ruled out (egress, dispatch, adapter pin) and the two candidate causes that are not yet distinguished.

That issue is the thing to watch for the OAuth half of this entry. When it closes, the credential caveat in the `ModelCatalog` comment is what needs revisiting.

PR #1730, opened 2026-09-07
feat: /admin/broker, an operator view of the egress credential broker

jhgaylor:
## What

A new admin tab, **`/admin/broker`**, for the egress credential broker (ADR 0019). Since `BROKER_TENANTS=*` on 2026-09-04 every tenant's sandbox dials it, and its state was only on Grafana, in the per-conversation `/egress` endpoint, or in SQL. This is the glance.

Sections, top to bottom:

- **Health** — listener up on this replica, the tenant ratchet (`all` / count / none), live sessions with the expired-but-unswept count, when the derived CA ends, the log retention window.
- **Traffic** over a window (1h / 24h / 7d, in the URL) — requests split into credential attached / passed through / denied / failed, with how many conversations and tenants produced them; the busiest hosts with their outcome split; the bindings whose credential the proxy attached and the **variable names** it attached (never values).
- **Denied** and **Failed** — the recent rows an operator acts on, each linked to its conversation and owner, with a one-line explanation of what each usually means (`client_closed` is the sandbox hanging up, `upstream_*` is the origin).
- **Live sessions** — who holds a proxy token, its unmatched-host policy, the variables it brokers, minted and expiry; soonest to expire first.

A deployment that does not broker gets a banner saying so and whatever the log still holds.

## Shape

`Fountain.Broker.Native.Insights._unsafe_overview_admin/1` is the one query module: `_unsafe_` because it spans every tenant by design, with the admin page as its only caller. Every aggregate is bounded by the window and reads the `inserted_at` index, so the page is a handful of range scans rather than the unscoped aggregate on a busy log table that took the broker down on 2026-09-04. Prod's log is ~46k rows/day, ~77k total, so this is cheap. Nothing decrypts a rule; `rules_ciphertext` and `token_hash` never leave the query.

## Tests

- `broker_insights_test.exs` (async): zeros on an empty log, the outcome split and host/binding tables, window bounds and the fallback for an unknown window, variable-name merging across rows, live vs expired sessions with the credential names from `meta`, and the cascade on tenant delete.
- `admin_broker_live_test.exs` (async): admin / user / anonymous access, the not-brokering banner, rows rendered with links to the conversation and owner, and the window control.

`mix precommit` green locally: 4,412 core tests + both extension suites, 0 failures; credo no issues; dialyzer passed; release assembled.

## Docs

`docs/configuration.md` names the page beside `BROKER_LOG_RETENTION_HOURS`; `docs/guides/operate/observability.md` points to it under the broker alerts. No ADR change: nothing about the broker's behaviour moves.

https://claude.ai/code/session_01YChmLndKqQbBMrSrxoQpXE


jhgaylor:
Jake asked me to collect occasional feedback while Fountain work continues. This run exposed the accounting/budget problem; the follow-up is tracked in #1731 and #1732, including a host deadline that can stop a blocked tool call.

When you resume this PR, please note any remaining Review Loop friction—especially repeated resolved findings, unnecessary retries, or a fix that invalidates CI. A run link plus one concrete example is enough. No need to retry solely to generate feedback.

BinaryBourbon (approved):
**Approving.** One disclosure first: I wrote the third commit on this branch (`df09694`, the fixes for my own earlier review), so my read of that commit is not independent. The first two commits and everything below I looked at fresh.

### The door

`/admin/broker` is declared inside `live_session :admin`, behind both `:require_authenticated_user` and `:require_admin`, so the page is gated the same way every other admin section is. `Insights` is `_unsafe_` throughout and says why in its moduledoc — the queries span every tenant by design, and the one caller is behind `require_admin`. That is the documented legitimate caller shape.

Nothing here decrypts a rule or renders a credential. `live_sessions/1` selects neither `rules_ciphertext` nor `token_hash`, and nils `meta` once it has pulled the variable *names* out of it. The request table shows `path` only, which the schema keeps query-string-free on purpose. This is the right side of the "never record values" line in 0013.

### The seven findings from the earlier review

All addressed, and each one is pinned by a test that was verified by reverting the fix and watching that test — and only that test — fail:

- Denied is no longer described as host-policy refusals alone. It now names the `403` placeholder/unsafe-credential cases, the `413`, and the `502 credential_missing`, which is the fault this page exists to catch and was previously mislabelled as an agent hitting an allow list.
- `credential_missing` is the one refusal that writes an `error`, so `count(r.error)` had it in Denied *and* Failed at once. Failed is now `error IS NOT NULL AND outcome <> 'denied'` in both the window counts and `top_hosts`, and `where_failed`/`error_counts` match. The three outcome shares partition to 100% again.
- `live_sessions` is newest-minted first instead of soonest-to-expire, which is what makes the 50-row cap show the tokens sandboxes actually hold rather than a conversation's stalest duplicates.
- The post-mount authorization recheck now sends each of the three revocation cases where `Live.Hooks` sends it, instead of collapsing all three onto the login form.
- `recent_query/2` orders on the indexed `inserted_at`, so a window with no refusals stops at the window edge instead of walking the retained log backwards through the primary key.
- `keys_by_service` takes the ten selected binding names.
- `Native.ca_expires_at/0` is public and both the page and `emit_telemetry/0` derive from it, which makes the "one function" comment true rather than aspirational.

One detail I want to call out as correct because it is easy to get wrong: `live_sessions_total` is snapshotted from the *same* read as the list, not from `sessions.live`. `sessions.live` is refreshed by the 30s health tick while the list is not, so comparing those two would have shown the truncation warning as soon as a single new session was minted.

### Not a blocker, but worth an issue

This is now the only admin LiveView that rechecks authorization after mount, and the pages where it matters most do not. `Accounts.update_user_role/3` does not bump `session_version`, so a demoted admin's session stays valid, and `AdminLive.Users`' `toggle_admin` and delete handlers never revisit the actor's current role. A demoted operator with an open tab can still promote and delete accounts until they reload. That is pre-existing on `main` and squarely outside this diff, but this PR is what makes the asymmetry visible, so it should not go unrecorded.

Two smaller notes, neither worth holding the merge:

- The recheck costs one `get_user/1` per open admin tab per 30s tick. Fine at operator scale, but it is a real query on a timer.
- After a health tick the "Live sessions" tile can read higher than the table beneath it. The page states that health and traffic refresh on different clocks, so this is legible rather than wrong.

### Gates

Full CI green on `df09694`: six test partitions, coverage, Elixir static analysis, release boot and the API contract, SDK clients, both Swift jobs, docs prose, secret scan, `CI required`. The earlier `Release boot` failure on this branch was `setup-beam` taking a Gateway Error from `builds.hex.pm` before any code ran, and it passed on the current head first try.

PR #1747, opened 2026-09-08
Persist deadline events and delivery intent atomically

jhgaylor:
Deadline expiration currently changes the turn row without guaranteeing that a blocked actor publishes its result. This commits one failure event, matching tenant webhook jobs, and a retryable local notification with the failed turn. Late completion or interruption reuses that event; retention cannot authorize a contradictory replacement. Notification retries keep the same event id.

Stacked on #1746. **Draft; public bounded execution remains disabled.** Trusted provider identity and the remaining actor/sandbox lifecycle gates still block activation. Delivery is at-least-once, and local event durability does not prove remote termination or stopped billing.

Validation: full precommit passes **4,682 tests and 6 doctests, zero failures**; 74 focused tests and 20 separate-connection PostgreSQL races pass. Tests cover rollback, enqueue failure/recovery, ordinary webhook compatibility, notification retry, competing late publishers, deletion, and tenant isolation. No provider operations occurred. ADR 0046 and `decisions/evidence/deadline-events.json` record the scope and the earlier deliberately interrupted validation.

Tracks #1732. Internal event durability only; `sdk-no-release` applies.

[CI passes at `b339383a7404597b3e3f5fe3b04edecc160c8de0`](https://github.com/BinaryBourbon/fountain/actions/runs/34176205708), including all six test partitions, coverage, static analysis, release boot, API contracts, SDKs and secret checks. Activation gates above still apply.


jhgaylor (commented):

BinaryBourbon (commented):
## Review: atomic outcome + delivery intent is the right shape

Commenting rather than blocking — the mechanism here is sound and my objections to this stack are on #1744, #1745 and #1746, which this sits on.

What I like: recording the terminal stage event, the webhook job and the notification job in the same transaction as the failed turn removes the whole class of "the turn is failed but nobody was told", and using the persisted `log_events.id` as the dedup identity for both streams and callbacks is the right choice — it is the only id that survives a restart. Keeping `deadline_event_id` on `turn_executions` so the identity outlives transcript retention is a nice touch, and the tenant recheck in `_unsafe_record!/2` (`Repo.get_by(Conversation, id:, user_id:)` plus `turn.conversation_id == owner.id`) is the right paranoia for a system writer.

Three things to look at:

- **`TurnDeadlineNotification` is on `queue: :webhooks` with `max_attempts: 20`.** That queue delivers real outbound webhooks to customers; a local `_unsafe_notify_stage/1` fan-out competing for its concurrency, with 20 attempts, can delay actual deliveries during an incident — which is exactly when a lot of deadline events get produced at once. A separate queue (or the default one) and a much smaller attempt count would isolate them.
- **`_unsafe_record!/2` returns `nil` when the ownership recheck fails**, and the caller stores that in `deadline_event_id`. That is correct behaviour, but it is silent: an execution whose conversation changed hands ends with no event, no webhook and no log line saying so. A `Logger.warning` there would make it findable.
- **The `!` name with a conditional body.** `_unsafe_record!/2` raises on the writes but returns `nil` on the ownership miss; `Webhooks.dispatch_stage!/1` and `Oban.insert!/1` inside it will raise inside the caller's transaction and take the turn write with them. Given the rule in this repo that a best-effort recording must never take its mutation down, it is worth a sentence in the moduledoc saying that here the atomicity is deliberate and a failure to record *should* roll the turn back.

Needs a rebase along with the rest of the stack — see #1745 for why.

jhgaylor:
Force-pushed `5b1a43ee`. Rebased onto the fixed #1746 and all three notes addressed.

- **Off the customer queue.** `queue: :webhooks, max_attempts: 20` → `queue: :maintenance, max_attempts: 3`. The webhook job is already committed beside the event, so this worker only notifies local subscribers; leaving it on `:webhooks` meant a deadline storm competed with outbound delivery exactly when that matters. Three attempts because the event is durable — a subscriber that missed the first three reads the outcome from the transcript on reconnect.
- **The two silent paths now say what happened.** `_unsafe_record!/2` returning `nil` on a failed tenant recheck, and the worker skipping a deleted transcript. Both correct, both previously leaving no trace that a persisted deadline outcome reached nobody. `Logger.warning` rather than `info`, since `config/test.exs` sets the level to `:warning` and a regression can then assert on it.
- **The `!` naming is now explained rather than changed.** `DeadlineEvents`' moduledoc says why the atomicity here is the exception to ADR 0013 and not a violation: elsewhere a recording is best-effort so a failed insert cannot take its mutation down; here it is the reverse, because a durably failed turn that nothing was told is the bug this module exists to prevent. The three `!` calls raise into the caller's transaction and *should* roll the turn write back.

The `_unsafe_check_saved_execution_allowance` conflict from the re-cut resolved cleanly — the rebase only needed the `Labels`/`ExecutionAllowance` aliases and a second `Mimic.copy` line.

**Verification.** Three new regressions. Reverting the queue and both log lines fails all three. Full core suite: **6 doctests, 5,002 tests, 0 failures, 2 skipped**. Format, `--warnings-as-errors`, `credo --strict` (7,303 mods/funs, no issues) and `okf validate decisions` clean. ADR 0046's deadline-supervisor row now names what this PR delivers.

On to #1748.

BinaryBourbon (commented):
## Review: right direction on the queue, wrong destination

Re-reviewed at `b01c634`. The three notes are all answered, and the moduledoc paragraph on why the atomicity here is the exception to ADR 0013 rather than a violation of it is exactly what I was after — it says the quiet part (these `!` calls *should* roll the turn write back) instead of leaving the next reader to infer it. The two `Logger.warning` lines are the right level, and the tests assert on them because `config/test.exs` sits at `:warning`.

**But `:maintenance` is `concurrency: 1`.** From `config/config.exs`:

```elixir
queues: [maintenance: 1, credits: 5, exports: 1, mailer: 5, schedules: 5, webhooks: 10],
```

and eight workers already live on it: `RetentionPruner`, `SandboxReaper` (hourly, a paginated provider list plus deletes), `AutonomousTurnReaper`, `ClaimablePrincipalSweep`, `BrokerReaper`, `SecretExpirySweeper`, `TeamScheduler`, `UnverifiedAccountPruner`.

So a user-visible notification now queues single-file behind whichever sweep is mid-run, and a deadline storm — the scenario the move was made for — serializes one job at a time. That trades starving customer webhooks for starving on a sweep, and the config file itself already makes the argument against it: `exports` and `webhooks` are separate queues precisely "so a tenant's slow receiver never sits in front of a maintenance sweep or an email".

The ask is one line rather than a redesign: give it its own queue (`notifications: 5`, or whatever name fits) and point `TurnDeadlineNotification` at it. `max_attempts: 3` is right either way.

Commenting rather than blocking, because the consequence is latency and not correctness — the event is durable and a reconnecting client reads the outcome off the transcript, which is the same reason three attempts is enough. But a notification whose whole job is to be prompt should not be behind the retention pruner.

The rest looks right to me. `assert notification.queue == ... and delivery.queue == "webhooks"` in the same test is a good shape: it pins the separation, not just the move.

jhgaylor:
Pushed `cc4e1df9` — you're right, and the config file made the argument for me.

`notifications: 5`, its own queue, worker pointed at it. `maintenance` being `concurrency: 1` behind eight sweeps is worse than what I moved it off: a deadline storm would have serialized one job at a time behind whichever sweep was mid-run, which is the scenario the move existed to fix. And the comment block above `queues:` already reasons this way twice — `exports` so a data export isn't stuck behind a sweep, `webhooks` so a slow receiver isn't in front of a sweep or an email. Work whose whole value is being prompt gets the same treatment, and the comment now says so in those terms.

`max_attempts: 3` unchanged. The regression needed exactly one line, because it was already asserting the separation rather than the destination — `notification.queue` *and* `delivery.queue == "webhooks"` in the same test. That shape paid for itself two reviews later.

Full core suite: **6 doctests, 5,002 tests, 0 failures, 2 skipped**.

BinaryBourbon (approved):
## Review: its own queue, with the reason written where the next person will look

Re-reviewed at `cc4e1df`. This is the change I asked for, and the part I did not ask for is the better half of it: the `config.exs` comment now makes the argument in the same place the file already makes it twice for `exports` and `webhooks`, so `notifications: 5` reads as a continuation of a policy rather than a one-off. The worker's own comment says what it is *not* on and why for both rejected queues, which is what stops the next move putting it back.

I checked the things a new Oban queue can quietly need, and it needs none of them:

- `OpsGauges.known_queues/0` derives the list from `Application.get_env(:fountain, Oban)[:queues]`, so `fountain_oban_queue_depth` picks up `notifications` with no edit. `ops_gauges_test.exs`'s `~w(maintenance credits exports)` is a sample, not an enumeration.
- Nothing in `runtime.exs` overrides `:queues`, and `config/test.exs` is `testing: :manual`, so the queue name is all the tests need.
- No migration: the queue is a string on `oban_jobs`. During a rolling deploy an old replica simply does not run it, and the new ones do.

`max_attempts: 3` unchanged, and the regression still pins the separation — the notification's queue *and* `delivery.queue == "webhooks"` in the same assertion — rather than the destination, so this cost one line as it should have.

Approving. That clears everything I raised on this PR across both rounds.

PR #1748, opened 2026-09-08
Prepare bounded command transport and guarded ACP writes

jhgaylor:
A bounded turn needs its provider command tracked even when the conversation actor blocks or the turn succeeds. This adds a supervised transport that binds trusted command identity, checks durable authority before each ACP write, and retains cleanup obligations for every outcome. Unknown spawns and uncertain writes are never replayed. Successful turns require confirmed cleanup before a fresh connection can proceed.

Stacked on #1747. **Draft; public enforcement remains disabled.** The conversation actor does not select this transport yet. Activation still requires tracked adapter setup, complete actor/cancellation/recovery wiring, released Sprites identity support, and live acceptance. Local shutdown and a completed reply do not prove remote termination or stopped billing.

Validation: full precommit **4,699 tests and 6 doctests, zero failures**; 66 focused tests; 20 independent PostgreSQL races. A real ACP peer runs through the guarded writer against a simulated provider, then refuses a post-completion prompt. Tests cover identity forgery/conflicts, delayed spawn results, blocked writes, owner death, durable close acknowledgment, orphan task timeouts, terminal outcome arbitration and redacted crash state. No provider operations occurred. ADR 0046 and `decisions/evidence/execution-transport.json` record scope and evidence.

Tracks #1732. Internal preparation only; `sdk-no-release` applies.

[All 24 CI checks pass or intentionally skip at `9b1fb2ea8a26e9d32823d7c0e5c0a8ed5f2b4a83`](https://github.com/BinaryBourbon/fountain/actions/runs/34179844997).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -137,6 +137,11 @@ upgrade, is in
   immediate error. Starts carrying images or naming a `sandbox_id` never
   queue.
 
+- Prepare a supervised bounded-command transport that binds provider identity
+  before stdin, rechecks journal authority for each write, and retains uncertain
+  operations. Successful bounded replies now require connection retirement before
+  a fresh successor. Actor/spawn integration and public activation remain gated.
+
 - Commit deadline failure events and delivery jobs with the failed turn. Late
   completion and interruption reuse the original event, and notification retries
   retain its id. Public bounded execution remains disabled.
--- a/apps/fountain/lib/fountain/application.ex
+++ b/apps/fountain/lib/fountain/application.ex
@@ -57,6 +57,7 @@ defmodule Fountain.Application do
         # (#1040). Supervised and unlinked, a crash here is a log line.
         {Task.Supervisor, name: Fountain.TaskSupervisor},
         Fountain.PlatformChatGPT.Refresher,
+        {DynamicSupervisor, name: Fountain.ExecutionTransportSupervisor, strategy: :one_for_one},
         FountainWeb.Plugs.RateLimit.Sweeper,
         Fountain.Conversations.Redaction,
         Fountain.FeatureFlags.Cache,
--- a/apps/fountain/lib/fountain/conversations/execution_guard.ex
+++ b/apps/fountain/lib/fountain/conversations/execution_guard.ex
@@ -58,17 +58,9 @@ defmodule Fountain.Conversations.ExecutionGuard do
           if sandbox.user_id != conv.user_id, do: Repo.rollback(:ownership_changed)
           if sandbox.status != "ready", do: Repo.rollback(:sandbox_not_ready)
 
-          prior = prior_connection(connection_id)
-          if prior && prior.state != "completed", do: Repo.rollback(:connection_retired)
-
-          if prior &&
-               (prior.conversation_id != conv.id || prior.sandbox_id != sandbox.id ||
-                  prior.sandbox_name != sandbox.sprite_name || prior.provider != sandbox.provider ||
-                  prior.user_id != conv.user_id),
-             do: Repo.rollback(:connection_owned_elsewhere)
-
-          if prior && is_nil(prior.provider_session_id),
-            do: Repo.rollback(:connection_unidentified)
+          # Bounded turns never inherit a warm process: it may retain background
+          # work or SDK allowances from its previous prompt.
+          if prior_connection(connection_id), do: Repo.rollback(:connection_retired)
 
           limits = resolve_turn_limits(conv)
           enforce_deadline_ceiling!(turn, deadline_at, limits)
@@ -82,7 +74,6 @@ defmodule Fountain.Conversations.ExecutionGuard do
             sandbox_name: sandbox.sprite_name,
             provider: sandbox.provider,
             connection_id: connection_id,
-            provider_session_id: prior && prior.provider_session_id,
             deadline_at: deadline_at
           }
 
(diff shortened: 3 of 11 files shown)


jhgaylor (commented):

BinaryBourbon (commented):
## Review: spawn-intent-before-I/O is right; the provider story needs stating

Commenting. The transport is the piece that makes the journal mean something, and the ordering is correct: `_unsafe_claim_spawn/2` persists intent before `DynamicSupervisor.start_child`, stdin stays closed until identity binds, each write rechecks the journal, and a failed write retires the whole execution instead of retrying it. `restart: :temporary` is the right supervision choice for something that must never be replayed. Buffer caps (`@buffer_bytes`, `@buffer_frames`) and the `:deadline` / `:drain_end` pair are good defensive detail.

Points:

- **It is Sprites-only, and that is not visible from the outside.** `_unsafe_start/5` matches `%TurnExecution{provider: "sprites"}` and otherwise returns `:provider_not_supported`. Combined with #1749's `if sandbox.provider != "sprites", do: Repo.rollback(:provider_not_supported)` and #1746's four-entry `@providers` map, the stack reads as multi-provider but is not. E2B, Daytona and self-hosted runners are a real part of this product (ADR 0018, ADR 0022) — the moduledoc and ADR 0046 should say plainly that bounded turns are Sprites-only for now and what the others get instead (today: silently no bound, which is the outcome that worries me most, because a caller that asks for a limit and is refused is fine, and one that asks and is ignored is not).
- **`remaining` is computed once in `init/1`** from `execution.deadline_at`. That is correct while the process lives, and the journal plus #1746's coordinator cover the restart. Worth one line saying so, because the timer looks like the enforcement mechanism at a glance and it is not.
- **`call/3` rescues `:exit` into `{:error, :transport_unavailable}`.** Fine, but that swallows a genuine timeout as well as a dead process, and the two mean different things to the journal. If a 30 s `GenServer.call` times out while the transport is alive and mid-write, the caller sees `:transport_unavailable` and the execution may still be writing. Distinguishing them would make the failure mode readable.

Blocked behind #1744/#1745/#1746 and needs the same rebase.

jhgaylor:
Force-pushed `9e429567`. Rebased onto the fixed #1747, all three notes addressed, plus one thing the rebase surfaced that I think matters more than any of them.

**Sprites-only is now stated rather than implied.** The moduledoc and ADR 0046 both say it, with the reason: binding a provider-issued session id from trusted control metadata is a per-adapter capability and only the Sprites adapter has it, so without one there is nothing to terminate by name. A regression asserts `e2b`, `daytona` and `runner` are refused **without** recording spawn intent — the turn stays retryable instead of landing in the `awaiting_identity` fence. The half I cared about holds: those providers get a refusal at admission, not a silently unbounded turn.

**A timeout and a dead transport are different answers.** `call/3` collapsed both into `:transport_unavailable`, which reads as "nothing happened" and is only true of the second — a transport that did not answer in 30 s may be mid-write. Timeouts now return `:transport_timeout`. `write/2` gained an optional timeout so the regression costs 50 ms instead of 30 s.

**The `init/1` timer** is documented as a local convenience rather than the mechanism; the guarantee is `deadline_at` on the row, which the coordinator acts on whether or not the process survived.

---

**The thing worth your attention: this PR changes completion semantics, and I only noticed because a #1744 regression failed on it.**

`complete/3` now routes a successful turn through `stop/4`, so a completed bounded turn lands in `ready` rather than `completed`, and `prior_connection/1` refuses *all* connection reuse. The comment gives the reason and I agree with it — a runtime that answered can still hold background work, which is exactly the shape behind the phantom follow-up turns.

But the consequence is that **the termination fence now reaches the happy path**. Every successful bounded turn owes a provider cleanup, so a lost or failed `terminate_session` on a turn that worked perfectly fences its conversation and its shared home — the same dead-end I flagged on #1744, now on the common path rather than the error path. The ageing exit added there is load-bearing because of this PR, not despite it. I have added a regression for that corollary (a successful turn whose cleanup is lost ages out, with its own `completed` outcome left untouched) and ADR 0046 now records both the change and why.

That is worth a second look from you, because it is the sort of change that reads as a small refactor in the diff — three call sites moved to `stop/4` — and is actually a shift in what the journal promises.

**Verification.** Two new regressions here plus one updated from #1744. Reverting the provider guard and the timeout clause fails three tests. Full core suite: **6 doctests, 5,022 tests, 0 failures, 2 skipped**. Format, `--warnings-as-errors`, `credo --strict` (7,355 mods/funs, no issues), `okf validate decisions` clean.

On to #1749 — the one I requested changes on.

BinaryBourbon (approved):
## Review: the provider story is stated, and stated in the right place

Re-reviewed at `6d51262`. All three notes answered.

**Sprites-only.** The moduledoc and ADR 0046 now say it plainly, and the regression is better than the sentence: for `e2b`, `daytona` and `runner` it asserts `spawn_submitted_at == nil` and `state == "active"` after the refusal. That is the half I actually cared about — a refused provider leaves a retryable turn rather than an `awaiting_identity` row somebody has to age out. "A caller that asks for a limit and is told no is fine; one that asks and is ignored is not" reads back as the rule it is.

**Timeout versus dead transport.** `:exit, {:timeout, _}` ahead of `:exit, _` is the right order, and the test distinguishes them with a live-but-silent process and a genuinely dead one instead of asserting the clause exists.

One thing worth a sentence somewhere, because it will otherwise be assumed: nothing acts on the difference yet. At the stack tip, `:transport_timeout` appears only in `execution_transport.ex` and its test — `TurnMachine.command_writer/2` hands both to the ACP writer identically, and `Connection._unsafe_spawn_bounded/5` closes on `{:error, _}` either way. Today this buys a truthful return value and a readable log line, which is enough; a line in the moduledoc saying "no caller branches on this yet" stops the next reader concluding the journal already treats them differently.

**The `:deadline` timer as convenience rather than mechanism** is now said out loud, which was the point.

**The semantic conflict is the interesting part of this re-cut.** Every bounded connection owing cleanup — a successful reply included — so a completed turn lands in `ready` and `prior_connection/1` refuses all reuse is right, and it is right for the reason given (a runtime that answered can still hold background work; that is the phantom-follow-up shape). Updating the #1744 regression instead of weakening the state machine is the correct direction, and the companion test — a turn that answered correctly whose cleanup is then lost ages out with its own `completed` outcome untouched — is the one that makes the ageing exit load-bearing rather than a corner case. ADR 0046 records it.

Approving.

Issue #1431, opened 2026-09-03
27 operations declare an error schema for 422 that CastAndValidate does not render


jhgaylor:
Found by the schema guard (#1427).

## The mismatch

`OpenApiSpex.Plug.CastAndValidate` rejects a malformed request itself, before
the action runs, and renders its own body:

```json
{"errors":[{"message":"Invalid integer. Got: string","title":"Invalid value","source":{"pointer":"/limit"}}]}
```

An array of objects. The operations that install that plug declare either
`Schemas.Error` (`{"error": "..."}`) or `Schemas.ChangesetError`
(`{"errors": {"field": ["message"]}}`) for 422. So the document describes the
changeset's 422 and the wire sometimes carries the validator's, and a client
that branches on the shape gets whichever one the failure happened to be.

27 operations are affected. The full list is in
`FountainWeb.SchemaGuardAllowlist` under `:cast_and_validate_422`.

## Why it matters

`sdk/typescript`'s `ValidationError.fieldErrors` reads `body.errors` as an
object of field to messages, which is the `ChangesetError` shape. Against a
`CastAndValidate` 422 it silently returns `{}` — the caller is told the request
was invalid and not which field, on exactly the failures where the field is the
only useful information.

## Options

- **Teach `CastAndValidate` to render `ChangesetError`.** It takes a
  `:render_error` module, so one module implementing the existing shape makes
  the wire match the document everywhere at once. Best fix if the shapes can be
  reconciled: `{"errors": {"limit": ["Invalid integer. Got: string"]}}` is
  expressible from what the validator has.
- **Declare a second 422 schema** and make the document honest about there
  being two shapes. Cheap, but it pushes the branch onto every client.

The first is one change; the second is four SDKs and their conformance
scenarios.

## Where it is recorded now

`FountainWeb.SchemaGuardAllowlist`, as `:cast_and_validate_422`, one entry per
operation so a 28th cannot join silently.

Issue #413, opened 2026-08-03
ops: a dropped sprite WebSocket is ignored — the conversation wedges busy and the sprite bills for 24 hours


jhgaylor:
When the WebSocket between Fountain and sprites.dev drops mid-turn, `Sprites.Command` sends `{:error, %{ref: ...}, reason}` to its owner and stops. The handler discards it:

`apps/fountain/lib/fountain/conversations/conversation_server.ex:1040-1043`

```elixir
def handle_info({:error, _ref, reason}, state) do
  Logger.error("sprite command error: #{inspect(reason)}")
  {:noreply, state}
end
```

`current_command`, `current_command_ref` and `current_turn` all stay set. The command process is started with `GenServer.start`, not `start_link` (`deps/sprites/lib/sprites/command.ex:50`), and `ConversationServer` neither links nor monitors it — the only `Process.monitor` in the file is the provision watchdog watching itself. So nothing else ever notices.

## Consequences

- The turn row stays `status: "running"` with `ended_at: nil`, permanently.
- The conversation stays `running`.
- `handle_call({:send_prompt, ...})` returns `{:error, :busy}` (`:842-843`) because `state.current_command` is still set, so every further prompt gets a 400 `conversation_busy` — forever.
- No `turn/done` or `turn/failed` is ever published, so streaming clients simply stop receiving anything.

Both recovery mechanisms are disabled by the same stale state: `Lifecycle.check/4` receives `state.current_command != nil` as `busy?` (`:1059`), which suppresses idle reclaim, and `SandboxReaper.expire_abandoned_sandboxes/0` explicitly rejects sandboxes whose server is alive (`sandbox_reaper.ex:167`).

The sprite therefore bills until the 24-hour `max_lifetime` bound. `POST /interrupt` does recover it — that handler tolerates a dead `cmd_pid` — but nothing surfaces to the user that this is what they need to do.

## Fix direction

Monitor the command process, or match the ref in the `{:error, ...}` clause and treat it the same way as the `{:exit, ...}` path: clear `current_command`, mark the turn failed, publish `turn`/`failed`, return the conversation to `idle`.

A regression test can send `{:error, ref, :closed}` to a server with a live `current_command` and assert the turn row and the next `send_prompt` both recover.

🤖 Audited and filed by Claude Code

jhgaylor:
Part of the 2026-08-03 re-audit — tracker: #416

PR #1749, opened 2026-09-08
Integrate bounded turn admission and actor lifecycle

jhgaylor:
Bounded turns now commit their journal with admission, run adapter setup inside the tracked command, and retire the connection after every outcome. Cancellation persists before contacting a blocked actor. Restarted actors retire old execution before reattachment; deleting a parent retains cleanup against the original sandbox binding. A process exit without an ACP reply cannot imply success.

Stacked on #1748. **Draft; public enforcement remains disabled.** Released Sprites identity support, atomic late-write fencing, the remaining sandbox lifecycle audit and live acceptance still block activation. Initial provisioning remains separate, and local cleanup does not prove remote termination or stopped billing.

Validation: full precommit **4,716 tests and 6 doctests, zero failures**; 17 lifecycle regressions drive real actors/ACP peers with simulated providers. Sixty new independent PostgreSQL races cover admission, cancellation and deletion, alongside 20 existing deadline races. No provider operations occurred in those proofs. The first full gate caught the server size limit: extracting `TurnLaunch` reduced its pin from 2,767 to 2,706. ADR 0046 and `decisions/evidence/bounded-lifecycle.json` retain evidence and prior corrected failures.

Consumes published `managoat_runtimes` 0.4.2. Tracks #1732. Server integration only; `sdk-no-release` applies.

[CI passed](https://github.com/BinaryBourbon/fountain/actions/runs/34182962640) at `42e40c5954392580b196889d2e5a1d32175c752b`: all 24 checks passed or intentionally skipped.


BinaryBourbon (changes requested):
## Review: this rewrites `conversation_server.ex` on a base that `main` has moved out from under

Requesting changes — on merge mechanics, not on the design, which I think is largely right.

### The rebase is not routine here

This branch forked at `59c5ebc8`. Since then `main` has landed **five commits in `conversation_server.ex`** and **eighteen in `conversations.ex`** — the two files this PR reworks most:

```
086143a6 fix(conversations): refuse a turn whose agent no longer carries a command (#1634)
ee6d551e fix(skills): mount skills on a runtime the library does not know (#1634)
e78a417e feat: execution-limit admission (#1792)
317b00d5 fix: refuse new turns from stale sandbox actors (#1764)
0cea1e7e feat: run the codex runtime on the platform's ChatGPT account (ADR 0047)
```

`e78a417e`, `1fe98654`, `cc455bfc`, `f1d02d2a` and `7a7177bb` in particular land allowance persistence and the fleet-reservation foreign-key fix inside the same launch path this PR restructures — including the rule that nothing under `Quotas.with_sandbox_reservation/3`'s process-wide advisory lock may wait on another transaction. Rebasing 537 changed lines of `conversation_server.ex` across that is where this PR's bugs will come from, and a clean textual merge will not tell you whether the lock ordering still holds. I would want the rebase done, and `reserve_initial_conversation/4` re-read afterwards, before this is reviewable on its merits.

Please also check the size pin after the rebase: the file goes 2,767 → 2,706 on this branch, which is the right direction, but `main` is at 2,774 and the pin only ratchets down.

### On the design

Extracting `turn_launch.ex` is the right call and the admission gate reads well. Specific notes:

- **`_unsafe_admission_gate/1` and `_unsafe_autonomous_turn/2` fence on `open_execution?/1`, which is not limited to bounded turns.** Once a conversation has any unresolved journal row, every new prompt, every wake and every autonomous turn is refused with `:execution_fenced`. That is correct while limits are configured, and today `limits` is `%{}` so no row is ever written — but it means the dead-end I described on #1744 is not confined to the bounded path once anyone sets a ceiling. The stuck-state recovery has to exist before this gate does.
- **`_unsafe_autonomous_turn/2` refuses whenever any limit resolves** (`{:ok, limits} when map_size(limits) == 0 -> :ok`, everything else rolls back `:execution_fenced`). So configuring an account ceiling silently turns off scheduled/autonomous turns for that account. That is a defensible choice, but it is a product decision and I could not find it written down in ADR 0046 or the changelog entry. Teammate schedules are a shipped feature; this would stop them.
- **`{:execution_limits_invalid, "wall_time_seconds_required"}`** when limits are present without a wall clock is the right refusal. Worth confirming the API surface turns it into a 422 with the one-shape error body (#1431) rather than a 500.
- `_unsafe_interrupt/1` picking `order_by: [desc: e.inserted_at], limit: 1` under `FOR UPDATE` is fine given the partial unique index on open rows per conversation, but the index is what makes it safe — a comment pointing at `turn_executions_open_conversation_index` would save the next reader the trip.

jhgaylor:
Force-pushed `d3c1af1f`. I did the rebase first, as asked, and it turned up three protections this branch predated and would have reverted — so this is a re-cut: main's code is the base, and the journal integration is layered on it.

### What the rebase found

**1. `_unsafe_admit_turn/4` re-implemented admission and lost two checks.** It replaced `Conversations._unsafe_create_turn_on_sandbox/3` with `conv.sandbox_id != sandbox_id`, which drops the `attached?` check — the one proving the conversation is still on a **non-terminal** sandbox owned by the same tenant (#1761, #1764) — and the `_unsafe_check_saved_execution_allowance/1` recheck under the row locks. It also predated the `FOR UPDATE`/foreign-key ordering from #1790/#1793.

Now `ExecutionGuard._unsafe_register_bounded/3`, a step *inside* that transaction rather than a replacement for it. A turn refused for an unavailable sandbox still answers `:sandbox_unavailable`, before the journal is consulted at all — one of the lifecycle tests asserted `:sandbox_not_ready` there, and that expectation moving back is the check working.

**2. `turn_launch.ex` had zero references to `CodexChatGPT.peer_auth/2`.** Main's launch calls it twice. Taking this branch's copy would have silently reverted ADR 0047's codex-on-ChatGPT authentication, which is live in production. The extraction is still the right idea — `conversation_server.ex`'s pin only ratchets down and the bounded lifecycle needs room — so `TurnLaunch` is now main's `run_fresh_turn/7` verbatim plus this branch's bounded spawn branch and `bounded_sdk_limits/2`, with `fail_turn_before_start/6` passed in because it writes through the actor's logger and owns `current_turn`. Pin 2774 → 2739.

**3. `_unsafe_autonomous_turn/2` refused all background work under any allowance, and dropped #1764's sandbox binding** by calling `_unsafe_create_turn/1` instead of `_unsafe_create_turn_on_sandbox/3`. Both gone. Autonomous turns now go through the same admission, so they get a journal and a deadline the coordinator expires like any other. Refusing them would have turned one account ceiling into "no schedules and no background follow-ups for this account" — a product decision affecting a shipped feature that nothing had written down. ADR 0046 now records the choice; if you want the refusal instead, that is a reasonable position but it needs saying out loud.

### What I kept, because it is the point of the PR

The per-message `_unsafe_authorize_write/3` gate in `handle_info`, the bounded transport spawn, retirement, and the error propagation out of `kick_turn/4`. That last one I first reverted to main's shape and a lifecycle test caught it immediately — a fenced second prompt was answered `:ok`. It is now merged rather than chosen: `kick_turn/4` returns `{:error, reason, state}` so the refusal reaches the caller **and** main's `drop_connection/2` still happens.

**Two of main's own tests changed as a result.** `conversation_server_acp_test.exs`'s "a stale actor cannot start prompt work after retire/move" asserted `send_prompt` replying `:ok`. It now asserts `{:error, :sandbox_unavailable}`. Those tests' real subject — no turn, no peer, no timer, no caller calls — is untouched; the old `:ok` told a client its prompt was accepted when no turn would ever run. Flagging it because changing a main test inside a feature PR deserves a second opinion.

Also ported `bounded_lifecycle_test.exs` off `conversations.execution_limits` onto `execution_allowances`, matching the #1745 re-cut.

**Verification.** 18 lifecycle tests pass, including the rewritten autonomous one. Full core suite: **6 doctests, 5,039 tests, 0 failures, 2 skipped**. Format, `--warnings-as-errors`, `credo --strict` (7,396 mods/funs, no issues — it wanted an `# ownership:` line on the new `_unsafe_open_execution?/1` call, which it has), `okf validate decisions` valid.

On to #1750.

BinaryBourbon (changes requested):
## Review: the rebase is done and done properly; the per-message gate is what I would change now

Re-reviewed at `1d4be91`. The rebase was the blocking ask and it was not treated as a merge exercise, which is why it found things.

**Verified rather than taken on trust.** `086143a6`, `ee6d551e`, `e78a417e`, `317b00d5`, `0cea1e7e`, `1fe98654`, `cc455bfc`, `f1d02d2a` and `7a7177bb` are all ancestors of this branch. `_unsafe_register_bounded/3` is a step *inside* `Conversations._unsafe_create_turn_on_sandbox/3` rather than a second admission beside it, so the advisory lock, the `FOR UPDATE` on the parent, the `attached?` non-terminal-sandbox check (#1761, #1764) and `_unsafe_check_saved_execution_allowance/1` all still run, in that order, and a turn refused for an unavailable sandbox still answers `:sandbox_unavailable` before the journal is consulted. `turn_launch.ex` carries `CodexChatGPT.peer_auth/2` (ADR 0047 intact). Autonomous turns go through the one admission path. That is the shape I wanted and the three near-reverts were worth the trip.

The `TurnLaunch` extraction reads as main's body plus the bounded branch, and passing `fail_before_start` in rather than duplicating it is the right seam.

### Requesting changes on one thing: `handle_info/2` now costs a locking write transaction per message

```elixir
def handle_info(message, state) do
  if execution = Map.get(state, :turn_execution) do
    case ExecutionGuard._unsafe_authorize_write(execution.id, execution.connection_id) do
```

`_unsafe_authorize_write/3` goes through `with_execution/2`, which is a transaction that does `Repo.get`, `lock_parent` (`FOR UPDATE` on the conversation), the journal row `FOR UPDATE`, `lock_turn`, then `Repo.get(Turn, ...)` and `current_binding?/1`'s two-table join. Six queries and three row locks.

That is not a per-prompt gate. It runs for every message the actor receives while a bounded turn is open: every `{:stdout, %{ref: ref}, data}` chunk, every `{:acp, ref, payload}` report, `:lifecycle_check`, every `:DOWN`. A chatty ACP turn is hundreds to thousands of those. And because it takes the parent `FOR UPDATE`, it serializes against everything else that locks the parent — admission, `_unsafe_release_parent/3`, `reset_sandbox/2`, and the coordinator's own `_unsafe_expire/2` and `_unsafe_claim_termination/1`. The hotter the turn, the more the deadline coordinator queues behind the very turn it is supposed to expire.

The outbound direction is already covered by #1748: `ExecutionTransport` calls the same `_unsafe_authorize_write/2` before each write, and writes are few. What this adds is authorization of the *inbound* stream, which cannot itself reach the provider.

What I would like instead: keep the authoritative locked check where a provider write or a terminal outcome actually happens — the transport, and `_unsafe_complete/3` — and make the actor-side gate cheap. An unlocked read of `state`, `connection_id` and `deadline_at` off the journal row would answer "is this still mine, is it past its deadline" for a fraction of the cost, or gate only the messages that can produce a write or an ending. It is the same argument I made on #1746's per-second poll, one layer in: inert today because `enforced_controls/1` is `[]`, but this is the shape that ships.

### Smaller

- `retire_bounded_turn/1` hard-matches `{:ok, decision} = ExecutionGuard._unsafe_complete(execution.id, "interrupted")`. The only `{:error, _}` reachable is `:not_found`, and the journal deliberately carries no FK to its parent (`20260907214000_add_turn_execution_guards.exs` says why), so the row can outlive a deleted conversation — which is exactly the moment an actor is still draining its mailbox. A `MatchError` there takes the actor down. A clause that logs and clears `turn_execution` costs less than a restart.
- `run_turn/6` skips `TurnMachine.generate_title/4` for a bounded turn and nothing says why. A conversation whose first turn is bounded silently never gets a title. If the reason is "do not spend inference against an allowance", a comment settles it; if it is not deliberate, it is a regression for anyone who turns limits on.
- `TurnLaunch.bounded_sdk_limits/2` hard-matches `{:ok, limits}` from `Managoat.Runtimes.ACP.execution_limits/2`. Safe, because `Connection._unsafe_spawn_bounded/5` called the same function through a `with` a few frames earlier — but the safety is at a distance and deserves the sentence.
- Still open from last time, trivial: `_unsafe_interrupt/1`'s `order_by: [desc: :inserted_at], limit: 1` under `FOR UPDATE` has no comment pointing at `turn_executions_open_conversation_index`, which is what makes it safe.
- The pin will want another look on the next rebase. `7bcdb52e` (#1886) and `8a8f9a76` (#1853) have taken `conversation_server.ex` on `main` to 2,567 lines while `main`'s `@pin` still reads 2,774. Rebasing this stack on top lands the file well under the 2,733 the stack ends at, so the pin should come down again rather than be inherited — the number is a record of where the file is, not a budget.

Checks are green on this head.

jhgaylor:
Pushed `db842e58`. The per-message gate is fixed, and the four smaller notes with it.

**`handle_info/2` is one unlocked read.** `ExecutionGuard._unsafe_actor_gate/3` reads three columns — state, connection, deadline — and answers only "is this still mine, still inside its deadline". The inbound stream cannot reach the provider by itself, so it never needed write authorization; the locked check stays where a write or a terminal outcome happens, which is the transport (#1748) and `_unsafe_complete/3`. `:retire` is not the durable decision either — retirement takes the locks a frame later and `_unsafe_complete/3` arbitrates completion against expiry there, so a missing row, a superseded connection, a closed state and a passed deadline all converge on the same authoritative write.

The regression **measures** the property rather than asserting it: one query, matching `^SELECT`, no `FOR UPDATE`, no `FOR SHARE`, no `BEGIN`. Collected from `[:fountain, :repo, :query]` telemetry and asserted in the test body, because a raising handler is detached and fails nothing (#1427). My first attempt tried to prove lock-freedom by holding `FOR UPDATE` in another process — which cannot work under the SQL Sandbox's single shared connection, so it was measuring the sandbox, not the gate.

**`retire_bounded_turn/1` no longer takes the actor down.** Your point about the missing FK was the whole thing. It now fails the local turn — the journal was what would have ended it, so with the row gone the actor has to — clears `turn_execution` and closes.

Worth saying how that one went, because it is the more useful half: **my first version of that test passed against the unfixed code.** It asserted the gate returned `:retire` and that `_unsafe_complete/3` returned `{:error, :not_found}`, and never drove the actor at all. Reverting the fix broke nothing, which is how I found it. Rewritten to start a real bounded server and `send(pid, :lifecycle_check)`, it reproduces the `MatchError` — and then failed a second time on `wait_idle`, because clearing `turn_execution` without ending `current_turn` left the actor believing a turn was still running forever. That second failure was a real defect the first version would have shipped.

**Both halves moved into `Fountain.Conversations.BoundedTurn`** — journal logic the actor calls rather than actor logic, with `finish_acp_turn/4` passed in for the same reason `TurnLaunch` takes `fail_before_start`. That is also where the 45 lines came from.

**The pin comes down rather than being inherited: 2739 → 2729**, which is where the file actually is at the tip. Taking your framing that the number is a record and not a budget.

**The three comments.** `bounded_sdk_limits/2`'s hard match is safe because admission already called the same function through a `with` that rolls the turn back, so a journal cannot exist for a runtime that refuses its own limits — crashing there is the right answer to an admission bug. `_unsafe_interrupt/1`'s `limit: 1` is exact because `turn_executions_open_conversation_index` is unique and partial to the open states; the `order_by` only matters if that invariant were dropped. And the title: titling is a second inference call the journal neither bounds nor prices, so spending it under a wall-clock ceiling would be usage the caller asked to cap and cannot see — deliberate, but the consequence is that a conversation whose *first* turn is bounded has no title until an unbounded one follows, since titling runs once. ADR 0046 records that as a gap rather than leaving it silent.

**One thing I did not do, and you flagged it yourself.** `main` has moved 17 commits since this stack's base — including `7bcdb52e` and `8a8f9a76` taking `conversation_server.ex` to 2,567, and `a7bfb8ba fix: fence sandbox resets until deletion succeeds (#1768)`, which touches the `reset_sandbox/2` path #1744 changes. Rebasing nine PRs onto that is its own exercise and I did not fold it into this round. It should happen before any of this merges, and `a7bfb8ba` against #1744's `force:` option is the specific thing to look at.

**Verification.** 21 lifecycle and size tests pass. Reverting the gate fails the query-count test; reverting the retirement fix reproduces the `MatchError`. Full core suite at this commit: **6 doctests, 5,042 tests, 0 failures**; at the restacked tip (#1752): **5,095 tests, 0 failures**. `credo --strict` 7,441 mods/funs no issues, `okf validate decisions` valid.

#1750–#1752 are rebased on this and pushed; the chain is clean end to end.

BinaryBourbon (changes requested):
## Review: the gate is fixed properly; the new `:not_found` branch leaks the connection it was meant to put away

Re-reviewed at `db842e5`. The blocking point is answered, and answered at the right altitude — `_unsafe_actor_gate/3` is one unlocked `SELECT` of three columns by primary key, the authoritative checks stay at the transport and at `_unsafe_complete/3`, and `BoundedTurn` is the right home for both halves. I particularly like that `:retire` is explicitly *not* the durable decision: a missing row, a superseded connection, a closed state and a passed deadline all converge on the same locked write one frame later, so narrowing the gate narrows cost and not correctness.

I checked the narrowing rather than taking it: dropping `:identity_unconfirmed` is safe because `_unsafe_claim_spawn/2` leaves the row `active` and the transport owns the stream until `await_ready/2` returns, so the actor sees no message in that window; dropping `:turn_not_running` and `current_binding?/1` is safe because every turn write already goes through `_unsafe_write_turn/3`, and an ownership change that matters retires the journal (`_unsafe_retire_sandbox/1`), which the state check then catches. And the deadline arbitration still lands on `expire/2` inside `complete/3`, so a turn that ran out of time still ends `failed` with its reason rather than `interrupted`.

The query-count test is the right kind of test — measured, collected in the handler and asserted in the body for #1427's reason. The three answered notes are answered accurately; I checked the `bounded_sdk_limits/2` one against `_unsafe_register_bounded/3` and the `with` it describes is really there.

### One thing, and it is in the new code: `abandon/3` closes nothing

```elixir
defp abandon(state, reason, finish_turn) do
  ...
  state = %{state | turn_execution: nil}
  ...
  |> Connection.close_bounded()
end
```

`Connection.close_bounded/1`'s first clause is `def close_bounded(%{turn_execution: nil} = state), do: state`. Clearing `turn_execution` first therefore makes the close at the end of that pipeline a **no-op**, and it is the only thing on this path that would have done the cleanup. What survives:

- `execution_transport` is never sent `:retire`, so the transport keeps the sprite command until its own `:deadline` and `:drain_end` fire — bounded, but it is holding a session nobody is reading;
- `stop_peer/1` is never called and `acp_peer` / `acp_peer_mon` stay set, so the peer lives and its later `{:DOWN, ...}` arrives against state that no longer expects it;
- `current_command` and `current_command_ref` stay set, which is the part that actually bites. `turn_execution` is now nil, so the next prompt takes the unbounded path, finds `Connection.alive?/1` true and resumes onto that connection — and every `handle_execution_info({:acp, ref, payload}, %{current_command_ref: ref})` in the meantime drives a turn this function already failed. That is the #413 shape the rest of this module works hard to avoid.

The success path is fine, for the reason the broken one is not: `finish/3` passes a state that still carries `turn_execution`, so `finish_acp_turn/4`'s own trailing `if state.turn_execution, do: close_bounded_connection(state)` does the work and the outer call is the no-op.

The fix is to let `close_bounded/1` do the clearing it already does — drop the `%{state | turn_execution: nil}` line and keep the pipeline, or close before nilling. Nothing else in `abandon/3` needs the field to be nil early: `close_bounded/1` touches no journal, only local state.

The test cannot catch this as written — `assert is_nil(:sys.get_state(pid).turn_execution)` passes either way, because the broken line is what sets it. Worth asserting `execution_transport`, `current_command_ref` and `acp_peer` are nil too, which is the property the paragraph in the commit message actually claims.

### Smaller

- The one-query assertion sends **every** `[:fountain, :repo, :query]` event in the VM to the test pid, and asserts the global count is one. `ConversationServerCase` is `async: false` so no other test is running, which makes it safe today — but any supervised process that touches the repo in that microsecond turns a correct gate into a red build, and a count of one is the easiest assertion in the world to make flaky. Filtering the drained queries to the ones naming `turn_executions` before counting keeps the property and removes the ambient dependency.
- `defp drain_queries/1` is defined inside the `describe` block. Legal, and it compiles into a module function all the same, but it reads as scoped when it is not.

Pin lands at 2,735 with 45 lines moved out rather than raised, which is the right way to pay for this. Checks are still running on this head; I am not counting them either way.

Everything else across both rounds on this PR is closed.

jhgaylor:
Pushed `76221235`. You're right, and it's the more embarrassing kind of right — the leak was in the clause I added to fix a crash.

**`abandon/3` closed nothing.** `Connection.close_bounded/1`'s first clause is `def close_bounded(%{turn_execution: nil} = state), do: state`, so nilling the field first made the close at the end of that pipeline a no-op — and it was the only thing on the path that sends `:retire` to the transport, calls `stop_peer/1` and drops `current_command`/`current_command_ref`. I verified all six fields it clears in the second clause.

And your read on why it matters is the right one: leaving that set is worse than the `MatchError` it replaced. With `turn_execution` nil the next prompt takes the unbounded path, finds `Connection.alive?/1` true and resumes onto a connection whose turn this function already failed — with every `handle_execution_info({:acp, ref, _}, %{current_command_ref: ref})` in between driving it. A crash loses a server; this loses the invariant.

The field now stays set through `abandon/3`, and both exits close for the reason you gave: with a turn, `finish_turn`'s own trailing `if state.turn_execution, do: close_bounded_connection(state)` does it — the same thing `finish/3` relies on — and without one the pipeline's call does it. `close_bounded/1` clears the field either way.

**The test could not see it, exactly as you said.** `assert is_nil(state.turn_execution)` passed either way, because the broken line was what set it. It now asserts the whole teardown — `execution_transport`, `current_command`, `current_command_ref`, `acp_peer`, `acp_peer_mon` — and reintroducing the early nil fails on `execution_transport`. That is the **second time on this PR** a first-version test passed against the unfixed code, which is worth more than either fix: I am now reverting every fix before believing its test, and both times the revert is what found the real defect.

**Both smaller notes taken.** The query assertion filters drained queries to the ones naming `turn_executions` before counting, so an ambient repo call cannot redden a correct gate; the no-transaction check still spans everything drained, since that property is genuinely global. `drain_queries/1` moved out of the `describe` block.

**Verification.** 20 lifecycle tests pass; reverting the fix fails on `execution_transport`. Full core suite at this commit: **6 doctests, 5,042 tests, 0 failures**; at the restacked tip (#1752): **5,095 tests, 0 failures**. `credo --strict` 7,403 mods/funs no issues. Pin still 2,729, matching the file.

#1750–#1752 rebased on this and pushed; chain clean end to end. Thanks for checking the narrowing rather than taking it — the `:identity_unconfirmed` window was the one I was least sure of.

BinaryBourbon (approved):
## Review: the leak is closed, and the test now fails for the right reason

Re-reviewed at `7622123`. This is the fix, and it is the smaller of the two available ones — the field stays set and `close_bounded/1` does the clearing that is already its job, rather than a second teardown being written beside it.

Both exits check out:

- with a `current_turn`, `finish_turn` is `finish_acp_turn/4`, whose trailing `if state.turn_execution, do: close_bounded_connection(state)` now runs, so the transport gets `:retire`, the peer stops, and `current_command_ref` goes — and the outer `|> Connection.close_bounded()` is the no-op instead of the whole path being one;
- with no turn, the pipeline's own call does it against a state that still carries the journal.

Same shape as `finish/3`, which is what makes it easy to keep true.

The test change is the part I would point at if someone asked what a regression is for. `assert is_nil(state.turn_execution)` was compatible with both the bug and the fix, because the broken line was the one setting it; asserting `execution_transport`, `current_command`, `current_command_ref`, `acp_peer` and `acp_peer_mon` pins the property the comment claims rather than the symptom nearest to hand. Noting in the commit message that this is the second time on this PR a first-draft test passed against unfixed code is the right instinct — that is the check worth keeping: revert the fix and watch it go red, and note which assertion catches it.

The two smaller notes are also answered, and answered better than I asked: filtering to `turn_executions` before counting keeps the one-query property while the `begin`/`savepoint` check still spans everything drained, so the no-transaction half did not get narrowed along with the count. `drain_queries/1` out of the `describe` block.

### One leftover, prose only

The comment block above `abandon/3` still ends with the sentence written for the old body:

> Clearing `turn_execution` drops this server back onto the unbounded path for whatever is left in its mailbox, rather than re-entering the gate against a row that is not there.

Twelve lines later the new comment opens with "`turn_execution` stays set through this". Both are true — `close_bounded/1` does the clearing, just later — but read in sequence they contradict each other, and the first one is the explanation of a line that no longer exists. Fold the surviving half into the new block ("the field is cleared by `close_bounded/1`, which drops this server back onto the unbounded path") and delete the rest.

Not blocking. Checks are green on this head, and with this the whole of #1749 across three rounds is closed for me.

PR #1750, opened 2026-09-08
Fence bounded transcripts and deduplicate delayed usage

jhgaylor:
Cancellation could race an actor's deadline check and still admit a late transcript write. Bounded output and turn-tagged stages now check and write under the journal locks. Retired turns reject stale edits; terminal stages must match the stored outcome. Duplicate usage deliveries lock the parent and turn before incrementing totals, while allowing delayed accounting after retirement.

Stacked on #1749; **draft, public enforcement disabled**. Remaining gates include conversation/session metadata, untagged lifecycle events, sandbox transfer/parking/incarnation safety, released Sprites identity support and live provider acceptance.

Full precommit passes **4,728 tests + 6 doctests, zero failures**. The pinned-toolchain focused suite passes 90 tests, including 12 new regressions. Independent PostgreSQL proofs pass 60 new races and four lock-delay cases, plus 20 existing deadline races. No provider operations occurred. Tracks #1732.

Evidence: `decisions/evidence/bounded-transcript.json`. The first full gate caught an analytics fixture with an invented turn ID; it now inserts a real completed turn and retains all assertions.

[All 24 CI checks passed or intentionally skipped](https://github.com/BinaryBourbon/fountain/actions/runs/34184936407) at `751f8b6db1ec5b86b5d0a70474adf652c06f7ed4`.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2225,9 +2225,6 @@ defmodule Fountain.Conversations do
     # used to raise inside the transaction and take the turn's usage
     # recording with it. Anything that is not a non-negative integer counts
     # as nothing, which is what an unreported figure already counts as.
-    input = counter_value(Map.get(usage, "input"))
-    output = counter_value(Map.get(usage, "output"))
-
     Repo.transaction(fn ->
       # Same parent-before-turn lock order as ExecutionGuard. Delayed accounting
       # survives retirement, but duplicate deliveries cannot debit twice.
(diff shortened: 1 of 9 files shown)


jhgaylor (commented):

BinaryBourbon (commented):
## Review: the writer-side fence is the right place for it

Commenting. Enforcing an elapsed deadline in the turn writer rather than waiting for a coordinator tick is the correct instinct — it closes the window where a late ACP response lands between expiry and the next scan, and it means the guarantee does not depend on the worker being up. The test name says it exactly: "an elapsed deadline is enforced by the writer without a coordinator tick".

Deduplicating delayed usage against the persisted outcome is also the right shape; usage arriving after a turn is already failed was the original complaint in #1732 and this is where it has to be handled.

Notes:

- `Output` changes here affect the unbounded path too. Worth confirming in the PR body which transcript events change shape for a turn that was never bounded, because the conversations app and the team app both parse these (`?blocks=true`), and neither is in this repo.
- Same rebase requirement as the rest of the stack (#1745).

No objection to the approach; holding off approving only because the base is unresolved.

jhgaylor:
Force-pushed `87913aab`. Rebased onto the re-cut #1749, and the note answered with assertions rather than prose.

**Which transcript shapes change for an unbounded turn: none.** The fence can only suppress. A retired bounded turn's output is neither written nor broadcast — `log!/1` returns `nil` for `kind: "output"` behind it, and the `Output` change skips the publish rather than broadcasting a nil. An unbounded turn's event keeps its `id`, `kind`, `stream`, `turn_id` and `data` exactly as before, and there is now a regression on a conversation with no journal row at all that asserts every one of those fields plus the sidebar broadcast. That matters because the conversations and team apps are out of this repo (ADR 0034) and cannot be updated in lockstep.

**One change does reach the unbounded path, and it is worth calling out.** A second usage delivery for a turn that already has usage is now refused with `:already_recorded` instead of overwriting it. A retried provider delivery used to double-count the conversation's `usage_input_tokens`/`usage_output_tokens`, so this is a fix — but it is a fix on the common path, not a bounded-turn detail, and the PR body did not say so. Second regression covers it: the first delivery sticks, the second is refused, and the counters stay at the first values. ADR 0046 records both.

I checked the one thing that would have made the dedup dangerous: `field :usage, :map` has no default, so it is `nil` until a delivery lands and `is_map(current.usage)` cannot refuse a legitimate first write.

The rebase itself needed only the `_unsafe_write_event/3` addition merged beside #1748's `_unsafe_write_turn/3` docstring — no semantic conflict.

**Verification.** 14 transcript tests pass. Full core suite: **6 doctests, 5,053 tests, 0 failures, 2 skipped**. Format, `credo --strict` (7,401 mods/funs, no issues), `okf validate decisions` valid.

On to #1751.

BinaryBourbon (approved):
## Review: the question is answered by assertion, which is the only way it could be

Re-reviewed at `32eb823`. My note was "say which transcript events change shape for a turn that was never bounded", and the answer — none — is now pinned rather than reasoned about. That matters more here than usual: the two apps that read these events live in other repos (ADR 0034) and cannot be updated in lockstep, so "the fence can only suppress" needs to be a test and not a paragraph. `id`, `kind`, `stream`, `turn_id`, `data` and the `{:log_event, _}` broadcast all asserted on a conversation with no journal at all.

**The one change that does reach the unbounded path** is called out honestly, and I went and checked it rather than taking the reasoning: `_unsafe_record_turn_usage/2` has exactly one caller, `TurnMachine.handle({:done, stop_reason, usage}, ctx)`, whose own comment says the figure is recorded once because the prompt response is the only place the runtime reports it, and `record_usage/2` logs rather than raises on refusal. So a second delivery really is a re-delivery, keeping the first is right, and there is no cumulative-report case where this would under-count. The conversation counters were being double-incremented before; `:already_recorded` is the fix, not a behaviour regression.

ADR 0046 records both halves. Approving.

PR #1751, opened 2026-09-08
Fence parent state and orphan recovery by turn generation

jhgaylor:
A late completion or session callback could overwrite parent state after cancellation and successor admission. Parent/session writes now check the exact turn generation under database locks. Admission commits running status with the turn; orphan recovery retires execution first and preserves uncertain spawns and deadline outcomes.

Stacked on #1750; **draft, public enforcement disabled**. Remaining gates: sandbox release/transfer/park/reset/incarnation safety, released Sprites identity support and live provider acceptance. Tracks #1732.

Full precommit: **4,755 tests + 6 doctests, zero failures**. Includes 27 new regressions and two real-actor callback barriers. Independent PostgreSQL proofs pass 80 new races and four delayed-lock cases, plus 20 earlier deadline races. Provider acknowledgments are synthetic; no provider operations occurred.

Evidence: `decisions/evidence/turn-parent-fences.json`. Records corrected fixture expectations, Credo findings and the webhook source-catalogue failure; no checks were weakened. ADR validation and secret scanning pass.

[CI at `8381e6a7`](https://github.com/BinaryBourbon/fountain/actions/runs/34187729039) passes the test partitions and other checks but fails the dependency audit: EEF-CVE-2026-32686 now includes Decimal 3.1.1. The [maintainer advisory](https://github.com/ericmj/decimal/security/advisories/GHSA-rhv4-8758-jx7v) lists 3.0.0 as patched; the current EEF feed lacks that fixed boundary. This discrepancy is under investigation; CI has not been waived.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1176,6 +1176,42 @@ defmodule Fountain.Conversations do
     end)
   end
 
+  @doc "Idle only the latest ended turn's still-running parent."
+  def _unsafe_idle_after_turn(%Turn{} = turn),
+    do: write_turn_parent(turn, :idle, %{status: "idle"})
+
+  @doc "Apply a runtime session report only for the current, unretired turn."
+  def _unsafe_set_turn_session(%Turn{} = turn, session_id),
+    do: write_turn_parent(turn, :session, %{runtime_session_id: session_id})
+
+  @doc "Clear an idle legacy peer's session only if its observed identity is still current."
+  def _unsafe_clear_idle_session(conv_id, expected) do
+    # ownership: this is the already-owned actor's conversation and session.
+    ExecutionGuard._unsafe_clear_idle_session(conv_id, expected, fn current ->
+      current |> Conversation.changeset(%{runtime_session_id: nil}) |> Repo.update()
+    end)
+    |> notify_parent_change()
+  end
+
+  defp write_turn_parent(turn, mode, attrs) do
+    # ownership: the calling actor/recovery path already owns this exact turn.
+    result =
+      ExecutionGuard._unsafe_write_parent(turn, mode, fn current ->
+        current |> Conversation.changeset(attrs) |> Repo.update()
+      end)
+
+    notify_parent_change(result)
+  end
+
+  defp notify_parent_change(result) do
+    case result do
+      {:ok, %{applied: true, conversation: conv}} -> broadcast_sidebar_update(conv.user_id)
+      _ -> :ok
+    end
+
+    result
+  end
+
   @doc """
   Re-resolve the Agent, Environment and Vault for an existing conversation,
   on the machine it is already running (#1565).
@@ -1726,6 +1762,15 @@ defmodule Fountain.Conversations do
 
         unless attached?, do: Repo.rollback(:sandbox_unavailable)
 
+        # A terminated or failed parent takes no more turns. `attached?` above
+        # checks the machine; this checks the conversation, which a retired
+        # actor can still reach with a queued prompt.
+        if Repo.exists?(
+             from c in Conversation,
+               where: c.id == ^conv_id and c.status in ["terminated", "failed"]
+           ),
+           do: Repo.rollback(:not_running)
+
         case _unsafe_check_saved_execution_allowance(conv_id) do
           :ok -> :ok
           {:error, reason} -> Repo.rollback(reason)
@@ -1762,11 +1807,21 @@ defmodule Fountain.Conversations do
         end
       end)
 
-    with {:ok, turn} <- result do
-      record_turn_usage(turn)
-      {:ok, turn}
-    end
+    record_started_turn(result)
   end
+
+  defp record_started_turn({:ok, turn}) do
+    record_turn_usage(turn)
+
+    case Repo.get(Conversation, turn.conversation_id) do
+      nil -> :ok
+      conv -> broadcast_sidebar_update(conv.user_id)
+    end
+
+    {:ok, turn}
+  end
+
+  defp record_started_turn(error), do: error
 
   @doc """
   Save an initial resolved allowance once, scoped to its conversation owner.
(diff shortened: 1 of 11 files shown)


jhgaylor (commented):

BinaryBourbon (commented):
## Review: generation-fencing parent state is right; the red CI is staleness, not this PR

Commenting.

**On the red checks:** `Elixir static analysis` and `CI required` fail on `scripts/hex-audit-gate.exs`, not on anything in this diff — the `decimal 3.1.1` advisory `EEF-CVE-2026-32686`. `main` acknowledged that in `62654eea` / `f645be21` (#1753), which this branch predates by ~60 commits. `mix credo --strict` reports "6928 mods/funs, found no issues" and the compile is clean. One rebase clears both, and the run on the checks tab is from 2026-09-08, so it is three days stale in any case.

**On the change:** fencing parent state and orphan recovery by turn generation is the right answer to the problem — an orphan reconciler that cannot tell "this turn" from "a later turn on the same conversation" is how a recovery path ends up clobbering live work, and a monotonic generation is the cheapest way to express it. The `turn_machine.ex` changes look consistent with that.

Two things:

- The 100 changed lines in `turn_machine.ex` and 118 in `conversations.ex` overlap the same area `main` has been moving (labels #1841–#1849, the acp runtime #1832–#1839, admission #1787–#1793). Re-read `reserve_initial_conversation/4` and the orphan reconciler after the rebase rather than trusting a clean merge.
- `scripts/verify-turn-parent-races.exs` plus `decisions/evidence/turn-parent-fences.json` is a good pattern and I am glad it is here. One ask: the evidence file records `"reason": "Unknown spawn recovery correctly returns failed with awaiting_identity and spawn_unconfirmed; corrected the expectation instead of weakening uncertainty policy."` — that is exactly the dead-end I have written up on #1744. The script asserting it as correct is fine; ADR 0046 asserting it as *complete* is not, and that is the line I would like changed.

jhgaylor:
Force-pushed `62969ec8`. Rebased onto the re-cut #1750, both notes addressed.

**The red CI was staleness, confirmed.** `hex-audit-gate.exs` was failing on the `decimal` advisory `EEF-CVE-2026-32686`, which main acknowledged in #1753 sixty commits after this branch forked. It exits 0 here. One advisory line remains: #1746's dependency bump moved the locked Decimal artifact, so the acknowledgment no longer matches anything and *could* be removed — that is main's call, not this stack's, and it does not block.

**The evidence file no longer reads as if the dead-end were the design.** `turn-parent-fences.json` said unknown spawn recovery "correctly returns failed with `awaiting_identity`". That is true as an immediate outcome and the expectation was right to correct — but stated alone it reads as the *final* state, which is precisely what I could find no way out of when reviewing #1744. It now says `awaiting_identity` is an obligation with an age and names the sweep that writes it off.

### Two things the rebase changed

**`_unsafe_create_autonomous_turn/1` is dropped.** It admitted background work through `_unsafe_create_turn/1`, which skips the sandbox binding #1764 added, and leaned on the `_unsafe_autonomous_turn/2` helper the #1749 re-cut removed for the same reason. Autonomous turns now use the one admission path, so `turn_parent_test.exs`'s "closed parents refuse user and background admission" exercises both doors as the same door — which is what it was really asserting.

**Two improvements this branch had that main's admission lacked, now moved into it:**

- a terminated or failed *parent* refuses a turn with `:not_running`. `attached?` checks the machine; a retired actor can still reach the conversation with a queued prompt. This branch had the check inside `_unsafe_admit_turn`; it is now beside `attached?`.
- the parent goes `running` inside the admission transaction. A reader could previously see a `running` turn under an `idle` parent for the width of the launch, and a launch that failed left the pair disagreeing. `update_all` rather than `update_conversation/2`, because that one audits and an audit insert must not run inside a transaction (ADR 0013). One of your own tests — "admission commits running status even when an older snapshot still said running" — caught that I had dropped this, which is the test doing its job.

**Kept, because it is the point of the PR:** `_unsafe_write_parent/3`, `_unsafe_recover_turn/2`, `_unsafe_clear_idle_session/2`, and `session_plan/2` returning `{:error, :execution_fenced}` so even the placeholder session write is tied to the admitted turn. That last one is a nice catch — the placeholder write was the one thing that could outlive a cancellation.

**Verification.** 28 parent/actor/size tests pass. Full core suite: **6 doctests, 5,080 tests, 0 failures, 2 skipped**. Format, `--warnings-as-errors`, `credo --strict` (7,429 mods/funs, no issues), `hex-audit-gate.exs` exit 0, `okf validate decisions` valid. Size pin 2739 → 2733.

On to #1752, the last one.

BinaryBourbon (approved):
## Review: the evidence file no longer asserts the dead-end as the design

Re-reviewed at `37e3965`. Both notes answered, and the rebase cleared the red `hex-audit-gate` as expected — checks are green on this head.

**The evidence file.** `turn-parent-fences.json` now says `awaiting_identity` is the intended *immediate* outcome and not the final state, and names `_unsafe_retire_unresolved/2` as the way out. That was the line I wanted changed and it is the right change: the script asserting the immediate behaviour was always correct; it was the absence of an exit that made it read as policy.

**The second autonomous door is gone.** `_unsafe_create_autonomous_turn/1` bypassed #1764's sandbox binding, and dropping it so `Connection.open_autonomous_turn/3` goes through `_unsafe_create_turn_on_sandbox/3` makes "closed parents refuse user and background admission" a test of one door rather than two similar ones.

**The parent flip moving into the admission transaction** is the right call — a `running` turn under an `idle` parent was visible for the width of the launch, and a failed launch left the pair disagreeing. The `update_all`-not-`update_conversation/2` comment gives the audit reason (ADR 0013), which is half of it. The other half is unwritten and worth a clause: dropping `update_conversation/2` also drops its `broadcast_sidebar_update/1`, and the reason that is safe is that `record_started_turn/1` still broadcasts after the transaction commits. Two separate facts, one currently left for the reader to rediscover.

**One thing to confirm rather than change.** `Connection.open_autonomous_turn/3` puts `status: "running"` in its attrs, so the new in-transaction flip fires for background follow-ups too — where previously the flip lived in the launch, which an autonomous turn never takes, and the parent stayed `idle` while background work ran. `_unsafe_idle_after_turn/1` still puts it back, and the new behaviour is arguably the more honest one, but it is a status change the conversations and team apps will see and it is not mentioned in the commit message or ADR 0046.

The generation fencing this PR is actually for — `_unsafe_write_parent/3`, `_unsafe_recover_turn/2`, `_unsafe_clear_idle_session/2`, and `session_plan/2` refusing with `{:error, :execution_fenced}` so even the placeholder session write is tied to the admitted turn — reads consistently after the re-cut. Approving.

PR #1774, opened 2026-09-09
fix: refuse unenforced execution limits before admission

jhgaylor:
Launch requests containing `execution_limits` could return 200/201 while Fountain ignored the limits. Fresh channel requests could replace the binding without honoring them.

Check requested limits after the tenant-scoped agent lookup, before launch, attachment or channel mutation. Nonempty controls return `422 execution_limits_unsupported`; malformed values return `422 execution_limits_invalid`. Omitted, null and empty limits preserve existing behavior. All controls remain unsupported until enforcement is integrated.

Stacked on #1773 (`feat/execution-limit-policy`): four files, +171/-1. Saved allowances and runtime enforcement remain separate work in #1732.

Validation: seven baseline regressions reproduced; 155 related tests passed, including refusal without row/channel changes and ordinary admission. Full precommit passed 4,717 tests and six doctests. CI and secret scanning passed on `71ce96c8`; advisory prose reports reviewed.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1792,7 +1792,7 @@ defmodule Fountain.Conversations do
   # ── high-level lifecycle ──────────────────────────────────────────────────────────
 
   alias Fountain.Agents
-  alias Fountain.Conversations.ConversationServer
+  alias Fountain.Conversations.{ConversationServer, ExecutionLimits}
 
   @doc """
   Like `start_conversation/2`, but a conversation already bound to
@@ -1834,6 +1834,7 @@ defmodule Fountain.Conversations do
       )
       when is_binary(channel_id) and channel_id != "" do
     with %Agents.Agent{} = agent <- Agents.get_agent(agent_id, user_id) || {:error, :not_found},
+         :ok <- check_execution_limits(attrs["execution_limits"]),
          {:ok, vault_id} <- resolve_vault_id(attrs["vault_id"], user_id, agent),
          {:ok, env_id} <- resolve_environment_id(attrs["environment_id"], user_id, agent) do
       case find_channel_conversation(user_id, agent.id, vault_id, env_id, channel_id) do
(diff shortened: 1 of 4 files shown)


BinaryBourbon (approved):
Approving.

The refusal lands in the right place: after the tenant-scoped `Agents.get_agent/2` (so a foreign agent still 404s before any limit validation) and before every write on all three entry points — `start_conversation_in_channel`, `start_conversation` and `attach_conversation`. Those are the three doors into a conversation row, so there is no way in that skips it.

`FallbackController` matches the house shape: `error` + `errors` for the validation case, `error` + `message` for the policy refusal, exactly like `vault_not_allowed` / `permission_policy_widens` above it.

Checked the edge cases against `ExecutionLimits.normalize/1`: `nil` -> `{:ok, %{}}` -> admitted; `[]` falls to the catch-all clause -> `object_required`; `%{}` -> admitted. The test file covers all three plus the "does not echo input" property, which is the part that would be easy to regress later.

One note for whoever merges the stack: this commit also exists on `fix/account-ceiling-launch-preflight` as `d4ea7e76`, where #1785 builds `check_execution_limits/2` on top of it. Land this first and rebase that chain, or land #1785 and close this — merging both as merges will conflict in `conversations.ex` and add/add-conflict on `execution_limit_admission_test.exs`.

BinaryBourbon (approved):
Approving.

The refusal lands in the right place: after the tenant-scoped `Agents.get_agent/2` (so a foreign agent still 404s before any limit validation) and before every write on all three entry points — `start_conversation_in_channel`, `start_conversation` and `attach_conversation`. Those are the three doors into a conversation row, so there is no way in that skips it.

`FallbackController` matches the house shape: `error` + `errors` for the validation case, `error` + `message` for the policy refusal, exactly like `vault_not_allowed` / `permission_policy_widens` above it.

Checked the edge cases against `ExecutionLimits.normalize/1`: `nil` -> `{:ok, %{}}` -> admitted; `[]` falls to the catch-all clause -> `object_required`; `%{}` -> admitted. The test file covers all three plus the "does not echo input" property, which is the part that would be easy to regress later.

One note for whoever merges the stack: this commit also exists on `fix/account-ceiling-launch-preflight` as `d4ea7e76`, where #1785 builds `check_execution_limits/2` on top of it. Land this first and rebase that chain, or land #1785 and close this — merging both as merges will conflict in `conversations.ex` and add/add-conflict on `execution_limit_admission_test.exs`.

PR #1775, opened 2026-09-09
feat(conversations): store versioned execution allowances

jhgaylor:
Store each conversation's resolved allowance separately from ordinary conversation metadata. Partial updates retain omitted controls; widening fails; revision checks reject stale writes. A second database writer cannot replace a committed tighter allowance. Duplicate inserts fail, and rollback refuses to discard saved allowances.

Storage primitive only: this adds no user setting, endpoint, admission caller or runtime enforcement. Based on #1773; a focused replacement for part of #1745/#1754. Tenant-scoped admission and later-turn/recovery integration remain in #1732, alongside the requested-limit refusal in #1774.

Validation: 46 affected tests passed, including separate PostgreSQL connections with an observed lock wait. Removing revision checking produces four regression failures. Empty migration down/up passed; occupied rollback preserved both the allowance and migration version. The first full run caught retained test sandboxes; cleanup now removes them explicitly and asserts their absence. Full precommit passed 4,719 tests and six doctests; the dedicated database was empty afterward. Secret scanning passed. [CI passed](https://github.com/BinaryBourbon/fountain/actions/runs/34331964389) on `241d884f`. No provider calls or live execution-limit acceptance.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/execution_allowance.ex
+++ b/apps/fountain/lib/fountain/conversations/execution_allowance.ex
@@ -1,0 +1,53 @@
+defmodule Fountain.Conversations.ExecutionAllowance do
+  @moduledoc """
+  Versioned storage for a conversation's resolved execution allowance.
+
+  This is a storage primitive, not an admission or enforcement API. No service
+  path writes these records yet. Admission must establish conversation ownership,
+  resolve current ceilings and prove runtime support before saving an allowance.
+  Later turns and recovery must consult it before this becomes a usable setting.
+
+  Updates must use `narrow_changeset/2`. A stale revision fails with
+  `Ecto.StaleEntryError` (or a changeset error with `stale_error_field: :revision`).
+  Reload and revalidate the request after a conflict; never retry the old map as
+  a replacement. An opaque revision avoids counter rollover accepting old writes.
+  This record does not reset or replace an in-flight turn's deadline or usage.
+  """
+
+  use Ecto.Schema
+  import Ecto.Changeset
+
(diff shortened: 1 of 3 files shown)


BinaryBourbon (approved):
Approving.

The storage design is right: separate table so an ordinary conversation update cannot touch policy, opaque UUID revision instead of a counter, `optimistic_lock` on every narrowing, and `new_changeset` constrained by the PK so a duplicate insert cannot replace a committed allowance. The "stale omission cannot restore another field's old allowance" test is the one that matters most and it is there.

Two things I want on the record rather than blocking on:

**`unboxed_run` is new to this repo.** Nothing else under `apps/` or `ee/` uses it. The race test commits rows to `fountain_test` outside the SQL Sandbox. Within one `mix test` run this is safe — ExUnit runs sync cases after async ones, so nothing overlaps — and the `after` block cleans up. The exposure is two suites sharing one database, which happens locally with parallel worktrees: a committed conversation is visible to another run's unscoped assertions (`conversations_start_test.exs:361` asserts `Repo.aggregate(Conversation, :count) == 0`). It is also unrecoverable if the BEAM is killed mid-test. Worth a line in the moduledoc saying the fixtures are committed on purpose and must be cleaned, so the next person adding a race test here copies the discipline and not just the shape.

**The raising `down` is also new.** No other migration in `priv/repo/migrations` raises on rollback. Refusing to discard saved allowances is defensible, but it means a rollback during an incident fails hard rather than degrading, and the operator has to know to empty the table first. Given the current state (nothing writes these rows) the risk is nil; it is worth remembering when this becomes load-bearing.

Minor: `insert_allowance/1` is duplicated verbatim in both test modules.

PR #1776, opened 2026-09-09
fix(conversations): check saved allowances before turn admission

jhgaylor:
A saved execution allowance could be ignored when a user or autonomous turn opened. Check it inside the shared turn transaction: unsupported or malformed limits refuse admission without a turn or usage event. Missing and empty allowances retain ordinary behavior.

Lock the conversation row to serialize first allowance inserts through its foreign key, then hold the allowance row's share lock through turn insertion. Both write orders are covered with distinct PostgreSQL connections and observed lock waits.

Based on #1775; a focused replacement for part of #1745/#1754. This adds no policy-writing API or runtime enforcement. Synchronous prompt, channel, wake and recovery gates remain separate work in #1732; #1774 handles requested limits at launch.

Validation: eight failures reproduced on the parent; 47 related tests passed, including four database races, both turn openers, tenant separation, JSON-null corruption and fixture cleanup. Full precommit passed 4,730 tests and six doctests; the dedicated database was empty afterward. Secret scanning passed. [CI passed](https://github.com/BinaryBourbon/fountain/actions/runs/34334257814) on `163a4dff`. No live execution-limit acceptance is claimed.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -12,6 +12,7 @@ defmodule Fountain.Conversations do
 
   alias Fountain.Audit
   alias Fountain.Conversations.{Blocks, Conversation, LogEvent, Sandbox, Turn, TurnImage}
+  alias Fountain.Conversations.{ExecutionAllowance, ExecutionLimits}
   alias Fountain.Repo
 
   # ── on the _unsafe_ prefix ────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


BinaryBourbon (approved):
Approving.

The lock reasoning holds up. Admission takes the sandbox advisory lock, then `Conversation FOR UPDATE`, then `ExecutionAllowance FOR SHARE`, and keeps both row locks through the turn insert. Because `FOR UPDATE` conflicts with the `FOR KEY SHARE` that the allowance's foreign key needs, the conversation lock fences a *first* insert in the window where there is no allowance row to lock yet. That is the case a naive `SELECT ... FOR SHARE` on the allowance alone would miss, and the comment says so.

I walked the deadlock cases against the other writers in the stack and could not build a cycle: `narrow_changeset` updates only `limits`/`revision`, so it never re-checks the FK and never wants the conversation row, and nothing that holds an allowance lock goes on to want the sandbox advisory lock.

Behavior is preserved on the paths that do not have an allowance: the new `Repo.one(... FOR UPDATE) || Repo.rollback(:sandbox_unavailable)` returns the same refusal the `attached?` check already produced for a missing conversation, so no caller sees a new error.

`check_saved_execution_allowance/1` re-normalizing an already-normalized stored map is the right kind of paranoid — it is what makes the `'null'::jsonb` case in the test a refusal instead of "unrestricted", and that is exactly the failure mode worth defending against.

The added contention is on a path the per-sandbox advisory lock already serializes, so the extra row lock costs close to nothing.

PR #1777, opened 2026-09-09
fix: refuse unsupported saved allowances before wake

jhgaylor:
A dormant conversation with an unsupported saved allowance could probe/resume a provider, start a worker and queue a prompt before turn admission refused it. Check the saved allowance first; prompt-triggered wake returns HTTP 422 without those effects. Missing or empty allowances retain ordinary behavior. Reconnecting to interrupt an existing running turn remains available, without a new prompt.

Stacked on #1776 (`fix/saved-allowance-turn-admission`). The two fallback clauses are reused **identically** from sibling #1774; no other sibling changes are imported. Focused replacement for part of #1745/#1754, tracked in #1732.

Validation: 12 new tests, including six sandbox states, real HTTP responses, tenant isolation and cancellation; eight fail against the parent. All 182 related tests pass. Full precommit passes 4,742 tests plus six doctests, zero failures; diff secret scan passes.

This is a preflight, not a lock held across provider operations; #1776 rechecks within turn admission. Active-server prompt admission, channel reuse, direct boot recovery, allowance saving/current ceilings and runtime enforcement remain separate. No new execution controls are enabled, and this is not live enforcement acceptance.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2993,11 +2993,19 @@ defmodule Fountain.Conversations do
   (`terminated`, `failed`) — those don't auto-resume.
   """
   def wake_conversation(conv_id, initial_prompt \\ nil) do
-    # Ownership: called from ConversationServer (which established ownership
-    # before starting) and the boot-time rehydrator sweep. The agent fetched
-    # below is the conversation's own agent_id, same tenant by construction.
+    wake_conversation_for(conv_id, initial_prompt, :work)
+  end
+
+  defp wake_conversation_for(conv_id, initial_prompt, purpose) do
+    # Ownership is established by callers before reaching this internal wake
+    # path. The agent fetched below is the conversation's own agent_id,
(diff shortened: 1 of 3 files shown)


BinaryBourbon (approved):
Approving.

Splitting `wake_conversation/2` into `wake_conversation_for/3` and exempting `:interrupt` is the right call. Cancellation has to stay reachable no matter what policy says, and a saved allowance blocking the one way to stop a running turn would be a genuinely bad failure. The "a saved limit does not prevent reconnecting to interrupt a running turn" test pins that, and "cancellation does not wake an idle conversation" pins the other half.

The check sits after `assert_resumable/1` and before the agent lookup, so nothing has touched the provider, started a worker or queued a prompt when it refuses. The six-sandbox-status loop with `refute_side_effects/0` is a good way to prove that — it asserts on the absence of the probe, the resume, the worker start and the queued prompt, not just on the return value.

Noted that the two `FallbackController` clauses are byte-identical to #1774's, as the description says. Same merge-order caveat I left there.

PR #1779, opened 2026-09-09
fix: guard saved limits before queued prompt handoff

jhgaylor:
A queued initial prompt could complete autonomous work and close its connection before turn admission refused an unsupported saved allowance. Check saved policy when consuming the cast, before handoff; refusal preserves the current turn, connection, activity and usage. Cancellation and ordinary empty-policy handoff remain available.

Stacked on #1778 (`fix/saved-allowance-active-prompt`), reusing its policy helper. Focused replacement for part of #1745/#1754; tracked in #1732.

Validation: six new real-server tests cover idle/autonomous refusals for unsupported/malformed policy, cancellation, empty-policy handoff and busy-user preservation. Four fail against the parent. All 165 related tests pass. Full precommit passes 4,754 tests plus six doctests, zero failures; diff secret scan passes.

A cast has no synchronous caller response. This is preflight, with #1776's transactional recheck retained for intervening policy writes. It adds no atomic prompt receipt, retroactive enforcement, policy writer or runtime controls. Channel/boot admission, ceilings and live bounded-execution acceptance remain required.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -1671,15 +1671,16 @@ defmodule Fountain.Conversations.ConversationServer do
     else
       conv = Conversations._unsafe_get_conversation!(state.conversation_id)
 
-      case TurnMachine.gate(conv.user_id, state.inference_source) do
-        :ok ->
-          state = close_autonomous_turn(state, "superseded_by_prompt")
-          agent = if conv.agent_id, do: Agents._unsafe_get_agent!(conv.agent_id)
-          {:noreply, kick_turn(state, prompt, agent, images)}
-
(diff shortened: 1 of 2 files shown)


BinaryBourbon (approved):
Approving.

Same treatment as #1778 applied to the cast path, which is the right symmetry: a queued initial prompt should not be able to close a live autonomous turn and its connection just because it arrived before the policy was checked.

The `case` -> `with` rewrite is mechanical and the log line is preserved. The replacement comment is more accurate than the one it removes: after #1777 the wake door does refuse synchronously, but this is no longer only a backstop that "has to not run the turn" — it now also has to not destroy existing work, and the new wording says that.

Relying on same-sender mailbox ordering to wait out the cast before asserting is legitimate here and the comment explains it, so nobody reads it later as a missing sync.

PR #1781, opened 2026-09-10
fix: refuse saved limits before boot recovery

jhgaylor:
Boot recovery started conversation servers directly, bypassing the saved-allowance wake guard. Check saved policy before agent/runtime lookup and worker start. Unsupported or malformed policy skips that conversation without changing its rows; other tenants and absent/empty allowances still recover normally.

Stacked on #1780 (`fix/saved-allowance-channel-resume`). Focused replacement for part of #1745/#1754; tracked in #1732.

Validation: six new sweep tests cover idle/running conversations, all three controls, repeated refusal, corrupt maps/JSON null, mixed tenants, existing servers and non-ready sandboxes. Five fail on the parent. All 168 related tests pass, including cancellation and real-server regressions. Full precommit passed: 4,770 tests and 6 doctests, zero failures (2 skipped, 7 excluded).

This is preflight; turn admission retains its transactional recheck. It does not stop already-running provider work, enforce deadlines, add policy writes or activate new recovery behavior. Existing cancellation remains available. Runtime integration and live bounded-execution acceptance remain required.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/rehydrator.ex
+++ b/apps/fountain/lib/fountain/conversations/rehydrator.ex
@@ -11,6 +11,10 @@ defmodule Fountain.Conversations.Rehydrator do
   sandboxes from a crashed mid-provision are left as-is — the user's next
   action lazily resolves them via `wake_conversation`.
 
+  Saved execution allowances must be enforceable before a server is started.
+  Refused rows remain unchanged; this preflight does not stop provider work
+  that was already running. Turn admission still rechecks the saved policy.
+
   ## Clustered boot
 
   `run/1` fires on every node as it boots, but libcluster needs a few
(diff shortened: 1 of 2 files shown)


BinaryBourbon (approved):
Approving.

This is the gap that mattered most in the set: `Rehydrator.spawn_server/1` starts `ConversationServer` directly rather than through `wake_conversation/2`, so #1777's guard did not cover boot. Closing it in `spawn_server/1` rather than in the sweep query is the right level — the existing `else` already turns `{:error, reason}` into a skip-and-log, so the new clause needs no new control flow.

The log assertion is the detail I would have asked for if it were not there: `inspect({:execution_limits_invalid, "unknown_field"})` is safe precisely because `normalize/1` returns the fixed string `"unknown_field"` and never the offending key, and the test asserts both that the reason appears and that the private key and value do not.

Converting the file to `use Fountain.DataCase, async: false` is justified and the comment says why — the sweep reads every resumable row, so it cannot run concurrently with fixtures on independent connections. That is a real requirement rather than a convenience, which is the bar for `async: false` here.

Worth knowing for later: `assert sweep() == 2` is an exact count over all resumable conversations, so it will notice if a future committed-row fixture leaks. That is arguably a feature.

PR #1782, opened 2026-09-10
fix: refuse narrowing corrupt saved allowances

jhgaylor:
Narrowing a saved JSON `null` allowance treated it as unrestricted: even an omitted request replaced it with `{}` and advanced its revision. Reject non-object saved policy before narrowing so corrupt limits, revision and timestamps remain unchanged. Explicitly empty allowances still narrow normally.

Stacked directly on #1775 (`feat/versioned-execution-allowances`); this storage repair does not need the later admission guards. Further extraction from #1745/#1754, tracked in #1732.

Validation: reproduced the JSON-null replacement on the parent. Three new database tests cover null and malformed saved policy, omitted/empty/nonempty requests, row preservation, normalized errors and valid empty-policy narrowing. All 32 allowance/policy tests pass, including the independent-connection PostgreSQL stale-writer race. Full precommit passed: 4,722 tests and 6 doctests, zero failures (2 skipped, 7 excluded).

No public policy writer or runtime control is enabled. Tenant-scoped saving, current-ceiling checks and live bounded-execution acceptance remain required.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/execution_allowance.ex
+++ b/apps/fountain/lib/fountain/conversations/execution_allowance.ex
@@ -12,6 +12,7 @@ defmodule Fountain.Conversations.ExecutionAllowance do
   Reload and revalidate the request after a conflict; never retry the old map as
   a replacement. An opaque revision avoids counter rollover accepting old writes.
   This record does not reset or replace an in-flight turn's deadline or usage.
+  A null saved policy is corrupt, not an unrestricted allowance to narrow.
   """
 
   use Ecto.Schema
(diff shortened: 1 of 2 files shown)


BinaryBourbon (approved):
Approving.

Good catch, and the reasoning in the one-line moduledoc addition is the whole fix: a JSON `null` in `limits` is corruption, not "no restrictions to narrow". Before this, `for_resume(nil, nil, nil, nil)` normalized the stored `nil` to `%{}` and happily wrote `%{}` back over it while advancing the revision, so a corrupt row quietly became an unrestricted one and the timestamps lost the evidence.

Guarding with `when is_map(limits)` on the head rather than inside `put_limits/2` is the right shape — the fallback clause cannot accidentally acquire an `optimistic_lock` and attempt a write, because it never reaches one.

The three-request loop (`nil`, `%{}`, nonempty) matters: the omitted-request case is the one that looked harmless and was not.

Stacking this directly on #1775 rather than behind the admission guards is correct — it is a storage repair and nothing in #1776..#1781 is a prerequisite for it.

PR #1783, opened 2026-09-10
feat: narrow execution allowances by owner

jhgaylor:
Add an owner-scoped context operation for narrowing an existing execution allowance. Missing and foreign records return the same error; omitted controls stay in force, wider or corrupt policy is refused, and concurrent updates revalidate the locked current value. Unchanged requests leave the row untouched.

The context audits successful changes after its transaction, recording only changed control names with caller attribution. Narrowing does not start work or reset an active turn. Initial allowance creation, current host/account ceiling checks and HTTP enrollment remain separate.

Stacked on #1781 (`fix/saved-allowance-boot-guard`) so saved-policy admission and recovery guards are present. Focused replacement for part of #1745/#1754, tracked in #1732. The context rejects stored JSON null itself; sibling #1782 still repairs direct storage-primitive callers.

Validation: 117 related tests pass, including tenant denial, no-op/error audit behavior, unchanged active state, unsupported-turn refusal, and two observed PostgreSQL lock interleavings for concurrent scoped writers. The existing stale-writer and admission races remain covered. Full precommit passed: 4,778 tests and 6 doctests, zero failures (2 skipped, 7 excluded).

No endpoint, runtime control or new recovery behavior is enabled. Bounded live stop/restart/deletion acceptance remains required.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1152,6 +1152,72 @@ defmodule Fountain.Conversations do
   end
 
   @doc """
+  Narrow an existing allowance owned by `user_id`, retaining omitted controls.
+
+  This edits future policy only: it neither admits work nor resets an active
+  turn's usage/deadline. Initial allowance creation and current-ceiling checks
+  remain admission responsibilities. Missing and foreign records return the
+  same error. Concurrent writers revalidate against the latest locked value.
+  """
+  def narrow_execution_allowance(conversation_id, user_id, request, opts \\ []) do
+    result =
+      Repo.transaction(fn ->
+        # Keep ownership stable through the write; turn admission locks this
+        # conversation before its allowance too. No sandbox/provider work here.
+        Repo.one(
+          from c in Conversation,
+            where: c.id == ^conversation_id and c.user_id == ^user_id,
+            select: c.id,
+            lock: "FOR SHARE"
+        ) || Repo.rollback(:not_found)
+
+        allowance =
+          Repo.one(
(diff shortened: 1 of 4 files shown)


BinaryBourbon (approved):
Approving.

The audit placement is right where the contributor guide requires it — `Audit.record/1` is outside `Repo.transaction/1`, so a failed audit insert cannot abort the narrowing it is describing. The metadata is control *names* only, never values, which is the other half of that rule.

Lock ordering checked against the rest of the stack and it is deliberate and safe:

- narrowing: `Conversation FOR SHARE` -> `ExecutionAllowance FOR UPDATE`
- turn admission (#1776): `Conversation FOR UPDATE` -> `ExecutionAllowance FOR SHARE`

Both take the conversation row first, and the two conversation-level modes conflict, so one always waits at the first lock rather than acquiring one and blocking on the other. Two concurrent narrowings share `FOR SHARE` and then serialize on the allowance, with only one lock left to contend for. No cycle either way.

The no-op detection is the subtle bit and it is correct: `optimistic_lock` always stages a `:revision` change, so testing `not Map.has_key?(changeset.changes, :limits)` is what distinguishes "nothing to write" from "write a new revision", and it keeps `updated_at` and the revision stable for an omitted or identical request. The test asserts `updated == narrowed` across both `nil` and `%{}`.

Rejecting a non-map `limits` in the context rather than relying on the schema is right given this branch does not carry #1782 — and the two are consistent, so whichever lands first the other stays correct.

The `:disjoint` race is the good addition: proving two scoped writers keep *each other's* tighter fields is a stronger property than proving the loser fails.

PR #1784, opened 2026-09-10
feat: store typed account execution ceilings

jhgaylor:
Add typed storage for operator-owned account execution ceilings, with an empty default for existing accounts. A dedicated changeset validates the three supported controls; registration, OAuth, principal creation and ordinary profile changes cannot set or clear them. The database rejects SQL/JSON null and non-object values.

Rollback locks the user table and refuses to discard configured ceilings. Once all overrides are empty, rollback removes only the new column and constraint.

Stacked directly on #1773 (`feat/execution-limit-policy`). Focused replacement for the account-storage portion of #1745/#1754; tracked in #1732.

Validation: 40 account/policy tests pass. On a fresh local database, an existing account received the empty default; a configured ceiling survived a refused rollback; empty rollback and reapply preserved the account. Full precommit passed: 4,714 tests and 6 doctests, zero failures (2 skipped, 7 excluded).

Storage only: admission does not read this field yet. Leave overrides empty until current-ceiling admission and runtime enforcement are integrated. No operator endpoint, host setting, allowance creation or runtime control is enabled; a future write context must authorize the operator and audit changes.

### DIFF
--- a/apps/fountain/lib/fountain/accounts/user.ex
+++ b/apps/fountain/lib/fountain/accounts/user.ex
@@ -26,6 +26,10 @@ defmodule Fountain.Accounts.User do
     # `Fountain.Credits`, in the same transaction as the ledger row; never
     # cast from user input.
     field :credit_balance_cents, :integer, default: 0
+    # Operator-owned per-turn ceiling storage. Admission does not consume this
+    # field yet; leave it empty until ceiling enforcement is integrated.
+    # Never cast it through registration, OAuth, principal or profile changes.
+    field :execution_limits, :map, default: %{}
     field :role, :string, default: "user"
     # A claimable principal (ADR 0044): a tenant with no identity, opened by a
     # trusted application before its visitor has an account and claimed later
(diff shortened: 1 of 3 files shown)


BinaryBourbon (approved):
Approving.

Storage-only and genuinely inert, which the moduledoc-level comment on the field states plainly. The important property is that the ceiling can only be written through `execution_limits_changeset/2` — the test exercising `registration_changeset`, `oauth_registration_changeset`, `principal_changeset`, `theme_changeset` and `preferences_changeset` and asserting the field is untouched is exactly the coverage that keeps it that way when someone later adds a field to one of those casts.

Defending at both layers is the right instinct: the changeset for shape, and `jsonb_typeof(execution_limits) = 'object'` plus `NOT NULL` for anything that reaches the column another way. The test that drives `'null'`, `'[]'`, `'1'` and `'"value"'` through raw SQL and asserts `:check_violation` versus `:not_null_violation` proves the constraint is actually doing work rather than being decorative — JSON null slipping past a `NOT NULL` column is precisely the trap #1776 and #1782 are also defending against, so having the database refuse it here too is worth it.

Reusing `Conversations.ExecutionLimits` from `Accounts.User` is a cross-context reference, but it is a pure policy module with no repo access, and duplicating the validator would be worse.

Same note as #1775 on the raising `down`: no other migration here does that. Fine while nothing writes the field; remember it exists when something does.

PR #1754, opened 2026-09-08
Draft reference: execution limits and lifecycle investigation

jhgaylor:
Remaining extraction tracker: [Fountain #1864](https://github.com/managoat/fountain/issues/1864) — PR boundaries, sequencing and acceptance criteria.

Eight small replacements from [Fountain #1754](https://github.com/managoat/fountain/pull/1754) are published:
- [#1796](https://github.com/managoat/fountain/pull/1796): isolate global pricing tests (1 file).
- [#1797](https://github.com/managoat/fountain/pull/1797): delete only the owned broker session (4 files).
- [#1818](https://github.com/managoat/fountain/pull/1818): share the image byte limit (2 files).
- [#1820](https://github.com/managoat/fountain/pull/1820): record trusted provider identity (6 files; recording only).
- [#1826](https://github.com/managoat/fountain/pull/1826): require committed state before launching workers (2 files).
- [#1827](https://github.com/managoat/fountain/pull/1827): revoke only the failed provision's token (4 files).
- [#1831](https://github.com/managoat/fountain/pull/1831): preserve replacement/completed/cancelled state after a stale startup error (2 files).
- [#1840](https://github.com/managoat/fountain/pull/1840): validate opening input before allocation (5 files).

All eight pass full local precommit and CI and are ready for review. Another agent owns merges after review and CI.

Existing reset PR [#1768](https://github.com/managoat/fountain/pull/1768) is also updated and ready: an enclosing-transaction guard adds 18 lines across two existing files. Its 62 focused tests and full precommit (4,892 tests +6 doctests) pass; the new regression fails on the old handler. [Current-base CI](https://github.com/managoat/fountain/actions/runs/34481301915) passes on `da12a039`. The base change cancelled the superseded push run before tests.

This extraction pass is complete: eight new focused PRs plus the reset update. The residual 211-path audit against main `bed264af` found no additional standalone small cuts worth publishing. Remaining work needs coherent protocol replacements:
- Provider create intent/results with retained ownership and capacity; then confirmed deletion and recovery. Account deletion and billing must retain uncertain operations.
- Committed holder transfer and actor launch ownership, with startup/teardown/recovery tests. Preserve [#1801](https://github.com/managoat/fountain/pull/1801)'s merged identity filter; a waiting helper or replacement callback alone omits the required transfer guarantee.
- Durable prompt persistence, admission, claim and delivery through an actual worker, including HTTP outcomes and replay tests.

[...]

PR #1785, opened 2026-09-10
fix: check account ceilings before launch admission

jhgaylor:
Launches ignored stored account ceilings when callers omitted `execution_limits`. Resolve each request against a fresh, owner-scoped account read before launch, sandbox attachment or channel changes. Omitted/null/empty requests inherit the ceiling; wider requests return `422 execution_limits_widen`. Malformed policy fails without exposing its contents, and unsupported controls remain refused.

Stacked on #1784. Reuses #1774's requested-limit preflight and its nine tests; the remaining nine regressions cover account inheritance, changed policy and tenant isolation. Five files, +289/-3 relative to #1784.

Validation: eight reproduced account-policy regressions; 187 related tests passed. Full precommit passed 4,732 tests and 6 doctests with zero failures. [CI passed](https://github.com/BinaryBourbon/fountain/actions/runs/34434666719). Documentation reports reviewed: destink clean; docs-style findings are in unchanged files, and Vale suggestions retain precise API terminology.

This is read-only preflight, not an atomic reservation against concurrent policy changes. Keep account overrides empty until later-turn/recovery checks and runtime enforcement are integrated. Host ceilings and initial allowance persistence remain separate. No runtime control, operator endpoint or deployment is enabled.

Extraction tracking: #1745, #1754; acceptance remains #1732.

### DIFF
--- a/apps/fountain/lib/fountain/accounts/user.ex
+++ b/apps/fountain/lib/fountain/accounts/user.ex
@@ -26,8 +26,8 @@ defmodule Fountain.Accounts.User do
     # `Fountain.Credits`, in the same transaction as the ledger row; never
     # cast from user input.
     field :credit_balance_cents, :integer, default: 0
-    # Operator-owned per-turn ceiling storage. Admission does not consume this
-    # field yet; leave it empty until ceiling enforcement is integrated.
+    # Operator-owned per-turn ceilings, read by launch preflight. Later-turn and
+    # recovery ceiling checks remain unwired; leave overrides empty until integrated.
     # Never cast it through registration, OAuth, principal or profile changes.
     field :execution_limits, :map, default: %{}
     field :role, :string, default: "user"
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1792,7 +1792,7 @@ defmodule Fountain.Conversations do
   # ── high-level lifecycle ──────────────────────────────────────────────────────────
 
   alias Fountain.Agents
-  alias Fountain.Conversations.ConversationServer
+  alias Fountain.Conversations.{ConversationServer, ExecutionLimits}
 
   @doc """
   Like `start_conversation/2`, but a conversation already bound to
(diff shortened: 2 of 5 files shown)


BinaryBourbon (approved):
Approving.

Reading the ceiling fresh from `Accounts.get_user/1` on every launch, rather than from anything the request carries or a cached copy, is the important property — and "account ceilings are reread on each resume" pins it against a future optimization that caches the user.

`ExecutionLimits.resolve(nil, ceiling, request)` gives the inheritance semantics the description claims: an omitted, `nil` or `%{}` request merges over the ceiling and keeps it, so omission cannot be used to escape a configured limit. That is the actual bug and the four-path matrix covers it.

Separating `execution_limits_widen` into its own 422 with `errors.execution_limits` is worth the extra clause. "Invalid" and "you asked for more than you are allowed" send a caller to different places, and clamping silently would hand back a tighter policy than requested with no way to learn the ask was wrong — the same reasoning `permission_policy_widens` records just above it.

The corrupt-account-policy case falling to `{:error, {:execution_limits_invalid, "object_required"}}` rather than crashing is right, and the test asserts the private key and value never reach the response.

Two notes, neither blocking:

The `"account_execution_limits" => %{}` key the inheritance tests put in the request params is not read by anything, so that part of the assertion is vacuous. I take the intent — proving a request-supplied field cannot shadow the stored ceiling — but as written it only proves an unrecognized key is ignored. A comment saying that is what it is for would stop the next reader hunting for the handler.

`check_execution_limits/2` calls `Accounts.get_user/1` and `Billing.check_spend/1` a few lines later reads the same user again. Not worth restructuring the `with` chain for, just noting it.

Merge-order caveat, as on #1774: this branch carries a cherry-picked copy of #1774's commit as `d4ea7e76`. `check_execution_limits/2` here strictly supersedes the `/1` version there. Land #1774 first and rebase this chain, or land this and close #1774 — merging both will conflict in `conversations.ex`, `fallback_controller.ex` and add/add on `execution_limit_admission_test.exs`.

PR #1786, opened 2026-09-10
feat: create initial execution allowances by owner

jhgaylor:
Initial execution allowances only had an unscoped storage changeset. Add `Conversations.create_execution_allowance/3` to lock the conversation's ownership through insertion and record a successful creation in the audit trail. Missing and foreign conversations return the same error before validation. Duplicate inserts cannot replace a saved policy, revision or timestamp.

Stacked on #1783. Four files, +245/-1; 44 added context lines, with tests and storage documentation making up the rest.

Validation: 125 related tests passed, including observed PostgreSQL lock waits for a competing insert and an ownership change. Tests also cover redacted audit metadata, invalid inputs, unchanged active work and refusal of unsupported execution. Full precommit passed 4,786 tests and 6 doctests with zero failures. [CI passed](https://github.com/BinaryBourbon/fountain/actions/runs/34436161379).

This is internal persistence, with no launch or HTTP caller. Admission must still resolve trusted current ceilings and verify runtime support before using it. Account-ceiling preflight remains on sibling #1785; host settings and complete runtime enforcement remain separate.

Extraction: #1745, #1754. Live acceptance remains #1732.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1152,6 +1152,50 @@ defmodule Fountain.Conversations do
   end
 
   @doc """
+  Save an initial resolved allowance once, scoped to its conversation owner.
+
+  Internal persistence only: the caller must resolve trusted current ceilings
+  and prove runtime support before admission. This function does not admit work
+  or reset an active turn. No launch or HTTP path calls it yet. A duplicate
+  fails without replacing the saved policy; use `narrow_execution_allowance/3`
+  for subsequent changes. Ownership stays locked through insertion.
+  """
+  def create_execution_allowance(conversation_id, user_id, resolved_limits, opts \\ []) do
+    result =
+      Repo.transaction(fn ->
+        Repo.one(
+          from c in Conversation,
+            where: c.id == ^conversation_id and c.user_id == ^user_id,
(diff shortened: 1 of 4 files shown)


BinaryBourbon (approved):
Approving.

`Conversation FOR SHARE` before the insert is the right lock: it is strong enough to stabilize ownership and compatible with the `FOR KEY SHARE` the allowance's own foreign key needs, so the lock does not fight the insert it is protecting.

The `:ownership` race is the case I would not have thought to write and it is the valuable one — it proves the conversation is re-read *under* the lock rather than trusted from before the wait, so a row that changed owner while the writer was blocked produces `{:error, :not_found}` and no audit row. The `:duplicate` half proves the PK constraint still wins after a lock wait rather than the transaction retrying blindly.

Audit is outside the transaction, metadata is control names only, and `Enum.filter(ExecutionLimits.keys(), ...)` gives a stable order so the assertion is not seed-dependent.

Returning the same `{:error, :not_found}` for missing and foreign conversations before any request validation is the correct disclosure boundary — a caller cannot use a malformed request to learn whether an id exists.

Adding both this and `narrow_execution_allowance/3` to `audit_guardrail_test.exs` rather than the exclusion list is the right choice.

PR #1787, opened 2026-09-10
feat: validate host execution ceilings before launch

jhgaylor:
Host execution ceilings had no configuration or launch reader. Add `FOUNTAIN_EXECUTION_LIMITS`, parsed at boot with the existing typed validator. Invalid input refuses startup without echoing its contents. Launch preflight now resolves requests against both host and current account ceilings; omission cannot remove a limit, and requests cannot widen the stricter ceiling.

Stacked on #1785. Six files, +166/-10. The setting defaults to an empty map; the test runtime ignores operator shell policy, and tests that change application configuration run serially.

Validation: all 11 new tests failed against the parent; 65 related tests now pass, covering boot parsing, four launch paths, inherited controls, widening, redaction and configuration guardrails. Full precommit passed 4,743 tests and 6 doctests with zero failures. [CI passed](https://github.com/BinaryBourbon/fountain/actions/runs/34440792649). Prose reports reviewed: destink clean; docs-style findings are in unchanged files; Vale advice retains precise API terminology.

Keep host/account ceilings empty until runtime enforcement and later-turn/recovery ceiling checks are integrated. Nonempty effective limits still return `422 execution_limits_unsupported`. This is preflight, not an atomic reservation or aggregate spend cap. Initial allowance persistence remains on sibling #1786 and is not wired into launch admission.

Extraction: #1745, #1754. Live acceptance remains #1732.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -329,6 +329,15 @@ STRIPE_WEBHOOK_SECRET=
 # uses it and no sandbox meter reports it. An abuse ceiling, not an allowance.
 # BUZZ_IDENTITY_CEILING=10
 
+# Per-turn host execution ceiling, as JSON: wall_time_seconds and
+# max_model_turns are positive integers, max_estimated_cost_usd is a positive
+# number. Read at boot; invalid input refuses startup. A launch request
+# inherits the stricter of this and the account ceiling, and cannot widen it.
+# Leave unset: nothing enforces these controls yet, so a nonempty effective
+# ceiling refuses the launch with 422 execution_limits_unsupported. This is a
+# per-turn bound, not an aggregate spend cap.
+# FOUNTAIN_EXECUTION_LIMITS={"wall_time_seconds":3600,"max_model_turns":50}
+
 # Prepaid credits (ADR 0030): what the customer pays, in whole cents. Distinct
 # from the provider costs above. The turn-hour price defaults to 25 and is a
 # placeholder until a provider invoice has been reconciled. The four comms
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2038,13 +2038,15 @@ defmodule Fountain.Conversations do
   defp check_execution_limits(user_id, request) do
     # Ownership: each caller just fetched the agent by this authenticated user.
     # Read the current account policy, never a request-supplied or cached map.
-    case Fountain.Accounts.get_user(user_id) do
-      %Fountain.Accounts.User{execution_limits: ceiling} when is_map(ceiling) ->
-        with {:ok, limits} <- ExecutionLimits.resolve(nil, ceiling, request) do
+    case {Fountain.Accounts.get_user(user_id),
+          Application.get_env(:fountain, :execution_limit_ceiling, %{})} do
+      {%Fountain.Accounts.User{execution_limits: ceiling}, host}
+      when is_map(ceiling) and is_map(host) ->
+        with {:ok, limits} <- ExecutionLimits.resolve(host, ceiling, request) do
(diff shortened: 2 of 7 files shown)


BinaryBourbon (changes requested):
The mechanism is right and I have no concerns with the code. One small gap to close before merge.

**`FOUNTAIN_EXECUTION_LIMITS` is missing from `.env.example`.**

`docs/configuration.md` gets a good row, but `.env.example` is the other half of that contract — `CLAUDE.md` points at it as "the full list", and every comparable knob has a commented entry there (`SANDBOX_FLEET_CEILING=20` at line 323, `CREDIT_OPENING_CENTS=500` at 337, `BROKER_LISTEN_PORT` at 76). A self-hoster copying `.env.example` will not learn this variable exists. It is a one-line addition next to the other sandbox/ceiling entries, and it is the kind of thing that never gets added once the PR that introduced the variable is merged.

Everything else looks good:

Refusing boot on invalid input is the correct failure mode — a host ceiling that silently parses to `%{}` because of a typo would read as "no ceiling configured" everywhere it is displayed, which is the failure this stack keeps designing against. `raise "FOUNTAIN_EXECUTION_LIMITS is invalid: #{field}"` names the variable without echoing the value, and the test asserts both halves.

`if env == :test, do: %{}` is worth having. Without it an operator's exported shell variable would leak into every local test run and produce refusals nobody could explain from the diff. The `read("private-value", :test) == %{}` case pins it.

Calling a `Fountain.*` module from `config/runtime.exs` matches existing precedent there (`Fountain.PublicUrl.absolute/2`, `Fountain.Extensions.validate!/0`), and `from_json_env/1` is pure, so the config-provider boot phase is fine.

Flipping `execution_limit_admission_test.exs` to `async: false` is correct rather than a concession — the module writes application env, and doing that from an async module is a known flake source in this repo. The `Application.fetch_env/2` + `on_exit` restore handles the "not set at all" case properly.

The `{2, 10}` / `{10, 2}` pair in the widening test is the right shape: it proves the *stricter* of host and account wins in both directions rather than one of them always dominating.

Re-request review once `.env.example` has the entry and I will approve.

BinaryBourbon:

Base moved to `main`. **My changes-requested review stands**: `FOUNTAIN_EXECUTION_LIMITS` still needs its `.env.example` entry.

jhgaylor:
Addressed the review.

Added `FOUNTAIN_EXECUTION_LIMITS` to `.env.example`, next to the other ceiling knobs, with the example commented out so the default stays empty. The comment says what the three fields accept, that invalid input refuses boot, that a request inherits the stricter of host and account, and that it should stay unset until enforcement exists — since a nonempty effective ceiling currently refuses the launch.

No code change; `docs/configuration.md` already had the row.

BinaryBourbon (approved):
Approving. The `.env.example` gap is closed and that was the only thing holding this.

The entry does more than name the variable: it says what the three fields accept, that invalid input refuses boot rather than degrading to empty, that a request inherits the stricter of host and account, and that it should stay unset until enforcement exists. Someone copying `.env.example` now learns all of that without going to `docs/configuration.md`. The example value stays commented, so the default is still empty — which matters here, because an uncommented one would refuse every launch on this deployment.

Everything I said about the mechanism last time stands. Refusing boot on invalid input is the right failure mode for a ceiling — one that silently parsed to `%{}` would read as "no ceiling configured" everywhere it is displayed, which is the failure this whole stack keeps designing against. `if env == :test, do: %{}` earns its place: without it an exported shell variable would leak into local test runs and produce refusals nobody could explain from the diff.

Disclosure: I wrote the `.env.example` commit myself in response to my own review, so this approval covers a one-line documentation change I authored. The code under it is unchanged from what I reviewed before.

PR #1788, opened 2026-09-10
chore: join execution admission prerequisites

jhgaylor:
Join #1787's host/account launch checks with #1786's saved-allowance storage and recovery guards for the attachment-persistence follow-up in #1789.

Base: #1786. This preserves both branches' commits. The diff is 10 files, +590, mostly existing tests from #1784/#1785/#1787. Resolution keeps the widening HTTP error and both channel guards, and removes a duplicate alias. It adds no new runtime capability or launch writer.

Validation: 81 combined admission/storage/configuration tests passed in a fresh database with both migrations. The first test attempt stopped before tests because the new database was unmigrated; migration resolved it. Full precommit passed 4,821 tests and 6 doctests with zero failures. [CI passed](https://github.com/BinaryBourbon/fountain/actions/runs/34443300234). The inherited documentation retains the previously reviewed prose reports.

Keep host/account ceilings empty until later-turn/recovery ceiling checks and runtime enforcement integrate. This preflight is not an atomic policy reservation or aggregate spend cap. Existing activation and live-acceptance holds remain.

Extraction: #1745, #1754. Acceptance: #1732.

### DIFF
--- a/apps/fountain/lib/fountain/accounts/user.ex
+++ b/apps/fountain/lib/fountain/accounts/user.ex
@@ -26,8 +26,8 @@ defmodule Fountain.Accounts.User do
     # `Fountain.Credits`, in the same transaction as the ledger row; never
     # cast from user input.
     field :credit_balance_cents, :integer, default: 0
-    # Operator-owned per-turn ceiling storage. Admission does not consume this
-    # field yet; leave it empty until ceiling enforcement is integrated.
+    # Operator-owned per-turn ceilings, read by launch preflight. Later-turn and
+    # recovery ceiling checks remain unwired; leave overrides empty until integrated.
     # Never cast it through registration, OAuth, principal or profile changes.
     field :execution_limits, :map, default: %{}
     field :role, :string, default: "user"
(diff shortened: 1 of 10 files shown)


BinaryBourbon (approved):
Approving.

I verified the merge resolution rather than taking the description's word for it. On the resulting tree:

- `conversations.ex` keeps both guards and they are distinct things, which is the resolution that was easy to get wrong: `check_execution_limits/2` on the three admission `with` chains (lines 1989, 2141, 2680) for the *requested* limits, and `_unsafe_check_saved_execution_allowance/1` on the channel-resume branch (1999), turn admission (1132) and the wake path (3151) for the *saved* allowance. Dropping either would have looked green.
- The duplicate alias is gone: line 15 carries `{ExecutionAllowance, ExecutionLimits}` and the lifecycle section's `alias ... ConversationServer` is untouched.
- `fallback_controller.ex` has exactly one clause each for `execution_limits_invalid`, `execution_limits_widen` and `execution_limits_unsupported`, no duplicates.

Diff is `+590/-0` against #1786 with no deletions, which is what a clean join of two additive stacks should look like.

Doing the join as an explicit PR rather than letting #1789 carry an implicit merge is the right call — it makes the resolution reviewable on its own instead of hiding it inside a feature diff.

BinaryBourbon:
Closing: this PR existed to join the saved-allowance and account-ceiling stacks, and #1792 merged both to `main` as `e78a417e`. The join is moot.

Its conflict resolution was not wasted — I used it as the reference when resolving #1792's merge, and verified #1792's tree equals this one minus #1787's commit plus #1782's, with both deltas fully accounted for.

PR #1789, opened 2026-09-10
feat: persist attachment allowances before prompt delivery

jhgaylor:
Attachments could deliver a prompt without a saved initial allowance, and an account ceiling changed after early preflight was not rechecked. Insert the attached conversation and its resolved allowance in one transaction, holding the account, agent and sandbox ownership stable. Recheck sandbox identity and current host/account limits before either insert. A refused allowance rolls back the conversation.

Activation analytics, creation audits and prompt delivery occur after commit. Existing conversations and their allowances are preserved. Unsupported controls remain refused.

Stacked on #1788, which explicitly joins the two prerequisite stacks. Five files, +265/-34. Validation: five parent HTTP regressions; 147 related tests passed, including a real PostgreSQL lock wait where a concurrent account-policy change causes refusal without creation records or audit. Full precommit passed 4,826 tests and 6 doctests with zero failures. [CI passed](https://github.com/BinaryBourbon/fountain/actions/runs/34444174037). Prose reports reviewed: destink clean; docs-style findings are in unchanged files; Vale suggestions retain precise API terminology.

This covers attachment to an existing sandbox, including reuse of an existing home. Fresh worker creation still needs atomic reservation and allowance persistence. Later-turn/recovery current-ceiling checks and runtime enforcement remain required; keep configured ceilings empty until integrated. This does not guarantee an atomic provider operation or enforce aggregate spend.

Extraction: #1745, #1754. Live acceptance: #1732.


BinaryBourbon (changes requested):
The transactional design is good — committing the conversation and its allowance together, with analytics, audit and prompt delivery after commit, is the right shape, and the rollback test proves it. Two things to address first.

**1. `create_conversation/1`'s comment now asserts an invariant that is false.**

The attach path no longer goes through `create_conversation/1`; it inserts the changeset directly and calls `Fountain.Activation.conversation_created/1` itself. But the comment inside `create_conversation/1` still reads:

> Both create paths go through this write, so the funnel's third step cannot be missed by a new door — and it does not depend on the landing page still being open.

That was a claim about a chokepoint, and after this PR the chokepoint does not exist. On the tip of the stack `create_conversation/1` has exactly one caller left (`Fountain.Team`), and `Activation.conversation_created/1` has three call sites. Nothing enforces that a fourth create path remembers to call it — `activation_test.exs` only unit-tests the function, so a new door that misses it fails no test.

This repo has already lost call sites this way once (the system prompt, unwired twice by squash merges). Either restore the chokepoint by having `create_attached_conversation/3` insert through `create_conversation/1` inside the transaction and move only the Activation call out, or delete the claim from the comment and add a guard test that every conversation-insert site fires the activation seam. What should not ship is a comment telling the next contributor the property is guaranteed by construction when it is not.

**2. `users` row lock inside admission.**

`create_attached_conversation/3` takes `SELECT ... FROM users ... FOR SHARE`, which conflicts with the `FOR UPDATE` that `Credits.insert_and_move/3` (`ee/lib/fountain/credits.ex:581`) holds across a ledger insert plus `consume_lots/3` plus the balance update. So an attach can now block behind an in-flight credit posting for the same tenant.

Here that is bounded to one tenant and I would not block on it alone. It becomes a real problem in #1790, where the same lock moves inside the global fleet advisory lock — see my review there. Worth deciding on both together: if the ceiling read does not need `FOR SHARE` (the docs already say this is "not an atomic reservation against later policy changes"), dropping it solves both.

The rest reads well:

Re-running `check_attachable/4` inside the transaction against re-locked `agent` and `sandbox` rows closes the check-then-act window on sandbox identity, and the "ceiling changed after early preflight" test — stubbing `RuntimeDispatch.for_agent/1` to mutate the ceiling mid-`with` — is a nice way to force the interleaving deterministically without a second connection.

Extracting `record_execution_allowance_created/3` and having it read `allowance.conversation_id` rather than a passed-in id removes a way for the two to drift.

`resolve_admission_limits/2` returning the resolved map, with `check_execution_limits/2` as a thin `:ok` wrapper over it, is the right factoring — the preflight and the persisted value now come from one function, so they cannot disagree.

The `send_prompt` stub asserting the allowance row is already readable is a good way to state "before prompt delivery" as an executable claim rather than a comment.

BinaryBourbon:

jhgaylor:
Addressed both points.

**The stale invariant.** You were right that it was worse than a comment: `Fountain.Activation`'s moduledoc made the same claim, so there were two places telling the next contributor the funnel step was guaranteed by construction when it was not.

Rather than only correcting the prose, split the seam in two: `insert_conversation_row/1` is now the one place a conversation row is written, and `after_conversation_created/1` the one place the funnel step fires. They are separate on purpose — a caller inside a transaction has to report only after it commits, or a rolled-back write claims a request that never happened. Both comments now describe what is true.

The guarantee is `conversation_creation_seam_test.exs`, which drives every door and fails if one stops reporting. It has no exclusion list. I verified it by breaking each door in turn and watching the right test fail, and there is a case asserting a rolled-back write reports nothing.

**The `users` lock.** Dropped. It was only ever an ownership recheck (`select: u.id`), and `users` is the row `Credits.insert_and_move/3` holds `FOR UPDATE` across a ledger insert, lot consumption and the balance move. The agent and sandbox locks stay: those guard a real check-then-act on attachability and contend with nothing outside the tenant.

The race test that asserted a lock wait now asserts what actually matters, still on two real connections: a ceiling committed after the early preflight is honoured by the in-transaction recheck.

BinaryBourbon (approved):
Approving. Both points are addressed, and the first one better than I asked for.

**The invariant.** I asked for the claim to be corrected or the chokepoint restored. What landed does the more useful third thing: it separates the two obligations that were tangled together. `insert_conversation_row/1` is the one place a row is written; `after_conversation_created/1` is the one place the funnel step fires; and they are deliberately not nested, because a caller inside a transaction must report only after it commits — otherwise a rolled-back write claims a request that never happened. That distinction is exactly why the original chokepoint could not simply be restored, and now it is written down instead of being an accident of where the call sat.

It also caught the second copy. `Fountain.Activation`'s moduledoc made the same "instrumented by construction" claim, so there were two places telling the next contributor the guarantee held. Both now describe what is true.

**The guard.** `conversation_creation_seam_test.exs` drives every door and has no exclusion list, and the header tells someone who trips it that the fix is a call to the seam rather than an entry in a list. The `reject/1` case covering a rolled-back write is the right companion — it pins the reason the two seams are separate, not just that they both exist. I am satisfied it guards rather than merely passes: the description reports each door was broken in turn and the matching test failed, and the shape of the assertion (`assert_received {:reported, ^id}` against a per-door pin) makes a false pass hard.

One weakness worth naming rather than blocking on: `@doors` is a hand-maintained list, so a new create path is caught only once someone adds it there. Deriving it would need AST work for little gain, and the failure mode is benign — a missing door means no coverage, not a false green. The header comment is the mitigation and it is the right one.

**The lock.** Dropped, with the reasoning recorded at the point of edit. Keeping the `agents` and `sandbox` locks is the right line: those guard a genuine check-then-act on attachability and contend with nothing outside the tenant, whereas `users` is the row every credit posting serialises on.

The reworked race test still uses two real connections and still proves something that can regress — a ceiling committed after the early preflight is honoured — instead of asserting a lock ordering that no longer exists.

Disclosure: I wrote the fix commit after my own changes-requested review, so this approves my own code. The PR's original commit is unchanged.

PR #1790, opened 2026-09-10
Persist fresh launch allowances atomically

jhgaylor:
A failed fresh conversation insert could leave a reserved sandbox, and launches had no saved initial execution allowance. Commit the sandbox, conversation and resolved allowance together under the existing quota locks. Recheck account/agent ownership and current ceilings before inserting; start the worker and emit creation events after commit.

Failed inserts roll everything back. A worker startup failure still retains the admitted allowance with its failed conversation and sandbox. Unsupported controls remain refused.

Stacked on #1789; five files, +190/-40. Four reproduced parent failures; 229 related tests passed, including a real PostgreSQL lock wait on an account-ceiling update. Full precommit passed 4,832 tests and 6 doctests with zero failures. Channel rotation retains the old allowance and gives the new conversation its own policy. Staged secret scan passed; prose reports reviewed. [CI passed](https://github.com/BinaryBourbon/fountain/actions/runs/34447732132). Later-turn/recovery checks against current ceilings and runtime enforcement remain separate.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -911,7 +911,8 @@ defmodule Fountain.Conversations do
   # *not* inside `insert_conversation_row/1`: a caller in a transaction must
   # fire it after that transaction commits, so a rolled-back write reports no
   # request. Every door that inserts a conversation calls it exactly once —
-  # `create_conversation/1`, `create_attached_conversation/3` — and
+  # `create_conversation/1`, `create_attached_conversation/3`,
+  # `reserve_initial_conversation/3` — and
   # `conversation_creation_seam_test.exs` drives each of them and fails if one
   # stops firing.
   defp after_conversation_created(%Conversation{} = conv) do
@@ -2235,7 +2236,7 @@ defmodule Fountain.Conversations do
              },
              attrs["execution_limits"]
            ) do
-      Fountain.Activation.conversation_created(conv)
+      after_conversation_created(conv)
       record_execution_allowance_created(allowance, user_id, opts)
 
       # Recorded here rather than in either branch below: both of them return
@@ -2344,27 +2345,33 @@ defmodule Fountain.Conversations do
   # reservation. Analytics, audit, worker startup and prompts run after commit.
   defp reserve_initial_conversation(sandbox_attrs, conversation_attrs, request) do
     Fountain.Quotas.with_sandbox_reservation(conversation_attrs.user_id, fn ->
+      # Nothing in here may take a row lock. `with_sandbox_reservation/3` holds
+      # `pg_advisory_xact_lock(@fleet_lock_key)` — one lock shared by every
+      # tenant, the thing that enforces SANDBOX_FLEET_CEILING — for the whole
+      # of this function. A `SELECT ... FOR SHARE` on `users` conflicts with
+      # the `FOR UPDATE` that `Credits.insert_and_move/3` holds across a ledger
+      # insert, lot consumption and the balance move, so one tenant's credit
+      # posting would stall provisioning for every other tenant. `agents` is
+      # the same story against a `last_used_at` stamp. These stay plain
+      # ownership rechecks: MVCC reads never block, the ceiling below is read
+      # the same way, and the inserts' foreign keys enforce integrity.
       Repo.one(
         from u in Fountain.Accounts.User,
           where: u.id == ^conversation_attrs.user_id,
-          select: u.id,
-          lock: "FOR SHARE"
+          select: u.id
       ) || Repo.rollback(:not_found)
 
       Repo.one(
         from a in Agents.Agent,
           where:
             a.id == ^conversation_attrs.agent_id and a.user_id == ^conversation_attrs.user_id,
-          select: a.id,
-          lock: "FOR SHARE"
+          select: a.id
       ) || Repo.rollback(:not_found)
 
       with {:ok, limits} <- resolve_admission_limits(conversation_attrs.user_id, request),
            {:ok, sandbox} <- create_sandbox(sandbox_attrs),
            {:ok, conv} <-
-             %Conversation{}
-             |> Conversation.changeset(Map.put(conversation_attrs, :sandbox_id, sandbox.id))
-             |> Repo.insert(),
+             insert_conversation_row(Map.put(conversation_attrs, :sandbox_id, sandbox.id)),
            {:ok, allowance} <-
              conv.id |> ExecutionAllowance.new_changeset(limits) |> Repo.insert() do
         {:ok, {sandbox, conv, allowance}}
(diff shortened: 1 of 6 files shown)


BinaryBourbon (changes requested):
Folding the sandbox reservation, the conversation and the allowance into one transaction is clearly right, and the "failed fresh conversation validation rolls back its sandbox reservation" test fixes a leak worth fixing on its own. But the way the ceiling recheck is done puts a blocking row lock inside a process-wide lock, and I do not think that trade is worth it.

**Blocker: `users`/`agents` row locks now run inside the global fleet advisory lock.**

`Quotas.with_sandbox_reservation/3` opens with

```elixir
Repo.query!("SELECT pg_advisory_xact_lock($1, $2)", [@lock_namespace, @fleet_lock_key])
```

which is a single lock shared by **every tenant** — it is what enforces `SANDBOX_FLEET_CEILING`, and `CLAUDE.md` calls out that it is taken before the per-user one. Until now the closure it wraps was just `create_sandbox/1`: inserts and reads, nothing that waits on another transaction. The three checks above it (`check_fleet_ceiling/1`, `Credits.gate/1`, `check_sandbox_quota/2`) are all unlocked reads.

`reserve_initial_conversation/3` now adds, inside that closure:

```elixir
Repo.one(from u in Fountain.Accounts.User, where: u.id == ^..., select: u.id, lock: "FOR SHARE")
Repo.one(from a in Agents.Agent,           where: a.id == ^..., lock: "FOR SHARE")
```

`FOR SHARE` conflicts with the `FOR NO KEY UPDATE` a plain `UPDATE` takes, so either of these can now block. The concrete case is `Credits.insert_and_move/3` (`ee/lib/fountain/credits.ex:578-615`), which takes the user row `FOR UPDATE` and holds it across a ledger insert, `consume_lots/3` — potentially many `credit_ledger` rows for an account with a long grant history — and the balance `update_all`. Any turn burn, pack purchase, grant, expiry or refund clawback for that tenant is such a transaction, and `CreditPricer` runs them in batches seven days back.

The result: while one tenant's credit posting is in flight, a launch for that tenant blocks on `users FOR SHARE` **while holding the fleet lock**, so every other tenant's launch queues behind it. Today the fleet lock is only ever held for the duration of local inserts; after this it can be held for the duration of an unrelated transaction owned by a different subsystem. That is a cross-tenant availability regression on the provisioning path, and it will present as "sandbox creation is slow for everyone" with nothing in the launch path to point at.

The `agents` lock has the same shape against agent updates, including the `last_used_at` stamp.

What I would do instead, in rough order of preference:

1. Drop `lock: "FOR SHARE"` from both. The reads are already inside the transaction, so they see a consistent snapshot for the rest of it, and the PR's own documentation says this is "not an atomic reservation against later policy changes". The lock is buying atomicity against a rare operator action (an admin editing a ceiling) at the cost of a global stall on a common one.
2. If the lock is genuinely wanted, take it *before* `with_sandbox_reservation/3` — resolve the limits in the outer transaction-free scope and pass the resolved map in, so any wait happens outside the fleet lock.
3. If neither is acceptable, at minimum set a short `lock_timeout` for the locked statements so a blocked launch fails fast instead of holding the fleet lock open.

Whichever way it goes, the same call in `create_attached_conversation/3` (#1789) should move with it — that one is not under the fleet lock, so it is much less severe, but it is the same lock against the same writer.

**Also carried from #1789:** `create_conversation/1`'s "Both create paths go through this write" comment is now false for both create paths, and `Activation.conversation_created/1` has three unguarded call sites. Details in my review there.

The rest is good and I would like to approve once the lock question is settled:

Returning `{sandbox, conv, allowance}` out of the closure and firing Activation plus the allowance audit after `with_sandbox_reservation/3` returns is correct — nothing side-effecting happens inside the transaction.

`stub(Horde.DynamicSupervisor, :start_child, ...)` asserting `refute Repo.in_transaction?()` and reading the allowance row back is a good executable statement of "committed before the worker starts", much better than asserting on ordering indirectly.

"worker startup failure retains the admitted allowance with its failed rows" is the right call on semantics — the reservation is spent and the rows exist, so the allowance belongs with them.

The `%Conversation{} |> changeset |> insert` inside the closure correctly puts `sandbox_id` from the freshly created sandbox rather than threading it through the attrs map, so the two cannot drift.

BinaryBourbon:

jhgaylor:
Addressed. You were right, and I took option 1.

Both `FOR SHARE` locks are gone from inside `with_sandbox_reservation/3`. I traced the case you named and it holds: `Credits.insert_and_move/3` (ee/lib/fountain/credits.ex:578) takes the user row `FOR UPDATE` and holds it across a ledger insert, `consume_lots/3` and the balance `update_all` — and CreditPricer runs those in batches seven days back. `FOR SHARE` conflicts with that, so one tenant's credit posting could park provisioning for every other tenant, with nothing in the launch path to point at.

They were only ownership rechecks (`select: u.id` / `select: a.id`). The ceiling was already read separately and unlocked, MVCC reads never block, and the inserts' foreign keys are what enforce integrity. There is a comment on the closure now saying nothing in there may take a row lock, and why — the next person adding a check inside that function needs to know it runs under a process-wide lock.

What is lost is atomicity against a ceiling committing mid-transaction. That is a window the early preflight already leaves wide open, and this PR's own docs call it "not an atomic reservation against later policy changes". The authoritative gate is untouched: turn admission re-reads the saved allowance under row locks on every turn.

So the race test no longer asserts a lock wait — there is none to assert. It now asserts the property that matters, on both `:attach` and `:fresh` with two real connections: a ceiling committed after the early preflight is honoured by the recheck.

Also routed the fresh path through `insert_conversation_row/1` and `after_conversation_created/1`, so the third create door is covered by the seam guard added in #1789.

BinaryBourbon (approved):
Approving. The blocker is resolved, and resolved the right way.

Both `FOR SHARE` locks are gone from inside `with_sandbox_reservation/3`, and the comment left on the closure is the part that will still be doing work in a year: it tells the next person adding a check in there that the function runs under a process-wide lock and why that constrains what they may write. That is the failure mode that produced this finding — the constraint was real but invisible at the point of edit.

Dropping them rather than bounding them is the correct trade. They were `select: u.id` / `select: a.id`, pure ownership rechecks; the ceiling was already read unlocked a few lines down, MVCC reads never block, and the inserts' foreign keys are what actually enforce integrity. What they bought was atomicity against a ceiling committing mid-transaction — a window the early preflight already leaves wide open — and the authoritative gate is untouched: turn admission re-reads the saved allowance under row locks on every turn.

The race test change is the honest one. There is no lock wait left to assert, so asserting one would have been theatre. It now proves the property that survives: on both `:attach` and `:fresh`, across two real connections, a ceiling committed after the early preflight is honoured by the in-transaction recheck. It gets there by holding the admitting process inside a stubbed `RuntimeDispatch.for_agent/1` — which is the seam immediately after the preflight — so the interleaving is deterministic rather than hoped for.

**One thing this does not fix, stated so the fix is not read as more complete than it is.** Inserting a row with a `user_id` foreign key takes `KEY SHARE` on that `users` row, and `Credits.insert_and_move/3` takes it `FOR UPDATE`, which conflicts with `KEY SHARE`. So the sandbox insert inside the fleet lock can still wait on a credit posting — and has been able to since #330, long before this stack. This PR removes the exposure the stack *added*; it does not remove the class. Worth its own issue rather than growing this PR: the shape of a fix is different (move the reservation's writes off the user row, or bound the reservation as a whole), and it wants measurement first.

Disclosure: I wrote the fix commit myself after my own changes-requested review, so this approves my own code. The original PR content is unchanged apart from the two locks, the race test, and routing the fresh path through the seam from #1789.

PR #1793, opened 2026-09-10
fix: keep tenant foreign-key waits outside fleet reservation

jhgaylor:
Fresh launches could still stall every tenant while a credit posting held `users FOR UPDATE`. #1790 removed the explicit `FOR SHARE` from inside `with_sandbox_reservation/3`, but that was only half of it: `create_sandbox/1` and the conversation insert take `KEY SHARE` on `users` through their foreign keys, and `KEY SHARE` conflicts with the `FOR UPDATE` that `Credits.insert_and_move/3` holds across a ledger insert, lot consumption and the balance move. The wait moved from an explicit lock to an implicit one and stayed under the global fleet advisory lock.

Take the account and scoped agent rows `FOR SHARE` in an outer transaction, before the reservation. That does two things at once: it settles the wait before `pg_advisory_xact_lock(@fleet_lock_key)` is taken, and — because `FOR SHARE` is compatible with `KEY SHARE` and is already held by this transaction — it satisfies the foreign-key checks inside, so the inserts cannot block on that row either. Both halves are needed; acquiring the lock is what makes the inserts non-blocking, not just the reordering.

The rotation unbind from #1791 moves out with them. It is the same category of wait, on a row turn admission takes `FOR UPDATE`, and leaving it inside would keep a wait under the fleet lock that this change exists to remove.

## The tradeoff

**This does not remove the wait, it changes who does the waiting.** The outer transaction holds `users FOR SHARE` for its whole life, including the wait for the global fleet lock, and `FOR SHARE` conflicts with `FOR UPDATE`. So a tenant's credit postings now queue behind that tenant's in-flight launch, which may itself be queued behind every other tenant's launch. Turn burns, purchases, grants, expiry and refund clawbacks all post through `Credits.insert_and_move/3`.

A tenant-scoped wait is strictly better than a fleet-wide one, which is why this is the right trade — but it is a real inversion, and someone debugging a slow credit posting should start here. Recorded in the comment on `reserve_initial_conversation/4` as well as here.

## The nested-transaction trap

The nested `Repo.transaction` inside `with_sandbox_reservation/3` joins the outer one rather than opening another, so the sandbox, conversation and allowance still commit or roll back together. The `case result` re-rollback is load-bearing rather than defensive: a nested rollback the outer transaction does not re-raise with its reason reaches the caller as `{:error, :rollback}`. Without it, every refusal `with_sandbox_reservation/3` produces — `:insufficient_credits` (402 with `upgrade_url`), `:fleet_full` (503), a quota changeset (422) — would fall through `FallbackController` as a 500. Verified directly rather than assumed.

## Validation

Two contention tests, for the account row and the agent row: a second tenant's launch commits while the first is still blocked on its own row, on real independent connections with an observed lock wait. Asserting that the blocked launch eventually acquires its lock would prove nothing; the second tenant committing is the property.

Both fail on `997ed65f`. Reverting only the two `lock: "FOR SHARE"` clauses, keeping the restructuring, also fails them — so the locks carry the fix.

Compile with `--warnings-as-errors`, format at both roots, `credo --strict` (6994 mods/funs, no issues), 131 tests across the admission and allowance suites, no leftover fixtures in the database.

Rebased onto `main` after #1787, #1789, #1790 and #1791 landed; it was still based on `feat/fresh-initial-execution-allowance` at the pre-rebase `fc005df5`, which inflated the diff to +882.

Trackers: #1732, #1754.

https://claude.ai/code/session_01EzUPkXKdmnZYeJ1sqjpnp9

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2458,42 +2458,73 @@ defmodule Fountain.Conversations do
     end
   end
 
-  # Keep fleet -> user advisory lock order, then stabilize ownership and
-  # recheck policy. A failed conversation or allowance must release the entire
-  # reservation. Analytics, audit, worker startup and prompts run after commit.
+  # Tenant row waits happen here, before the fleet lock, and the reservation
+  # runs inside them. `with_sandbox_reservation/3` holds
+  # `pg_advisory_xact_lock(@fleet_lock_key)` — one lock shared by every tenant —
+  # so anything that can wait on another transaction must be settled before it
+  # is taken, or one account stalls provisioning for all of them.
+  #
+  # An unlocked read was not enough: `create_sandbox/1` and the conversation
+  # insert take `KEY SHARE` on `users` through their foreign keys, and
+  # `Credits.insert_and_move/3` holds that row `FOR UPDATE` across a ledger
+  # insert, lot consumption and the balance move. Taking `FOR SHARE` out here
+  # both settles the wait outside the fleet lock and satisfies those foreign
+  # keys, so the inserts below cannot block on it. The rotation unbind is the
+  # same category of wait and joins them.
+  #
+  # This is one transaction: the nested `Repo.transaction` inside
+  # `with_sandbox_reservation/3` joins it rather than opening another, so the
+  # sandbox, conversation and allowance still commit or roll back together.
+  # That is also why the `case` below re-raises the inner rollback with its
+  # reason: a nested rollback the outer transaction does not re-raise reaches
+  # the caller as `{:error, :rollback}`, which would turn every credits, quota
+  # and fleet refusal into a 500 instead of a 402, 422 or 503.
+  #
+  # The wait does not disappear, it changes hands. This transaction holds
+  # `users FOR SHARE` for its whole life, the fleet-lock wait included, and
+  # `FOR SHARE` conflicts with `FOR UPDATE` — so this tenant's credit postings
+  # now queue behind its own in-flight launch, which may itself be queued
+  # behind every other tenant's. Turn burns, purchases, grants, expiry and
+  # refund clawbacks all post through `Credits.insert_and_move/3`. A
+  # tenant-scoped wait beats a fleet-wide one, which is why it is the right
+  # trade, but a slow credit posting starts here.
   defp reserve_initial_conversation(sandbox_attrs, conversation_attrs, request, opts) do
-    Fountain.Quotas.with_sandbox_reservation(conversation_attrs.user_id, fn ->
-      # Nothing in here may take a row lock. `with_sandbox_reservation/3` holds
-      # `pg_advisory_xact_lock(@fleet_lock_key)` — one lock shared by every
-      # tenant, the thing that enforces SANDBOX_FLEET_CEILING — for the whole
-      # of this function. A `SELECT ... FOR SHARE` on `users` conflicts with
-      # the `FOR UPDATE` that `Credits.insert_and_move/3` holds across a ledger
-      # insert, lot consumption and the balance move, so one tenant's credit
-      # posting would stall provisioning for every other tenant. `agents` is
-      # the same story against a `last_used_at` stamp. These stay plain
-      # ownership rechecks: MVCC reads never block, the ceiling below is read
-      # the same way, and the inserts' foreign keys enforce integrity.
+    Repo.transaction(fn ->
       Repo.one(
         from u in Fountain.Accounts.User,
           where: u.id == ^conversation_attrs.user_id,
-          select: u.id
+          select: u.id,
+          lock: "FOR SHARE"
       ) || Repo.rollback(:not_found)
 
       Repo.one(
         from a in Agents.Agent,
           where:
             a.id == ^conversation_attrs.agent_id and a.user_id == ^conversation_attrs.user_id,
-          select: a.id
+          select: a.id,
+          lock: "FOR SHARE"
       ) || Repo.rollback(:not_found)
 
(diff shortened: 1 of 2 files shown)


jhgaylor:
Addressed the review, and rebased onto `main` (this was still based on `feat/fresh-initial-execution-allowance` at the pre-rebase `fc005df5`, which is why the diff read +882; it is +168/-24 against `main`).

**The tradeoff is now written down in both places.** You asked for a line in the description; I put it in the comment on `reserve_initial_conversation/4` as well, because that is where someone debugging a slow credit posting will actually be looking, and a PR description is not in the file. It says plainly that the wait changes hands rather than disappearing: this transaction holds `users FOR SHARE` across the fleet-lock wait, `FOR SHARE` conflicts with `FOR UPDATE`, so the tenant's own credit postings queue behind its in-flight launch — and names the five things that post through `insert_and_move/3`.

**Also recorded the nested-transaction trap in the comment**, since you nearly filed the `case result` block as dead code and the next reader has the same information you had. It now says why re-raising with the reason is load-bearing: without it every credits, quota and fleet refusal reaches the caller as `{:error, :rollback}` and becomes a 500 instead of a 402, 422 or 503.

**On the rebase conflict**, agreed with your resolution and it is what I pushed: `unbind_rotated_channel` belongs in the outer transaction with the other two. It waits on a row turn admission takes `FOR UPDATE`, so leaving it inside would have preserved exactly the class of stall this change removes — and the 250ms bound from #1791 stays as defence rather than as the only thing standing between a busy channel and the fleet.

Re-verified after the change: compile with `--warnings-as-errors`, format at both roots, `credo --strict` clean, 78 tests across the admission and allowance suites, no leftover fixtures.

BinaryBourbon (approved):
This is a good catch and the mechanism is right. I verified the parts I could not take on faith, and I have one tradeoff I want stated in the PR before it lands rather than found later.

**The finding is real and I reproduced it.** #1790 removed the explicit `FOR SHARE` from inside `with_sandbox_reservation/3`, and I said at the time that this removed the exposure the stack added but not the class — `create_sandbox/1` and the conversation insert take `KEY SHARE` on `users` through their foreign keys, and `KEY SHARE` conflicts with the `FOR UPDATE` that `Credits.insert_and_move/3` holds. This PR is that follow-up, and its two tests fail on `997ed65f` for exactly that reason. I reverted the two `lock: "FOR SHARE"` clauses while keeping the outer transaction and watched both fail, so the locks are what carries the fix, not the restructuring alone.

**The mechanism is subtler than "move the lock earlier" and worth spelling out**, because the next reader will wonder why acquiring a lock *helps*. Taking `FOR SHARE` in the outer transaction does two things at once: it settles the wait before `pg_advisory_xact_lock(@fleet_lock_key)` is taken, and — because `FOR SHARE` is compatible with `KEY SHARE` and is already held by this transaction — it satisfies the foreign-key checks inside, so the inserts cannot block on that row either. Both halves are needed. The comment covers this now.

**The `case result` re-rollback is load-bearing and I nearly flagged it as dead code.** It is not. I probed the semantics directly rather than trusting my memory of them:

- Nested `Repo.transaction` joins the outer transaction; the inner returns `{:error, reason}` as normal.
- If the outer then completes normally, the caller gets `{:error, :rollback}` — the reason is *lost*.
- Re-raising with `Repo.rollback(reason)`, which is what this PR does, restores it: the caller sees `{:error, :insufficient_credits}`, identical to the single-transaction baseline.

Without those three lines every refusal `with_sandbox_reservation/3` produces — `:insufficient_credits` (402 with `upgrade_url`), `:fleet_full` (503), a quota changeset (422) — would reach `FallbackController` as `:rollback` and fall through to a 500. That is a trap worth a comment, and it now has one.

**The test is the right shape.** Asserting that the blocked launch acquires its lock eventually would prove nothing; asserting that *a second tenant's launch commits while the first is still blocked* is the property. Doing it for both the account and the agent row, on real independent connections with an observed lock wait, is thorough.

**The tradeoff I want written down.** This does not make the wait disappear, it moves who does the waiting. The outer transaction now holds `users FOR SHARE` for its whole life — including the wait for the global fleet lock — and `FOR SHARE` conflicts with `FOR UPDATE`, so a tenant's credit postings now queue behind that tenant's in-flight launch, which may itself be queued behind every other tenant's launch. Turn burns, purchases, grants, expiry and refund clawbacks all go through `insert_and_move/3`.

That is the right trade — a tenant-scoped wait is strictly better than a fleet-wide one — but it is a real inversion and it is not in the PR description. Please add a line saying so, so the next person reading a slow credit posting has somewhere to start. No code change needed for that.

Rebased onto `main` (it was still based on `feat/fresh-initial-execution-allowance` at the pre-rebase `fc005df5`, which inflated the diff to +882). The only conflict was #1791's `unbind_rotated_channel` and the `opts` parameter, which this predates; I put the unbind in the outer transaction with the other two, since it is the same category of wait — a row turn admission takes `FOR UPDATE` — and leaving it inside would have kept a wait under the fleet lock that this PR exists to remove.

Local verification on the rebased tree: compile with `--warnings-as-errors`, format at both roots, `credo --strict` (6994 mods/funs, no issues), 131 tests across the admission and allowance suites, no leftover fixtures.

**The tradeoff landed in a better place than I asked for.** I asked for a line in the description; it went into the comment on `reserve_initial_conversation/4` as well, which is the right call — the description is not in the file, and someone tracing a slow credit posting is reading the function. The nested-transaction trap is recorded there too, which closes the gap that had me halfway to filing the `case result` block as dead code.

CI green on `95efd5bc`: 19 pass, 4 skipping, `CI required` pass.

Approving. Disclosure: I did not author the change, but I resolved its rebase conflict onto `main` and moved the rotation unbind into the outer transaction, so that part is mine and has not had a second reader.

PR #1794, opened 2026-09-10
ADR 0048: the repo moves to the managoat organization

jhgaylor:
Decision only. Nothing in this PR moves anything.

`BinaryBourbon/fountain` is a personal **user** account, and it is the last part of the project still there: the nine graduated libraries, `demos`, `goatherd`, `airlock`, the three `fountain-template-*` starters and `review-loop` are all in the `managoat` org, and `scripts/graduate-library.sh` already writes `managoat/` source URLs. ADR 0048 decides the transfer to `managoat/fountain` — repo name and project name unchanged, per 0034's "the project is not renamed".

The useful half of the file is the fallout, split by whether the transfer forces it:

**Forced** (a GHCR namespace belongs to its owner, and a repo's `GITHUB_TOKEN` cannot write another owner's packages, so these break the moment the repo moves): `ghcr.io/managoat/fountain`, `ghcr.io/managoat/fountain-manifests`, `github.com/managoat/fountain/cli` and the Buzz provider module beside it, `managoat/homebrew-tap`, and the four SDK manifests' repository URLs.

**Not forced** (the name carries no owner, so nothing breaks on the day): the npm scope, the PyPI name, the hex name. Renaming an SDK package breaks every consumer's import — the template repos, `demos`, `fountain-conversations`, `fountain-team`, Workbench, `salon` — so it is deliberately uncoupled from a transfer that breaks none of them. Target names are fixed here anyway, which is what keeps interim work from being wasted: nothing merged from now on writes a new `binarybourbon` or `agentshit` coordinate.

Also in the file:

- **The cutover order**, including the one sequence that can break prod. `build.yml:35` and `publish-manifests.yml:142` hold the image and artifact paths as literals rather than deriving them from `github.repository`, so the first build after the transfer fails loudly instead of publishing where nobody looks — but the gap between the transfer and home-cloud's `OCIRepository` url is a merge freeze either way.
- **What does not transfer and must be verified by hand**: the five repository secrets, the `github-pages` and `pypi` environments, the ruleset and its bypass actor, PyPI trusted publishing (it names owner + repo + workflow filename, and a failed publish is invisible to every gate here), npm provenance's repository claim.
- **Three breaks we accept**: `go install github.com/BinaryBourbon/...` once `go.mod` declares the new path, `binarybourbon.github.io/fountain/*` (Pages does not follow a transfer redirect — that is the #1011 tombstone), and published image pins that keep working but drift from the docs.

`okf validate decisions` and `okf lint decisions` are clean (0 errors; the 4 warnings are pre-existing actor-convention ones on other files), and `decisions/index.md` is regenerated in the same commit.

Reviewing this is mostly checking two judgment calls: renaming the Go module paths rather than keeping the old ones declared, and deferring the SDK package renames instead of doing everything in one window.


jhgaylor:
**Cutover step 0, the "before" state recorded** (read-only, from the cluster just now):

| What | Value |
|---|---|
| Deployment image | `ghcr.io/binarybourbon/fountain:sha-7a7177bba0b42c6dad2c3e050bac249093b41890` |
| Running digest (both replicas) | `ghcr.io/binarybourbon/fountain@sha256:d93b613405ae7c5a8f8eca097d366c8bdc6a3c3a146d5aa1be35fb3fa2169aff` |
| Flux `OCIRepository/fountain` | `oci://ghcr.io/binarybourbon/fountain-manifests`, `latest@sha256:ff466976a88e970cd8ff31c1058009b4cc240f45887ed44d687ec153113ffb11` |
| Self-host pin | `v0.16.0` in `.env.compose.example`, matching `mix.exs` — which is also why `ci.yml`'s quick-start job takes its release-bump exemption on the sweep PR instead of failing |

That digest is the rollback target for the whole window: nothing deploys between the transfer and the home-cloud flip, so if the flip goes wrong the cluster is already running this and `flux tag artifact oci://ghcr.io/managoat/fountain-manifests:sha-<old> --tag latest` is not needed — re-pointing the `OCIRepository` back is enough.

Both execution PRs are open as drafts and green: BinaryBourbon/fountain#1795 (the in-repo sweep) and jhgaylor/home-cloud#221 (the Flux url and the release-asset URLs).

jhgaylor:
**The cutover, as commands.** Everything below is ready to run in order; steps 0, 4's prerequisites and the "before" state are already done.

**1. Freeze.** Stop merging to `main`. Nothing else needed — `build.yml` only fires on a CI-successful push to main.

**2. Transfer.**

```bash
gh api -X POST repos/BinaryBourbon/fountain/transfer -f new_owner=managoat
gh api -X POST repos/BinaryBourbon/homebrew-tap/transfer -f new_owner=managoat
gh repo view managoat/fountain --json nameWithOwner,isInOrganization
```

The token needs admin on the repo and the right to create repositories in the org; both hold for the account that owns them.

**3. Verify what survived** (expect the first four to come back populated; recreate anything missing before merging a thing):

```bash
gh secret list   --repo managoat/fountain      # FLUX_WEBHOOK_URL, HEX_API_KEY, HOMEBREW_TAP_TOKEN, RELEASE_BUMP_TOKEN, RENDER_DEPLOY_HOOK_URL
gh variable list --repo managoat/fountain      # REVIEW_LOOP_ENABLED, REVIEW_LOOP_URL
gh api repos/managoat/fountain/environments --jq '.environments[].name'   # github-pages, pypi
gh api repos/managoat/fountain/rulesets --jq '.[] | {id, name, enforcement}'
#   ^ the id is 21689465 today and scripts/ci/require-checks.py carries it as
#     a default; if it changed, that default changes with it.
gh api repos/managoat/fountain/actions/permissions
gh api repos/managoat/fountain/collaborators/BinaryBourbon/permission --jq .permission
```

That last one is the estate-medic bot identity, which held Write by owning the repo. Grant it back, and keep it out of admin so it still cannot approve its own PR:

```bash
gh api -X PUT repos/managoat/fountain/collaborators/BinaryBourbon -f permission=push
```

**4. Copy the released tags into the new namespace**, so the documented self-host pin resolves and `ci.yml`'s quick-start check keeps booting a real image instead of taking its release-bump exemption:

```bash
echo "$GHCR_WRITE_TOKEN" | docker login ghcr.io -u jhgaylor --password-stdin
docker buildx imagetools create -t ghcr.io/managoat/fountain:v0.16.0 ghcr.io/binarybourbon/fountain:v0.16.0
docker buildx imagetools create -t ghcr.io/managoat/fountain:latest  ghcr.io/binarybourbon/fountain:v0.16.0
docker manifest inspect ghcr.io/managoat/fountain:v0.16.0 | jq '.manifests[].platform'
```

`v0.16.0` is a two-platform index (`linux/amd64`, `linux/arm64`); `imagetools create` copies the index rather than flattening it, which matters because the cluster is arm64. Then make both new packages **public** and link them to the repo — the `OCIRepository` has no `secretRef`, so Flux fetches anonymously.

**5. Merge the sweep** (#1795 — mark ready, it is green and carries the `sdk-no-release` label):

```bash
gh pr ready 1795 --repo managoat/fountain && gh pr merge 1795 --repo managoat/fountain --squash --admin
gh run watch --repo managoat/fountain "$(gh run list --repo managoat/fountain --workflow build.yml --limit 1 --json databaseId --jq '.[0].databaseId')"
docker manifest inspect "ghcr.io/managoat/fountain:sha-$(gh api repos/managoat/fountain/commits/main --jq .sha)" > /dev/null && echo published
```

**6. Flip Flux** (jhgaylor/home-cloud#221), then confirm the cluster actually moved rather than trusting the merge:

```bash
flux reconcile source oci fountain -n flux-system
kubectl get ocirepository -n flux-system fountain -o jsonpath='{.spec.url}{"\t"}{.status.artifact.revision}{"\n"}'
kubectl rollout status deploy/fountain -n fountain --timeout=5m
kubectl get pods -n fountain -l app=fountain -o jsonpath='{range .items[*]}{.status.containerStatuses[0].imageID}{"\n"}{end}'
```

The image IDs must leave `sha256:d93b6134…` and the url must read `oci://ghcr.io/managoat/fountain-manifests`. If it does not, re-point the `OCIRepository` back and the cluster is already where it was — no rollback tag needed.

**7. The things no PR can do.** PyPI: edit the trusted publisher on `fountain-agent-sdk` to owner `managoat`, repo `fountain`, workflow `python-sdk-publish.yml`, environment `pypi` — do this in this window, because the next Python SDK release fails silently-ish without it. npm: nothing to configure, provenance reads `package.json`, which #1795 already fixed. Hex: update the package links on the next publish. `managoat/review-loop-action`: check its OIDC allow-claim for the old `owner/repo`. And the GitHub **Project** (`users/BinaryBourbon/projects/1`) stays on the account — rebuild or copy it under the org when you want it there, and re-point `.agents/skills/fountain-project-gardener`.

**8. Unfreeze.** Then the deferred half of 0048: the npm scope, the PyPI name, and whether hex needs one — each its own PR with both names published for a window.

PR #1795, opened 2026-09-10
Point every coordinate at managoat/fountain (0048 execution, hold for the transfer)

jhgaylor:
**Draft on purpose. Do not merge before the repo is transferred** — it names a repository and two container packages that do not exist until then. Step 4 of the cutover in #1794.

128 files, 272 insertions. Every change is one of three kinds.

**Coordinates of this repo** (rewritten): the two GHCR paths in `build.yml`, `publish-manifests.yml`, `release.yml` and `ci.yml`'s quick-start pin check; the five published image pins plus the three regexes in `release_pin_test.exs` that gate them; both Go module paths, every import, and the `-ldflags` version stamp; the Homebrew tap checkout, the generated formula and the ten documented install lines; release-asset URLs including the Dockerfile's buzz-acp download; the four SDK manifests' repository URLs; and the `gh`/docs URLs in `CLAUDE.md`, `CONTRIBUTING.md`, `SETUP.md`, `README.md`, the issue template, both skill trees, `docs/`, and the runtime strings users actually read (`api_spec.ex`, `llms_controller.ex`, `marketing_html.ex`, the bundled skill, `priv/help`).

**Names that carry no owner** (untouched): `@agentshit/fountain-sdk`, `fountain-agent-sdk`, `fountain_sdk`. Only their repository metadata moved. Renaming a package breaks every consumer's import, so that is its own PR with a deprecation window, per 0048.

**Records and other repos** (left alone): both CHANGELOGs and `decisions/` keep the name they were written with — the links redirect, and rewriting a record of what happened is wrong. `priv/help/skills.md`'s `BinaryBourbon/skills` and the docs' example `BinaryBourbon/fountain-api-skill` are different repositories; `deployed/matrix.mjs` now accepts either owner so a `matrix.json` written before the move still validates.

Verified locally: `go build`/`go vet`/`go test ./...` green in both modules (`cli/` and `apps/fountain_buzz/cli/`), `py_compile` on `require-checks.py`, `node --check` on the two `deployed/` modules, `json.load` on `matrix.example.json`, `yaml.safe_load` on every workflow and issue template. `sdk/contract/contract.json` does not project `info.description`, so the `api_spec.ex` wording change leaves the committed contract current — `build.sh --check` in CI is the confirmation.

Three things this PR surfaced that the ADR now records (see the second commit on #1794):

1. **Copy the released image tags into the new namespace before merging this.** `ci.yml`'s quick-start job hard-fails when the pinned tag has no image, with one exemption: when the pin equals `mix.exs`'s version it assumes a release bump and skips. `v0.16.0` happens to satisfy that today, so this PR stays green — by luck, and with the compose boot check skipped rather than passing. Copying `v0.16.0` (and `latest`) from `ghcr.io/binarybourbon/fountain` keeps the check real and keeps every published self-host pin resolvable at the new path.
2. **A user-owned Project does not transfer.** `users/BinaryBourbon/projects/1` stays where it is; `.agents/skills/fountain-project-gardener` is pointed at the repo, not the project, and the project itself needs its own move.
3. **`require-checks.py` also defaults to a numeric ruleset id** (`21689465`), which is worth re-reading after the transfer alongside the secrets and environments.

Out of scope but worth a look: `BinaryBourbon/skills` and `*/fountain-api-skill` are used as documentation examples and neither repository exists.


jhgaylor:
Evidence note, so the red run above is not read as a real failure.

The second push (the Python gate parity commit, which touches one workflow file and nothing else) came back with four reds: partitions 5 and 6, "Core distribution boots without the extensions", and "Elixir 1.15 / OTP 26 compatibility". All four carry the same single annotation:

```
Could not fetch /builds/otp/amd64/ubuntu-24.04/builds.txt from any hex.pm mirror
Action fetch /builds/otp/amd64/ubuntu-24.04/builds.txt failed for mirror https://builds.hex.pm, with Error: Gateway Error
```

That is `setup-beam` failing to install a toolchain, before any of this repo's code runs — `builds.hex.pm` was briefly returning gateway errors around 10:04–10:05 UTC. It is serving 200 again now. Corroboration: the first run of this PR, on the same tree apart from that one workflow file, had all six partitions green.

Failed jobs re-run rather than re-pushed, so the earlier attempt's logs stay attached to the run.

PR #1796, opened 2026-09-10
fix(tests): serialize global credit-pricing changes

jhgaylor:
The credit-pricer suite changes global message prices while running asynchronously. A full-suite run at seed `828329` made `CreditsTest` observe those temporary prices and fail its default-price assertion.

Run the pricer suite synchronously so its existing `on_exit` restoration completes before other suites read pricing. One file, +2/-1, extracted from #1754 onto `main`.

Validation: both affected suites pass (38 tests, original seed). Rebased onto `ea7b8be5`; full `mix precommit --seed 828329` passes: 4,885 tests +6 doctests, zero failures. CI passes on this revision ([run](https://github.com/managoat/fountain/actions/runs/34471070392)); earlier attempts were blocked by Hex mirror connectivity.

### DIFF
--- a/ee/test/fountain/workers/credit_pricer_test.exs
+++ b/ee/test/fountain/workers/credit_pricer_test.exs
@@ -4,7 +4,8 @@ defmodule Fountain.Workers.CreditPricerTest do
   providers Fountain pays for, only from the configured instant.
   """
 
-  use Fountain.DataCase, async: true
+  # Message pricing changes global :credits configuration read by other suites.
+  use Fountain.DataCase, async: false
 
   alias Fountain.Billing
   alias Fountain.Credits


jhgaylor (commented):
**Reviewed — no changes requested.** One file, `async: true` → `async: false`, and it's the justified kind.

The suite writes `Application.put_env(:fountain, :credits, …)` at line 138 and `:credits_enabled` at 261. That's global application env, which lives entirely outside the SQL Sandbox, so the existing `on_exit` restoration is racing every other async module that reads pricing rather than protecting anything. `CreditsTest` observing temporary message prices and failing its default-price assertion is exactly the shape that produces. CLAUDE.md carves out global mutable state as the legitimate reason to go sync, and this is it.

Good that the seed is in the PR body — replaying the failing seed is the only way to tell a real fix from seed luck, and `828329` is the evidence.

Nit, take it or leave it: only two `describe` blocks touch global config, and the whole module now runs sync to cover them. Splitting those into their own `async: false` module would keep the rest of the file parallel. Not worth a re-push unless you're already in there — the file is small and the sync cost is trivial next to a flake that fails an unrelated suite.

BinaryBourbon (approved):

BinaryBourbon (approved):
**Approving.** One file, `async: true` → `async: false`, and it is the carved-out kind.

The module writes `Application.put_env(:fountain, :credits, …)` at line 138 and `:credits_enabled` at 261. Application env lives outside the SQL Sandbox, so the `on_exit` restoration bounds the window but does not close it — any async module reading pricing in that window sees the test's values. CLAUDE.md names global mutable state as the one legitimate reason to go sync, and this is it. The replayed seed in the PR body is the right evidence; branch-vs-main with fresh seeds would only have measured seed luck.

Nit, not blocking: line 262 restores `:credits_enabled` to a hardcoded `true` rather than to the value captured beforehand, unlike line 127 which captures `cfg` first. Sync execution makes that harmless today, but it is a latent trap if the test-config default ever changes. Capturing it costs one line.

Second nit: only two `describe` blocks touch global config and the whole module now runs sync for them. Splitting those out would keep the rest parallel — not worth a re-push, the file is small.

PR #1797, opened 2026-09-10
feat(broker): add tenant-scoped session cleanup

jhgaylor:
Worker cleanup needs to revoke its own broker token without removing a replacement session for the same conversation. Add `Broker.release_session/3`, scoped by tenant, conversation and token hash; repeated cleanup is harmless and works after broker shutdown or tenant unenrollment. Existing proxy tunnels are unaffected.

Extracted from #1754 onto #1796 (the one-file test-isolation fix): four files, +85 lines. This is an internal primitive; lifecycle caller integration remains separate.

Validation: 50 broker tests pass, covering replacement preservation, all three scope predicates, repeated cleanup and disabled-broker cleanup. Full `mix precommit --seed 828329` passes: 4,886 tests +6 doctests, zero failures.

Rebased onto current main through #1796; the fresh full gate passes (4,888 tests +6 doctests). CI passes for this head ([run](https://github.com/managoat/fountain/actions/runs/34473233390)). The earlier Hex connectivity failure remains recorded in [the old run](https://github.com/managoat/fountain/actions/runs/34465468509).

Part of #1864

### DIFF
--- a/apps/fountain/lib/fountain/broker.ex
+++ b/apps/fountain/lib/fountain/broker.ex
@@ -571,6 +571,15 @@ defmodule Fountain.Broker do
     end
   end
 
+  @doc """
+  Delete the session matching tenant, conversation and token, even after the
+  broker is disabled. Repeated cleanup succeeds without touching other tokens.
+  Subsequent lookups refuse the token; existing proxy tunnels are not closed.
+  """
+  @spec release_session(String.t(), String.t(), String.t()) :: :ok
+  def release_session(user_id, conversation_id, token),
+    do: Native.release_session(user_id, conversation_id, token)
+
   @typedoc "One outbound request the broker handled for a conversation."
   @type egress_event :: %{
           id: integer(),
--- a/apps/fountain/lib/fountain/broker/native.ex
+++ b/apps/fountain/lib/fountain/broker/native.ex
@@ -436,6 +436,11 @@ defmodule Fountain.Broker.Native do
   def release(conversation_id) when is_binary(conversation_id),
     do: Sessions.release(conversation_id)
 
+  @doc "Revoke one session matching tenant, conversation and token."
+  @spec release_session(String.t(), String.t(), String.t()) :: :ok
+  def release_session(user_id, conversation_id, token),
+    do: Sessions.release_session(user_id, conversation_id, token)
+
   @doc """
   The conversation's egress rows, newest first (gate 4, #1486). The same
   shape the Agent Vault backend answered with, `status`, `latency_ms` and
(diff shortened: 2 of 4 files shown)


jhgaylor (commented):
**Reviewed — no changes requested.** Small, tenant-scoped, and tested for the things that actually matter.

`release_session/3` keying on `{user_id, conversation_id, token_hash}` rather than conversation alone is the right primitive — `release/1` deleting every session of a conversation is precisely what makes replacement tokens collateral, and this gives the caller a way to revoke only its own. Scoped by `user_id` so it needs no `_unsafe_` prefix, hashing the token rather than comparing it, and `delete_all` making repeated cleanup naturally idempotent are all right.

The three tests cover the cases I'd have asked for: a replacement surviving, each of the three scope predicates failing independently (the `for` loop over wrong-tenant / wrong-conversation / wrong-token is a nice way to write that), and cleanup still working after `broker_tenants` is emptied and the listener is gone. That last one is the non-obvious requirement — cleanup has to work on a path where the broker is already disabled — and it's pinned.

Checked that the new test's `put_env`/`delete_env` calls are covered by the module's existing `setup`-level `on_exit` restoration and that the file is already `async: false`, so this doesn't add a sibling of the flake #1796 fixes.

Landing the primitive ahead of its caller is the right call given the stack.

BinaryBourbon (approved):

BinaryBourbon (approved):
**Approving.** Additive, three thin layers down to one scoped `delete_all`, and the tests cover the axes that matter.

The scoping is right: `user_id` **and** `conversation_id` **and** `token_hash`, so the operation cannot be widened by a caller that only knows the conversation. That is the whole point next to the existing `release/1`, which deletes every session on the conversation — the over-broad revoke that #1827 is about to stop calling.

What I checked:

- **Repeat-safe.** `delete_all` on a non-matching set is `{0, nil}` and still answers `:ok`, so cleanup on an already-clean row is not an error. Tested directly.
- **Works with the broker off.** `release_session/3` goes to `Native` unconditionally rather than through `backend()`, so unenrolling a tenant or dropping `BROKER_LISTEN_PORT` does not strand rows. The third test proves both, which is the case a `backend()`-dispatched version would have silently skipped.
- **No global-env leak.** The module is `async: false` and `setup` captures and restores all three broker keys, so the `put_env`/`delete_env` inside the third test cannot escape — the same class #1796 is fixing one PR down the stack.

The docstring is honest about the limit that matters operationally: an existing CONNECT tunnel is not torn down, only the next lookup is refused. Worth keeping that sentence when this grows a caller.

PR #1799, opened 2026-09-10
fix: a save that changes nothing records nothing, and apply says so

### LINKED ISSUE #1680: Bulk apply reports `updated` for a document that changed nothing, so a client cannot tell a no-op from a write

jhgaylor:
`Repo.update` on an empty changeset skips the SQL, but `Environments`,
`Vaults`, `Agents` and `Webhooks` audited the `{:ok, _}` either way. An
idempotent re-apply of a manifest therefore wrote one `*.updated` row per
resource, each naming an empty changed-field list, about a record nobody had
touched. That is what CLAUDE.md's "only record what happened" forbids. The
four contexts now gate the recording on a non-empty changeset, the way
`Fountain.Team.Schedules.update_schedule/3` already did.

The same re-apply also *reported* an update it had not made. `Manifest` now
compares the record before and after over the schema's own columns (timestamps
left out) and reports a third verdict, `unchanged`, so a second apply of the
same file says plainly that it wrote nothing. Inline `spec.secrets` are the
exception and keep reporting `upserted`: the stored ciphertext cannot be
compared with the plaintext given, so they are re-encrypted on every apply.

`fountain apply` prints `=` for such a row, next to `+` and `~`.

This is the first of eight PRs that together land #1636 (one manifest
reconciles a whole estate). #1675 is the combined branch and stays open as the
reference until the stack is in.

Closes #1680

Validation:

- `mix test` over the touched suites, `mix precommit` clean.
- `go test ./...` in `cli/`.
- The contract and TypeScript types are regenerated; the SDK version bump for
  the whole stack lands in the last PR, so no publish happens here.

---

**The stack** (each PR is based on the one above it; review and merge top down):

1. `stack/1680-unchanged-verdict` — a save that changes nothing records nothing (#1680) ← **this one**
2. `stack/1636-apply-isolation` — a raise in one apply document fails that row
3. `stack/1636-cotenant-identity` — a co-tenant follows a replacement only if it declared the same identity
4. `stack/1636-update-teammate` — Team.update_teammate/4
5. `stack/1636-cli-group-by-kind` — CLI: group apply documents by kind
6. `stack/1636-teammate-kind` — apply reconciles Teammate documents
7. `stack/1636-schedule-kind` — apply reconciles Schedule documents
8. `stack/1636-webhook-kind` — apply reconciles Webhook documents (#1636)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,12 @@ upgrade, is in
 
 ### Fixed
 
+- Updating an environment, vault, agent or webhook endpoint with the values it
+  already holds no longer records an `*.updated` audit event naming no changed
+  fields, and `fountain apply` reports those rows as `unchanged` rather than
+  claiming an update. A re-apply of the same manifest now writes nothing and
+  says so (#1680).
+
 - A secret edited or rotated in an environment or vault during a brokered
   conversation reaches the broker before the next turn. The broker's copy was
   split once, at provisioning, and only the tenant's connection tokens were
--- a/apps/fountain/lib/fountain/agents.ex
+++ b/apps/fountain/lib/fountain/agents.ex
@@ -192,9 +192,14 @@ defmodule Fountain.Agents do
         Repo.update(changeset)
       end
 
+    # A save that moves nothing records nothing, the same rule
+    # `Fountain.Environments.update_environment/3` follows (#1680).
     result =
-      result
-      |> audited("agent.updated", merge_metadata(opts, Audit.changed_fields(changeset)))
+      if changeset.changes == %{} do
+        result
+      else
+        audited(result, "agent.updated", merge_metadata(opts, Audit.changed_fields(changeset)))
+      end
 
     # Only once the new identity is the committed one: a home torn down
     # against an update that then failed would be rebuilt for nothing.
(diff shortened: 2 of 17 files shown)


jhgaylor:
Labelled `sdk-no-release`. This PR regenerates `sdk/contract/contract.json` and the TypeScript types but leaves the version alone on purpose: the whole stack publishes once, from #1806, as 1.24.0. Four intermediate releases for one feature would each describe a half-finished apply vocabulary, and `latest` would move three times on the way there.

PR #1804, opened 2026-09-10
feat(api): bulk apply reconciles Teammate documents

jhgaylor:
`POST /api/apply` reconciled `Environment`, `Vault` and `Agent`. Putting an
agent on the team was a separate call afterwards, so a checked-in manifest
could not declare who the team is.

A `Teammate` document does that. It is keyed by the document name, which is
what the teammate is called; `agent`, `environment` and `vault` resolve by
name, against the manifest first and then against the tenant's own records,
the way an Agent's `environment` already did. Adding one opens the teammate's
conversation through `Fountain.Team.add_teammate/4`; re-applying goes through
`Fountain.Team.update_teammate/4` (the PR before this one) and provisions no
second computer.

Three refusals are worth calling out, each failing its own row and no other:

- **A `Teammate` document is the whole teammate.** Unlike the other kinds,
  where an absent `spec` key leaves that column alone, dropping `environment`
  or `vault` clears that binding. That is what makes the document a
  declaration rather than a patch, and it is said in all three places a reader
  looks.
- **Two documents may not name the same agent.** A teammate is one agent's
  membership of the team, so two documents describe one record: without this
  the second renames the first's conversation on every pass and the manifest
  is never idempotent. Two documents sharing a *name* are refused for the same
  reason.
- **A reference that resolves to nothing fails.** Including another tenant's
  agent, which resolves to nothing here by construction.

Result-row error keys are strings now whichever branch produced them.
`traverse_errors/2` keys by field atom while the hand-built refusals key by
string, and a caller reading a row should not have to know which one it got.
The wire shape does not change; JSON stringifies either.

Sixth of eight PRs toward #1636. #1675 is the combined branch and stays open
as the reference until the stack is in.

Validation: `mix precommit`, `go test ./...` in `cli/`, the SDK contract check
and the TypeScript verifier. The SDK version bump for the whole stack lands in
the last PR, so no publish happens here.

---

**The stack** (each PR is based on the one above it; review and merge top down):

1. `stack/1680-unchanged-verdict` — a save that changes nothing records nothing (#1680)
2. `stack/1636-apply-isolation` — a raise in one apply document fails that row
3. `stack/1636-cotenant-identity` — a co-tenant follows a replacement only if it declared the same identity
4. `stack/1636-update-teammate` — Team.update_teammate/4
5. `stack/1636-cli-group-by-kind` — CLI: group apply documents by kind
6. `stack/1636-teammate-kind` — apply reconciles Teammate documents ← **this one**
7. `stack/1636-schedule-kind` — apply reconciles Schedule documents
8. `stack/1636-webhook-kind` — apply reconciles Webhook documents (#1636)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,12 @@ upgrade, is in
   existing launches retain `owner` behavior. Applications processing mutually
   untrusted work can keep all Fountain API authority on their service host.
 
+- A `fountain apply` manifest can declare team membership. A `Teammate`
+  document names its agent, environment and vault, and the apply puts the agent
+  on the team, which opens its conversation and provisions its computer.
+  Re-applying moves the name and the bindings and provisions no second
+  computer (#1636).
+
 ### Fixed
 
 - A teammate can be moved to a different environment or vault. Fountain retires
--- a/apps/fountain/lib/fountain/manifest.ex
+++ b/apps/fountain/lib/fountain/manifest.ex
@@ -3,25 +3,52 @@ defmodule Fountain.Manifest do
   Bulk apply for compiled IaC manifests (`fountain apply`).
 
   Takes the full list of resources the CLI compiled from a `fountain.yml`
-  and reconciles them against the tenant's records in one pass:
-  environments first, then vaults, then agents — the same fixed ordering
-  `fountain apply` has always used, so an Agent doc can reference an
-  Environment by name regardless of document order. Agent `environment`
-  references resolve against environments in this manifest first, then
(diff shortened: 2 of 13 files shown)


jhgaylor:
Labelled `sdk-no-release`. This PR regenerates `sdk/contract/contract.json` and the TypeScript types but leaves the version alone on purpose: the whole stack publishes once, from #1806, as 1.24.0. Four intermediate releases for one feature would each describe a half-finished apply vocabulary, and `latest` would move three times on the way there.

PR #1818, opened 2026-09-10
refactor(images): share the prompt image byte limit

jhgaylor:
Move the existing 10 MiB prompt-image limit into `Fountain.Images` and delegate the HTTP decoder's limit to it. Future durable ingestion can share the same ceiling; current acceptance, rejection and error text are unchanged.

Two files, +4/-4, extracted from #1754. Stacked on #1796 to keep the known global-pricing test race out of validation.

Validation: 10 image-ingestion tests pass, including oversized-image rejection. Full `mix precommit --seed 828329` passes: 4,885 tests +6 doctests, zero failures.

CI passes on this revision ([run](https://github.com/managoat/fountain/actions/runs/34472143358)).

Part of #1864

### DIFF
--- a/apps/fountain/lib/fountain/images.ex
+++ b/apps/fountain/lib/fountain/images.ex
@@ -12,6 +12,8 @@ defmodule Fountain.Images do
 
   @valid_media_types ~w(image/png image/jpeg image/gif image/webp)
 
+  def max_prompt_image_bytes, do: 10 * 1024 * 1024
+
   def valid_media_types, do: @valid_media_types
 
   def valid_media_type?(media_type), do: media_type in @valid_media_types
--- a/apps/fountain/lib/fountain_web/prompt_images.ex
+++ b/apps/fountain/lib/fountain_web/prompt_images.ex
@@ -14,9 +14,7 @@ defmodule FountainWeb.PromptImages do
   `{:error, message}`.
   """
 
-  @max_image_bytes 10 * 1024 * 1024
-
-  def max_image_bytes, do: @max_image_bytes
+  def max_image_bytes, do: Fountain.Images.max_prompt_image_bytes()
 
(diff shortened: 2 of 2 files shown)


jhgaylor (commented):
**Reviewed — no changes requested.** +4/-4, behaviour identical, and it puts the constant where the next consumer can reach it.

`FountainWeb.PromptImages` is a web-layer decoder; a byte ceiling that domain code also needs doesn't belong there. Moving it to `Fountain.Images` beside `valid_media_types/0` and delegating `max_image_bytes/0` keeps the existing public function working, so nothing downstream has to change. #1840 then consumes it from the domain side, which is what makes this worth doing separately.

Nit: `validate_size/1`'s message still hardcodes `"image exceeds the 10MB limit"` while the number it's checking now comes from another module. Pre-existing, and interpolating `Images.max_prompt_image_bytes()` would make the text uglier — but if the limit ever moves, that string is the thing that silently lies.

BinaryBourbon (approved):

BinaryBourbon (approved):
**Approving.** Straight move of `10 * 1024 * 1024` from `FountainWeb.PromptImages` to `Fountain.Images`, with the web module now delegating. Correct direction: `Fountain.Images` is core, so `Fountain.Conversations.PromptInput` in #1840 can read the same number without the context reaching into `FountainWeb`. `max_image_bytes/0` keeps its name and its callers, and `validate_size/1` switched from the attribute to the function so there is one source.

One thing left behind, worth a follow-up rather than a re-push: the *message* is still an independent literal — `"image exceeds the 10MB limit"` in `validate_size/1`. Change the constant and the string silently lies. It was already that way before this PR, so not a regression, but this is the PR that makes the constant shareable and the string the last copy.

PR #1827, opened 2026-09-10
Preserve replacement broker tokens when provisioning fails

jhgaylor:
A returned provisioning error revoked every broker session for the conversation, including a replacement worker's token. Preserve the mint result outside `with` and revoke only that attempt's tenant/conversation/token; a failed mint owns no token to revoke. Cleanup also works after brokerage is disabled.

Four files, +71/-27, extracted from #1754 onto #1797. Move session preparation and cleanup into Egress, reducing ConversationServer by 12 lines and lowering its size pin. Full actor/provider ownership fencing remains separate.

Validation: 22 broker/size tests pass, with real session rows. Both new behavior cases fail on the unchanged caller. Full `mix precommit --seed 828329` passes: 4,890 tests +6 doctests, zero failures. Staged secret scan passes. The initial full gate caught the server size limit; the helper extraction and lower pin fix that failure. CI passes ([run](https://github.com/managoat/fountain/actions/runs/34476805041)).

Part of #1864

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -864,7 +864,10 @@ defmodule Fountain.Conversations.ConversationServer do
         # The broker session is minted before the env is built, because the
         # env carries it; the CA is installed before anything dials out,
         # because nothing dials out without it (ADR 0019 gate 1a).
-        with {:ok, state} <- broker_prepare(state),
+        # Keep the result outside `with`: its else cannot see the minted state.
+        prepared = Egress.prepare_state(state)
+
+        with {:ok, state} <- prepared,
              sprite_env = build_sprite_env(state, agent, env, secrets, sandbox_url),
              # A real step, not best effort: an agent whose MCP servers could
              # not be written would otherwise run without them and report
@@ -933,7 +936,7 @@ defmodule Fountain.Conversations.ConversationServer do
           {:error, reason} ->
             Logger.error("provision step failed: #{inspect(reason)}")
             _ = Managoat.Sandbox.destroy(handle)
-            Egress.release(state.user_id, state.conversation_id)
+            Egress.release_prepared(prepared)
             {:ok, _} = Conversations.update_sandbox(sandbox, %{status: "failed"})
 
             Output.publish_stage(state.conversation_id, "provision", "failed", %{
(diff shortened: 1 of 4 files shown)


jhgaylor (commented):
**Reviewed — no changes requested.** Good catch and a clean fix.

The bug is real and the diagnosis is exact: `Egress.release(user_id, conversation_id)` in the provisioning error branch revokes *every* session for the conversation, so a replacement worker that had already minted its own token loses it to an unrelated attempt's failure. Hoisting `prepared` out of the `with` so the error branch can still see the minted state is the minimal fix — `with`'s `else` genuinely cannot reach it otherwise, and the comment says that rather than leaving the next reader to wonder why the call moved.

`release_prepared/1` is the part I looked at hardest and it's right:

- `{:ok, %{broker: %{token: token}}}` revokes exactly that attempt's token.
- A failed mint is `{:error, _}` and falls to `_ -> :ok`, which matches "a failed mint owns no token to revoke".
- An unbrokered conversation is `{:ok, %{broker: nil}}`, doesn't match the first clause, also `:ok`.

I checked that `prepare_state/1` is the only mint in the provisioning path, so there's no earlier session the narrow release would now miss.

The test change is the right shape too — `reject(Fountain.Broker, :release, 1)` alongside `reject(…, :release_session, 3)` asserts that a failed mint revokes *nothing*, which is stronger than the old stub that merely observed a release happening. Renaming the test to say what it now proves is appreciated.

Moving preparation and cleanup into `Egress` and dropping the server pin 2774 → 2762 is the right direction for #1369, and it makes this a net subtraction from `ConversationServer` rather than an addition.

BinaryBourbon (approved):

BinaryBourbon (approved):
**Approving.** This fixes a real over-broad revoke, and the test proves it rather than asserting the code shape.

The bug: the provisioning failure handler called `Egress.release(user_id, conversation_id)`, which deletes **every** session row on the conversation. Anything that minted a replacement between the original prepare and the failure — a re-prepare, a concurrent attach — lost its token to a cleanup that had no business touching it. `release_prepared/1` now revokes exactly the token this preparation returned, and `Broker.release_session/3` (#1797) is scoped tightly enough that it cannot do anything else.

Hoisting `prepared` out of the `with` is the right shape and the comment says why: the `else` branch cannot see the rebound `state`, so the token would not have been reachable there. Reusing the already-evaluated result instead of re-minting is what makes "revoke only what this attempt owns" true.

The new test is the good kind — it stubs only the provider boundary and lets `Broker.Native.prepare/4` write real session rows, so `lookup(original) == :error` and `lookup(replacement) == {:ok, _}` are statements about the database, not about a mock. Running it in both broker-enabled and broker-disabled shapes covers the case where a revoke would have gone through `backend()` and silently no-op'd.

Two observations, neither blocking:

- `release_prepared/1` takes the *result tuple* rather than a session, which is unusual for a public function — `{:ok, state}` in, `:ok` out, and a bare `_` clause swallowing everything else. It reads fine at the one call site because `prepared` is right there, but a second caller passing a session by mistake gets a silent `:ok`. A `@spec` naming the tuple would make the contract visible.
- `prepared` holds the token minted at that instant. If a later provisioning step ever re-mints (the OAuth-refusal re-prepare does this at turn time, not here), the failure handler would revoke the stale token and leak the new one. Today no provisioning step does, so this is correct as written — worth a comment so it stays correct.

`@pin` 2774 → 2762 is downward, as the size test requires.

PR #1831, opened 2026-09-10
Preserve current state after a stale worker-start failure

jhgaylor:
A delayed initial worker-start error could mark a replacement binding, completed conversation or cancellation as failed. Lock and recheck the original tenant/machine binding and pending state before atomically failing both rows. Only an applied failure restores a rotated channel; missing or reassigned parents return not found. Usage is recorded after commit.

Two files, +169/-6, extracted from #1754 onto #1826. This is initial-start error handling; actor-incarnation and uncertain provider recovery remain separate. A changed binding leaves the original sandbox untouched.

Validation: six launch-path regressions cover ordinary failure, replacement, completed work, cancellation, deletion and reassignment. Five fail on the unchanged caller. All 82 focused regression/rotation/admission tests pass, including an assertion that usage sees both failed rows after commit; the earlier start-path run passed 54 tests. Full `mix precommit --seed 828329` passes: 4,894 tests +6 doctests, zero failures. Staged secret scan passes. CI passes ([run](https://github.com/managoat/fountain/actions/runs/34479441710)).

Part of #1864

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2615,12 +2615,17 @@ defmodule Fountain.Conversations do
             "ConversationServer failed to start for conv #{conv.id}: #{inspect(reason)}"
           )
 
-          update_conversation(conv, %{status: "failed"})
-          update_sandbox(sandbox, %{status: "failed"})
-          restore_rotated_channel(conv, opts)
-          result = _unsafe_get_conversation!(conv.id)
-          broadcast_sidebar_update(user_id)
-          {:ok, result}
+          if fail_initial_start(conv, sandbox) == :failed,
+            do: restore_rotated_channel(conv, opts)
+
+          case get_conversation(conv.id, user_id) do
+            nil ->
+              {:error, :not_found}
+
+            result ->
+              broadcast_sidebar_update(user_id)
+              {:ok, result}
+          end
       end
     else
       nil ->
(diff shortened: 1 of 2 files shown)


BinaryBourbon (approved):
Approving. The locked recheck is the right fix and the lock order is right.

Replacing the unconditional `update_conversation(conv, %{status: "failed"})` with a locked recheck of the *original* binding is correct: a delayed `start_child` error arriving after a replacement has rebound the sandbox, or after the work completed, or after cancellation, would otherwise stamp `failed` over state it knows nothing about. Taking the advisory lock then `parent` then `machine` matches turn admission's machine → parent → sandbox order, which is what keeps this from deadlocking against it.

Details I checked:

- `_unsafe_running_turns_elsewhere(sandbox.id, nil)` — the `nil` clause exists and drops the `c.id != ^conv_id` predicate, so it counts every running turn on the machine. That's what you want here: a replacement's turn should block the stale failure.
- `record_sandbox_usage/2` is called *after* `Repo.transaction` returns, and the test asserts `refute Repo.in_transaction?()` inside the Billing mock. That's the right way to pin "metering follows commit" — an assertion on ordering rather than on the outcome.
- Only `:failed` triggers `restore_rotated_channel/2`, so a stale error can't roll a channel back onto a conversation that has moved on.

The return-path change is a quiet improvement worth calling out: `_unsafe_get_conversation!(conv.id)` → `get_conversation(conv.id, user_id)` with `nil -> {:error, :not_found}`. It stops raising on a deleted conversation, and it makes the read tenant-scoped where it previously wasn't.

`pending_initial_binding?/4` comparing `Map.take(parent, [...])` against a literal map is compact and reads fine, though it'll silently start passing if a field is ever renamed rather than failing loudly. Minor.

Note for whoever merges the stack: this sits on #1826, so it inherits that PR's commit-boundary guard.

BinaryBourbon (approved):
**Approving.** The old code wrote `update_conversation(conv, %{status: "failed"})` and `update_sandbox(sandbox, %{status: "failed"})` against the caller's stale structs, unconditionally and outside any shared lock. A `start_child` error that arrived after the conversation had already been rebound to a replacement sandbox would mark the *live* pair failed. `fail_initial_start/2` fixes that by re-reading both rows under the per-sandbox advisory lock and writing only when the binding is still the pending one it created.

`pending_initial_binding?/4` comparing `Map.take/2` against a literal map is the right call: it fails closed on every field at once, and adding a field to the tuple is a one-line change rather than a new `and` clause someone forgets. Requiring `status == "pending"` on *both* rows is what makes "still the original, still unresolved" precise.

The test matrix covers the cases that matter and each one asserts the negative — rebind keeps both machines, already-provisioned is not overwritten, a cancelled parent keeps `terminated`, deleted and reassigned parents return `{:error, :not_found}` with the sandbox left `pending` for the reaper. Checking `refute Repo.in_transaction?()` and the persisted status *inside* the `Billing.record_usage` expectation is a neat way to pin that metering follows the commit.

Three notes, none blocking:

- **The lock-order comment does not describe the code.** It says "machine -> parent -> sandbox lock order," but the sequence is advisory(`phash2(sandbox.id)`) → `Conversation` FOR UPDATE → `Sandbox` FOR UPDATE. That matters because `do_reset_sandbox/2` takes the same advisory key and then goes sandbox-row → conversation-rows, the opposite row order. The advisory lock is what makes that safe, and the comment is the only place a future reader learns it. Please say so explicitly rather than naming an order the code does not follow.
- **`start_conversation/2` can now return `{:error, :not_found}`** where it previously returned `{:ok, conv}`. That is the right answer for a concurrently deleted or reassigned parent, and `get_conversation/2` being tenant-scoped is what makes the reassigned case work. It is still a new terminal value on a documented function — worth a line in the `start_conversation/2` docstring, since callers that only match `{:ok, _}` and `{:error, :insufficient_credits}` now have a third outcome.
- **`restore_rotated_channel/2` is now skipped on the `:stale` path.** Deliberate and correct when the replacement won — the channel belongs to the live conversation. But in the deleted/reassigned cases nothing restores it and nothing owns it either. If that is acceptable, say so in the comment; if it is not, it is a follow-up, not this PR.

PR #1832, opened 2026-09-10
feat(runtimes): Fountain.CommandRuntime, the acp runtime's module (#1634)

jhgaylor:
Every runtime Fountain has is an LLM coding agent. A deterministic program
that wants a warm machine, a vault, a thread, a schedule and a human who can
read what happened has nowhere to run: `Managoat.Runtimes` is a closed map of
four CLIs, each with a pinned adapter, a provider and an inference credential.

This adds the runtime module and the host dispatch that resolves it, and
nothing else. `"acp"` is not yet a value `Agent.changeset/2` accepts and the
`runtime_command` column it reads does not exist, so no request reaches this
code until the next PR in the stack opens the door.

- `Fountain.CommandRuntime` implements `Managoat.Runtimes` for a command the
  agent names. It writes no config, prepares no sandbox and builds no legacy
  command, so `Provisioning`'s `function_exported?` guards skip it. Its
  `default_env/2` ignores the credentials it is handed and exports one
  variable, `FOUNTAIN_SKILLS_DIR`.
- `Fountain.RuntimeDispatch` answers `acp` itself: `for_agent/1`,
  `acp_enabled?/1` and an `install/3` that has nothing to install. It gains
  `acp_model/2` (always nil for `acp`, so the pin path that reports
  `model`/`failed` is unreachable there), and the two predicates the schema
  needs next, `model_required?/1` and `command_required?/1`.
- `command/1` becomes `command/2`. The argv of this runtime is a property of
  the agent rather than of the runtime name, so the agent has to be in hand;
  every other clause ignores it. `TurnMachine.command/7` passes it through.

---

| | PR | What it is |
|---|---|---|

### DIFF
--- a/apps/fountain/lib/fountain/command_runtime.ex
+++ b/apps/fountain/lib/fountain/command_runtime.ex
@@ -1,0 +1,110 @@
+defmodule Fountain.CommandRuntime do
+  @moduledoc """
+  The `acp` runtime: launch the command the agent names and speak ACP to it.
+
+  Every other runtime Fountain has is a coding agent driven by a model.
+  `Managoat.Runtimes` knows four of them by name, installs a pinned adapter
+  for each and hands each one an inference credential. This one knows
+  nothing. The agent carries a `runtime_command`, Fountain runs it inside the
+  sandbox, and whatever comes back over stdio is the Agent Client Protocol
+  (ADR 0014) exactly as it is for claude or codex. That is the whole runtime.
+
+  It exists for a deterministic program that wants what Fountain gives an
+  agent and not what a model gives one. A convergent operation on a schedule,
+  inside a persistent sandbox that holds the repo and the toolchain, with the
+  run readable as a turn in a teammate's thread (#1634).
+
+  ## The name
+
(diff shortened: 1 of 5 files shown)


lex00 (approved):

jhgaylor:
`Build and Test (partition 6)` failed on `Fountain.TeamTest` "a co-tenant of the retired computer keeps its own binding, either wake order" — filed as #1862, not this change.

It exits rather than asserts: a `ConversationServer` for a *different* conversation (`b01703b7`) outlives its SQL Sandbox owner, Horde replicates it through the CRDT, the supervisor shuts down mid-`add_children`, and this test's `start_child` is simply the next caller to get the corpse. Every frame in the trace names `Managoat.Runtimes.Claude`.

Evidence it is not this PR: #1834, #1836 and #1838 all contain these exact commits and passed the same partition in the same round, and this PR passed partition 6 on both of its previous rounds with identical content. It touches `command_runtime.ex`, `runtime_dispatch.ex`, `turn_machine.ex`, `provisioning.ex` and one new test file — none of them near Team, Horde or `ConversationSupervisor`.

Re-running the failed job.

PR #1833, opened 2026-09-10
feat(agents): agents.runtime_command, and acp becomes a runtime you can name (#1634)

jhgaylor:
Opens the door the previous PR built behind. `acp` joins the runtime enum, a
new `runtime_command` column carries the command it launches, and `model`
becomes optional for that one runtime.

- **The column.** `agents.runtime_command` is text: a value may be a whole
  shell line ("cd /srv/app && exec bin/agent acp"), and a length rule belongs
  in the changeset if one is ever wanted. `agents.model` loses its NOT NULL in
  the same migration, because whether a model is required is a rule about the
  pair of columns and belongs where both are in hand.
- **Both refusals are 422s naming the field.** A missing `runtime_command` on
  `acp`, or one carried by any other runtime, which resolves its own
  executable from a pinned table. A `runtime_command` that is stored and never
  run reads as configuration and is not.
- **A model is still checked when it is given.** Optional on `acp` and inert
  there, but it still has to parse and name a known provider, because a value
  that is stored and ignored is worse than one that is refused.
- **`model` is nullable on the wire**, in the response and in both request
  bodies, so an agent converted to `acp` can clear the model it no longer uses
  with `{"model": null}`. CastAndValidate runs before the changeset, so a
  non-nullable string there would reject the null with a 400 and leave the
  stale value on the row forever. `AgentRequest` drops `model` from `required`
  and the changeset enforces it instead, so a missing model is a 422 rather
  than a 400.
- Model suggestions return nothing for `acp`: advice to fill in a field that
  does nothing is worse than no advice.
- The export carries the field, so an account export stays a restorable copy.

Nothing runs a turn on this runtime any better than before this PR — the
dispatch that spawns it landed below, and the skills mount, the refusal path,
the console and the docs are above.

`sdk-no-release`: `contract.json` and the generated TypeScript types move
here, and the version bump is the last PR in the stack.

---

| | PR | What it is |
|---|---|---|

### DIFF
--- a/apps/fountain/lib/fountain/agents.ex
+++ b/apps/fountain/lib/fountain/agents.ex
@@ -373,6 +373,7 @@ defmodule Fountain.Agents do
     :system,
     :model,
     :runtime,
+    :runtime_command,
     :sandbox_provider,
     :sandbox_mode,
     :skills,
--- a/apps/fountain/lib/fountain/agents/agent.ex
+++ b/apps/fountain/lib/fountain/agents/agent.ex
@@ -4,12 +4,17 @@ defmodule Fountain.Agents.Agent do
 
   alias Fountain.Accounts.User
   alias Fountain.Environments.Environment
+  alias Fountain.RuntimeDispatch
   alias Managoat.Runtimes.Model
 
   @primary_key {:id, :binary_id, autogenerate: true}
   @foreign_key_type :binary_id
 
-  @runtimes ~w(claude codex gemini opencode)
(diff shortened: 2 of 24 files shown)


lex00 (approved):

jhgaylor:
The first run of `A change to the SDK says whether it releases` failed because the run started about two minutes before the `sdk-no-release` label landed on this PR. The label is what the gate's own error message asks for when a PR changes the published surface without releasing, and the version bump is #1839. Re-running.

PR #1836, opened 2026-09-10
fix(conversations): refuse a turn whose agent no longer carries a command (#1634)

jhgaylor:
An agent's deletion nilifies its conversations' pointer, and on the `acp`
runtime the argv lives on the agent. A live `ConversationServer` could
therefore be asked for a turn it had nothing to spawn for, and
`RuntimeDispatch.command/2`'s `{:ok, argv} = CommandRuntime.argv(agent)` raised
a `MatchError` inside the server, after a turn row already existed. Every
other runtime resolves its argv from a pinned table and cannot reach this.

`TurnMachine.open/4` refuses it instead, before a turn row is written, for the
same reason `:at_capacity` is refused there: there is nothing to record, and a
`turn`/`failed` stage event says what happened and what to do about it. The
server treats `:no_command` as it treats `:at_capacity` — stay idle, drop
nothing — while `{:error, _}` keeps dropping the connection as before.

Backing the guard out puts the `MatchError` back in the one test that covers
this and leaves the other ten green.

---

| | PR | What it is |
|---|---|---|

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -2299,9 +2299,9 @@ defmodule Fountain.Conversations.ConversationServer do
   defp kick_turn(state, prompt, agent, images) do
     state = touch_activity(state)
 
-    case TurnMachine.open(state.conversation_id, state.sandbox_id, prompt) do
+    case TurnMachine.open(state.conversation_id, state.sandbox_id, prompt, agent) do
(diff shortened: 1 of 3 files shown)


lex00 (approved):

jhgaylor:
The `Swift SDK (ubuntu-24.04)` failure on the first run is #1861, not this change. This stack touches no Swift file, and #1837, #1838 and #1839 are stacked on top of this branch — so each ran a strict superset of this tree, in the same three minutes, and all three passed. Re-running the job with that recorded.

PR #1809, opened 2026-09-10
Publish the TypeScript SDK as @managoat/fountain-sdk (hold for npm setup)

jhgaylor:
**Draft, because merging this publishes** — and the npm side needs two things from you first (below).

The package follows the repo into the org: `@agentshit/fountain-sdk` → **`@managoat/fountain-sdk`**, 1.24.0 continuing the same version line so a consumer changes the dependency and the import specifier and nothing else.

This is not only tidiness. npm publishing here is **OIDC trusted publishing with no `NPM_TOKEN`**, and a trusted publisher names the owner, the repo *and* the workflow file — so the binding on `@agentshit/fountain-sdk` still names `BinaryBourbon/fountain`, and the next publish under the old name would fail exactly the way the PyPI one would have. I got this wrong earlier in the session when I said npm needed no configuration; provenance reads `package.json`, but the publisher binding is a separate thing and the transfer broke it.

**Before merging, on npm:**

1. The **`@managoat` scope must exist** and your account must be able to publish into it.
2. **`@managoat/fountain-sdk` needs a trusted publisher** naming `managoat/fountain` + `.github/workflows/sdk-publish.yml`. If npm won't let you configure a publisher for a package that doesn't exist yet (PyPI allows pending publishers; I couldn't verify npm's behaviour without your account), then the first version has to go up with a token once, after which OIDC takes over. Either way, merging before that is a failed publish.
3. Afterwards, worth doing: `npm deprecate "@agentshit/fountain-sdk" "moved to @managoat/fountain-sdk"`. Its published versions stay resolvable forever either way — it just stops receiving new ones.

**Renamed everywhere the name is live** (27 files): `package.json`, `package-lock.json`, the tsconfig path mapping, `src/http.ts`, `src/config.ts`, `src/node.ts`, the contract manifest, `sdk-publish.yml`'s messages and release tag, the SDK README and examples, docs (`sdk`, `quickstart`, `tour`, `open-source`, `build/team-chat`, the snippet), `README.md`, the `build-fountain-app` skill, and the runtime strings the console and `/llms.txt` render — including the onboarding and marketing tests that assert them.

**Left with the name it shipped under:** both CHANGELOGs' history and `decisions/`.

Also in here, because the next release needs it and it was the gap I flagged: a `[Unreleased]` entry for **the org move itself** — image path, tap, Go module paths, what redirects and what doesn't.

[...]

PR #1839, opened 2026-09-10
docs(runtimes): the acp runtime, ADR 0049 and SDK 1.26.0 (#1634)

jhgaylor:
The last PR of the stack. Everything the seven below it built is now
documented and released.

- **`docs/catalog/runtimes/index.md`** gains the runtime, a row in the
  comparison table, a skills-layout row, and *Running a program as an agent*:
  the manifest, the five things that follow from the command being a shell
  line, and how the turn is priced.
- **Two traps have their own paragraphs**, because both are silent and both
  cost a turn. Write `exec` before the program, or the login shell stays as the
  parent and an interrupt can leave the program running on a persistent
  sandbox. And nothing may print on stdout before the program starts, because a
  profile's banner lands in front of the first protocol message.
- **What "the same isolation" means is stated where it is weakest.** The free
  string is safe because it runs where an environment's setup script already
  runs. On `sandbox_provider: runner` with the default backend that is a
  directory and the daemon's own user, so the page says so and links to
  trusted mode.
- `docs/concepts/agent.md`, `docs/concepts/permissions.md` (both tables, since
  whether the command asks and how long a grant lasts are its decisions, not
  Fountain's), `docs/sdk.md`, `docs/reference/glossary.md`, `docs/build/index.md`
  and the three "all four runtimes speak ACP" lines.
- **SDK 1.25.0.** `Agent["model"]` is `string | null`, and the two request
  bodies take `string | null` optionally. A client that assumed a string needs
  a null check; nothing else on the wire changed shape.

The docs suite, `docs-style.py` (its two findings are pre-existing pages I did
not touch), vale and destink all pass.

---

| | PR | What it is |
|---|---|---|


lex00 (commented):
Two things before this one goes in:

- `CHANGELOG.md` line 50 has a stray `>>>>>>> theirs` marker that this commit introduced (it is in none of the source branches).
- `main` moved to SDK 1.25.0 with #1809 after this stack was cut, so this release collides; it needs 1.26.0, and #1849 then needs 1.27.0 or the other way round depending on which lands first.

Approving the rest of the stack.

jhgaylor:
Rebased onto the current `main`. #1809 landed while this stack's CI was running and took **1.25.0** for the `@managoat/fountain-sdk` rename, so this bump is now **1.26.0** and its changelog entry sits above main's. All four edit sites moved together: `package.json`, `package-lock.json` (two spots), `src/http.ts`'s `USER_AGENT` and `CHANGELOG.md`.

Worth noting the collision was invisible to the tooling: GitHub reported `MERGEABLE` the whole time, because two releases claiming the same version number is a semantic conflict and not a textual one.

PR #1852, opened 2026-09-10
feat(agents): permission_policy accepts ask_timeout, a key that names no tool (#1635)

lex00:
Re-cut of #1678 as a stack of nine on `main`, under the no-big-PRs rule. This is the policy half: `ask_timeout` as the one permission_policy key that is seconds rather than a verdict, stripped before the map reaches the library and narrowed the same way. Nothing reads it yet.

Part 1 of 9 for #1635.


jhgaylor (dismissed):
Requesting changes — two mechanical blockers, nothing wrong with the substance.

`Fountain.PermissionPolicy` is the right shape for this. Pulling the non-tool keys out into one module, with `verdicts/1` as the single chokepoint before anything reaches `Managoat.ACP.Permissions`, means the library never has to learn that some keys aren't tools. I checked every `ACP.Permissions` call site on the stack tip and they all go through it. The narrowing direction (shorter wait is stricter, and an agent that set no `ask_timeout` is bounded by the global ceiling rather than by nothing) is the non-obvious half and it's both right and explained.

**1. Conflicts with `main`.** `apps/fountain/lib/fountain/agents/agent.ex` — `main` gained `alias Fountain.RuntimeDispatch` where this branch adds `alias Fountain.PermissionPolicy`. Both are wanted; it's an alias-ordering conflict only. The rest of the stack rebases clean apart from the SDK changelog on #1860.

**2. `A change to the SDK says whether it releases` is red.** This PR changes `sdk/typescript/src/generated/openapi.ts` but leaves the version alone, because the bump is deliberately saved for #1860. That sequencing is right — concurrent bumps racing on `main` is the failure mode we've hit before. The gate's own escape hatch is the fix: add the **`sdk-no-release`** label to this PR and to #1859. Happy to apply it if you'd rather I did.

Nit, non-blocking: `reserved_errors/1` here (and `validate_reserved_keys/1` in `conversations.ex`) matches the literal `"ask_timeout"` rather than walking `PermissionPolicy.reserved_keys()`. Today that's the only entry so the behaviour is identical, but a second reserved key would be stripped from verdict validation by `verdicts/1` and then validated by nothing at all — it'd be accepted with any value. Iterating `reserved_keys/0` with a per-key validator closes that off.

BinaryBourbon (changes requested):
Reviewed the whole `stack:1635` set (#1852 → #1860) against issue #1635, not just this PR. The design is right and the stack is cut well: the detach seam is the turn row rather than a process, so the deadline survives the suspend the feature exists to allow; nothing is resolved before the gates that would refuse to deliver it; and the connection-drop reasoning — both shippable adapters number requests from 0 *per turn*, so a kept peer would swallow the resume turn's first request against a stale hold — is the kind of thing usually found in production rather than in review. CI is green on all nine.

One blocking finding. It lives here because the validator does.

## `ask_timeout` has no upper bound, and the deadline it produces crashes the server that writes it

`valid_ask_timeout?/1` accepts any positive integer, and the five schema blocks declare `minimum: 1` with no `maximum`. The moduledoc declines a ceiling deliberately:

> Deliberately not bounded above: the point of a detached request is that it can wait for days, and a ceiling here would be a second, quieter copy of the idle bound this key exists to escape.

That is right about the idle bound. It is not right about `timestamp`. The value reaches `turns.permission_deadline` via `DetachedRequest.deadline/2`, and Postgres timestamps stop at 294276 AD.

Measured on `stack/1635-9-docs`:

```
create_agent(permission_policy: %{"ask_timeout" => 999_999_999_999_999})
  -> :ok   (accepted and stored)

DetachedRequest.deadline(999_999_999_999_999 * 1000)
  -> #DateTime<31690765-03-15 06:36:04Z>

Pending.detach(%Pending{}, turn)
  -> ** (FunctionClauseError) no function clause matching in
        Postgrex.Extensions.Timestamp.encode_elixir/1
```

`Pending.detach/2` writes that row as `{:ok, turn} = Conversations._unsafe_update_turn(...)`, inside `apply_effect(state, :detach_permission)`, inside the `handle_info` that processes the `waiting` frame. So the raise is uncaught in a GenServer callback: the `ConversationServer` dies *after* `:detach_permission` and *before* `{:finish, …}`, leaving the turn `running` with `pending_permission` set and `waiting` false — the one row shape that neither `Pending.holds?/2`, nor the sweep's `waiting == true` query, nor `restore_permission_timer/2` will ever resolve.

Reachable two ways, both tenant input: `POST /api/agents` with `permission_policy.ask_timeout`, and `_meta.fountain.timeout` on the request itself, which is unbounded on its own whenever the policy names no `ask_timeout` (`DetachedRequest.timeout_ms/2`, the `{request, nil}` clause). Blast radius is the tenant's own conversation, so this is low severity and not a security issue. It is still a crash on a value the door accepted, and the absurd-input shape is exactly what a fuzzed SDK client sends.

The fix is a ceiling that is visibly not the idle bound — 365 days, or anything under ~9.2e12 seconds — applied where the seconds are parsed rather than where they are spent, so both halves get it. Which is also the answer to the next point.

## `to_positive_seconds/1` here and `positive_seconds/1` in #1854 are the same function

`Fountain.PermissionPolicy.to_positive_seconds/1` (`permission_policy.ex:134`) and `DetachedRequest.positive_seconds/1` (`detached_request.ex:203`) are byte-identical, three clauses each. `DetachedRequest` already depends on this module's concepts; one clamped `seconds/1` exported from here and called from both is the de-duplication and the ceiling in the same edit. Left a matching note on #1854.

## Non-blocking: the schema fragment is now copy-pasted five times

The `ask_timeout` property plus the `oneOf` union is repeated verbatim across `AgentResponse`, `AgentRequest`, `AgentUpdateRequest`, `Conversation` and `ConversationCreateRequest` — five copies of a 22-line block carrying the same 8-line description. The duplication predates this PR (the `additionalProperties` enum was already 5×), but the fragment roughly triples in size here, and the file's own idiom is a named nested schema module (`UsageTotal`, and `PendingPermissionRequest` in #1859). A `PermissionPolicy` schema module referenced in all five would collapse it. Not asking for it in this PR: it turns the inline object into a `$ref` and renames the generated TS type, which is a contract decision of its own.

Everything else here reads correct. `verdicts/1` as the single chokepoint into `Managoat.ACP.Permissions` is the right seam, and `permission_policy_test.exs` proving *why* it exists — that a reserved key left in comes back out of `effective/2` as `"ask_timeout" => "auto_deny"` — is the test I would have asked for. `keep_reserved/2` catching the console form silently dropping a key it cannot render is a nice catch. The narrowing direction (shorter is stricter, and the global ceiling standing in where the agent set nothing) is argued correctly and covered both ways.

Happy to re-approve on the ceiling alone.

jhgaylor:
Pushed `4529ce85`. The ceiling is in, and it is one parser rather than two.

**`Fountain.PermissionPolicy.seconds/1`** is now the only place a number of seconds is read on this path, bounded at `max_ask_timeout_seconds/0` = 365 days. The moduledoc says why that is not the thing I argued against — the idle bound is minutes and exists because a held request keeps a sandbox from parking, which a detached request does not do; this one exists because the value lands in a `timestamp` column that stops at 294276 AD. A year is further out than anything the feature is for, so hitting it reads as a typo caught rather than a limit imposed. Your measurement is in the commit message and in the test comments, since the failure mode is worth not rediscovering: the raise is inside the `handle_info` that detaches, so the server died after `:detach_permission` and before `{:finish, …}` and left the turn in the one row shape nothing resolves.

Details:

- **Refused at the doors, nil everywhere else.** Both doors that store a policy now reject an over-ceiling value (`valid_ask_timeout?/1` → `seconds/1`), and the changeset message names the bound, because "not a positive number of seconds" about 1e15 reads as a lie. Every read path treats unreadable as nil, so a value nothing can parse falls back to the global ceiling rather than to no bound at all — there is a `check_narrows/3` test for exactly that, so "the agent's value is nil" can never come to mean "the launch may ask for anything".
- **The ceiling is inclusive**, tested at `max` and `max + 1`, so the refusal is a bound and not an off-by-one.
- **`maximum:`** added to all five schema blocks, and the regenerated `openapi.ts` is description-only — the type surface does not move, so `sdk-no-release` still holds. `contract.json` is unchanged, since it records types and requiredness rather than numeric bounds.

The duplicate parser is gone on #1854, which now reads `_meta.fountain.timeout` through this one — that half is the one that needed it more, since it has no door to be refused at.

I checked the tests by weakening only the ceiling and leaving the refactor in place, so the failures are behavioural rather than "function undefined": `create_agent` returned `{:ok, …}` where it now errors, and the `_meta` half produced the year-31690765 deadline again.

Local: `mix format --check-formatted`, `mix compile --warnings-as-errors`, `mix credo --strict` (no issues), and 370 tests green at the stack tip across `permission_policy`, `agents`, `conversations_start`, `pending`, `detached_request`, `conversation_server_acp`, the size pin and the conversation controller. CI is running on all nine; the chain is rebased and every PR is `MERGEABLE`.

Left alone deliberately: the five-way schema duplication, for the `$ref` reason you gave. Also still open is my own earlier nit — `reserved_errors/1` and `validate_reserved_keys/1` match the literal `"ask_timeout"` rather than walking `reserved_keys/0`. One reserved key exists, so the behaviour is identical today; a second one would need a per-key validator anyway, which is more design than this PR should carry. Worth an issue if a second key ever gets proposed.

BinaryBourbon (approved):
Approving. `4529ce85` fixes the blocking finding, and fixes it in the place I asked for rather than at the call site.

## What I checked

`seconds/1` is the single parser now, `to_positive_seconds/1` is gone, and `valid_ask_timeout?/1` and `ask_timeout_seconds/1` both route through it. Both doors that store a policy refuse an over-ceiling value — `Agent.reserved_errors/1` with a message that names the bound, and `Conversations.validate_reserved_keys/1` with `:permission_policy_invalid` — and every other reader takes it as nil and falls back to the global ceiling rather than to no bound. That is the right direction: a stored value nothing can parse becomes a shorter wait, never a longer one.

I reverted the constant to check the tests bite. With `@max_ask_timeout_seconds` at `99_999_999_999_999_999`, five of the new assertions fail:

```
1) seconds/1 a value past the ceiling would not fit in a timestamp, so it is not read
2) seconds/1 the ceiling is a year, and it is inclusive
3) the deadline the ceiling itself is honoured, so the bound is not an off-by-one
4) the deadline the sandbox cannot name a deadline no timestamp can hold
5) permission_policy (#939) an ask_timeout past the ceiling is refused at the door, not at the deadline
```

The ones that survive are the `max + 1` cases, which are consistency checks and cannot fail by construction. That is fine, because `assert max == 365 * 24 * 60 * 60` pins the constant itself and the literal `999_999_999_999_999` cases pin the property. Between them the guard cannot quietly stop guarding.

162 tests, 0 failures across `permission_policy_test.exs`, `agents_test.exs`, `detached_request_test.exs` and `conversations_start_test.exs` on the stack tip.

I also checked the `sdk-no-release` claim rather than taking it. The `openapi.ts` diff is five description strings and nothing else, and `sdk/contract/contract.json` records no `maximum` at all — so `scripts/sdk-contract/build.sh --check` reads the same artifact it did before and the label holds. `maximum` does reach the served spec, so an API caller gets a 422 from `OpenApiSpex` before the context sees the value, which is the door I would rather it met.

The reasoning in the moduledoc is the part worth keeping. "The ceiling, and what it is not" says the thing that was actually wrong with the old comment: the old one declined a ceiling for a true reason about the idle bound, and that reason has nothing to do with `timestamp`. A reader who only remembers "no ceiling, on purpose" now finds out why there is one.

## Standing, non-blocking

The five-times-duplicated schema fragment grew again — `maximum:` plus three more lines of description in each of `AgentResponse`, `AgentRequest`, `AgentUpdateRequest`, `Conversation` and `ConversationCreateRequest`. Same note as last time: a named `PermissionPolicy` schema module would collapse it, and that is a `$ref` and a generated-type rename, so it is a change of its own and not this PR's.

## The rest of the stack

I diffed each of #1855 through #1859 against the commit I approved and their own diffs are byte-identical; they moved only by carrying `4529ce85` and `b5917356` under them. Those approvals stand without a re-read. #1853 is unchanged. #1854 and #1860 get their own notes.

PR #1853, opened 2026-09-10
refactor(conversations): the pending family's state round trip moves into Pending

lex00:
The five private round-trip adapters in the server become `Pending` functions over the server's state. No behaviour changes; the server shrinks.

Part 2 of 9 for #1635, on #1852.

Part of #1635

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -1538,17 +1538,8 @@ defmodule Fountain.Conversations.ConversationServer do
   # another, so a second answer to the same request is "too late" rather
   # than an error in the caller.
   def handle_call({:answer_permission, request_id, option_id}, _from, state) do
-    {reply, turn, pending} =
-      Pending.answer_permission(
-        Pending.from_state(state),
-        state.conversation_id,
-        state.current_turn,
-        state.acp_peer,
-        request_id,
-        option_id
-      )
-
-    {:reply, reply, %{Pending.into_state(state, pending) | current_turn: turn}}
+    {reply, state} = Pending.answer(state, request_id, option_id)
+    {:reply, reply, state}
   end
 
   # A call needs a turn to belong to, and the client following that turn is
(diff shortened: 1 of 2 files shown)


jhgaylor (approved):
Approving. Good subtraction: five private adapters in the server collapse into five `Pending` functions over state, `over/2` carries the `from_state` → operate → `into_state` + write-back round trip once instead of per call site, and the server loses the boilerplate. Exactly the shape #1369 asks for. Green across the board.

One nit, and it's about stack hygiene rather than the code:

`restore_permission_timer/2` gains a `%{waiting: true} -> nil` clause and a docstring paragraph about detached requests — but `Turn` has no `:waiting` field until #1854. At this commit the pattern can never match (a `%Turn{}` without the key simply fails the match), so the clause is inert and nothing in the suite can exercise it. A PR titled as a pure refactor is carrying a slice of the next one's feature.

Not worth a re-push on its own if it costs you a rebase of everything above it — the clause is harmless where it sits and it's correct the moment #1854 lands. Flagging it because we've had the same forward-reference pattern in the labels stack, and the cheap version is to move both the clause and the docstring paragraph into #1854 where the field and a test for it exist together.

BinaryBourbon (approved):
Approving. A clean subtraction: five private adapters in `ConversationServer` that each did `from_state` → operation → `into_state` → write the turn back, collapsed into `over/2` and five named entry points on the family that owns them. `conversation_server.ex` loses 25 lines and the size pin drops with it in #1855, which is the direction that file is only allowed to move.

Two things I checked rather than assumed:

- `Pending.answer/2` is not written through `over/2`, because it returns a reply as well as state. Right call — bending `over/2` to carry a third element would cost more than the one explicit function.
- `restore_permission_timer/2` gaining a `%{waiting: true} -> nil` clause is forward-referencing #1854's flag, but the docstring says so plainly ("The reattach path only ever passes a `running` turn, and a detached one is `completed`, so this is the belt to that braces") rather than describing unbuilt behaviour as live. That is the CLAUDE.md rule satisfied, and the belt is worth having: arming the in-turn timer over a detached request would deny it at the five-minute ceiling the detach exists to escape.

No findings.

PR #1854, opened 2026-09-10
feat(conversations): the waiting turn row, and DetachedRequest (#1635)

lex00:
Two columns on `turns` (`waiting`, `permission_deadline`) and `Fountain.Conversations.DetachedRequest`, the pure half: the deadline rule, the fail-closed option check, and the one-line resume prompt. Nothing writes the columns yet.

Part 3 of 9 for #1635, on #1853.

Part of #1635


jhgaylor (approved):
Approving. Clean, self-contained, green.

The migration is the part I looked hardest at and it's right. `pending_permission` genuinely cannot distinguish "held inside a running turn" from "outlived one" — the row is identical — so `waiting` earning its own column rather than a key inside the JSON is correct, and the partial index over it is exactly the sweep's query. `permission_deadline` as a column for the same reason: a process timer cannot survive the suspend this whole path exists to allow, and the comment says so.

`DetachedRequest` reads well as a value module. Three details worth calling out as right:

- `positive_seconds/1` returning nil rather than 0 for garbage, so a typo in `_meta.fountain.timeout` falls back to the policy instead of becoming "deny immediately".
- `timeout_ms/2` taking the **smaller** of request and policy. The request half is written inside the sandbox and the policy half is the tenant's, so an agent bounding its own wait is fine and an agent extending the tenant's is not. That asymmetry is the whole security property here and it's both implemented and explained.
- `deny_option_id/1` going through `Permissions.deny_outcome/1` so expiry can only ever name an option the agent itself offered — the fail-closed rule the in-turn path gets from the peer, reapplied where there's no peer left to apply it.

`request_line/2` gating the JSON decode behind a `String.contains?` so an ordinary turn's thousands of updates pay a substring search is a nice touch.

BinaryBourbon (changes requested):
Requesting one change, which is the other half of the finding on #1852. The design here I have no argument with — notes below.

The row-not-a-timer decision is the load-bearing one and it is argued in the right place: a two-day wait outlives every process involved, so `permission_deadline` is a column with a partial index over `waiting`, and the migration comment says exactly why `pending_permission` alone could not have been the discriminator (the row looks identical whether the request is held inside a running turn or has outlived one). `waiting` being the thing every sweep reads, and a waiting turn being `completed` so nothing hunting `running` rows finds it, is the detail that makes the rest safe.

The resume-prompt-as-JSON-line choice is the one I expected to argue with and did not. `Managoat.ACP.Peer.prompt/3` takes text and images and offers no protocol-extension hook, and keeping the payload in the exact shape a `_meta` key would carry, under the same name, makes the eventual move a move rather than a redesign. Worth keeping that sentence in the moduledoc.

`request_line/2` is nicely defensive: the substring guard before `classify_line/1` so an ordinary turn's thousands of updates pay a `String.contains?` and nothing else, and `params_for/2` matching on the minted id so a wrong-order relay degrades to `nil` params and the policy fallback rather than to a wrong deadline. That ordering is a guarantee of the `managoat_acp` package rather than of this repo, so I am glad the fail-safe is there and that #1855's `_meta` test would catch a regression in it.

**The change:** `positive_seconds/1` (`detached_request.ex:203`) is byte-identical to `Fountain.PermissionPolicy.to_positive_seconds/1` (`permission_policy.ex:134`), and both are unbounded above. As measured on #1852, `ask_timeout`/`_meta.fountain.timeout` of 1e15 seconds produces a year-31690765 deadline, which `Pending.detach/2` then hands to Postgrex and crashes the `ConversationServer` on. `timeout_ms/2`'s `{request, nil}` clause is the sandbox-supplied half of that: with no policy `ask_timeout` to `min/2` against, the request's own value stands unbounded. One clamped `seconds/1` exported from `Fountain.PermissionPolicy` and called from here fixes the duplication and both reachable paths in the same edit. Details and the measurement are on #1852.

Nothing else. `deny_option_id/1` deferring to `Permissions.deny_outcome/1` so expiry can never name an option the agent did not offer, and `offered?/2` applying the peer's fail-closed rule where no peer is left to apply it, are both right.

jhgaylor:
Pushed `b5917356`. `positive_seconds/1` is gone; `timeout_seconds/1` reads `_meta.fountain.timeout` through `Fountain.PermissionPolicy.seconds/1`, which is the bounded parser added on #1852.

You were right that this is the half that needed it. `ask_timeout` arrives at a door that validates and can refuse with a message; `_meta.fountain.timeout` arrives on a `session/request_permission` that has already been raised, so there is nothing to refuse it *to*. A parser of its own is precisely how this half would have kept the unbounded behaviour after the other half was fixed — and `timeout_ms/2`'s `{request, nil}` clause has no policy value to `min/2` against, so with no ceiling the sandbox alone could name 1e15 seconds and crash the server writing the row.

Over-ceiling falls back rather than clamping, which is what every other unreadable value here already did: to the policy, else the global ask timeout. Falling back is the shorter wait, and short is the safe direction.

Two tests, both verified to fail with the ceiling weakened and the refactor kept:

- the sandbox cannot name a deadline no `timestamp` can hold — asserts the fallback with a policy set *and* with none, since the `min/2` in the first case would otherwise mask it;
- the ceiling itself is honoured, asserting `deadline(max * 1000).year < 294_276`, so the bound is checked against the thing that actually broke rather than against itself.

`deadline/2`'s docstring now records what bounds it and what happened when nothing did.

Nothing else in this PR moved — the row, the index, the wire shape of the resume prompt and `request_line/2` are as you reviewed them.

BinaryBourbon (approved):
Approving. `b5917356` is the other half, and it closes the path that mattered more.

`positive_seconds/1` is gone and `timeout_seconds/1` reads `_meta.fountain.timeout` through `PermissionPolicy.seconds/1`, so the `{request, nil}` clause in `timeout_ms/2` — the one with no policy value to `min/2` against — is now bounded by the same ceiling as the policy half. That was the reachable-with-no-door case: a `session/request_permission` has already been raised by the time we see it, so there was nowhere to refuse it to, and a parser of its own is exactly how this half would have kept the unbounded behaviour after #1852 was fixed.

Over-ceiling falling back to the policy, then to the global ceiling, rather than clamping to the ceiling, is the right call and the docstring says why: falling back is a shorter wait, and short is the safe direction. Same rule as every other unreadable value on this path, so there is nothing new for a reader to learn.

The two new tests earn their place. The first asserts both the symbolic `max + 1` and the literal `999_999_999_999_999`, and with the ceiling reverted both fail; the second checks the ceiling itself is storable and that `deadline(max * 1000).year < 294_276`, which is the assertion that ties the bound to the reason for it rather than to a number somebody picked.

## Non-blocking: the crash site itself is still unguarded

`Pending.detach/2` reads the deadline back off the row:

```elixir
request
|> Map.get("detached_timeout_ms")
|> case do
  ms when is_integer(ms) and ms > 0 -> ms
  _ -> Lifecycle.ask_timeout_ms()
end
|> DetachedRequest.deadline()
```

That guard is unbounded, and it is the exact line that raised. It is unreachable with a bad value now, because `Pending.ask/7` is the only writer of `detached_timeout_ms` and it goes through `timeout_ms/2`; the stack lands as one merge, so there is no mixed-version window where an old writer meets a new reader. So this is defence in depth and not a bug. But the clause is one `and ms <= PermissionPolicy.max_ask_timeout_seconds() * 1000` away from the consumer refusing what the producer already refuses, and the cost of getting it wrong here is a dead `ConversationServer` rather than a short wait. Your call whether it is worth the line.

Everything else from the previous read stands — the row-not-a-timer decision, `request_line/2`'s substring guard, and `params_for/2` degrading to `nil` params on a wrong-order relay.

PR #1855, opened 2026-09-10
feat(conversations): a turn that ends waiting leaves its request open (#1635)

lex00:
The `waiting` stop reason: the request is detached instead of denied, the turn ends `completed`, the deadline goes on the row and the connection is dropped so the resume turn's first request cannot collide with the held one. One deviation from #1678: `waiting` joins `end_turn` as a completed stop reason, because 8a05804c narrowed that rule after the branch was cut.

Part 4 of 9 for #1635, on #1854.

Part of #1635

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -21,7 +21,7 @@ defmodule Fountain.Conversations.ConversationServer do
   }
 
   alias Fountain.Conversations.{CallbackKey, Checkpoints, CodexChatGPT, Connection}
-  alias Fountain.Conversations.{Conversation, Egress}
+  alias Fountain.Conversations.{Conversation, DetachedRequest, Egress}
   alias Fountain.Conversations.{Lifecycle, McpServers, Output, Pending, Provisioning, Reapply}
   alias Fountain.Conversations.{Reattachment, Redaction, SpriteEnv, TurnMachine}
 
@@ -487,6 +487,10 @@ defmodule Fountain.Conversations.ConversationServer do
       acp_peer: nil,
       # Timer refusing an unanswered permission request (#940).
       permission_timer: nil,
+      # The last `session/request_permission` the peer relayed, as
+      # `DetachedRequest.request_line/2` reads it (#1635): the ask that follows
+      # carries no params, and the per-request timeout is in theirs.
+      acp_request_params: nil,
       # Caller-tool calls parked on the turn (#1202): id => %{name, arguments,
       # turn_id, waiter, timer, result}. `result` is nil while parked and the
       # answer once resolved; entries are dropped when the turn ends.
@@ -1456,9 +1460,14 @@ defmodule Fountain.Conversations.ConversationServer do
   # editor (#708) are peer clients of this door, not fallbacks for one
   # another, so a second answer to the same request is "too late" rather
   # than an error in the caller.
+  # `Pending.holds?/2` is the guard a detached request needs (#1635).
   def handle_call({:answer_permission, request_id, option_id}, _from, state) do
-    {reply, state} = Pending.answer(state, request_id, option_id)
-    {:reply, reply, state}
+    if Pending.holds?(state.current_turn, request_id) do
+      {reply, state} = Pending.answer(state, request_id, option_id)
+      {:reply, reply, state}
+    else
+      {:reply, {:error, :no_pending_permission}, state}
+    end
   end
 
   # A call needs a turn to belong to, and the client following that turn is
@@ -1692,7 +1701,7 @@ defmodule Fountain.Conversations.ConversationServer do
   end
 
   def handle_info({:permission_timeout, request_id}, state) do
-    {:noreply, Pending.resolve(state, request_id, "timeout", nil)}
+    {:noreply, Pending.resolve_if_held(state, request_id, "timeout", nil)}
   end
 
   # The caller never answered a parked tool call (#1202). The agent gets an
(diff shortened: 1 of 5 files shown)


jhgaylor (dismissed):
Requesting changes. The design is right — one verified bug and one format break.

The reasoning for dropping the connection at detach is the best comment in the stack. Both shippable adapters numbering `session/request_permission` from 0 per turn means the resume turn's first request collides with the peer's stale hold and gets swallowed silently — no response, no card, no timeout, an agent blocked with `SANDBOX_MAX_LIFETIME_HOURS` off by default. That's a trap worth the paragraph it got. Effect ordering (`:detach_permission` before `{:finish, …}` so `finish/4` sees the marked row, `{:drop_connection, …}` after) is correct, and the size pin moving 2774 → 2749 is the right direction.

---

**1. A stale in-turn permission timeout resolves a detached request's card. (verified)**

`handle_call({:answer_permission, …})` gained a `Pending.holds?/2` guard in this PR. The two `handle_info({:permission_timeout, request_id}, state)` clauses did not, and they need the same one.

`Pending.detach/2` calls `Process.cancel_timer/1`, which does not recall a message already in the mailbox. If the five-minute in-turn timer fires in the window before the `{:done, "waiting", …}` frame is processed, `Pending.resolve(state, request_id, "timeout", nil)` runs against a turn that has already detached.

I reproduced it against the stack tip by sending `{:permission_timeout, request_id}` to the server immediately after the detach:

```
PROBE request/done stages: [
  %{"option_id" => nil, "outcome" => "timeout", "request_id" => "499.20a8434…"}
]
PROBE audit rows: [
  %{action: "conversation.permission_denied", actor: "sprite",
    metadata: %{"tool" => nil, "verdict" => "timeout"}}
]
PROBE pending_requests after stale timeout: [ %{options: [...], deadline: ~U[...], ...} ]
```

The good news is the turn row survives — `state.current_turn` is already nil, so `resolve_permission/7` takes its `nil -> nil` branch and never writes. The request stays answerable and the sweep still owns it. What's damaged is everything around it:

- a `request`/`done` event with `outcome: "timeout"` goes out, so every attached client resolves the permission card as denied while the request is in fact still open and answerable. The client has no reason to ever show it again.
- a `conversation.permission_denied` audit row is written for a denial that did not happen, with `tool: nil` because `pending_tool(nil)` has nothing to read. 0013 is explicit that the trail records only what happened.

Guarding both clauses with the `holds?/2` you already wrote covers it:

```elixir
def handle_info({:permission_timeout, request_id}, state) do
  if Pending.holds?(state.current_turn, request_id) do
    {:noreply, Pending.resolve(state, request_id, "timeout", nil)}
  else
    {:noreply, state}
  end
end
```

Worth a test alongside the existing detach ones — it's a one-liner to provoke, since the message can just be sent.

Separately, and pre-existing rather than yours: `resolve_permission/7` never checks that `request_id` matches the request on the turn. `holds?/2` is the natural place that check now lives, which is another reason to route the timeout path through it.

**2. `mix format --check-formatted` fails**, and it cascades to #1856 and everything above.

```
apps/fountain/test/fountain/conversations/conversation_server_acp_test.exs
1844 -|      (stray blank line before the closing `end`)
```

BinaryBourbon (approved):
Approving. This is the heart of the stack and it is careful in the places that matter.

The three decisions I went looking for are all made and all documented:

- **`waiting` is `completed`, not a fourth status.** `status = if stop_reason in ["end_turn", "waiting"]` keeps the `8a05804c` rule that an unrecognised stop reason is a failure, and `waiting?/2` gates the detach on a request actually being held — so a `waiting` frame with nothing outstanding is an ordinary completed turn, which is both the sane reading and a test here.
- **The connection goes with the detach.** The argument in the `handle/3` comment is the one that earns the departure from #817's sandbox-scoped connection: the peer's single `pending_permission` slot is cleared only by an answer or a denial, the detached path sends neither, and both adapters number requests from 0 per turn — so a kept peer matches the resume turn's first request against the stale hold and swallows it. No response, no card, no timeout, and `SANDBOX_MAX_LIFETIME_HOURS` is 0 by default, so nothing reclaims it. #1857's colliding-id test is the regression guard for exactly that, and it asserts the *card and the timer*, not just the id.
- **`resolve_if_held/4` for the in-flight timer.** `Process.cancel_timer/1` does not recall a message already in the mailbox, so the in-turn timeout can land against a turn that has since detached; resolving it would publish a `request`/`done` denying a request that is still open and write a `conversation.permission_denied` row for something that did not happen. The test drives the real mailbox rather than asserting on the guard, which is the version of that test worth having.

`waiting_meta/1` prefixing `waiting_request_id` rather than emitting a bare `request_id` on a `turn` event is a small thing that would have been a real client bug — a card pairing on `request_id` would have settled against an event that is not about it — and the test asserts the absence (`refute Map.has_key?(done, "request_id")`), not just the presence.

One non-blocking nit: the six `Pending.ask/7` call sites updated in `pending_test.exs` pass `60_000` where the seventh parameter is now `params :: map() | nil` (the request's `session/request_permission` params). It is harmless — `DetachedRequest.timeout_seconds/1` falls through to its catch-all, so those tests exercise the policy/ceiling fallback and none of them assert on `detached_timeout_ms` — but it reads as if the argument were still a timeout, and nothing catches it: dialyzer runs `MIX_ENV=dev` and never sees test files. `nil` at those six sites would say what is meant.

Also worth noting for whoever merges: `Pending.ask/7` now does two DB reads per ask (`_unsafe_get_conversation!` + `agent_for/1`) to resolve the effective `ask_timeout`. Asks are rare enough that this is fine, and the ownership comment is the correct GenServer-established-ownership shape, so no change requested — just flagging that it is on the ask path now.

Size pin 2774 → 2749, in the allowed direction.

PR #1857, opened 2026-09-10
test(conversations): the resume turn, end to end (#1635)

lex00:
The server-driven tests for the resume turn: detach, answer, fresh handshake, the prompt, a colliding request id, a restart in between, the sandbox's own token, the audit row and the webhooks.

Part 6 of 9 for #1635, on #1856.

Part of #1635

### DIFF
--- a/apps/fountain/test/fountain/conversations/conversation_server_acp_test.exs
+++ b/apps/fountain/test/fountain/conversations/conversation_server_acp_test.exs
@@ -1672,6 +1672,35 @@ defmodule Fountain.Conversations.ConversationServerACPTest do
     # cannot see them. A detached answer goes through `send_prompt/4`, which
     # asks the registry first, so the resume turn has to be able to find this
     # server rather than waking a second one.
+    defp register(pid, conv_id) do
+      Mimic.stub(Horde.Registry, :lookup, fn
+        Fountain.ConversationRegistry, ^conv_id -> [{pid, nil}]
+        _registry, _key -> []
+      end)
+    end
+
+    # The detach closes the connection, so a resume turn re-handshakes:
+    # `initialize`, `session/resume` (the caps advertise it), the model pin,
+    # then the prompt. Returns the prompt's params.
+    #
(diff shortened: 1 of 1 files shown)


jhgaylor (approved):
Approving. Tests only, green, and testing the right things.

The one I most wanted to see is here: *"a restart between the detach and the answer changes nothing: the row drives it."* That's the property the whole design rests on — the deadline is on the turn row precisely because no process survives a two-day wait — and a test that kills the server between the halves is the only way to know it holds.

`drive_to_resume_prompt/2` re-handshaking (`initialize`, `session/resume`, `session/set_model`, then the prompt) is what makes the dropped-connection decision in #1855 observable rather than just asserted in a comment, and the note about `models` on the session response not being decoration will save whoever writes the next one an hour.

Also glad to see the negative cases pinned, not just the happy path: `waiting` with nothing held staying an ordinary completed turn, and a request still held when the turn ends **normally** still being denied. The second is the one that would rot quietly if `waiting` ever leaked into the general path.

BinaryBourbon (approved):
Approving. A test-only stage, and the four tests in it are the ones that would actually have caught the bugs this design is avoiding.

- **The colliding request id.** Drives request id `0` in the detached turn, answers, resumes, then raises `0` again and asserts the second request is minted distinctly, carded on the row, announced on the stream *and armed with a timer* that produces a real `reject_once` response down the wire. That is the whole reason the connection is dropped in #1855, verified end to end rather than by asserting the drop.
- **The restart.** `GenServer.stop(pid)` between the detach and the answer, then the answer over the wake path. "Nothing about a detached request lives in a process" is the claim the whole feature rests on, and this is the test that makes it one.
- `drive_to_resume_prompt/2` answering `session/resume` with `"models" => %{}` because a runtime exposing no model selection fails the turn is the sort of harness detail that costs an hour when it is missing. Good that it is commented.
- The webhook test asserting the *order* of the two Oban payload types, rather than that both exist, is the stronger version.

No findings.

PR #1858, opened 2026-09-10
feat(workers): deny a request that outlived its turn and then ran out of time (#1635)

lex00:
The Oban sweep that denies a detached request past its deadline, from the options the agent itself offered, audited to the sweep, delivered as a resume turn. A conversation that cannot take the turn keeps the row for the next minute.

Part 7 of 9 for #1635, on #1857.

Part of #1635

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -4262,6 +4262,48 @@ defmodule Fountain.Conversations do
     end
   end
 
+  @doc """
+  Deny a detached request whose deadline has passed, and tell the agent
+  (#1635).
+
+  The denial picks from the options the agent itself offered, never an id it
+  did not send. Recorded as `conversation.permission_denied` with the sweep as
+  the actor, because no human was at the keyboard and saying otherwise would
+  be a lie about who decided.
+  """
+  @spec _unsafe_expire_detached_request(Turn.t(), keyword()) ::
+          :ok | {:error, term()}
+  def _unsafe_expire_detached_request(%Turn{} = turn, opts \\ []) do
+    request = turn.pending_permission
+    option_id = DetachedRequest.deny_option_id(request)
+    actor = Keyword.get(opts, :actor, "system:detached_request_sweeper")
+
+    # The gate first, for the same reason the answer door applies it first: a
+    # request resolved into a prompt nobody can deliver is gone, the agent is
+    # never told, and there is no second copy to retry from. The deadline has
+    # passed either way, so leaving the row is the safe half of the trade —
+    # the sweep is back in a minute.
+    #
+    # A conversation that is over is the exception: no turn will ever carry
+    # the denial, so the request is resolved and the card stops waiting.
+    case _unsafe_resume_gate(turn.conversation_id) do
+      :ok ->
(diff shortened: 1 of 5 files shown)


jhgaylor (dismissed):
Requesting changes for a one-line format break that this PR introduces itself.

```
apps/fountain/lib/fountain/conversations.ex
  13   13  |  alias Fountain.Audit
       14 +|
  14   15  |  alias Fountain.Conversations.{
```

The diff deletes the blank line after `alias Fountain.Audit` that #1856 left in place. `Elixir static analysis` is red here and on #1859 and #1860 as a result. Put it back and three PRs go green.

The sweeper itself is good and I have no substantive changes to ask for.

- Leaving the row untouched when `_unsafe_resume_gate/1` says `:busy` or the account is refused, and coming back in a minute, is the same trade #1856 makes at the answer door and it's the right one. A deadline that has passed is owed either way; a denial resolved into a prompt nobody delivers is gone.
- `{:error, :gone}` being the exception — resolve, record, don't try to open a turn — is correct. No turn will ever carry that denial, so leaving the card up forever would be the worse failure.
- `actor: "system:detached_request_sweeper"` rather than reusing `sprite`. The vocabulary in 0013 is closed and `system:<worker>` is the right slot; the docstring saying *why* (no human was at the keyboard, and claiming otherwise lies about who decided) is the part that keeps it from drifting back.
- `@sweep_limit 25` with `order_by permission_deadline` so a backlog can't wake a hundred sandboxes in one tick, and the minute grain justified against deadlines measured in hours.

One thing I'd like confirmed rather than changed: `record_permission_denied/4` grew an `opts` with a default, and the `/3` callers are unchanged. Worth a glance that the in-turn path still records `sprite` — the existing tests should cover it, I just didn't trace every caller.

BinaryBourbon (approved):
Approving. The sweep is the right mechanism and the reasons are in the moduledoc where the next person will find them: a process timer cannot span a two-day wait that the sandbox parks through, a deploy in the middle would drop it silently, so the deadline is a column and this is what reads it.

The details that make it safe rather than merely correct:

- **The gate runs before the row is touched**, same as the answer door, with `{:error, :gone}` as the deliberate exception — a terminated conversation will never carry the denial, so resolving it is what stops the sweep finding the same row every minute forever. Both branches are tested, including the "and then it finds nothing" follow-up assertion.
- **`@sweep_limit 25`**, so a backlog cannot wake a hundred sandboxes in one tick.
- **The partial index is the query.** `where waiting == true and permission_deadline <= now order by permission_deadline limit 25` matches `turns_waiting_permission_deadline_index` from #1854 on its leading column, which is what makes a per-minute cron defensible.
- **`actor: "system:detached_request_sweeper"`.** Within the closed vocabulary CLAUDE.md defines, and `record_permission_denied/4` gaining the opt rather than hardcoding `sprite` keeps ADR 0013's "only record what happened" — no human decided this, and saying `sprite` would have been the wrong lie rather than no lie.
- The "a turn that is not waiting is never a candidate, whatever its deadline" test is the guard against this sweep reaching into a request the process timer owns. Worth having explicitly.

One observation, no change requested: the sweep counts `Enum.count(&expire/1)` and logs at `:info` per expiry plus a batch line, so a 25-row backlog writes 26 info lines a minute. Fine at this grain given how rarely the query is non-empty.

No findings.

PR #1859, opened 2026-09-10
feat(api): a conversation lists the requests that outlived its turns, and answers one (#1635)

lex00:
`pending_requests` on GET /api/conversations/{id}, the answer endpoint taking a detached request with its own codes for busy (400) and recorded-but-undelivered (409), `waiting` on a turn, and the regenerated contract and types.

Part 8 of 9 for #1635, on #1858.

Part of #1635

### DIFF
--- a/apps/fountain/lib/fountain_web/controllers/conversation_controller.ex
+++ b/apps/fountain/lib/fountain_web/controllers/conversation_controller.ex
@@ -252,8 +252,18 @@ defmodule FountainWeb.ConversationController do
     # last_active_at, as the plain fetch would, is worse than not serving
     # the fields at all.
     case Conversations.get_conversation_with_activity(id, user.id) do
-      nil -> {:error, :not_found}
-      conv -> render(conn, :show, conversation: conv)
+      nil ->
+        {:error, :not_found}
+
+      conv ->
+        # Ownership: established by the tenant-scoped fetch above. Requests
+        # that outlived a turn (#1635) are served here rather than on the
+        # list, because a conversation is idle while one waits and the card
+        # has to survive a client reload.
+        render(conn, :show,
+          conversation: conv,
+          pending_requests: Conversations._unsafe_list_pending_requests(conv.id)
+        )
     end
   end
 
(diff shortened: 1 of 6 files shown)


jhgaylor (dismissed):
Requesting changes for the label only — the code is fine.

**`A change to the SDK says whether it releases` is red.** Same cause as #1852: this PR changes `sdk/typescript/src/generated/openapi.ts` while the version bump waits for #1860. That's the right sequencing, and the gate's own escape hatch is the answer — add the **`sdk-no-release`** label here and on #1852. Say the word and I'll apply both.

(The `Elixir static analysis` red is #1858's deleted blank line, not yours.)

On the API surface itself, no changes wanted:

- `pending_requests` on `show` only, with the reasoning in the code — the conversation is *idle* while a request waits, so a client that reloads has nowhere else to learn the card is still up, and a list endpoint would pay a query per row for something almost always empty. That's the right place and the right justification.
- `permission_answer_not_delivered` as its own error with its own message. A 409 normally tells a client to give up on a request somebody else took; this one has to say the opposite — the answer landed, the agent hasn't heard it, send a prompt. Reusing the generic 409 would have been actively misleading.
- The catch-all `answer_response({:error, _} = err, _conn)` falling through to the `FallbackController` rather than becoming a `FunctionClauseError` 500, with the reference to why `do_prompt/5` does the same.
- `waiting` added to `turn_data/1` so a client reading turns doesn't have to infer it, and the `operation` docs updated with the new response codes rather than just the happy path.

The `pending_requests` options being the agent's list verbatim, with the "never send an option the agent did not offer" comment on the serializer, keeps that rule visible at the surface a client actually reads.

BinaryBourbon (approved):
Approving. The API surface is the right size for the feature and the seven-gate checklist for a route change is all here — operation declaration with the new statuses, the contract artifact regenerated, `openapi.ts` regenerated, the named schema, and the prose.

Decisions I agree with:

- **`pending_requests` on `show` only.** The reason given is the honest one: the conversation is idle while such a request waits, so a client that reloads has nowhere else to learn the card is still up, and putting it on the list would cost a query per row for something almost always empty. The schema description says so in prose because OpenAPI cannot, which is the right place to be explicit rather than silent.
- **`waiting` on the turn payload**, so a client rendering a transcript can label the turn without cross-referencing the conversation.
- **`:busy` → 400 `conversation_busy` rather than 409.** It is not a "somebody else got there first"; the request is untouched and the same answer will work in a moment. Distinct code, distinct meaning, and the controller test asserts the request survives the 400.
- `PendingPermissionRequest.options` as `additionalProperties: true` objects, verbatim from the agent, with the description saying an id from another runtime is refused. Correct not to model the ACP option shape here.

`ConversationJSON.show/1`'s three clauses order correctly — the `resumed` clause needs both keys and does not shadow the new one — and the empty-list case is tested, so a client can rely on the key's presence on `show` rather than branching on absence.

No findings.

PR #1865, opened 2026-09-11
fix(tests): stop leaking crash-looping ConversationServers into the Horde supervisor (#1862)

### LINKED ISSUE #1862: Flake: a ConversationServer outliving its SQL Sandbox owner can take down the whole :fountain application mid-suite
`Fountain.TeamTest` "update_teammate/4 a co-tenant of the retired computer keeps its own binding, either wake order" (`apps/fountain/test/fountain/team_test.exs:409`) exited instead of asserting.

**The failure**

```
** (exit) exited in: GenServer.call(Fountain.ConversationSupervisor, {:start_child, … conversation_id: "a33ee7c3…" …}, :infinity)
    ** (EXIT) exited in: GenServer.call(Fountain.ConversationSupervisor.ProcessesSupervisor,
        {:start_child, {1392265132474425998…, {… conversation_id: "b01703b7…" …}}}, :infinity)
        ** (EXIT) no process: the process is not alive or there's no process currently
                  associated with the given name

  lib/fountain/conversations.ex:3768: start_conversation_server/4
  lib/fountain/conversations.ex:3855: create_fresh_sandbox_and_start/4
  test/fountain/team_test.exs:447 → :410
```

Note the two different conversation ids. The test asks to start `a33ee7c3`; the call that dies underneath is for `b01703b7`, a conversation from elsewhere in the partition. Just before it, `Fountain.ConversationSupervisor` is terminating on a `:crdt_update` that carries `b01703b7`, and `Horde.SignalShutdown.terminate/2` is calling `:horde_shutting_down`. The same `b01703b7` server then dies with `DBConnection.OwnershipError … using mode :manual`, which is a test process that has already finished.

So the sequence looks like: one test's ConversationServer outlives its owner, Horde replicates it through the CRDT, the supervisor shuts down mid-`add_children`, and an unrelated test's `start_child` is the one that gets the corpse. The wake-order test is the victim, not the cause — it is simply the next caller.

**How often, out of what**

One in four, in one CI round, across trees that contain each other. Four PRs of the #1634 stack ran `Build and Test (partition 6)` on 2026-09-10/11:

| PR | partition 6 | note |
|---|---|---|
| #1832 | **fail** | https://github.com/managoat/fountain/actions/runs/34551683445/job/103115768923 |
| #1834 | pass | contains #1832's commits |
| #1836 | pass | contains #1832's commits |
| #1838 | pass | contains #1832's commits |

#1832 had also passed this same partition on its two previous CI rounds with identical content. It touches five files — `command_runtime.ex`, `runtime_dispatch.ex`, `turn_machine.ex`, `provisioning.ex` and one new test file — and none of them are near `Team`, Horde or `ConversationSupervisor`. The runtime in every frame of the trace is `Managoat.Runtimes.Claude`.

**Where a fix probably belongs**

The test at :409 loops `for order <- [:teammate_first, :cotenant_first]`, starting real servers under the shared Horde supervisor. Whatever leaves `b01703b7` alive after its owning test ends is the actual bug — a `ConversationServer` that survives its SQL Sandbox owner will always eventually take someone else down, and which test it lands on is luck. `DataCase` already waits for tasks a test started; the equivalent for servers started through `start_conversation_server/4` is what seems to be missing.

Filed rather than fixed in place: found while landing an unrelated stack (#1634), and per CONTRIBUTING the campaign's diffs should stay about the campaign.

jhgaylor:
Two tests called `Conversations.wake_conversation/1` without stubbing
`Horde.DynamicSupervisor.start_child/2`, so the wake put a **real**
`ConversationServer` under the supervisor that every test in the partition
shares. In an `async: true` test that server holds no SQL Sandbox connection,
so it raises `DBConnection.OwnershipError` in `handle_continue(:provision)` —
and its child spec is `restart: :transient`, so Horde puts it straight back and
it raises again. A crash loop, for the rest of the run.

Measured over a whole-suite run: **409 of those raises before, 0 after**, and
the suite log drops from 21,561 lines to 4,715.

That churn is what #1862 reported. The visible failure was a *different* test
exiting inside `GenServer.call(ConversationSupervisor, {:start_child, …})` for a
conversation it had never heard of, while Horde was shutting the supervisor
down mid-`add_children`. Which test gets hit is scheduling luck, which is why
it reads as a new flake each time.

- `team_test.exs` already had an `inert_start_child/0` helper and eleven tests
  called it; the co-tenant wake-order test did not. Promoted to
  `setup :inert_start_child` for the file rather than adding a twelfth call:
  nothing in the file asserts on the real start, and forgetting it is invisible
  locally and lands on somebody else's test in CI.
- `conversations_wake_test.exs` — one test, "an idle conversation whose sandbox
  was reclaimed is still resumable", whose wake succeeds and so starts a
  server. Stubbed inline, matching that file's idiom; its assertion
  (`refute match?({:error, :gone}, …)`) is unchanged either way.

Every other test file that wakes or starts a conversation already stubbed
correctly — audited by running each one and counting the raises.

Verified by reverting: removing the `setup` line puts team_test back to 204
raises, and removing the inline stub puts the wake test back to 101. Full
suite: 4807 core, 144 fountain_buzz, 33 fountain_support, 0 failures.

**What this does not do.** It removes the mechanism, not the race. I could not
reproduce the original cross-test failure on demand — it needs a particular
interleaving inside a partition — so this fixes the cause that is demonstrable
and leaves the Horde shutdown path as it is. If it recurs, the next step is a
teardown guard that fails a test which leaves a `ConversationServer` alive for
a conversation it created. That needs the factory to record ids per test and
`DataCase` to check the registry, and because ExUnit runs `on_exit` in a
separate process from the test it needs an ETS table keyed by test pid — real
machinery, not worth inventing on the back of two missing stub calls.

Closes #1862

Closes #1862

### DIFF
--- a/apps/fountain/test/fountain/conversations_wake_test.exs
+++ b/apps/fountain/test/fountain/conversations_wake_test.exs
@@ -40,6 +40,14 @@ defmodule Fountain.ConversationsWakeTest do
       sandbox = insert_sandbox(user_id: user.id, status: "terminated")
       conv = insert_conversation(user_id: user.id, agent: agent, sandbox: sandbox, status: "idle")
 
+      # The wake succeeds here, so it starts a server. Without this it is a
+      # real one under the shared Horde supervisor, with no sandbox connection
+      # in an async test and `restart: :transient` to put it back after every
+      # raise (#1862).
+      stub(Horde.DynamicSupervisor, :start_child, fn _supervisor, _child_spec ->
+        {:ok, spawn(fn -> Process.sleep(:infinity) end)}
+      end)
+
       refute match?({:error, :gone}, Conversations.wake_conversation(conv.id))
     end
 
--- a/apps/fountain/test/fountain/team_test.exs
+++ b/apps/fountain/test/fountain/team_test.exs
@@ -15,10 +15,24 @@ defmodule Fountain.TeamTest do
     )
   end
 
-  defp inert_start_child do
+  # Every test in this file that reaches a wake or a start would otherwise put
+  # a real `ConversationServer` under the shared Horde supervisor (#1862). It
+  # has no sandbox connection in an async test, so it raises on
+  # `handle_continue(:provision)`, and `restart: :transient` puts it straight
+  # back — a crash loop that churns the CRDT for every other test in the
+  # partition. One test forgetting the stub cost 155 of those in a single run.
+  #
+  # Stubbed for the whole file rather than per test: nothing here asserts on
+  # the real start, and the failure mode of forgetting is invisible locally
+  # and lands on somebody else's test in CI.
+  setup :inert_start_child
+
+  defp inert_start_child(_context \\ nil) do
     stub(Horde.DynamicSupervisor, :start_child, fn _sup, _spec ->
       {:ok, spawn(fn -> Process.sleep(:infinity) end)}
     end)
+
+    :ok
   end
 
   describe "list_teammates/1" do


BinaryBourbon (approved):
Two test files, no production change: `team_test.exs` promotes its existing `inert_start_child/0` helper to a file-wide `setup :inert_start_child`, and `conversations_wake_test.exs` adds the same inline stub to the one test whose wake succeeds. Both stop a real `ConversationServer` going under the shared `Fountain.ConversationSupervisor` from an `async: true` test, where it has no SQL Sandbox connection, raises in `handle_continue(:provision)`, and is put straight back by `restart: :transient`. That is the mechanism #1862 describes, and the change removes it. I found no defect — **approving**.

I verified the claims by reverting, on the PR head, in a scratch worktree:

- `team_test.exs` + `conversations_wake_test.exs` together: 65 tests, 0 failures, 0 `DBConnection.OwnershipError`.
- Delete the `setup :inert_start_child` line → 204 raises in `team_test.exs` alone (47 tests still "pass"), exactly the number in the PR body.
- Delete the inline stub → 101 raises in `conversations_wake_test.exs`, also as stated.
- Full core suite on the head: 6 doctests, 4814 tests, 2 skipped, 0 `OwnershipError` from a `ConversationServer`.

Mechanics that could have gone wrong and did not: ExUnit resolves `setup :atom` as a local call, so the private `inert_start_child/1` is reachable; the added `:ok` return is load-bearing, because `Mimic.stub/3` returns the module atom and ExUnit rejects that as a callback result; the `\\ nil` default keeps the eleven existing zero-arity call sites compiling; `Horde.DynamicSupervisor` is already `Mimic.copy`'d in `test_helper.exs`; and Mimic's default private mode keeps the stub off every other process, so the app's own `Horde.DynamicSupervisor.start_child/2` callers (the rehydrator, `FountainBuzz.Manager`) are untouched. Nothing in `team_test.exs` asserts on a real start — `Team.live?/1` reads the conversation status, not the registry, and `open_fresh_conversation/3` creates the row without a server — so the file-wide stub changes no assertion. No sleep-based wait, no new `async: false`, no pool change.

Observations, none blocking:

- `apps/fountain/test/fountain/team_test.exs:28` — the eleven surviving `inert_start_child()` calls are now redundant, and leaving them in reads as if the explicit call were still required, which is the habit the `setup` is there to retire. Dropping them would also let the helper be `defp inert_start_child(_context)` with no default. Taste, not correctness.
- `apps/fountain/test/fountain/conversations_wake_test.exs:45` — the body says this matches the file's idiom, but the file's other stubs return `spawn(fn -> :ok end)` (an already-dead pid). The version here is the better one; no behavioural difference in this test, since `start_conversation_server/4` only casts to the pid and this test passes no prompt.
- Out of scope, for a separate `flake` issue: my full-suite run on this head had one failure, `FountainWeb.RunnersLiveTest` "lists runners with live status and forgets one" (`apps/fountain/test/fountain_web/live/runners_live_test.exs:9`), exiting on `** (DBConnection.OwnershipError) cannot find ownership process for #PID<...> (:proc_lib)` raised in `Fountain.Runners.touch/1` from `Managoat.Runner.Connection.handle_info/2` on a `:heartbeat`, under `Managoat.Runner.FakeDaemon.Socket`. Same family as #1862 — a process outliving its SQL Sandbox owner — but a different mechanism (`on_exit` stops the fake daemon in a separate process, so a heartbeat can land after the test process is gone), untouched by this PR and present on `main`. It passed six times in isolation, so it needs load to show. No `flake` issue exists for it yet.

PR #1866, opened 2026-09-11
chore(deps): ignore the setup-swift v3 prerelease

jhgaylor:
Adds an `ignore` entry to the `github-actions` block of `.github/dependabot.yml` so Dependabot stops proposing the `swift-actions/setup-swift` 3.x line while it is a beta. Supersedes and closes #1760.

## Why #1760 was rejected

Three findings, all verified against the run and against upstream.

### 1. It breaks Linux

`Swift SDK (ubuntu-24.04)` fails on [run 34479405394](https://github.com/managoat/fountain/actions/runs/34479405394/job/102878176781); `Swift SDK (macos-15)` passes, and `CI required` is red as a result.

The v3 beta rewrote toolchain setup on top of Swiftly, which GPG-verifies the download. The runner has no public key for the signing key:

```
[command]/usr/bin/gpg --verify /home/runner/work/_temp/ae57031d-... /home/runner/work/_temp/7a73010f-...
gpg: Signature made Fri Oct 17 15:00:17 2025 UTC
gpg:                using RSA key E813C892820A6FA13755B268F167DF1ACF9CE069
gpg: Can't check signature: No public key
##[error]Unexpected error, unable to continue. Please report at https://github.com/swift-actions/setup-swift/issues
Error: The process '/usr/bin/gpg' failed with exit code 2
```

The beta's own release notes add a `skip-verify-signature` flag to turn this check off, which is not a trade we should take to land a dependency bump.

### 2. The "v3" tag is a prerelease

The SHA #1760 pins, `364295d9c23900ce04d4e5cc708387921b4e50f9`, carries **both** tags upstream:

| ref | resolves to |
|---|---|
| `refs/tags/v3` | `364295d9c23900ce04d4e5cc708387921b4e50f9` |
| `refs/tags/v3.0.0-beta.1` | `364295d9c23900ce04d4e5cc708387921b4e50f9` |
| `refs/tags/v2` | `7ca6abe6b3b0e8b5421b88be48feee39cbf52c6a` (= `v2.4.0`) |

The release is flagged `prerelease: true` on GitHub. `v2` still points at v2.4.0, which is exactly what both workflows pin today, so staying put costs nothing. Dependabot proposed it only because the maintainer moved the floating `v3` tag onto the beta commit, and a bare `v3` reads as plain version 3 rather than as a prerelease.

### 3. It also edits `release.yml`

The bump touches both call sites:

- `.github/workflows/ci.yml:1153`
- `.github/workflows/release.yml:546`

So merging would take the Swift SDK **release** job down on Linux as well, and a failed publish is invisible to every gate this repo has. That failure mode is the one from the Python SDK publish incident, and it is worth avoiding by construction.

## The change

```yaml
  - package-ecosystem: github-actions
    directory: /
    schedule:
      interval: weekly
    groups:
      actions:
        patterns: ["*"]
    ignore:
      - dependency-name: swift-actions/setup-swift
        versions: ["3.x"]
```

Placed in the `github-actions` `updates:` block, since that is the ecosystem that raised #1760 (`ignore` is evaluated before the `actions` group is assembled, so a grouped PR will no longer pick the bump up). 2.x minor and patch updates still flow, and a future 4.x is untouched.

### One caveat, recorded in the config comment

Dependabot reads the floating `v3` tag as plain version `3`, so it cannot distinguish today's beta from a future stable `v3.0.0` — there is no ignore expression that suppresses only the prerelease. The entry therefore covers the 3.x line as a whole, and the comment above it says to delete the entry when a stable v3.0.0 ships and to re-test both the `ubuntu-24.04` and `macos-15` jobs at that point. Reviewing a major bump of the action that sets up the Swift toolchain by hand is the right amount of ceremony either way.

`.github/dependabot.yml` parses (`yaml.safe_load`).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,21 @@ updates:
     groups:
       actions:
         patterns: ["*"]
+    ignore:
+      # swift-actions/setup-swift v3 is a beta: the floating `v3` tag and
+      # `v3.0.0-beta.1` are the same commit (364295d), while `v2` still points
+      # at v2.4.0, which is what ci.yml and release.yml pin. The beta rewrote
+      # toolchain setup on top of Swiftly, which GPG-verifies the download, and
+      # the ubuntu-24.04 runner has no public key for it: `Swift SDK
+      # (ubuntu-24.04)` dies on `gpg: Can't check signature: No public key` ->
+      # `The process '/usr/bin/gpg' failed with exit code 2`. macos-15 passes,
+      # so a bump looks half-green while taking the Swift SDK *release* job
+      # down too, and a failed publish is invisible to every gate here (#1760).
+      # Dependabot reads the floating `v3` tag as plain version 3, so it cannot
+      # tell the beta from a future stable — this ignores the 3.x line whole.
+      # Delete this entry when a stable v3.0.0 ships and re-test both jobs.
+      - dependency-name: swift-actions/setup-swift
+        versions: ["3.x"]
 
   - package-ecosystem: mix
     directory: /
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,9 +22,12 @@ updates:
       # down too, and a failed publish is invisible to every gate here (#1760).
       # Dependabot reads the floating `v3` tag as plain version 3, so it cannot
       # tell the beta from a future stable — this ignores the 3.x line whole.
+      # `github-actions` parses `versions` as a Bundler requirement, so the
+      # range is spelled `~> 3.0`; an npm-style `3.x` parses to `= 3.x`, which
+      # matches no version at all and silently ignores nothing.
       # Delete this entry when a stable v3.0.0 ships and re-test both jobs.
       - dependency-name: swift-actions/setup-swift
-        versions: ["3.x"]
+        versions: ["~> 3.0"]
 
   - package-ecosystem: mix
     directory: /


BinaryBourbon (changes requested):
The reasoning in the PR body holds up: I re-verified the upstream refs and both call sites, and every factual claim checks out. The problem is the one line that has to work. `versions: ["3.x"]` is not a range in the requirement syntax the `github-actions` ecosystem uses, so Dependabot will parse it, accept it, match nothing with it, and reopen the bump on the next weekly run. That is precisely the failure mode this PR exists to prevent, and it is indistinguishable from success until the PR shows up again. Requesting changes for that one line; the rest of the change (placement, dependency name, the comment) is right and should stay as is.

## Finding

**`.github/dependabot.yml:27` — `versions: ["3.x"]` matches no version, so the ignore is a silent no-op.**

`x`-ranges only work for ecosystems whose requirement class understands them (npm). `github-actions` does not. The chain:

- The hosted service hands each `ignore` entry to the updater as a single `version-requirement` string (`updater/lib/dependabot/job/ignore_condition.rb`), and `Job#calculate_update_config` (`updater/lib/dependabot/job.rb:715-721`) rebuilds it into a `Config::IgnoreCondition` verbatim. `Config::File#ignore_conditions` and `IgnoreCondition#ignored_versions` also pass the strings through untouched. There is no x-range normalization at any layer.
- The filter is `GitCommitChecker#tag_included_in_ignore_requirements?` → `requirement_class.requirements_array(req)` → `Dependabot::GithubActions::Requirement`, which is a bare `Gem::Requirement` subclass whose own source comment says "Lifted from the bundler package manager".
- `Gem::Requirement.new("3.x")` does not raise. It parses to `[["=", Gem::Version("3.x")]]`, an equality against a literal letter-segment version. Measured: `satisfied_by?` is `false` for `3`, `3.0.0`, `3.1.0`, `3.0.0-beta.1`, `4.0.0` and `2.4.0`.
- The version Dependabot computed for this bump is `3`, from the floating `v3` tag: #1760's title is "bump swift-actions/setup-swift from 2.4.0 to 3". `= 3.x` does not match `3`, so the `v3` tag survives `tag_included_in_ignore_requirements?` exactly as it did before this PR.

GitHub's own docs are consistent with this: `versions` takes "the standard pattern for the package manager", and the examples given are Bundler `~> 2.0` and npm `^1.0.0`. Bundler syntax is what applies here.

**Failure scenario.** The weekly `github-actions` run reopens the grouped `actions` PR carrying setup-swift `2.4.0 -> 3`, with the config comment above it asserting that it cannot. If someone merges it on the strength of that comment, `Swift SDK (ubuntu-24.04)` (`.github/workflows/ci.yml:1153`) goes red on the GPG check, and the release job (`.github/workflows/release.yml:546`) fails a Swift SDK publish, which is the failure mode no gate here reports. Nothing in this repo lints `dependabot.yml`, so the no-op is invisible locally too.

**Fix.** Use Bundler-style requirement syntax:

```yaml
      - dependency-name: swift-actions/setup-swift
        versions: ["~> 3.0"]
```

`>= 3.0.0, < 4.0.0` works too, but keep the comma **inside one array element** (`versions: [">= 3.0.0, < 4.0.0"]`): `GithubActions::Requirement#initialize` splits a single string on commas and ANDs the parts, whereas two separate array elements become two independent conditions that are OR'd, and `>= 3.0.0` OR `< 4.0.0` ignores every version there is.

Measured for both spellings: matches `3`, `3.0.0`, `3.9.9`; does not match `2.4.0` or `4.0.0`. So the stated intent holds, 2.x minor and patch still flow and a future 4.x is untouched.

## Verified, no defect

- `dependency-name: swift-actions/setup-swift` matches the action name exactly at `.github/workflows/ci.yml:1153` and `.github/workflows/release.yml:546`. `wildcard_match?` is an exact match with no `*`, so the name is right.
- The `# v2.4.0` comment is truthful, not the `# v4`-beside-a-stale-SHA trap: upstream `refs/tags/v2.4.0` is `7ca6abe6b3b0e8b5421b88be48feee39cbf52c6a`, which is the SHA both workflows pin.
- The PR body's upstream claims all check out. `refs/tags/v3` is an annotated tag dereferencing to `364295d9c23900ce04d4e5cc708387921b4e50f9`, the same commit as `v3.0.0-beta.1`, whose release carries `prerelease: true`. `v2` still points at v2.4.0.
- No duplicate or contradicting entry: this is the only `ignore:` block in the file, and no other ecosystem names setup-swift.
- Placement in the `github-actions` block is correct, and the claim about grouping is correct: ignore conditions are applied per dependency in the update checker before the `actions` group is assembled.
- The entry blocks a future stable v3 as well as the beta. That is inherent to the floating tag rather than a defect, the comment says so, and hand-reviewing a major bump of the Swift toolchain action is the right ceremony.

jhgaylor:
Fixed in b09604bb — thanks, this was exactly the failure mode that would have looked like success.

`.github/dependabot.yml:27` is now:

```yaml
      - dependency-name: swift-actions/setup-swift
        versions: ["~> 3.0"]
```

Went with `~> 3.0` rather than the comma form, so the split-vs-OR trap you flagged does not arise at all.

**Measured**, not reasoned — `Gem::Requirement#satisfied_by?` over the value read back out of the file with `YAML.load_file`, since nothing here lints this config:

```
req "~> 3.0" => #<Gem::Requirement [["~>", "3.0"]]>
  3:     true
  3.0.0: true
  3.9.9: true
  2.4.0: false
  4.0.0: false
```

And the old spelling, confirming your parse:

```
req "3.x" => #<Gem::Requirement [["=", Gem::Version "3.x"]]>
  3: false   3.0.0: false   3.9.9: false   2.4.0: false   4.0.0: false
```

`3` is the version that has to match — it is what Dependabot computed from the floating `v3` tag for #1760 ("from 2.4.0 to 3") — and it does. 2.x minor/patch still flow, a future 4.x is untouched.

One correction to your write-up, on a detail that changes nothing here. The literal string `3.0.0-beta.1` does **not** satisfy `~> 3.0`: `Gem::Version` canonicalises it to `3.0.0.pre.beta.1`, and a `String` segment sorts below a missing (`0`) one, so it compares *less than* `3.0`. It fails `>= 3.0.0.pre, < 4.0.0` for the same reason — no `>= 3.0.x` spelling matches it. It does not matter in practice, because a prerelease tag is not a candidate while the pin is a stable version; what Dependabot actually offers is `3` from the floating tag, and that is blocked.

The YAML still parses under both `ruby -ryaml` and `python3 -c "import yaml"`. Placement, `dependency-name` and the existing comment are unchanged. I did add three lines to the comment recording that this ecosystem parses `versions` as a Bundler requirement and that an x-range silently matches nothing, so the next edit does not reintroduce one.

BinaryBourbon (approved):
Re-reviewed at `b09604bb`. The fix is correct, and I reproduced it independently against the file as committed rather than against the diff. Approving.

I also chased the prerelease question, because this PR's title says it ignores a *prerelease* while `~> 3.0` provably does not match `3.0.0-beta.1`. It holds up: the beta is excluded one filter later, by a different mechanism, and the entry blocks the only thing Dependabot would ever offer for the 3 line. Details below, since that reasoning is worth having on the record next to the config comment.

## Measured

Read back with `YAML.load_file` from `.github/dependabot.yml` at `b09604bb`, then fed through `Gem::Requirement` the way `Dependabot::GithubActions::Requirement` does (split each `versions` element on commas, then `satisfied_by?`):

```
versions: ["~> 3.0"]  =>  [["~>", Gem::Version("3.0")]]

  2.3.0                                          matched=false
  2.4.0                                          matched=false
  2.9.9                                          matched=false
  3             (canonical 3)                    matched=true
  3.0.0                                          matched=true
  3.0.1                                          matched=true
  3.9.9                                          matched=true
  3.0.0-beta.1  (canonical 3.0.0.pre.beta.1)     matched=false
  4 / 4.0.0 / 5.0.0                              matched=false
```

That is the fixing agent's result exactly. One `ignore` entry in the file, `dependency-name` unchanged, the other five ecosystem blocks byte-identical, and the YAML parses. The intent stated in the body holds: 2.x minor and patch still flow, a future 4.x is untouched.

## The prerelease question

`~> 3.0` does not match the beta, and no spelling of the range could: `Gem::Version` canonicalises `3.0.0-beta.1` to `3.0.0.pre.beta.1`, and a string segment sorts below a missing one, so it compares less than `3.0`. The PR is still accurate about what it achieves, because the beta never reaches the ignore filter. `GitCommitChecker#allowed_versions` (`common/lib/dependabot/git_commit_checker.rb:418-430`) applies two rejections in order, the ignore requirements first and the prerelease rejection second:

| upstream ref | `version_tag?` | version Dependabot computes | prerelease? | rejected by |
|---|---|---|---|---|
| `7ca6abe6…` (the current pin) | false | n/a | n/a | n/a, it is the pin |
| `v3` | true | `3` | false | **the ignore**, `~> 3.0` matches `3` |
| `v3.0.0-beta.1` | true | `3.0.0.pre.beta.1` | true | the prerelease filter |
| `v2` / `v2.4.0` | true | `2` / `2.4.0` | false | neither, which is intended |

The prerelease filter fires because `wants_prerelease?` is false: it requires `pinned_ref_looks_like_version?`, and the pinned ref is a 40-hex SHA, which `VERSION_REGEX` does not match (no leading `v`, no dot, not a date). Measured all four refs against that regex to confirm.

So the beta commit is unreachable by two independent routes, and the route this PR closes is the one that actually produced #1760: Dependabot titled it "bump swift-actions/setup-swift from 2.4.0 to 3", i.e. version `3` off the floating tag, which `~> 3.0` now blocks. The title and body are not overstating.

## The added comment

Lines 25-27 are accurate, and they are the right three lines to leave behind:

- `github-actions` does parse `versions` as a Bundler requirement. `Dependabot::GithubActions::Requirement` is a `Gem::Requirement` subclass whose own source comment reads "Lifted from the bundler package manager", and the string reaches it untouched through `Config::File#ignore_conditions` and `Job#calculate_update_config`.
- `3.x` does parse to `= 3.x`, an equality against a literal letter-segment version, and it does match nothing. Re-measured: false for every value in the table above, including `3` and `3.0.0`.

Non-blocking, take it or leave it: the comment explains why the entry is spelled as a 3.x range but not why it need not cover `3.0.0-beta.1`. A reader who checks `~> 3.0` against the beta will find it does not match and may think the entry is wrong again. Half a line, something like "the `v3.0.0-beta.1` tag is separately excluded as a prerelease; `v3` is the one that needs ignoring", would close that loop. Not worth another round on its own.

PR #1869, opened 2026-09-11
feat(sandbox): a table for work waiting on capacity (#1033)

jhgaylor:
Re-cut of #1481 as a stack of seven on `main`, under the no-big-PRs rule. The original is 2,171 lines across 44 files and 207 commits behind; this is re-derived against today's code rather than rebased.

This is the table and the tenant-scoped half of the context. `sandbox_requests` holds a start or a scheduled run before it has a sandbox or a conversation — a `queued` conversation status was the alternative, and it would put a row with no machine into every conversation list, every stream and every client enum.

`Fountain.SandboxQueue` gets `enqueue/2`, `list_queued/1`, `get_request/2`, `position/1` and `cancel_request/2`, plus the `SANDBOX_QUEUE_*` config keys. The depth bound is a check followed by an insert, so it takes a per-tenant advisory lock in its own namespace (4317), for the reason `Quotas.with_sandbox_reservation/3` holds one (#330). Cancellation is a compare-and-swap, not a read-then-write, so it can never reach a row the drainer has already claimed.

Audit rows for `sandbox_request.enqueued` and `sandbox_request.cancelled` are in `audit_guardrail_test.exs`, recorded outside the transaction.

**Defers:** the drain (part 2), everything that enqueues (parts 3 and 4), the HTTP endpoints (part 5), telemetry and retention (part 6), and ADR 0042 and the docs (part 7). Nothing enqueues and nothing drains yet.

Part 1 of 7 for #1033.

---

### Review fixes (round 2)

- The schedule dedup moved **inside** the advisory lock, so two firings of one schedule cannot both read "no live request" and both insert. `insert_bounded/1` returns `{:deduplicated, _}` or `{:inserted, _}`, and only an insert audits.
- `@active_statuses` is now `Request.active_statuses()` rather than a second copy of `~w(queued starting)`.
- **New `sandbox_key_id` column** — the restriction a request arrived under, so the replay can be held to ADR 0045 an hour later. Not a `references/2`, deliberately: a nilified id would turn a refusal into a permission the request never had. `schedule_id` gains the same note.
- New public `schedule_request/2`, which part 4 uses to keep a schedule row saying *waiting* instead of repeating a capacity error.

### DIFF
--- a/apps/fountain/lib/fountain/sandbox_queue.ex
+++ b/apps/fountain/lib/fountain/sandbox_queue.ex
@@ -1,0 +1,249 @@
+defmodule Fountain.SandboxQueue do
+  @moduledoc """
+  The bounded per-tenant queue for sandbox capacity (#1033, ADR 0042).
+
+  A start reaches a capacity limit two ways: the tenant's own cap, funded by
+  its credit balance under ADR 0031, and `SANDBOX_FLEET_CEILING` across the
+  whole deployment. `Quotas.with_sandbox_reservation/3` refuses both
+  immediately, which is the right answer for a caller that can retry and the
+  wrong one for a cron firing with nobody there to retry it.
+
+  This module holds that work instead. It never raises a limit: the queue
+  delays admission and every replay re-enters the same reservation, the same
+  credit gate and the same platform-inference gate.
+
+  Bounded twice. A tenant holds at most `SANDBOX_QUEUE_MAX_DEPTH` active
+  requests, and a request waits at most `SANDBOX_QUEUE_MAX_WAIT_SECONDS`.
+  Beyond the depth bound the caller keeps its immediate capacity error.
+  """
+
+  import Ecto.Query, only: [from: 2]
+
+  alias Fountain.Audit
+  alias Fountain.Repo
+  alias Fountain.SandboxQueue.Request
+
+  @default_max_depth 10
+
+  # One source of truth. `Request.active_statuses/0` is what the ops gauge
+  # reads too, and two copies would let the depth bound and the gauge disagree
+  # silently if a status were ever added.
+  @active_statuses Request.active_statuses()
+
+  # Its own advisory-lock namespace. The depth bound counts rows in
+  # `sandbox_requests` and has nothing to serialize against a sandbox
+  # reservation, and every namespace here hashes a different kind of id into
+  # the same 32 bits: sharing one would let a `phash2` collision between a
+  # user and a sandbox block an unrelated writer. Taken: 4315
+  # `Fountain.Quotas`, 4316 `Conversations`, 4331 `Fountain.Connections`.
+  @lock_namespace 4317
+
+  @doc """
+  Queue a start or a scheduled run, subject to the per-tenant depth bound.
+
+  `params` is a map the caller builds key by key — never a request body cast
+  wholesale. Returns `{:error, :queue_full}` at the depth bound, which the
+  caller turns back into the capacity error it was about to send.
+
+  A `schedule_run` is deduplicated against the schedule's own live request, so
+  a cron that keeps firing while the first one waits does not stack ten copies
+  of the same run. That dedup wins over the depth bound: it removes a row
+  rather than adding one.
+  """
+  def enqueue(params, opts \\ []) do
+    with {:ok, outcome} <- insert_bounded(params) do
+      case outcome do
+        # The schedule already had a live request. Nothing was written, so
+        # nothing is recorded: a trail that logs an attempt as a change is
+        # worse than no trail (ADR 0013).
+        {:deduplicated, request} ->
+          {:ok, request}
+
+        # Audited outside the transaction: `Audit.record/1` is best-effort
+        # by rescuing, and a rescue does not survive a transaction — a
+        # failed audit insert would abort the enclosing one and take the
+        # request with it.
+        {:inserted, request} ->
+          audited(request, "sandbox_request.enqueued", opts)
+          emit_depth(request.user_id)
+          {:ok, request}
+      end
+    end
+  end
+
+  @doc """
+  The tenant's live request for a schedule, if it has one.
+
+  What lets a surface say "waiting for a free slot" rather than repeating the
+  capacity error a replay just met again.
+  """
+  def schedule_request(user_id, schedule_id)
+      when is_binary(user_id) and is_binary(schedule_id) do
+    existing_schedule_request(%{user_id: user_id, schedule_id: schedule_id})
+  end
+
+  # The depth bound is a check followed by an insert, so it needs the same
+  # protection `Quotas.with_sandbox_reservation/3` gives the sandbox cap: two
+  # requests that both read "room for one more" at the last slot is precisely
+  # how #330 got past that cap before its advisory lock existed.
+  #
+  # The schedule dedup reads under the same lock, for the same reason: two
+  # firings of one schedule that both read "no live request" would both insert,
+  # which is the stacking the dedup exists to prevent.
+  defp insert_bounded(params) do
+    Repo.transaction(fn ->
+      Repo.query!("SELECT pg_advisory_xact_lock($1, $2)", [
+        @lock_namespace,
+        :erlang.phash2(params.user_id)
+      ])
+
+      case existing_schedule_request(params) do
+        %Request{} = request ->
+          {:deduplicated, request}
+
+        nil ->
+          with :ok <- check_depth(params.user_id),
+               {:ok, request} <-
+                 %Request{}
+                 |> Request.changeset(Map.put_new(params, :status, "queued"))
+                 |> Repo.insert() do
+            {:inserted, request}
+          else
+            {:error, reason} -> Repo.rollback(reason)
+          end
+      end
+    end)
+  end
+
+  @doc "List a tenant's waiting requests, oldest first."
+  def list_queued(user_id) when is_binary(user_id), do: Repo.all(queued_query(user_id))
+
+  @doc "Get a request scoped to its tenant."
+  def get_request(id, user_id) when is_binary(id) and is_binary(user_id) do
+    case Ecto.UUID.cast(id) do
+      {:ok, request_id} -> Repo.get_by(Request, id: request_id, user_id: user_id)
+      :error -> nil
+    end
+  end
+
+  @doc """
+  A waiting request's one-based position.
+
+  Counts claimed (`starting`) rows as well as waiting ones: a request being
+  replayed right now is still ahead of this one, and leaving it out reports
+  `position: 1` to a caller that has someone in front of it. `nil` once the
+  request is no longer waiting.
+  """
+  def position(%Request{status: "queued"} = request) do
+    Repo.aggregate(
+      from(r in Request,
+        where:
+          r.user_id == ^request.user_id and r.status in ^@active_statuses and
+            (r.inserted_at < ^request.inserted_at or
+               (r.inserted_at == ^request.inserted_at and r.id < ^request.id))
+      ),
+      :count
+    ) + 1
+  end
+
+  def position(%Request{}), do: nil
+
+  @doc """
+  Cancel a request if it is still waiting.
+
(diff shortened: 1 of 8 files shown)


BinaryBourbon:
Pushed a fix across this stack: the two `SANDBOX_QUEUE_*` rows in `docs/configuration.md` moved from #1876 (the docs PR) into #1869, where `config/runtime.exs` first reads them.

`config_reference_test.exs` asserts that every env var read in `config/runtime.exs` is documented in `docs/configuration.md`. Reads landed in PR 1 and the docs landed in PR 7, so the guard failed on **all six** PRs in between:

```
config/runtime.exs reads env vars that docs/configuration.md does not document:
  SANDBOX_QUEUE_MAX_DEPTH, SANDBOX_QUEUE_MAX_WAIT_SECONDS
```

Only #1870 and #1871 had surfaced it — the rest were still queueing and would have gone red the same way.

The general rule for this stack and the others in flight: **a guard test that spans code and docs needs both in the same PR.** Splitting a change by layer is usually right, but not across a check that reads one and asserts on the other.

#1869 now carries both rows; PRs 2–7 were rebased and force-pushed with `--force-with-lease`, and #1876's diff no longer duplicates them. `config_reference_test.exs` passes on #1869 (11 tests, 0 failures). Nothing else changed.

BinaryBourbon:
Reviewed the whole `stack:1033` set (#1869 → #1870 → #1871 → #1872 → #1873 → #1874 → #1876) against `main` at `4ca555da`, as one change. This comment is the stack-level summary; the per-part notes are on each PR.

**Verdict: approving parts 1, 3, 5, 6 and 7; requesting changes on #1870 and #1872.** Two defects, both small fixes, both in the replay path rather than in the table or the drain machinery.

## What I checked

- The claim fence, end to end: `claim_next/2`'s swap on `(status, updated_at)`, the same fence on `release/2` and `finish/3`, and the two regression tests. The design argument in #1870's description is correct, and the tests do exercise it — stubbing `Billing.check_spend/1` lands inside `start_conversation/2`'s `with` at `conversations.ex:2501`, before the reservation, so the row really is moved out from under the replay and the assertions are not vacuous.
- That every replay re-enters the real gates. `attempt/1` goes through `start_or_resume_conversation/2`, and `Quotas.with_sandbox_reservation/3` holds `check_fleet_ceiling` → `Credits.gate` → `check_sandbox_quota` under both advisory locks, so `docs/api.md`'s "must pass the credit gate and the inference gate again" is true and the queue genuinely cannot raise a cap.
- The advisory-lock namespace. 4317 is free (4315 Quotas, 4316 Conversations, 4331 Connections) and nothing in `sandbox_requests` needs serialising against a sandbox reservation, so a separate namespace is right.
- The poke choke point. `maybe_poke_sandbox_queue/2` is after the commit, inside `update_sandbox/2`, and only fires on a transition *out of* `Quotas.active_statuses/0` — so `suspended` frees a slot (ADR 0017) and `pending -> ready` frees nothing, which the drainer test pins. The rescue matches the metering precedent at the same choke point.
- The Oban uniqueness argument. `states: :scheduled` really is the only group that does not swallow a poke arriving behind a running drain. Worth knowing that it makes the effective dedup window ~1s rather than the nominal 30, since a job leaves `scheduled` as soon as its `schedule_in: 1` elapses — deliberate, and the right side of the trade, but it is why the churn note on #1872 matters.
- Seven gates for the three new routes: operation declarations, a `forbidden` response on each, enums derived from `Request.kinds/0` / `statuses/0` with guard rows, the `omissions.json` claim, regenerated `contract.json` and `openapi.ts`. All present. `system:sandbox_queue` satisfies `audit_actor_vocabulary_test.exs`'s `^system:[a-z][a-z0-9_]*$`.
- The SDK bump. `main` is at 1.27.0, npm's latest is 1.27.0, and all four edits (`package.json`, both `package-lock.json` occurrences, `USER_AGENT`, `CHANGELOG.md`) are consistent at 1.28.0.

## The two blockers

1. **A queued start loses its `source`** (#1870). `attempt/1` restores `user_id` and `agent_id` onto `request.attrs` but never `request.source`, and `start_conversation/2` defaults `attrs["source"] || "api"`. So an agent fan-out that queues is recorded as `source: "api"`. One line, and the `source` column is already there holding the right value.
2. **The replay drops `sandbox_key_id`, bypassing the ADR 0045 label rule** (#1872). Details on that PR — a `sprite`-scoped token can reach `POST /api/conversations` (that is what `source: "agent"` *is*), and the replay passes only `actor:`, so `sandbox_owns?(_conv, nil) -> true`.

## Worth fixing, not blocking

- The schedule row's `last_error` reverts from `"waiting for a free sandbox slot"` to `"sandbox quota: 5/5"` on the first replay attempt, and each attempt re-stamps `last_run_at`, writes a `team.schedule.fired` audit row and broadcasts. ADR 0042 decision 3 and issue #1033 point 7 both promise the opposite. On #1872.
- `sandbox_request.started`, `.failed` and `.expired` are emitted but nothing asserts them; `finish/3`'s `case updated.status` has no fallback clause. ADR 0042 decision 8 claims all five transitions are covered. On #1870.
- Five smaller notes spread across #1869, #1873, #1874 and #1876.

## Housekeeping

- The `docs/configuration.md` rows for `SANDBOX_QUEUE_MAX_DEPTH` / `SANDBOX_QUEUE_MAX_WAIT_SECONDS` were added in part 1 and reverted in part 2, which failed `Fountain.ConfigReferenceTest` on #1870 (partition 2) and #1871 (partition 3) in run `34560380231`. Fixed by the force-push to parts 2–6 during this review; the rows are now present on every branch. Flagging it only so the two red runs are not mistaken for flakes.
- Three open PRs claim SDK 1.28.0: this one, #1860 (#1635) and #1885 (#1125). Ascending order at merge time, so whichever lands second re-rolls the four edits.

Nothing else from #1481 looks dropped. The re-cut is a clear improvement on it — the claim fence and the `labels` / `execution_limits` / `sandbox_api_access` additions to the allow list are both real bug fixes relative to that branch.

BinaryBourbon (approved):
Part 1 of 7. The table, the schema and the tenant-scoped half of the context. Approving — this part is sound on its own and nothing enqueues or drains yet.

The separate-table choice is right and the reasoning in the moduledoc is the reasoning I would want: a `queued` conversation status would have to be taught to every conversation query, the SSE stream and four client enums, and the row has no machine to point at. The two indexes match the three reads exactly (the FIFO/depth/position composite, and the partial one for the fan-out probe), and cancellation being a compare-and-swap rather than a read-then-write is the detail that makes `DELETE` safe against the drainer in part 5.

`position/1` counting `starting` as well as `queued` is correct and easy to get wrong the other way — a claimed row is still ahead of you — and the tie-break on `id` matches `queued_query/1`'s `order_by`, so position and list order cannot disagree.

Three notes, none blocking:

**The schedule dedup is outside the lock.** `enqueue/2` calls `existing_schedule_request/1` *before* `insert_bounded/1` takes `pg_advisory_xact_lock(4317, phash2(user_id))`, and there is no partial unique index on `(user_id, schedule_id)` for active rows. Two firings of the same schedule landing together can both read `nil` and both insert — the same shape as the depth-bound race the lock exists for (#330). In practice `Workers.TeamScheduleRun`'s `unique: [keys: [:schedule_id, :fired_at]]` makes it hard to hit, so it is not a blocker, but moving the dedup read inside the transaction after the lock is free and removes the argument entirely.

**`@active_statuses` is declared twice.** `Request.active_statuses/0` exists and part 6 uses it for the ops gauge, while `SandboxQueue` keeps its own `~w(queued starting)` copy. Two places to edit if a status is ever added, and the drift would be silent — the depth bound and the gauge would simply disagree. Point the module attribute at `Request.active_statuses/0`.

**`schedule_id` has no foreign key.** Deliberate as far as I can tell, and part 2 handles the dangling case properly (`:schedule_deleted` is terminal, with a test). Worth a line in the migration saying so, because the column sits next to two `references(...)` and reads like an omission.

jhgaylor:
All three notes addressed, plus the column the ADR 0045 fix needs. Force-pushed.

**The schedule dedup now reads under the lock.** It moved inside `insert_bounded/1`, after `pg_advisory_xact_lock(4317, phash2(user_id))`, so the argument about two firings both reading "no live request" is gone rather than merely unlikely. `insert_bounded/1` returns `{:deduplicated, request}` or `{:inserted, request}` so the audit and the depth telemetry still fire only on a real insert — a deduplicated firing wrote nothing, so it records nothing. Pinned by *"a deduplicated firing records nothing, because nothing was written"*.

**One source of truth for the live statuses.** `@active_statuses Request.active_statuses()` — resolved at compile time, so the queries keep their `^@active_statuses` pin and the ops gauge in part 6 reads the same list.

**`schedule_id` says why it has no foreign key**, and so does the new `sandbox_key_id` beside it. Both want a dangling id rather than a nilified one: a dangling `schedule_id` replays as `:schedule_deleted` (terminal), and a dangling `sandbox_key_id` fails the ownership comparison in `set_conversation_labels/4`. `nilify_all` on either would turn a refusal into a permission the request never had, which is the wrong direction to fail in.

**New: a `sandbox_key_id` column** — the fix for the ADR 0045 bypass found on #1872. It is a restriction rather than a credential, and the schema docstring says so. Two tests here (it round-trips; `schedule_request/2` is tenant-scoped and stops at a terminal row), with the behaviour it protects tested on #1870 and #1872.

`schedule_request/2` is also new and public: part 4 needs "does this schedule have live work" to keep the row saying *waiting* rather than repeating a capacity error.

BinaryBourbon (approved):
Re-reviewed the delta since my approval (`3fc08ee2` → `00b2e064`). Approving again, and the change is larger than "add the column I asked for on #1872".

## The column

`sandbox_key_id` lands here rather than in a follow-up migration, which is the right call while part 1 is unmerged. The migration comment is the part I want to keep: **neither** id is a `references/2`, and the reason given — a nilified `sandbox_key_id` would read as "no sandbox restriction applies" and turn a refusal into a permission the original request never had — is the actual security property, not a style note. A dangling `schedule_id` replaying as `:schedule_deleted` is the same shape. Good instinct to write down why a foreign key is *wrong* here; that is exactly the decision somebody will otherwise "fix" in six months.

`sandbox_key_id` is absent from `SandboxQueueJSON.data/2`, which is correct — it is a restriction the server holds the caller to, not something the caller needs read back.

## The dedup moved under the lock

This I did not ask for and it is the best change in the delta. Previously `existing_schedule_request/1` ran outside `insert_bounded/1`, so two tickers firing one schedule could both read "no live request" and both insert — the exact stacking the dedup exists to prevent, and the same shape as #330 getting past the sandbox cap before its advisory lock. It now reads inside the transaction that already holds `pg_advisory_xact_lock(4317, phash2(user_id))`, and the comment says so.

The `{:inserted, _}` / `{:deduplicated, _}` split is what makes the audit behaviour fall out rather than be remembered: a dedup writes nothing, so it records nothing (ADR 0013), and `"a deduplicated firing records nothing, because nothing was written"` pins it at three firings → one row → one event.

## `@active_statuses Request.active_statuses()`

Right direction. It costs a compile-time dependency on `Request` (touching the schema recompiles this module), which is cheap and worth it — the gauge in part 6 reads the same function, and two hand-copied `~w(queued starting)` lists that drift is a silent bug in the depth bound rather than a loud one.

## Verified locally

Migrated a private database at the stack tip and ran the affected files (`sandbox_queue_test.exs`, `sandbox_queue_drain_test.exs`, `team/schedules_test.exs`, `sandbox_queue_controller_test.exs`, `audit_guardrail_test.exs`, `conversation_controller_test.exs`, `marketing_pricing_test.exs`, `ops_gauges_test.exs`, `retention_pruner_test.exs`, `docs_test.exs`): 298 tests, 0 failures. `mix credo --strict` clean, `mix format --check-formatted` clean, `scripts/sdk-contract/build.sh --check` ok. CI itself has reported nothing on any of these heads yet — every job on the 05:26 runs is `cancelled` by concurrency, which `gh pr checks` renders as `fail`. Worth a glance before merging rather than trusting that column.

Nothing blocking here.

PR #1870, opened 2026-09-11
feat(sandbox): claim before replay, fenced on the version claimed (#1033)

jhgaylor:
Stacked on #1869.

The claim/drain state machine. `drain/1` expires overdue work, then replays one tenant's queue in FIFO order, and a replay ends three ways:

- a **capacity** refusal (`sandbox_quota_exceeded`, `fleet_full`) returns the claim and stops the pass, because every request behind it would meet the same wall;
- a **transient** refusal (`:busy`, `:provisioning`, `:runner_offline`, `:sandbox_at_capacity`) returns the claim and the drain moves past it, for the reason `Workers.TeamScheduleRun` snoozes on that same list — the turn in flight ends, the home finishes building, the runner comes back, and treating those as terminal would destroy the prompt over a condition that clears by itself;
- anything else is terminal.

**The one place this departs from #1481's design.** Every write out of a live status is a compare-and-swap fenced on the `(status, updated_at)` the drain observed, and the claim is the same swap: `claim_next/2` takes a `queued` row, *or* a `starting` row whose claim outran five minutes because the worker holding it died. #1481 recovered stale claims in a separate `update_all` pass at the top of each drain and then wrote outcomes with a blind `Repo.update/1`. That leaves a window: a replay slow enough to lose its claim writes `"started"` and its own conversation id over the row a recovering drain has already replayed, so one request reads as one conversation while two are running. Folding recovery into the claim and fencing every later write closes it. `sandbox_queue_drain_test.exs` has the two regression tests ("a claim recovered mid-replay does not have its outcome overwritten", "a claim lost before release leaves the recovering drain's row alone").

Also here: `any_active_requests?/0` and `user_ids_with_active_requests/0`, the two reads the fan-out in part 3 needs.

**Defers:** nothing calls `drain/1` yet — the worker and the poke are part 3. Producers are part 4, endpoints part 5, telemetry declarations part 6, ADR and docs part 7.

Part 2 of 7 for #1033.

---

### Review fixes (round 2)

- **`attempt/1` restores `source`.** It was dropped, so every queued fan-out replayed as `source: "api"` instead of `"agent"`.
- **`attempt/1` carries `sandbox_key_id`** through `replay_opts/1`. Without it a `sprite`-scoped token could queue a `channel_id` start with `labels` and have the drainer merge them into a conversation the token does not own — the write ADR 0045 has `set_conversation_labels/4` refuse at the door.
- `finish/3` names its event from the status instead of a two-arm `case` that would be a `CaseClauseError` inside a drain pass.
- `claim_next/2` adds a lost row to `skipped`, so the loop terminates structurally.
- Audit assertions for `sandbox_request.started`, `.failed` and `.expired`, plus one that a transient release records nothing.

Both replay fixes were verified by reverting them: the ADR 0045 test fails with `started: 1` (the relabel going through) and the provenance test with `left: "api"`.

---

### Review fixes (round 3)

`SandboxQueue.actor/0` exposes `"system:sandbox_queue"` the way `Team.Schedules.actor/0` exposes its own, so a surface can tell the drainer's replay from a person's own call without hardcoding the string. Used by part 4 to decide which callers are silent about a wait.

Part of #1033

### DIFF
--- a/apps/fountain/lib/fountain/sandbox_queue.ex
+++ b/apps/fountain/lib/fountain/sandbox_queue.ex
@@ -15,20 +15,60 @@ defmodule Fountain.SandboxQueue do
   Bounded twice. A tenant holds at most `SANDBOX_QUEUE_MAX_DEPTH` active
   requests, and a request waits at most `SANDBOX_QUEUE_MAX_WAIT_SECONDS`.
   Beyond the depth bound the caller keeps its immediate capacity error.
+
+  ## Claiming
+
+  Several replicas can be told a slot freed at the same instant, so a drain
+  claims before it replays. Every write that moves a row out of a live status
+  is a compare-and-swap fenced on the `(status, updated_at)` the drain
+  observed, and a claim is the same compare-and-swap: `claim_next/2` will take
+  a `queued` row, or a `starting` row whose claim outran
+  `@claim_timeout_seconds` because the worker holding it died. Only the
+  replica that won the swap can write that row's outcome, so a zombie replay
+  cannot overwrite the row a recovering drain has since taken.
   """
 
   import Ecto.Query, only: [from: 2]
+
+  require Logger
 
   alias Fountain.Audit
   alias Fountain.Repo
   alias Fountain.SandboxQueue.Request
 
   @default_max_depth 10
+  @default_max_wait_seconds 3600
 
   # One source of truth. `Request.active_statuses/0` is what the ops gauge
   # reads too, and two copies would let the depth bound and the gauge disagree
   # silently if a status were ever added.
   @active_statuses Request.active_statuses()
+
+  # A worker can die after its compare-and-swap and before its replay
+  # finishes. Both replay paths normally return in milliseconds — provisioning
+  # happens in the ConversationServer, not under the claim — so five minutes
+  # identifies an abandoned claim with a wide margin.
+  @claim_timeout_seconds 300
+
+  # Not this request's fault and not this tenant's ceiling: the turn in flight
+  # ends, the home sandbox finishes provisioning, the runner comes back. The
+  # request goes back in line rather than burning its prompt on a condition
+  # that clears by itself. `Fountain.Workers.TeamScheduleRun` snoozes on
+  # exactly this list, for exactly this reason.
+  @transient_errors ~w(busy provisioning runner_offline sandbox_at_capacity)a
+
+  # Every replay and every terminal write the drain makes is attributed to the
+  # queue, not to whoever originally asked. The audit vocabulary is closed
+  # (ADR 0013) and `system:<worker>` is the shape a background caller takes.
+  @system_actor "system:sandbox_queue"
+
+  @doc """
+  The audit actor a replay runs as.
+
+  Exposed the way `Team.Schedules.actor/0` is, so a surface can tell the
+  drainer's replay from a person's own call without hardcoding the string.
+  """
+  def actor, do: @system_actor
 
   # Its own advisory-lock namespace. The depth bound counts rows in
   # `sandbox_requests` and has nothing to serialize against a sandbox
@@ -156,7 +196,7 @@ defmodule Fountain.SandboxQueue do
   reached a `starting` row would abandon a start already in flight.
   """
   def cancel_request(%Request{} = request, opts \\ []) do
-    case terminate(request.id, "queued", %{status: "cancelled", attrs: %{}}) do
+    case swap(request.id, "queued", %{status: "cancelled", attrs: %{}}) do
       {:ok, cancelled} ->
         audited(cancelled, "sandbox_request.cancelled", opts)
         emit_depth(cancelled.user_id)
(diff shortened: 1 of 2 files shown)


BinaryBourbon (changes requested):
Part 2 of 7. The claim/drain state machine. Requesting changes for one small defect and one test gap; the fence itself is good work and I would take it as-is otherwise.

## The fence

I went looking for a hole in this and did not find one. `claim_next/2` swapping on the exact `(status, updated_at)` it read, with stale-claim recovery folded into the same swap rather than run as a separate `update_all` pass, closes the window the description names — and the two regression tests are real tests, not shape tests. I checked that the `Billing.check_spend/1` stub actually fires: it is called at `conversations.ex:2501`, inside `start_conversation/2`'s `with` and before the reservation, so the row genuinely moves out from under the replay and `started: 1` with `status: "cancelled"` is the assertion it looks like. `release/2` tolerating a lost swap instead of raising is the right call too — raising there would fail the job and strand every other request the tenant has waiting.

One precision note on the framing, no code change: the fence prevents the *bookkeeping* double-start, not the double *start*. A zombie replay that outran its five minutes still creates a conversation, and after this change that conversation is simply orphaned from the request row (the `:stale` branch logs and drops it). That is strictly better than #1481 writing `started` and its own conversation id over a row a recovering drain had already replayed, and the orphan is visible in the tenant's own conversation list rather than invisible spend — but "latent double-start" in the description slightly oversells what the fence fixes. The five-minute margin is what keeps it rare, and the justification for that margin (provisioning happens in the `ConversationServer`, not under the claim) is correct.

The three-way classification is right, and `:sandbox_at_capacity` belongs in the transient set for exactly the reason given — it is one shared machine's turn ceiling (ADR 0023), not the tenant's. `skipped` keeping the loop from re-claiming what it just released is the detail that makes the transient path terminate.

## Blocking: a queued start loses its `source`

`attempt/1` rebuilds the launch attrs as:

```elixir
attrs =
  request.attrs
  |> Map.put("user_id", request.user_id)
  |> Map.put("agent_id", request.agent_id)
```

`source` is not restored, and `start_conversation/2` reads `attrs["source"] || "api"` (`conversations.ex:2564`). Part 4 deliberately keeps `source` out of `@queued_attr_keys` and stores it in the request's own `source` column instead — so the right value is sitting on the row, and the replay just never reads it back.

The case that matters is a sandbox fan-out: `infer_provenance/1` sets `source: "agent"` from the parent-conversation header, a `sprite`-scoped token can reach `POST /api/conversations`, and that is precisely the caller most likely to meet the tenant cap. Queue it and the conversation records `source: "api"` instead. That is a provenance field `GET /api/conversations` filters on and the analytics sink reads.

`Map.put("source", request.source)` when it is non-nil, plus a row in the drain test asserting the replayed conversation keeps `"agent"`.

Note the asymmetry that makes this read as an oversight rather than a decision: `parent_conversation_id` is *also* set by that path from a header, and it **is** in the allow list. The intent was clearly to preserve provenance.

## Blocking: three of the five audit transitions are untested

`finish/3` records `sandbox_request.started` and `sandbox_request.failed`, and `expire_overdue/1` records `sandbox_request.expired`. Nothing in the stack asserts any of the three — `audit_guardrail_test.exs` covers `enqueued` and `cancelled` only, and a grep across the stack's tests finds those two strings and no others. ADR 0042 decision 8 says "Audit events cover enqueue, start, cancellation, expiry and failure", so today the ADR is ahead of the tests on three of five.

They are cheap to pin — the drain tests already produce all three outcomes (`started`, the `insufficient_credits` failure, the two-hour-old expiry), so it is an extra assertion on existing setups rather than new tests.

While you are in there: `finish/3`'s

```elixir
action =
  case updated.status do
    "started" -> "sandbox_request.started"
    "failed" -> "sandbox_request.failed"
  end
```

has no fallback. It is exhaustive over today's two callers, but it is a `CaseClauseError` raised inside a drain pass the moment a third terminal outcome is written through `finish/3`, and it would take the rest of that tenant's queue down with it. A `status -> "sandbox_request.#{status}"` fallback, or an explicit raise with a message, costs nothing.

## Non-blocking

**`claim_next/2` recurses without a bound on `:stale`.** Each lost swap calls itself with the same `skipped` list. In practice it terminates — a row another replica claimed is now `starting` with a fresh `updated_at` and no longer matches the query — but a row being claimed and released in a tight loop would be re-read each time. A depth counter, or adding the lost id to `skipped`, makes termination structural rather than incidental.

**`expire_overdue/1` only expires `queued` rows.** Right choice (a `starting` row is claimed work, and recovery is the claim's job), but it means nothing bounds a `starting` row's lifetime — see the note on #1874 about the `RetentionPruner` docstring's "they cannot outlive `SANDBOX_QUEUE_MAX_WAIT_SECONDS` anyway", which is only true while some drain keeps running for that tenant.

jhgaylor:
Both findings fixed, and the two non-blocking notes with them. Force-pushed.

**`source` was lost on replay.** `attempt/1` now restores it, guarded so a request that never had one does not write `nil` over `start_conversation/2`'s `|| "api"` default:

```elixir
|> put_unless_nil("source", request.source)
```

Test: *"keeps the provenance the API inferred instead of defaulting to api"*. Verified by reverting the fix — it fails with `left: "api", right: "agent"`, which is the fan-out being relabelled as a plain API start.

**The replay also dropped `sandbox_key_id`** (raised on #1872, fixed across parts 1, 2 and 4 because the column, the producer and the replay are in different stages). `attempt/1` now builds its opts from the request rather than hardcoding the actor:

```elixir
defp replay_opts(%Request{sandbox_key_id: nil}), do: [actor: @system_actor]

defp replay_opts(%Request{sandbox_key_id: key_id}),
  do: [actor: @system_actor, sandbox_key_id: key_id]
```

Three tests under *"what the door passed survives the replay"*, including the one that matters most: reverting the fix makes *"holds a sprite token to the conversation it owns (ADR 0045)"* fail with `left: %{failed: 1}, right: %{started: 1, ...}` — the replay successfully relabelling a conversation the token does not own. The third test pins the other direction, that a request with no restriction still resumes and labels, so the fix cannot quietly become a blanket refusal.

**The three untested audit transitions are pinned.** `sandbox_request.started`, `.failed` and `.expired` each have an assertion on the action, the `system:sandbox_queue` actor and the resource, on the setups that already produced those outcomes. Added one more while I was there — *"a release records nothing, because the request is still waiting"* — since a transient release is an attempt rather than a change, and that is the direction ADR 0013 cares about.

**`finish/3` no longer has a two-arm `case`.** The status names its own event:

```elixir
audited(updated, "sandbox_request.#{updated.status}", actor: @system_actor)
```

**`claim_next/2` terminates structurally.** A lost swap now adds the row to `skipped`, so each turn round either claims a row or removes a candidate rather than relying on the row no longer matching the query.

On the framing note: you are right that the fence settles the bookkeeping rather than the work, and ADR 0042 decision 6 now says so explicitly (part 7) — the zombie's conversation is orphaned from the request and stays visible in the tenant's list; what the fence removes is one request reading as one conversation while two run.

BinaryBourbon (approved):
Re-reviewed the delta since my changes-requested (`ccfe1dfd` → `3385706b`). **Both blockers are cleared and both non-blocking notes were taken.** Approving.

## Blocker 1: the queued start lost its `source` — fixed

`attempt/1` now ends `|> put_unless_nil("source", request.source)`, and the test is the case I named rather than a shape test: a request stored with `source: "agent"` replays into a conversation whose `source` is `"agent"`, not the `attrs["source"] || "api"` default. `put_unless_nil/3` rather than a blind `Map.put` is right — a `nil` written into `attrs` would defeat the `||` fallback and store a `nil` source.

## Blocker 2: the three untested audit transitions — fixed, and the `case` with it

Four new assertions, all on setups that already existed:

- `sandbox_request.started` — actor, resource type, resource id, and `metadata["conversation_id"]` against the row's own `conversation_id`.
- `sandbox_request.failed` — `metadata["error"] == "insufficient_credits"`, reached by draining the opening credit.
- `sandbox_request.expired` — on the two-hour-old row.
- `"a release records nothing, because the request is still waiting"` — the negative case, which is the one I would have forgotten. A transient release is an attempt, not a change, and asserting `queue_actions(user) == ["sandbox_request.enqueued"]` pins that as an equality rather than a "does not contain".

Both `refute Map.has_key?(event.metadata, "prompt")` assertions are worth keeping even though nothing today writes one.

And `audited(updated, "sandbox_request.#{updated.status}", ...)` replaces the two-arm `case`. The status names its own event, so a third terminal outcome cannot raise a `CaseClauseError` inside a drain pass and strand the rest of the tenant's queue. There is no audit-action allowlist to keep in step, so the interpolation costs nothing.

## Non-blocking, both taken

`claim_next/2` on `:stale` now recurses with `[request.id | skipped]`, so each turn round either claims a row or removes a candidate. Termination is structural instead of resting on the row no longer matching the query. The comment says exactly that.

The `RetentionPruner` docstring note went to #1874 and reads correctly there now.

## The ADR 0045 fix, which I raised on #1872

The carry lands here: `replay_opts/1` passes `sandbox_key_id` alongside the actor when the request has one, `[actor: @system_actor]` when it does not. I traced it end to end — `start_or_resume_conversation/2` → `resume_labels/3` → `set_conversation_labels/4` → `sandbox_owns?/2` — and the two tests are the two halves that matter:

- a sprite-scoped request whose `channel_id` resolves to another sandbox's conversation fails with `error == "sprite_may_not_label_another_conversation"`, and `theirs.labels` is still `%{}`;
- a request with **no** restriction still resumes and labels, so the fix cannot quietly become a blanket refusal.

That second test is the one that makes this a fix rather than a patch.

## One thing that is not yours, flagged here because the seam is

`expire_overdue/1` marks a `schedule_run` request `expired` and tells nobody. The schedule row keeps saying `waiting for a free sandbox slot` with `last_run_at` still `nil`, indefinitely for a one-off. That is part 4's behaviour, so the blocker is on #1872 — but the seam it needs is here, in `finish/3` and `expire_overdue/1`. Fine with me wherever it lands.

## Verified locally

`sandbox_queue_test.exs`, `sandbox_queue_drain_test.exs`, `team/schedules_test.exs`, `sandbox_queue_controller_test.exs`, `audit_guardrail_test.exs`: 151 tests, 0 failures, on a database migrated to this tree. `mix credo --strict` clean. No CI verdict exists on this head yet — the 05:26 jobs are all `cancelled` by concurrency, which `gh pr checks` prints as `fail`.

BinaryBourbon (approved):
Re-reviewed the delta since my approval (`3385706b` → `0d016ab7`). Eight lines: `SandboxQueue.actor/0`. Approving.

```elixir
@doc """
The audit actor a replay runs as.

Exposed the way `Team.Schedules.actor/0` is, so a surface can tell the
drainer's replay from a person's own call without hardcoding the string.
"""
def actor, do: @system_actor
```

This is the seam part 4 needed to separate "did this caller fire" from "is anybody waiting", and exposing it beats the alternative of `Schedules` carrying its own copy of `"system:sandbox_queue"`. The actor vocabulary is closed (ADR 0013) and `system:<worker>` strings are exactly the kind of thing that goes stale in a second location — `Schedules.actor/0` already set the precedent and the docstring points at it.

`@system_actor` stays private and the accessor is the only way out, so the string has one definition.

Nothing else moved in this PR. Both of my original blockers stay fixed, and the queue-side half of #1872's expiry fix (`note_schedule_outcome/1` on `expire_overdue/1`) lands in part 4 rather than here, which is the right place — this PR owns the claim and the fence, and the schedule notification is the producer's business.

## Verified locally

At the stack tip, on a database migrated to this tree: the stack's affected files run 302 tests, 0 failures. `mix credo --strict` clean, `mix format --check-formatted` clean, `MIX_ENV=test mix compile --force --warnings-as-errors` clean — that last one because part 4 makes `SandboxQueue` and `Team.Schedules` call each other. Both directions are fully-qualified runtime calls, so there is no compile cycle.

Issue #1040, opened 2026-08-23
API auth stamps last_used_at in a linked Task.async that is never awaited — a pool blip kills the request (and flakes CI)


jhgaylor:
## What happened

Partition 3 failed on #1036 with a test that has nothing to do with the change:

```
1) test GET/PATCH/DELETE /api/team/:agent_id/schedules/:id show, update and delete,
   scoped to the tenant and the path's agent (FountainWeb.TeamScheduleControllerTest)
   test/fountain_web/controllers/team_schedule_controller_test.exs:154
   ** (EXIT from #PID<0.2273.0>) an exception was raised:
       ** (DBConnection.ConnectionError) could not checkout the connection owned by
          #PID<0.2273.0> ... Reason: connection not available and request was dropped
          from queue after 453ms
          lib/fountain/accounts.ex:798: Fountain.Accounts.Mimic.Original.Module.touch_api_key/1
          (elixir 1.19.2) lib/task/supervised.ex:105: Task.Supervised.invoke_mfa/2
   Function: #Function<0.24697844/0 in FountainWeb.Plugs.TenantAPIAuth.call/2>
```

Run: https://github.com/BinaryBourbon/fountain/actions/runs/32631856247 (partition 3). Passed on re-run of the same commit. I also hit the identical signature twice locally, in `ApiKeyControllerTest` and again here, whenever a second BEAM was contending for the 20-connection test pool. So it is load-sensitive rather than a one-off.

## Why it fails the test

`FountainWeb.Plugs.TenantAPIAuth.call/2:33`:

```elixir
Task.async(fn -> Accounts.touch_api_key(raw_key) end)
```

`Task.async/1` **links** to the caller, and this one is never awaited. When `touch_api_key/1` cannot check out a connection it raises, the task exits abnormally, and the exit signal propagates down the link to the process that started it. Under the SQL Sandbox that process is the test, so a transient pool timeout while stamping `last_used_at` kills a test whose own assertions were fine. The test looks flaky; the plug is what is fragile.

## The part that is not just a test problem

The same link exists in production. If the task raises while the request process is still alive, the request process takes the exit too — so a blip stamping `last_used_at`, a column nothing reads on the hot path, can kill an API request that had already authenticated successfully. The window is narrow (the task has to crash before the conn process finishes) but it is real, and it is pure downside: the write is best-effort by intent and the caller never wants its result.

`FountainWeb.PasswordResetController:76` has the same shape:

```elixir
[...]

PR #1871, opened 2026-09-11
feat(workers): drain the queue when a sandbox frees its slot (#1033)

jhgaylor:
Stacked on #1870.

What makes a freed slot drain the queue. `Conversations.update_sandbox/2` is the choke point every sandbox status change goes through, and a transition out of a cap-counting status frees both a tenant slot and the deployment-wide fleet slot — so the poke fans out to every tenant with live work, not just this one. A slot another tenant freed is exactly what lets this one start.

The cost on the caller's path is one existence probe against the partial index, and almost every status change happens with an empty queue. When there is work it is one Oban insert, and the job does the scan that finds the tenants — never a scan plus an insert per waiting tenant on the path that wrote the row. The poke is best-effort for the same reason metering is at that choke point: the row is already committed, nearly every call site matches `{:ok, _}`, and a failed poke must not take down a caller that only wanted to write a status.

A drain is Oban work rather than a task started from the request path. It outlives the request, it has to survive a replica going away mid-drain, and an unawaited `Task.async` linked to a web request would take that request down when a replay failed (#1040).

Uniqueness is `[:user_id, :scope]` over `:scheduled` only. Every Oban state group that includes `executing` would let a poke reporting capacity a running drain has already read past vanish into the uniqueness window. A five-minute cron is the lost-poke and expiry backstop.

`sandbox_queue_drainer_test.exs` covers the three job shapes and five properties of the poke, including that `suspended` frees a slot (it is outside `Quotas.active_statuses/0`, ADR 0017), that `pending -> ready` frees nothing, and that a raising poke never takes down the caller.

**Defers:** producers (part 4), endpoints (part 5), the metric declarations for the events already emitted (part 6), ADR and docs (part 7).

Part 3 of 7 for #1033.

Part of #1033

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -263,11 +263,39 @@ defmodule Fountain.Conversations do
     case result do
       {:ok, {was, updated}} ->
         record_sandbox_usage(was, updated)
+        maybe_poke_sandbox_queue(was, updated)
         {:ok, updated}
 
       {:error, _} = error ->
         error
     end
+  end
+
+  # A transition out of a cap-counting status frees a tenant slot and the
+  # deployment-wide fleet slot at once, so every tenant with live queue work
+  # wants draining — not just this one (ADR 0042 decision 5). This is the
+  # choke point every sandbox status change goes through and almost every one
+  # of them happens with an empty queue, so the cost here is one existence
+  # probe against a partial index. When there is work it is one Oban insert,
+  # and the job does the scan that finds the tenants.
+  defp maybe_poke_sandbox_queue(was, %Sandbox{} = updated) do
+    active = Fountain.Quotas.active_statuses()
+
+    if was in active and updated.status not in active and
+         Fountain.SandboxQueue.any_active_requests?() do
+      Fountain.Workers.SandboxQueueDrainer.poke_all_later()
+    end
+
+    :ok
+  rescue
+    # Best-effort for the same reason `Billing.record_usage/5` rescues at this
+    # choke point: the row is already committed, nearly every call site matches
+    # `{:ok, _}` (ConversationServer's terminate path, `Accounts.Deletion`,
+    # `SandboxReaper`), and a failed poke must not take down a caller that only
+    # wanted to write a status. The cron backstop drains anyway.
+    e ->
+      Logger.warning("sandbox queue poke failed: #{Exception.message(e)}")
+      :ok
   end
 
   defp prevent_sandbox_revival(changeset) do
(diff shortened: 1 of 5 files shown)


BinaryBourbon (approved):
Part 3 of 7. The poke and the worker. Approving — no findings.

`update_sandbox/2` is the right choke point and the guard is the right guard: `was in active and updated.status not in active` means `suspended` frees a slot (ADR 0017 keeps it out of `Quotas.active_statuses/0`) while `pending -> ready` frees nothing, and `sandbox_queue_drainer_test.exs` pins both. Fanning out to every tenant rather than only this one is correct for the reason given — the fleet slot is shared, so the tenant that gets to use a freed slot is usually not the one that freed it.

The cost argument holds up: one `Repo.exists?` against the partial index from part 1 on the common path, and the scan that finds the waiting tenants moved into the job rather than run on the path that wrote the row. Best-effort with a rescue is consistent with `Billing.record_usage/5` at the same choke point, and the "nearly every call site matches `{:ok, _}`" justification is accurate — `ConversationServer`'s terminate path, `Accounts.Deletion` and `SandboxReaper` all do.

Oban over `Task.async` needs no defending after #1040, and the `states: :scheduled` argument is right: every group that includes `executing` would let a poke reporting capacity a running drain has already read past disappear into the window, which is the one failure mode that actually stalls the queue.

One thing worth writing down somewhere, since the code comment reads as if the window were 30 seconds: with `schedule_in: 1`, a job leaves `scheduled` about a second after insert, so `period: 30` over `states: :scheduled` is an effective dedup window of ~1s, not 30. That is the correct side of the trade — a swallowed poke is worse than a redundant drain — but it means sustained sandbox churn with a non-empty queue inserts roughly one fan-out per second plus one job per waiting tenant, and each of those does a real replay attempt. It is the mechanism behind the audit/broadcast churn I raised on #1872. Nothing to change here.

`Mimic.copy(Fountain.Workers.SandboxQueueDrainer)` in `test_helper.exs` and the five poke properties (including that a raising poke does not take the caller down) are exactly the right things to have pinned.

BinaryBourbon (commented):
Re-checked this one after the stack was pushed. **No change of its own.** The cumulative diff at `2fa86d51` is byte-for-byte the cumulative diff at `3385706b`, so everything new in `581ea919` → `2fa86d51` arrived from part 2 underneath it. My approval stands unmodified.

The parts of #1870's delta that reach into this PR's territory both land well: `claim_next/2` skipping the row it lost the swap on makes the drain loop's termination structural rather than incidental, and `finish/3` deriving its audit action from the status removes the `CaseClauseError` that would have taken down the rest of a tenant's queue at the moment somebody added a third terminal outcome.

`team/schedules_test.exs` and the queue suites pass at the stack tip (151 tests, 0 failures) on a database migrated to this tree.

One thing to watch before merging rather than after: CI has reported nothing on this head. Every job on the 05:26 run is `cancelled` by concurrency, which `gh pr checks` renders as `fail`.

Issue #1126, opened 2026-08-25
Credits cleanup 1/3: ledger correctness bugs and customer-facing falsehoods left by ADR 0031


jhgaylor:
From the 2026-08-25 billing review after ADR 0031 (credits are the product). The pricer no-op and the comped-rent bug are already fixed in #1124; this issue is the rest of the **bugs**. One PR; every item gets a regression test. Files are `ee/lib/...` unless noted.

## Money / correctness
1. **Expiry can claw back paid credit.** `workers/credit_granter.ex` `expire_grant/1` reads `Credits.unspent_of/2` outside the lock, then `Credits.debit/4` consumes inside it; a burn in between makes `consume_lots` (credits.ex) fall past the named lot into earliest-expiry then purchased money. Fix: when `lot_id:` is given, `consume_lots` takes only from that lot (cap at its `remaining_cents`); compute the amount inside the transaction.
2. **Refund after dispute double-claws.** `credits/purchases.ex` `refund/1` computes "already clawed back" over `clawback_refund` rows only; a charge disputed then refunded is taken twice. Sum `clawback_refund` + `clawback_dispute` for the charge (dispute rows carry the charge id in metadata).
3. **Expired-but-unswept credit is spendable until the 06:23 sweep.** `live_grants/2` excludes expired lots, so `summary/2`'s `expiring_cents`/`purchased_cents` disagree with `credit_balance_cents` for up to 24h, and `check_balance/1` reads the cache. Run the expiry pass on the pricer's 10-minute tick (or have the pricer call it), and consider subtracting expired-lot remainders in `check_balance/1`.
4. **`Rent.charge/3` checks the balance before idempotency** (`credits/rent.ex`): a re-charge of an already-paid month on a short balance returns `:insufficient_credits` and stamps `rent_due_at`, starting a release clock for rent already collected. Check the idempotency key first.
5. **`credit_ledger` ordering.** The open-lots partial index is `(user_id, expires_at, inserted_at)` but `consume_lots` and `live_grants` order by `expires_at, seq`; `list_entries/2` orders by `inserted_at, id` (random uuid, second-resolution) so a grant and its burn in the same second render in random order on the billing page. Migration: index third column → `seq`; `list_entries/2` → `desc: e.seq`.
6. `Credits.audit_action/1` has no fallback clause (raises after the row commits); `rebuild_lots/1` ends in `then(fn {:ok, n} -> n end)` and crashes on rollback.

## Customer-facing text that is now false
[...]

PR #1872, opened 2026-09-11
feat(api): queue: true on a start, and a cron firing that waits (#1033)

jhgaylor:
Stacked on #1871.

The two producers, and they opt in differently on purpose.

**`queue: true` on `POST /api/conversations`** turns a tenant-cap 429 or a fleet-cap 503 into 202 with the request and its position. Every existing caller keeps the error its client already handles. Starts carrying images do not queue — the server does not hold image bytes for an hour — and starts naming an explicit `sandbox_id` do not either, being an attach to a machine the caller already has rather than a request for capacity. At the depth bound the caller gets the capacity error it was about to get anyway: the queue delays the cap, it never raises it.

What a queued start stores is an **allow list** of the launch keys `start_conversation/2` reads, not a drop list of the ones this path sets. `ConversationCreateRequest` does not set `additionalProperties: false`, so a drop list would park whatever else a caller sent in the table until the request expired. The list is today's keys, which is why it carries three #1481 did not have: `labels` (#1637), `execution_limits` and `sandbox_api_access`.

**A cron firing of a teammate schedule** opts in by itself, because it has nobody present to retry it — a 09:00 run that meets a fan-out at 08:59 is otherwise simply lost. The gate is the audit actor, so the page's and the API's "Run now" still gets the refusal: queueing there would answer a person with an error while a conversation started on its own up to an hour later, with no request id in the response to watch or cancel. The queue's own replay (`actor: "system:sandbox_queue"`) is excluded by the same test, so a replay cannot stack a second request behind the one it is replaying. `describe_error(:fleet_full)` gains the words it was missing.

Generated artifacts (`sdk/contract/contract.json`, `sdk/typescript/src/generated/openapi.ts`) are rebuilt with `scripts/sdk-contract/build.sh` and `npm run generate`, not hand-edited. **No SDK version bump here** — the bump lands once, in part 7. Label this `sdk-no-release`.

**Defers:** `/api/sandbox-queue` (part 5), telemetry declarations (part 6), ADR 0042 and the docs (part 7).

Part 4 of 7 for #1033.

---

### Review fixes (round 2)

- **`maybe_enqueue/4` stores `SandboxKey.id(conn)`**, closing the ADR 0045 bypass: a queued start is now held to the same label rule as the door it came through. Column in part 1, replay in part 2.
- **A waiting firing keeps saying it is waiting.** `await_capacity/2` replaces `maybe_queue_run/3` and answers "is this work waiting" rather than "who may enqueue" — so the drainer's own replay, which may meet the ceiling again, no longer flips `last_error` back to `sandbox quota: N/N`. Who enqueues is unchanged, and both existing gate tests pass untouched.
- **A waiting firing is not a run.** It stamps no `last_run_at` and records no `team.schedule.fired`; `sandbox_request.enqueued` is the event that happened. Previously each replay attempt wrote a firing row and re-stamped the timestamp.
- The `sandbox_id` clause is documented as belt and braces — `attach_conversation/4` takes no reservation, so it cannot raise a capacity error for it to intercept, and no honest test of it exists.

---

### Review fixes (round 3)

Two regressions the round-2 fix introduced, both found by re-review and both reproduced as tests before changing anything:

- **A person's "Run now" during a queued window recorded nothing.** `waiting?` answered "is anybody waiting" where it needed "did *this caller* wait", so `unless waiting?` suppressed the audit row for a human action too. Now two values: `waiting?` decides what the row says, `queued_by_us?` decides whether this caller fired. A person refused while work waits is stamped and recorded with the refusal they got, while the row goes on reporting the wait.
- **The row said "waiting" forever once the request died.** Nothing retracted `@waiting` on expiry or cancellation, and a `one_off` has no next firing to self-correct. `SandboxQueue` now calls `Schedules.note_queued_run_ended/3` from `expire_overdue/1` and `cancel_request/2` — `timed out waiting for a free sandbox slot` or `the queued run was cancelled`.

Four new tests, plus `SandboxQueue.actor/0` in part 2 and ADR 0042 decision 3 extended to cover both halves.

Part of #1033

### DIFF
--- a/apps/fountain/lib/fountain/sandbox_queue.ex
+++ b/apps/fountain/lib/fountain/sandbox_queue.ex
@@ -199,6 +199,7 @@ defmodule Fountain.SandboxQueue do
     case swap(request.id, "queued", %{status: "cancelled", attrs: %{}}) do
       {:ok, cancelled} ->
         audited(cancelled, "sandbox_request.cancelled", opts)
+        note_schedule_outcome(cancelled)
         emit_depth(cancelled.user_id)
         {:ok, cancelled}
 
@@ -402,6 +403,17 @@ defmodule Fountain.SandboxQueue do
   defp put_unless_nil(attrs, _key, nil), do: attrs
   defp put_unless_nil(attrs, key, value), do: Map.put(attrs, key, value)
 
+  # A `schedule_run` that ended without running has to tell its schedule, or
+  # that row keeps reporting a wait that is over. `started` and `failed` went
+  # through `run_schedule/2`, which wrote the row itself; `expired` and
+  # `cancelled` never reach it.
+  defp note_schedule_outcome(%Request{kind: "schedule_run", schedule_id: id} = request)
+       when is_binary(id) do
+    Fountain.Team.Schedules.note_queued_run_ended(id, request.user_id, request.status)
+  end
+
+  defp note_schedule_outcome(%Request{}), do: :ok
+
   defp expire_overdue(user_id) do
     cutoff = DateTime.add(DateTime.utc_now(), -max_wait_seconds(), :second)
 
@@ -415,6 +427,7 @@ defmodule Fountain.SandboxQueue do
       case swap(request.id, "queued", %{status: "expired", attrs: %{}}) do
         {:ok, expired} ->
           audited(expired, "sandbox_request.expired", actor: @system_actor)
+          note_schedule_outcome(expired)
           emit_completed(expired, request.inserted_at)
           true
 
--- a/apps/fountain/lib/fountain/team/schedules.ex
+++ b/apps/fountain/lib/fountain/team/schedules.ex
@@ -36,6 +36,49 @@ defmodule Fountain.Team.Schedules do
 
   @actor "system:team_scheduler"
 
+  @doc """
+  What a schedule row says once its queued run ended without ever running.
+
+  `Fountain.SandboxQueue` calls this on the two terminal transitions that never
+  reach `run_schedule/2`: a request that expired at the wait bound, and one
+  somebody cancelled. Without it `last_error` keeps reporting a wait that is
+  over — a cron schedule self-corrects at its next firing, but a `one_off` has
+  no next firing, so its row would claim it was waiting for a slot nothing is
+  waiting for.
+
+  No `last_run_at`: nothing ran. Not audited either, because
+  `sandbox_request.expired` and `sandbox_request.cancelled` already record the
+  event this mirrors onto the schedule, and the trail is not a second copy of
+  it (ADR 0013).
+  """
+  def note_queued_run_ended(schedule_id, user_id, status)
+      when is_binary(schedule_id) and is_binary(user_id) and is_binary(status) do
+    case get_schedule(schedule_id, user_id) do
+      nil ->
+        :ok
+
+      schedule ->
+        {:ok, _} =
+          schedule
+          |> Schedule.run_changeset(%{last_error: describe_queue_outcome(status)})
+          |> Repo.update()
+
+        Team.broadcast_schedules_changed(user_id)
+        :ok
+    end
+  end
+
+  defp describe_queue_outcome("expired"), do: "timed out waiting for a free sandbox slot"
+  defp describe_queue_outcome("cancelled"), do: "the queued run was cancelled"
+  defp describe_queue_outcome(status), do: "the queued run ended: #{status}"
+
+  # What the row says while work sits in the queue. Reported by every caller,
+  # not only the one that enqueued: the drainer replays with
+  # `actor: "system:sandbox_queue"`, and a replay that met the ceiling again
+  # must leave the row saying "waiting" rather than flip it back to the
+  # capacity error the queue exists to absorb (ADR 0042 decision 3).
+  @waiting "waiting for a free sandbox slot"
+
   @doc "The audit actor a run fires as."
   def actor, do: @actor
 
@@ -161,9 +204,14 @@ defmodule Fountain.Team.Schedules do
   now". Returns `{:ok, conv}` with the conversation the prompt went to, or
   the `Team.send_message/5` / `Conversations.start_conversation/2` error
   unchanged (`:busy`, `:provisioning`, `:not_found`, `{:sandbox_quota_exceeded, _}`,
-  ...). Either way `last_run_at` is stamped, and `last_conversation_id` /
+  ...). A firing that ran stamps `last_run_at`, and `last_conversation_id` /
   `last_error` say how it went; the caller decides whether an error is worth
   retrying (the worker snoozes on `:busy` and `:provisioning`).
+
+  A capacity refusal is queued for the cron caller and nobody else — see
+  `await_capacity/2`. "Run now" gets the error unchanged. A firing that is
+  waiting on capacity stamps no `last_run_at` and records nothing: it is not a
+  run, and the request's own trail says what happened.
 
   `opts` is audit attribution: `Fountain.Workers.TeamScheduleRun` passes
   `actor: "system:team_scheduler"`, the UI passes the socket's. Recorded as
@@ -177,31 +225,69 @@ defmodule Fountain.Team.Schedules do
 
     now = DateTime.utc_now() |> DateTime.truncate(:second)
 
-    run_attrs =
+    # Two different questions, and conflating them is what made a person's
+    # refused "Run now" stop being recorded at all.
+    #
+    # `waiting?` decides what the ROW says. A live request means this
+    # schedule's work is waiting, whoever asked.
+    #
+    # `queued_by_us?` decides whether this CALLER fired. The cron caller
+    # queued instead of firing, and the drainer is replaying a firing already
+    # on the trail; both are silent. Anybody else fired and got an answer, and
+    # a refusal that writes the row is a mutation that records (ADR 0013).
+    waiting? =
       case result do
-        {:ok, conv} -> %{last_run_at: now, last_conversation_id: conv.id, last_error: nil}
-        {:error, reason} -> %{last_run_at: now, last_error: describe_error(reason)}
+        {:error, {:sandbox_quota_exceeded, _}} -> await_capacity(schedule, opts)
+        {:error, :fleet_full} -> await_capacity(schedule, opts)
+        _ -> false
+      end
+
+    queued_by_us? = Keyword.get(opts, :actor) in [@actor, Fountain.SandboxQueue.actor()]
+
+    run_attrs =
+      cond do
+        # Nothing fired, so `last_run_at` keeps saying when the schedule last
+        # actually ran.
+        waiting? and queued_by_us? -> %{last_error: @waiting}
+        # Somebody fired and was refused while the work waits. The attempt is
+        # stamped and recorded; the row still reports the wait, because that is
+        # what is true of the schedule.
+        waiting? -> %{last_run_at: now, last_error: @waiting}
+        true -> ran_attrs(result, now)
       end
 
     {:ok, _} = schedule |> Schedule.run_changeset(run_attrs) |> Repo.update()
     Team.broadcast_schedules_changed(schedule.user_id)
 
-    record(
-      "team.schedule.fired",
-      schedule,
-      opts,
-      Map.merge(describe(schedule), %{
-        "outcome" => if(match?({:ok, _}, result), do: "ok", else: run_attrs.last_error),
-        "conversation_id" =>
-          case result do
-            {:ok, conv} -> conv.id
-            _ -> nil
-          end
-      })
-    )
+    unless waiting? and queued_by_us? do
+      record(
+        "team.schedule.fired",
+        schedule,
+        opts,
+        Map.merge(describe(schedule), %{
+          # The outcome this caller got, not what the row displays: a person
+          # refused while work waits was refused, and the trail should say so.
+          "outcome" => outcome(result),
+          "conversation_id" =>
+            case result do
+              {:ok, conv} -> conv.id
+              _ -> nil
+            end
+        })
+      )
+    end
 
     result
   end
+
+  defp outcome({:ok, _conv}), do: "ok"
+  defp outcome({:error, reason}), do: describe_error(reason)
+
+  defp ran_attrs({:ok, conv}, now),
+    do: %{last_run_at: now, last_conversation_id: conv.id, last_error: nil}
+
+  defp ran_attrs({:error, reason}, now),
+    do: %{last_run_at: now, last_error: describe_error(reason)}
 
   # A fresh conversation with what the teammate has: its agent, and the
   # environment override and vault its current team conversation carries.
@@ -282,6 +368,42 @@ defmodule Fountain.Team.Schedules do
     end
   end
 
+  # Scheduled work opts into the bounded queue: unlike an interactive caller,
+  # a cron firing has nobody present to retry it when capacity frees, and a
+  # 09:00 run that meets a fan-out at 08:59 is simply lost.
+  #
+  # Only the cron caller enqueues. `run_schedule/2` is also the page's and the
+  # API's "Run now" (`POST /api/team/:agent_id/schedules/:id/run`), and
+  # queueing there would hand a person a 429 or 503 while a conversation
+  # started on its own up to an hour later, with no request id in the response
+  # to watch or cancel. ADR 0042 decision 3 buys the queue with "a cron firing
+  # has nobody there to retry it", which is exactly what somebody pressing a
+  # button is not. The queue's own replay is excluded by the same test: the
+  # drainer already holds the claim, so re-entering would stack a second
+  # request behind the one being replayed.
+  #
+  # Returns whether the work is waiting, which is a different question from
+  # who was allowed to put it there. A depth-bound refusal leaves no request,
+  # so the caller gets the capacity error it was about to get anyway.
+  defp await_capacity(%Schedule{} = schedule, opts) do
+    if Keyword.get(opts, :actor) == @actor, do: enqueue_run(schedule, opts)
+
+    Fountain.SandboxQueue.schedule_request(schedule.user_id, schedule.id) != nil
+  end
+
+  defp enqueue_run(%Schedule{} = schedule, opts) do
+    Fountain.SandboxQueue.enqueue(
+      %{
+        user_id: schedule.user_id,
+        agent_id: schedule.agent_id,
+        kind: "schedule_run",
+        schedule_id: schedule.id,
+        source: "schedule"
+      },
+      opts
+    )
+  end
+
   # What a run's failure reads as on the row and in the UI. Never the prompt.
   def describe_error(:busy), do: "teammate was busy"
 
@@ -291,6 +413,7 @@ defmodule Fountain.Team.Schedules do
   def describe_error(:provisioning), do: "teammate's computer was still starting"
   def describe_error(:not_found), do: "agent is not on the team"
   def describe_error(:insufficient_credits), do: "out of credit"
+  def describe_error(:fleet_full), do: "sandbox fleet is full"
   def describe_error(:runner_offline), do: "teammate's machine is offline"
   def describe_error(:sprite_probe_failed), do: "could not reach the sandbox provider"
 
--- a/apps/fountain/lib/fountain_web/controllers/conversation_controller.ex
+++ b/apps/fountain/lib/fountain_web/controllers/conversation_controller.ex
@@ -492,6 +492,9 @@ defmodule FountainWeb.ConversationController do
       conflict: {"Conflicting state", "application/json", Schemas.Error},
       service_unavailable: {"Sandbox or fleet unavailable", "application/json", Schemas.Error},
       created: {"Conversation", "application/json", Schemas.ConversationResponse},
+      accepted:
+        {"Queued for sandbox capacity", "application/json", Schemas.SandboxRequestResponse},
+      too_many_requests: {"Tenant concurrency cap reached", "application/json", Schemas.Error},
       ok:
         {"Conversation (resumed by channel_id)", "application/json", Schemas.ConversationResponse},
       forbidden:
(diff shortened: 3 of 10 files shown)


BinaryBourbon (changes requested):
Part 4 of 7. The two producers. Requesting changes — one security-relevant gap and one visibility regression against what the ADR promises. The opt-in design itself is right and I would not change it.

Both opt-in decisions are well argued and the *asymmetry* is the good part: `queue: true` keeps every existing client's contract, and gating the schedule path on the audit actor rather than on a parameter means "Run now" keeps its refusal and the queue's own replay cannot re-enter the queue. I confirmed `@actor` is `"system:team_scheduler"` and that `Workers.TeamScheduleRun` passes `Schedules.actor()`, so the gate discriminates the three callers it claims to. The allow list being an allow list is right for the stated reason — `ConversationCreateRequest` has no `additionalProperties: false`, so a drop list would park arbitrary caller JSON in the table for an hour — and I checked it against every `attrs[...]` read in `start_conversation/2` and `start_or_resume_conversation/2`: the list is complete except for `source`, which is intentional here and is the bug I raised on #1870.

`params["queue"] == true` being strict is fine — `CastAndValidate` rejects `"true"` for a boolean before the controller sees it, so there is no silently-ignored-flag path.

## Blocking: the replay drops `sandbox_key_id` and bypasses the ADR 0045 label rule

The door builds `opts = SandboxKey.opts(conn) ++ Audited.attribution(conn)`. The replay in part 2 builds `actor: "system:sandbox_queue"` and nothing else. `sandbox_key_id` is a *restriction* opt, and its absence is read as "no sandbox restriction applies":

```elixir
defp sandbox_owns?(_conv, nil), do: true
defp sandbox_owns?(%Conversation{callback_api_key_id: id}, key_id), do: id == key_id
```

So the path is: a `sprite`-scoped token posts `POST /api/conversations` with `queue: true`, a `channel_id` and `labels`. `POST /api/conversations` is in the plain `[:accepts_json, :api]` scope, not behind `:require_full_scope`, so a sandbox token reaches it — that is what `source: "agent"` is. At the door, `find_channel_conversation/5` returns `nil` (otherwise there is no capacity error to queue), so the request is stored with `channel_id` and `labels` in `attrs`. Within the next hour a conversation binds to that `(user, agent, vault, env, channel_id)` key. The drain replays, `start_or_resume_conversation/2` now takes the resume branch, and `resume_labels(conv, attrs["labels"], opts)` → `set_conversation_labels/4` with no `sandbox_key_id` → `sandbox_owns?(_, nil) -> true` → the labels land on a conversation that token does not own.

That is the exact write `set_conversation_labels/4`'s own docstring says must be refused, and its reasoning applies verbatim: *"the rule has to hold on the door rather than on whichever of them remembered."* The queue is a new door, and it forgot. Narrow to exploit — it needs a sprite token, a capacity refusal, and a channel binding appearing inside the wait window — but ADR 0045 exists for precisely this class, and an hour-long queue is a better window than the synchronous path ever offered.

Cheapest correct fix: carry the restriction. Add a `sandbox_key_id` column to `sandbox_requests` (part 1's migration is unmerged, so it can go in there), set it from `SandboxKey.id(conn)` in `maybe_enqueue/4`, and have `attempt/1` pass `sandbox_key_id: request.sandbox_key_id` alongside the actor when it is non-nil. A test that queues as a sprite token and asserts the replay refuses to relabel another conversation would pin it.

If you would rather not grow the schema: refuse to queue when `SandboxKey.id(conn)` is non-nil and the body carries both `channel_id` and `labels`, in the same breath as images and `sandbox_id`. Less good — it declines useful work to avoid carrying one field — but it closes the hole and is three lines. Either way it wants a line in ADR 0042 decision 2 next to the images and `sandbox_id` exclusions, because "which restrictions survive a replay" is the general question and this is the first instance of it.

## Blocking: the queued schedule stops saying it is queued

`maybe_queue_run/3` writes `"waiting for a free sandbox slot"` on `last_error`, and ADR 0042 decision 3 states that as the behaviour. It survives until the first drain pass and no longer.

On replay the actor is `"system:sandbox_queue"`, so `maybe_queue_run/3` falls to `describe_error(reason)` and `last_error` becomes `"sandbox quota: 5/5"` again. Since `run_schedule/2` runs its full tail on every attempt, each replay also re-stamps `last_run_at`, writes a `team.schedule.fired` audit row with an error outcome, and calls `Team.broadcast_schedules_changed/1`. With the ~1s effective poke dedup (see my note on #1871), a schedule waiting out a real capacity spike can accumulate a lot of those, and the one thing a person looking at the row wants to know — *it is queued, not failed* — is the thing that gets overwritten first.

Issue #1033 point 7 asks for exactly the opposite: "something in the team/conversations apps that says *waiting for a free slot* rather than showing a conversation that looks stuck".

I checked that this is display-only and does not corrupt scheduling: `next_run_at` is computed independently and nothing reads `last_run_at` to decide the next firing, so there is no double-fire or skipped-fire here. That is why this is a visibility bug rather than a correctness one.

Two ways out, either fine:

- Have the queue's replay skip the row-stamping tail — pass an opt through `run_schedule/2` that suppresses the `last_run_at` / `last_error` / audit / broadcast update while the request is still live, and let the terminal outcome write the row once. The `sandbox_request.*` audit events already carry the per-attempt history, so nothing is lost.
- Or keep the tail but make the queued state win: when a live `schedule_run` request exists for this schedule, write `"waiting for a free sandbox slot"` regardless of actor. Smaller change, keeps the row honest, still writes an audit row per attempt.

Whichever you pick, ADR 0042 decision 3's sentence should end up describing what the row actually says after the second attempt.

## Non-blocking

`describe_error(:fleet_full)` gaining words is a good catch and the test for it (`"a full fleet is described in words too (#1033)"`) is the right shape — the same gap #1126 fixed for the quota atom.

The `schedules_test.exs` trio ("the cron firing waits", "does not stack copies", "the queue's own replay does not re-enter the queue") covers all three arms of the actor gate, which is what I would want given the gate is the whole design.

jhgaylor:
Both blockers fixed. Force-pushed.

**The ADR 0045 bypass.** You traced it exactly right, including that `POST /api/conversations` is in the plain `[:accepts_json, :api]` scope. Fixed by carrying the restriction rather than by refusing to queue, so a sprite token keeps the feature:

- part 1 adds a `sandbox_key_id` column. Deliberately **not** a `references/2` — a `nilify_all` would turn the restriction into a permission the request never had when a key is deleted, and a dangling id fails the ownership comparison instead, which is the safe direction. The migration says so, next to `schedule_id` for the same reason;
- this PR sets it from `SandboxKey.id(conn)` in `maybe_enqueue/4`;
- part 2 passes it back through `replay_opts/1`.

Three tests: the sprite token's request stores the key id, an owner's own key stores `nil` (so a full-scope caller cannot pick a restriction up by accident), and on part 2 the end-to-end refusal. Reverting the replay half makes that last one fail with `started: 1` — the relabel going through.

ADR 0042 decision 2 now states the general rule, since this will not be the last one: **a replay carries the restrictions the door was under, not only its attributes.**

**The schedule row stopping saying it is queued.** Took your first option, and it turned out to fold the audit churn in with it. `run_schedule/2` now asks whether the work is waiting rather than who was allowed to queue it:

```elixir
defp await_capacity(%Schedule{} = schedule, opts) do
  if Keyword.get(opts, :actor) == @actor, do: enqueue_run(schedule, opts)

  Fountain.SandboxQueue.schedule_request(schedule.user_id, schedule.id) != nil
end
```

The gate on *who enqueues* is untouched, so "Run now" still gets its refusal and the queue's replay still cannot re-enter the queue — both existing tests pass unchanged. What changes is that every caller reports a wait that is already underway. And since a waiting firing is not a firing:

- it stamps no `last_run_at` (the row keeps saying when the schedule last actually ran);
- it records no `team.schedule.fired`. `sandbox_request.enqueued` is the event that happened, and a trail that logged each replay attempt as a firing would report a dozen for one run — the "only record what happened" rule.

Three tests: the row and trail after a waiting firing, a replay that meets the ceiling again still saying "waiting", and a depth-bound refusal reporting the capacity error with `last_run_at` stamped, because that request really was refused.

`await_capacity/2` returning "is it waiting" rather than a message is also why `describe_error/1` no longer needs to be called from two places.

**On your `sandbox_id` note (raised on #1873):** I went to add that test and could not write an honest one. `attach_conversation/4` takes no sandbox reservation, so it cannot return `sandbox_quota_exceeded` or `:fleet_full` — there is no capacity error at that door for the clause to intercept. My first attempt "passed" only by getting a 422 `sandbox_identity_mismatch` from a sandbox built for another agent, which tests nothing. So the clause is belt and braces, and the comment now says that plus why it stays: a future attach that *does* reserve must not start queueing by accident.

BinaryBourbon (changes requested):
Re-reviewed the delta since my changes-requested (`8fd82389` → `fe83cca7`). **Both original blockers are properly fixed.** Requesting changes again, on two things the second fix introduced — both narrow, both cheap, and I would not have found either by reading. I ran them.

## Blocker 1: the replay dropped `sandbox_key_id` — fixed, and fixed the good way

You took the schema option rather than the three-line refusal, which is the right trade: the queue now carries the restriction instead of declining work to avoid carrying it. `SandboxKey.id(conn)` at the door, a column in part 1's migration, `replay_opts/1` in part 2. Two controller tests pin both ends — a sprite token's queued start stores `key.id`, a full-scope caller's stores `nil` and does not pick a restriction up by accident.

And the generalisation made it into ADR 0042 decision 2: *"Anything that narrows what a request may do has to survive the wait, or the queue is a way around it."* That sentence is what I actually wanted out of this, more than the column.

## Blocker 2: the queued schedule stopped saying it was queued — fixed, and then some

You went further than either option I offered. A waiting firing now stamps no `last_run_at`, writes `@waiting` on `last_error`, and records no `team.schedule.fired` — because it did not fire, and `sandbox_request.*` is the trail for the wait. `"a replay that meets the ceiling again keeps saying it is waiting"` drains once with capacity still full and asserts the row survives it, which is the regression I asked for. `"a refusal past the depth bound reports the capacity error, not a wait"` covers the other arm.

That is a better answer than "make the queued state win". But the predicate it rests on is not quite the one the behaviour needs.

---

## Blocking: a person's "Run now" during a queued window records nothing at all

`await_capacity/2` (`schedules.ex:329`) decides `waiting?` from **whether a live request exists**, not from whether this caller put one there or is replaying one:

```elixir
if Keyword.get(opts, :actor) == @actor, do: enqueue_run(schedule, opts)

Fountain.SandboxQueue.schedule_request(schedule.user_id, schedule.id) != nil
```

Who enqueues is gated on the actor. Who is *told they are waiting* is not. So a person pressing "Run now" on a schedule that already has a cron request in the queue takes the `waiting?` branch: `unless waiting?` (`schedules.ex:208`) suppresses the audit row, and `run_attrs` (`:199`) writes only `last_error`.

I ran it rather than reasoning about it — cron fires and queues, then `run_schedule(schedule, actor: "ui", request_ip: "1.2.3.4")`:

```
PROBE A: fired rows before=0 after=0
PROBE A: last_run_at=nil last_error="waiting for a free sandbox slot"
```

The person's action updated the row and left **zero** audit trace. There is no `sandbox_request.*` event either, because they did not enqueue. Before this delta it recorded `team.schedule.fired` with `outcome: "sandbox quota: 5/5"`; on `main` it always records one. A mutation that records nothing is the thing `audit_guardrail_test.exs` exists to prevent, and it stays green here because `run_schedule/2` still audits on its other branch.

The window is not exotic: a queue full of schedule requests and a "Run now" that meets the ceiling both require capacity pressure, so this is the *likely* coincidence, not the unlikely one.

The return value is right — the person gets `{:error, {:sandbox_quota_exceeded, _}}` unchanged, exactly as decision 3 promises. It is only the row and the trail that go quiet.

Fix: make `waiting?` mean "this caller is the one waiting", not "somebody is". The cron actor and the drainer's replay actor are the two callers that wait; everyone else fired and was refused. Something like

```elixir
waiting? = Keyword.get(opts, :actor) in [@actor, Fountain.SandboxQueue.actor()] and request?
```

with `SandboxQueue.actor/0` exposing `"system:sandbox_queue"` the way `Schedules.actor/0` already exposes its own. A test that presses "Run now" against a live request and asserts a `team.schedule.fired` row still lands would pin it.

## Blocking: once the request dies, the row still says it is waiting — forever, for a one-off

Nothing retracts `@waiting`. `expire_overdue/1` marks the request `expired` and tells the schedule nothing; `cancel_request/2` (part 5, `DELETE /api/sandbox/requests/:id`) is the same. Measured, with the request aged past the wait bound and one drain pass:

```
PROBE B: after expiry last_error="waiting for a free sandbox slot" last_run_at=nil
```

A cron schedule self-corrects at its next firing. A `one_off` has no next firing, so the row claims it is waiting for a slot that nothing is waiting for, permanently — and `list_queued/1` only returns `queued` rows, so `GET /api/sandbox/requests` will not show the expired one either. The person looking at the schedule sees something that looks stuck with nothing to click.

Issue #1033 point 7 asks for *"something in the team/conversations apps that says waiting for a free slot rather than showing a conversation that looks stuck"*. This is that sentence with the sign flipped: the wait is over and the row is the thing that looks stuck.

The old behaviour was wrong for a few seconds; this is wrong until something else happens to the schedule. That is why I am blocking rather than filing it.

Fix: a terminal transition on a `schedule_run` request should write the outcome back — `last_run_at` plus `"timed out waiting for a free sandbox slot"`, or the failure reason. The seam is `finish/3` and `expire_overdue/1` in part 2; landing it there or here is your call, they are both unmerged. Whichever, decision 3's paragraph should end by saying what the row reads after the wait ends, since it now says what it reads during.

## Non-blocking

The `maybe_enqueue/4` comment now says the `sandbox_id` arm is belt and braces with no test behind it, because `attach_conversation/4` takes no reservation and cannot raise either capacity error. Good — an untested arm that says so is fine; an untested arm that reads as covered is not.

`ran_attrs/2` split out of the `case` reads better than what it replaced, and `enqueue_run/2` no longer having to decide what the row says is the right division.

## Verified locally

At the stack tip, on a database migrated to this tree: `sandbox_queue_test.exs`, `sandbox_queue_drain_test.exs`, `team/schedules_test.exs`, `sandbox_queue_controller_test.exs`, `audit_guardrail_test.exs` — 151 tests, 0 failures; `conversation_controller_test.exs`, `marketing_pricing_test.exs`, `ops_gauges_test.exs`, `retention_pruner_test.exs`, `docs_test.exs` — 147 tests, 0 failures. `mix credo --strict` clean, `mix format --check-formatted` clean, `scripts/sdk-contract/build.sh --check` ok. Neither probe above is in the suite; both pass as written because nothing asserts on them.

Note for whoever merges: CI has reported no verdict on any head in this stack. Every job on the 05:26 runs is `cancelled` by concurrency, and `gh pr checks` prints `cancelled` as `fail` with a zero duration. That column is not evidence either way right now.

jhgaylor:
Both blockers confirmed and fixed. I reproduced your two probes as tests before changing anything, and they measured exactly what you measured:

```
PROBE A: fired rows before=0 after=0
PROBE A: last_run_at=nil last_error="waiting for a free sandbox slot"
PROBE B: after expiry last_error="waiting for a free sandbox slot" last_run_at=nil
```

Both are mine and both are worse than what they replaced. Thank you for running them rather than reading.

## Blocker 1: a person's "Run now" recorded nothing

You diagnosed the cause precisely — the predicate answered "is anybody waiting" where the behaviour needed "did this caller wait". I did not take your suggested fix, because it trades the bug for a smaller copy of the original one: making `waiting?` false for a person's call sends them down `ran_attrs/2`, which writes `last_error: "sandbox quota: 2/2"` over `@waiting` — the row flipping back to the capacity error, just triggered by a button instead of by the drainer.

The two questions are genuinely separate, so they are now two values:

```elixir
# `waiting?` decides what the ROW says. A live request means this schedule's
# work is waiting, whoever asked.
# `queued_by_us?` decides whether this CALLER fired.
queued_by_us? = Keyword.get(opts, :actor) in [@actor, Fountain.SandboxQueue.actor()]

run_attrs =
  cond do
    waiting? and queued_by_us? -> %{last_error: @waiting}
    waiting? -> %{last_run_at: now, last_error: @waiting}
    true -> ran_attrs(result, now)
  end
```

So a person refused during a queued window is **stamped and recorded**, while the row goes on reporting the wait — which is what is true of the schedule. `SandboxQueue.actor/0` is exposed in part 2 next to `@system_actor`, as you suggested.

One thing I changed beyond the predicate: the audit outcome no longer comes from `run_attrs.last_error`. A person refused while work waits got a refusal, and the trail should say `sandbox quota: 2/2` even though the row displays the wait. New `outcome/1` does that.

After the fix: `PROBE A: fired rows before=0 after=1`, `last_run_at` stamped, `last_error` still `"waiting for a free sandbox slot"`. Pinned by *"a person's \"Run now\" during a queued window is still recorded"*, which asserts the actor is `ui` and the recorded outcome contains `sandbox quota`.

## Blocker 2: the row said "waiting" forever

Also real, also mine, and the one-off case is the one that makes it bad. Fixed at the seam you pointed at. `SandboxQueue` now tells the schedule when a `schedule_run` ends without running:

```elixir
# `started` and `failed` went through `run_schedule/2`, which wrote the row
# itself; `expired` and `cancelled` never reach it.
defp note_schedule_outcome(%Request{kind: "schedule_run", schedule_id: id} = request)
     when is_binary(id) do
  Fountain.Team.Schedules.note_queued_run_ended(id, request.user_id, request.status)
end
```

called from `expire_overdue/1` and `cancel_request/2`. `note_queued_run_ended/3` writes `timed out waiting for a free sandbox slot` or `the queued run was cancelled`, stamps no `last_run_at` because nothing ran, and is deliberately not audited — `sandbox_request.expired` / `.cancelled` already record the event it mirrors, and the trail is not a second copy of it.

`PROBE B: after expiry last_error="timed out waiting for a free sandbox slot"`. Three tests: expiry, cancellation, and a `start` request ending touching no schedule at all (so the `kind` guard cannot rot into a lookup on a request that has no schedule).

I put both in part 4 rather than part 2 — the queue-side hook is two lines, and keeping the schedule's semantics in the PR that owns `run_schedule/2` seemed better than splitting them. Say the word if you would rather they moved.

ADR 0042 decision 3 now has both halves: what the row says during the wait and who is silent, and what it says once the wait ends without a run.

## On the CI note

Agreed, and thank you for writing it down — the 05:26 runs are all `cancelled` by concurrency because an atomic push of all seven branches updates each PR's head *and* its base, so GitHub queues several `pull_request` runs per branch and kills the older ones. `CI required` then fails on `"changes": {"result": "cancelled"}` rather than on anything in the tree. The runs that replaced them were filling in green (15 to 17 successes per PR) before this push restarted them.

## Verified

67 tests in the three queue and schedule files, then 516 across everything the stack touches, on the pinned toolchain. `mix format --check-formatted` clean, `okf validate decisions` passes, `decisions/index.md` unchanged. Both of your probes now pass; neither is in the suite, but the four tests above are.

BinaryBourbon (approved):
Re-reviewed the delta since my changes-requested (`fe83cca7` → `e4c7d7bd`). **Both blockers are fixed.** Approving. I re-ran the two probes from my last review and added two more to check the fixes did not move the arms that were already right.

## Blocker 1: the person's "Run now" — fixed, and the fix is better than what I suggested

I proposed folding the actor into `waiting?`. You split the question instead:

```elixir
waiting?      = ...                       # what the ROW says
queued_by_us? = actor in [@actor, SandboxQueue.actor()]   # whether this CALLER fired
```

That is the right decomposition and the comment says why in two sentences. The `cond` then has three honest arms instead of two overloaded ones, and the middle one — *"Somebody fired and was refused while the work waits. The attempt is stamped and recorded; the row still reports the wait, because that is what is true of the schedule"* — is a distinction I did not make and should have. The row describes the schedule; the trail describes the caller. Those were never the same fact.

**You also caught a bug I missed.** The old audit payload was `"outcome" => if(match?({:ok,_}, result), do: "ok", else: run_attrs.last_error)` — it read the outcome off the *row*. Gate the audit without touching that and the person's newly-restored event would have recorded `outcome: "waiting for a free sandbox slot"`, which is the one thing they did not get. `outcome/1` reads `result` instead. The test pins the divergence rather than the agreement:

```elixir
assert event.metadata["outcome"] =~ "sandbox quota"
assert reloaded.last_error == "waiting for a free sandbox slot"
```

Re-ran my probe, same scenario as before:

```
BEFORE  PROBE A: fired rows before=0 after=0   last_run_at=nil
AFTER   PROBE A: fired rows before=0 after=1   last_run_at=~U[2026-09-11 06:26:47Z]
                 outcome="sandbox quota: 2/2"  last_error="waiting for a free sandbox slot"
```

And the two arms that were already correct still are — I checked rather than assumed, because `queued_by_us?` is a new term in both of them:

```
PROBE C (cron, refused past the depth bound): fired=1 outcome="sandbox quota: 2/2" last_run_at=true
PROBE D (drainer replay meets the ceiling):   fired=0 last_error="waiting..." last_run_at=nil
```

The depth-bound refusal records, because the cron fired and got an answer; the replay stays silent, because its firing is already on the trail. Both right.

## Blocker 2: the row that said it was waiting forever — fixed

`note_queued_run_ended/3` on the two terminal transitions that never reach `run_schedule/2`, called from `expire_overdue/1` and `cancel_request/2`:

```
BEFORE  PROBE B: after expiry last_error="waiting for a free sandbox slot"
AFTER   PROBE B: after expiry last_error="timed out waiting for a free sandbox slot"
```

Four things I checked about the shape, all good:

- **`note_schedule_outcome/1` pattern-matches on `kind: "schedule_run"` with a binary `schedule_id`**, and `"a start request ending touches no schedule"` asserts a cancelled `start` leaves the schedule's own row alone. That negative test is the one that stops this becoming a function that guesses.
- **Neither call site is inside a transaction.** `drain/1` calls `expire_overdue/1` outside one and `swap/3` is a bare `update_all`, so the `Repo.update` and the `broadcast_schedules_changed/1` are not the audit-in-a-transaction trap ADR 0013 warns about.
- **Not audited, and the docstring says why**: `sandbox_request.expired` and `sandbox_request.cancelled` already record the event this mirrors, and the trail is not a second copy of tenant state.
- **No `last_run_at`.** Nothing ran. Correct, and stated.

`describe_queue_outcome/1` has a catch-all third clause, which is the same lesson as `"sandbox_request.#{status}"` in part 2 — a future terminal status degrades to a readable string instead of a `FunctionClauseError` raised inside a drain pass.

I also went looking for the hole this leaves: a `starting` row whose drain died is not expired by `expire_overdue/1`, so its schedule would keep saying "waiting". It closes on its own — the `*/5 * * * *` cron fans out over `user_ids_with_active_requests/0`, which includes `starting`, and `claim_next/2` reclaims a claim older than five minutes. So every path out of a live request now writes the row.

## Non-blocking: the exclusion belongs on the record, not only in the docstring

`note_queued_run_ended/3` is a public context function that writes tenant-owned state and deliberately does not audit. The reason is good and it is written down — but it is written in `schedules.ex`, and `audit_guardrail_test.exs` exists precisely so that *"the exclusion is a decision on the record rather than an omission"*. Its `@deliberately_silent` map is the record, and somebody auditing coverage in six months reads that map, not the context.

One line:

```elixir
"Schedules.note_queued_run_ended/3" =>
  "mirrors sandbox_request.expired / .cancelled onto the schedule row; the trail already has it",
```

The same applies, more weakly, to `run_schedule/2`'s new silent branch — though there the caller's action is on the trail as `sandbox_request.enqueued` in the same call, so it is genuinely covered.

## Verified locally

At the stack tip, on a database migrated to this tree: `sandbox_queue_test.exs`, `sandbox_queue_drain_test.exs`, `team/schedules_test.exs`, `sandbox_queue_controller_test.exs`, `audit_guardrail_test.exs`, `conversation_controller_test.exs`, `ops_gauges_test.exs`, `retention_pruner_test.exs`, `docs_test.exs`, `marketing_pricing_test.exs` — 302 tests, 0 failures. `mix credo --strict` clean, `mix format --check-formatted` clean, and `MIX_ENV=test mix compile --force --warnings-as-errors` clean, which I ran specifically because this PR makes `SandboxQueue` and `Team.Schedules` mutually dependent; both directions are fully-qualified runtime calls, so there is no compile cycle.

Still worth checking the real CI verdict before merging rather than the checks column — the runs on these heads were still queued when I looked.

PR #1873, opened 2026-09-11
feat(api): list, read and cancel a sandbox request (#1033)

jhgaylor:
Stacked on #1872.

Three routes, read and cancel only. Work enters the queue through `POST /api/conversations` with `queue: true` or through a schedule's own cron firing; nothing posts to `/api/sandbox-queue`.

- `GET /api/sandbox-queue` — the waiting list, in position order.
- `GET /api/sandbox-queue/:id` — where a 202 caller learns the outcome and the conversation id.
- `DELETE /api/sandbox-queue/:id` — gives up work that is still waiting. A request the drainer has already claimed is 404 rather than being cancelled out from under a start in flight.

Another tenant's request is 404 and never 403 — the same answer a request that does not exist gets. The id of a row you do not own is not something this API confirms.

The SDKs **omit** the three rather than claim them, with the reason recorded in `omissions.json`: every SDK's run handle promises an immediate conversation, and a queued start hands back a sandbox request first. Wrapping that properly means a queued-run handle that owns the polling and the cancellation, which is separate work; until then the raw HTTP client reaches them.

Seven gates for a new API route: operation declarations ✓, a `forbidden` response on each ✓, enums derived from `Request.kinds/0` and `Request.statuses/0` rather than duplicated (guard row added in part 4) ✓, `scripts/sdk-contract/build.sh --check` green with the omission claim ✓, `npm run generate` ✓, prose gates in part 7.

Generated artifacts are rebuilt, not hand-edited. **No SDK version bump here** — part 7 owns it. Label this `sdk-no-release`.

**Defers:** telemetry and retention (part 6), ADR 0042 and the docs (part 7).

Part 5 of 7 for #1033.

---

### Review fixes (round 2)

- `index`'s description no longer enumerates the terminal statuses, because a request the drainer has **claimed** is neither "already started" nor listed. It now says only waiting requests appear, and to read the rest by id.

Part of #1033

### DIFF
--- a/apps/fountain/lib/fountain_web/controllers/sandbox_queue_controller.ex
+++ b/apps/fountain/lib/fountain_web/controllers/sandbox_queue_controller.ex
@@ -1,0 +1,77 @@
+defmodule FountainWeb.SandboxQueueController do
+  @moduledoc false
+  use FountainWeb, :controller
+  use OpenApiSpex.ControllerSpecs
+
+  alias Fountain.SandboxQueue
+  alias FountainWeb.{Audited, Schemas}
+
+  action_fallback FountainWeb.FallbackController
+  plug OpenApiSpex.Plug.CastAndValidate, replace_params: false
+
+  tags(["Sandbox Queue"])
+
+  operation(:index,
+    summary: "List queued sandbox requests",
+    description:
+      "The caller's waiting requests, oldest first, each with its one-based position. " <>
+        "Only requests that are still waiting appear here. One the drainer has already " <>
+        "claimed is not listed, and neither is one that finished; read either by id " <>
+        "instead.",
+    responses: [
+      ok: {"Sandbox requests", "application/json", Schemas.SandboxRequestListResponse},
+      forbidden: {"Forbidden", "application/json", Schemas.Error}
+    ]
+  )
+
+  def index(conn, _params) do
+    render(conn, :index, requests: SandboxQueue.list_queued(conn.assigns.current_user.id))
+  end
+
+  operation(:show,
+    summary: "Get a sandbox request",
+    description:
+      "The request's current status. `conversation_id` is set once it started; `error` " <>
+        "says why if it failed. A request nobody else owns reads as 404.",
+    parameters: [id: [in: :path, type: :string, required: true]],
+    responses: [
+      ok: {"Sandbox request", "application/json", Schemas.SandboxRequestResponse},
+      not_found: {"Not found", "application/json", Schemas.Error},
+      forbidden: {"Forbidden", "application/json", Schemas.Error}
+    ]
+  )
+
+  def show(conn, %{"id" => id}) do
+    user = conn.assigns.current_user
+
+    case SandboxQueue.get_request(id, user.id) do
+      nil -> {:error, :not_found}
+      request -> render(conn, :show, request: request, position: SandboxQueue.position(request))
+    end
+  end
+
+  operation(:delete,
+    summary: "Cancel a queued sandbox request",
+    description:
+      "Gives up a request that is still waiting. A request the drainer has already " <>
+        "claimed reads as 404 rather than being cancelled out from under a start in " <>
+        "flight.",
+    parameters: [id: [in: :path, type: :string, required: true]],
+    responses: [
+      no_content: "Cancelled",
+      not_found: {"Not found or no longer queued", "application/json", Schemas.Error},
+      forbidden: {"Forbidden", "application/json", Schemas.Error}
+    ]
+  )
+
+  def delete(conn, %{"id" => id}) do
+    user = conn.assigns.current_user
+
+    with %{} = request <- SandboxQueue.get_request(id, user.id) || {:error, :not_found},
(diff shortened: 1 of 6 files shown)


BinaryBourbon (approved):
Part 5 of 7. The three routes. Approving.

Read-and-cancel-only is the right shape, and 404-not-403 for another tenant's id is the right answer — the id of a row you do not own is not something the API should confirm. `DELETE` going through part 1's compare-and-swap rather than a read-then-write is what keeps it from cancelling a start already in flight, and returning 404 for a claimed row (rather than 409) is defensible: from the caller's side the thing it asked to cancel is no longer cancellable, and it can read the outcome by id.

Seven gates all present — operation declarations, a `forbidden` response on each of the three, enums derived from `Request.kinds/0` / `statuses/0` with the guard rows from part 4, the `omissions.json` claim, and regenerated `contract.json` / `openapi.ts`. The omission reason is the honest one: every SDK's run handle promises an immediate conversation, and a queued start hands back a request first, so wrapping it properly is a queued-run handle and that is separate work. Better recorded than half-built.

One wording fix. `index`'s description says "A request that has already started, failed, expired or been cancelled is not listed; read it by id instead", but `list_queued/1` filters `status == "queued"` only — so a request the drainer has *claimed* (`starting`) also vanishes from the list while it is very much still live. A caller that polls `GET /api/sandbox-queue` right after its 202 can see an empty list with its request mid-replay and reasonably conclude the request is gone. Either say "waiting" instead of enumerating the terminal statuses, or include `starting` in the list (`position/1` already returns `nil` for it, and the JSON view already handles a null position).

Minor: `a start naming an explicit sandbox_id is never queued` has no test, while the images half does (`conversation_controller_test.exs` in part 4). Same two-line shape as the images case.

jhgaylor:
Wording fixed. Force-pushed.

`index`'s description no longer enumerates the terminal statuses, because that left out the case you spotted — a request the drainer has claimed is neither "already started" nor listed:

> Only requests that are still waiting appear here. One the drainer has already claimed is not listed, and neither is one that finished; read either by id instead.

I kept `list_queued/1` filtering `queued` alone rather than adding `starting`, since the JSON view derives position from list order (`Enum.with_index(1)`) while `position/1` returns `nil` for a claimed row — including it would need those two reconciled for no gain over reading by id.

**On the `sandbox_id` test I said was missing:** I tried to write it and could not do so honestly. `attach_conversation/4` takes no sandbox reservation, so that path cannot return `sandbox_quota_exceeded` or `:fleet_full` — there is no capacity error for the clause to intercept. My first attempt only "passed" by hitting a 422 `sandbox_identity_mismatch` from a sandbox built for a different agent, which tests nothing about queueing. So I withdrew the request and documented the clause in `maybe_enqueue/4` instead: belt and braces, kept so a future attach that *does* reserve cannot start queueing by accident. Details on #1872.

BinaryBourbon (approved):
Re-reviewed the delta since my approval (`053f8802` → `e90b871e`). Five lines, all of them the `index` operation's `description`. Approving.

The old wording listed the four terminal statuses and left `starting` out, which read as "everything except these four is here" when the truth is "only `queued` is here" — `queued_query/1` filters on that one status. A claimed request is neither listed nor terminal, and a caller polling this endpoint would have watched its request vanish for the seconds the drainer held it and concluded it was gone. The new sentence says it directly: *"One the drainer has already claimed is not listed, and neither is one that finished; read either by id instead."*

`scripts/sdk-contract/build.sh --check` is happy with the regenerated spec (167 operations, 193 schemas), so nothing downstream needs a bump for this.

Also worth noting from part 1's delta: `sandbox_key_id` is deliberately absent from `SandboxQueueJSON.data/2`. That is right — it is a restriction the server holds the replay to, not state the caller needs back.

`sandbox_queue_controller_test.exs` passes at the stack tip. CI has reported no verdict on this head; the 05:26 jobs are all `cancelled` by concurrency and `gh pr checks` prints that as `fail`.

jhgaylor:
CI caught my mistake on this one. `Release boot and the API contract` failed on parts 5, 6 and 7 — the three that carry the `/api/sandbox-queue` routes — with:

```
sdk/typescript/src/generated/openapi.ts is stale.
-  @description ... A request that has already started, failed, expired or been cancelled is not listed; read it by id instead.
+  @description ... Only requests that are still waiting appear here. One the drainer has already claimed is not listed, and neither is one that finished; read either by id instead.
```

The `index` description I rewrote is embedded in the generated TypeScript types, and I edited the operation without rebuilding them — exactly the thing this PR's own description says not to do ("Generated artifacts are rebuilt, not hand-edited"). Ran `scripts/sdk-contract/build.sh` then `npm run generate`, which changed one line, and rebased 6 and 7 so they carry it.

Worth recording *why* the wire-contract check stayed green while this failed, because the pair is confusing: `sdk/contract/contract.json` is the document projected down to shape alone with descriptions dropped, so it did not move at all — `SDK wire contract: ok (167 operations, 193 schemas)` both before and after. A prose-only edit to an operation is invisible to the contract and visible to the generated types. The two checks are not redundant.

Verified on the stack tip: `build.sh --check` ok, and `npm run generate` now leaves `src/generated/openapi.ts` unchanged.

PR #1874, opened 2026-09-11
feat(telemetry): what the sandbox queue looks like from outside (#1033)

jhgaylor:
Stacked on #1873.

Three series and a retention window. Parts 1 to 3 already emit these events; this declares the metrics over them, which is the direction `metrics_test.exs` checks ("every subscribed fountain event has a live producer").

- **`fountain.sandbox_queue.tenant_depth`** — a distribution observed after every queue mutation rather than polled, so a spike between two poller ticks is still in the histogram.
- **`fountain.sandbox_queue.completed.{count,wait_ms}`** — outcomes counted as they happen, tagged by status and kind, with the wait that preceded each one. The wait histogram is the number that says whether the queue is a wait or a black hole, and the evidence for revisiting the one-hour bound.
- **`fountain.sandbox_queue.requests.count`** — the `OpsGauges` poll, over **live rows only**. A `last_value` across terminal rows climbs forever and reports nothing about the queue now, which is why the Oban gauge excludes `completed` too. `ops_gauges_test.exs` asserts both halves: the live statuses are emitted with zeros, and the four terminal ones are not emitted at all.

`RetentionPruner` expires finished requests after 30 days. A finished row holds no prompt — every terminal transition erases `attrs` — so what is left is provenance and an outcome the audit trail already carries at its own, longer window. A `queued` or `starting` row is live work and is never pruned, whatever the window says; two tests pin that.

**Defers:** ADR 0042, the docs and the SDK version bump (part 7).

Part 6 of 7 for #1033.

---

### Review fixes (round 2)

- `emit_status_counts/3` takes a `where: r.status in ^statuses` — a no-op for the two existing callers, and the difference between a full-table group-by and an index read for this one, every ten seconds.
- The `RetentionPruner` docstring no longer claims a claimed row cannot outlive `SANDBOX_QUEUE_MAX_WAIT_SECONDS`. Nothing expires a `starting` row; the drain reclaiming an abandoned claim is what bounds it.

Part of #1033

### DIFF
--- a/apps/fountain/lib/fountain/ops_gauges.ex
+++ b/apps/fountain/lib/fountain/ops_gauges.ex
@@ -40,6 +40,17 @@ defmodule Fountain.OpsGauges do
         Fountain.Conversations.Sandbox.statuses()
       )
 
+      # Waiting and claimed only, for the same reason the Oban states above
+      # exclude `completed`: the terminal statuses are history rows, so a
+      # `last_value` over them only ever climbs and says nothing about queue
+      # health now. Outcomes are counted as they happen, on
+      # `[:fountain, :sandbox_queue, :completed]`.
+      emit_status_counts(
+        [:fountain, :sandbox_queue, :requests],
+        Fountain.SandboxQueue.Request,
+        Fountain.SandboxQueue.Request.active_statuses()
+      )
+
       emit_sandbox_provider_counts()
       emit_oban_depths()
     end)
@@ -76,9 +87,18 @@ defmodule Fountain.OpsGauges do
     end
   end
 
+  # Scoped to the statuses asked for, which is a no-op for a caller passing its
+  # schema's whole list and the difference between a full-table group-by and an
+  # index read for one passing a subset. This runs every ten seconds.
   defp emit_status_counts(event, schema, statuses) do
     counts =
-      Repo.all(from(r in schema, group_by: r.status, select: {r.status, count(r.id)}))
+      Repo.all(
+        from(r in schema,
+          where: r.status in ^statuses,
+          group_by: r.status,
+          select: {r.status, count(r.id)}
+        )
+      )
       |> Map.new()
 
     for status <- statuses do
--- a/apps/fountain/lib/fountain/workers/retention_pruner.ex
+++ b/apps/fountain/lib/fountain/workers/retention_pruner.ex
@@ -24,6 +24,15 @@ defmodule Fountain.Workers.RetentionPruner do
   attempt of a failing endpoint writes another. It is a debugging aid, not an
   event store, so it gets the shortest window of the lot.
 
+  `sandbox_requests` is the same shape of bookkeeping (ADR 0042): a finished
+  request holds no prompt, because every terminal transition erases `attrs`.
+  What is left is provenance and an outcome that the audit trail already
+  carries at its own, longer window. Waiting and claimed rows are never pruned
+  whatever the window says: the cutoff must not be what decides that a row is
+  live. A waiting row is bounded by `SANDBOX_QUEUE_MAX_WAIT_SECONDS`, and a
+  claimed one by the drain reclaiming an abandoned claim rather than by any
+  expiry, so neither is the pruner's to reason about.
+
   `usage_events` gets the longest window because it is the input to billing
   history, and `turn_images` is deliberately absent — those rows are owned by
   their turn and go when the conversation does.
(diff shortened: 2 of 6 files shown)


BinaryBourbon (approved):
Part 6 of 7. Telemetry and retention. Approving.

The three series are the right three, and the direction is right: depth observed after every mutation rather than polled means a spike between two poller ticks is still in the histogram, and the wait histogram is genuinely the number that decides whether the one-hour bound is defensible. Bucket choices look sensible against that bound (up to `3_600_000`, so the top bucket is "expired" rather than an overflow).

Excluding the terminal statuses from the `last_value` gauge is the detail I would have looked for — a `last_value` over history rows only ever climbs and says nothing about the queue now, which is the same reason the Oban gauge drops `completed`. `ops_gauges_test.exs` asserting both halves (live statuses emitted with zeros, the four terminal ones not emitted at all) is the right pair, because the zeros are what make an alert able to recover.

Two notes, neither blocking:

**The gauge query scans the whole table.** `emit_status_counts/3` groups over the schema with no `where`, then emits only the statuses in the list — so every 10 seconds this does a full `sandbox_requests` group-by to read two numbers, and with a 30-day retention the history dominates the table. Adding `where: r.status in ^statuses` to `emit_status_counts/3` is a no-op for the two existing callers (both pass their schema's complete status list) and lets this one use part 1's partial index. Small, and the poller is the one place where a cheap query matters.

**The `RetentionPruner` docstring overclaims slightly.** "Waiting and claimed rows are never pruned whatever the window says; they cannot outlive `SANDBOX_QUEUE_MAX_WAIT_SECONDS` anyway" — the first half is enforced by `do_prune(:sandbox_requests, _)`'s `status in ^terminal` and pinned by two tests, good. The second half is only true of `queued` rows: `expire_overdue/1` filters `status == "queued"`, so nothing expires a `starting` row, and one whose drain never runs again (cron off, or a tenant the fan-out stops finding) sits `starting` past the bound and is never pruned. In practice the five-minute cron recovers it inside ten minutes, so this is a docstring fix rather than a code one — say that recovery, not expiry, bounds a claimed row.

jhgaylor:
Both notes fixed. Force-pushed.

**The gauge no longer scans the whole table.** `emit_status_counts/3` takes a `where: r.status in ^statuses`, which is a no-op for the two existing callers (each passes its schema's complete status list) and lets this one use part 1's partial index instead of grouping 30 days of history every ten seconds to read two numbers.

**The pruner docstring no longer overclaims.** You were right that "they cannot outlive `SANDBOX_QUEUE_MAX_WAIT_SECONDS`" is only true of `queued` rows — `expire_overdue/1` filters on that status, so nothing expires a claimed one. It now says what actually bounds each:

> A waiting row is bounded by `SANDBOX_QUEUE_MAX_WAIT_SECONDS`, and a claimed one by the drain reclaiming an abandoned claim rather than by any expiry, so neither is the pruner's to reason about.

`ops_gauges_test.exs` and the two pruner tests pass unchanged — neither change alters which events are emitted or which rows are deleted.

BinaryBourbon (approved):
Re-reviewed the delta since my approval (`25d61712` → `8bb25fa4`). Two changes, both taking notes I left. Approving.

## `emit_status_counts/3` is now scoped to the statuses it reports

```elixir
where: r.status in ^statuses,
```

A no-op for a caller passing its schema's whole status list, and the difference between a full-table group-by and an index read for one passing a subset — which `sandbox_requests` does, reporting `Request.active_statuses/0` while the table also holds every terminal row until the pruner gets to it. This runs every ten seconds, and a queue table's history is the part that grows without bound between prunes. The comment gives that reasoning at the call site, which is where somebody will otherwise wonder why the filter is there.

No behaviour change: the `for status <- statuses` loop below only ever emitted the statuses it was given, so the rows the old query fetched and discarded were pure cost.

And with part 1's `@active_statuses Request.active_statuses()`, the gauge and the depth bound now read one list rather than two copies of `~w(queued starting)`. That was the actual risk — a status added to the schema and to one of the two copies is a silently wrong depth bound, not a visibly wrong gauge.

## The `RetentionPruner` docstring

The old sentence justified never pruning live rows with *"they cannot outlive `SANDBOX_QUEUE_MAX_WAIT_SECONDS` anyway"*, which is only true while some drain keeps running for that tenant — and a `starting` row is bounded by the drain reclaiming an abandoned claim, not by any expiry at all. The rewrite leads with the rule instead: *"the cutoff must not be what decides that a row is live"*, then names what actually bounds each of the two. That is the version that survives somebody changing the wait bound.

## Verified locally

`ops_gauges_test.exs` and `workers/retention_pruner_test.exs` pass at the stack tip, along with the rest of the stack's files (298 tests across the two batches, 0 failures) on a database migrated to this tree. `mix credo --strict` clean.

CI has reported no verdict on this head — every job on the 05:26 run is `cancelled` by concurrency, which `gh pr checks` renders as `fail`. Worth a real look before merging.

PR #1877, opened 2026-09-11
feat(oauth): a tenant registers, renames and removes its own clients (#1125)

jhgaylor:
Part 2 of 9 for #1125 (re-cut of #1489). Based on #1875.

The second registry, scoped by `user_id` like every other tenant-owned thing: `list_clients/1`, `get_client_record/2`, `create_client/3`, `update_client/3` and `delete_client/2` on `Fountain.OAuth`. Each records its own event, so every door is covered by construction (ADR 0013); `audit_guardrail_test.exs` carries the three new rows.

Two refusals are worth reading twice.

- **Twenty-five clients an account.** An abuse ceiling rather than an allowance, because a row will shortly widen the deployment's CORS allowlist and registration is self-serve.
- **A published client can be changed or deleted by nobody but an operator.** Publication moves the trust boundary from the owner to every account, and the row's `client_id` is random, so a deletion would break sign-in for all of them with an id nobody can recreate.

`client_id`, the owner and `published` are dropped from update attributes rather than rejected, so a surface that hands back what it read does not fail for sending a field it was given.

The authorization flow still reads only the config registry. Reading these rows into it is the next PR.

Part of #1125

### DIFF
--- a/apps/fountain/lib/fountain/oauth.ex
+++ b/apps/fountain/lib/fountain/oauth.ex
@@ -22,10 +22,17 @@ defmodule Fountain.OAuth do
 
   ## Clients
 
-  A registry in application config, not a table: there are two of them and
-  they are ours. `config :fountain, Fountain.OAuth, clients: [%{id, name,
-  redirect_uris}]` (runtime.exs reads `OAUTH_CLIENTS` as JSON). Redirect
-  URIs match **exactly**.
+  Two registries. Application config holds the operator's — `config :fountain,
+  Fountain.OAuth, clients: [%{id, name, redirect_uris}]`, which runtime.exs
+  reads from `OAUTH_CLIENTS` as JSON — and the `oauth_clients` table holds
+  the ones a tenant registers for itself (#1125). Redirect URIs match
+  **exactly**.
+
+  The client functions at the bottom of this module own the second registry:
+  a tenant's own rows, scoped by `user_id`, audited, and capped. A row starts
+  unpublished, and the authorization flow does not read the table yet. The
+  next change makes it, under the development-mode rule that is what makes a
+  self-chosen redirect URI safe.
 
   ## Tokens are API keys
 
@@ -37,9 +44,17 @@ defmodule Fountain.OAuth do
   the host, not the library's.
   """
 
+  import Ecto.Query, warn: false
+
   use Managoat.OAuth, otp_app: :fountain, host: Fountain.OAuth.Host
 
-  alias Fountain.Accounts
+  alias Fountain.{Accounts, Audit, Repo}
+  alias Fountain.OAuth.Client
+
+  # An abuse ceiling, not an allowance: every row here will widen the
+  # deployment's CORS allowlist (ADR 0021), so registration must not be
+  # unbounded.
+  @max_clients_per_account 25
 
   @doc """
   Revoke the token (an API key) presented by an app that is signing out.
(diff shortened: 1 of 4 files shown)


BinaryBourbon (approved):
The context half. Tenant scoping is right everywhere, the audit calls sit in the context rather than the callers (so the console in #1883 and the API in #1882 are covered by construction), and all three actions are on the `audit_guardrail_test.exs` list — which is the check that actually stops this rotting.

Things I checked and liked:

- **`get_client_record/2` guards the UUID cast.** The comment names the real failure — a mistyped id raising through phoenix_ecto into a 400 rather than the documented 404 — and `_test.exs` pins `"foo"` and `""`. That is the sort of thing normally found in production.
- **`update_client/3`'s `Map.drop` is belt-and-braces.** The changeset's cast allowlist already refuses `published` and `user_id`; the drop is redundant but harmless, and the atom-keyed test proves the *cast* is what is actually holding, not the drop. Good that the test attacks the non-dropped path.
- **`prevent_published_update/2` and the delete refusal.** Refusing *deletion* of a published client, not just editing, is the half people forget. The reason given (random `client_id`, nobody can recreate it, every account's sign-in breaks) is the correct one.
- **Rejected writes leave no trail**, tested both ways. Matches the audit rule in CLAUDE.md.

Three nits:

**`list_clients/1` has no tie-break.** `order_by: [desc: c.inserted_at]` over `timestamps(type: :utc_datetime)` — second precision — so two clients registered in the same second come back in whatever order Postgres feels like, and the console list in #1883 can reorder between renders. `order_by: [desc: c.inserted_at, desc: c.id]` costs nothing.

**The cap is check-then-insert.** Two concurrent `POST`s both read 24 and both insert. It is an abuse ceiling rather than an allowance, so I would not add a lock for it — but it is worth knowing that 25 is a soft 25.

**`@max_clients_per_account 25` is a module attribute, not config.** The ADR amendment in #1885 says explicitly that an operator raises it by editing the constant and that nobody has asked — so this is a decision on the record rather than an oversight, and I am not asking for it to change. Flagging only because every other ceiling in this codebase (`TEAM_CONTACT_CEILING`, `SANDBOX_FLEET_CEILING`) reads from the environment, so a self-hoster will look for `OAUTH_CLIENT_CEILING` first.

One question, not for this PR: there is no operator switch that turns self-registration off. A deployment that deliberately left `OAUTH_CLIENTS` and `API_CORS_ORIGINS` empty now gets working OAuth clients, and CORS origins, from any verified account. I think #1879's argument is sound and this admits nobody who does not already hold a bearer key — but "any verified user may register" is a product decision with no knob, and the self-host story is a thing this repo cares about. Worth a line in the ADR's "still not done" if the answer is "deliberately, ask later".

Approving.

jhgaylor:

- **`list_clients/1` breaks the tie on `id`.** Second-precision timestamps meant two clients registered in the same second came back in the planner's order and the console list could reorder itself between renders. Test asserts the exact documented ordering rather than a sort that would be flaky across a second boundary.
- **An update that moves nothing records nothing.** This is the audit finding raised on #1882, fixed at its source: `update_client/3` returns `{:ok, client}` before the audit when the changeset is valid and empty. `Repo.update/1` issued no statement in that case anyway, so the only thing the old path produced was an `oauth_client.updated` row with `"changed" => []` — the shape ADR 0013 rules out. Two tests, one for a PATCH that repeats what it read and one for `%{}`.

Both verified by reverting the guard: the tests fail without it.

Not changed, as discussed in review: the check-then-insert race on the 25-client cap (it is an abuse ceiling, a lock buys nothing) and the ceiling staying a module attribute (the ADR says so on the record). The missing operator kill-switch is now written down in ADR 0021's "Still not done" — see #1885.

BinaryBourbon (approved):
Both nits fixed here, and this PR absorbed the audit finding I raised on #1882 — which is where it belonged, since the console in #1883 takes the same path.

## The no-op update

`if changeset.valid? and changeset.changes == %{}` is the right predicate, and the `valid?` half is what makes it safe: a published client's refusal is an *invalid* changeset with no changes, so it still falls to `Repo.update/1` and still returns `{:error, _}`. The "a refused update leaves no trail" test proves that did not move.

I checked the paths that reach an empty change set, because the fix is only correct if `changes` is genuinely empty in every "moved nothing" case rather than merely in the `%{}` one:

| Request | Why `changes` is empty |
|---|---|
| `%{}` | nothing cast |
| `%{"name" => same}` | `cast/3` drops a value equal to the data |
| `%{"name" => " same "}` | `update_change/3` trims, then `put_change/3` drops it for equalling the data |
| `%{"redirect_uris" => same_list}` | `validate_redirect_uris/1`'s `put_change` drops it; `put_origin_keys/1` likewise |

Both new tests are on the two ends of that table. The middle two rows are not tested and I would not add them — they are Ecto's behaviour, not this function's.

The self-healing property survives, which is the thing I would have worried about: a row whose stored `origin_keys` drifted, or whose `redirect_uris` still holds pre-#1875 duplicates, produces a *different* value from `put_origin_keys` / the uniq, so it is not a no-op and the write lands. A PATCH repeating what it read is still the repair path.

The docstring gained "A call that moves no field succeeds and records nothing" as its second line, which is the sentence an API caller needs, and the inline comment names ADR 0013 rather than restating it. Good.

## The tie-break

`order_by: [desc: c.inserted_at, desc: c.id]` — and the comment says why, which is the half that stops someone tidying it away.

The test is stronger than it looks at first read. It builds `expected` by sorting `list_clients/1`'s own output, so it is asserting "the output is already in that order" rather than comparing two independent orderings. Drop the `desc: c.id` and five rows sharing a second come back in heap order while `expected` is id-desc, so it fails roughly 119 times in 120. Worth knowing it is probabilistic rather than deterministic; at five rows that is close enough to a guard that I would leave it.

Elixir sorting the UUID *strings* and Postgres sorting the `uuid` *values* agree here, for what it is worth — the canonical text form is fixed-width lowercase hex with dashes at fixed offsets, so lexicographic order over the text matches byte order over the value. Not worth a comment.

Approving.

PR #1878, opened 2026-09-11
feat(oauth): a registered client signs in its owner and nobody else (#1125)

jhgaylor:
Part 3 of 9 for #1125 (re-cut of #1489). Based on #1877.

The authorization flow reads both registries now, and the second one is safe because of this rule rather than because of its redirect allowlist.

**The wrong answer, on the record.** Wildcard the sandbox domain in the redirect allowlist so an app built inside a sprite can point at a production Fountain. That is a phishing kit: anyone with any sandbox could start a flow with their own PKCE challenge and their own box as `redirect_uri`, and a consenting user's full-scope key would land there. PKCE does not help when the attacker initiates the flow.

**What is built instead.** An unpublished client authorizes only its owner. Every other account is rendered an error page, never redirected, and the identity check runs *before* the redirect check so a stranger's error page discloses no registration. Because the only account such a client can capture belongs to the person who registered it, its owner may name any HTTPS redirect or an HTTP loopback one. Config wins on an id collision, so a row cannot shadow a first-party client.

A loopback redirect matches on any port for an unpublished client (RFC 8252 §7.3) and not at all for a published one, because the port a local dev server lands on is not a fact anybody registered. `authorize/3` hands the library the validated requested URI so its own exact match reaches the same conclusion.

`oauth.authorized` now carries the client's row id when there is one.

Left to the next two PRs: CORS still reads only `API_CORS_ORIGINS`, and the consent page's `form-action` header still names every configured client.

Part of #1125

### DIFF
--- a/apps/fountain/lib/fountain/oauth.ex
+++ b/apps/fountain/lib/fountain/oauth.ex
@@ -25,14 +25,27 @@ defmodule Fountain.OAuth do
   Two registries. Application config holds the operator's — `config :fountain,
   Fountain.OAuth, clients: [%{id, name, redirect_uris}]`, which runtime.exs
   reads from `OAUTH_CLIENTS` as JSON — and the `oauth_clients` table holds
-  the ones a tenant registers for itself (#1125). Redirect URIs match
-  **exactly**.
+  the ones a tenant registers for itself (#1125). Config wins on a collision,
+  so a row can never shadow a first-party client.
 
-  The client functions at the bottom of this module own the second registry:
-  a tenant's own rows, scoped by `user_id`, audited, and capped. A row starts
-  unpublished, and the authorization flow does not read the table yet. The
-  next change makes it, under the development-mode rule that is what makes a
-  self-chosen redirect URI safe.
+  **Development mode is the security boundary, not the redirect allowlist.**
+  A row starts unpublished, and an unpublished client authorizes only its
+  owner: every other account is rendered an error page, never redirected.
+  That is what makes a self-chosen redirect URI safe — the only account such
+  a client can capture belongs to the person who registered it — and it is
+  why its owner may name any HTTPS redirect or an HTTP loopback one.
+  `published` is an operator flip with no self-serve path, and a published
+  client is an ordinary first-party client on config's terms.
+
+  Redirect URIs match **exactly**, except that an unpublished loopback
+  redirect matches on any port (RFC 8252 §7.3), because the port a local dev
+  server lands on is not a fact anybody registered.
+
+  `redirect_registered?/2` is the **only** redirect gate in the server.
+  `grant_client/2` hands the library a client whose registered list is the
+  requested URI, so `Managoat.OAuth.Clients.validate_request/2` always agrees
+  and is no longer a second opinion. Loosening the check here loosens
+  everything; there is nothing behind it.
 
   ## Tokens are API keys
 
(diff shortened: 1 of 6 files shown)


BinaryBourbon (approved):
This is the one that matters, and it holds up. I went looking for a way around the owner gate rather than reading it sympathetically; I did not find one.

**Paths to an authorization code, all of them:**

| Path | Gated? |
|---|---|
| `GET /oauth/authorize` | `validate_request(params, current_user.id)` ✅ |
| `POST /oauth/authorize` (decision) | re-validates rather than trusting the `GET` ✅, and there is a test that POSTs straight to the decision as a stranger |
| `OAuth.authorize/3` | re-runs `validate_request/2` itself, so a future caller that skips the controller is still gated ✅ |
| Device grant | `start_device_grant/0` takes no `client_id`, so the registry is not reachable from it ✅ |
| `Managoat.OAuth.authorize/4` | one call site, inside the override ✅ |

**The ordering is the good part.** `authorizable_by?/2` before `redirect_registered?/2` means a stranger cannot use the error page as an oracle for someone else's registered URIs, and the test asserts `:development_mode` for a stranger asking with a *wrong* redirect — which is the assertion that would fail if someone later "tidied" the `with` into redirect-first. Keep that test's comment.

**`exchange/1` is safe with a DB client.** I checked `Managoat.OAuth.Codes.exchange/3`: it binds against the stored row's `client_id` and `redirect_uri` and never touches the registry, so the any-port loopback code is bound to the port that actually asked for it. The "a loopback code is bound to the port that asked for it" test is proving something real.

**`loopback_match?/2` compares everything but the port** — scheme, downcased host, userinfo, path, query, fragment — and refuses to apply at all unless *both* sides are loopback. `published` short-circuits to exact-only before it is ever reached. The four-way negative test (other path / other host / other scheme / a fragment) covers the shapes that matter.

One thing to keep visible rather than change: **`grant_client/2` hands the library a config whose only registered URI is the requested one**, which makes `Clients.validate_request/2`'s exact-match a tautology. That is correct given Fountain has already validated, and the comment says so — but it does mean `redirect_registered?/2` is now the single redirect gate in the system, with the library's check no longer a backstop. If that function ever gets refactored, nothing else catches the mistake. Worth one sentence in the moduledoc saying so out loud.

Nit: the consent page's development-mode banner uses raw `bg-amber-50 text-amber-700` while the rest of that template is on `var(--color-*)` tokens. Cosmetic, and #1883 does the same thing, so it is at least consistent with itself.

Approving.

jhgaylor:
Review feedback applied; force-pushed (amended commit). No behaviour change — the point raised was that the *reason* the design is safe is invisible at the one place a future refactor would touch.

The moduledoc now says outright that `redirect_registered?/2` is the only redirect gate in the server, and `grant_client/2` carries a comment saying its own construction is what makes that true (the library's exact match cannot fail once it is handed the requested URI, so it is not a backstop). Loosening the check here loosens everything; there is nothing behind it.

BinaryBourbon (approved):
Documentation only on this push, and it is the sentence I asked for.

> `redirect_registered?/2` is the **only** redirect gate in the server. `grant_client/2` hands the library a client whose registered list is the requested URI, so `Managoat.OAuth.Clients.validate_request/2` always agrees and is no longer a second opinion. Loosening the check here loosens everything; there is nothing behind it.

That is in the moduledoc, where a reader arrives, and the matching comment at `grant_client/2` adds the operational half — "never call this with a URI `validate_request/2` has not already accepted" — which is the instruction a future caller needs rather than a restatement of the observation. Two places, two different jobs, neither redundant.

Nothing else moved; the ordering guarantee and its test are untouched. The amber-chip note stays cosmetic and stays a stack-wide choice.

Approving.

PR #1879, opened 2026-09-11
feat(oauth): CORS admits a registered client's redirect origins (#1125)

jhgaylor:
Part 4 of 9 for #1125 (re-cut of #1489). Based on #1878.

One registration now covers both halves of "an app on another origin": the sign-in and the API calls that follow it. Without this, registering a client would leave the app able to complete OAuth and then fail its first `fetch`, and the operator would still be editing `API_CORS_ORIGINS`.

A preflight carries no authentication, so the origin is the only thing the plug can key on — which is exactly why "an origin some client registered" is a sound predicate. Admitting an origin admits nobody who does not already hold a bearer key, and `Access-Control-Allow-Credentials` is still never sent, so no cookie crosses either way.

`API_CORS_ORIGINS` is still checked first and short-circuits, so the lookup only runs for an origin the operator did not list, and it is one containment against the GIN index on `origin_keys`. An unlisted origin therefore costs one indexed read per request; the `:api` pipeline's pre-auth IP rate limit is what bounds that. Loopback matches on any port here too: the sign-in and the first API call must not be able to disagree about which port counts.

Left to the next PR: the consent page's `form-action` header.

Part of #1125

### DIFF
--- a/apps/fountain/lib/fountain/oauth.ex
+++ b/apps/fountain/lib/fountain/oauth.ex
@@ -316,6 +316,35 @@ defmodule Fountain.OAuth do
     end
   end
 
+  @doc """
+  Whether an origin belongs to an operator or a tenant client.
+
+  CORS still requires a bearer key. This predicate grants no account access;
+  it only lets a browser present a key it already holds from an origin some
+  client registered.
+  """
+  @spec registered_origin?(term()) :: boolean()
+  def registered_origin?(origin) when is_binary(origin) do
+    case Client.origin_key(origin) do
+      nil -> false
+      key -> key in config_origin_keys() or Repo.exists?(origin_key_query(key))
+    end
+  end
+
+  def registered_origin?(_), do: false
+
+  defp origin_key_query(key) do
+    from c in Client, where: fragment("? @> ?", c.origin_keys, ^[key])
+  end
+
+  defp config_origin_keys do
+    config_clients()
+    |> Enum.flat_map(& &1.redirect_uris)
+    |> Client.origins_of()
+    |> Enum.map(&Client.origin_key/1)
+    |> Enum.reject(&is_nil/1)
+  end
+
   defp client_count(user_id) do
     Repo.aggregate(from(c in Client, where: c.user_id == ^user_id), :count)
   end
--- a/apps/fountain/lib/fountain_web/plugs/cors.ex
+++ b/apps/fountain/lib/fountain_web/plugs/cors.ex
@@ -3,14 +3,34 @@ defmodule FountainWeb.Plugs.Cors do
   CORS for `/api/*`, for a browser client on another origin — the standalone
   team app (#810) is the first.
 
-  Off by default: with no configured origins the plug does nothing, and the
-  API stays same-origin + non-browser clients only, as it always was. Set
-  `API_CORS_ORIGINS` (comma-separated, exact origins such as
-  `https://team.example.com`, or `*`) to allow browsers there to call the API
-  with a bearer token. No cookies are ever allowed across origins
+  Two sources, checked in order. `API_CORS_ORIGINS` is the operator's
+  (comma-separated, exact origins such as `https://team.example.com`, or
+  `*`). A registered OAuth client's redirect origin is the second, so
+  registering an app is enough to call the API from it and no operator has to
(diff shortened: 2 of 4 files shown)


BinaryBourbon (approved):
The security argument is correct and I want to say so explicitly, because "any tenant can add an origin to the deployment's CORS allowlist" reads alarming until you check the one thing it depends on: `put_cors_headers/2` sets no `Access-Control-Allow-Credentials`, and nothing in this PR adds one. So an admitted origin gets to *read a response it already had the bearer key to fetch*. No cookie crosses, and a `credentials: 'include'` fetch is blocked by the browser for want of the header. The moduledoc makes that argument in the right place. Keep the "never sent" sentence adjacent to `put_cors_headers/2` forever.

The loopback any-port key earning its keep here is the real design win — sign-in and the first `/api` call agreeing is exactly the half-working failure the `origin_key` column exists to prevent, and the test asking on `:5174` for a client registered on `:5173` pins it.

**One thing to consider before merge, not a blocker.**

`FountainWeb.Plugs.Cors` is plugged at `endpoint.ex:116` — *before* `Plug.Session` and before the router, so before authentication and before both `RateLimit` passes in the `:api` pipeline. `registered_origin?/1` now puts a database round trip on that path: every `/api/*` request carrying an `Origin` header that is not in `API_CORS_ORIGINS` does one `@>` lookup, including preflights, including requests from origins nobody ever registered. The config check short-circuits first, which helps deployments that set `API_CORS_ORIGINS` — but the deployments this feature is *for* are precisely the ones that leave it empty, so they take the query every time.

The index makes each query cheap; my concern is the shape, not the cost of one. This is an unauthenticated, un-rate-limited, attacker-triggerable DB query, on a repo that has already been bitten once by unscoped query volume plus a small pool (the 2026-09-04 broker OOM was demo apps at 14 rps against an unscoped aggregate). A small ETS cache keyed on the origin key with a short TTL, invalidated on client write, would take it off the hot path entirely; even a negative cache would, since the common case is a miss.

If the answer is "measure it first, the index is fine" that is a reasonable answer — I would just rather it be written down than discovered.

Two smaller notes:

- `config_origin_keys/0` recomputes `Config.load!` → normalize → flat_map → origins → keys on every call. Trivial next to the query, but it is in the same hot path and it is pure config.
- The plug is no longer "off by default": the old `allowed?(_, [])` returned false without so much as a lookup. The moduledoc rewrite says this, which is the important part.

Approving.

jhgaylor:

The answer to the hot-path question is "measure first", and that is now written into the plug's moduledoc rather than left to be rediscovered: this plug runs ahead of the session, the router and both rate-limit passes, so an unauthenticated request with an unrecognised `Origin` reaches the database once per request, preflights included. The moduledoc records why that is accepted for now — `API_CORS_ORIGINS` short-circuits, `origin_keys` is GIN-indexed — and that if the measurement ever says "cache", the invalidation on every client write and delete is the design, not the cache.

No code change. The `Allow-Credentials`-is-never-sent argument the whole predicate rests on stays adjacent to `put_cors_headers/2`, as it should.

BinaryBourbon (approved):
Moduledoc only, and the answer is the reasonable one I said I would accept — written down rather than discovered.

> That is the accepted cost for now, not an oversight. `API_CORS_ORIGINS` is checked first and short-circuits; `origin_keys` is GIN-indexed, so the miss that the common case is costs one indexed containment; and a cache would have to be invalidated on every client write and every deletion, which is the part that goes wrong quietly. Measure before adding one — and if the measurement says cache, the invalidation is the design, not the cache.

Three things I like about how it is written. It states the position of the plug in the stack explicitly ("ahead of the session, the router and both of the `:api` pipeline's rate-limit passes"), so the next reader does not have to go and count plugs in `endpoint.ex` to see why it matters. It names the miss as the common case rather than the exception, which is the fact that makes the index argument work. And the last clause pre-empts the shortcut version of the fix: someone adding an ETS cache without an invalidation story now has to do it against a sentence telling them not to.

`config_origin_keys/0` still recomputes `Config.load!` on every call — I flagged it as trivial next to the query and it still is, so no objection to leaving it. It just means the "config is checked first and short-circuits" sentence is doing slightly more work than the code.

Approving.

PR #1881, opened 2026-09-11
fix(docs): remove a stray conflict marker from the changelog

jhgaylor:
`CHANGELOG.md` line 67 on `main` contains a leftover `>>>>>>> theirs` conflict marker from a botched resolution.

This is not just repo furniture. `Fountain.Docs` declares the repo-root changelog on its `extra_resources:` line (`apps/fountain/lib/fountain/docs.ex:48`) and `docs/changelog.md` pulls it in as a snippet, with `Changelog: changelog.md` in `docs/nav.yml`. So the marker is embedded at compile time and served to readers at `/docs/changelog`.

The paragraph above it (the `Agent.model` client note) is intact and the bullet below it is the next entry, so the marker is vestigial — the line just goes. No other markers in the file.

Found while rebasing the #1635 stack; unrelated to it, so it lands on its own.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,7 +64,6 @@ upgrade, is in
   `AgentUpdate["model"]` are `string | null` and optional. A client that
   assumed a string needs a null check. Nothing else on the wire changed
   shape.
->>>>>>> theirs
 
 - Environment `setup_timeout_seconds` (1–900, default 120) lets cold repository
   toolchain setup run within an explicit bound. It persists through API/spec


BinaryBourbon (approved):
Verified the whole claim chain, in a worktree off `pr-1881`. Approving.

## The marker is vestigial

`git log -S'>>>>>>> theirs' -- CHANGELOG.md` traces it to `9da64a10` (the #1839 stack tip), where the entire surrounding block — the `runtime_command` entry, the free-string rationale, the `Agent.model` client note — arrives in a single hunk with the marker as its last line. The resolver kept *theirs* wholesale and dropped the `<<<<<<<` / `=======` pair but not the tail, so no *ours* content was lost. The paragraph above the marker is complete and the bullet below it is the next entry, exactly as the PR body says.

It is also the only one: all three marker forms across `main`'s `CHANGELOG.md` match line 67 and nothing else; on this branch they match nothing.

## It is live on the public manual

The embedding claim holds — `extra_resources: ["CHANGELOG.md"]` in `apps/fountain/lib/fountain/docs.ex:48`, `docs/changelog.md` is a bare `--8<-- "CHANGELOG.md"`, and `docs/nav.yml:150` mounts it at `Changelog: changelog.md`.

`>>>>>>>` is seven nested blockquotes in markdown, so it renders as an indented box holding the word "theirs" in the middle of the unreleased notes. Production right now:

```
$ curl -s https://managoat.com/docs/changelog | grep -B2 -A2 theirs
<blockquote><blockquote><p>theirs</p></blockquote></blockquote>
```

## CI

All six partitions green, plus static analysis, the release/contract job and the SDK clients. `mergeable: MERGEABLE`, blocked only on `REVIEW_REQUIRED`.

## One gap this PR correctly leaves alone

No conflict-marker guard exists anywhere in the tree — a grep for the marker patterns across `.ex`, `.exs`, `.sh`, `.yml`, `.py` and `.mjs` finds no check.

`CHANGELOG.md` also sits in a seam. The `changes` job keeps it off the docs-only allowlist on purpose (`ci.yml:115`, right call — the release-bump flow must not skip the suite), but that also leaves `docs_touched` false for a changelog-only diff, so **Docs** and **Docs prose gates** both skipped on this PR. The prose linters cover `docs/` and not the repo root, so the one file that is both a published docs page and prose-gate-exempt is where this landed.

The fix for the class is a tree-wide marker check rather than a changelog one, so it belongs in its own PR. Filed separately.

jhgaylor:
Follow-up for the class gap noted in the review: #1898.

PR #1882, opened 2026-09-11
feat(api): register and manage OAuth clients over /api (#1125)

jhgaylor:
Part 6 of 9 for #1125 (re-cut of #1489). Based on #1880.

`GET/POST /api/oauth/clients` and `GET/PATCH/DELETE /api/oauth/clients/:id`, with the OpenAPI operations, the three schemas and the regenerated wire contract.

**Full scope, not the sprite scope #1125 originally proposed.** A registered client is a standing way to obtain a full-scope thirty-day key with one consent, which is exactly the escalation the sprite scope exists to prevent: a sandbox's per-conversation token must not be able to leave one behind. Registering from the console, the CLI or the API with the owner's own key still removes the operator, which was the point.

Two details worth naming:

- Another account's client answers 404 rather than 403, like every other tenant-scoped resource, and **so does an id that is not a UUID** — casting a path segment to `:binary_id` raises, and phoenix_ecto turns that into a 400, so the documented 404 would never reach anybody who mistyped an id.
- `origins` is derived for the response rather than being the stored lookup key, so a caller sees the origins it will actually be calling `/api` from.

**Fixed against #1489:** the original declared these routes in neither `sdk/contract/manifests/*` nor `sdk/contract/omissions.json`, so `scripts/sdk-contract/build.sh --check` would have failed. They are recorded as an omission here: registering is a once-per-app setup step a person does in the console or with `fountain oauth-client`, and an SDK caller's work starts after it, with the key the flow mints.

`sdk-no-release`: `src/generated/openapi.ts` is regenerated here with no version bump. The bump is in #1885, the last PR of the stack.

**On size.** This is the largest PR in the stack at ~1,190 lines, and 747 of those are machine-written: `sdk/contract/contract.json` and `src/generated/openapi.ts` are regenerated by `scripts/sdk-contract/build.sh` and `npm run generate`, not edited. The hand-written half is ~440 lines across the controller, the view, the router and the schemas.

Left to the next PRs: the console page, the CLI, and the docs.

Part of #1125

### DIFF
--- a/apps/fountain/lib/fountain_web/controllers/oauth_client_controller.ex
+++ b/apps/fountain/lib/fountain_web/controllers/oauth_client_controller.ex
@@ -1,0 +1,150 @@
+defmodule FountainWeb.OAuthClientController do
+  @moduledoc """
+  The OAuth clients an account registered for itself (#1125).
+
+      GET    /api/oauth/clients      — the account's clients
+      POST   /api/oauth/clients      — register one
+      GET    /api/oauth/clients/:id  — one of them
+      PATCH  /api/oauth/clients/:id  — rename it, or change its redirect URIs
+      DELETE /api/oauth/clients/:id  — unregister it
+
+  Full scope, for the reason API key management is full scope: a client is a
+  standing way to obtain a full-scope key with one consent, so a sandbox's
+  per-conversation token must not be able to leave one behind (see
+  `Fountain.Accounts.ApiKey`).
+
+  Registering here is enough to *use* the client: it can sign in its owner at
+  `/oauth/authorize`, and its redirect origins are admitted by CORS, so no
+  operator has to touch `OAUTH_CLIENTS` or `API_CORS_ORIGINS`.
+  """
+
+  use FountainWeb, :controller
+  use OpenApiSpex.ControllerSpecs
+
+  alias Fountain.OAuth
+  alias FountainWeb.Audited
+  alias FountainWeb.Schemas
+
+  action_fallback FountainWeb.FallbackController
+
+  tags(["OAuth clients"])
+
+  operation(:index,
+    summary: "List OAuth clients",
+    description:
+      "The OAuth clients this account registered. Each one is in development " <>
+        "mode until an operator publishes it, which means it signs in its " <>
+        "owner and refuses every other account.",
+    responses: [
+      ok: {"Clients", "application/json", Schemas.OAuthClientListResponse},
+      unauthorized: {"Missing or invalid key", "application/json", Schemas.Error},
+      forbidden: {"Insufficient scope", "application/json", Schemas.Error}
+    ]
+  )
+
+  def index(conn, _params) do
+    user = conn.assigns.current_user
+    render(conn, :index, clients: OAuth.list_clients(user.id))
+  end
+
+  operation(:create,
+    summary: "Register an OAuth client",
+    description:
+      "Register an app that can offer \"Sign in with Fountain\". The response " <>
+        "carries the generated `client_id` to put in the app. Redirect URIs " <>
+        "match exactly, must be https unless they are loopback, and a loopback " <>
+        "URI matches on any port. The client starts in development mode, so it " <>
+        "signs in nobody but you. A sandbox's public HTTPS URL is valid.",
+    request_body: {"Client", "application/json", Schemas.OAuthClientRequest, required: true},
+    responses: [
+      created: {"Client", "application/json", Schemas.OAuthClient},
+      unprocessable_entity: {"Validation error", "application/json", Schemas.Error},
+      unauthorized: {"Missing or invalid key", "application/json", Schemas.Error},
+      forbidden: {"Insufficient scope", "application/json", Schemas.Error}
+    ]
+  )
+
+  def create(conn, params) do
+    user = conn.assigns.current_user
+
+    with {:ok, client} <-
+           OAuth.create_client(user.id, client_attrs(params), Audited.attribution(conn)) do
+      conn |> put_status(:created) |> render(:show, client: client)
+    end
+  end
+
+  operation(:show,
+    summary: "Get an OAuth client",
+    parameters: [id: [in: :path, type: :string, description: "Client record id"]],
+    responses: [
+      ok: {"Client", "application/json", Schemas.OAuthClient},
+      not_found: {"No such client", "application/json", Schemas.Error},
+      unauthorized: {"Missing or invalid key", "application/json", Schemas.Error},
+      forbidden: {"Insufficient scope", "application/json", Schemas.Error}
+    ]
+  )
+
+  def show(conn, %{"id" => id}) do
+    user = conn.assigns.current_user
+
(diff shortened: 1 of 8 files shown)


jhgaylor:
**Note on the red "A change to the SDK says whether it releases".**

That run ([34560373305](https://github.com/managoat/fountain/actions/runs/34560373305)) fired on `opened`, before the `sdk-no-release` label existed on this PR, and correctly said what it says: the PR changes `sdk/typescript/src/generated/openapi.ts` and leaves the version at 1.27.0.

Adding the label fired a second run, [34560380801](https://github.com/managoat/fountain/actions/runs/34560380801), which is **skipped** — the workflow listens for `labeled` for exactly this reason. The escape hatch is deliberate and the sentence it asks for is in the description: the wire types are regenerated here, and the bump that publishes them is #1885, the last PR of the stack.

Nothing to re-run; the stale conclusion is the first run's.

BinaryBourbon (changes requested):
The API surface itself is right — full scope with the escalation argument spelled out, 404-not-403 for another account's client on all three verbs, the non-UUID id landing as 404 rather than a cast error, and a scope test that hits every route with a sprite key. The seven gates for a new route are all walked: operation declarations, the 403 response, the contract entries, the omissions reason, and the regenerated TypeScript types.

Two things before this merges.

## 1. The required check is red and cannot go green as written

`A change to the SDK says whether it releases` fails on this PR:

> This PR changes what the SDK ships (`sdk/typescript/src/generated/openapi.ts`) but leaves the version at 1.27.0.

That is the gate working correctly — it compares this PR's diff against *its own base*, and the `1.28.0` bump lives at the tip of the stack in #1885. Bumping here instead would be the wrong fix: the house rule is that SDK versions race on `main`, so the bump belongs in the last PR, which is where you put it.

The gate names the right escape itself: **add the `sdk-no-release` label to this PR**. The label exists on the repo ("Changes the SDK's published surface deliberately without cutting a release") and describes this situation exactly. #1885 then carries the release for the whole stack.

Worth checking whether the same thing bites any future stack that touches generated SDK output in a middle PR — the label is the answer every time, but it is not discoverable until the check goes red.

## 2. A no-op `PATCH` writes an audit row

`OAuthClientUpdateRequest` requires nothing, so `PATCH /api/oauth/clients/:id` with `{}` is a legal request. It goes:

```
client_attrs(%{"id" => ...})  #=> %{}
Client.changeset(client, %{}) #=> changes: %{}   (put_change of an identical
                              #    redirect_uris / origin_keys is a no-op)
Repo.update/1                 #=> {:ok, client}, no query issued
audited(...)                  #=> oauth_client.updated, metadata "changed" => []
```

CLAUDE.md is explicit that this is not allowed: *"Only record what happened. A rejected changeset or a no-op sync records nothing; a trail that logs attempts as changes is worse than no trail."* An `oauth_client.updated` row that says nothing changed is exactly the shape that rule names.

It reaches the console too — #1883's `save` with unmodified fields takes the same path — so the fix belongs in `update_client/3` in #1877 rather than in this controller: return `{:ok, client}` early when `changeset.changes == %{}`, or make `audited/3` skip an empty `changed` list. A test asserting that a PATCH with no changes leaves no trail would pin it, alongside the "a refused update leaves no trail" test that is already there.

## Nit

`origins` in the JSON (and the Go CLI's `origins:` line in #1884) is `Client.origins_of/1`, so a loopback client reports `http://localhost:5173` while CORS actually admits `http://localhost` on **any** port. The comment says the field exists so a caller "sees the origins it will actually be calling `/api` from", which is the one job it does not quite do for the loopback case that the whole any-port design is built around. Either derive it from `origin_key/1` or say "any port" in the schema description.

Requesting changes for the red check; the audit row is small but it is a documented rule.

jhgaylor:
Both review items addressed; force-pushed (amended commit).

**1. The SDK release gate.** Labelled `sdk-no-release`. The bump stays at the tip of the stack in #1885, which is where a version bump belongs so it does not race another one on `main`; this PR changes the published surface without cutting the release, which is exactly what the label is for. The gate's workflow listens for `labeled`, so it re-evaluates on its own.

**2. The no-op PATCH audit row.** Fixed at its source in #1877 rather than here — `update_client/3` now returns before the audit when the changeset is valid and empty, so the console (#1883) gets the same fix for free. Verified by reverting the guard. This PR adds the API-level half: a `PATCH` that repeats what it read is a 200, not a 422.

**3. The `origins` nit.** Took the "say it in the description" option rather than changing the field: `origins` stays the real origin list, and the schema now says a loopback origin is admitted on any port, not only the port shown. `src/generated/openapi.ts` carries the same sentence.

One note on that last file. Regenerating it end to end here (`mix openapi.spec.json` then `npm run generate`) produced a 6,059-line reordering on top of the one-line change — the path key order in the serialised spec came out different on this machine, while the file content was otherwise identical (verified by diffing both files sorted: the only line that differs is the description). Since the committed file is the one CI already regenerated and matched on this PR, I applied the single description line to it directly rather than committing a reordering that CI would only have to undo. Worth knowing that this reordering is reachable, because it would look like a real diff to anyone who regenerates locally.

BinaryBourbon (approved):
Both blockers are cleared. Switching from changes-requested to approve.

## 1. The required check

`sdk-no-release` is on the PR and `A change to the SDK says whether it releases` is now `SKIPPED`. That is the right shape: the label records that this PR changes the SDK's published surface deliberately without cutting a release, and #1885 carries the 1.28.0 bump for the whole stack, where a version bump belongs so it does not race another on `main`.

`stack:1125` is on it too, so the label is discoverable from the stack rather than only from the red check that caused it.

## 2. The no-op audit row

Fixed in #1877's `update_client/3` rather than in this controller, which is where I asked for it, so the console in #1883 is covered by the same change. Details on that PR; the short version is that the early return is guarded on `changeset.valid?` as well as `changes == %{}`, so a published client's refusal still goes through `Repo.update/1` and still returns an error.

The test added here is the one that pins the *HTTP* half of it:

```elixir
test "a PATCH that moves nothing is a 200, not a 422"
```

Worth having separately from the context test, because "no field of `OAuthClientUpdateRequest` is required" is a statement about the schema, and the thing that would break it is a schema edit rather than a context edit. The comment says exactly that.

## 3. The `origins` nit

Taken the schema-description way rather than the `origin_key/1` way:

> Origins derived from redirect_uris and admitted by CORS. A loopback origin is admitted on any port, not only the port shown here.

That is one of the two fixes I offered and it is the cheaper one — the field keeps returning the concrete origins a caller can display, and the caveat rides along in the generated types, which is where a TypeScript consumer will actually read it. `sdk/typescript/src/generated/openapi.ts` is regenerated with it; nothing else generates from OpenAPI, so there is no second artifact to chase.

#1883 went further and put the caveat in the console card. The **CLI's `origins:` line in #1884 did not** — `printOAuthClient` still prints `strings.Join(a.Origins, ", ")` bare, and a schema description does not reach a terminal. Noted there rather than here; it is a leftover of the same nit, not a new one.

Approving.

PR #1883, opened 2026-09-11
feat(console): Account, then OAuth apps (#1125)

jhgaylor:
Part 7 of 9 for #1125 (re-cut of #1489). Based on #1882.

Register an app, see its generated `client_id`, edit its redirect URIs and remove it, at `/account/oauth-apps`. Console work in the console's sense: account configuration, not an interactive application.

The page says what the mode means rather than hiding it. Every app is badged **In development**, the copy says it signs in nobody but you, and the delete confirmation says what deleting does and does not do — new sign-ins stop, keys already issued stay valid until revoked under API keys. A published app offers neither Edit nor Delete, because an operator owns that registration now.

Redirect URIs are a textarea, one per line; a comma is accepted as a separator because it is the common near-miss.

One test is about a race rather than a feature: a submit already in flight arrives after the cancel that closed the form, and the save must be a no-op rather than taking the view down.

Left to the next PRs: the CLI, and the docs.

Part of #1125

### DIFF
--- a/apps/fountain/lib/fountain_web/components/layouts.ex
+++ b/apps/fountain/lib/fountain_web/components/layouts.ex
@@ -99,6 +99,7 @@ defmodule FountainWeb.Layouts do
                 current={@current_path}
               />
               <.nav_link href={~p"/account/runners"} label="Runners" current={@current_path} />
+              <.nav_link href={~p"/account/oauth-apps"} label="OAuth apps" current={@current_path} />
               <.nav_link
                 :if={
                   assigns[:current_user] &&
--- a/apps/fountain/lib/fountain_web/live/oauth_clients_live/index.ex
+++ b/apps/fountain/lib/fountain_web/live/oauth_clients_live/index.ex
@@ -1,0 +1,263 @@
+defmodule FountainWeb.OAuthClientsLive.Index do
+  @moduledoc """
+  Account → OAuth apps (#1125): the clients this account registered so that an
+  app it is building can offer "Sign in with Fountain".
+
+  Every client here is in development mode, which is what makes the redirect
+  URI field free-form: it signs in its owner and renders an error page for
+  everyone else, so the only account a badly-chosen redirect can send anywhere
+  is the owner's own.
+  """
+  use FountainWeb, :live_view
+
+  alias Fountain.OAuth
+  alias Fountain.OAuth.Client
+  alias FountainWeb.Audited
+
+  @impl true
+  def mount(_params, _session, socket) do
+    user = socket.assigns.current_user
+
+    {:ok,
+     socket
+     |> assign(:page_title, "OAuth apps")
+     |> assign(:user_id, user.id)
+     |> assign(:clients, OAuth.list_clients(user.id))
+     |> assign(:editing, nil)
+     |> assign(:form_errors, [])
+     |> assign(:form, blank_form())}
+  end
+
+  @impl true
+  def handle_event("new", _params, socket) do
+    {:noreply, socket |> assign(:editing, :new) |> assign(:form, blank_form()) |> clear_errors()}
+  end
+
+  def handle_event("edit", %{"id" => id}, socket) do
+    case OAuth.get_client_record(id, socket.assigns.user_id) do
+      nil ->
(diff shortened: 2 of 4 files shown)


BinaryBourbon (approved):
Console half. Scoping is right — both `edit` and `delete` re-fetch through `get_client_record/2` rather than trusting `phx-value-id`, and the cross-tenant test asserts the other account's `client_id` never reaches the HTML. The published-client branch hides Edit and Delete *and* the context refuses them anyway, so the UI is a convenience rather than the enforcement, which is the right way round.

The "a save with no form open is a no-op, not a crash" test is the good one: a submit already in flight behind the cancel that closed the form is a real LiveView race, and `:no_form` handles it without a flash nobody asked for. Worth keeping the comment.

Notes:

**`split_uris/1` splits on commas as well as newlines.** "One URI per line is what people paste; commas are a common near-miss" is fair, but a redirect URI may legally carry a comma in its query string, and that one gets silently cut into two invalid URIs that the changeset then rejects with a confusing message. Rare enough that I would not block on it; if you keep it, the helper text under the textarea says "One per line" and does not mention commas, so the behaviour is undocumented either way. (Same shape as the `StringSlice` note on #1884, which splits on commas for the same reason.)

**`Enum.join(Client.origins_of(c.redirect_uris), ", ")`** under each card has the loopback under-reporting I raised on #1882: it will say `Calls /api from http://localhost:5173` when the actual rule is any port on `http://localhost`. Same fix, same place.

**Raw Tailwind in the "In development" chip** (`bg-amber-50 text-amber-700`) while the rest of the template is on `var(--color-*)`. Consistent with the consent page in #1878, so it is a stack-wide choice rather than a slip here — just noting it so it is a choice.

**`error_messages/1` duplicates the `traverse_errors` block** found in eight other LiveViews in this app. Not a finding — it matches the surrounding code, which is the standard I would rather this met than a new shared helper introduced in a feature PR.

Approving.

jhgaylor:

- **The card no longer under-reports loopback.** "Calls /api from …" now goes through `origins_line/1`, which appends "(and a loopback origin on any port)" when any registered URI is loopback — the port shown is not the only one CORS admits. Pinned by an assertion in the existing registration test, which already registers a `localhost` URI.
- **The comma split is documented.** The helper text under the textarea now says "One per line, or separated by commas", so the behaviour the code deliberately has is the behaviour the field advertises. The split itself is unchanged.

The no-op-save audit row is fixed in #1877, so the console gets it through the context rather than needing its own guard.

BinaryBourbon (approved):
Both notes from the first pass are answered, one by fixing and one by documenting.

## The loopback under-reporting

`origins_line/1` is the fix, and it is better than the schema-description version in #1882 because the console is where a person reads this and then goes and configures something:

```
Calls /api from http://localhost:5173 (and a loopback origin on any port)
```

`loopback_uri?/1` is safe on garbage — `URI.parse/1` returns `host: nil` and `Client.loopback?/1` has a `def loopback?(_), do: false` clause — so a half-typed URI in the list does not take the card down. The test asserting `html =~ "any port"` with the comment naming why (":5173 is not the only origin CORS admits") is the right pin; a future refactor that drops the suffix fails on a string rather than on a layout.

One wording nit, take it or leave it: for a client registered with *only* a loopback URI the line reads `http://localhost:5173 (and a loopback origin on any port)`, where "and" suggests a second origin rather than the same one relaxed. "on any port" alone would read truer there, but branching on it is not worth the code.

## The comma split

Kept, and documented instead:

> One per line, or separated by commas. https, unless the host is localhost or 127.0.0.1.

That closes the half I actually objected to — the behaviour was undocumented either way. A redirect URI carrying a comma in its query string still gets cut, but the field now says so, and the changeset rejects both halves with a message pointing at them.

It does leave the console and the CLI disagreeing about one input: #1884 moved `--redirect-uri` to `StringArray` so a comma there is preserved verbatim, while the same URI pasted into this textarea is split. Both are defensible for their surface — a textarea is a paste target, a flag is typed once — and I would not unify them. Worth knowing the two exist.

Nothing else moved. Scoping, the `:no_form` race test and the raw-Tailwind chip are untouched.

Approving.

PR #1886, opened 2026-09-11
refactor(conversations): four function families move out of the server (#1565)

jhgaylor:
Re-cut of #1569 as a stack of eight on `main`, under the no-big-PRs rule. #1569 was 2,351 lines across 56 files and 161 commits behind; this stack is re-derived against today's code rather than rebased.

This is the room-making half. Four function families leave `ConversationServer` for the modules that already own their subject, with no behaviour change:

- the callback-key revocation in `terminate/2` becomes `CallbackKey.revoke/2`;
- the orphaning of a turn a normally-stopping server leaves behind becomes `TurnMachine.orphan_on_normal_stop/3`;
- the `format_status/1` redactor becomes `Redaction.server_state/1` and `Redaction.server_status/1`;
- the checkpoint pair becomes `Fountain.Conversations.Checkpoints`, a module of its own beside the `Provisioning.restore_checkpoint/2` and `create_checkpoint/2` it calls.

That last one is a module of its own rather than more of `Provisioning` for a reason worth knowing: those two calls are what the server tests stub, and **Mimic cannot intercept a call a module makes to itself**. Folding the family into `Provisioning` compiled, passed every targeted test, and silently took two warm-start tests off the stub and onto the real Sprites client. The full suite caught it.

2,774 lines to 2,627.

**Why the size pin does not move here.** Later links of this stack add to the server, so lowering it now would mean raising it two PRs later, and the pin only ever goes down. ~~#1891 lowers it once, to 2,716, which is what the file is with everything this stack adds.~~

**Corrected 2026-09-11.** #1891 did lower it to 2,716, which left zero headroom, and two rounds of review fixes below it took the file to 2,739 — turning #1891, #1892 and #1893 red on the pin they had set themselves. The pin now stays at main's 2,774 for the whole stack, which still passes at every link because the stack is a net reduction (2,774 to 2,739). Lowering it is a follow-up once the shrink is on `main` and the number has stopped moving; #1891 carries that rule in the size test's moduledoc.

`events_test.exs` gets `checkpoints.ex` on its source list. The webhook catalogue guard reads `publish_stage` call sites out of a named set of files, and the `checkpoint_restore` stage moved with the code; without the line the guard reports three catalogue entries as stale. The same class of thing as the Mimic seam above, and also caught by the full suite rather than by a targeted one.

`checkpoint_test.exs`, `home_checkpoint_test.exs`, `conversation_server_redaction_test.exs` and `conversation_server_shutdown_revoke_test.exs` are the existing proof and all pass unchanged.

## The stack

| | PR | What |
|---|---|---|
| 1 | this | server subtraction, no behaviour change |
| 2 | #1887 | the revision, fingerprint and applied-skills columns |
| 3 | #1888 | what can be reconfigured in place, and what cannot |
| 4 | #1889 | `Conversations.reapply_conversation/3` |
| 6 | #1891 | skills reconciliation |
| 7 | #1892 | `POST /api/conversations/:id/reapply` |
| 8 | #1893 | the manual, three SDKs and the version bump |

Part 1 of 8 for #1565.

---

| | PR | What it covers |
|---|---|---|

### DIFF
--- a/apps/fountain/lib/fountain/conversations/callback_key.ex
+++ b/apps/fountain/lib/fountain/conversations/callback_key.ex
@@ -156,4 +156,36 @@ defmodule Fountain.Conversations.CallbackKey do
         {:error, conv}
     end
   end
+
+  @doc """
+  Revoke the key a stopping server minted, if the row still names it.
+
+  A server that stops for any reason gives its sprite's callback key back.
+  The row check keeps a server that already handed the conversation on from
+  revoking its successor's key: reading the row's id at call time made a
+  dying duplicate (Horde's CRDT merge mass-terminates losers) revoke the
+  SURVIVING server's live credential, and its sprite then 401'd on every
+  callback and sub-agent spawn, surfaced nowhere. Best-effort: a key that is
+  already gone or inert is not an error.
+
+  If the BEAM dies hard (SIGKILL is untrappable) the `api_keys` row stays, but
+  it is not dangerous. `api_key_opts/0` sets an `expires_at`, so an unrevoked
+  key stops authenticating on its own and RetentionPruner deletes long-expired
+  rows. `SandboxReaper` is the sprite half, which does not self-heal.
+  """
+  def revoke(conversation_id, key_id) when is_binary(conversation_id) and is_binary(key_id) do
+    # ownership: the caller is the conversation's own server, which
+    # established ownership before it started. This read only confirms the row
+    # still names this key, and attributes the revocation to that owner.
+    case Conversations._unsafe_get_conversation(conversation_id) do
+      %Conversation{user_id: user_id, callback_api_key_id: ^key_id} when is_binary(user_id) ->
+        _ = Accounts.revoke_api_key(user_id, key_id, actor: "system:conversation_server")
+        :ok
+
+      _ ->
+        :ok
+    end
+  end
+
+  def revoke(_conversation_id, _key_id), do: :ok
 end
--- a/apps/fountain/lib/fountain/conversations/checkpoints.ex
+++ b/apps/fountain/lib/fountain/conversations/checkpoints.ex
@@ -1,0 +1,109 @@
+defmodule Fountain.Conversations.Checkpoints do
+  @moduledoc """
+  An environment's disk checkpoint: the warm start that restores one onto a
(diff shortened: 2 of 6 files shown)


BinaryBourbon (approved):
Approving. I read this as a pure move and checked each of the four families for behaviour drift rather than trusting the diff shape:

- **`CallbackKey.revoke/2`** — the old `if state.conversation_id && state.callback_api_key_id` becomes `when is_binary(...) and is_binary(...)` with a `:ok` fallback clause. Equivalent for every value those fields can hold, and the row-still-names-this-key check survives intact, which is the part that matters (the dying-Horde-duplicate bug in the docstring).
- **`TurnMachine.orphan_on_normal_stop/3`** — `reason == :normal and state.current_turn` becomes a `:normal` head plus `not is_nil(turn)`; `current_turn` is a `%Turn{}` or nil, so same. The `rescue` moved inside the function and still returns `:ok`, so it cannot take the rest of `terminate/2` with it.
- **`Redaction.server_state/1`** — same five fields, same `[REDACTED]`, `redact`/`redact_map` renamed to `secret`/`secrets` only.
- **`Checkpoints`** — moved verbatim, including the `ok when ok == :ok or (is_tuple(ok) and elem(ok, 0) == :ok)` clause and the off-by-default flag.

Ordering in `terminate/2` is preserved (redaction delete, then revoke, then orphan), which is the one thing a move like this can quietly get wrong.

Two things I checked that were easy to miss and are right: `checkpoints.ex` is on `events_test.exs`'s `@sources` (the `checkpoint_restore` stage moved with the code), and `Fountain.Conversations.Turn` already declares `@type t`, so the new `@spec` does not trip the type gate.

**On the size pin.** I went looking for the missing `@pin` change before reading your explanation, so for the record I agree with it: lowering to 2,627 here and raising to 2,716 in #1891 would break "never raise it" twice to satisfy it once, and #1891 setting it to the stack's true final value is the honest version. Worth stating plainly in the tracker though — with #1890 adding 76 lines back, the ratchet ends the stack at 2,716, so this eight-PR effort nets the server 58 lines smaller, not 147. That is fine, but "server shrink" in the branch name is doing some work.

No blocking comments. The three findings I have on this stack are on #1888, #1889 and #1890.

PR #1887, opened 2026-09-11
feat(conversations): the three columns a reapply reads (#1565)

jhgaylor:
`conversations.configuration_revision` counts committed selections. `sandboxes.build_fingerprint` records the digest of the Environment fields provisioning turned into disk state, and `sandboxes.applied_skills` the skill selection the machine was last reconciled to.

`Fountain.Conversations.Reapply` arrives with the digest alone. `fingerprint/1` covers packages, repositories, the setup script and the network policy, and deliberately not variables or the checkpoint: a variable reaches a running machine on its next spawn, and a checkpoint only ever applies to a machine being built. Provisioning writes the digest and the mounted skills when the machine reaches `ready`, so a row written from here on answers "would this selection need the disk built again" for itself rather than by inference.

Nothing reads either column yet. The rule that does is the next link, and the reapply that calls it the one after.

Part 2 of 8 for #1565, on #1886.

---

| | PR | What it covers |
|---|---|---|

Part of #1565


BinaryBourbon (approved):
Approving. The three columns are the right three, the migration is reversible and ordered after `20260910130000_add_labels_to_conversations`, and both new fields are added to the `Sandbox` changeset's cast list (easy to add a field to the schema and forget that half).

I like that `build_fingerprint` and `applied_skills` are written at the `ready` transition rather than at create — that is what makes them describe the disk instead of the intent — and that `fingerprint/1` deliberately excludes `env_vars` and `checkpoint_id`, with a test that pins the exclusion rather than just asserting the inclusion.

One non-blocking note, on `fingerprint/1`:

```elixir
:erlang.term_to_binary({env.packages, env.repositories, env.setup_script,
                        env.networking_type, env.networking_config})
```

This digest is **persisted**, so it has to stay comparable across restarts and upgrades, and `term_to_binary` is not specified to be stable for maps — the external term format leaves map ordering undefined, and only flatmaps (under 32 keys) happen to serialise in key order today. `packages` and `networking_config` are JSONB-loaded maps, so they go through that path. An OTP bump that changed map encoding would make every stored fingerprint mismatch at once.

It fails closed (`{:rebuild_required, ...}` rather than a wrong reconfiguration), so this is not urgent, and the blast radius is "every reapply starts refusing" rather than anything unsafe. But since the value outlives the VM that computed it, a canonical encoding would cost almost nothing here — `Jason.encode!` over a term with sorted keys, or an explicit iolist of the five fields — and would remove the dependency on an unspecified property. Worth a line in the docstring at minimum, saying the digest is only ever compared against one produced by the same release.

Two smaller observations, neither needing action:

- `applied_skills` has no DB default, so pre-existing rows read `nil` rather than `[]`. #1889 and #1891 both spell `sandbox.applied_skills || ...`, so this is handled — just noting the schema and the code disagree about the empty value.
- `fingerprint/1` truncates to 32 hex chars (128 bits). Fine for a change-detector; not a problem.

Issue #848, opened 2026-08-19
agents: the system prompt is stored but never delivered to the runtime


jhgaylor:
`agents.system` is stored, edited in the form, exported — and never reaches a sandbox. Nothing under `lib/fountain` reads it except exports and the forms; under ACP `session/new`/`session/prompt` carry `cwd`, `mcpServers` and the user's text. Every agent has been running on its CLI's default persona; found when a freshly specced team-lead answered as plain Claude Code ("no roster here, where should I look?"). Fix: write it to each runtime's user-level instructions file at provision and at reattach.

PR #1888, opened 2026-09-11
feat(conversations): what can be reconfigured in place, and what cannot (#1565)

jhgaylor:
`Reapply.check/2` answers one question: can the machine this conversation is already on be reconfigured into the selection being asked for.

Variables, the system prompt, skills, MCP servers, the model and the permission policy are either process environment or a file, and the wake path has rewritten both of those under a live sandbox since #848. A different runtime, different packages, different repositories, a different setup script or a different network policy are not: the ACP adapter is installed before the network policy that would now block installing another, `git clone` refuses a checkout that already exists, a setup script that starts services fails on its second run, and the egress rules are written exactly once.

So `check/2` compares the digest the machine recorded with the digest of the selection, and returns `{:error, {:rebuild_required, field}}` naming the first field that forced it. A row written before this stack has no digest and falls back to the environment it records, which still catches a selection pointing somewhere else. `explain/1` is the sentence each blocker gets, for the API error and the log.

The network policy is the cautious one of the five. Nothing has ever re-run `Egress.apply_policy/4` against a live machine, so rather than assume it is idempotent and find out in production, a networking change is refused. Relaxing that is a one-line change to `fingerprint/1` once somebody has shown the re-application is safe.

`:shared_sandbox` is declared here and reported by the context, which is the only place that can see a machine's cotenants. Nothing calls `check/2` yet.

Part 3 of 8 for #1565, on #1887.

---

| | PR | What it covers |
|---|---|---|

Part of #1565

### DIFF
--- a/apps/fountain/lib/fountain/conversations/reapply.ex
+++ b/apps/fountain/lib/fountain/conversations/reapply.ex
@@ -37,11 +37,41 @@ defmodule Fountain.Conversations.Reapply do
   is refused. Relaxing that is a one-line change to `fingerprint/1`, once
   somebody has shown the re-application is safe.
 
-  This module is built across the #1565 stack. This first part is the digest
-  alone: the rule that reads it lands next.
+  One gap is worth naming. A skill whose source is a GitHub repository is a
+  clone, and by the time a reapply runs the machine's network policy is
+  already in force. Bundled skills are file writes and always land; a remote
+  one may not, under a restrictive policy.
+
+  So a reapply that needs either of those is refused, and says which field
+  forced it. Start a new conversation for that, or build the machine under
+  the conversation again with `DELETE /api/sandboxes/:id` (#1071).
+
+  ## The cotenant rule
+
+  Skills, the instructions file and `.mcp.json` sit at per-sandbox paths
+  (`Managoat.Runtimes.Layout`), not per-conversation ones, so rewriting them
+  rewrites them for every conversation on that machine. Sharing only happens
+  on a persistent home or an explicit `sandbox_id` attach, and
+  `Conversations.check_attachable/4` already pins every conversation on a
+  machine to one `(user, agent, environment, vault)`. A conversation that has
+  the machine to itself can therefore be reconfigured freely; one that shares
+  it may only be reapplied to the selection its cotenants already have, which
+  is what a refresh is. The context applies that rule and reports it as the
+  `:shared_sandbox` blocker below.
   """
 
+  alias Fountain.Conversations.Sandbox
   alias Fountain.Environments.Environment
+
+  @typedoc "Why a selection cannot be applied to the machine that is already there."
+  @type blocker ::
+          :runtime
+          | :packages
+          | :repositories
+          | :setup_script
+          | :networking
+          | :environment
+          | :shared_sandbox
 
   @doc """
   The digest of the Environment fields that provisioning turns into disk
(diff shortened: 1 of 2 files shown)


BinaryBourbon (changes requested):
The rule is well-shaped — runtime checked before anything else, fingerprint equality as the fast path, `nil` sandbox as trivially applicable, and the legacy fallback to the recorded environment is exactly the right reading of a null column. The `explain/1` sentences are genuinely useful error copy.

One thing I want changed before this goes in, because the docstring makes a promise the code does not keep on the primary path.

### `build_field/2` cannot name the field on a refresh

`check/2` says the refusal "says which field forced it", and `explain/1` has a sentence per field so a client can tell `packages` from `setup_script`. That works when the reapply is a **rebind** to a different Environment row. It does not work when it is a **refresh** — the case issue #1565 lists first ("an Agent, Environment or Vault changed and the user wants the existing conversation's next machine to receive the new system prompt, skills, MCP config, variables or secrets").

On a refresh, #1889 calls it like this:

```elixir
target_environment: environment_for(target_environment_id, conv.user_id),
built_with: sandbox && environment_for(sandbox.environment_id, conv.user_id)
```

Both are fresh reads of the **same row**. So when someone edits their environment's `packages` in place and reapplies:

- `built_fingerprint/2` returns the digest stored at build time — computed from the *old* packages;
- `fingerprint(target_env)` is computed from the *new* packages;
- they differ, so the `cond` falls to `build_field(built_with, target_env)`;
- but `built_with == target_env`, every branch compares equal, and it returns the catch-all `:environment`.

The user gets "the selected environment builds the machine differently", not "the selected environment installs different packages". The stored digest knows something changed; nothing left on the row knows *what*.

### Why the suite does not catch it

`reapply_check_test.exs` passes `built_with: ctx.env` — the pre-edit struct still held in the test process — against `target_environment: rebuilt`. That is an input the production call site cannot construct, since it re-reads both sides from the database. The per-field assertions in #1889 (`refuses an environment whose build inputs differ`, `refuses a networking change`) both `insert_env` a *second* environment, so they exercise the rebind path too. Nothing in the stack covers refresh-after-in-place-edit, which is why this reads as covered.

### Two ways out, either is fine by me

1. **Make the claim true.** Store the digest per field rather than as one string — `build_fingerprint` as a small JSON object of four digests, or four columns. `build_field/2` then diffs the stored digests against freshly computed ones and needs no second copy of the old environment at all. This also drops the `built_with` parameter and the legacy-fallback asymmetry.
2. **Soften the claim.** Keep the single digest, say in the docstring and in `explain(:environment)` that the field is only nameable when the selection points at a different Environment, and add the missing test asserting `{:rebuild_required, :environment}` for the in-place-edit case so the behaviour is pinned rather than accidental.

I would take (1) if the per-field message is worth a column, and (2) if it is not — but the docstring and the code should agree either way. CLAUDE.md is explicit that a document describing behaviour that is not in the code is the defect, and this is the same class one level down.

### Non-blocking

- `blocker()` includes `:shared_sandbox`, which #1889 produces and this module only explains. Fine as a shared vocabulary, but the type is now "why a reapply was refused" rather than the "why the machine cannot be reconfigured" its `@typedoc` says, since sharing is not a property of the machine's build.

jhgaylor:
Taken the second of the two routes you offered, in `4b7b5d2`.

You were right that the docstring and the code disagree, and right about why the suite could not see it: `reapply_check_test.exs` fed `built_with` a pre-edit struct that the production call site cannot produce, because `assert_applicable_in_place/4` reads both sides from the database.

I went with **soften the claim** rather than per-field digests. The field name is a nicety and the remedy is identical for every blocker, so it did not seem worth a column shape change reaching back into #1887 and re-running the stack. If you disagree, the per-field version is a contained change and I am happy to do it — say so and I will.

What landed:

- `check/2` gains a **What the refusal can name** section saying the field is nameable only when the selection moves to a different Environment, that a refresh of one edited in place is the case that cannot, and why (the stored digest catches the move, nothing left on the row says which input moved).
- `build_field/2`'s comment names both `:environment` fallbacks rather than only the legacy one.
- The rebind test is renamed to say it is the rebind, with a comment that the field is nameable only while `built_with` and `target_environment` are two rows.
- A new test pins the refresh: same row on both sides, `{:rebuild_required, :environment}`.

`explain(:environment)` was already true in both cases, so it is unchanged.

`mix format --check-formatted` clean, 9 tests 0 failures in that file, 121 across the reapply/webhook/audit set.

BinaryBourbon (approved):
Route (2), and it is the route I would have picked too. The claim and the code agree now.

I checked that the new test is not vacuous, because a pinning test for "the digest knows, the row does not" is easy to write as a tautology. `setup` stores `build_fingerprint: Reapply.fingerprint(env)` before the edit, so `built_fingerprint/2` really does return the stale digest and the `cond` really does fall through to `build_field/2` with two equal structs. Passing `edited` on both sides is the production shape, which is what the old `built_with: ctx.env` was not.

The rest of what landed is right: the rebind test renamed to say it is the rebind, `build_field/2`'s comment naming both `:environment` fallbacks, and `explain(:environment)` left alone because it was already true either way.

### One line left over

The moduledoc still carries the unqualified version of the promise:

> So a reapply that needs either of those is refused, and says which field forced it.

That is the sentence whose class started this — `check/2` now says the field is nameable only on a rebind, and the summary three paragraphs above it says it always is. A reader meets the moduledoc first. "and says which field forced it, where it can" plus a pointer to `check/2` closes it; no code change.

Approving, because that is an edit and not a decision.

On the `:shared_sandbox` typedoc point from last time: still fine as a shared vocabulary, no change wanted.

BinaryBourbon:
Re-checked at `4b7b5d2`: no new commit since my approval, so the one line from that review is still open.

`reapply.ex:45` still reads "So a reapply that needs either of those is refused, and says which field forced it." `check/2` now says the field is nameable only on a rebind; the moduledoc three paragraphs above still says it always is, and the moduledoc is what a reader meets first.

Non-blocking, as it was. Approval stands — just do not let it merge as the one sentence in the module that makes the promise you removed.

PR #1889, opened 2026-09-11
feat(conversations): reapply an Agent, Environment and Vault in place (#1565)

jhgaylor:
`Conversations.reapply_conversation/3` re-resolves a conversation's selection and writes it to the row, keeping the conversation, its transcript and the machine it is already running on. An omitted field keeps what is selected, an explicit null clears the Environment override or the Vault, and an empty map is a refresh.

Four things it is careful about:

- **One transaction, under the lock turn admission takes.** The conversation row is re-read `FOR UPDATE` inside it, so two reapplications preserve each other's omitted fields instead of each writing from a stale copy, and a turn cannot start on the configuration being replaced. The write itself is one guarded statement (`not exists` a running turn), so a caller that lost the race is told it is busy rather than silently overwritten.
- **The machine's binding identity moves with the conversation**, because `check_attachable/4` is what a later attach is matched against. A move onto a persistent home that already exists comes back as a changeset error on `:home` and rolls the whole reapply back, leaving both bindings where they were: two homes for one identity is exactly what the partial index prevents.
- **A machine with cotenants only accepts the selection they already have.** Skills, instructions and `.mcp.json` are per-machine paths, so anything else would reconfigure a conversation nobody asked about. Reported as `{:rebuild_required, :shared_sandbox}`.
- **A `pending` conversation is reapplicable once its machine is ready.** Nothing writes `idle` until a turn ends, so a conversation created without a prompt sits at `pending` for good; refusing every one of those would put "I picked the wrong agent before I sent anything" permanently out of reach. A provision genuinely in flight is still a retry.

Audited as `conversation.configuration_reapplied`, outside the transaction and naming the fields that moved rather than copying values. `audit_guardrail_test.exs` gets its row.

`applied_skills` is carried forward rather than replaced: it records what is on the disk, and the newly selected skills only become the recorded set once something installs them. That something is part 6.

**What this link deliberately leaves:** no live server is told yet, and turn admission does not check the revision this writes. Both are part 5.

### Differences from #1569

`main` moved under this. The original wrapped the reapply and the attach path in a shared `Reapply.transaction/2`; today's `create_attached_conversation/3` already rechecks `check_attachable/4` inside a transaction holding `FOR SHARE` on the sandbox row, and a reapply's `update_sandbox` takes the conflicting lock, so attach needs no change at all. The lock helper is private to `Conversations` instead, beside the namespace constant it uses.

Part 4 of 8 for #1565, on #1888.

---

| | PR | What it covers |
|---|---|---|

Part of #1565


BinaryBourbon (changes requested):
The mechanics here are the strongest part of the stack and I checked the load-bearing claims rather than taking them on trust:

- The advisory lock really is the one turn admission takes — namespace `4316` with `:erlang.phash2(sandbox_id)`, identical to the existing acquisition in `_unsafe_create_turn_on_sandbox`. The serialisation claim holds.
- `unique_constraint(:home, name: :sandboxes_home_identity_index)` is on the `Sandbox` changeset and the partial index exists with `nulls_distinct: false`, so the "move onto an existing home comes back as a changeset error and rolls back" story is real and not a `Ecto.ConstraintError` waiting to 500.
- The audit is recorded outside the transaction, with a comment saying why. `audit_guardrail_test.exs` gets its row. `_unsafe_current_version_id/1` sits directly after the scoped `get_agent/2` with the ownership comment. All three CLAUDE.md rules that usually get missed here are met.
- Every `resolve_*` in the `with` returns `{:error, _}`, so `with_sandbox_lock/2`'s `case` cannot `CaseClauseError` inside the transaction. I checked all five.
- Re-running `resolve_permission_policy/2` against the *new* agent is a nice touch — a rebind cannot widen permissions past the incoming agent's base policy.

The test list is genuinely thorough; `a conflicting persistent home rolls back both bindings` and `omitted fields use the latest committed selection` are the two I would have asked for.

One blocking ask, and it is paperwork rather than code.

### ADR 0023 needs its amendment

Issue #1565 says it outright: *"ADR 0023 defines persistent-home identity and shared-sandbox behavior; this feature should amend its outcome if the implementation adds a new transition."*

This adds one, and `Reapply.update_identity/4` is where it lands. ADR 0023's Outcome currently reads:

> **Identity key.** `(user_id, agent_id, environment_id, vault_id)` stays. The disk is materialized from the environment and vault at provision — env vars, packages, cloned repos, setup scripts — so a machine built from one environment is not a machine built from another, **and re-materializing in place is not built.**

and its lifecycle table says:

> | agent deleted / vault or env changed | n/a | **destroy the home** (identity no longer exists); a "reset home" action does the same on demand |

After this PR, a changed environment or vault does not destroy the home — it *moves the home's identity onto the new selection* and keeps the disk. That is a good decision, and the reasoning in `Reapply`'s moduledoc is better than what is in 0023 today. But right now a reader who starts where CLAUDE.md tells them to start (`decisions/index.md`) is told the opposite of what the code does, and `okf backlinks decisions 0023` will not lead anyone here.

What I think it needs, in this PR since this is the PR that makes the transition:

- an amendment to 0023 recording that a home's identity key is now mutable through one narrow door, that the door refuses when the build inputs differ (so "a machine built from one environment is not a machine built from another" still holds — this is the part that makes the change *consistent* with 0023 rather than a reversal of it), and that it refuses outright when the machine has cotenants;
- the lifecycle row for "vault or env changed" corrected, since destroy-the-home is no longer the only answer;
- `adr_status` / `verified` / `stale_after` frontmatter touched per the template, plus the `scripts/decisions-index.sh` regeneration in the same PR that CLAUDE.md requires.

I am not asking for a new ADR — 0023 is the one that is now wrong, and an amendment to it is cheaper and more findable than 0044.

### Non-blocking

- **`:shared_sandbox` under the `rebuild_required` error code.** #1892 renders it as `409 {"error":"rebuild_required","field":"shared_sandbox"}`, but rebuilding is not the remedy and the message correctly tells the caller to start a new conversation instead. The blocker vocabulary is doing double duty for "this machine cannot be reshaped" and "this machine is not yours alone". A separate error code would read better to a client that switches on `error`; keeping one code but documenting that `field` can be a non-field is the cheaper alternative.
- `assert_reapplicable/1`'s handling of `pending` is the right call and the comment explaining why is one of the more useful in the stack. No change wanted, just noting I checked it against `reapply_provision_in_flight?/1` and agree the sandbox status is the correct discriminator rather than the conversation status.

jhgaylor:
ADR 0023 amended in `ff0bb29`. You were right that this is the PR that makes the transition, so it is the PR that owes the amendment.

The amendment keeps 0023's reasoning rather than reversing it, because that reasoning is exactly what the door refuses on: the build inputs must be unchanged (the digest), the runtime must be unchanged (the adapter is installed before the network policy), and the machine must have no cotenants. "A machine built from one environment is not a machine built from another" still holds; what moves is the selection, not the shape of the disk.

Three edits beyond the new block:

- the lifecycle row for **agent deleted / vault or env changed** is corrected, since destroy-the-home is no longer the only answer;
- the resolved **Identity key** bullet records that the key is no longer immutable, with a pointer to the amendment;
- the partial unique index is written down as load-bearing **on update as well as on insert**, which is what makes a collision roll the reapply back rather than produce two homes for one identity.

One wording call worth flagging: the amendment names `Fountain.Conversations.reapply_conversation/3` rather than `POST /api/conversations/:id/reapply`, because the route does not exist until #1892 and the stack merges top down. It mentions the route as what the same stack adds. That keeps it honest at this point in the chain rather than describing unbuilt behaviour as built.

Frontmatter `description`, `generated.at` and `verified.at` are re-stamped, and `scripts/decisions-index.sh` was re-run in the same commit. `okf validate decisions` passes; its 4 warnings are pre-existing and none of them names 0023.

On the non-blocking `:shared_sandbox` point: left as is for now. Splitting it out of `rebuild_required` changes the error vocabulary #1892 documents and the SDKs in #1893, and it is a rename rather than a fix. Happy to file it as a follow-up if you want it tracked.

BinaryBourbon (approved):
The amendment is the right shape, and keeping 0023's reasoning rather than reversing it is what makes it readable as a continuation instead of a contradiction. The three refusals are the load-bearing part and they are all there.

Checked rather than taken on trust:

- **`generated.at` re-stamped to today is correct.** The template defines it as "ISO 8601 datetime of the last meaningful edit", not the creation date, so moving it off 2026-08-17 is right and I was ready to object to it.
- **The index row matches the frontmatter `description` exactly**, so `scripts/decisions-index.sh` really was re-run against this file rather than hand-edited.
- **The partial-index sentence is true on update.** `unique_constraint(:home, name: :sandboxes_home_identity_index)` is on the changeset and `nulls_distinct: false` is on the index, which is what makes a collision a rolled-back changeset error rather than two homes — the same thing I verified in the code last round, now written down where the next reader will find it.
- The corrected lifecycle row and the amended **Identity key** bullet both point at the amendment, so none of the three places a reader can land tells them the old story.

### One wording nit, which is the one you flagged inverted

You were careful in the amendment block — "the same stack puts it on the API as `POST /api/conversations/:id/reapply`" — precisely because the route does not exist at this link. Good call. The lifecycle table row then says it flat:

> a *selection* change that leaves the build inputs alone no longer destroys anything — `POST /api/conversations/:id/reapply` moves the home's identity onto the new triple and keeps the disk

`verified` is now stamped on this file, and that row describes a route this tree does not serve. Same treatment as the block above it, or name `reapply_conversation/3` there too.

Not blocking: it is one clause, the stack merges as a unit, and the honest version is three words away.

On `:shared_sandbox`: a follow-up issue is enough. I do not want the error vocabulary churning underneath #1892 and #1893 for a rename, but the question ("`field` can be a non-field") should not evaporate with this thread.

BinaryBourbon:
Re-checked at `ff0bb29`: no new commit since my approval, so the one line from that review is still open.

`decisions/0023-persistent-agent-sandbox.md:396` — the amended lifecycle row — still names `POST /api/conversations/:id/reapply` in flat present tense, which is what the amendment block above it deliberately avoided because the route lands in #1892. This file now carries `verified`.

Non-blocking, as it was. Approval stands.

PR #1890, opened 2026-09-11
feat(conversations): a turn is admitted against the configuration its server read (#1565)

jhgaylor:
The reapply of #1889 writes a new selection and bumps a revision. This is what makes a live server honour it.

`reapply_conversation/3` now calls `ConversationServer.refresh_configuration/2` after it commits, and only publishes the `configuration`/`done` lifecycle event when that succeeds. A refresh that fails is returned to the caller: telling every follower the machine is configured, over a server that never reloaded, is worse than an error. A server that is not running needs nothing, because its next wake builds from the row.

The server keeps the revision it loaded and hands it to turn admission. `_unsafe_create_turn_on_sandbox/4` compares it under the sandbox lock it already holds and answers `{:error, :configuration_changed}` on a mismatch, which `TurnMachine.open/5` surfaces as `:configuration_changed`. That is not a refusal: the server drops its connection and rebuilds from the row before delivering the prompt, so a prompt that raced a reapply runs against the new selection rather than the old one. A caller with no revision to offer is not checked, so nothing else changes shape.

`kick_turn/4` therefore returns a callback tuple rather than a state, because that one outcome needs a continuation.

`configuration`/`done` joins the webhook stage vocabulary.

Two tests are the point of this link: a stale revision cannot open a turn on either capacity shape, and a server that never got the notification reloads on the next prompt and spawns with the new environment.

Part 5 of 8 for #1565, on #1889.

---

| | PR | What it covers |
|---|---|---|

Part of #1565

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1181,6 +1181,20 @@ defmodule Fountain.Conversations do
   `{:error, {:rebuild_required, field}}` rather than silently applied or
   silently ignored. `Fountain.Conversations.Reapply` owns that rule and says
   why for each field.
+
+  ## What `{:ok, conv}` promises
+
+  That the selection is committed, and that no turn can open against the
+  previous one: `configuration_revision` moved, and turn admission compares it
+  with the revision the live server loaded.
+
+  It does not promise the running machine has already been reconfigured. A
+  server is told after the commit, and it can be mid-provision or gone by then.
+  Neither loses the change — the next wake builds from the row — so neither is
+  a failure of this call, and reporting one would hand the caller an error for
+  a selection that is already committed. The `configuration` stage event says
+  which of the two happened: `done` when a machine is configured now, `failed`
+  when it is selected and the machine has yet to catch up.
   """
   @spec reapply_conversation(Conversation.t(), map(), keyword()) ::
           {:ok, Conversation.t()} | {:error, term()}
@@ -1198,6 +1212,8 @@ defmodule Fountain.Conversations do
                do: do_reapply_conversation(current, attrs),
                else: {:error, :provisioning}
            end) do
+      metadata = reapply_metadata(previous, updated)
+
       # Outside the transaction: a failed audit insert would abort the
       # enclosing one and take the reapply with it.
       Audit.record(%{
@@ -1207,35 +1223,70 @@ defmodule Fountain.Conversations do
         resource_id: updated.id,
         actor: Keyword.get(opts, :actor, "self"),
         request_ip: Keyword.get(opts, :request_ip),
-        metadata: reapply_metadata(previous, updated)
+        metadata: metadata
       })
 
       broadcast_sidebar_update(updated.user_id)
-
-      # A refresh that fails is the answer, not a footnote: publishing a
-      # successful lifecycle event over a server that never reloaded would
-      # tell every follower the machine is configured when it is not.
-      with :ok <-
-             Fountain.Conversations.ConversationServer.refresh_configuration(
-               updated.id,
-               updated.configuration_revision
-             ) do
-        metadata = reapply_metadata(previous, updated)
-
-        publish_stage(updated.id, "configuration", "done", %{
-          event: "reapplied",
-          previous: metadata["previous"],
-          current: metadata["current"],
-          changed_fields: metadata["changed_fields"],
-          message:
-            "The configuration was reapplied on this machine. The transcript and the " <>
-              "files on disk are kept; the next prompt starts a new runtime session."
-        })
-
-        {:ok, updated}
-      end
+      announce_reapply(updated, metadata)
+      {:ok, updated}
     end
   end
+
+  # The selection is committed by the time this runs, so it is not in doubt and
+  # the caller is not told otherwise. What is still in doubt is whether the
+  # machine running now has read it, and that gets an event of its own rather
+  # than a `done` that would tell every follower the machine is configured when
+  # it is not.
+  defp announce_reapply(conv, metadata) do
+    common = %{
+      event: "reapplied",
+      previous: metadata["previous"],
+      current: metadata["current"],
+      changed_fields: metadata["changed_fields"]
+    }
+
+    case Fountain.Conversations.ConversationServer.refresh_configuration(
+           conv.id,
+           conv.configuration_revision
+         ) do
+      :ok ->
+        publish_stage(
+          conv.id,
+          "configuration",
+          "done",
+          Map.put(
+            common,
+            :message,
+            "The configuration was reapplied on this machine. The transcript and the " <>
+              "files on disk are kept; the next prompt starts a new runtime session."
+          )
+        )
+
+      # Total on purpose. This runs after the commit, so an unexpected shape
+      # here must become an event rather than a CaseClauseError that 500s a
+      # reapply which already happened.
+      other ->
+        publish_stage(
+          conv.id,
+          "configuration",
+          "failed",
+          common
+          |> Map.put(:reason, refresh_reason(other))
+          |> Map.put(
+            :message,
+            "The configuration is selected, but the machine it is running on has not " <>
+              "read it yet. It is applied when this conversation next wakes, and no " <>
+              "turn can run against the previous selection in the meantime."
+          )
+        )
+    end
+
+    :ok
+  end
+
+  defp refresh_reason({:error, reason}), do: refresh_reason(reason)
+  defp refresh_reason(reason) when is_atom(reason) or is_binary(reason), do: to_string(reason)
+  defp refresh_reason(other), do: inspect(other)
 
   defp do_reapply_conversation(conv, attrs) do
     agent_id = reapply_value(attrs, "agent_id", conv.agent_id)
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1211,7 +1211,29 @@ defmodule Fountain.Conversations do
       })
 
       broadcast_sidebar_update(updated.user_id)
-      {:ok, updated}
+
+      # A refresh that fails is the answer, not a footnote: publishing a
+      # successful lifecycle event over a server that never reloaded would
+      # tell every follower the machine is configured when it is not.
+      with :ok <-
+             Fountain.Conversations.ConversationServer.refresh_configuration(
+               updated.id,
+               updated.configuration_revision
+             ) do
+        metadata = reapply_metadata(previous, updated)
+
+        publish_stage(updated.id, "configuration", "done", %{
+          event: "reapplied",
+          previous: metadata["previous"],
+          current: metadata["current"],
+          changed_fields: metadata["changed_fields"],
+          message:
+            "The configuration was reapplied on this machine. The transcript and the " <>
+              "files on disk are kept; the next prompt starts a new runtime session."
+        })
+
+        {:ok, updated}
+      end
     end
   end
 
@@ -1565,6 +1587,10 @@ defmodule Fountain.Conversations do
   Create a turn on a sandbox that may be shared, refusing when the runtime's
   capacity is used up by another conversation's running turn.
 
+  `revision` is the conversation's `configuration_revision` as the caller
+  understands it, or nil for a caller with none; a mismatch answers
+  `{:error, :configuration_changed}` (#1565).
+
   `capacity` is `Managoat.Runtimes.ACP.concurrency/1`. All runtimes take the
   per-sandbox advisory lock and verify that the conversation still belongs
   to this nonterminal sandbox. An integer capacity also limits concurrent
@@ -1573,7 +1599,7 @@ defmodule Fountain.Conversations do
   nonempty allowance refuses the turn. Refusal writes no turn.
   Usage is recorded after the transaction commits, never inside it.
   """
-  def _unsafe_create_turn_on_sandbox(attrs, sandbox_id, capacity)
+  def _unsafe_create_turn_on_sandbox(attrs, sandbox_id, capacity, revision \\ nil)
       when is_binary(sandbox_id) and
              (capacity == :unbounded or (is_integer(capacity) and capacity > 0)) do
     conv_id = Map.fetch!(attrs, :conversation_id)
@@ -1588,12 +1614,21 @@ defmodule Fountain.Conversations do
         # The allowance's FK takes KEY SHARE on this row when first inserted.
         # UPDATE also fences that first insert when there is no allowance row
         # to lock yet. Keep both locks through the turn insert.
-        Repo.one(
-          from c in Conversation,
-            where: c.id == ^conv_id,
-            select: c.id,
-            lock: "FOR UPDATE"
-        ) || Repo.rollback(:sandbox_unavailable)
+        conv =
+          Repo.one(
+            from c in Conversation,
+              where: c.id == ^conv_id,
+              select: %{id: c.id, configuration_revision: c.configuration_revision},
+              lock: "FOR UPDATE"
+          ) || Repo.rollback(:sandbox_unavailable)
+
+        # The server passes the revision it loaded. A reapply committed since
+        # then means this turn would run against settings the server has not
+        # read, so it is refused here rather than started wrong (#1565). A
+        # caller with no revision to offer is not checked.
+        if not is_nil(revision) and conv.configuration_revision != revision do
+          Repo.rollback(:configuration_changed)
+        end
 
         attached? =
           Repo.exists?(
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1233,57 +1233,70 @@ defmodule Fountain.Conversations do
   end
 
   # The selection is committed by the time this runs, so it is not in doubt and
-  # the caller is not told otherwise. What is still in doubt is whether the
-  # machine running now has read it, and that gets an event of its own rather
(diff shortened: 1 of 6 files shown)


BinaryBourbon (changes requested):
This is the subtlest PR in the stack and I spent most of my time on it. The control flow is correct where I expected it to be fragile:

- `handle_continue({:reapply_prompt, ...})` pattern-matches `{:noreply, fresh}` and passes everything else through. I traced every return from `handle_continue(:provision, ...)` — it is `{:noreply, state}` or `{:stop, :normal, state}`, never a three-tuple — so nothing silently swallows the queued prompt.
- The retry cannot spin. `provision_with_rows/3` sets `configuration_revision: conv.configuration_revision` in the single state-assembly point *before* `dispatch_provision/7` branches, so both the fresh-provision and the reattach arms come back with the current revision. If that assignment lived in either arm instead, `kick_turn` → `:configuration_changed` → `:reapply_prompt` → `:provision` → `kick_turn` would be a hot loop against the database. Worth a comment saying the placement is load-bearing.
- `replying_ok/1` covers both shapes `kick_turn/4` can now return, and the new `handle_call` clauses are all above the state-redacting catch-all, so an unmatched call still cannot put plaintext secrets in a crash report.
- The revision guard itself is in the right place: inside the same transaction, after the `FOR UPDATE`, so it cannot race the write in #1889.

One blocking problem.

### A reapply that succeeded can be reported to the caller as a failure

The sequence in `reapply_conversation/3` is: commit the transaction → record the audit event → **then** call the live server, with the result gating the return value.

```elixir
with :ok <- ConversationServer.refresh_configuration(updated.id, updated.configuration_revision) do
  publish_stage(...)
  {:ok, updated}
end
```

There is no `else`, so anything that is not `:ok` becomes the return value of `reapply_conversation/3` — after the row has been written and the audit recorded. `refresh_configuration/2` returns `:ok` when `whereis/1` finds nothing, but hands the call to `call_server/2` otherwise, and `call_server/2` catches:

- `:exit, {:noproc, _}` → `{:error, :not_running}` — the server stopped between the `whereis` and the call;
- `:exit, {:timeout, _}` → `{:error, :provisioning}` — the call sat behind a long `handle_continue(:provision)` for 30s.

Both are the *same situation* as `whereis/1` returning nil: there is no live server holding stale configuration, and your own docstring says why that is fine — "the next wake builds from the row, which already says what to build". But instead of `:ok` the caller gets an error, and #1892 renders it as `422 {"error":"not_running"}` (via the terminal safety net, which also logs `fallback: unmapped error atom :not_running`) or `503 provisioning` with a `Retry-After`. The client is told to retry an operation that already took effect; retrying bumps `configuration_revision` again.

The `{:error, :conversation_busy}` case has the same shape and is the one the test pins:

```elixir
test "a refresh that fails is the answer, not a footnote", ctx do
  Mimic.expect(..., fn _, _ -> {:error, :conversation_busy} end)
  assert {:error, :conversation_busy} = Conversations.reapply_conversation(ctx.conv, %{})
  refute ... &(&1.stage == "configuration")
end
```

The test asserts the error comes back and that no `configuration` event is published. It never asserts what happened to the **row** — which is updated, with a bumped revision and an audit row saying the selection moved. So the test reads as proof the failure is clean, and the thing that is not clean is the part it does not look at.

I think the comment above this block is half right. Publishing `configuration`/`done` over a server that never reloaded would indeed lie to followers. But returning an error for a change that committed lies to the caller, which is worse, and right now there is no third option because `Fountain.Webhooks.Events` only admits `{"configuration", ~w(done)}` — there is no way to tell followers the selection moved but the machine has not caught up yet.

Suggestions, in the order I would try them:

1. **Fold `:not_running` into success.** It is definitionally the `whereis → nil` case observed a few microseconds later; `refresh_configuration/2` should normalise it rather than the caller. That removes the most likely instance of this on its own.
2. **Add `failed` to the `configuration` stage vocabulary** and publish it when the refresh genuinely could not be applied, so followers get the true state instead of silence. That is a one-line change to `events.ex` and a row in `docs/reference/webhooks.md`.
3. **Decide what `{:ok, updated}` means when the machine is behind**, and say it in the docstring and in `docs/api.md`. The revision guard already guarantees no turn can run on stale configuration, so "committed, machine catches up on next wake" is a perfectly good success — it just has to be stated rather than left to whether a `GenServer.call` happened to land.

Whatever you pick, the test should grow an assertion about the row, because that is the invariant that is actually at stake.

### Non-blocking

- `_unsafe_create_turn_on_sandbox/4`'s `revision \\ nil` means the autonomous-turn path in `connection.ex:279` is unchecked. I think that is fine — a reapply drops the connection, so there is no peer left to narrate an out-of-turn cycle — but the escape hatch is silent, and "a caller with no revision to offer is not checked" is doing a lot of work for a guard this stack describes as the thing that stops a turn on unread settings. A sentence naming the one caller that takes the exemption, and why it cannot matter, would close it.

jhgaylor:
Fixed in `e9a92ef`. This was the best catch of the three: the operation could commit, audit, and then tell the caller it failed.

Taken all three of your suggestions.

**1. `:not_running` folds into success.** `refresh_configuration/2` now normalises it, with a comment saying it is the `whereis/1` miss the first clause already answers `:ok` for, observed a few microseconds later.

**2. `configuration` gained `failed`.** Published when the refresh did not land, carrying the reason. As you spotted, the previous shape could not report this at all, since the vocabulary held only `done` and the error path published nothing, so followers got silence.

**3. The commit is the success.** `reapply_conversation/3` returns `{:ok, conv}` once the row is written. Its docstring has a **What `{:ok, conv}` promises** section: the selection is committed and no turn can open against the previous one, and explicitly *not* that the running machine has already been reconfigured, because a server can be mid-provision or gone and neither loses the change.

You were also right about the test. It asserted the error and the absent event and never looked at the row, which is why it read as proof the failure was clean. It now asserts the row moved, the revision is 1, and the event says `failed`; its opposite number asserts `done`.

One extra, unprompted: the failure branch is total (`other ->`) rather than matching `{:error, reason}`, with a `refresh_reason/1` that handles a non-atom. It runs after the commit, so an unexpected shape there would have been a `CaseClauseError` on a reapply that already happened, which is the same class of bug you reported one level down.

On the non-blocking autonomous-turn point: agreed the exemption is silent, but I left `connection.ex` alone rather than widen this PR. A reapply drops the connection, so there is no peer left to narrate an out-of-turn cycle, and a comment naming the one caller that takes the exemption is a better fit for a follow-up than for the PR you asked to change. Say the word if you would rather it land here.

`docs/reference/webhooks.md` and `docs/api.md` follow in #1893. 174 tests 0 failures across the reapply, controller, skills, audit, webhook and docs suites; `credo --strict` clean on the three changed modules.

BinaryBourbon (changes requested):
The caller-facing half is fixed, and fixed in the right place. I re-checked all three:

- `{:ok, updated}` comes back once the row is written, and **What `{:ok, conv}` promises** says what that does and does not cover. That is the answer I wanted and the docstring is better than the one I asked for.
- The failure branch is total and `refresh_reason/1` handles a non-`{:error, _}` shape. Agreed a `CaseClauseError` after the commit would have been the same bug one level down.
- The test asserts the row, the revision and the event. That is the invariant that was actually at stake, and its `done` counterpart is a good pairing.

The problem is that the lie moved rather than went away. It used to be told to the caller. It is now told to every follower.

### `configuration.done` is published when no machine has read the selection

`docs/api.md` in #1893 — the page this stack adds — defines the pair:

> The status is `done` when a machine holds the new selection, and `failed` when the selection stands and the machine has yet to read it.

and `reapply_conversation/3`'s new docstring says the same thing. `announce_reapply/2` does not implement it. It switches on `refresh_configuration/2`, whose `:ok` means *"no live server is left holding the previous selection"* — which includes **"there is no server, and nothing was rewritten anywhere."**

So the canonical target of this whole feature gets `done`: an idle conversation whose server has stopped. That is every suspended one (`suspend_sandbox/2` ends `{:stop, :normal, %{state | handle: nil}}`), every conversation on a node that has restarted since its last turn, and anything parked by 0017. Acceptance criterion one of #1565 is the idle conversation, so this is the path, not an edge of it.

Measured on `e9a92ef`, no server started, nothing stubbed, a real `reapply_conversation/3` moving `vault_id`:

```
stage: "configuration"   state: "done"
data: {"message":"The configuration was reapplied on this machine. The transcript and the
       files on disk are kept; the next prompt starts a new runtime session.",
       "changed_fields":["vault_id"], ...}
```

The sprite is parked. Nothing on it was rewritten. `failed` — "selected, and the machine has yet to read it" — is the true statement, and it is the statement you added the status for.

### `settled/1` makes it worse, and nothing tests it

This one is on me. I asked for `:not_running` to fold into success, and that ask was only ever about the **return value**. Suggestion 3 landed as well and returns `{:ok, updated}` for every refresh outcome, so `settled/1` no longer changes anything a caller sees. The only thing left for it to do is move `:not_running` off the truthful `failed` event and onto the false `done` one. Given 3, 1 is redundant at best — drop it.

It is also invisible to the suite. Nothing calls `ConversationServer.refresh_configuration/1,2`: `conversation_server_refresh_test.exs` goes at the handler with `GenServer.call(pid, ...)`, and `conversation_reapply_test.exs` replaces the whole function with Mimic. I deleted both `settled/1` clauses and re-ran the reapply, refresh, check and webhook files — **47 tests, 0 failures**. Whatever it ends up doing, it needs a test that would notice.

### What I would do

`refresh_configuration/2` is answering two different questions with one atom. Give it three answers — a live server reloaded it, there was no live server to reload, it refused or timed out — and publish `done` only for the first. `{:ok, :reloaded}` / `{:ok, :no_server}` / `{:error, reason}` is one shape; `announce_reapply/2` asking `whereis/1` for itself is another. Note there is a third "no machine was reconfigured" hiding in the handler: `handle_call({:refresh_configuration, _})` answers `:ok` when `state.handle` is nil, which `conversation_server_refresh_test.exs` pins as "a server holding no machine has nothing to do".

The caller contract does not move. `{:ok, conv}` for all of them, exactly as the docstring now says. The only change is which sentence the followers get, and #1893's page already says which sentence that should be.

### Non-blocking

- `announce_reapply/2`'s own reasoning — "this runs after the commit, so an unexpected shape here must become an event rather than a `CaseClauseError` that 500s a reapply which already happened" — applies to `publish_stage/4` too. It calls `log!/1`, which raises on an insert failure, straight out through a committed reapply. The totality guards the `case` and not what is inside it.
- The `configuration_revision` assignment in `provision_with_rows/3` still carries no comment saying its placement is load-bearing. Same hot-loop risk as last time, still one sentence.
- Autonomous-turn exemption: agreed, follow-up.

jhgaylor:
Right on all three, and the second one is a good catch on my own fix. Landed in ``97d77ea``.

### The lie did move, and the docs were the half that was right

Confirmed your measurement before changing anything. `refresh_configuration/2`'s `:ok` meant "no live server is left holding the previous selection", `announce_reapply/2` read it as "a machine holds the new selection", and those are not the same fact. The tell is that `docs/api.md` and the `reapply_conversation/3` docstring both already described the behaviour you wanted; the code was the half that disagreed, so **nothing in the manual changed this round.**

`refresh_configuration/2` now answers three questions separately:

- `{:ok, :reloaded}` — a live server has the new selection;
- `{:ok, :no_server}` / `{:ok, :no_machine}` — nothing to tell, so nothing was rewritten;
- `{:error, reason}` — a live server holds the previous selection and could not be told.

`announce_reapply/2` publishes `done` for the first only. You were right that there was a third "nothing was reconfigured" hiding in the handler: the `handle: nil` clause answered `:ok` too, and now answers `{:ok, :no_machine}`.

### `settled/1` is gone

Agreed, and agreed it was my misreading of your suggestion 1 rather than yours. Once suggestion 3 landed, the only thing it could still do was move `:not_running` off the truthful event onto the false one. Deleted.

### The coverage gap, closed and checked

You were right that nothing reached the function. Two additions:

- `conversation_server_refresh_test.exs` gains a test that calls `ConversationServer.refresh_configuration/2` itself. Since this harness's servers are deliberately outside Horde, `whereis/1` misses and the assertion is the real production answer for an idle conversation whose server has stopped.
- The context test you measured — `keeps the conversation, the transcript and the machine`, no server, nothing stubbed — now asserts `failed` and `reason == "no_server"`. That flip is the bug, pinned.

I checked these bite rather than assuming it. Restoring the conflation (`nil -> {:ok, :reloaded}`) and re-running gives **29 tests, 2 failures**, and they are exactly those two. Clean run is 48 tests 0 failures on the four affected files, and the whole `test/fountain/conversations/` tree plus the audit and webhook guards is green.

### Both non-blocking items

- `announce_reapply/2` is wrapped, so `publish_stage/4`'s `log!/1` cannot raise out through a committed reapply either. You were right that the totality guarded the `case` and not its body.
- The `configuration_revision` assignment in `provision_with_rows/3` now carries the sentence: one state assembly, both arms of `dispatch_provision/7` come through it, and in either arm instead the `kick_turn -> :reapply_prompt -> :provision` cycle would spin against the database.
- Autonomous-turn exemption: follow-up, as agreed.

### One observation, not an objection

`failed` is now the status for the most common reapply — an idle conversation with no live server — where nothing actually failed. I went looking for whether that misleads a subscriber and it mostly cannot: the filter shapes are an exact type, one stage (`conversation.configuration.*`), or `*`, so there is no `*.failed` to accidentally alert on across stages, and anyone naming the exact type has read what it means on the page. So the word is a little lossy for "nothing to tell" but cheap. Flagging it in case you would rather have a third status than a second meaning for `failed`; I did not add one, since the page you reviewed already defines the pair.

BinaryBourbon (approved):
This is right, and the three-way answer is cleaner than the shape I suggested — putting `:no_server` and `:no_machine` side by side in the `@spec` and the docstring makes the distinction a type rather than a convention.

I re-measured rather than reading it, because the thing I reported was a claim about behaviour and not about code.

**The event flipped.** Same probe as before on `97d77ea`: a real `reapply_conversation/3` moving `vault_id`, no server started, nothing stubbed. `state: "failed"`, `reason: "no_server"`, and the message now says "No machine has read it yet" instead of asserting one did. The caller still gets `{:ok, updated}`.

**The guard guards.** Restoring the conflation — `nil -> {:ok, :reloaded}` and nothing else — and re-running the reapply and refresh files gives **29 tests, 2 failures**, and they are exactly `keeps the conversation, the transcript and the machine` and `with no server registered, the public entry point says so`. Your number, reproduced, and both new assertions bite. That is the part I would have come back about if only the code had moved.

**`:no_machine` was the third one.** `handle_call(:refresh_configuration, _from, %{handle: nil})` answered `:ok` before and answers `{:ok, :no_machine}` now, with the handler test pinning it. Nothing on disk was rewritten there either, so it belongs on the same side of the line.

Also checked: `announce_reapply/2` is wrapped as a whole, so `publish_stage/4`'s `log!/1` can no longer raise out through a committed reapply — that was the gap between your comment and what it covered. `refresh_reason({:ok, reason})` is what makes `no_server` reach the payload as a word. The `configuration_revision` placement sentence landed and says the two things that matter: one assembly, both arms of `dispatch_provision/7`. `credo --strict` finds nothing on either module, `mix format --check-formatted` is clean, and the reapply/refresh/check/webhook/audit set is 122 tests 0 failures here.

### One clause up, the same conflation survives

```elixir
def handle_call({:refresh_configuration, revision}, from, state) do
  if revision == state.configuration_revision,
    do: {:reply, {:ok, :reloaded}, state},
    else: handle_call(:refresh_configuration, from, state)
end
```

That fast path answers `:reloaded` without consulting `handle`, and it sits **above** the `%{handle: nil}` clause, so it outranks it. A server that already holds the committed revision and holds no machine reports that a machine read the selection.

The window is narrow — the server has to have loaded the row in the microseconds between the commit and this call, and such a server is usually provisioning from that very revision, where the call times out into `{:error, :provisioning}` instead. So this is a nit, not a repeat of the bug. But it is the same mistake shape one clause up, and the fix is to fall through to the clauses below rather than reply from here; they answer `:reloaded` for a server with a machine anyway.

### On your observation, since you asked

I would take the third status, and I would spend the rebase on it now.

Your filter analysis is right as far as it goes — there is no cross-stage `*.failed` to catch this by accident. But `conversation.configuration.failed` is an entirely ordinary filter for somebody to set, and the *normal* outcome of this endpoint now carries it: reapply an idle conversation, which is what the feature is for, and a subscriber watching for configuration failures gets woken. "Nothing failed" and "a machine could not be told" are as different as the two facts you just separated one layer down, and `failed` is the one word in this vocabulary anybody will page on.

The change is small and it is smaller now than later: one entry in `Events`' `configuration` row, the status word in
`docs/reference/webhooks.md` and in the sentence I approved in `docs/api.md`, and the message text already says the
right thing. A stage-specific word is not a new idea in that catalogue either — `interrupted` is already one, carried
by `turn` and `reattach` and by nothing else. An event type is a wire contract — renaming it after the first subscriber exists is the expensive version of this conversation.

Your call, and the approval does not hang on it. If you leave it, leave it deliberately rather than by running out of rebases.

### One thing this PR causes further up, which is not this PR's to fix

These 13 lines take `conversation_server.ex` to 2,735. That is fine here — the pin on this branch is still `main`'s 2,774, and I ran `conversation_server_size_test.exs` at `97d77ea` to be sure: 1 test, 0 failures.

#1891 lowers the pin to 2,716, which was the file's exact length when that PR was written. It has no headroom, so every line this PR has grown since — 10 in `e9a92ef`, 13 in `97d77ea` — lands on it. At the stack tip the file is 2,739 and that test fails. Raised on #1891, since the pin is its edit; noting it here because this is where the lines come from, and because any further fix to the server in this PR moves the number again.

PR #1891, opened 2026-09-11
feat(sandboxes): reconcile skills rather than only install them (#1565)

jhgaylor:
Installing was enough while a conversation's agent was fixed at launch. It is not once the agent can be reapplied: a skill the agent no longer names is still on the disk and the runtime still reads it, so removing one from an agent changed nothing until the machine was rebuilt.

`SandboxSkills.reconcile/4` deletes what Fountain put under the skills root and no longer selects, and only that. `mount/3` is now the empty-history call to it, so a fresh provision is unchanged.

Three things make "only that" true:

- **A manifest at the skills root** maps each selected skill to the directories its install produced. A GitHub install without `--skill` picks its own names, so the manifest records what appeared rather than what was asked for.
- **A seed for disks written before the manifest existed**: the conversation's recorded Agent version names the skills that were installed, and skills.sh's own source lock names the directories an unnamed GitHub install produced. An entry neither can account for is somebody else's and is left where it is.
- **Only direct children of the skills root are removed**, each matched against a conservative name pattern, so neither a forged manifest nor a legacy name can turn reconciliation into a delete elsewhere on the disk.

A remote skill that is still selected keeps its directory when the reinstall fails, which is what an offline machine under a restrictive network policy looks like: the copy on the disk is the working one.

`Reapply.mount_skills/3` runs it on every wake, beside the `.env`, `.mcp.json` and instructions rewrites the wake path has always done, and records the new selection only when the pass succeeded. So a sleeping conversation applies a reapply when it next comes up, and a live one applies it when its server reloads.

The reconciliation tests run against a real directory rather than a mock, so "preserves unrelated files" is a statement about a filesystem.

### The size pin

The pin stays at main's value while the stack is in flight. This PR changes only the test's documentation; lowering the pin is a follow-up after the stack lands, when the server's size has stopped moving.

### Difference from #1569

The original resolved a runtime string through `Managoat.Runtimes.for_runtime/1`. This one goes through `Fountain.RuntimeDispatch`, for the reason #1634 gives: the library knows four runtimes and `acp` is not one of them, so the library's dispatcher would have left an acp sandbox with no skills and said nothing.

Part 6 of 8 for #1565, on #1890.

---

| | PR | What it covers |
|---|---|---|

Part of #1565

### Queue preparation

Restacked onto current `main`. The skills implementation is unchanged (`git range-diff`); the size test passes its structural check at 2,689 lines against a 2,762-line pin. Queue before #1892 and #1893.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -1150,6 +1150,10 @@ defmodule Fountain.Conversations.ConversationServer do
         Fountain.Conversations.Identity.disk_env(sprite_env)
       )
 
+      # An agent's skills reach the existing computer on its next wake too,
+      # and a skill it no longer names is taken off the disk (#1565).
+      Reapply.mount_skills(handle, conv, agent)
+
       # Same for the agent's system prompt: an edit reaches the existing
       # computer on its next wake (#848).
       runtime = conv.runtime || (agent && agent.runtime) || "claude"
--- a/apps/fountain/lib/fountain/conversations/reapply.ex
+++ b/apps/fountain/lib/fountain/conversations/reapply.ex
@@ -247,4 +247,33 @@ defmodule Fountain.Conversations.Reapply do
       version -> version.config["skills"] || []
     end
   end
+
+  @doc """
+  Reconcile the machine's skills with the conversation's current selection,
+  then record what is now on it.
+
+  Run on every wake, not only after a live reapply: a sleeping conversation
+  whose selection changed applies it when it next comes up, and a machine
+  built before the manifest existed gets one on its first pass. The recorded
+  set is only advanced when the reconciliation succeeded, so a failed pass
+  leaves the next one the same work rather than a wrong picture of the disk.
+  """
+  @spec mount_skills(Managoat.Sandbox.Handle.t(), map(), map() | nil) :: :ok | {:error, term()}
+  def mount_skills(handle, conv, agent) do
+    # ownership: conv came from the tenant-scoped API fetch or its own server.
+    sandbox = Conversations._unsafe_get_sandbox!(conv.sandbox_id)
+    skills = (agent && agent.skills) || []
+    runtime = conv.runtime || (agent && agent.runtime) || "claude"
+
+    with :ok <-
+           Fountain.SandboxSkills.reconcile(
+             handle,
+             runtime,
+             skills,
+             sandbox.applied_skills || previous_skills(conv)
+           ),
+         {:ok, _} <- Conversations.update_sandbox(sandbox, %{applied_skills: skills}) do
+      :ok
+    end
+  end
 end
--- a/apps/fountain/lib/fountain/sandbox_skills.ex
+++ b/apps/fountain/lib/fountain/sandbox_skills.ex
@@ -14,6 +14,9 @@ defmodule Fountain.SandboxSkills do
   require Logger
 
   @bundle_root "sprite_skills"
+  # Written at the skills root: which directories each Fountain-managed skill
+  # put there, so a later reconciliation knows what it owns.
+  @manifest_name ".fountain-managed-skills"
   @fountain_skill_name "fountain"
   # Every sandbox gets these, in this order: the API skill, then the team
   # set-up Q&A (#851) — so a first teammate can answer "/create-team".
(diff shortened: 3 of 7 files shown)


BinaryBourbon (approved):
Approving. This is the PR in the stack that runs `rm -rf` against a tenant's disk driven by a manifest read off that same disk, so I audited the deletion path specifically rather than reviewing it like the others.

**The containment argument holds.** Every name that can reach `remove/2` passes `safe_name?/1` — `\A[A-Za-z0-9][A-Za-z0-9._-]*\z` — on all four routes in: `decode_manifest/1` filters the on-disk manifest, `legacy_manifest/2` filters the recovered skills.sh names, `entries/1` filters the listing diff, and `names/1` filters the agent's own skill names. That regex rejects anything containing `/` and anything starting with `.`, so `..`, `.`, absolute paths and traversal are all out, and `Path.join(root, name)` cannot escape the root. `quote_shell/1` does the single-quote doubling correctly. A forged manifest can name a sibling skill directory but nothing outside the root, which is the property the docstring claims.

Two consequences of that regex I checked and liked: `.fountain-managed-skills` starts with `.`, so the manifest can never appear in a listing diff and be attributed to a skill, and can never be selected for deletion. And `obsolete_names/2` subtracting `retained ++ wanted` means a directory name two skills both produce is never deleted while either is still selected.

**`mount/3` delegating to `reconcile(handle, module, skills, [])` is safe on a fresh machine** — `previous` is empty, the manifest file does not exist, `managed` is `%{}`, so `obsolete_names/2` returns `[]` and nothing is removed on a first provision. Good that the two paths converged rather than staying separate.

**The retained-on-failure behaviour is the right call.** Keeping a remote skill's directory when its reinstall fails, and folding `Map.get(managed, identity(skill), [])` back into the recorded names, is what stops an offline machine under a restrictive egress policy from deleting a working copy on the next pass. That is the failure mode I went looking for and it is already handled.

Three non-blocking notes:

- **The return value is discarded at the call site.** `Reapply.mount_skills(handle, conv, agent)` in the reattach path is not matched, so a reconciliation that fails on a wake is invisible — no log line, no stage event. The "recorded set only advances on success" design means it self-heals on the next wake, which is why this is not blocking, but a `case`/`Logger.warning` would cost nothing and would turn "my removed skill is still active" from a mystery into a grep. `reconcile/4` already logs when the *runtime* cannot be resolved; the other failure modes are silent.
- **Per-skill installs replace one batched call.** `mount/3` used to be a single `Managoat.Runtimes.Skills.install(handle, bundled() ++ skills, ...)`; `install_selected/5` now invokes the library once per skill, plus two `listing/1` execs around each remote one. On a provision with several GitHub skills that is a noticeable number of round trips added to the path a first turn waits on. Probably fine, and the per-skill attribution is what makes the manifest possible — worth measuring once against a realistic agent rather than assuming.
- **`identity/1` keys the manifest on `[source, name]`**, so two selected skills sharing both collide into one entry. Not reachable with any sane agent config, and the failure is benign (one entry's names recorded under the other), but it is an assumption worth a line.

The `sandbox_skills_test.exs` additions cover the cases I would have asked for, including the legacy-recovery path and the "leave what you do not own" case.

jhgaylor:
Rebased only, no content change. The parent moved for the review fixes in #1888 (`4b7b5d2`), #1889 (`ff0bb29`), #1890 (`e9a92ef`) and #1893 (`5c1e48d`); this branch's own commit is unchanged apart from its new base.

BinaryBourbon (approved):
Rebase verified, not taken on trust: `git diff stack/1565-5-revision..stack/1565-6-skills` against the diff I reviewed at `6f8d1b4` differs in exactly two lines, both `@@` hunk headers whose offsets moved because the parents grew (`conversation_server.ex` 1122→1132, `reapply.ex` 231→247). No content change.

Approval stands.

jhgaylor:

BinaryBourbon (changes requested):
Withdrawing the approval: this PR is red at the stack tip, and the cause is its own edit to the pin.

```
apps/fountain/lib/fountain/conversations/conversation_server.ex is 2739 lines,
over the pin of 2716. The server only shrinks (#1369): move the new code into a
Fountain.Conversations.* module rather than raising the pin.
```

`mix test test/fountain/conversations/conversation_server_size_test.exs` at `fae75c1`, and at #1892 and #1893 which inherit it. The post-rebase CI runs were still queued when I looked, so this is ahead of them rather than a report of them.

### Why, and why it was green when I approved it

`@pin 2716` was the file's exact length at the time — I checked, and at `6f8d1b4` the count was 2716 against a pin of 2716. Zero headroom. Every line added to `conversation_server.ex` anywhere **below** this PR in the stack therefore breaks it:

| | count | pin | |
|---|---|---|---|
| `6f8d1b4` (approved, round 1) | 2716 | 2716 | ok, exactly |
| `e9a92ef` (#1890 review fixes, round 1) | 2722 | 2716 | over by 6 |
| `97d77ea` (#1890 review fixes, round 2) | 2735 | 2716 | over by 19 |
| `fae75c1` (here, +4) | 2739 | 2716 | **over by 23** |

This is mine to own as much as yours. At the second round I verified this branch was rebase-only by diffing its patch against the one I had already read, found two hunk headers of difference, and said the approval stands. A byte-identical patch is not a still-green branch when the patch contains an absolute number measured against the rebased file. I will not take "rebase only" as "nothing to re-check" on a pin, an allowlist or a line count again.

### What I would do

Leave the pin alone in this stack. `main` is at 2,774 and the tip of the stack is 2,739, so the whole stack is already a 35-line reduction and `@pin 2774` passes every link of it without touching the test. Lowering it is a one-line follow-up once the stack has landed and the number has stopped moving.

If you would rather keep the lowering here, 2,739 is today's correct value — but it is correct only until the next fix below it, and there have been two rounds of those. The test's own docstring says the pin is "the file's line count on `main` at the last move", which argues for the follow-up: nothing in this stack is on `main` yet.

Either way it wants a sentence saying why the number is what it is, because the next person to review a stack that grows the server will hit exactly this.

Nothing else in the PR has changed and nothing else needs to.

jhgaylor:
Confirmed and fixed in `400afeb`. I reproduced it before changing anything rather than taking the table on faith:

```
apps/fountain/lib/fountain/conversations/conversation_server.ex is 2739 lines,
over the pin of 2716.
1 test, 1 failure
```

and your counts are exact at every sha — `6f8d1b4` 2716/2716, `e9a92ef` 2722, `97d77ea` 2735, `fae75c1` 2739.

Taken your recommendation: **the pin goes back to 2,774**, main's value and the number the test's own docstring describes ("the file's line count on `main` at the last move"). Nothing in this stack is on `main` yet. I checked 2,774 passes at all eleven commits rather than only at the tip:

| | lines | |
|---|---|---|
| `main` | 2774 | pass |
| `c130228` … `ff0bb29` | 2627–2636 | pass |
| `762c258` | 2712 | pass |
| `e9a92ef` / `97d77ea` | 2722 / 2735 | pass |
| `400afeb` (tip of this PR) | 2739 | pass |

The stack is still a net reduction, 2,774 down to 2,739. Lowering it is a follow-up once the shrink is on `main`, exactly as you say — doing it here would be correct only until the next fix below it, and there have been two of those already.

The moduledoc gets the sentence you asked for, generalised past this stack since the next person will hit the same thing:

> **A stack in flight does not move it.** The pin is the length on `main`, and a number measured against a branch is stale the moment any link below it grows the file, which is what review rounds do. #1565 lowered it mid-stack to the tip's exact length, left zero headroom, and went red two rounds later when a fix four PRs down added lines. Land the shrink first, then lower the pin in a follow-up, when the number has stopped moving.

I have also corrected **#1886's body**, which promised "#1891 lowers it once, to 2,716" and was the origin of the plan. It now records what actually happened and where the lowering went.

On your own note about the round-two approval: the same applies to me on the other side. I described those pushes as "rebased only, no content change" on #1891, #1892 and #1893, which was true of the patches and false of the tree, because the thing that broke was a number measured against a file those PRs do not touch. A rebase note is not a claim that the branch still passes.

89 tests 0 failures at the stack tip across the size, skills, refresh, reapply, controller, docs and webhook suites. Nothing else in the PR changed.

BinaryBourbon (approved):
Fixed, and the fix is smaller than the bug — which is the right shape for this.

Re-checked rather than reading the table:

- **The pin is back to main's 2,774 and 1891's net diff no longer touches the number.** `git diff 97d77ea..400afeb` on that file is the moduledoc paragraph and nothing else.
- **It passes at every commit, not only at the tip.** I recomputed pin-versus-length at all fourteen commits from `c130228` to `b8d8776`: every one passes except `fae75c1`, which this PR's second commit supersedes. `mix test test/fountain/conversations/conversation_server_size_test.exs` at `b8d8776`: 1 test, 0 failures.
- **#1892 and #1893 are byte-identical rebases** of the diffs I approved, hunk headers included, so the red on both is gone with nothing else moved.
- **The tree is green, not just that one test.** The same broad run that found the pin last round — the whole
  `conversations/` tree plus `webhooks/`, `sandbox_skills_test.exs`, the conversation controller, `docs_test.exs` and the
  audit guardrail, at `b8d8776`: **1,221 tests, 0 failures, 2 skipped**.
- **#1886's body is corrected properly** — the old plan struck through, the correction dated, and it says where the lowering went instead of quietly deleting the sentence that caused this.

The moduledoc paragraph is the part I value most here. Generalising it past #1565 was the right call; "a number measured against a branch is stale the moment any link below it grows the file, which is what review rounds do" is the whole lesson in one sentence, and it is in the file that enforces the rule rather than in a PR body nobody will reopen.

And yes — symmetric. "Rebased only, no content change" and "the approval stands" were both true of the patches and both silent about the tree. The patch is not the unit that passes CI.

### Three small things, none blocking

**The moduledoc now describes a lowering that will not be in `main`.** It says #1565 "lowered it mid-stack to the tip's exact length, left zero headroom, and went red two rounds later". All of that happened, but it happened on a branch and this PR backs it out, so a reader who goes to `git log` for it finds nothing. Six words fix it — "… and went red two rounds later; it was backed out before merge" — and the lesson survives intact. Worth it precisely because the paragraph is written for someone who will trust it.

**The paragraph above the new one still says the thing that caused this.** The opening still reads "each sub-issue moves a function family into a module under `Fountain.Conversations.*` and lowers `@pin` to the file's new length in the same PR" — which is the instruction #1886 followed, and which the new paragraph now contradicts without amending. #1886 deliberately does not lower it, so as written the docstring tells a reader to do the thing four lines before telling them not to. A clause on the first one ("in the PR that lands on `main`; see below for a stack") reconciles them.

**`fae75c1` is still in this PR's history as a commit where the test fails.** Harmless under a squash. If this stack is landed the way the others have been — first PR squashed, the rest merged with `--merge` — that commit reaches `main` and `git bisect` over this test hits it. Squashing #1891, or folding `400afeb` in as a fixup, costs nothing and removes the trap.

Approving.

PR #1892, opened 2026-09-11
feat(api): POST /api/conversations/:id/reapply (#1565)

jhgaylor:
The door onto the context function the stack has been building. An empty body reapplies the current selection; `agent_id`, `environment_id` and `vault_id` change one each; an omitted field keeps what is selected and an explicit null clears the Environment override or the Vault.

## The status codes

This is the decision worth reviewing on its own, because four different things can go wrong and they mean four different things to a client.

| Status | Body | When | Why not something else |
|---|---|---|---|
| `409` | `conversation_busy` | a turn is running | not 503: waiting does not help by itself, the caller waits for the turn to end or interrupts it |
| `409` | `rebuild_required` + `field` | the machine cannot be reconfigured into this selection | not 422: nothing about the request is malformed, the selection is perfectly valid |
| `503` | `provisioning` + `Retry-After` | the machine is still being built | here retrying is exactly right, which is why it is not one of the 409s |
| `410` | `conversation_terminated` | the conversation has ended | the id was real and there is nothing to come back for |

422 keeps its usual meaning: a selection the agent's allowlists forbid.

The `field` on `rebuild_required` is what lets a client say "your agent runs a different runtime" rather than "something needs a rebuild", and the message names the two ways forward (a new conversation, or `DELETE /api/sandboxes/:id`).

Both new refusals live in `FallbackController` rather than the controller, because `:conversation_busy` is a fact about a conversation that other doors will want to report the same way.

## Gates

The operation declares every status it can render, 403 included. The contract and the TypeScript types are regenerated, never hand-edited. `Conversation.reapply()` is the hand-written method that claims the operation so `scripts/sdk-contract/build.sh --check` passes; its test asserts the one thing types cannot, that an omitted field and an explicit null reach the wire differently.

**This PR changes `sdk/typescript/src/generated/openapi.ts` without bumping the version, so it carries `sdk-no-release`.** The bump is #1893.

Part 7 of 8 for #1565, on #1891.

---

| | PR | What it covers |
|---|---|---|

Part of #1565

### Queue preparation

Targets `main` for the merge queue and retains #1891 as its ancestor. Queue after #1891 and before #1893. The API patch is unchanged by the restack (`git range-diff`); generated OpenAPI and contract artifacts match regeneration against the updated server.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -1150,6 +1150,10 @@ defmodule Fountain.Conversations.ConversationServer do
         Fountain.Conversations.Identity.disk_env(sprite_env)
       )
 
+      # An agent's skills reach the existing computer on its next wake too,
+      # and a skill it no longer names is taken off the disk (#1565).
+      Reapply.mount_skills(handle, conv, agent)
+
       # Same for the agent's system prompt: an edit reaches the existing
       # computer on its next wake (#848).
       runtime = conv.runtime || (agent && agent.runtime) || "claude"
--- a/apps/fountain/lib/fountain/conversations/reapply.ex
+++ b/apps/fountain/lib/fountain/conversations/reapply.ex
@@ -247,4 +247,33 @@ defmodule Fountain.Conversations.Reapply do
       version -> version.config["skills"] || []
     end
   end
+
+  @doc """
+  Reconcile the machine's skills with the conversation's current selection,
+  then record what is now on it.
+
+  Run on every wake, not only after a live reapply: a sleeping conversation
+  whose selection changed applies it when it next comes up, and a machine
+  built before the manifest existed gets one on its first pass. The recorded
+  set is only advanced when the reconciliation succeeded, so a failed pass
+  leaves the next one the same work rather than a wrong picture of the disk.
+  """
+  @spec mount_skills(Managoat.Sandbox.Handle.t(), map(), map() | nil) :: :ok | {:error, term()}
+  def mount_skills(handle, conv, agent) do
+    # ownership: conv came from the tenant-scoped API fetch or its own server.
+    sandbox = Conversations._unsafe_get_sandbox!(conv.sandbox_id)
+    skills = (agent && agent.skills) || []
+    runtime = conv.runtime || (agent && agent.runtime) || "claude"
+
+    with :ok <-
+           Fountain.SandboxSkills.reconcile(
+             handle,
+             runtime,
+             skills,
+             sandbox.applied_skills || previous_skills(conv)
+           ),
+         {:ok, _} <- Conversations.update_sandbox(sandbox, %{applied_skills: skills}) do
+      :ok
+    end
+  end
 end
--- a/apps/fountain/lib/fountain/sandbox_skills.ex
+++ b/apps/fountain/lib/fountain/sandbox_skills.ex
@@ -14,6 +14,9 @@ defmodule Fountain.SandboxSkills do
   require Logger
 
   @bundle_root "sprite_skills"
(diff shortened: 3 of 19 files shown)


BinaryBourbon (approved):
Approving. I walked this against the seven gates a new `/api` route has to clear, because that list is where routes usually bleed:

1. `operation(...)` declared on a controller that already uses `OpenApiSpex.ControllerSpecs` — visible to `api_spec_test`. ✔
2. Every status the route can render is declared, `forbidden:` included, so nothing lands in `SchemaGuardAllowlist`. ✔
3. `ConversationReapplyRequest` has no `:enum`, so no `@derived` row needed. ✔
4. `sdk/contract/contract.json` regenerated. ✔
5. `sdk/contract/manifests/typescript.json` claims the operation, so `--check` has its claimant and no `omissions.json` entry is needed. ✔
6. `sdk/typescript/src/generated/openapi.ts` regenerated, and the hand-written method is separate from it. ✔
7. `docs/api.md` lands in #1893 with the prose gates. ✔

The status-code contract is the part I would have argued about and I agree with where it came out: 409 rather than 422 for `rebuild_required`, because the selection is valid and only this machine cannot take it; 409 rather than 503 for `conversation_busy`, because waiting is not what resolves it. Both new fallback clauses carry the reasoning in a comment, which is the house style and genuinely useful here.

I also confirmed the unmapped-atom safety net catches the refusals this route can produce that have no clause of their own — `:not_running` renders as `422 {"error":"not_running"}` with a logged warning rather than a 500. That is the net doing its job, but it is also the symptom of the #1890 problem, so see that review.

The TypeScript `reapply()` gets the three-state body right — `!== undefined` rather than truthiness, so `environmentId: null` serialises as `null` and an omitted key stays omitted. That is the one thing this API's semantics depend on and it is easy to write as `if (options.environmentId)`.

Two nits, neither blocking:

- **The controller forwards the whole `params` map as `attrs`**, including `"conversation_id"`. `reapply_value/3` only reads three keys so nothing leaks, but `Map.take(params, ~w(agent_id environment_id vault_id))` would make the contract at the door match the contract in the schema, and would stop a future field in `Conversation` from becoming settable by accident if `reapply_value/3` ever grows a loop over keys.
- **`forbidden:` is declared but I could not find what enforces it** for this action specifically — it looks inherited from the pipeline rather than from anything in `reapply/2`. If that is right it is correct and I am only noting that the declaration and the mechanism are far apart; if sprite keys are meant to be refused here by something route-specific, it is not in this diff.

jhgaylor:
Rebased only, no content change. The parent moved for the review fixes in #1888 (`4b7b5d2`), #1889 (`ff0bb29`), #1890 (`e9a92ef`) and #1893 (`5c1e48d`); this branch's own commit is unchanged apart from its new base.

BinaryBourbon (approved):
Rebase verified: `git diff stack/1565-6-skills..stack/1565-7-api` is byte-identical to the diff I reviewed at `7a333fc`, hunk headers included. No content change.

Approval stands.

One thing to carry into the merge, not a change here: #1890 still publishes `configuration.done` when nothing read the selection, and this is the PR that renders the 409 vocabulary the `field` travels in. Nothing to do until that lands.

jhgaylor:

BinaryBourbon:
Heads-up rather than a change request: the `conversation_server_size_test.exs` failure you will see here is inherited, not yours. `@pin 2716` comes from #1891 and the file is 2,739 lines at this commit. Nothing in this PR touches `conversation_server.ex`. Raised on #1891; approval here stands.

jhgaylor:
Rebased only — the parent moved for the size-pin fix in #1891 (`400afeb`). No change to this PR's own commit. Note the pin fix is what makes this branch green again; it was red on `conversation_server_size_test.exs` inherited from #1891.

PR #1893, opened 2026-09-11
docs: reapply a conversation's configuration, and the four SDKs (#1565)

### LINKED ISSUE #1565: Conversations: reapply or change the Agent, Environment and Vault without starting a new thread

jhgaylor:
The manual and the three SDKs #1892 left, plus the one version bump the stack makes.

`docs/api.md` gets the endpoint, the three-state body, what moves with it (the machine's binding identity, the skills reconciliation, the revision) and every refusal by status. `docs/concepts/conversation.md` gets the short version beside prompts and interrupts, and `docs/reference/webhooks.md` gets the `configuration`/`done` stage.

Python, Elixir and Swift get the method, each in its own idiom for the one thing the wire format needs: an omitted field and an explicit null are different requests. Python uses an `_UNSET` sentinel, Elixir reads `Keyword.has_key?`, and Swift has a three-case `ConversationBindingUpdate` rather than a plain optional that cannot tell "leave it" from "remove it". Each SDK's test asserts the encoded body, not just the call.

Versions: TypeScript 1.30.0 → **1.31.0**, Python 0.1.1 → 0.2.0, Elixir 0.1.0 → 0.2.0. Swift is unversioned.

The TypeScript bump is re-cut on top of 1.30.0 on `main`, preserving the intervening OAuth, sandbox-queue, and detached-permission releases.

> [!IMPORTANT]
> **Re-check the TypeScript version against `main` immediately before merge.** Concurrent bumps must land in ascending order, and this one has already been re-bumped once for exactly that reason. Merging a bump publishes, and a publish moves the `latest` tag.

Checked: `docs-style.py` (two pre-existing findings, neither on a page this touches), `vale lint` (no new hit on any of the six gating rules), `destink` (112 pages clean), `docs_test.exs`, and all four SDK suites and contract checks including `swift test` (82 tests).

## What this stack carries from #1569, and what it does not

Everything #1569 decided is here, re-derived against today's code. Two deliberate departures, both noted on the PR that makes them:

- **The shared attach/reapply transaction** (#1889). `main`'s `create_attached_conversation/3` already rechecks `check_attachable/4` under `FOR SHARE` on the sandbox row, which a reapply's `update_sandbox` conflicts with, so the attach path needs no change and the lock helper is private to `Conversations`.
- **Runtime resolution in the skills reconciliation** (#1891) goes through `Fountain.RuntimeDispatch`, not the library's own dispatcher, for the reason #1634 gives.

One thing #1569 got wrong and this does not: its docs and SDK doc comments said a reapply moves the conversation to "a fresh machine", which contradicts the change itself. The machine is kept; that is the whole point.

Out of scope in #1569 and still out of scope here: the conversations UI, which lives in its own repository.

Part 8 of 8 for #1565, on #1892.

---

| | PR | What it covers |
|---|---|---|

Closes #1565

### Queue preparation

Targets `main` for the merge queue and retains #1891 and #1892 as ancestors. Queue after #1892. TypeScript is now 1.31.0, above main's 1.30.0; Python and Elixir remain 0.2.0. The restack preserves the reviewed feature patches and intervening changelog entries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,16 @@ upgrade, is in
   Connections off is the broker: an account is offered the feature only while
   `BROKER_TENANTS` names it.
 
+### Added
+
+- `POST /api/conversations/:id/reapply` re-selects a conversation's Agent,
+  Environment and Vault on the machine it is already running, keeping the
+  conversation, its transcript and the files on its disk. A selection that would
+  need the machine built again is refused, naming what forced it. Adds the
+  `configuration` webhook stage and three columns
+  (`conversations.configuration_revision`, `sandboxes.build_fingerprint`,
+  `sandboxes.applied_skills`) (#1565).
+
 ### Changed
 
 - **The claude runtime's ACP adapter moves to 0.75.1, and a fresh sandbox now
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -1150,6 +1150,10 @@ defmodule Fountain.Conversations.ConversationServer do
         Fountain.Conversations.Identity.disk_env(sprite_env)
       )
 
+      # An agent's skills reach the existing computer on its next wake too,
+      # and a skill it no longer names is taken off the disk (#1565).
+      Reapply.mount_skills(handle, conv, agent)
+
       # Same for the agent's system prompt: an edit reaches the existing
       # computer on its next wake (#848).
       runtime = conv.runtime || (agent && agent.runtime) || "claude"
--- a/apps/fountain/lib/fountain/conversations/reapply.ex
+++ b/apps/fountain/lib/fountain/conversations/reapply.ex
@@ -247,4 +247,33 @@ defmodule Fountain.Conversations.Reapply do
       version -> version.config["skills"] || []
     end
   end
+
+  @doc """
+  Reconcile the machine's skills with the conversation's current selection,
+  then record what is now on it.
+
+  Run on every wake, not only after a live reapply: a sleeping conversation
+  whose selection changed applies it when it next comes up, and a machine
+  built before the manifest existed gets one on its first pass. The recorded
+  set is only advanced when the reconciliation succeeded, so a failed pass
+  leaves the next one the same work rather than a wrong picture of the disk.
+  """
+  @spec mount_skills(Managoat.Sandbox.Handle.t(), map(), map() | nil) :: :ok | {:error, term()}
+  def mount_skills(handle, conv, agent) do
+    # ownership: conv came from the tenant-scoped API fetch or its own server.
+    sandbox = Conversations._unsafe_get_sandbox!(conv.sandbox_id)
+    skills = (agent && agent.skills) || []
+    runtime = conv.runtime || (agent && agent.runtime) || "claude"
+
+    with :ok <-
+           Fountain.SandboxSkills.reconcile(
+             handle,
+             runtime,
+             skills,
+             sandbox.applied_skills || previous_skills(conv)
+           ),
+         {:ok, _} <- Conversations.update_sandbox(sandbox, %{applied_skills: skills}) do
+      :ok
+    end
+  end
 end
--- a/apps/fountain/lib/fountain/sandbox_skills.ex
+++ b/apps/fountain/lib/fountain/sandbox_skills.ex
@@ -14,6 +14,9 @@ defmodule Fountain.SandboxSkills do
   require Logger
 
   @bundle_root "sprite_skills"
+  # Written at the skills root: which directories each Fountain-managed skill
+  # put there, so a later reconciliation knows what it owns.
+  @manifest_name ".fountain-managed-skills"
   @fountain_skill_name "fountain"
   # Every sandbox gets these, in this order: the API skill, then the team
   # set-up Q&A (#851) — so a first teammate can answer "/create-team".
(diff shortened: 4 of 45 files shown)


jhgaylor:
Verified on the stack tip after the rebase onto current `main`:

- core suite: **6 doctests, 4,975 tests, 0 failures, 2 skipped**
- `mix credo --strict`: no issues; `mix format --check-formatted`: clean
- TypeScript 106 tests + typecheck + verify-contract; Python 19 + verify-contract; Elixir SDK 52 + contract; Swift 82
- `scripts/sdk-contract/build.sh --check`: ok (165 operations, 191 schemas); `npm run generate` produces no drift
- `docs-style.py` (two pre-existing findings, neither on a page this stack touches), `vale lint`, `destink` (112 pages clean)

Two regressions the full suite caught that no targeted run did, both from the subtraction in #1886 and both fixed there:

1. **Mimic cannot intercept a call a module makes to itself.** Folding the checkpoint family into `Provisioning` meant `attempt_warm_start/3` called `restore_checkpoint/2` as a self-call, which took two warm-start tests off their stub and onto the real Sprites client. Fixed by giving the family its own module, `Fountain.Conversations.Checkpoints`.
2. **The webhook catalogue guard reads `publish_stage` call sites out of a named file list.** Moving the `checkpoint_restore` stage out of `ConversationServer` made three catalogue entries read as stale. Fixed by adding `checkpoints.ex` to that list.

jhgaylor:

| PR | stack | branch |
|---|---|---|

Instead:

BinaryBourbon (approved):
Approving. The docs read well in STE and say the things a reader actually needs — that the machine stays, that a sleeping conversation applies the change on its next wake, that an entry with no ownership record is left alone, and that `DELETE /api/sandboxes/{id}` is the escape hatch when a rebuild is genuinely needed. `docs/reference/webhooks.md` gets the `configuration`/`done` row to match `Fountain.Webhooks.Events`.

Making the three-state body explicit in each SDK's own idiom — Python `_UNSET`, Elixir `Keyword.has_key?`, Swift's three-case `ConversationBindingUpdate` — is the right amount of care. A plain optional in Swift genuinely cannot express "leave it" versus "remove it", and asserting the encoded body rather than the call is what makes those tests worth having.

### The TypeScript 1.28.0 collision is real — here is the list

Your `[!IMPORTANT]` note is correct to be worried. Concretely, **four open PRs currently claim `1.28.0`**:

| PR | Tracker |
|---|---|
| #1893 | this one (#1565) |
| #1885 | #1125, tenant-registered OAuth clients |
| #1876 | #1033, bounded capacity queue |
| #1860 | #1635, requests that outlive a turn |

`main` is at `1.27.0`. Whichever lands first takes `1.28.0` and the other three need re-bumping. They will conflict in `package.json` and `CHANGELOG.md`, so git will stop a silent double-publish — but the conflict lands on whoever merges second, and a bump is four edits (`package.json`, `package-lock.json`, `USER_AGENT` in `src/http.ts`, the `CHANGELOG.md` heading), so it is easy to resolve three of four and ship a `USER_AGENT` that disagrees with the published version.

Python `0.1.1 → 0.2.0` and Elixir `0.1.0 → 0.2.0` have no competitors among the open PRs, so those two are uncontested.

### One thing missing

**The root `CHANGELOG.md` gets no `[Unreleased]` entry.** This stack ships a new public endpoint, a migration that adds three columns, and a new webhook stage — all three are things a self-hoster reads that file to find before taking an upgrade. `CHANGELOG.md` is itself a published docs page, so the entry is user-facing rather than bookkeeping.

Not blocking the approval, because this is the natural PR to add it to and it is two lines, but please add it before the stack merges rather than after. Something like:

```markdown
### Added

- `POST /api/conversations/:id/reapply` re-selects a conversation's Agent,
  Environment and Vault on the machine it is already running (#1565).
```

### Non-blocking

- The `curl` example relies on `-d` implying `POST`, which matches the rest of `docs/api.md`, so no change wanted — just confirming it is deliberate and not a dropped `-X POST`.
- `docs/api.md` says the `409 rebuild_required` response carries "a `field` that names the cause". See my review on #1888: on a refresh of an environment edited in place, that field is always the generic `environment`. If #1888 takes the "soften the claim" route, this paragraph needs the same softening.

jhgaylor:
Both points addressed in `5c1e48d`.

**Root `CHANGELOG.md`** gets its `[Unreleased]` entry, naming the endpoint, the `configuration` webhook stage and the three columns. You were right that this is the natural PR for it and that the file is published under `/docs` rather than being bookkeeping.

**`docs/api.md` and `docs/reference/webhooks.md`** carry the `failed` status the `configuration` stage gained in #1890, and say the request succeeds once the selection is committed, with the event reporting whether the machine has read it.

**`docs/api.md` also stops overpromising the `field`**, per your cross-reference to #1888. I took the "soften the claim" route there, so the page now says Fountain answers `environment` when you edit one Environment in place, and why.

On the **1.28.0 collision**: thank you for naming the other three. I have left the bump at 1.28.0 for now rather than pre-emptively moving to 1.29.0, because whichever of #1885, #1876, #1860 and this one lands first takes it and the rest have to re-bump anyway. The `[!IMPORTANT]` note stands, and I will re-check against `main` immediately before merge. The four-edit shape is the trap to watch when resolving that conflict.

The curl example is deliberate, matching the rest of the page.

Gates re-run on the tip: `docs-style.py` (the same two pre-existing findings, neither on a page this touches), `vale lint` measured **per file before and after** rather than in aggregate (identical counts on every gating rule, and `webhooks.md` clean), `destink` (112 pages clean), `docs_test.exs` + `events_test.exs` 39 tests 0 failures.

BinaryBourbon (approved):
Both points landed, and the `CHANGELOG.md` entry is the right size — the endpoint, the stage, the three columns, no re-telling of the ADR.

`docs/api.md` is the part I want to single out, because it is now the clearest statement of the contract anywhere in the stack:

> The status is `done` when a machine holds the new selection, and `failed` when the selection stands and the machine has yet to read it. The request succeeds in both cases.

That is exactly right, and it is the sentence I am holding #1890 to. As of `e9a92ef` the code does not meet it: `announce_reapply/2` publishes `done` whenever no live server is holding the previous selection, which includes there being no server at all — the idle, suspended conversation this feature is mostly for. I have asked for the code to move, not the page, so I expect this file to stand as written. If that argument goes the other way and `done` ends up meaning "the request was applied", these two lines have to change with it. They cannot merge disagreeing.

The `field` softening matches #1888's route and says why in two sentences, which is as much as a reader needs. The `webhooks.md` row reads correctly for both statuses.

### Non-blocking

- The new entry makes a **second `### Added` under `[Unreleased]`** — there is already one further down, at what is now line 46. The section is triplicated already from earlier merges so this does not make it meaningfully worse, but folding into the existing heading is free while you are in the file.
- While you are there: `>>>>>>> theirs` is still sitting at line 77 of this file on your branch. It is inherited from `main` and #1881 removes it, so not yours — just be aware you are both editing `CHANGELOG.md` and it is a published `/docs` page in the meantime.

### The 1.28.0 collision

Your reasoning for leaving it is sound on its own terms and I am not going to block a docs PR over it. Recording where I still disagree, for whoever merges first: "re-check against `main` immediately before merge" is a manual step that has to fire correctly four times, and the failure mode is not a red build — it is a version bump that merges green and publishes nothing. `sdk-no-release` on all four plus one release PR afterwards costs one PR and removes the ordering constraint entirely. Your call, and the `[!IMPORTANT]` note plus the four-edit warning are a reasonable second best.

Approving.

jhgaylor:

BinaryBourbon:
Heads-up rather than a change request: the `conversation_server_size_test.exs` failure you will see here is inherited, not yours. `@pin 2716` comes from #1891 and the file is 2,739 lines at this commit. Nothing in this PR touches `conversation_server.ex`. Raised on #1891; approval here stands.

jhgaylor:
Rebased only — the parent moved for the size-pin fix in #1891 (`400afeb`). No change to this PR's own commit. Note the pin fix is what makes this branch green again; it was red on `conversation_server_size_test.exs` inherited from #1891.

PR #1903, opened 2026-09-11
chore(deps): ignore TypeScript majors for the TypeScript SDK

jhgaylor:
Stops #1759 ("bump typescript from 5.9.3 to 6.0.3 in /sdk/typescript") from reopening weekly and burning a full CI run, until upstream makes the bump possible.

## Why the bump cannot land

`sdk/typescript` generates `src/generated/openapi.ts` with `openapi-typescript`, and that generator peers on the TypeScript 5 line. Measured against the registry today (2026-09-11):

```
$ npm view openapi-typescript@latest version peerDependencies
version = '7.13.0'
peerDependencies = { typescript: '^5.x' }

$ npm view openapi-typescript dist-tags --json
{ "swagger-v2": "5.4.2", "next": "7.0.0-rc.1", "latest": "7.13.0" }
```

Every published release on the current line carries the same peer, so there is no generator version to pair a TypeScript 6 bump with:

| version | `peerDependencies` |
| --- | --- |
| 7.7.3 | `{"typescript": "^5.x"}` |
| 7.8.0 | `{"typescript": "^5.x"}` |
| 7.9.0 | `{"typescript": "^5.x"}` |
| 7.9.1 | `{"typescript": "^5.x"}` |
| 7.10.0 | `{"typescript": "^5.x"}` |
| 7.10.1 | `{"typescript": "^5.x"}` |
| 7.12.0 | `{"typescript": "^5.x"}` |
| 7.13.0 (`latest`) | `{"typescript": "^5.x"}` |
| 7.0.0-rc.1 (`next`) | `{"typescript": "^5.x"}` |

The peer is load-bearing, not decorative. Unpacking the published tarball, the generator builds its AST through whichever compiler is installed:

```
$ tar -xzf openapi-typescript-7.13.0.tgz
$ head -2 package/dist/lib/ts.mjs
import { parseRef } from '@redocly/openapi-core/lib/ref-utils.js';
import ts from 'typescript';
```

Ten more `dist/` modules import it the same way. So `--legacy-peer-deps` is not an escape hatch: it would run a TS-5-targeting generator against the newer compiler API, and the output is diff-gated by the `SDK types match the spec` step, which regenerates the file and fails on any change.

The consequence in CI is that `npm ci` exits 1 on `ERESOLVE` at the `SDK install` step of both affected jobs, so every gate that would tell us whether TypeScript 6 is actually safe for this package — typecheck, tests, build, browser bundle, `verify-contract`, conformance, the freshness diff — never runs. `sdk-publish.yml` uses the same install command, so the break would surface at publish time rather than at PR time if it landed.

## What this changes

The `sdk-typescript` group is `update-types: [minor, patch]`, so a major arrives as its own ungrouped PR. The ignore condition already recorded against #1759 is `typescript [>= 7.a, < 8]`, which does not cover 6.x — hence the weekly reopen.

This adds an `ignore:` entry on the `/sdk/typescript` npm update:

```yaml
    ignore:
      - dependency-name: typescript
        update-types: ["version-update:semver-major"]
```

**Majors as a class rather than `6.x` alone**, because the peer pins the 5 line: 6, 7 and anything after it break identically, and a `6.x`-only range would just move the weekly PR to the next major. Minors and patches still arrive through the existing group, so `typescript@^5.7.0` keeps getting updated.

### How the entry was verified

Nothing in this repo lints `dependabot.yml` — which is how the sibling bug in #1866 shipped as a silent no-op — so the entry was checked directly:

- Parsed with `yaml.safe_load` and validated against the SchemaStore [`dependabot-2.0.json`](https://json.schemastore.org/dependabot-2.0.json) schema: **valid**. `update-types` is an accepted key inside `ignore`, and `version-update:semver-major` is in its enum.
- Asserted structurally that the entry lands on the `npm` + `/sdk/typescript` update and nowhere else, that the `/scripts/destink` npm entry is untouched, and that the `sdk-typescript` group still reads `[minor, patch]`.
- `5.9.3 -> 6.0.3` is a major version update by definition, so it is what this condition suppresses.

This is the **npm** ecosystem, so npm requirement semantics apply — unlike the `github-actions` block in #1866, where `versions` parses as a Bundler requirement. Using `update-types` sidesteps range spelling entirely.

## Follow-ups

- **Remove this entry when `openapi-typescript` declares support for a newer TypeScript major**, and bump both together. The comment in the file says so.
- #1759 should be closed once this merges. I have commented there with this evidence and a link here, but left the close to you.
- **#1866 is concurrently editing the `ignore:` block of the same file** (its `github-actions` entry, ~35 lines above this hunk). The two hunks are far apart and should not textually conflict, but whichever merges second may want a trivial rebase.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -62,8 +62,17 @@ updates:
       # Majors as a class, not 6.x alone: the peer pins the 5 line, so 6, 7 and
       # anything after it break identically. Minors and patches still arrive,
       # grouped above.
+      # `ignore` withholds security updates too, not just version updates, so a
+      # TypeScript advisory fixed only in 6.x or later would also be held back.
+      # Accepted: it is a devDependency, never shipped to SDK consumers, and the
+      # 5.x line still gets patches through the group.
       # Delete this entry when openapi-typescript declares support for a newer
-      # TypeScript major, and bump both together (#1759).
+      # TypeScript major, and bump both together (#1759). Deleting it does not
+      # restore 7.x on its own: #1557 used `@dependabot ignore this major
+      # version`, recording `typescript [>= 7.a, < 8]` in dependabot's central
+      # state, which no file here shows. Run
+      # `@dependabot show typescript ignore conditions` to see what is stored,
+      # then `@dependabot unignore typescript [>= 7.a, < 8]` to clear it.
       - dependency-name: typescript
         update-types: ["version-update:semver-major"]
 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -45,6 +45,27 @@ updates:
     groups:
       sdk-typescript:
         update-types: [minor, patch]
+    ignore:
+      # The SDK generates `src/generated/openapi.ts` with openapi-typescript,
+      # which builds its AST through the compiler API it imports directly
+      # (`import ts from 'typescript'`, dist/lib/ts.mjs:2) and so declares
+      # `peerDependencies: {"typescript": "^5.x"}`. Every published release
+      # carries that peer, 7.7.3 through 7.13.0 (`latest`); the `next` tag is
+      # the older 7.0.0-rc.1. A TypeScript major above 5 therefore makes the
+      # tree unresolvable and `npm ci` exits 1 on ERESOLVE before typecheck,
+      # tests, build, `verify-contract`, conformance or the generated-types
+      # freshness diff get to say whether the compiler itself is fine
+      # (ci.yml `SDK install`, and sdk-publish.yml runs the same install).
+      # `--legacy-peer-deps` is not the way out: it would run a TS-5-targeting
+      # generator against the newer compiler API, and the `SDK types match the
+      # spec` step fails on any diff in the regenerated output.
+      # Majors as a class, not 6.x alone: the peer pins the 5 line, so 6, 7 and
+      # anything after it break identically. Minors and patches still arrive,
+      # grouped above.
+      # Delete this entry when openapi-typescript declares support for a newer
+      # TypeScript major, and bump both together (#1759).
+      - dependency-name: typescript
+        update-types: ["version-update:semver-major"]
 
   - package-ecosystem: npm
     directory: /scripts/destink


BinaryBourbon (approved):
Approving. I verified the mechanism independently rather than reading the summary, and the entry does what it claims: it parses, it validates, it lands on the one update it should, and `update-types` is the right lever for this. The majors-as-a-class choice is not just defensible, it is empirically proven in this repo — see the #1557 evidence below, which I think is the strongest argument available for the design and is worth adding to the PR body. Two non-blocking notes, one of which is a real gap in the removal condition.

### What I verified

- **`update-types` is a first-class key inside `ignore`, and it is not ecosystem-specific.** In the SchemaStore `dependabot-2.0.json` schema, `definitions.update.properties.ignore.items.properties.update-types` `$ref`s the shared `update-types` definition, whose enum is exactly `["version-update:semver-major", "version-update:semver-minor", "version-update:semver-patch"]`. Nothing in the schema conditions that key on `package-ecosystem`, and GitHub's options reference documents `version-update:semver-major` as "major releases" generally. Validating `.github/dependabot.yml` at head `43dce9c1` against that schema with `jsonschema` passes. So this is not the #1866 failure mode: there is no requirement string to be parsed by an ecosystem-specific grammar, which is the whole reason `versions: ["3.x"]` could silently match nothing there. `5.9.3 -> 6.0.3` is a semver-major update, so it is suppressed.

- **Placement is right and nothing else is touched.** Parsing the file at head and dumping each update entry: the `ignore` key appears on `npm` `/sdk/typescript` and on no other entry — `github-actions`, `mix`, `gomod`, `docker` and `npm` `/scripts/destink` all come back with no `ignore`. The `sdk-typescript` group still reads `update-types: [minor, patch]`, so `typescript@^5.7.0` keeps receiving minors and patches through the group; this condition only removes the ungrouped major PR.

- **`dependency-name: typescript` is an exact match, which matters here.** The schema documents wildcards as opt-in (`optionally using * to match zero or more characters`), so this does not match `openapi-typescript`. That is the right outcome and worth stating out loud: the generator's own major bump, which is the single event that would unblock #1759, still arrives as a PR. The ignore does not hide its own removal condition.

- **The design argument is proven by history in this repo, not just by reasoning.** #1557 ("bump typescript from 5.9.3 to 7.0.2", closed 2026-09-04) carries `@dependabot ignore this major version` from @jhgaylor, and Dependabot's reply is `OK, I won't notify you about version 7.x.x again`. That recorded precisely the `typescript [>= 7.a, < 8]` condition displayed on #1759 — and #1759 opened four days later regardless. A per-major ignore has already been tried here and already failed to hold. GitHub's docs also back the file-over-comment choice directly: comment-recorded preferences are "stored for the repository centrally", and "for repositories with more than one contributor it is better to explicitly define the dependencies and versions to ignore in the configuration file."

- **Composition with that recorded condition is a union, not a conflict.** The file entry (all majors) is a strict superset of the stored `[>= 7.a, < 8]`, so the stored one becomes redundant but harmless. No interaction to worry about at merge time.

- **No semantic conflict with #1866.** Its hunk adds an `ignore:` block to the `github-actions` update entry; this one adds a separate `ignore:` block to the `npm` `/sdk/typescript` entry. They are different list items under `updates:`, so whichever lands second needs at most a textual rebase, and the merged file parses with both conditions on their own ecosystems. I confirmed the two blocks target different entries by parsing each side.

- **Nothing in the repo validates this file.** The only other mention of dependabot outside the config is a comment at `.github/workflows/ci.yml:994`. The body's claim is accurate, and it is why the schema check above was worth doing rather than assuming.

### Two notes, neither blocking

1. **`ignore` suppresses security updates too, and the comment does not say so.** GitHub's options reference marks `ignore` as applying to Dependabot security updates as well as version updates, so a TypeScript advisory whose only fix ships in a 6.x or later release would also be withheld. I think this is the right trade anyway and would not hold the PR for it: `typescript` is a devDependency that is never shipped to SDK consumers, the 5.x line still receives patches through the group, and the compiler's recent advisory-shaped fixes are editor and ATA surface (`Harden ATA package name filtering` in the 6.0.3 notes on #1759) that this build does not expose. Worth one clause in the comment block so the next reader does not have to rediscover it.

2. **The removal condition as written is necessary but not sufficient.** The comment says to delete this entry when `openapi-typescript` declares support for a newer TypeScript major. Deleting it will not by itself restore TypeScript 7 updates, because the condition #1557 recorded lives in Dependabot's central state rather than in this file and will keep suppressing the 7.x line after the YAML is gone — invisibly, since nothing in the repo shows it. Suggest the comment also name the cleanup: `@dependabot unignore typescript [>= 7.a, < 8]` (or `@dependabot unignore typescript` to clear all of them), with `@dependabot show typescript ignore conditions` to inspect the stored state first. Both commands are documented, and the bracketed form matches the exact rendering Dependabot prints on #1759.

jhgaylor:
Both documentation notes from the review are closed in `327840ea`, in the comment block on the `/sdk/typescript` ignore entry.

**The removal condition now names the full cleanup.** The previous wording ("delete this entry when openapi-typescript declares support for a newer TypeScript major") was necessary but not sufficient, exactly as you said: the `typescript [>= 7.a, < 8]` condition #1557 recorded lives in Dependabot's central state, so deleting the YAML leaves 7.x suppressed with nothing in the repo to show it. Added:

```
      # Delete this entry when openapi-typescript declares support for a newer
      # TypeScript major, and bump both together (#1759). Deleting it does not
      # restore 7.x on its own: #1557 used `@dependabot ignore this major
      # version`, recording `typescript [>= 7.a, < 8]` in dependabot's central
      # state, which no file here shows. Run
      # `@dependabot show typescript ignore conditions` to see what is stored,
      # then `@dependabot unignore typescript [>= 7.a, < 8]` to clear it.
```

Both command spellings were checked against GitHub's current [Dependabot pull request comment commands](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-pull-request-comment-commands) reference rather than written from memory, since a wrong command here would repeat the failure mode this PR exists to fix. The documented forms are `@dependabot show DEPENDENCY_NAME ignore conditions` and `@dependabot unignore DEPENDENCY_NAME IGNORE_CONDITION`; the bracketed `[>= 7.a, < 8]` substituted for `IGNORE_CONDITION` is the exact rendering Dependabot printed on #1759.

**The security-update caveat is stated.** Not a change to the entry, just the fact recorded so it is not rediscovered:

```
      # `ignore` withholds security updates too, not just version updates, so a
      # TypeScript advisory fixed only in 6.x or later would also be held back.
      # Accepted: it is a devDependency, never shipped to SDK consumers, and the
      # 5.x line still gets patches through the group.
```

`ignore` is marked as supported by both version updates and security updates in the [options reference](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference), and the how-to states it directly: Dependabot ignores those dependencies "when it opens pull requests for version updates and security updates."

**The change is comment-only, and the structural diff is empty.** Dumping `yaml.safe_load` of the file to sorted JSON at `43dce9c1` and at `327840ea` and diffing the two produces no output, so the parsed configuration is identical: same `ignore` entry, same `update-types`, same group, nothing else touched. The file still parses and still validates against the SchemaStore `dependabot-2.0.json` schema. Your approval therefore still applies to the same entry.

PR #1911, opened 2026-09-11
fix(broker): refuse to boot when tenants are named with no listener (#1686)

### LINKED ISSUE #1686: Broker fails open when BROKER_TENANTS is set and BROKER_LISTEN_PORT is missing
## Symptom

With `BROKER_TENANTS=*` (production since #1553/home-cloud#163) and `BROKER_LISTEN_PORT` dropped from the pod env, boot succeeds and every tenant silently gets plaintext GitHub, inference and connection tokens in the sandbox. Connections and secret bindings keep "working" unbrokered.

## Mechanism

- `config/runtime.exs:240-286` raises only for "listen port set, proxy URL unset". Nothing checks "tenants set, listener unset".
- `Fountain.Broker.enabled_for?/1` (`apps/fountain/lib/fountain/broker.ex:146-152`) is `configured?() and ...`, so a missing listener makes it false for every tenant, and every egress path falls back to direct credentials.
- The listener gauge is emitted only on the native backend (`broker/native.ex` emit_telemetry), so the alert series goes absent rather than zero (the #1169 trap).

This is the envFrom drift that crash-looped #1495, in the other direction. #1495 made `BROKER_TOKEN` a warning; that variable is dead, but the listener port is load-bearing.

## Fix

Raise at boot when `broker_tenants != []` and `broker_listen_port` is nil. Add an `absent()` arm to the listener alert in home-cloud.

---

jhgaylor:
## The mechanism

`Fountain.Broker.enabled_for?/1` is `configured?() and <tenant is on the ratchet>`, and `configured?/0` is only true when `:broker_listen_port` is an integer (`apps/fountain/lib/fountain/broker.ex:139-152`). So a deployment that runs `BROKER_TENANTS=*` (production, since #1553) and loses `BROKER_LISTEN_PORT` from its pod env does not half-broker anyone. Brokerage turns itself off for every tenant, and each sandbox gets plaintext GitHub tokens, inference keys and connection tokens instead. The console keeps working, the connections keep working, and nothing in the logs says so.

`config/runtime.exs` already raised for the inverse case — "listen port set, proxy URL unset" — but nothing checked "tenants set, listener unset". The listener gauge is emitted only on the native backend, so the alert series goes absent rather than zero (the #1169 trap), and the monitoring does not catch it either.

This is the `envFrom` drift of #1495 in the other direction. #1495 made `BROKER_TOKEN` a warning because that variable was inert; the listener port is load-bearing, so it gets a raise.

## The fix

A boot guard in `config/runtime.exs`, next to the existing one: raise when `broker_tenants != []` (`:all` or a non-empty id list) and `broker_listen_port` is nil. Blank or separators-only `BROKER_TENANTS` still trims to nobody and still boots with no listener — the guard is on who is named, not on the variable being present, so the inert state the ratchet starts from is unaffected.

`docs/configuration.md` and `.env.example` carry the new hard requirement: `BROKER_LISTEN_PORT`'s Required column is now "With `BROKER_TENANTS`.", and the `BROKER_TENANTS` row says what boot refuses and why.

## How it was verified

- Two cases added to the existing `describe "the egress broker (ADR 0019)"` block in `apps/fountain/test/fountain/runtime_config_test.exs`, following the established `Config.Reader.read!(@runtime_exs, env: :prod)` pattern that file already uses. One asserts the raise for `*`, a single id and a padded list; the other pins that blank and `" , "` still boot with no listener.
- **Reverted the guard and re-ran**: the new case failed with `Expected exception RuntimeError but nothing was raised` (22 tests, 1 failure). Restored the guard: 22 tests, 0 failures.
- `mix test test/fountain/runtime_config_test.exs test/fountain/config_reference_test.exs` — 33 tests, 0 failures. `compose_passthrough_config_test.exs` — 16 tests, 0 failures.
- `mix format` from `apps/fountain` on the mise-pinned toolchain (no diff), and `mix compile --warnings-as-errors` clean.
- Three prose gates on the touched page: `python3 scripts/docs-style.py` reports only the two pre-existing findings in other files; `vale --config .vale-ste.yml docs/configuration.md` gives 0 errors and the two passive-voice warnings it already had at lines 82 and 163; `node scripts/destink/destink.mjs docs/configuration.md` is clean.

## Deliberately left out

The `absent()` arm on the listener gauge alert. That lives in the home-cloud repository, not this one, and is **still owed**.

Closes #1686


BinaryBourbon (approved):
Approving. The guard is in the right place, the condition is the right one, and both new tests are load-bearing. Details of what I checked and how.

## The guard fires only where it should

`config/runtime.exs:280` sits after `broker_listen_port` is parsed (line 205) and after `broker_tenants` is normalised (lines 253-270), so both operands are assigned. `broker_tenants` is exactly one of `[]`, `:all`, or a non-empty list of ids by that point, which makes `!= []` true for `:all` and for a list and false only for the inert case. I confirmed the three post-parse shapes by reading the `case` rather than trusting the name.

Nothing that legitimately names tenants can be bricked by this, because there is no configuration where the listener runs elsewhere. `Fountain.Broker.backend/0` (`apps/fountain/lib/fountain/broker.ex:137`) is `if is_integer(Application.get_env(:fountain, :broker_listen_port)), do: :native`, `@type backend :: :native` is the whole type since the ADR 0019 flip, and grepping `broker_listen_port` across `apps/`, `ee/` and `config/` turns up only `backend/0` and `Broker.Native`'s `child_spec` port. So any deployment the guard now refuses was already brokering nobody — the raise converts a silent no-op into a boot failure and takes no working configuration with it.

Places I checked for an accidental trip:

- `.github/workflows/` sets no `BROKER_*` at all, so the prod release boot check, the compose boot and the core-distribution boot are unaffected. All three are green on this PR.
- `docker-compose.yml:192` is `BROKER_TENANTS: ${BROKER_TENANTS:-}`, which reaches the app as `""` and trims to `[]`.
- `.env.example` and `.env.compose.example` keep `BROKER_TENANTS` commented out.
- The existing `"BROKER_TENANTS is blank for nobody, a list for some, `*` for everyone"` case sets `BROKER_LISTEN_PORT` in its `on` map, so it does not collide; `runtime_config_test.exs` is `async: false` and the describe's `on_exit` clears all five broker vars per test, so no ordering leak between the new cases and the ones around them.

The block is not wrapped in `config_env()`, so this raises in dev and test too — same as the `BROKER_LISTEN_PORT`/`BROKER_PROXY_URL` guard two lines up, and correct for a variable that has the same meaning in every environment.

## The tests are load-bearing

I ran both experiments in the worktree and restored it (`git status` clean, `d0db1fd3`).

- Deleted the `if` block at `config/runtime.exs:280-285`: `runtime_config_test.exs:324` fails with `Expected exception RuntimeError but nothing was raised` — 22 tests, 1 failure.
- Kept the raise but widened its condition to `System.get_env("BROKER_TENANTS") != nil`: `runtime_config_test.exs:338` (`"tenants that trim away to nobody still boot without a listener"`) fails — 22 tests, 1 failure.

So the second case is not filler; it pins the guard to who is named rather than to the variable being present, which is the part that would otherwise regress a manifest that sets `BROKER_TENANTS=` unconditionally.

Unmodified: `mix test test/fountain/runtime_config_test.exs test/fountain/config_reference_test.exs test/fountain/compose_passthrough_config_test.exs` — 49 tests, 0 failures. `mix format --check-formatted` exits 0 both at the umbrella root (which is what covers `config/runtime.exs`) and in `apps/fountain`.

## Prose gates

`python3 scripts/docs-style.py` reports 2 problems in 112 files, both in pages this PR does not touch (`docs/catalog/runtimes/codex.md:62`, `docs/concepts/environment.md:73`). `vale --config .vale-ste.yml docs/configuration.md` is 0 errors / 2 warnings, and both warnings are `STE.PassiveVoice` far from the edited rows; everything reported on lines 53 and 55 is `STE.Vocabulary`, which advises and does not gate. `node scripts/destink/destink.mjs docs/configuration.md` is clean.

## Two notes, neither blocking

1. `config/runtime.exs:281-283` — the message tells an operator to add the port, but not that the other way out is blanking `BROKER_TENANTS`. The case where that matters is a rollback: someone pulling the listener deliberately hits this raise and the message reads as if adding a port is the only exit. One clause naming the inert spelling would save that reader a trip to the docs.

2. `.env.compose.example:145` still says "List the tenants you want brokered" with no mention of the new hard requirement, while `.env.example:74-77` now spells it out. Both files list the port first and leave every key commented, so nobody boots into the raise by following either — it is an asymmetry, not a defect.

Separately, and only so it is on the record: "Closes #1686" will close the issue while the `absent()` arm on the listener gauge alert is still owed in home-cloud, which the PR body says plainly. Worth a tracking issue there so the monitoring half does not get closed along with the boot half.

jhgaylor:

PR #1912, opened 2026-09-11
fix(principals): clamp a principal's sandbox cap to the application's own (#1687)

### LINKED ISSUE #1687: Claimable principals: max_live_sandboxes is written unclamped to sandbox_limit_override
## Symptom

Any full-scope key can set a per-user sandbox cap that bypasses the balance clamp and `SANDBOX_CAP_CEILING`. A $5 account minting principals with `max_cost_usd: 0.01` and `max_live_sandboxes: 100000` can provision up to `SANDBOX_FLEET_CEILING` on one cent each.

## Mechanism

- `apps/fountain/lib/fountain/principals.ex:230`: `max_live_sandboxes: limits |> Map.get("max_live_sandboxes") |> as_integer(1) |> max(0)`. Floored at zero, no ceiling.
- `principals.ex:328`: `Accounts.create_principal_user(sandbox_limit_override: attrs.max_live_sandboxes)`.
- `quotas.ex:114`: `resolve_limit(override, _balance, _comped) when is_integer(override), do: override` honours the override before any balance clamp or cap ceiling.

`sandbox_limit_override` was an admin-only lever (`accounts.ex:1044`) until #1564.

## Fix

Clamp to `SANDBOX_CAP_CEILING` in `cast_create/1`, or better, clamp to the *application's own* effective `Quotas.sandbox_limit/1` so a principal can never out-provision the account that funds it.

---

jhgaylor:
## Mechanism

`Fountain.Principals.cast_create/1` read `limits.max_live_sandboxes` off the
request, floored it at zero, and wrote it to `users.sandbox_limit_override`
(`principals.ex:328`). `Quotas.resolve_limit/3` answers with that override
*first*:

```elixir
defp resolve_limit(override, _balance, _comped) when is_integer(override), do: override
```

so neither the balance rule nor `SANDBOX_CAP_CEILING` ever ran for a
principal. Any full-scope key could open principals with
`max_cost_usd: 0.01` and `max_live_sandboxes: 100000` and take
`SANDBOX_FLEET_CEILING` a cent at a time. `sandbox_limit_override` was an
admin-only lever (`accounts.ex:1044`) until #1564 taught the principals route
to write it, and nothing re-bounded it on the way.

## The fix

`cast_create/2` now takes the application and clamps the cap to what that
account may actually hand on — the issue's preferred option, in the same
`clamp(0, ceiling)` shape `grant_cents` already uses against
`max_grant_cents`:

```elixir
max_live_sandboxes:
  limits
  |> Map.get("max_live_sandboxes")
  |> as_integer(1)
  |> clamp(0, application_cap(application)),

defp application_cap(%User{} = application) do
  %{cap_ceiling: ceiling} = Fountain.Quotas.settings()

  application.id
  |> Fountain.Quotas.sandbox_limit()
  |> min(ceiling)
  |> max(1)
end
```

Each of the three steps exists for a different failure, and the comment above
the function says which:

- **the application's own cap** is the fix for the reported bug: a principal
  can never out-provision the account funding it. Read by **id**, not off the
  `%User{}` the caller passed, for the reason `check_application_allowed/2`
  already gives about a struct's cached balance.
- **`min(cap_ceiling)`** because `resolve_limit/3` returns an application's
  *own* `sandbox_limit_override` unclamped. An admin lever of 50 under a
  ceiling of 20 would otherwise be inherited by every principal it opens. (An
  earlier revision of this PR claimed the ceiling came along for free; that
  was only true of the balance-derived branch. Review caught it.)
- **`max(1)`** because `check_application_allowed/2` only checks the balance
  for a grant above zero, so a drained application asking `max_cost_usd: 0`
  passes it with an effective cap of 0. A zero written to the override is
  *durable* — `resolve_limit/3` answers with it before the balance rule, no
  claim clears it, only an operator can undo it — so that principal would
  report 0/0 even after being claimed and funded, with `Credits.gate` saying
  `:ok`. That is the 0/0 quota `quotas.ex:124` and `:244` say never to show,
  and it would be a regression, since the cap defaulted to 1 before this
  branch. The gate is the balance (ADR 0031): `check_spend/1` refuses an
  unfunded principal at every door that spends and stops refusing once it has
  credit, so a floor of 1 costs nothing and heals itself.

An **explicitly** requested `0` is still honoured, with a test pinning it:
that is an application asking for a principal that runs nothing, which
`ClaimableUser.changeset/2` has always permitted. The floor bounds the clamp,
not the caller.

Nothing else moves: the audit event already records `max_live_sandboxes`, and
it now records the clamped value — the one that took effect.

## Behaviour change worth knowing

A verified account holds the $5 opening credit, so its own cap is 2. An
application asking for more than it can itself run now gets its own number
rather than the number it asked for. The existing pass-through test asked for
3 and asserted 3 — it was asserting the bug — and now asks for the
application's own cap and asserts that.

## Verification

`apps/fountain` on the mise-pinned toolchain (Elixir 1.19.2 / OTP 28):

- `MIX_ENV=test mix compile --warnings-as-errors` — clean
- `mix credo --strict` on the changed files (from the umbrella root; credo is
  not a dep of `apps/fountain`) — no issues
- `mix format --check-formatted` — clean
- principals + principals_ceiling + quotas + quotas_fleet_ceiling +
  claimable_user_controller + audit_guardrail — **168 tests, 0 failures**
- `docs_test.exs` — 29/29 after the docs edit
- `okf validate decisions` valid, regenerated index unchanged
- vale on the edited page: 44 → 52 `STE.Vocabulary` suggestions (advisory;
  the added words are ones the page already uses), **0 errors, 0 warnings**
  on the six gating rules

**Every guard was verified by reverting the line it guards**, not the whole
change:

| Reverted | Failing test | Measured |
|---|---|---|
| the clamp (`\|> max(0)` back) | `cannot out-provision the account that funds it` | `left: 100000, right: 2` |
| the clamp, richly funded | `still held to the deployment's cap ceiling` | `left: 100000, right: 20` |
| `\|> max(1)` | `empty balance cannot mint a principal funding cannot revive` | `left: 0, right: 1` |
| `\|> min(ceiling)` | `admin override does not reach its principals` | `left: 50, right: 20` |

The zero-balance test walks the whole sequence: drain the application with a
`burn_turn` debit, open a principal, claim it, fund it 100,000 cents, then
assert `Credits.gate == :ok` **and** `Quotas.sandbox_limit > 0`.

Two further tests pass either way by design — the pass-through and the
default of one — and guard against the clamp becoming "always the
application's cap".

## Documented

ADR 0044 decision 3 and `docs/build/anonymous-visitors.md` now say the asked
limits are bounded by what the application itself holds.

## Not fixed here

**The aggregate.** A $5 account can still open 500 principals at 2 slots each,
so ~1,000 slots, bounded only by `SANDBOX_FLEET_CEILING` and
`max_outstanding_per_application`. Per-principal clamping cannot reach that;
it needs an aggregate over a tenant's live principals, or a cap funded from
the *remaining* balance rather than the whole of it. Out of scope for a P1
clamp, but it is the next question this area asks.

**The OpenAPI request schema** documents neither this clamp nor the
pre-existing `max_cost_usd` one. Adding a description there regenerates the
spec and the SDK contract artifact; the two clamps are worth documenting
together in a follow-up.

Closes #1687

### DIFF
--- a/apps/fountain/lib/fountain/principals.ex
+++ b/apps/fountain/lib/fountain/principals.ex
@@ -29,8 +29,10 @@ defmodule Fountain.Principals do
     * `max_live_sandboxes` is written to `users.sandbox_limit_override`, so
       `Fountain.Quotas` needs no knowledge of principals: the override branch
       answers before the balance rule is ever consulted. Because it answers
-      first, it is clamped at creation to the application's own effective cap,
-      so a principal can never out-provision the account funding it (#1687).
+      first, it is clamped at creation to the application's own effective cap
+      under the deployment ceiling, so a principal can never out-provision the
+      account funding it — and never below one, because a zero written there is
+      a gate that funding cannot reopen (#1687).
     * `max_cost_usd` is a credit grant moved from the application's balance
       into the principal's, so budget exhaustion is `Billing.check_spend/1`
       refusing at every door that spends (ADR 0031). With credits off nothing
@@ -205,10 +207,6 @@ defmodule Fountain.Principals do
   # *before* the balance rule and the cap ceiling are ever consulted. Unclamped,
   # a full-scope key could mint principals on a cent each that out-provision the
   # account funding them, up to `SANDBOX_FLEET_CEILING` (#1687).
-  #
-  # Read by **id**, for the reason `check_application_allowed/2` gives about the
-  # balance: the `%User{}` the caller loaded carries a cached cap input that may
-  # predate its own opening credit.
   defp cast_create(%User{} = application, params) do
     %{max_ttl_seconds: max_ttl, default_ttl_seconds: default_ttl, max_grant_cents: max_grant} =
       settings()
@@ -243,10 +241,40 @@ defmodule Fountain.Principals do
              limits
              |> Map.get("max_live_sandboxes")
              |> as_integer(1)
-             |> clamp(0, Fountain.Quotas.sandbox_limit(application.id)),
+             |> clamp(0, application_cap(application)),
            metadata: Map.get(params, "metadata", %{})
          }}
     end
+  end
+
+  # What the application may hand a principal: its own effective cap, itself
+  # held to the deployment ceiling, and never below one.
+  #
+  # Read by **id**, for the reason `check_application_allowed/2` gives about
+  # the balance: the `%User{}` the caller loaded carries a cached cap input
+  # that may predate its own opening credit.
+  #
+  # `min(cap_ceiling)` because `resolve_limit/3` returns an application's *own*
+  # `sandbox_limit_override` unclamped — an admin lever of 50 under a ceiling
+  # of 20 would otherwise be inherited by every principal it opens.
+  #
+  # `max(1)` because an application with an empty balance has an effective cap
+  # of zero, and a zero written here is durable: `resolve_limit/3` answers with
+  # the override before the balance rule, no claim clears it, and only an
+  # operator can undo it. That is the 0/0 quota `Fountain.Quotas` says never to
+  # show, on a principal that funding cannot reopen. The gate is the balance
+  # (ADR 0031) — `check_spend/1` refuses a principal with no credit at every
+  # door that spends, and stops refusing the moment it has some — so a cap of
+  # one on a broke application's principal costs nothing and heals itself. An
+  # *explicit* zero is still honoured: that is an application asking for a
+  # principal that runs nothing, which the schema has always allowed.
+  defp application_cap(%User{} = application) do
+    %{cap_ceiling: ceiling} = Fountain.Quotas.settings()
+
+    application.id
+    |> Fountain.Quotas.sandbox_limit()
+    |> min(ceiling)
+    |> max(1)
   end
 
   # The application funds its principals out of its own balance, so it must
--- a/apps/fountain/lib/fountain/principals.ex
+++ b/apps/fountain/lib/fountain/principals.ex
@@ -28,7 +28,9 @@ defmodule Fountain.Principals do
 
     * `max_live_sandboxes` is written to `users.sandbox_limit_override`, so
       `Fountain.Quotas` needs no knowledge of principals: the override branch
-      answers before the balance rule is ever consulted.
+      answers before the balance rule is ever consulted. Because it answers
+      first, it is clamped at creation to the application's own effective cap,
+      so a principal can never out-provision the account funding it (#1687).
     * `max_cost_usd` is a credit grant moved from the application's balance
       into the principal's, so budget exhaustion is `Billing.check_spend/1`
       refusing at every door that spends (ADR 0031). With credits off nothing
@@ -191,13 +193,23 @@ defmodule Fountain.Principals do
           {:ok, %{claimable: ClaimableUser.t(), api_key: String.t(), claim_token: String.t()}}
           | {:error, term()}
   def create_claimable(%User{} = application, params, opts \\ []) do
-    with {:ok, attrs} <- cast_create(params),
+    with {:ok, attrs} <- cast_create(application, params),
          {:ok, claimable} <- upsert_claimable(application, attrs, opts) do
       issue_credentials(claimable, opts)
     end
   end
 
-  defp cast_create(params) do
+  # A third ceiling in the spirit of the two in `check_application_allowed/2`,
+  # bounding a leaked application key: `max_live_sandboxes` lands on
+  # `users.sandbox_limit_override`, which `Quotas.resolve_limit/3` answers with
+  # *before* the balance rule and the cap ceiling are ever consulted. Unclamped,
+  # a full-scope key could mint principals on a cent each that out-provision the
+  # account funding them, up to `SANDBOX_FLEET_CEILING` (#1687).
+  #
+  # Read by **id**, for the reason `check_application_allowed/2` gives about the
+  # balance: the `%User{}` the caller loaded carries a cached cap input that may
+  # predate its own opening credit.
+  defp cast_create(%User{} = application, params) do
     %{max_ttl_seconds: max_ttl, default_ttl_seconds: default_ttl, max_grant_cents: max_grant} =
       settings()
 
@@ -227,7 +239,11 @@ defmodule Fountain.Principals do
            expires_at: DateTime.utc_now() |> DateTime.add(ttl, :second) |> truncate(),
            grant_cents:
              limits |> Map.get("max_cost_usd") |> usd_to_cents() |> clamp(0, max_grant),
-           max_live_sandboxes: limits |> Map.get("max_live_sandboxes") |> as_integer(1) |> max(0),
+           max_live_sandboxes:
+             limits
+             |> Map.get("max_live_sandboxes")
+             |> as_integer(1)
+             |> clamp(0, Fountain.Quotas.sandbox_limit(application.id)),
            metadata: Map.get(params, "metadata", %{})
          }}
     end
--- a/apps/fountain/test/fountain/principals_test.exs
+++ b/apps/fountain/test/fountain/principals_test.exs
@@ -13,6 +13,7 @@ defmodule Fountain.PrincipalsTest do
   alias Fountain.Credits
   alias Fountain.Principals
   alias Fountain.Principals.ClaimableUser
+  alias Fountain.Quotas
 
   defp application_account, do: insert_verified_user()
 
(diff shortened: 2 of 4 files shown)


BinaryBourbon (changes requested):
Reviewed against `CLAUDE.md`'s "Credits are the product" rules and ADR 0044. The
core of this is right: `sandbox_limit_override` genuinely is answered by
`resolve_limit/3` before the balance rule and the cap ceiling, so an unclamped
wire value was a real bypass, and clamping at creation is the right place to
stop it. I ran the numbers rather than reading them.

## What I verified

- **The old test was asserting the bug.** Reverted the clamp hunk in
  `principals.ex` (tests untouched) and ran the three files. Two of the new
  tests go red, and the failure prints the numbers: `left: 100000 / right: 2`
  and `left: 100000 / right: 20`. A verified application's own cap really is 2
  ($5 opening credit ÷ $2 reserve), so "asked 3, asserted 3" was asserting a
  principal above the account funding it. Restored; tree clean.
- **The rewrite is not vacuous** — for the two new tests. Note that the
  *renamed* pass-through test (`asked = Quotas.sandbox_limit(app.id)`) passes
  with the clamp reverted; it is a pass-through test, which is fine, but the
  guard for #1687 is carried entirely by the other two.
- **The DB read is not in a transaction.** `create_claimable/3` runs
  `cast_create/2` before `upsert_claimable/3`, and the `Repo.transaction` is
  inside `insert_claimable/3`. No advisory lock is held; `sandbox_limit/1` takes
  none. No lock-ordering interaction with `with_sandbox_reservation/3`.
- **Read by id, not off the struct** — correct, and for the reason the comment
  gives.
- **Idempotent replays** are unaffected: `upsert_claimable/3` returns the
  existing row, so a replay never re-clamps an existing principal's override.
- `mix format --check-formatted` and `mix credo --strict` clean on the changed
  files; 82 tests pass across `principals_test`, `quotas_test` and
  `claimable_user_controller_test`.
- The audit event and the JSON response both carry the clamped value, so the
  caller can observe its effective cap. Good.

## Finding 1 (blocking): a momentarily broke application mints permanently dead principals

`check_application_allowed/2` does **not** make the zero-balance branch
unreachable, on two counts: it runs *after* `cast_create/2`, and its balance
check is conditional on `grant_cents > 0` (`principals.ex:274`). An application
that asks for no introductory grant is never balance-checked at all.

Measured, on this branch:

```
app balance=0 limit=0
principal cap asked-default=1 got=0
principal override=0
# then fund the principal with 100,000 cents, as an owner would after a claim:
funded principal limit=0
gate=:ok
```

So an application whose balance has hit zero — not an abuser, just an account
between top-ups — keeps serving visitors, and every principal it opens in that
window gets `sandbox_limit_override = 0` written into its `users` row. That zero
is permanent. It survives the claim (nothing in `write_claim/3` clears it), and
`resolve_limit/3`'s override branch answers with it forever, so the owner who
claims and funds that principal gets `{:sandbox_quota_exceeded, %{count: 0,
limit: 0}}` with `Credits.gate/1` returning `:ok`. Only an admin can undo it.

That is precisely the outcome `quotas.ex:124` and `:244` say the design avoids:
"this account is refused as `insufficient_credits`, never as a 0/0 quota" /
"'out of credit' is the answer it should hear, not '0/0 sandboxes'". Before this
PR the same principal got the default of 1 and started working the moment money
arrived.

Two ways out, both small:

1. *(preferred)* refuse the create when the application's own effective limit is
   0 — an account that cannot start a sandbox itself has no business handing out
   principals — as a fourth clause in `check_application_allowed/2`, next to the
   ceilings it already owns. The caller gets a real error instead of a live
   credential for a machine that can never start, and no durable 0 is written.
2. clamp to `max(app_limit, 1)`, which keeps today's default behaviour for the
   broke case while still killing the `100000` bypass. Cheaper, but it leaves
   1 per principal when an admin has deliberately dropped an abusive
   application's override to 0.

## Finding 2 (non-blocking): "the deployment ceiling comes along for free" is not true when the application has an override

The PR body argues the `SANDBOX_CAP_CEILING` is inherited because
"`resolve_limit/3` mins every balance-derived cap against `cap_ceiling`". It
mins every *balance-derived* cap — the override branch returns first, unmapped.
Measured:

```
ceiling=20 app limit=50
principal cap=50 (ceiling is 20)
```

An operator who raises one trusted application to 50 now silently authorises 50
per principal, times `max_outstanding_per_application` (500), against a fleet
ceiling that is the only thing left holding the line. Arguably intended — the
operator did raise that account — but it is not what the description claims, and
`clamp(0, min(Quotas.sandbox_limit(application.id), Quotas.settings().cap_ceiling))`
would make the claim true and is strictly tighter than either option the issue
offered. At minimum the sentence in the PR body (and the new comment above
`cast_create/2`, which repeats the idea) should not assert the ceiling is
covered.

Residual worth recording either way: the clamp reduces the amplification but
does not remove it. A $5 account still gets `500 outstanding × 2 = 1,000`
sandbox slots at a cent apiece; `SANDBOX_FLEET_CEILING` is what bounds that, as
the issue acknowledged.

## Finding 3 (nit): the behaviour change is not documented anywhere a caller reads

`docs/build/anonymous-visitors.md` shows `limits.max_live_sandboxes` with no
hint that it is now bounded by the calling account's own cap, and ADR 0044
decision 3 still reads "writes `max_live_sandboxes` to
`users.sandbox_limit_override`" flat. One clause in each would keep an
application author from filing a bug when they ask for 5 and read back 2.

Finding 1 is the one I would like addressed before this merges; the rest is
yours to weigh.

jhgaylor:
Both findings confirmed and fixed in bc576322. Thank you — the zero-balance one was a regression I introduced, and I had the ceiling claim wrong.

## Blocking: the clamp could write a durable zero

Confirmed exactly as measured. `check_application_allowed/2` only checks the balance when `grant_cents > 0`, so a drained application asking for `max_cost_usd: 0` walks past it, its own effective cap is `0`, and my clamp handed that `0` to the principal. `resolve_limit/3` matches `is_integer(0)` and returns it before the balance rule, nothing in the claim path clears it, so the principal stayed at `0/0` after being claimed and funded while `Credits.gate` said `:ok`. That is the 0/0 quota `quotas.ex:124` and `:244` say never to show, and it is worse than the bug I was fixing because funding cannot reopen it.

Fixed by flooring, not refusing, per your reasoning and CLAUDE.md's: the gate is the balance, so a cap of 1 on a broke application's principal is refused by `check_spend/1` at every door that spends and heals itself the moment credit arrives.

One deliberate narrowing: the floor applies to the **clamp**, not to an explicitly requested `0`. `max_live_sandboxes: 0` still yields `0` — that is an application asking for a principal that runs nothing, which `ClaimableUser.changeset/2` has always permitted (`greater_than_or_equal_to: 0`) and which this PR should not remove. There is a test pinning it. Say the word if you would rather the floor be absolute.

## Non-blocking: the ceiling claim was wrong, so I took the clamp

You are right that `resolve_limit/3` returns an application's own `sandbox_limit_override` unclamped, so the ceiling did not come along for free. I took the clamp rather than dropping the claim. The extraction is now:

```elixir
defp application_cap(%User{} = application) do
  %{cap_ceiling: ceiling} = Fountain.Quotas.settings()

  application.id
  |> Fountain.Quotas.sandbox_limit()
  |> min(ceiling)
  |> max(1)
end
```

with a comment on each of `min` and `max` saying which failure it exists for. The PR body no longer claims the ceiling is free.

## Verification

Three tests, each verified by reverting **its own half** of the fix rather than the whole thing:

- dropping `|> max(1)`: `an application's empty balance cannot mint a principal funding cannot revive` fails at `left: 0, right: 1` — the zero-balance walk is create, claim, fund 100,000 cents, then assert `Credits.gate == :ok` and `Quotas.sandbox_limit > 0`, i.e. your whole sequence.
- dropping `|> min(ceiling)`: `an application's admin override does not reach its principals` fails at `left: 50, right: 20`, your number.
- `an explicitly requested cap of zero is still honoured` pins the narrowing above.

Gates, from `apps/fountain` on the pinned toolchain: `MIX_ENV=test mix compile --warnings-as-errors` clean, `mix credo --strict` on both files no issues, `mix format --check-formatted` clean, and principals + principals_ceiling + quotas + quotas_fleet_ceiling + claimable_user_controller + audit_guardrail at **168 tests, 0 failures**. `docs_test.exs` 29/29 after the docs edit.

## The nit, taken

ADR 0044 decision 3 and `docs/build/anonymous-visitors.md` now both say the asked limits are bounded by what the application itself holds. `okf validate decisions` is valid and the regenerated index is unchanged. Vale on the edited page is 44 → 52 `STE.Vocabulary` suggestions (advisory, and the added words are ones the page already uses); zero errors and zero warnings on the six rules that gate.

## Not fixed here, and worth stating rather than implying

A $5 account can still open 500 principals at 2 slots each, so ~1,000 slots bounded only by `SANDBOX_FLEET_CEILING` and `max_outstanding_per_application`. Per-principal clamping cannot reach that; it needs an aggregate over a tenant's live principals, or funding the cap from the remaining balance rather than the whole of it. I have left it out of a P1 clamp deliberately and noted it in the PR body.

BinaryBourbon (approved):
`bc576322` addresses both findings. Re-verified by reverting each half of
`application_cap/1` separately rather than taking the report on trust, and both
numbers match what you reported.

## Re-verified

- **`|> max(1)` guards the durable zero.** Dropped that one line: the
  zero-balance walk fails at `left: 0, right: 1`
  (`principals_test.exs:105`). Restored.
- **`|> min(ceiling)` guards the inherited override.** Dropped that one line
  instead: the override test fails at `left: 50, right: 20`
  (`principals_test.exs:138`). Restored; `git status` clean.
- The zero-balance walk now reproduces exactly the sequence I measured last
  round — drain, open, claim, fund 100,000 — and ends where it should:
  `Credits.gate == :ok` **and** `sandbox_limit > 0`. That is the case that was
  broken; it is now pinned.
- 85 tests green across `principals_test`, `quotas_test` and
  `claimable_user_controller_test`. `mix format --check-formatted` and
  `mix credo --strict` clean on the changed files. Vale reports nothing gating
  on the new docs paragraph (Vocabulary suggestions only, which do not gate).
  `okf validate` is green on the PR.
- ADR 0044 decision 3 and the docs page now state the bound, and the docs
  sentence pointing the reader at `max_live_sandboxes` in the response is the
  right place to have put it.

## Ruling on the narrowing: honour the explicit zero

Your narrowing is right, and I would keep it.

What I blocked on was not the value 0. It was Fountain converting a *transient
property of a third party* — the application's balance at that instant — into a
permanent property of a tenant row nobody had said anything about. An explicit
`max_live_sandboxes: 0` has none of that: the caller wrote it, the schema has
always accepted it, the response echoes it, and "a principal that holds
resources but runs no sandbox" is a coherent thing to ask for. Forcing it to 1
would be the API silently granting concurrency that somebody explicitly asked
not to have, which is the worse failure of the two. The test pinning it earns
its place.

I checked the two things you asked me to check.

**Is it as durable?** Yes, identically — I measured it: an explicitly-zeroed
principal, claimed and funded with 100,000 cents, reports `gate=:ok limit=0`.
That is still an operator-only unstick. It is caller-chosen, so I would not
block on it, but the claimer is a *different* party from the application that
chose it, and they cannot self-service. One clause in
`docs/build/anonymous-visitors.md` next to your new paragraph — that a cap of
zero is kept by the principal across a claim — would close it. Non-blocking.

**Can anything produce a zero without the caller meaning it?** One thing can,
and it is the only one. Measured on this commit:

```
asked=-1     -> cap=0
asked="-1"   -> cap=0
asked=0      -> cap=0
asked="0"    -> cap=0
asked=nil    -> cap=1
asked="banana" -> cap=1
```

`clamp(0, application_cap(application))` floors a negative at 0, so `-1` — a
typo, or the "unlimited" sentinel some clients use — buys a permanently dead
principal, silently, with a 201. It predates this branch (`max(0)` did the
same), but this branch is what makes 0 a deliberate value, so it is now the one
input that means something the caller did not say. Either treat only a literal
zero as the explicit zero (`0 -> 0; n -> clamp(n, 1, cap)`), or add
`minimum: 0` to `max_live_sandboxes` in the Grant request schema
(`claimable_user_controller.ex:66`) so it is a 422 instead. I prefer the schema
bound: it is the caller's mistake and the problem is that it is silent. While
you are in there, `expires_in` carries "Clamped to the deployment's maximum."
and `max_live_sandboxes` carries no description at all, which is now the less
obvious of the two.

## One consequence of the floor I asked for, for the record

`max(1)` also lifts an application whose `sandbox_limit_override` an operator
set to **0** — the abuse lever `quotas.ex` documents ("drop it to zero during
abuse"). That application can still open principals at 1 each, up to
`max_outstanding_per_application`. I think that is acceptable, because
suspension is the actual stop and it refuses API-key auth outright, so the
route is closed to a suspended account. But it does mean the cap lever alone no
longer throttles an application that opens principals. If you would rather keep
both properties, the two zeros are distinguishable — a balance-derived zero is
transient and deserves the floor, an operator's explicit override of zero is
neither transient nor un-self-serviceable and could pass through. Not something
I would hold the PR for; worth a line in the comment if you agree with the
reasoning, or a follow-up if you do not.

Approving. The security hole the issue reported is closed, the regression I
found is closed and pinned by a test that fails without its line, and the
aggregate residual is now stated in the PR body rather than implied.

jhgaylor:

PR #1913, opened 2026-09-11
fix(conversations): restart the turn when the runtime session is gone (#1667)

### LINKED ISSUE #1667: session_gone fails the turn instead of restarting the session — every conversation idle across a deploy looks dead
## Symptom

Reported as "my conversations are by and large non-responsive". Prod was healthy the whole time — the API served 2907×200 with zero 5xx, and a fresh `fountain run` completed end to end.

The discriminating test: **run a turn on a new conversation, then on an existing idle one.**

```
$ fountain run "Salon · Opus 5" -p "Reply with exactly: PONG."
▸ turn: started ▸ model: done
PONG▸ turn done                          # new conversation: works

$ fountain conv prompt ea00e769-864f-4c7f-be0a-39af940abd9a -p "Reply with exactly: PONG."
▸ turn: started ▸ session: done
▸ turn failed                            # existing conversation: fails

$ fountain conv prompt ea00e769-864f-4c7f-be0a-39af940abd9a -p "Reply with exactly: PONG."
▸ turn: started ▸ model: done
PONG▸ turn done                          # identical retry: works
```

The failure:

```
session|done   {"reason":"session_gone",
                "detail":"Resource not found: 59769299-239d-4282-aff7-47d3aa217e70",
                "event":"reset"}
turn|failed    {"reason":"The runtime session for this conversation is no longer on its
                sandbox: ... Send your prompt again — the next turn starts a fresh session
                on this same conversation. Its history is kept, but the agent will not
                remember the turns before this one."}
```

## Why this reads as an outage

This is #1657's designed behaviour: fail the turn, tell the user to resend. But the message only helps someone reading log events. In the conversations app a user sends a message, nothing comes back, and the conversation is indistinguishable from dead. It is a per-conversation symptom that no fleet-wide health check can see — which is exactly why prod looked green while conversations felt broken.

## Why sessions kept vanishing

**8 deploys on 2026-09-06, versus 0 the day before.** Each pod roll kills the sandbox ACP sessions, so every conversation that was idle across a deploy burns its next prompt. The same cause shows up in `turn|interrupted`, which went from 0/hour on every prior day to 4–5/hour:

```
hour   interrupted  started
11:00       5         25
10:00       4         29
09:00       4         53
08:00       1         44
07:00       0         10     <- and 0 for every hour before this
```

(Those interrupts are legitimate, for the record — every one is `audit_events.actor='api'` from a real client IP, i.e. deliberate stop calls, not a fault.)

## Proposed fix

Restart the session transparently and send the prompt, rather than failing the turn. The turn already knows the session is gone and already knows it must start a fresh one; making the user perform that retry by hand buys nothing and costs the appearance of a dead product.

If losing prior context must stay visible, surface it as a notice on the turn that *did* run, not as a failure of the turn that did not.

## Notes

The already-fixed morning outage (#1640 → #1663) is a separate matter; this survived it. See also the shared-sandbox collisions filed alongside this, which produce the same user-visible symptom by a different route.

jhgaylor:
The crash-report redaction prerequisite #1914 and transport deadline wakeup fix #2047 have landed. This branch includes current main with both prerequisites and preserves its reset/retirement guards.

When an ACP runtime reports that its saved session is gone during resume/load, the current prompt now continues once on a fresh runtime session. The transcript announces that the agent lost its memory, while preserving the conversation history and one turn-start/turn-done pair.

Recovery uses the released `managoat_acp` 0.4.2 `Peer.restart_session/1` API on the same initialized peer and remote command. The original prompt/images, authentication, model, request sequence, measurements and bounded execution remain in place. It never respawns a command, resets a journal, extends a deadline or admits a second turn. Failed creation and unrelated errors remain failures; each turn can recover only once.

The session-clear operation must pass the current-turn generation fence before recovery can write to the peer. A superseded legacy worker discards its local connection and measurements without touching its successor. Bounded recovery retains the existing deadline, write fencing and confirmed-stop cleanup path. Runtime error details are scrubbed from process status/crash reports.

Validation: compile, format, strict Credo, Dialyzer, Sobelow and the integrated production release assembly passed. Final complete `mix precommit --max-cases 4 --seed 299955` passed on signed commit `994c869587ac9e378073e2d8e621d352681714f5` with four BEAM schedulers and a fresh dedicated database: 6 doctests and 5,698 core tests, 144 Buzz tests and 33 Support tests, all with zero failures (2 skipped, 7 excluded). Focused regressions cover recovery, the one-retry guard, successor fencing, one command/prompt/turn, retained execution limits/deadline and expiry before the recovered prompt. The dependency's own release passed 234 tests, coverage above its 95% gate, static analysis and package checks; library PR: https://github.com/managoat/managoat_acp/pull/15.

The first full run overlapped several other suites and the unchanged docs anchor guard exceeded its normal 60-second timeout inside MDEx rendering (seed 299955). After stopping the contending runs and resetting the dedicated database, the unchanged docs module passed 29/29 twice, including the original seed, in 2.6 and 3.1 seconds. No timeout or assertion was loosened; the different outcomes are retained as timing evidence, without claiming a proven cause.

The actual production-built conversations React app was exercised locally with a mock API and live SSE. Recovery and the next turn render without duplicate/open turn groups or browser errors. This checks the real client against the recovery event sequence; it is not a production rollout or live-provider smoke.

Closes #1667


BinaryBourbon (approved):
Reviewed against the worktree at `p0/1667` (88576881). I went looking for a loop, a tenant leak, a double charge and a test that only proves itself, and did not find any of them. Approving. What I checked and how, then five nits, none blocking.

## Verified

**The no-loop guarantee holds, and the belt-and-braces claim is the real one.**

I traced the flag both ways.

- `TurnMachine.drive/4` (`turn_machine.ex:167-179`) calls `into_state(state, turn)` **before** reducing the effects, so `state.turn_session_retry` is already the detail by the time `{:restart_session, _}` is applied. `restart_session/2` then re-derives the machine with `from_state/1`, and `restart_attempt/2` clears `span`/`metrics`/`tracer` but not `session_retry` (`turn_machine.ex:706-724`). A second `session_gone` on the same row therefore lands in the `true ->` arm. Confirmed by mutation, below.
- Reset paths: `finish/4` (`:648`), `close_interrupted/1` (`:684`), `kick_turn/4` (`conversation_server.ex:2112-2113`), `fail_turn_before_start/6` (`:2353`). `open_autonomous_turn` does not clear it, but it only runs with no turn open, which means a `finish` already did.

The load-bearing claim checks out: `forget_runtime_session/4` returns `%{state | runtime_session_id: nil}`, so `session_plan/2` (`turn_machine.ex:1112-1117`) reads `:run`, and `Managoat.ACP.Peer.start_session/1` (`deps/managoat_acp/lib/managoat/acp/peer.ex:977`) dispatches `:run` to `start_new_session/1`, whose pending tag is `:new_session`. `:new_session` is not in `[:resume_session, :load_session]`, so the restarted attempt cannot re-enter this clause at all. The flag really is only the brace.

**Across a `ConversationServer` crash the flag is forgotten, and that is fine — I checked the path rather than assuming it.** A restarting server with a `running` turn row does not re-run the ACP handshake against it: `reattach_running_turn/1` (`conversation_server.ex:1212`) either attaches to the live sandbox session (no `session/resume`, so no `session_gone` possible) or calls `mark_orphan/3`. So the same turn row can never get a second restart out of a forgotten flag. A *new* turn gets a fresh budget of one, which is the design. And the restart is convergent rather than looping: `session_plan/2` persists a fresh placeholder and `session/new` overwrites it with a real id, so the "stale placeholder no rollout ever existed under" state this PR's first unit test describes now self-heals on the next prompt instead of failing every prompt forever.

**Tenant isolation is sound.** `_unsafe_list_turn_images/1` is reached only from `restart_session/2`, where `turn` is `state.current_turn` — a row this server created for its own conversation. That is CLAUDE.md's GenServer-has-already-established-ownership case, and the comment at `conversation_server.ex:2509-2510` names it. Shape round-trips correctly too: `TurnImage.data` is `:binary` and stored raw by `_unsafe_insert_turn_images/2`, and `Peer.image_blocks/1` (`peer.ex:1184-1188`) is what base64-encodes it, so reading back `%{media_type:, data:}` reproduces exactly what attempt 1 would have sent.

**Billing is correct, and the one thing I was most worried about is a non-issue.** `unpriced_turns/1` (`ee/lib/fountain/workers/credit_pricer.ex:152-179`) anti-joins on `burn_turn:<turn.id>` and prices `ended_at - started_at` off a single row, so one row is one key and one interval — `restart_attempt/2` touches neither `started_at` nor the row. More importantly: `usage.turn_started` is recorded by `record_turn_usage/1` from `_unsafe_create_turn/1` and `_unsafe_create_turn_on_sandbox/3` (`conversations.ex:1275`, `:1348`) — on **row creation**, not on the stage publish. So the double `turn|started` does not double-count the usage event that `Billing` aggregates turns from. Worth stating explicitly in the PR body, because it is the difference between a cosmetic duplicate and a metering bug.

**The double `turn|started` is not load-bearing anywhere in-repo.** I checked every consumer I could find:
- `openai_controller.ex:885-895` binds the turn on the first `started` only (the clause is guarded `%{turn_id: nil}`); the second matches nothing and falls through.
- `cli/internal/acp/prompt.go` `classifyStage` (`:287-330`) treats `turn`/`started` as non-terminal (and `session`/`done` as non-terminal too, so the new notice does not end a CLI turn early).
- `Fountain.Webhooks.Events` has `conversation.turn.started` in the catalogue but **not** in `@defaults`, so only an endpoint that opted in sees two.
- `sdk/typescript/src/turn.ts:66-69` does re-emit a second `turn-start` `RunEvent` with the same `turnId`/`turnNumber`. It resets no accumulated state, and attempt 1 produced no output, so nothing is lost.

**Tests fail against the old code.** I did the mutations myself rather than taking the table on trust.

Reverting the restart arm (`is_nil(turn.session_retry) ->` → `false ->`) reds exactly four:

```
1) the -32002 a replaced disk answers with is the same restart (TurnMachineTest)
2) a resume against a session that is not there restarts the turn (TurnMachineTest)
3) runs the prompt on a fresh session instead of failing the turn (ConversationServerACPTest)
4) says on the transcript that the agent's memory did not survive (ConversationServerACPTest)
138 tests, 4 failures
```

Removing the guard instead (`is_nil(turn.session_retry) ->` → `true ->`) reds exactly one, and only that one:

```
1) a second session_gone on the fresh session fails the turn (TurnMachineTest)
```

That matches the PR's table. The rewritten tests at 319/332/365 are carrying weight, not describing the diff. `git status` is clean; both mutations were restored from a copy.

**Gates.** `turn_machine_test` + `conversation_server_test` + `conversation_server_acp_test` + `conversation_server_size_test` + `webhooks/events_test`: 207 tests, 0 failures. The whole of `test/fountain/conversations/`: 939 tests, 0 failures, 2 skipped. `audit_guardrail_test` + `conversations_test`: 78, 0 (the new context function is a read, so the guardrail has nothing to say, but worth confirming). `mix format --check-formatted` from `apps/fountain`: clean. `mix credo --strict` over the three changed lib files: no issues. Size pin: `conversation_server.ex` is 2613 lines against `@pin 2774`.

## Nits

1. **`turn_machine.ex:529-532` — the log line is wrong in two of three branches.** The `Logger.warning` sits above the `cond`, so the second `session_gone` (which fails the turn) and the no-turn-open case both log "clearing it so **this turn starts a new one**". An operator grepping for the repeat failure gets told a restart happened when the turn was just failed. Move it into the arms, or word it neutrally.

2. **`conversation_server.ex:2505` — `restart_session/2` will raise if `state.current_turn` is nil.** `turn.prompt` on nil is a `BadMapError`, and the preceding `{:drop_connection, _}` effect *can* nil it: `drop_connection/2` → `close_autonomous_turn/2` → `finish_acp_turn/4` when `TurnMachine.autonomous_turn?/1`. I convinced myself it is unreachable today — a `resume_session`/`load_session` failure is a handshake failure, and an autonomous turn can only open after a session exists — but the function is depending on an invariant two modules away for its liveness. A `if turn do … else state end`, or asserting `turn.origin != "autonomous"` at the top of `session_gone/3`, costs one line and makes the assumption local. (Also: `agent_for/1` uses the non-bang `_unsafe_get_agent`, so a mid-turn agent deletion replays with `agent: nil` — same as `resume_acp_connection/5`'s refusal arm, so consistent, just noting it is deliberate.)

3. **"It bills less than today" overstates it** (PR body, and the docstring at `turn_machine.ex:695-698`). At `CREDIT_TURN_HOUR_CENTS` 25, one cent is 144 seconds, and `Credits.turn_cost_cents/1` rounds; a failed resume handshake is seconds, so today's dead turn row almost always prices at 0 and is never written. What actually changes is that the dead attempt's few seconds now ride on the successful turn's interval instead of rounding away. That is the right call and the amount is noise — but "no extra row, and the dead attempt's seconds fold into the one that ran" is the accurate sentence, and the docstring will be read as a claim later.

4. **`store_images/2` is log-and-continue on a rejected image** (`turn_machine.ex:1068-1077`), so a restarted turn replays only what persisted while attempt 1 would have sent the rejected one too. Arguably more correct, but it is a behaviour difference between the attempts that the "read back rather than stashed" paragraph does not mention.

5. **The "deliberately left out" note on the double `turn|started` should name the TypeScript SDK**, not only webhook subscribers — `sdk/typescript/src/turn.ts:66` is first-party and does emit `turn-start` twice. It is still not worth threading a flag through `run_fresh_turn/7` inside a size-pinned file, so I agree with the decision; I would just rather the note said which first-party surface it lands on. It would also be worth saying that `usage.turn_started` is *not* affected, since that is the reading someone will jump to.

None of these change the behaviour under test. Good diff — the separation between the pure machine and the server effect is kept, the restart is refused by the machine rather than remembered by the server, and the span of the abandoned attempt is closed rather than orphaned.

jhgaylor:
Pushed `54bc82fb`. All five nits addressed, plus one that turned into a behaviour change.

**The one that grew.** Nit 5 asked me to name the first-party surface in the "left out" note. Chasing it found the conversations app is not in `jhgaylor/fountain-conversations` any more (404 since the org move — `CLAUDE.md`'s surfaces table is stale) but in [`managoat/demos` `apps/fountain-conversations`](https://github.com/managoat/demos/tree/main/apps/fountain-conversations). Reading its `src/lib/blocks.ts:104`, it groups the timeline by pairing stage `started` events with their terminal partner, **not** by `turn_id`. It collapses a repeated `started` only when nothing came between the two — and here the two `session`/`done` events do — so a restarted turn would push a *nested* `turn` section whose outer half is never closed and renders as still running.

That is a turn that runs forever, on the exact surface this issue is about. So it is fixed, not documented: `run_fresh_turn/7` publishes `turn`/`started` only on a turn's first attempt, gated on `state.turn_session_retry`. The `session`/`done` restart notice already carries the turn id and says what happened. `sdk/typescript/src/turn.ts:65` no longer emits a second `turn-start` either. New assertion `["started", "done"]` in the end-to-end test, verified by reverting the gate — it fails with `["started", "started", "done"]`.

Thank you for the metering check. `usage.turn_started` comes from `record_turn_usage/1` at turn-*row creation* (`conversations.ex:1283`, `:1356`), not the stage publish, so a restart never meters twice — stated explicitly in the body now, since that is the reading someone will jump to and get wrong.

**Convergence.** Verified your `:new_session` chain myself: `forget_runtime_session/4` nils the id → `session_plan/2` returns `:run` → `Peer.start_session/1` sends `start_new_session/1` under tag `:new_session`, which the clause's `when tag in [:resume_session, :load_session]` does not admit. The restarted attempt cannot re-enter it. The body now leads with the consequence you drew out — `session/new` overwrites the stale placeholder, so the "no rollout ever existed" state that used to fail every prompt forever self-heals on the next one. The flag stays as the brace.

**The other four.**

1. `turn_machine.ex` — the `Logger.warning` moved into the three branches via `log_session_gone/4`; the repeat failure now says "already restarted once; failing the turn".
2. `conversation_server.ex` — `restart_session/2` has a `%{current_turn: nil}` clause that logs and returns state. 2631 lines against the 2774 pin.
3. Billing claim corrected in both the PR body and the docstring: at 25¢/turn-hour a cent buys 144 seconds, so the dead row almost always priced at zero. What changes is that the abandoned attempt's seconds fold into the turn that answered.
4. `_unsafe_list_turn_images/1`'s doc now says the replay is not byte-for-byte the caller's list — `_unsafe_insert_turn_images/2` halts on the first rejected image, so a restart sends the prefix that persisted, which is the more correct of the two and costs nothing since attempt 1 never reached `session/prompt`.

Re-ran clean: `compile --warnings-as-errors`, `format --check-formatted`, `credo --strict` on the changed files, `MIX_ENV=dev mix dialyzer`, and 207 tests / 0 failures across `turn_machine_test`, `conversation_server_acp_test`, `conversation_server_test`, `conversation_server_size_test` and `webhooks/events_test`.

Still owed and now said plainly in the body: the conversations app was read, not run. A build and a click-through on a conversation idle across a restart is worth doing before prod.

jhgaylor:

BinaryBourbon (approved):
Re-reviewed for `54bc82fb`, which landed after my approval of `88576881`. Stale-review dismissal is off on this branch, so that approval was standing against code I had not read. **Re-confirming it.** The behaviour change is correct and better than the thing I endorsed, and I was wrong to wave the double `turn|started` through as documentation — thanks for going and reading `blocks.ts`.

## The behaviour change

`run_fresh_turn/7` now gates the `turn`/`started` publish on `is_nil(state.turn_session_retry)` (`conversation_server.ex:2194-2207`). Verified in both directions by mutation:

```
if true  do   →  1 failure: left ["started","done"]  right ["started","started","done"]
if false do   →  1 failure: left ["started","done"]  right ["done"]
```

So the single new assertion pins the gate from both sides — over-publishing and over-suppressing each red it. That is a better test than I expected from one line.

**Conditioned correctly, by construction.** There are exactly three callers of `run_fresh_turn/7`:

- `run_turn/6:2145` — reached only from `kick_turn/4`, which clears the flag on its first line (`:2112-2113`) before `TurnMachine.open/4`. A genuine turn N+1 publishes `started`.
- `resume_acp_connection/5:2487` — reached from `run_turn/6`, so same clearing.
- `restart_session/2:2530` — the intended suppression; `drive/4` writes `into_state` before reducing effects, so the flag is already set.

Two independent clears keep a stale flag from following a restarted turn into the next one: `kick_turn/4` and `finish/4` (`turn_machine.ex:648`), either of which alone is sufficient.

**The terminal event is not gated.** `finish_acp_turn/4` → `TurnMachine.finish/4` publishes `turn`/`done` unconditionally, which is what `["started", "done"]` asserts.

**Nothing downstream needed the second `started`** — and you are right that my first-round check was made against a stream that had two, so I redid it as "does anything advance on the *second* one":

- `openai_controller.ex:885` binds `turn_id` on the first `started` (clause guarded `%{turn_id: nil}`); the second only ever matched nothing. Binding is unaffected.
- `sdk/typescript/src/turn.ts:65-69` sets `this.started`/`this.turnId` on the first; `applyOutput` gates on `started || turnId`, both already true. Nothing in the SDK reads a second. (Small correction to your note: `turn.ts` is **not** modified by this PR — `git diff --name-only` over both commits does not touch it. The SDK stops emitting a second `turn-start` because the server stops sending a second `started`, which is the better place to fix it.)
- `cli/internal/acp/prompt.go:287` `classifyStage` — non-terminal, `log.Debug` only.
- Webhooks: one `conversation.turn.started` instead of two, and it was never in `@defaults`.
- Metering: unchanged either way, `record_turn_usage/1` fires from row creation (`conversations.ex:1275`, `:1348`).

**Size pin:** `conversation_server.ex` is 2631 lines against `@pin 2774`. Confirmed with `wc -l`, and `conversation_server_size_test` passes.

**Gates:** turn_machine + conversation_server + conversation_server_acp + size + webhooks/events = 207 tests, 0 failures. `mix format --check-formatted` and `mix credo --strict` over the three changed lib files: clean. My four earlier nits are all taken, and the billing docstring rewrite at `turn_machine.ex:696-708` is now accurate.

I agree the build-and-click-through of `apps/fountain-conversations` is owed before deploy rather than before merge. The assertion pins the event sequence the app consumes, which is the part CI can hold.

## Two things for the author

**1. `resume_acp_connection/5` has the same double-`started`, and this gate does not cover it.** Not introduced here — it has been there since #817 — but it is the identical failure you just described, so it should not be left unnamed now that the mechanism is understood:

`Connection.resume/7` publishes `turn`/`started` at `connection.ex:115` (`mode: "continue"`, `connection: "reused"`) **before** calling `Peer.prompt/4`. When the idle peer refuses, `resume_acp_connection/5:2486-2488` drops the connection and calls `run_fresh_turn/7` on the same row — where `turn_session_retry` is nil, so the new gate does not fire and a second `started` goes out. That is the `blocks.ts` nesting shape exactly: one row, two `started`, one `done`, outer section closed by the end-of-stream sweep with `ended: null`.

There is no test on that path at all — `peer_refused_reuse` appears twice in `lib/` and zero times in `test/`. A follow-up issue is the right size for it. If you would rather fix it here, the honest gate is "has this row already announced itself" rather than "was this turn restarted"; `turn_session_retry` happens to answer that question for one of the two paths.

**2. The discriminating case is untested: a genuine turn after a restarted one.** I removed *both* clears (`kick_turn`'s and `finish/4`'s) so the flag latches on forever, and re-ran all of `test/fountain/conversations/` — **939 tests, 0 failures, 2 skipped**. Nothing notices, because no test prompts a second time after a restart. The clears are correct, and there are two of them, so this is belt-and-braces rather than a live bug. But a suppressed `started` on a real turn is a worse failure than the duplicate you just fixed — `blocks.ts` gets a `done` with no partner — and one more `prompt` at the end of `runs the prompt on a fresh session instead of failing the turn`, asserting `["started", "done", "started", "done"]`, would make the gate's condition load-bearing instead of merely correct.

## One nit

`conversation_server.ex:2492-2511` — the new `%{current_turn: nil}` clause is inserted directly under the four-paragraph comment that describes the *main* clause, with no `#` separator line, so "The row is gone. Unreachable: …" reads as the fifth paragraph of it and the whole block now appears to document the guard. Either a bare `  #` between them, or put the guard clause above the preamble.

Approval stands.

jhgaylor:
Pushed `38476e10` — the latch test, plus the comment nit. Flagging it here since stale-review dismissal is off and the approval is still showing green against this.

**Your correction is applied.** The body no longer implies this PR changed `sdk/typescript/src/turn.ts`. It does not — `git diff --name-only 86b9d695..HEAD` touches no SDK file. The SDK emits one `turn-start` because the server sends one `started`; the mapping at `turn.ts:65` is unchanged.

**The latch test, and a trap in its revert check.** Restart a turn, kill the idle peer so turn 2 takes `run_fresh_turn/7` — `Connection.resume/7` does not read the flag, so a resumed turn 2 would pass no matter what — then assert `["started", "done", "started", "done"]`.

Removing the two clears in `ConversationServer` leaves it **green**. That is why the reviewer's 939-test run found nothing holding them: there are **four** clears, and `TurnMachine.finish/4` is the one carrying this path — a turn that completed clears the flag on its way out, before the next prompt ever reaches `kick_turn/4`. With all four gone (`finish/4`, `close_interrupted/1`, `kick_turn/4`, `fail_turn_before_start/6`) the stream is `["started", "done", "done"]` and the test fails. The test's comment says this in place, so the next reader does not repeat the half-test and conclude the clears are unpinned.

`mix test test/fountain/conversations/` — **940 tests, 0 failures, 2 skipped** (the reviewer's 939 plus this one). `compile --warnings-as-errors`, `format --check-formatted` and `credo --strict` clean. 2632 lines against the 2774 pin.

One incidental: `next_write/1` now takes a timeout, defaulting to the old 1s. A turn that has to spawn a fresh adapter waits on `prepare_acp_adapter/3` before its peer writes anything, which outran one second locally — turn 2's four reads pass 5s. Existing call sites are untouched.

**Nit.** The `%{current_turn: nil}` guard clause had inherited the four paragraphs above it; there is a `#` break now and the guard has its own sentence.

**#1924 — agreed and not touched here.** `Connection.resume/7` publishing `started` at `connection.ex:114` before the idle peer may refuse, with `resume_acp_connection/5` then calling `run_fresh_turn/7` on the same row with the flag nil, is the same duplicate by a different route. And the framing is right: my gate asks "was this turn restarted" when the honest question for both paths is "has this row already announced itself" — a row-level fact, probably read off the turn's own state rather than a server flag. Its own PR.

Still owed, unchanged: the conversations app was read (`managoat/demos`, `src/lib/blocks.ts:104`), not run.

BinaryBourbon (approved):
Final pass, for `38476e10`. **Approval stands.** The new test is real, the timeout change is contained, and the `"models"` fixture is required rather than decorative. One correction to the record about my earlier revert, and one thing I got wrong in this round and caught.

## The test

**It is not vacuous in the other direction — and the peer kill is what makes it so.** This is the part I checked hardest, because a turn 2 that reuses the idle peer would go through `Connection.resume/7`, which publishes `turn`/`started` unconditionally and reads no flag, so the assertion would hold against a gate that was broken.

I removed only `GenServer.stop(:sys.get_state(pid).acp_peer)` from inside this test and re-ran it:

```
1) a later turn announces itself again — the restart flag does not latch
   code: %{"id" => init_id, "method" => "initialize"} = next_write(5_000)
```

Turn 2 goes straight to `session/prompt` on the peer the restart left behind, so the `initialize` match fails. The kill is load-bearing, and `initialize` + `session/resume` on the wire is itself the proof that turn 2 took `run_fresh_turn/7`. Confirmed the harness genuinely supports the other path too — `a second prompt reuses the connection — no handshake (#817)` at `:421` asserts exactly that — so this is not an artefact of the fixture never keeping a peer alive.

**The four-clear revert reproduces.** Measured on this commit, one test each:

| Mutation | Result |
|---|---|
| all four (`finish/4`, `close_interrupted/1`, `kick_turn/4`, `fail_turn_before_start/6`) | fails, `right: ["started", "done", "done"]` |
| the two in `ConversationServer` (`kick_turn/4` + `fail_turn_before_start/6`) | passes |
| baseline | passes |

Both match what the commit message and the test's own comment say. The comment is accurate as written — "removing only the two in the server leaves it green" — and it is the right thing to have recorded in place.

**One correction to the record on my earlier revert.** The mutation I ran last round was not the two `ConversationServer` clears; it was three — `kick_turn/4` plus *both* `turn_machine.ex` clears, because the `finish/4` and `close_interrupted/1` lines are byte-identical and my substitution was global. So it did include `finish/4`. Re-running that exact three-clear mutation against this commit fails the new test identically (`right: ["started", "done", "done"]`). My run came back clean last round for the reason I reported — no test prompted a second time after a restart — not because the mutation missed the clear that matters. The remaining gap in what I ran was `fail_turn_before_start/6`, which does not carry this path either way. Worth being exact about, because "the check was too small" and "the check was aimed at something the suite did not yet cover" call for different fixes, and this was the second.

**And one I did get wrong this round.** My first attempt at the vacuity check used a `perl -0pi -e 's/…/…/'` without `/g`, which replaced the *first* `GenServer.stop(:sys.get_state(pid).acp_peer)` in the file — one at `:320`, in an unrelated test — and left the new test untouched. It came back green and I nearly reported the kill as not load-bearing. I only caught it by reading the mutated region instead of trusting the exit code. Same shape as everything else in this batch: `git diff` said one line had changed, and it was the wrong line.

## The other items

**`next_write/1`.** `defp next_write(timeout \\ 1_000)` — the default is unchanged, and `git diff 88576881 HEAD` over the test file shows the `defp` line as the only `-` line touching it. All 54 bare `next_write()` calls keep the one-second bound; the only explicit timeouts are the four at `:1239-1250` in the new test. The three other files that use a `next_write` (`conversation_server_acp_runtime_test.exs`, `conversation_server_broker_test.exs`, `acp_model_contract_test.exs`) each define their own private copy, still `assert_receive {:wrote, line}, 1_000`, so nothing outside this file moved. The justification holds too: turn 2 spawns a fresh adapter and waits on `prepare_acp_adapter/3` before the peer writes anything, which is not bounded by the same second a write on an open connection is.

**`%{"models" => %{}}` on the `session/resume` answer is required, not an assumption.** `handle_response(:resume_session, result, state)` (`managoat_acp/peer.ex:808`) hands the result to `pin_model_then_prompt/2`, the same function `:new_session` uses, and that `cond` selects `session/set_model` on `Map.has_key?(result, "models")`. With a model configured and neither `"models"` nor `configOptions` in the response, the final arm is `fail(state, {:model_selection_failed, …})` — so without the key the peer would not prompt at all and the test could not reach its own assertion. It mirrors what `drive_to_prompt/2` already puts in the `session/new` answer, and upstream calls `pin_model_then_prompt/2` on a resume unconditionally, so a resume response carrying models is the shape the library expects rather than one the suite invented.

**Size pin:** 2632 against `@pin 2774`, by `wc -l`, and `conversation_server_size_test` passes.

**Comment nit taken** — `conversation_server.ex:2507` now separates the guard clause's paragraph and names it, so the preamble no longer reads as its documentation.

**My SDK correction is applied** — the body now says plainly that this PR does not touch the SDK and that the single `turn-start` follows from the server sending one `started`.

**#1924** is filed with the right framing and correctly out of this PR: `Connection.resume/7` publishes at `connection.ex:115` before `Peer.prompt/4`, and the refusal arm at `:2486-2488` re-runs the row through `run_fresh_turn/7` with the flag nil. That path has no test at all (`peer_refused_reuse` appears twice in `lib/`, zero times in `test/`), which is its own reason not to bolt a fix onto this PR.

## Gates

208 tests / 0 failures across turn_machine, conversation_server, conversation_server_acp, the size test and webhooks/events; the full `test/fountain/conversations/` tree at 940 tests, 0 failures, 2 skipped — matching the number in the commit message. `mix format --check-formatted` and `mix credo --strict` over the changed lib files: clean. Worktree restored to `38476e10` with an empty `git status` after every mutation.

Nothing further owed here beyond the build-and-click-through of `apps/fountain-conversations`, which the body records. Approving.

BinaryBourbon (commented):
Reviewed `38476e10fa7c87d4bdbe4b5064e00ff9a1cbb556`.

No actionable findings. Traced the ordered forget/drop/restart effects, stale peer-reference handling, reused turn/prompt/images, retry-flag clearing and suppression of a second `turn|started` event. The lost runtime memory is disclosed on the turn that actually runs.

Validation: source and test review at this head, including the restart integration cases and relevant ACP/connection behavior. ACP integration tests and a live conversations-app scenario were not rerun.

BinaryBourbon (approved):
**Approved** at `38476e10fa7c87d4bdbe4b5064e00ff9a1cbb556`.

The restart correctly retains the turn row/prompt/images, clears the missing runtime session, closes the abandoned peer/span, and emits one `turn|started` event. Rechecked the fresh-spawn latch regression and flag resets, stale-peer filtering and current review discussion; no blocking code finding.

Validation: [CI run 34587862230](https://github.com/managoat/fountain/actions/runs/34587862230) is verified at this exact head; six core partitions pass (5,211 tests in downloaded logs), and static, contract and required gates pass. I reviewed source/tests and CI logs; I did not rerun the live conversations app.

This code approval preserves the owner's deliberate **draft/merge-queue hold**. The documented conversations-app build and click-through on a conversation idle across restart remains a deployment prerequisite; it has not been completed in this review.

BinaryBourbon (commented):
Burndown verdict at `38476e10fa7c87d4bdbe4b5064e00ff9a1cbb556`: the existing code approval stands; this draft is not ready to merge into current main.

I rechecked the ordered forget/drop/restart effects, same-row prompt/image replay, retry reset paths, one-started-event gate and ACP regression tests. No new defect found in this head. Current-head `CI required` is green. This pass was source/test and CI-state review; I did not rerun ExUnit or the conversations app.

The integration work is substantive. Against main `bf8f890a5408119b97247aebfca6169d9989528f`, an isolated `git merge-tree` finds conflicts in `conversation_server.ex`:

- Preserve the `TurnLaunch.run` extraction and move the one-started-event gate into `turn_launch.ex`; retain bounded-execution cleanup when clearing `turn_session_retry` in `fail_turn_before_start`.
- Also handle the non-conflicting behavioral overlap: main's bounded `drop_connection` calls `BoundedTurn.retire`, which completes the execution as interrupted, finishes the turn and clears the connection. The proposed forget/drop/restart sequence would then reach the nil-turn guard without retrying. The engineering default should preserve the same admitted turn across attempt cleanup, retaining its execution fencing and confirmed-stop rules, with a missing-session regression through the bounded path. The current-head approval does not approve a hypothetical merge that has not implemented and tested this integration.

After the integration, rerun restart/next-turn ACP regressions, bounded-turn lifecycle tests and required CI, then review the new head. Before deployment, complete the documented conversations-app idle-across-restart click-through. No branch or draft state changed by this review.

BinaryBourbon (approved):
Approve `4f788a274876f8edbf88b59221cd679a6853962a` after independent source and integration review.

A missing resume/load session gets one fresh-session attempt on the same initialized ACP peer. The repair preserves the admitted turn, original input/options, command/transport, journal identity, deadline and execution limits, without a second turn-start event or command spawn. Published managoat_acp 0.4.2 provides the guarded pre-prompt restart; its independent review and release validation are recorded in managoat/managoat_acp#15.

The server requires a successful persisted session-clear fence before requesting restart. A rejected fence abandons only the stale actor's local state and cannot fail or idle a successor turn. I reviewed the bounded real-peer regressions, original-deadline enforcement, legacy successor fixture and one-retry/next-turn behavior. The retry detail is classified as secret in the exhaustive state guard and actual OTP status/crash coverage. The final merge preserves main's provisioning/wake retirement guards and both changelog entries.

No actionable source defect remains. Validation attribution: the coder's completed full precommit at this exact head passed compile, formatting, strict Credo, Dialyzer, Sobelow, production release assembly, 6 doctests + 5,628 core tests, 144 Buzz and 33 Support tests, all zero failures. The original native-docs timeout and unchanged rerun conditions are retained in the PR evidence; no assertion or project timeout was relaxed. I reviewed the source/tests rather than independently repeating that full suite.

I also exercised the actual production-built conversations client with a mock API and live SSE: recovery showed one completed turn, the lost-memory notice in Timeline, no extra running section, and a subsequent prompt completed normally. The app's 29 tests and build passed. This was a local client smoke test, not a real production restart; host/peer integration is covered by the server regressions.

Land after #1914 (with prerequisite #2047 already merged), and only with required current-head and merge-group checks passing.

BinaryBourbon (approved):
Approved after independent review of final integration 994c869587ac9e378073e2d8e621d352681714f5 against landed main 721bb174.

The signed merge retains all session-recovery redaction sentinel fixtures and assertions in both test conflicts, and keeps one recovery entry and one deadline entry in the changelog. Production recovery/fencing logic is unchanged from my previously reviewed head. Main's retired-or-resetting wake guard and identity-checked MachineEvents reset handling are preserved. The main-based diff is scoped to the 11 recovery files. No actionable findings.

The same initialized peer and admitted bounded turn survive one missing-session retry, retaining the original deadline and writer fence; rejected persisted session fences cannot restart or disturb a successor. The new recovery detail remains scrubbed from OTP status/crash reports. Released ACP 0.4.2 is locked.

I inspected the completed final-tree precommit log: static/type/security and production release gates passed, followed by 6 doctests and 5,698 core tests, Buzz 144 and Support 33 tests, zero failures (2 configured skips, 7 exclusions), seed 299955/max-cases 4. The coding agent ran the gate; I independently reviewed source, conflict resolutions and results. The production-built client smoke used a mock API/live SSE, not a production restart. Required PR and merge-group CI still gate landing normally.

Issue #826, opened 2026-08-19
api: GET /api/search — search across a user's conversations and turns


jhgaylor:
## What

A tenant-scoped search endpoint for the standalone clients (fountain-team, fountain-conversations), which today can only filter what they have already listed:

```
GET /api/search?q=<text>&limit=20[&agent_id=][&conversation_id=][&since=]
```

Returning hits over conversation titles, turn prompts, and assistant text (the `text` blocks of `Fountain.Conversations.Blocks`), each with enough to jump there:

```json
{"data": [{"conversation_id": "…", "agent_id": "…", "turn_id": "…", "kind": "prompt|reply|title",
           "snippet": "…", "ts": "…"}],
 "meta": {"limit": 20, "has_more": false}}
```

Postgres full-text (`to_tsvector`/`websearch_to_tsquery`) over `turns.prompt` + `conversations.title` is enough for a first cut; searching assistant text needs either a materialised per-turn text column (written when a turn ends, from the same block parse the API's `?blocks=true` uses) or a trigram index over `log_events.data` scoped to `stream = 'acp'` — the former is cheaper to query and does not index tool noise.

## Why

OpenMausBot's command palette (`/api/search` over messages, "jump to the message") is one of the features worth bringing to fountain-team; there is nothing on the API to build it on. Client side: jhgaylor/fountain-team issue (linked below).

Rate-limit like the other list endpoints; never index across tenants.

jhgaylor:
Client side: https://github.com/jhgaylor/fountain-team/issues/5

PR #1914, opened 2026-09-11
fix(conversations): keep brokered values, the proxy session, env credentials and resolved MCP servers out of crash reports (#1690)

### LINKED ISSUE #1690: ConversationServer crash reports print brokered values, the broker session, env credentials and resolved MCP servers
## Symptom

An unhandled raise in any ConversationServer callback prints `State:` in the crash report and the Sentry event with plaintext secrets, the #315 shape the redaction exists to prevent.

## Mechanism

`redact_state/1` at `apps/fountain/lib/fountain/conversations/conversation_server.ex:2242-2250` scrubs `sprite_env`, `tenant_key`, `inference_credentials` and `callback_token` only. Four state fields hold plaintext and are not scrubbed:

- `brokered` (`:447`, `:1441`): the real values of every brokered key (GITHUB_TOKEN, connection tokens, inference keys). `Broker.split` puts placeholders in the env and the values here.
- `broker` (`:466`): the proxy session including its token.
- `env_credentials` (`:462`, `:710`).
- `resolved_mcp_servers` (`:457`): the `${VAR}`-substituted MCP headers and env, added by #1511 this week.

`conversation_server_redaction_test.exs:56-62` asserts none of those values. `brokered`/`broker` date from #1136/#1150; #1511 added a fourth field without extending the redaction.

## Fix

Extend `redact_state/1` to the four fields and add four `refute` lines to the redaction test. Consider a guard test that every state field whose name matches `cred|token|secret|broker` is covered.

---

jhgaylor:
ConversationServer crash reports could expose broker credentials, resolved MCP secrets, adapter client credentials, turn content and buffered sandbox output. This change scrubs those fields through `format_status/1` while retaining identifiers, parameter keys and buffer size for diagnosis.

Cached ACP permission requests (`{request_id, params}`) are deeply redacted while retaining the request ID and keys. Bounded execution journals, including provider error text, are replaced wholesale. An exhaustive state-field guard requires every new server field to be classified.

The integration includes prerequisite #2047: required CI on the initial repair exposed an inherited early-deadline wake that could interrupt an unexpired turn. The prerequisite rechecks the absolute deadline and has a deterministic failing-before/passing-after regression. Land this PR after #2047.

Validation at signed `f05ade96530af653abc08a615cb0fc53aef1ed41`: all `mix precommit` stages passed on OTP 28.3 / Elixir 1.19.2, including compile, unused dependencies, formatting, strict Credo, Dialyzer, Sobelow, production release assembly and all suites: **6 doctests + 5,621 core tests, 144 Buzz tests, 33 Support tests; zero failures** (2 skipped, 7 excluded). The final run used seed `259632`, `+S 4:4` and `max_cases: 4` with normal assertions/timeouts. Redaction coverage includes structured status, nested values, the exhaustive guard, and two actual OTP callback-crash reports covering lifecycle corruption and the bounded journal.

Retained integration evidence: a preceding 58-test focused run encountered a DBConnection client-exited failure in an inherited 100 ms transport timeout fixture; a 17-test same-seed diagnostic passed that case but hit the 200 ms registration race tracked in #2024. Neither assertions nor timeouts were changed. The completed final full suite above includes both cases unchanged.

Closes #1690


BinaryBourbon (changes requested):
The four scrubs are right and the guard is genuinely fail-closed — I broke it
on purpose and it went red naming the field. My changes are all in
`@plaintext_fields`: two entries there hold credential-class material that I
think has to be scrubbed, and the PR's own argument is what convinced me.

## What I verified, and how

Everything from the worktree at `p0/1690` (4563c8a), `apps/fountain`, mise
toolchain, a dedicated test DB.

- **Baseline.** `mix test test/fountain/conversations/conversation_server_redaction_test.exs test/fountain/conversations/redaction_test.exs` — 17 tests, 0 failures. `mix format --check-formatted` clean.
- **The guard is fail-closed.** Added `fake_new_leaky_field: "some-plaintext-token"` to `init/1` at `conversation_server.ex:428`. Both guard tests went red, each naming `[:fake_new_leaky_field]` and nothing else. Removed; tree is clean (`git status --porcelain` empty).
- **Per-scrub reverts, four of nine.** Replaced each with the identity in turn:
  - `brokered`, `broker`, `resolved_mcp_servers` → 2 failures each (the `format_status` test and the sentinel guard).
  - `handle` → **1 failure, the guard only.** That confirms your claim about the pre-existing assertion.
- **The vacuous assertion, independently.** `Managoat.Sandbox.Handle` at `deps/managoat_sandbox/lib/managoat/sandbox/handle.ex:16` is `@derive {Inspect, only: [:provider, :name]}`, and `Sprites.Client` (`deps/sprites/lib/sprites/client.ex:12`) has no derive of its own. Rendering a handle wrapping `%Sprites.Sprite{client: %Sprites.Client{token: tok}}`:
  - `inspect(handle) =~ tok` → `false`
  - `inspect(handle, structs: false) =~ tok` → `true`

  So `refute rendered =~ @sprites_token` could not fail before, and `structs: false` is what makes it a real assertion. Your call.
- **`deep_redact/1` debuggability.** Fed the real shape through `server_state/1`: `%{"linear" => %{"url" => ..., "headers" => %{"Authorization" => "Bearer T"}}}` comes back as `%{"linear" => %{"headers" => %{"Authorization" => "[REDACTED]"}, "url" => "[REDACTED]"}}`. Server name, `headers`, `Authorization`, `env`, `LINEAR_TOKEN` all survive. Contract holds. `resolved_mcp_servers` is `agent.mcp_servers` (`conversation_server.ex:741`), and `Agent.mcp_servers` is `field :mcp_servers, :map` (`agent.ex:45`) — JSON-shaped, so maps/lists/binaries/numbers is the full reachable set and the clauses cover it. Terminates on all of them.

## Findings

### 1. `runner_replay` holds unredacted sandbox output — `conversation_server_redaction_test.exs:214`

`runner_replay` is `%{previous_id: integer, buffer: binary}` (`runner_replay.ex:15`), and `buffer` is raw bytes straight off the runner — `feed/2` at `conversation_server.ex:1643` appends `data` before anything logs it, up to `@max_line_bytes` = 4 MiB.

The allowlist reasons that the transcript path is protected by `Redaction.redact/2` at the single log writer. It is — but that is the *persisted* path, and these bytes have not reached it yet. That is the whole point of the buffer: it is the staging area before `log_with_replay_skip/3`. So a crash inside a runner reattach window (`reattachment.ex:61-82`, where `runner_replay` is set) prints, verbatim, the bytes that `redact/2` would have scrubbed on their way to `log_events`.

And this module's own moduledoc names exactly that content as credential-bearing — "an `env`, a `set -x`, a `cat .env` in someone's `setup_script`, or an agent that simply prints its environment". On the ACP path the tool results carrying that output *are* the bytes in this buffer.

Concrete failure: a raise during a runner reattach (the boundary-missing path at `conversation_server.ex:1912` is right there) puts up to 4 MiB of unredacted tool output into the crash report and the Sentry event body. The fix keeps the debugging signal — `previous_id` is what you actually want when a boundary scan goes wrong:

```elixir
runner_replay: state.runner_replay && %{state.runner_replay | buffer: @placeholder},
```

and moves the field into `sentinel_shapes/0` as `%{previous_id: 1, buffer: @sentinel}`.

`replay_dedup` in the same comment group I have no problem with: it holds `acp` lines read back out of `log_events`, which went through `redact/2` on the way in.

### 2. `current_command` is `handle` again, and the allowlist defends it with the argument the PR rejects — `conversation_server_redaction_test.exs:206`

The allowlist entry says `current_command`'s private is excluded from `inspect/1` by `@derive {Inspect, only: [:provider, :ref]}`. That is accurate (`deps/managoat_sandbox/lib/managoat/sandbox/command.ex:15`). But the PR body argues, correctly, that "a `@derive` on a dependency's struct is not a guarantee this server gets to make" — and scrubs `handle.private` on those grounds even though `Handle` carries the same kind of derive.

`current_command.private` is `%Sprites.Command{sprite: %Sprites.Sprite{client: %Sprites.Client{token: ...}, environment: ...}}` (`sprites.ex:413`, `deps/sprites/lib/sprites/command.ex:23`, `sprite.ex:6`). It is the same `%Sprites.Sprite{}` that `handle.private` holds. I rendered one:

```
current_command, plain inspect:   false
current_command, structs: false:  true      # the platform Sprites token
current_command, sprite env leak: true      # sprite.environment
```

So today it is not a live leak — and that is exactly the position `handle` was in before this PR. The asymmetry that bothers me is what happens next: if `managoat_sandbox` ever drops or widens that `@derive` (a dependency change, not a Fountain one), the platform token goes into every crash report and **no test here goes red**, because `:current_command` is allowlisted. The same change to `Handle` *is* caught, by this PR's guard. Same one-line shape as `handle`:

```elixir
current_command: state.current_command && %{state.current_command | private: nil},
```

`provider` and `ref` survive, so matching a report against `current_command_ref` still works.

## Two things I am not asking you to change

- **`current_turn` and `caller_calls`.** I think the call is defensible — it is tenant content, not credential material, and scrubbing the prompt costs the most debuggability of anything here. Worth knowing, though, that `%Turn{}` has no `@derive` of its own, so `prompt` and `reply_text` render in full in the plain-`inspect` crash report and therefore in the Sentry event. If that is a decision rather than an oversight (it reads like one), it is fine as written; I would take a follow-up that scrubs `prompt`/`reply_text` and keeps `id`/`turn_number`/`status`/`acp_prompt_id`.
- **`deep_redact/1` tuples and charlists.** `deep_redact(other)` at `redaction.ex:213` returns tuples and integer lists unchanged: `%{"a" => {:secret, "TOKEN"}}` and `[headers: "Bearer TOK"]` both survive intact (measured). Unreachable from an Ecto `:map` field today, so not a bug — but this is the one field in the module whose shape you deliberately do not want to assume, and a `deep_redact(t) when is_tuple(t)` clause is one line. Your call.

## Nits

- The comment group at `conversation_server_redaction_test.exs:209-215` says "the two reattach buffers" and then lists `replay_skip` alongside `runner_replay` and `replay_dedup`. `replay_skip` is a byte count per stream (`conversation_server.ex:501`), not a buffer — it belongs in the bookkeeping group above.

Happy to re-review quickly on the two scrubs.

jhgaylor:
Both blocking findings were right, and the `runner_replay` one was my own
reasoning applied where it did not hold. Fixed in 44f19c2a.

**`runner_replay`.** The justification I gave it covers `replay_dedup`, which
is read back from `log_events` by `_unsafe_recent_output_lines/3` and has
therefore been through `Redaction.redact/2`. It does not cover
`runner_replay`: `RunnerReplay.feed/2` is called from
`handle_info({:stdout, …})` on bytes that have not been persisted yet, so the
log writer never sees them. The buffer is now replaced with its byte size;
`previous_id` and the size stay, so a reattach crash is still diagnosable.

**`current_command`.** Scrubbed the same way `handle` is. You are right that
defending it with the dependency's `@derive` is the argument this PR rejects —
and the guard proves the point: with the scrub removed, `current_command`
fails **only** the guard test, exactly like `handle`.

Also taken, both of which you flagged as my call:

- **`current_turn`.** Rather than naming `prompt` and `reply_text`, the turn is
  rebuilt from a keep-list of identity fields (`id`, `conversation_id`,
  `turn_number`, `status`, `origin`, `exit_code`, `acp_prompt_id`, the
  timestamps). Everything else is replaced, so a text field added to the schema
  is redacted the day it appears instead of relying on someone noticing. Prompt
  text reaching a third-party processor was the deciding factor.
- **The `deep_redact/1` tuple and charlist gap.** Both are covered, with three
  tests in `redaction_test.exs` that reach `server_state/1` directly (the state
  field cannot carry either shape today, so a live-server test could not get
  there). Removing the tuple clause fails two of them; removing the charlist
  clause fails the third.

**Nit taken:** `replay_skip` is no longer grouped with the buffers; the comment
says it is a byte count taken from rows already written.

`@plaintext_fields` is down to `caller_calls`, `replay_dedup` and the
ids/config/plumbing groups, with the `replay_dedup`-versus-`runner_replay`
distinction written into the comment.

Verified the same way as the first round: each new scrub reverted in turn fails
at least one test, and the field guard is still fail-closed — a temporary extra
state field in `init/1` turns both guard tests red naming only it. 20 tests, 0
failures; compile, format and `credo --strict` clean. PR body updated to match.

BinaryBourbon (approved):
Both blocking findings are fixed, and I like the `current_turn` inversion better
than what I suggested. Approving. Everything below is verified against the
committed `44f19c2a`, not the description.

## The aborted-script window: nothing was lost

I diffed the clause lists rather than reading the file, as asked —
`4563c8af..44f19c2a` over `redaction.ex`, every `def`/`defp` line:

```
- defp deep_redact(list) when is_list(list), do: Enum.map(list, &deep_redact/1)
+ defp deep_redact([_ | _] = list),
+ defp deep_redact(list) when is_list(list), do: list
+ defp deep_redact(tuple) when is_tuple(tuple),
+ defp redact_each/1, charlist?/1, turn/1 (×3), replay/1 (×3)
```

One removal, and it is the clause the two new list clauses replace. Everything
else is purely additive, and the whole commit is 4 files. `mix compile
--warnings-as-errors` is clean, which is also where a clause left unreachable
by a partial restore would have shown up. Both the tuple and the charlist
clause are present and behave.

## Verified

From the worktree at `44f19c2a`, `apps/fountain`, mise toolchain, dedicated
test DB. Tree clean after every experiment.

- `mix test test/fountain/conversations/conversation_server_redaction_test.exs test/fountain/conversations/redaction_test.exs` — **20 tests, 0 failures** (17 before, +3 shape tests). `mix format --check-formatted` clean.
- **The guard is still fail-closed after the allowlist shrank.** Re-added `fake_new_leaky_field: "some-plaintext-token"` to `init/1`; both guard tests go red naming `[:fake_new_leaky_field]` and nothing else. Restored.
- **Revert check on all three new scrubs.** `current_turn` → 2 failures. `runner_replay` → 2 failures. `current_command` → 1 failure, **the guard only** — which is right, and is the same signature `handle` has, since `@derive` hides it from the plain `inspect` the other test renders. That asymmetry existing and being *caught* is the whole point of `structs: false`.
- **Behaviour, measured through `server_state/1`:**

| input (`resolved_mcp_servers`) | output |
|---|---|
| `%{"a" => {:bearer, "SECRET"}}` | `%{"a" => {:bearer, "[REDACTED]"}}` |
| `%{"a" => [headers: "Bearer SECRET"]}` | `%{"a" => [headers: "[REDACTED]"]}` |
| `%{"a" => ~c"SECRET"}` | `%{"a" => "[REDACTED]"}` |
| `%{"a" => [{:k, %{"x" => "SECRET"}}]}` | `%{"a" => [k: %{"x" => "[REDACTED]"}]}` |
| `%{"linear" => %{"url" => …, "headers" => %{"Authorization" => …}}}` | keys and nesting intact, leaves gone |

  `runner_replay` → `%{buffer: "[REDACTED] 17 bytes", previous_id: 41}`, and a
  shape without `:buffer` falls to `secrets/1` rather than raising. That is the
  right fallback ordering.

- **The `replay_dedup` / `runner_replay` distinction is correct**, and I checked it rather than taking the comment: `replay_dedup` is filled from `log_events` rows, which passed `redact/2` on the way in; `runner_replay.buffer` is appended straight from `handle_info({:stdout, …})` at `conversation_server.ex:1643`, before anything persists. The comment at `conversation_server_redaction_test.exs:243-248` now says exactly that, which is the part a future reader needs.

## The `@turn_identity_fields` keep-list — I checked every entry

The inversion is the better design: keeping identity rather than naming content
means a text column added to `turns` is redacted on the day it lands, and the
schema has form here (`reply_text` arrived with #826, `model_selection` later).
Rendered a fully-populated `%Turn{}` through it — `prompt`, `reply_text`,
`pending_permission`, `usage` and `model_selection` all become `[REDACTED]`,
`__meta__` and the unloaded associations are left readable, and the eleven
identity fields survive.

On the two you flagged:

- **`origin`** — `field :origin, :string, default: "user"` with `@origins ~w(user autonomous)` and `validate_inclusion(:origin, @origins)` at `turn.ex:80`. A closed two-value enum, not free text. Safe.
- **`acp_prompt_id`** — `field :acp_prompt_id, :integer`, the JSON-RPC id of the in-flight `session/prompt`. A counter. Safe, and it is one of the few things a reattach crash is actually diagnosed from.

The rest: `id`/`conversation_id` are UUIDs, `turn_number`/`exit_code` integers,
`status` a validated enum (`turn.ex:79`), and the four timestamps are
`:utc_datetime` (the schema is `updated_at: false`, so there is no twelfth
timestamp missing from the list). Nothing on the keep-list can carry tenant
text.

## Three nits, none blocking, none worth another round

1. **A short all-small-integer list is read as a charlist.** `List.ascii_printable?([80, 90])` is true, so `%{"timeout_ms" => [80, 90]}` renders as `"[REDACTED]"` while `%{"ports" => [8080, 9090]}` survives. Over-redaction, so the safe direction, and an MCP document of printable-range integers is not a real shape. Mentioning it only so the asymmetry is not a surprise later.

2. **`usage` and `model_selection` are collateral.** Both are non-content — token counts, and the provider/model the turn actually ran on — and both now read `[REDACTED]`. That is the price of the keep-list and I would pay it, but if a model-selection crash ever needs diagnosing, adding those two to `@turn_identity_fields` is the supported move and does not weaken the property.

3. **An improper list still raises, which is the one failure mode this module calls out by name.** `%{"a" => [1 | 2]}` and `%{"a" => ["SECRET" | "tail"]}` both match `deep_redact([_ | _])`, fail `charlist?`, and raise `FunctionClauseError` inside `Enum.map` — and a raise in `format_status/1` is OTP reporting the unredacted state, per the `secrets(%_{})` comment and the guard's own flunk message. Pre-existing (the old single `Enum.map` clause did the same), and unreachable from an Ecto `:map` — but so were tuples and charlists, and those clauses exist "for the day the field's shape changes". If you want the class gone rather than the shapes enumerated, the move is a `try/rescue` at the `format_status/1` call site (`conversation_server.ex:2047`) returning a wholly-scrubbed state on any raise. Follow-up, not this PR.

Good fix. The guard is the durable part of it.

jhgaylor:

BinaryBourbon (commented):
Reviewed `a1b0ecb18f24c791f7c6852cc9245a56dd6f8138` for stack #1690.

No actionable defect introduced by this PR found. Reviewed the redaction implementation, ConversationServer field shapes and format-status call site, and the new field-classification guard.

Validation: an isolated Elixir smoke check against the exact-head redaction module removed sentinels from environment, broker, credential, resolved MCP, and replay-buffer fields. This used a minimal Ecto association stub; full server crash/DB tests were not rerun.

BinaryBourbon (changes requested):
Request changes on `a1b0ecb18f24c791f7c6852cc9245a56dd6f8138`. This supersedes my earlier no-findings comment; that review did not inspect the failing integration checks.

**P1 — Redact the existing acp_request_params field before returning server state** (`apps/fountain/lib/fountain/conversations/redaction.ex:185–189`). ConversationServer already retains the last session/request_permission pair as `{request_id, params}`. DetachedRequest.request_line/2 extracts the complete params from the original relayed data, including tool arguments. server_state/1 leaves this field untouched, so an unhandled callback crash can print those arguments into the crash report/Sentry event even though the other request/turn content is scrubbed. Add a redacted representation and a matching sentinel shape; whitelisting these arguments as plaintext would preserve the leak.

This field predates the PR; the integration defect is that the new exhaustive field guard already fails on it and the redaction change leaves it exposed. [CI run 34593207530](https://github.com/managoat/fountain/actions/runs/34593207530) tested merge `7f87b873c802ab8a77c60850218ca033e5632b6f` of this exact head into `2c1c9a7ff0a839d22a8dd943d6bd55a7291c5c19`: partition 5 reports **858 tests, 2 failures**, both naming acp_request_params, and CI required fails. I also reproduced a synthetic secret surviving server_state/1 against the exact-head module in an isolated Elixir process. Full application tests were not rerun locally.

BinaryBourbon (changes requested):
Short-stack landing review of `a1b0ecb18f24c791f7c6852cc9245a56dd6f8138`: **request changes**. The existing blocker remains reproducible; the head has not changed since the previous review.

**P1 — Scrub `acp_request_params` before returning crash-report state** (`apps/fountain/lib/fountain/conversations/redaction.ex`, `server_state/1`). The server retains `{request_id, params}` from the relayed `session/request_permission` request. Those params can contain tool input and credentials, and this redactor leaves them intact. The new exhaustive classification/sentinel guard also fails for this field.

I loaded the exact-head redaction module in an isolated Elixir process using the existing Ecto beam, supplied a synthetic token inside `{17, %{"toolCall" => %{"rawInput" => %{"token" => sentinel}}}}`, and confirmed it survives `server_state/1`. A temporary in-memory variant using `acp_request_params: deep_redact(state.acp_request_params)` removes the token while retaining the request ID and parameter keys. This is a focused reproduction, not a full application test run or a committed fix.

The [failing partition-5 CI job](https://github.com/managoat/fountain/actions/runs/34593207530/job/103243164527) independently reports 858 tests, 2 failures, both naming `acp_request_params`.

To land: add the scrub, add its real tuple/nested-input shape to `sentinel_shapes/0` and the status/crash assertions, run the focused redaction/server tests and required checks, then obtain a fresh approval. There is no product decision needed for this repair. The earlier batch hold remains a separate maintainer landing decision.

BinaryBourbon (approved):
Approve `9787752bd1c855c7802bf443296ecd99a7385390` after independent review of the repaired current-main integration.

The two sensitive fields introduced on main are now covered: cached ACP permission parameters are recursively scrubbed using their actual `{request_id, params}` shape, and the bounded execution journal is scrubbed as a whole, including provider `last_error` text. The added plaintext classifications are backed by the producers: `configuration_revision` is an integer and `execution_transport` is a process handle.

The tests retain the existing lifecycle crash case and add a distinct bounded callback crash with the journal and request still present. They also check structured status output and the exhaustive state-field classification guard. That avoids obtaining a passing crash assertion by clearing the sensitive journal before the crash.

No actionable defect found in the final repair. Validation attribution: I reviewed the exact committed source, field producers, callback gate and tests; the coder reports 21 focused tests passing, including both actual OTP crash paths. I did not independently rerun those tests. The coder also reports full precommit passed: 6 doctests and 5,601 core tests, 144 Buzz tests, and 33 Support tests, all with zero failures; static and production-release checks passed. Current-head CI must pass before merge.

BinaryBourbon (approved):
Reviewed exact head f05ade96530af653abc08a615cb0fc53aef1ed41 independently. No actionable findings.

The redaction changes preserve useful identifiers while scrubbing broker/MCP/adapter credentials, cached ACP request payloads and the complete bounded execution journal. Current server fields remain exhaustively classified; actual OTP callback-crash coverage includes the journal and request secrets. The integration is an exact conflict-free merge of independently reviewed deadline prerequisite e627aaf9: recomputed merge tree matches committed tree139000c363139f692027515cb4e47e4d86bf7a56.

I ran final mix precommit at this unchanged head: all static checks and production release assembly passed; 6 doctests + 5,621 core tests + 144 Buzz + 33 Support tests, zero failures (2 skipped, 7 excluded), process exit0. Run used +S4/max_cases4/seed259632. Earlier 100–200ms transport fixture failures remain recorded in the PR body; assertions and timeouts were unchanged and both cases passed in the final complete run.

Approved. Preserve the documented landing order after prerequisite #2047.

PR #1915, opened 2026-09-11
fix(sandbox-files): redact before the caller's byte cap, not after

### LINKED ISSUE #1907: Caller-controlled max_bytes truncates before redaction on /file, exposing secret values a byte at a time
Output caps in `Fountain.SandboxFiles` are applied by the **shell** (`head -c`), before redaction runs in Elixir. `redact_with/2` is `:binary.replace/4` against the sandbox's known secret values, so it can only match a value present **whole**. When the cut lands inside a secret, the surviving prefix matches nothing and passes through into the response.

> **Updated after review of #1596.** This was first filed as a bounded, incidental leak. That was a grade too mild. On `GET /api/sandboxes/:id/file` **the cut position is chosen by the caller**, which turns an accidental boundary case into a repeatable one.

## The caller-controlled case (`/file`)

`read/3` takes `max_bytes` from `params["max_bytes"]` (`sandbox_files_controller.ex:103`), clamps it to 1…4 MiB, passes it to `head -c "$n"` **in the shell**, and redacts only what survived the cut. Because the caller picks `n`, they pick where the truncation falls relative to a secret's bytes — and a request whose `n` stops short of the full value never triggers a replacement at all. Repeating the read with successive values of `n` therefore reveals the value one byte at a time, and no single response ever contains enough of it to match.

This is a read of exactly what ADR 0039 decision 5 and the module's own docstring say is prevented.

**This is on `main`, not on a branch.** `/file`, its `max_bytes` parameter and this ordering all shipped in the original ADR 0039 PR and are published in **SDK 1.15.0**. #1596 did not introduce it; it surfaced it while fixing an adjacent defect in the same catch-all (the byte cap splitting a multi-byte character into invalid UTF-8, giving a 500 instead of a 422 — fixed in `5c081faa`).

Worth noting who holds the key: a full-scope token includes one issued to a third-party OAuth app the user authorized, which is a different principal from the tenant whose vault values are being protected.

## Other affected caps

- the failing-command diagnostic path (`head -c 4096` on git's stderr) — fixed cap, not caller-controlled, so exploiting it means getting git to place a secret across byte 4,096 of its own error message
- the diff cap

`/git-status` (#1596) is **not** affected in the same way: `status_records/1` drops a record the 1 MiB cap cut in half rather than returning it, so an entry path cannot leak a prefix.

## The ordering inside Elixir is already correct

Redaction must run **before** `to_text/1`, and does. Recoding first would rewrite a non-ASCII secret's own bytes (`é` = `C3 A9` → `C3 83 C2 A9`) out of its own match and stop redacting it — measured: the wrong order yields a body with no `[REDACTED]` at all. The defect is that truncation happens one layer earlier still, in the shell, where Elixir gets no chance to redact first.

## Suggested fix

Ask the shell for `max_bytes + <longest secret length - 1>` bytes of overlap, redact the result, then `binary_part/3` down to the cap. `secret_values/1` already sorts by size, so the overlap is available. That keeps the cap's memory guarantee while ensuring any secret crossing the caller's boundary is present whole at the moment redaction runs.

The same treatment applies to the diff and diagnostic caps, or they move into Elixir entirely — worth deciding together, since letting a script emit unbounded output before trimming has its own cost.

jhgaylor:
Fixes #1907

## The defect

`Fountain.SandboxFiles` applied its output caps in the **shell**, with
`head -c`, and redacted in Elixir over whatever survived the cut.
`redact_with/2` is `:binary.replace/4` against the sandbox's known secret
values, so it only matches a value that is present whole. A cut that lands
inside a value leaves a prefix that matches nothing, and that prefix goes
into the response.

On `GET /api/sandboxes/:id/file` the cut position is the caller's: `read/3`
takes `max_bytes` from `params["max_bytes"]`, clamps it to 1…4 MiB, and
hands it to `head -c "$n"`. So this is not an occasional boundary artifact
that depends on where a value happens to sit — the caller decides where the
boundary falls relative to a value, and can put it anywhere. A read of the
sandbox's `.env` is exactly what ADR 0039 decision 5 and the module's own
moduledoc say is prevented, and it matters who holds the key: a full-scope
token includes one issued to a third-party OAuth app the user authorized,
a different principal from the tenant whose vault values redaction protects.

This is on `main` and published in SDK 1.15.0. `/file`, `max_bytes` and this
ordering all shipped in the original ADR 0039 PR.

## The fix

Each script is asked for `overlap/1` bytes past the cap, redaction runs over
that, and only then is the result cut down to what the caller asked for.

`overlap/1` is one byte less than the longest known value, which
`secret_values/1` already sorts to the front. That width is exact rather
than generous: a value of length `L` with at least one byte inside the cap
starts at `max_bytes - 1` at the latest, so it ends at `max_bytes + L - 1`
at the latest. Anything wider buys nothing; anything narrower leaves the
worst case short.

The cut back down is taken in **the bytes the script produced**, not in the
redacted text, and that distinction is load-bearing. Redacting the whole
overlap window and then trimming the result to `max_bytes` looks equivalent
and is not: a value replaced by a shorter placeholder moves every byte
behind it closer to the front, which can carry a fragment out of the overlap
region and back inside the cap, unmatched. So `cut_at/3` finds the value
lying across the cap (at most one can, since matches do not overlap), keeps
up to its end so it is present whole, redacts, and caps the result. There is
a test for this variant specifically, and it is the only one of the three
that catches it.

Cutting a `[REDACTED]` marker in half is possible and is harmless — a
marker fragment carries nothing.

`truncated` now also reports the cut back down, which can happen when
`[REDACTED]` is longer than the value it replaces.

**Both caps that exist on `main` are fixed** — `read/3` and `diff/3` — since
one helper serves both. `diff/3` had the same shape twice over: the shell
cut at `max_bytes + 1`, and the Elixir `binary_part/3` before it also ran
ahead of redaction.

`/git-status` is not affected: `status_records/1` drops a record the cap cut
in half rather than returning it.

The ordering of redaction against `to_text/1` is untouched. Redaction still
runs first, because recoding first would rewrite a non-ASCII value's bytes
out of its own match (`é` = `C3 A9` → `C3 83 C2 A9`) and stop redacting it
entirely.

## Verified by reverting

Three reverts, each run against the new tests.

**A. The whole fix reverted** (`git checkout origin/main -- sandbox_files.ex`)
— 4 failures. Each fires on the overlap the script is asked for, before the
leak assertions are reached:

```
code:  assert n == max_bytes + byte_size(@straddled) - 1
left:  1
right: 20
```

**B. The overlap kept, the cut restored to `max_bytes` before redaction**
(`cut_at/3` returns `min(byte_size(bytes), max_bytes)`) — the vulnerable
ordering with the new shape. 3 failures, now on the leak itself:

```
1) diff/3: the same walk, one byte further out for the exact-fit probe
   max_bytes=3 leaked "Q" in "p=Q"

2) a value replaced earlier does not pull a fragment back inside the cap
   max_bytes=32 leaked "Q" in "[REDACTED]|Q"

3) read/3: no max_bytes anywhere across the value yields a fragment of it
   max_bytes=3 leaked "Q" in "p=Q"
```

**C. The overlap kept, the whole window redacted and the *result* trimmed**
(`cut_at/3` returns `byte_size(bytes)`) — the plausible-looking wrong fix.
1 failure, and only the shift test sees it, which is why that test is there:

```
1) a value replaced earlier does not pull a fragment back inside the cap
   max_bytes=12 leaked "Q" in "[REDACTED]|Q"
```

The fixed tree: `52 tests, 0 failures` over
`sandbox_files_test.exs`, `sandbox_files_controller_test.exs` and
`redaction_test.exs`.

## The tests

The straddle is constructed, not hoped for. The value shares not one
character with the text around it or with `[REDACTED]`, so each walk asserts
that **no prefix of it, down to a single byte, appears anywhere in the
response** — at every `max_bytes` from 1 to the length of the file, for both
`read/3` and `diff/3`. Each step also pins `byte_size(content) <= max_bytes`,
so the overlap cannot quietly return more than was asked for, and a separate
test pins the cap on a file that fits it exactly and is grown past it by the
placeholder.

## Overlap with #1596

#1596 is in the merge queue and edits the same module. It rewrites `diff/3`'s
`run` call (adding `roots/1` to the argv), the `redact/2` clause, and the
`# ── redaction ──` section, and it adds a `head -c 4096` cap of its own on
git's stderr. **These lines conflict**, and whichever of the two lands second
needs a rebase. Two notes for whoever does it:

- I used `redact_with(values, bytes)` with #1596's argument order on purpose,
  so the two definitions are the same function.
- The diagnostic cap `head -c 4096` does not exist on `main` — it arrives
  with #1596. It is fixed-width and not caller-steerable, so it is a
  different severity, but it has the same shape and after #1596 merges it is
  the one remaining uncovered cap in the module. Worth its own issue rather
  than a rebase note.

`max_max_bytes/0` and `default_max_bytes/0` are untouched, so the wire shape
does not move and `sdk/contract/contract.json` does not diff. The `truncated`
**descriptions** did change (see the rebase comment below), so
`sdk/typescript/src/generated/openapi.ts` is regenerated; the PR carries
`sdk-no-release` rather than a version bump.

## Checks

From `apps/fountain` on the mise-pinned toolchain: `mix format --check-formatted`,
`mix compile --warnings-as-errors`, `mix credo --strict` (798 files, no
issues), and the three test files above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,19 @@ upgrade, is in
   compatible endpoint is unchanged: it still synthesizes a caption for an
   image-only message, because clients of that dialect cannot always send one.
 
+- **`truncated` on a sandbox file or diff now means "this is not the whole
+  thing", not "the cap was reached"** (#1907). On
+  `GET /api/sandboxes/:id/file` and `GET /api/sandboxes/:id/diff` it is still
+  true when the file or diff is longer than `max_bytes`, and it is now also
+  true when redaction grew what was read past that cap — which happens when
+  `[REDACTED]` is longer than the value it stands in for. So a file that fits
+  `max_bytes` can come back `truncated: true` with its content cut. The
+  widening errs safe: `truncated: false` still means the bytes are complete,
+  which is the direction a caller depends on, and the new true case tells a
+  caller to ask for more. The field has been published since SDK 1.15.0; its
+  description changed with it, in the OpenAPI document and in the generated
+  TypeScript types.
+
 ### Added
 
 - **An `acp` runtime launches a named command, so a deterministic program can
@@ -550,6 +563,29 @@ upgrade, is in
   outbound WebSocket client to one known host, and Fountain serves HTTP with
   **Bandit, not Cowboy**, so cowlib is not on the inbound request path.
   Tracking upstream; `mix hex.audit` reports them and does not fail the build.
+- **A sandbox file read could return a fragment of a secret, at a byte offset
+  the caller chose** (#1907). `GET /api/sandboxes/:id/file` and
+  `GET /api/sandboxes/:id/diff` capped their output in the shell with
+  `head -c` and redacted in Elixir over whatever survived the cut. Redaction
+  matches a value's own bytes, so a cut through one left a prefix that matched
+  nothing and travelled on in the clear. On `/file` the cut lands at
+  `max_bytes`, which the caller sends, so this was not an occasional boundary
+  artifact that depended on where a value happened to sit: the caller decided
+  where the boundary fell relative to a value, and could walk it. Reading the
+  sandbox's `.env` that way is exactly what ADR 0039 decision 5 says is
+  prevented, and the reader need not be the tenant whose vault values
+  redaction protects — a `full`-scope key includes one issued to a
+  third-party OAuth app the user authorized. Both endpoints now ask their
+  script for enough bytes past the cap that a value lying across it arrives
+  whole (one less than the longest known value, which is the widest one can
+  straddle it), redact that, and only then cut the result down to what was
+  asked for. That last cut is taken in the bytes the script produced rather
+  than in the redacted text, because a value replaced by a shorter
+  `[REDACTED]` moves every byte behind it forward and could otherwise carry
+  an unmatched fragment back inside the cap. `/git-status` was never affected:
+  it drops a record its cap bisected rather than returning it. The change to
+  what `truncated` reports is under **Changed** above. Present since `/file`
+  and `max_bytes` shipped with ADR 0039, and published in SDK 1.15.0.
 
 ## [0.16.0] - 2026-09-03
 
--- a/apps/fountain/lib/fountain/sandbox_files.ex
+++ b/apps/fountain/lib/fountain/sandbox_files.ex
@@ -223,6 +223,10 @@ defmodule Fountain.SandboxFiles do
   itself when it is valid UTF-8 (`encoding: "utf-8"`) and base64 otherwise
   (`encoding: "base64"`); `size` is the whole file and `truncated` says
   whether `content` stopped short of it.
+
+  Redaction runs before the cap, never after: `max_bytes` is the caller's to
+  choose, so a cut taken first would let them place it inside a value and
+  read the piece in front of it (#1907).
   """
   @spec read(Sandbox.t(), String.t(), keyword()) ::
           {:ok,
@@ -239,16 +243,23 @@ defmodule Fountain.SandboxFiles do
 
     with :ok <- ready?(sandbox),
          {:ok, absolute} <- resolve_path(sandbox, path),
-         {:ok, output} <- run(sandbox, read_script(), [Integer.to_string(max_bytes), absolute]),
+         values = secret_values(sandbox),
+         # `overlap/1` bytes past the cap, so a secret lying across it is
+         # whole when redaction runs; `redact_to_cap/3` cuts back down.
+         {:ok, output} <-
+           run(sandbox, read_script(), [
+             Integer.to_string(max_bytes + overlap(values)),
+             absolute
+           ]),
          {:ok, size, bytes} <- parse_read(output) do
-      bytes = redact(sandbox, bytes)
+      {bytes, capped?} = redact_to_cap(values, bytes, max_bytes)
       {encoding, content} = encode(bytes)
 
       {:ok,
        %{
          path: absolute,
          size: size,
-         truncated: size > max_bytes,
+         truncated: size > max_bytes or capped?,
          encoding: encoding,
          content: content
        }}
@@ -280,27 +291,23 @@ defmodule Fountain.SandboxFiles do
     with :ok <- ready?(sandbox),
          {:ok, ref} <- validate_ref(ref),
          {:ok, absolute} <- resolve_path(sandbox, path),
-         # One byte past the cap tells truncation from an exact fit.
+         values = secret_values(sandbox),
+         # One byte past the cap tells truncation from an exact fit, and
+         # `overlap/1` past that is what lets redaction see a secret lying
+         # across the cap whole.
          {:ok, output} <-
            run(
              sandbox,
              diff_script(),
              [
                absolute,
-               Integer.to_string(max_bytes + 1),
+               Integer.to_string(max_bytes + 1 + overlap(values)),
                ref || "",
                if(staged, do: "1", else: "0")
              ] ++ roots(sandbox)
            ),
          {:ok, root, bytes} <- parse_diff(output) do
-      truncated = byte_size(bytes) > max_bytes
-
-      values = secret_values(sandbox)
-
-      text =
-        bytes
-        |> binary_part(0, min(byte_size(bytes), max_bytes))
-        |> then(&redact_with(values, &1))
+      {text, capped?} = redact_to_cap(values, bytes, max_bytes)
 
       {:ok,
        %{
@@ -312,7 +319,7 @@ defmodule Fountain.SandboxFiles do
          staged: staged,
          ref: ref,
          diff: to_text(text),
-         truncated: truncated
+         truncated: byte_size(bytes) > max_bytes or capped?
        }}
     end
   end
@@ -754,8 +761,8 @@ defmodule Fountain.SandboxFiles do
   defp redact(%Sandbox{} = sandbox, bytes) when is_binary(bytes),
     do: redact_with(secret_values(sandbox), bytes)
 
-  # `secret_values/1` reads the vault and the environment, so a caller with
-  # many strings to redact looks them up once and replaces many times.
+  # `secret_values/1` reads the vault and the environment, so a caller that
+  # needs the values for something else too looks them up once.
   defp redact_with([], bytes), do: bytes
 
   defp redact_with(values, bytes),
@@ -771,6 +778,53 @@ defmodule Fountain.SandboxFiles do
         renamed_from: change.renamed_from && to_text(redact_with(values, change.renamed_from))
     }
   end
+
+  # Redaction has to see a secret whole: `:binary.replace/4` matches the
+  # value's own bytes, so a cut through one leaves a prefix that matches
+  # nothing and travels on in the clear. The scripts cut first, with
+  # `head -c`, and on `read/3` the cut lands at `max_bytes` — which the
+  # caller chooses, making an incidental boundary case a repeatable one
+  # (#1907). So the scripts are asked for `overlap/1` bytes past the cap,
+  # redaction runs over that, and only then is the result cut to the cap.
+  #
+  # Returns the bytes and whether that last cut dropped anything, which it
+  # can when a value is shorter than the placeholder standing in for it.
+  defp redact_to_cap(values, bytes, max_bytes) do
+    kept = binary_part(bytes, 0, cut_at(values, bytes, max_bytes))
+    redacted = redact_with(values, kept)
+
+    if byte_size(redacted) > max_bytes,
+      do: {binary_part(redacted, 0, max_bytes), true},
+      else: {redacted, false}
+  end
+
+  # How much of `bytes` survives into redaction: the cap, carried forward to
+  # the end of a value lying across it. At most one value can, because
+  # matches do not overlap, and `:binary.matches/2` picks the same ones
+  # `:binary.replace/4` will.
+  #
+  # Cutting at the cap instead is the defect. Cutting *after* redacting is
+  # not the fix either: an earlier value replaced by a shorter placeholder
+  # shifts the bytes behind it, which can pull an unmatched fragment out of
+  # the overlap and back inside the cap.
+  defp cut_at([], bytes, max_bytes), do: min(byte_size(bytes), max_bytes)
+
+  defp cut_at(values, bytes, max_bytes) do
+    limit = min(byte_size(bytes), max_bytes)
+
+    bytes
+    |> :binary.matches(values)
+    |> Enum.find_value(limit, fn {start, length} ->
+      if start < limit and start + length > limit, do: start + length
+    end)
+  end
+
+  # How far past the cap a value can lie across it: one byte less than the
+  # longest, which `secret_values/1` sorts to the front. Ask a script for
+  # that many bytes more and any value with a byte inside the cap arrives
+  # whole.
+  defp overlap([]), do: 0
+  defp overlap([longest | _]), do: byte_size(longest) - 1
 
   # The identity's own values (what `.env` on that disk holds) plus what any
   # live server registered — the latter covers the inference credential and
--- a/apps/fountain/lib/fountain/sandbox_files.ex
+++ b/apps/fountain/lib/fountain/sandbox_files.ex
@@ -222,7 +222,8 @@ defmodule Fountain.SandboxFiles do
   #{@default_max_bytes}, at most #{@max_max_bytes}). `content` is the text
   itself when it is valid UTF-8 (`encoding: "utf-8"`) and base64 otherwise
   (`encoding: "base64"`); `size` is the whole file and `truncated` says
-  whether `content` stopped short of it.
+  whether `content` is short of it — because the file is longer than
+  `max_bytes`, or because redaction grew what was read past the cap.
 
   Redaction runs before the cap, never after: `max_bytes` is the caller's to
   choose, so a cut taken first would let them place it inside a value and
--- a/apps/fountain/lib/fountain_web/controllers/sandbox_files_controller.ex
+++ b/apps/fountain/lib/fountain_web/controllers/sandbox_files_controller.ex
@@ -79,7 +79,8 @@ defmodule FountainWeb.SandboxFilesController do
         "vault is replaced with `[REDACTED]`, as in the transcript. `content` is the text " <>
         "when it is valid UTF-8 (`encoding: utf-8`) and base64 otherwise " <>
         "(`encoding: base64`). `size` is the whole file; `truncated` says whether `content` " <>
-        "stopped at `max_bytes`. Full scope.",
+        "is short of it, which happens when the file is longer than `max_bytes` and also " <>
+        "when redaction grows what was read past that cap. Full scope.",
     parameters: [
       sandbox_id: [in: :path, type: :string, required: true],
       path: Keyword.put(@path_param, :required, true),
--- a/apps/fountain/lib/fountain_web/schemas.ex
+++ b/apps/fountain/lib/fountain_web/schemas.ex
@@ -200,7 +200,10 @@ defmodule FountainWeb.Schemas do
         size: %Schema{type: :integer, description: "The whole file, in bytes."},
         truncated: %Schema{
           type: :boolean,
-          description: "True when `content` stopped at `max_bytes` before the end of the file."
+          description:
+            "True when `content` is not the whole file: either the file is longer than " <>
+              "`max_bytes`, or redaction grew what was read past it. False means `content` " <>
+              "is everything."
         },
         encoding: %Schema{
           type: :string,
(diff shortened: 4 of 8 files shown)


BinaryBourbon (approved):
Reviewed `00c896e5` against `origin/main`. I did not take the reasoning on trust: I reimplemented `overlap/1`, `cut_at/3` and `redact_to_cap/3` alongside the sketch I originally suggested and ran both over 200,000 generated `.env`-shaped files, sweeping **every** `max_bytes` from 1 to the file's length, with the secrets actually planted so matches occur. 19 checks green, 4 skipped.

**Approving.** And the correction lands: my sketch was wrong, by a wide margin, and this PR is right.

## The shift claim holds, and it is not a corner

The mechanism as I was given it was inverted, and the direction is the whole thing. It is not that a *longer* placeholder shifts bytes toward the front — a longer placeholder pushes content out past the cap, which is harmless. It is a value **longer than the 10-byte placeholder** that shifts everything behind it *toward* the front, and can carry an unmatched fragment from beyond the overlap back inside the cap. `cut_at/3`'s comment states it correctly. Hand-built, two secrets of 30 and 20 bytes, `max_bytes = 12`:

```
sketch (redact the window, then trim) = "[REDACTED]QQ"   ← two raw bytes of the second value
implemented (cut in the raw bytes)    = "[REDACTED]"
```

Since every real API key is longer than ten bytes, that is the ordinary regime, not an exotic one. Over the 200,000 planted cases:

| | violations |
|---|---|
| implemented: output not a prefix of `redact(whole file)` | **0** |
| implemented: output longer than `max_bytes` | **0** |
| my sketch: output not a prefix of `redact(whole file)` | 46,382 |
| my sketch: output carries a ≥3-byte raw prefix of a secret | 40,570 |
| implemented with `overlap − 1` | 3,709 |

The last row is the answer to the minimality question from the other side: `L − 1` is not merely sufficient, it is exact — one byte narrower and the boundary case reappears at measurable frequency. Sufficiency I also checked by construction: a value with any byte inside the cap starts at `max_bytes − 1` at the latest and so ends at `max_bytes + L − 1` at the latest, which is inside the window the scripts are now asked for. And because `:binary.matches/2` returns non-overlapping matches, at most one can span the cap, so `Enum.find_value/3` returning the first is not a case of taking the earliest of several.

**The invariant I checked is the right one**: the output must be a prefix of `redact_with(values, whole_file)` — what an unbounded redaction followed by a cut would give. That is stronger than "no fragment leaked" and it is what `cut_at/3` buys, because cutting at a point no match straddles makes the redaction of the prefix a prefix of the redaction.

**The walk tests are genuinely exhaustive**, not sampled: `for max_bytes <- 1..byte_size(body)` in all three, each iteration asserting `byte_size(content) <= max_bytes` *and* that no prefix of the secret of any length 1..20 appears. I confirmed the cap independently — 0 over-cap answers in 200,000 cases — rather than relying on that assertion.

## `truncated`: the behaviour is right, three published sentences are now wrong

Reporting the cut back down is correct and errs in the safe direction: a caller that treats `truncated: false` as "this is the whole thing" is never misled, and the new true case tells them to ask for more. The old formula was also the one thing a client could not compute itself, so widening it loses nothing.

What the PR does not do is update what the field *says*, and it is now false for a case the PR's own test asserts (`body = "ab12345678cd"`, `max_bytes: 12`, file fits, `truncated` true):

- `apps/fountain/lib/fountain_web/schemas.ex:203` — "True when `content` stopped at `max_bytes` before the end of the file." It can now be true when the content did not stop before the end of the file.
- `schemas.ex:243` — same for `diff`.
- `sandbox_files.ex:166` — "`size` is the whole file and `truncated` says whether `content` stopped short of it", left intact below the new paragraph.

A developer reading the generated reference will request a larger `max_bytes`, get the same file back with `truncated: false`, and have learned it the hard way. Non-blocking — nothing is harmed and the semantics are the safer ones — but I think two string edits and a TypeScript regenerate belong in this PR rather than a follow-up, since this is the PR that changes the meaning, and a CHANGELOG line is warranted for a field published in 1.15.0.

## The rest of what was asked

- **The non-UTF-8 ordering is not defeated.** Walked every `max_bytes` over a body holding a 33-byte secret with multibyte characters throughout (`café`, `señor`, `™`, Cyrillic): zero leaks and zero over-cap answers, checked both on the raw output and after `to_text/1`. The reason it holds is structural — `cut_at/3` deliberately steps past a straddling match, so no cut in this path ever falls inside a secret, and the only cut that can land mid-character is the final one on already-redacted text, which contains no secret to split.
- **Nothing else on `main` caps ahead of redaction.** `main` has exactly two `head -c` caps, `read_script/0:347` and `diff_script/0:371`, and both are fixed here. `diff/3` had the shape twice and both are gone.
- **The diagnostic cap is out of scope, confirmed.** `head -c 4096` on git's stderr does not exist on `main`; it arrives with #1596. It is the same class but not the same severity — that cap is fixed at 4,096 and not caller-steerable, so it needs a secret to land across that exact byte rather than being placed there. Its own issue after #1596 lands is the right call. `/git-status` itself needs no overlap: `status_records/1` drops a record the cap bisected rather than returning it.

## Reconciling with #1596

The two `redact_with/2` definitions are byte-identical apart from two comment lines, so that half is one function under one name, not two ideas. The collision to watch is `diff/3`'s `run` call, where the two PRs edit the same expression from opposite ends. Four things a careless rebase drops, each silently:

1. `++ roots(sandbox)` — reopens #1596's confinement escape.
2. `+ overlap(values)` inside the byte argument — reopens #1907 for `/diff`.
3. `values = secret_values(sandbox)` bound twice, once in the `with` (this PR, needed before the `run`) and once in the body (#1596, for `repo_root`). Not a compile error, just `Crypto.load_tenant_key/1` plus three queries run twice per read.
4. #1596's exact-argument assertions — `assert args == [@home, "262145", "main", "1", @home]` and the controller's `{:exec_args, …}` — must move to `262145 + overlap`, and the honest fix is to update the number, not to loosen the match to `[_, _, _, _ | _]`.

Worth re-running the #1907 walk tests after the rebase rather than trusting green, since three of the four above leave the suite green on the half of the tree they do not touch.

## Also noticed, separate from this PR

`list/2` does not redact at all. Entry names from `GET /api/sandboxes/:id/files` go out raw, while `/git-status` redacts its paths in #1596 and `read/3` and `diff/3` redact their bytes here. ADR 0039 decision 5 says every byte that leaves passes the replacement, and a filename is a byte the agent chose. Not this PR's business and not a cap-ordering problem, but it is the last unredacted exit in the module and deserves its own issue.

jhgaylor:
Rebased onto `main` and closed the `truncated` documentation gap. Force-pushed, so **this supersedes the approval** — `57e9155`… (`57e59155`) is the new head.

## The rebase

The branch already sat on a base containing #1596 (`3f731281`), so the conflicting rebase was behind it; what remained was one commit, `96277135` (OAuth CORS), which touches no file here. Clean, no conflicts.

That is not evidence, though, and the four silent-drop items are exactly the ones a clean rebase does not report. Each checked by hand in the rebased tree:

1. **`++ roots(sandbox)`** — present, `sandbox_files.ex:307`. Not taken on trust: removed it and re-ran the three files. **3 failures**, on the argv assertions that exist to catch it:
   ```
   left:  ["/home/sprite/repo", "262145", "main", "1"]
   right: ["/home/sprite/repo", "262145", "main", "1", "/home/sprite"]
   ```
   Restored.

2. **`+ overlap(values)`** — present in both, `:251` (`max_bytes + overlap(values)`) and `:304` (`max_bytes + 1 + overlap(values)`). Same treatment: reverted both to the bare cap and re-ran. **6 failures** — all three #1907 walk tests, the placeholder-longer-than-value cap test, and the read argv assertion:
   ```
   left:  ["262144", "/home/sprite/.env"]
   right: ["262157", "/home/sprite/.env"]

   max_bytes=30 ... left: 1  right: 30
   max_bytes=20 ... left: 1  right: 20
   ```
   So the script is demonstrably asked for `cap + overlap`, not `cap`. Restored.

3. **No doubled `values = secret_values(sandbox)`** — one binding in `read/3`'s `with` (`:246`), one in `diff/3`'s (`:294`), and #1596's own one in `status/2` (`:368`), which is a different function. `grep` over the module returns exactly those three plus the `redact/2` clause. #1596's body binding is gone, not shadowed.

4. **The exact-argument assertions were updated, not loosened.** Every one is still a literal list — `assert args == [...]` / `assert_received {:exec_args, [...]}`, no wildcards anywhere. The numbers moved only where the overlap is nonzero: the controller's `/file` test carries a 14-byte vault value, so `1000` → `1013`; `sandbox_files_test.exs:189` reads `"#{262_144 + 13}"`. The diff assertions stay at `262145` because those sandboxes hold no secrets, so `overlap/1` is 0 there — and the revert above confirms they still fail when the argv changes, so they are still guarding.

**#1907 walk tests re-run on the rebased tree and passing**, along with everything else in the three files: **61 tests, 0 failures** (52 before the rebase; #1596 brought the rest).

## `truncated`

Fixed, plus a fourth instance the review did not name:

- `schemas.ex:203` (`SandboxFile`) and `:243` (`SandboxDiff`) — now "True when `content`/`diff` is not the whole file/diff: either it is longer than `max_bytes`, or redaction grew what was read past it. False means it is everything."
- `sandbox_files.ex` `read/3` docstring — `truncated` "says whether `content` is short of it — because the file is longer than `max_bytes`, or because redaction grew what was read past the cap."
- **`sandbox_files_controller.ex:82`**, the `show` operation description, which carried the same claim into the generated reference at operation level.
- **`sdk/typescript/src/client.ts:264`**, the hand-written `sandboxFile` JSDoc, which said it too.

No docs page and no other SDK's docstring describes the field (`grep` over `docs/`, `sdk/python`, `sdk/swift`, `sdk/elixir`, `cli/` turns up nothing), so nothing else was stale.

## Spec, types, contract, release

- `mix openapi.spec.json` + `jq empty`: ok.
- `scripts/sdk-contract/build.sh --check`: **ok, 165 operations, 193 schemas, no diff**. `sdk/contract/contract.json` is unchanged by design — the projection drops descriptions so a docs pass cannot diff the wire contract. `dist/openapi.json` is rebuilt and carries the new strings.
- `npm run generate`: three lines in `sdk/typescript/src/generated/openapi.ts` (the two schema descriptions and the operation description). `operations`/`schemas` stay key-sorted; the generator writes them.
- **`sdk-no-release`, deliberately.** The SDK change is descriptive text only — no type, no field, no runtime behaviour. Four open PRs (#1860, #1876, #1885, #1893) already claim `1.28.0`, and since the publish workflow ships whatever version lands on `main`, a fifth claim on the same number buys nothing but a merge-order hazard on a version npm can never republish. The corrected JSDoc rides out with whichever of those publishes next.
- `CHANGELOG.md`: a **Security** entry for the leak and a **Changed** entry for `truncated`, saying plainly that the field has been published since SDK 1.15.0 and that a file fitting `max_bytes` can now come back `truncated: true`.

## Checks

From `apps/fountain` on the mise-pinned toolchain (1.19.2/OTP 28 — Homebrew's `mix` here is 1.19.5 and formats differently):

- `mix format --check-formatted`, `mix compile --warnings-as-errors`, `mix credo --strict` (808 files, no issues).
- `sandbox_files_test.exs` + `sandbox_files_controller_test.exs` + `redaction_test.exs`: **61 tests, 0 failures**.
- `docs_test.exs`, `api_spec_test.exs`, `schema_guardrail_test.exs`, `schema_enum_guardrail_test.exs`, `schema_wrappers_test.exs`, `schemas_test.exs`: **65 tests, 0 failures**.
- TypeScript SDK: `typecheck` clean, `verify-contract` ok (73 operations, 49 schemas, 5 enums), `npm test` **109 pass, 0 fail**.

Prose gates were not run against a changed page: `CHANGELOG.md` is excluded from vale by name, and `docs-style.py` and `destink` walk only `docs/` and `apps/*/docs/`. No file under either changed.

BinaryBourbon (approved):
Re-reviewed `57e59155` after the force push. I did not take "the rebase was clean" as evidence: I extracted `redact_with/2`, `redact_to_cap/3`, `cut_at/3` and `overlap/1` **verbatim from this head** and re-ran the 200,000-case harness against the real functions rather than my reconstruction of them.

**Approving the substance.** One thing has to be fixed before this can queue, and it is not the code — see the last section.

## The overlap survived in substance, not just in argv

Against the helpers as they exist at `57e59155`, 200,000 generated `.env`-shaped files, every `max_bytes` from 1 to the file length:

| | |
|---|---|
| output not a prefix of `redact(whole file)` | **0** |
| output longer than `max_bytes` | **0** |
| output carrying a ≥3-byte raw prefix of a secret | **0** |
| same, with `overlap/1 − 1` | 3,627 |

The last row is the one that matters for a rebase review: it proves the width is still doing work at this head, not that an assertion still reads `262157`. The three helpers diff byte-identically against the revision I verified before, so the property carries over intact.

Your four hazards, checked against the file rather than the report:

1. `++ roots(sandbox)` — `sandbox_files.ex:308` (`diff/3`) and `:364` (`status/3`). Present.
2. `+ overlap(values)` — `:252` and `:305`. Present, and in `diff/3` it sits in the same expression as the `roots` append, which was the collision point: `Integer.to_string(max_bytes + 1 + overlap(values))` … `] ++ roots(sandbox)`. Both halves survived the merge of the two edits.
3. One `values` binding per function — `:247`, `:295`, `:369` — and `:369` is `status/3`, a different function. #1596's body binding in `diff/3` is gone rather than shadowed, and `repo_root: to_text(redact_with(values, root))` at `:319` still reads the surviving one.
4. Assertions updated, not loosened. Every one is still a literal list; `1000 → 1013` in the controller `/file` test and `262_144 + 13` in the unit test both carry the overlap explicitly, and the two `/diff` assertions correctly stay at `262145` because those sandboxes hold no values, so `overlap([])` is 0. That last detail is the one a loosened matcher would have hidden.

`status/3` deliberately has no overlap, which remains correct: its cap is not the caller's and `status_records/1` drops a bisected record rather than returning it.

Branch scope is clean — eight files, all of them this change plus the CHANGELOG and the two TypeScript files. The OAuth CORS work that appears in a head-to-head diff is the base moving, not something that rode in.

## The `truncated` wording

All five places now say the same true thing, and the sentence I would have written: not "the cap was reached" but "this is not the whole thing", with both causes named, and `truncated: false` explicitly promising completeness — which is the direction a caller actually depends on. `schemas.ex:203` and `:246`, the `read/3` docstring at `sandbox_files.ex:225`, the `show` operation description at `sandbox_files_controller.ex:82`, and the hand-written JSDoc at `client.ts:263`, with the three regenerated lines in `generated/openapi.ts` matching.

No sixth place says the old thing. The nearest is `@max_bytes_param` at `sandbox_files_controller.ex:43` — "`truncated` says whether the content stopped short" — which is shared by `/file` and `/diff` and is still accurate, if vaguer than its neighbours now. `docs/api.md`'s "Check truncation before you show a response as a complete file" is, if anything, more correct than it was.

**The contract-projection explanation checks out.** `sdk/contract/contract.json` carries only `type`, `required` and `enum` per property — `SandboxFile.truncated` is `{"required": true, "type": "boolean"}` and nothing else. The ten `"description"` strings in that file are all *property names* (`Agent.description`, `Vault.description`, and so on), not documentation. So `build.sh --check` reporting no diff is the projection behaving as designed, not a stale artifact.

The CHANGELOG is accurate on both entries, including the shift subtlety and the note that `/git-status` was never affected.

## The version-bump call: I agree, for a stronger reason than merge order

`client.ts` changed a JSDoc block and `generated/openapi.ts` changed three `@description` comments. No type, no method, no signature, no runtime behaviour. So the SDK's **published surface does not change** — `sdk-no-release` here is not a waiver of the rule, it is the accurate classification of the change, and the rule ("a version moves when the SDK's own public surface changes") is satisfied by not moving it. The merge-order argument is a second, independent reason: the gate reports base `1.27.0`, and four open PRs already claim `1.28.0`, so a fifth claim on a number npm cannot republish buys a collision and nothing else.

What it costs is small and worth stating: the hover text in an editor stays stale until whichever of those four ships. The server's own generated reference is correct the moment this merges, and that is the authority a caller checks; the stale sentence also fails safe, since anyone acting on "stopped at maxBytes" does the same thing they would do on the new text. If a later revision of this PR touches a type or a method, the label needs revisiting — that is the line, not the file list.

## What blocks queueing: the SDK release gate is red, and a re-run will not fix it

`A change to the SDK says whether it releases` **failed** on this head, 14s:

```
base: 1.27.0   head: 1.27.0 (unchanged)
shipped: sdk/typescript/src/client.ts, sdk/typescript/src/generated/openapi.ts
error: ... To change the published surface without releasing, add the "sdk-no-release" label.
```

The label *is* on the PR, so the job's `if:` did not see it: that condition reads `github.event.pull_request.labels`, which is frozen in the event payload, and the payload for this force-push's `synchronize` predates the label. The workflow's own header documents this exact trap and the remedy — `labeled`/`unlabeled` are in its `types:` list precisely so that **toggling the label fires a fresh event with a fresh payload**. Re-running the failed job replays the stale one and fails again.

So: remove and re-add `sdk-no-release`, confirm the job flips to *skipping*, and then queue. Nothing in the tree needs to change, which is why this is an approval and not a request for changes — but it is not mergeable as it stands, and a re-run is the one thing that will not clear it.

I did not run the suite: the shared test database is a migration behind and another session holds it, and for these particular claims the algorithm-level check against the extracted helpers is the stronger instrument anyway.

Issue #1916, opened 2026-09-11
Decide what a platform-inference turn with unknown tokens costs


jhgaylor:
Follow-up to #1685, which is the *visibility* half.

## Where we are

Since #1685 every ACP turn that reaches `session/prompt` carries its
inference source on the turn row from the moment the prompt is written:
`usage ->> 'inference'` is `platform` or `own`, with `model` beside it on a
platform turn. A turn that then ends without a `session/prompt` response —
an adapter exit, a sandbox deadline, a ConversationServer restart, an
interrupt, or a runtime that reports no usage — keeps that stamp and carries
no token figure.

Production over the seven days before #1685 was filed: 26 turns reported a
token figure, 319 ended with no usage map at all (204 completed, 65
interrupted, 50 failed). Whichever of those ran on the platform key spent
Fountain's tokens.

## The open decision

What does such a turn cost?

`Credits.InferenceRates.cost_cents/1` prices it at zero, so it writes no
`burn_inference` row and contributes nothing to
`Billing.platform_inference_spend_today/1` — which is what bounds the
deployment against `PLATFORM_INFERENCE_DAILY_CENTS`. #1685 deliberately left
that alone rather than invent a number: "priced at a floor" and "visible as a
zero-token turn" are different decisions with different money attached, and
picking a floor is a product call.

Things the decision has to answer:

- Is a floor charged per turn, or is the spend only *estimated* for the
  ceiling without a ledger row? The ceiling is a circuit breaker on
  Fountain's own bill; the ledger is what a tenant is charged. They do not
  have to agree.
- What is the figure derived from — the model's rate card against a median
  turn, the turn's elapsed seconds, the conversation's other turns?
- An interrupted turn and a turn that failed before the first token are not
  the same spend. Does status enter the estimate?
- A floor charged for a turn the runtime never billed anyone for is a charge
  for nothing; a zero for a turn that burned a full context window is
  Fountain paying for it. Both are wrong in one direction.

## What lands with it

`CreditPricer.unpriced_inference_turns/1` currently excludes rows with no
numeric token key, because the inference pass stops on a page that wrote
nothing and these rows outnumber the priceable ones ten to one — a full page
of them would starve the turns behind them. Whatever prices them has to
either re-admit them *and* give the pass a cursor that advances past rows it
cannot price, or run as a pass of its own.

The rows are on disk and selectable by the same
`usage ->> 'inference' = 'platform'` selector, so the decision can be taken
against real data whenever someone takes it.

[...]

Issue #1685, opened 2026-09-07
Platform inference is unbilled when a turn ends without a session/prompt response


jhgaylor:
## Symptom

A turn on the platform inference key that ends any way other than a `session/prompt` response spends tokens on Fountain's key with no ledger row, and does not count toward `PLATFORM_INFERENCE_DAILY_CENTS`.

Production, trailing 7 days: 26 completed turns carry `usage->>'inference' = 'platform'`; 319 turns ended with no usage map at all (204 completed, 65 interrupted, 50 failed). Whichever of those ran on the platform key are invisible to the pricer.

## Mechanism

- The `inference: platform` stamp is put on the usage map only inside `TurnMachine.handle({:done, ...})`, `apps/fountain/lib/fountain/conversations/turn_machine.ex:334-336` (`record_usage(turn, with_inference(usage, ctx))`).
- `Conversations._unsafe_record_turn_usage(turn, nil)` is a no-op (`conversations.ex:1354`).
- `CreditPricer.unpriced_inference_turns/1` selects on `usage ->> 'inference' = 'platform'` (`ee/lib/fountain/workers/credit_pricer.ex:214-233`), and `Billing.platform_inference_spend_today/1` (`ee/lib/fountain/billing.ex:52-63`) reads the ledger those rows produce.

So an adapter exit before the response, a sandbox deadline, a ConversationServer restart, an interrupt without a `cancelled` stop reason, or a runtime that reports no usage all leave no stamp. Same shape as #1459 (gemini), but for every runtime.

## Fix

Stamp the conversation's inference source on the turn row at turn start, next to `model_selection` in `record_model_selection/2`, so the pricer can see "platform turn with unknown tokens" and the ceiling can count a floor. Keep the `{:done, ...}` path as the source of the token figure.

---
Found in the 2026-09-07 review of the week on `main` (cb93a7ea).

jhgaylor:
Superseded by #1916. The mechanism this issue asked for landed in #1917: every ACP turn that reaches session/prompt carries its inference source from turn start (turn_machine.ex, turn.ex, with regressions in turn_usage_test.exs and conversation_server_platform_inference_test.exs). What such a turn costs when no token figure arrives is the open decision in #1916, which now owns the remainder, so closing this one so the question lives in one place.

Issue #1459, opened 2026-09-03
Platform inference on the gemini runtime is unbilled: gemini reports no ACP usage, so no burn_inference row exists


jhgaylor:
Found by running ADR 0038's acceptance walk against all three platform keys on production, 2026-09-03.

## What happens

One turn per runtime on the acceptance account (`jhgaylor+acceptance@gmail.com`), which holds no inference credential of its own, so all three ran on platform keys:

| runtime | model | turn | `turns.usage` | ledger |
|---|---|---|---|---|
| claude | `anthropic/claude-sonnet-5` | completed, 5s | `{"input":2,"output":41,"cache_read":24101,"cache_write":9122,"inference":"platform","model":"anthropic/claude-sonnet-5"}` | `-3` `burn_inference` |
| codex | `openai/gpt-5.3-codex` | completed, 5s | `{"input":12975,"output":10,"cache_read":0,"inference":"platform","model":"openai/gpt-5.3-codex"}` | `-2` `burn_inference` |
| gemini | `google/gemini-3.1-pro-preview` | completed, 8s | **NULL** | **none** |

All three answered correctly — the gemini turn replied "I am Gemini, a large language model built by Google." It ran, it cost Google money on our key, and it was free to the tenant.

## Why

`Managoat.ACP.Usage.from_prompt_result/1` reads `"usage"` or `"_meta"` off the ACP `session/prompt` result and returns `nil` when the result carries neither. gemini-cli's ACP adapter returns neither, so:

1. `usage` is `nil`, and `TurnMachine.with_inference/2` has nothing to stamp — the `"inference" => "platform"` marker and the `"model"` key it needs both live on that map;
2. `Workers.CreditPricer` prices platform turns from those token counts, so it sees nothing to price;
3. `turns.usage` lands NULL.

## Why it is worse than a missing debit

`Fountain.PlatformInference`'s daily ceiling is **measured from the ledger** (its own moduledoc says so). A turn that writes no ledger row is invisible to the circuit breaker, so gemini platform spend does not count against `PLATFORM_INFERENCE_DAILY_CENTS`. The $50/day bound is a bound on what we can *see*, and gemini is currently outside it. A deployment could run well past the ceiling on gemini alone without it ever tripping.

## Options

- **Get usage out of gemini.** If the adapter can report tokens (a newer gemini-cli, or a different field), this is a `Managoat.ACP.Usage` change plus a quirk entry, and everything downstream works unchanged.
[...]

PR #1917, opened 2026-09-11
fix(billing): stamp the inference source at turn start (#1685)

jhgaylor:
## The mechanism

A turn on the platform inference key that ended any way other than a
`session/prompt` response spent Fountain's tokens with no ledger row, and did
not count toward `PLATFORM_INFERENCE_DAILY_CENTS`.

The `inference: platform` stamp was applied only inside
`TurnMachine.handle({:done, ...})` (`turn_machine.ex`, `record_usage(turn,
with_inference(usage, ctx))`), which a turn reaches only when its prompt is
*answered*. `Conversations._unsafe_record_turn_usage(turn, nil)` is a no-op,
so an adapter exit before the response, a sandbox deadline, a
ConversationServer restart, an interrupt without a `cancelled` stop reason, or
a runtime that reports no usage all left the turn with no usage map at all.
`CreditPricer.unpriced_inference_turns/1` selects on
`usage ->> 'inference' = 'platform'`, and `Billing.platform_inference_spend_today/1`
reads the ledger those rows produce, so none of it was visible to either.

Production, trailing 7 days: 26 turns carried the stamp; 319 ended with no
usage map (204 completed, 65 interrupted, 50 failed). Same shape as #1459
(gemini), but for every runtime.

## The fix

The stamp now goes on the row at turn start, from the `{:model_selected, …}`
report alone. `Managoat.ACP.Peer` sends that one from `send_prompt/1`,
immediately before it writes `session/prompt`, so it lands exactly once per
ACP turn at the moment tokens are about to be spent.

**Its sibling `{:model_selection_failed, …}` does not stamp.** All three peer
sources of that report are in `phase: :setting_model`, strictly before
`send_prompt/1` — the clause's own message says "No prompt was sent." — so the
turn spent nothing. An earlier revision of this PR stamped there too; review
caught it. Those rows are inert today (they price at zero, are excluded from
the pricer's page, render `usage: null`) but they are durable and land in
exactly the population #1916 exists to price, where nothing downstream could
tell "died mid-inference" from "never prompted". Production over 30 days: 13
failed selections against 905 successful.

`record_model_selection/3` takes the extra attributes to write rather than the
ctx, so the stamp travels in the same update as the selection that earned it
and only the one clause supplies it.

Two things keep the early and late writes from fighting:

- `with_inference/2` derives the source for both, so they cannot disagree, and
  it is still gated on `PlatformInference.enabled?()` — a deployment holding
  no platform key stamps nothing and its turn rows are byte-for-byte what they
  were.
- `_unsafe_record_turn_usage/2` **merges** the end-of-turn figure over a
  stamp-only map instead of refusing it. The stamp holds no token count, so it
  has contributed to no conversation sum and there is nothing to double. A
  turn that answers its prompt ends with exactly the usage map it has today.
  An `"accounting"`-only map is *not* a stamp — that is a real end-of-turn
  record, and a second one is still `{:error, :already_recorded}`.

The API keeps reporting `usage: null` for a turn carrying only the stamp
(`ConversationJSON.turn_usage/1`, used by the turns endpoint and the team
view): it holds no token figure, and `{input: 0, output: 0}` would be a number
nobody measured.

## The pricing decision — deliberately not made

`Credits.InferenceRates.cost_cents/1` prices a stamp-only row at zero, so it
writes no `burn_inference` row and contributes nothing to the daily ceiling.
**That is unchanged, on purpose.** "Visible to the pricer as a zero-token
platform turn" and "priced at a floor" are different decisions with different
money attached, and picking a floor means picking a number nobody has chosen.
Filed as **#1916** with what the decision has to answer.

One change to the pricer was needed to avoid trading one billing bug for a
worse one. `unpriced_inference_turns/1` now also requires a numeric token key.
The inference pass stops on a page that wrote nothing, and these turns
outnumber the answered ones better than ten to one in production — leaving
them in the selector would let a full page of unpriceable rows hold the cursor
still and starve the priceable turns behind them. The rows are on disk and
selectable by the same `usage ->> 'inference'` selector whenever #1916 lands;
whatever prices them re-admits them with a cursor that advances, or runs as a
pass of its own.

The ledger rules are untouched: still idempotent per row, still never summed
on a gate.

## Verification

Ten new tests, then **reverted the fix and watched them fail**:

- `conversation_server_platform_inference_test.exs` — a platform turn whose
  adapter dies before the response, one that is interrupted, and one that
  fails before it can report anything all still carry
  `usage->>'inference' = 'platform'`; a BYO-credential turn is stamped `own`
  and never `platform`; a **failed model selection stamps nothing at all**; a
  deployment with no platform key stamps nothing; a
  normally-completing platform turn's final usage map equals
  `with_inference(%{"input" => 5, "output" => 3}, ctx)` and the conversation's
  sums move once; the API reports null for a stamp-only map.
- `turn_usage_test.exs` — the end-of-turn figure merges over the stamp rather
  than being refused, counts once, and a genuine second record is still
  refused; an `"accounting"`-only map is not treated as a stamp.
- `credits_inference_test.exs` — a stamp-only turn burns nothing; a full page
  of stamp-only turns (501, one more than the pricer's batch, all older) does
  not starve the priceable turn behind them.

With the fix reverted, 7 of the 10 fail; the failed-selection test was verified
separately by putting the stamp back on that clause, which fails it alone. The other two assert behaviour that is
*unchanged* by design (the self-hosted no-stamp case and the pricing decision
left open), so they pass either way — those two were checked by mutating the
guard instead: making `stamp_inference/3` bypass `with_inference/2` and making
`inference_stamp_only?/1` return `true` unconditionally fails 8 tests,
including the pre-existing "a second figure for the same turn is refused".

Also run green: `turn_machine_test.exs`, `conversation_controller_test.exs`,
`team_controller_test.exs`, `conversations_usage_test.exs`,
`dashboard_live_test.exs`, `audit_guardrail_test.exs`,
`platform_inference_test.exs`, `schema_controller_guard_test.exs`,
`credits_test.exs`. Plus `mix compile --warnings-as-errors`,
`mix credo --strict` on the six changed lib files (no issues), and
`mix format --check-formatted` at both the root and `apps/fountain`.

## Left out

- No backfill of the 319 existing rows: nothing knows what they spent.
- No change to `Billing.platform_inference_spend_today/1` — it reads the
  ledger, and no new ledger rows are written here.
- The OpenAPI schema is unchanged; `usage` was already nullable.

Refs #1685 — deliberately not `Closes`. This PR makes those turns
**attributable**: the source is on the row, and the 319-a-week that answered
nothing are now queryable as platform turns. It does not make them **billed**,
which is the symptom in the issue's title, because what they should cost is the
open decision in #1916. #1685 should stay open until that lands.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1759,11 +1759,25 @@ defmodule Fountain.Conversations do
   differs per runtime (a per-call delta, a thread total, a per-step figure).
   A second call for the same turn would double-count the conversation, so
   it refuses when the turn already carries a usage.
+
+  The one usage map that is not an end-of-turn record is the turn-start
+  inference stamp (#1685): it carries no token figure, so it has counted
+  towards nothing and there is nothing to double. This write merges over it
+  — the stamp's keys are the two `TurnMachine.with_inference/2` writes here
+  as well, so a turn that answers its prompt ends with the map it would have
+  carried with no stamp at all.
   """
   def _unsafe_record_turn_usage(%Turn{}, nil), do: :ok
-  def _unsafe_record_turn_usage(%Turn{usage: %{}}, _usage), do: {:error, :already_recorded}
 
-  def _unsafe_record_turn_usage(%Turn{} = turn, %{} = usage) do
+  def _unsafe_record_turn_usage(%Turn{usage: %{} = recorded} = turn, %{} = usage) do
+    if Turn.inference_stamp_only?(recorded),
+      do: write_turn_usage(turn, Map.merge(recorded, usage)),
+      else: {:error, :already_recorded}
+  end
+
+  def _unsafe_record_turn_usage(%Turn{} = turn, %{} = usage), do: write_turn_usage(turn, usage)
+
+  defp write_turn_usage(%Turn{} = turn, %{} = usage) do
     # `usage` is whatever the runtime reported. The map is stored as it came,
     # but the counters it increments are bigints: a string or an object here
     # used to raise inside the transaction and take the turn's usage
--- a/apps/fountain/lib/fountain/conversations/turn.ex
+++ b/apps/fountain/lib/fountain/conversations/turn.ex
@@ -40,6 +40,10 @@ defmodule Fountain.Conversations.Turn do
     # never summed from the live `usage_update`s (their meaning differs per
     # runtime). Optional "accounting" preserves adapter scope/version/completeness;
     # it may be the only key when token counts are unknown. nil when nothing was reported.
+    #
+    # Two more keys, written at turn start rather than at its end (#1685):
+    # "inference" (and, on a platform turn, "model") — see
+    # `inference_stamp_only?/1`.
     field :usage, :map
     # ACP selection evidence, distinct from the agent's saved configuration.
     field :model_selection, :map
@@ -56,6 +60,32 @@ defmodule Fountain.Conversations.Turn do
 
   def statuses, do: @statuses
   def origins, do: @origins
+
+  # The two keys the turn-start inference stamp writes (#1685). Both are also
+  # written by `TurnMachine.with_inference/2` at the end of a turn that
+  # answers its prompt, which is why the late write merges over the early one
+  # rather than colliding with it.
+  @inference_stamp_keys ~w(inference model)
+
+  @doc """
+  Whether this `usage` map is the turn-start inference stamp and nothing else
+  (#1685) — the turn ran on a known inference source, and no token figure has
+  been recorded for it.
+
+  Such a row exists so the platform-inference pass can see the turn at all: a
+  turn that ends any way other than a `session/prompt` response never reaches
+  `{:done, ...}`, and before #1685 left no trace of whose key it spent. It is
+  not an end-of-turn usage record, so the two places that treat a usage map as
+  one — the "already recorded" refusal in
+  `Conversations._unsafe_record_turn_usage/2` and the API's turn `usage` field
+  — ask this first.
+
+  An `"accounting"`-only map (a runtime that reported its scope but no counts)
+  is *not* a stamp: that is a real end-of-turn record, and a second one must
+  still be refused.
+  """
+  @spec inference_stamp_only?(map()) :: boolean()
+  def inference_stamp_only?(%{} = usage), do: Map.keys(usage) -- @inference_stamp_keys == []
 
   def changeset(turn, attrs) do
     turn
--- a/apps/fountain/lib/fountain/conversations/turn_machine.ex
+++ b/apps/fountain/lib/fountain/conversations/turn_machine.ex
@@ -247,7 +247,7 @@ defmodule Fountain.Conversations.TurnMachine do
     end
   end
 
-  def handle(%__MODULE__{} = turn, {:model_selected, requested, effective, source}, _ctx) do
+  def handle(%__MODULE__{} = turn, {:model_selected, requested, effective, source}, ctx) do
     selection = %{
       requested_model: requested,
       effective_model: effective,
@@ -255,7 +255,7 @@ defmodule Fountain.Conversations.TurnMachine do
       status: "selected"
     }
 
-    turn = record_model_selection(turn, selection)
+    turn = record_model_selection(turn, selection, ctx)
 
     publish_stage(
       turn.conversation_id,
@@ -273,7 +273,7 @@ defmodule Fountain.Conversations.TurnMachine do
     handle(turn, {:failed, {:model_selection_failed, requested, detail}}, ctx)
   end
 
-  def handle(%__MODULE__{} = turn, {:failed, {:model_selection_failed, requested, detail}}, _ctx) do
+  def handle(%__MODULE__{} = turn, {:failed, {:model_selection_failed, requested, detail}}, ctx) do
     message =
       "Could not select model #{requested}: #{detail}. No prompt was sent. " <>
         "Choose a model available in this runtime, or refresh its model catalog and check account access."
@@ -285,7 +285,7 @@ defmodule Fountain.Conversations.TurnMachine do
       error: message
     }
 
-    turn = record_model_selection(turn, selection)
+    turn = record_model_selection(turn, selection, ctx)
 
     publish_stage(
       turn.conversation_id,
@@ -754,11 +754,41 @@ defmodule Fountain.Conversations.TurnMachine do
 
   defp put_model(usage, _source, _model), do: usage
 
-  defp record_model_selection(%__MODULE__{row: nil} = turn, _selection), do: turn
+  defp record_model_selection(%__MODULE__{row: nil} = turn, _selection, _ctx), do: turn
 
-  defp record_model_selection(turn, selection) do
-    {:ok, row} = Conversations._unsafe_update_turn(turn.row, %{model_selection: selection})
+  defp record_model_selection(turn, selection, ctx) do
+    attrs = stamp_inference(%{model_selection: selection}, turn.row, ctx)
+    {:ok, row} = Conversations._unsafe_update_turn(turn.row, attrs)
     %{turn | row: row}
+  end
+
+  # The inference stamp, written at turn start rather than only at the end
+  # (#1685).
+  #
+  # `Managoat.ACP.Peer` reports `:model_selected` from `send_prompt/1`,
+  # immediately before it writes `session/prompt`, so this runs once per ACP
+  # turn at the moment tokens are about to be spent. Before #1685 the stamp
+  # was applied only in the `{:done, ...}` clause, which a turn reaches only
+  # when the prompt is *answered*: an adapter exit, a sandbox deadline, a
+  # server restart, an interrupt or a runtime that reports no usage all left
+  # a turn that spent Fountain's key with no record that it had.
+  #
+  # `with_inference/2` decides the source, exactly as the end-of-turn write
+  # does — one derivation, so the two writes cannot disagree. What it returns
+  # for an empty map is the stamp alone, and `%{}` on a deployment holding no
+  # platform key, which is what keeps a self-hosted install's turn rows
+  # unchanged.
+  #
+  # The end-of-turn write merges its token figures over this (see
+  # `Conversations._unsafe_record_turn_usage/2`), so a turn that does answer
+  # its prompt ends with the same map it carried before #1685.
+  defp stamp_inference(attrs, %{usage: usage}, _ctx) when is_map(usage), do: attrs
+
+  defp stamp_inference(attrs, _row, ctx) do
+    case with_inference(%{}, ctx) do
+      stamp when map_size(stamp) > 0 -> Map.put(attrs, :usage, stamp)
+      _ -> attrs
+    end
   end
 
   @spec record_usage(t(), map() | nil) :: :ok
--- a/apps/fountain/lib/fountain/conversations/turn_machine.ex
+++ b/apps/fountain/lib/fountain/conversations/turn_machine.ex
@@ -255,7 +255,9 @@ defmodule Fountain.Conversations.TurnMachine do
       status: "selected"
     }
 
-    turn = record_model_selection(turn, selection, ctx)
+    # The only report that precedes a prompt going out, so the only one that
+    # stamps (#1685).
+    turn = record_model_selection(turn, selection, inference_stamp(turn.row, ctx))
 
     publish_stage(
       turn.conversation_id,
@@ -273,7 +275,7 @@ defmodule Fountain.Conversations.TurnMachine do
     handle(turn, {:failed, {:model_selection_failed, requested, detail}}, ctx)
   end
 
-  def handle(%__MODULE__{} = turn, {:failed, {:model_selection_failed, requested, detail}}, ctx) do
+  def handle(%__MODULE__{} = turn, {:failed, {:model_selection_failed, requested, detail}}, _ctx) do
     message =
       "Could not select model #{requested}: #{detail}. No prompt was sent. " <>
         "Choose a model available in this runtime, or refresh its model catalog and check account access."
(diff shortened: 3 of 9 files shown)


BinaryBourbon (changes requested):
Reviewed against the four questions a billing change has to answer before it lands. The mechanism is right and the pricer change is safe; I found one place where the code does not do what the PR body and your own comment say it does, and it writes durable rows that #1916 will read.

## What I verified, and how

**The pricer change loses no revenue.** This was the one I spent the most on, so here is the whole chain.

`unpriced_inference_turns/1` used to select on `usage ->> 'inference' = 'platform'` alone; it now also requires a numeric `input`/`output`/`cache_read`/`cache_write`. A row is only lost if it was *priced* before and is excluded now. It cannot be:

- `InferenceRates.cost_cents/1` sums `tokens(usage, key) * rate`, and `tokens/2` is `n when is_integer(n) and n >= 0 -> n`, anything else `-> 0`. So `cost_cents > 0` requires at least one of those four keys to decode as an Elixir integer, and a jsonb integer is `jsonb_typeof = 'number'`. The new predicate is implied by the old behaviour, not a narrowing of it.
- The three shapes worth worrying about all priced at **zero before this PR**, so excluding them changes no money: token counts as JSON strings (`is_integer("100")` is false), floats (`is_integer(5.0)` is false, still admitted by `jsonb_typeof` anyway), and an `accounting`-only map (no token keys at all).
- A runtime reporting under different key names was never priceable either, for the same reason.
- Upstream closes it further: `Managoat.ACP.Usage.normalize/1` returns `nil` or a map whose `"input"`/`"output"` are non-negative **integers** (floats truncated, numeric strings parsed). It cannot emit a non-numeric token key, and `TurnMachine.record_usage/2` is the only writer of `turns.usage`.

Then I checked it against production rather than trusting the argument:

```
-- every token key present anywhere in the turns table
input|number|2059   output|number|2059   cache_read|number|2050   cache_write|number|1469
-- (no non-number rows at all)

-- every turn ever priced into the ledger, vs the new predicate
priced_total = 83 | now_excluded = 0

-- key shapes of platform-stamped usage maps
{cache_read,inference,input,model,output}                54
{cache_read,cache_write,inference,input,model,output}    22
{inference,input,model,output}                            6
{cache_write,inference,input,model,output}                1
```

Zero rows lost. Also re-derived your 7-day figures: 319 turns ended with no usage map at all, exactly as the body says (the stamped count has since moved to 74 as the window slid). The problem is real and large.

**The stall argument holds.** `price_inference/3` stops on `n == 0 or length(turns) < @batch` (`credit_pricer.ex:234`), and `unpriced_inference_turns/1` has no cursor beyond the anti-join, which only hides rows that *wrote* a ledger entry. A zero-cost row is never written, so it is never hidden, and `order_by: [asc: t.ended_at]` re-serves the same page forever. I deleted the `jsonb_typeof` clause and re-ran `credits_inference_test.exs`: the 501-row page test fails with `inference: 0` against `inference: 1`. The starvation is not hypothetical and excluding rather than filtering is the right call.

**No double counting.** `_unsafe_update_turn/2` goes through `Turn.changeset` and `Repo.update` only, so the stamp touches no conversation counter; the merged end-of-turn write is the single `update_all` increment. `Map.merge(recorded, usage)` puts the late write second, so a normally-completing turn's map is byte-identical to what it carried before this PR even if the two derivations ever disagreed. After the merge the row is no longer stamp-only, so a genuine second record is still `{:error, :already_recorded}`. Confirmed by reverting the merge clause: `turn_usage_test.exs` and the "ends with the map it had before the stamp" test both fail.

**The stamp's placement.** `Managoat.ACP.Peer.send_prompt/1` (`peer.ex:1154-1163`) reports `{:model_selected, ...}` and then writes `session/prompt` on the next line. The interrupt and adapter-exit paths do not stamp on their own, they inherit the stamp from a prompt that already went out, which is correct: those turns did spend tokens. `stamp_inference/3`'s `%{usage: usage} when is_map(usage)` clause makes a repeat a no-op, so a reattach cannot double-stamp.

**API surface.** `usage` is already `%Schema{oneOf: [TurnUsage], nullable: true}` at both `schemas.ex:784` and `:2525`, no schema file changes, so no SDK contract artifact moves. `ConversationJSON.turn_usage/1` and `TeamJSON` are the only two renderers of `turn.usage` in the tree, and both are updated.

**The tests.** All nine check out. Reverting the pricer clause fails 1, reverting `stamp_inference` fails 4, reverting the merge fails 2, reverting `turn_usage/1` fails 1. The remaining two are guards, and both fail when mutated: dropping the `PlatformInference.enabled?()` check in `with_inference/2` fails "a deployment holding no platform key stamps nothing", and adding `accounting` to `@inference_stamp_keys` fails "an accounting-only map is an end-of-turn record". `mix format --check-formatted` clean; turn_machine, turn_usage, platform-inference, conversation and team controller suites green (159 + 17 + 23 tests, 0 failures).

I am not asking for the pricing decision here. Leaving a token-less row at zero and filing #1916 is the right shape.

## Blocking: a turn whose model selection *failed* is stamped, and no prompt was sent

`turn_machine.ex:288` — the `{:failed, {:model_selection_failed, requested, detail}}` clause now calls `record_model_selection(turn, selection, ctx)`, which stamps through the same `stamp_inference/3`.

That clause is reached from three places in the peer (`peer.ex:566`, `:846`, `:1080`), all of them in `phase: :setting_model`, all of them strictly before `send_prompt/1`. The message the clause itself writes says so: *"No prompt was sent."* The turn spent zero tokens on anybody's key.

Proved it rather than read it. Driving only that clause with a platform ctx:

```
usage on a turn where NO PROMPT WAS SENT: %{"inference" => "platform", "model" => "anthropic/claude-opus-5"}
```

This contradicts the invariant the PR body states ("a turn that dies *before* the prompt went out, which spends nothing, is not stamped") and the comment at `turn_machine.ex:767-772` that repeats it.

Why it matters even though it costs nothing today. It is inert right now: `cost_cents/1` gives zero, the new `jsonb_typeof` clause keeps it out of the pricer's page, and the API reports null. But the stamped-and-token-less population is *precisely* the set #1916 exists to price. These rows are written now, they are permanent, and once they are on disk nothing distinguishes "died mid-inference, tokens unknown" from "never sent a prompt, tokens provably zero". Whatever floor #1916 picks would charge for the second. In production that is 13 failed selections in the last 30 days against 905 successful ones, so it is a steady trickle rather than a one-off.

The fix is one line: leave the failure clause on a stamp-free path (keep a `record_model_selection/2` for it, or pass `%{}` as its ctx), and give it a test alongside the five you already wrote. Worth doing here rather than in #1916, because #1916 will be reasoning about rows this PR creates.

## Non-blocking

1. `peer.ex:1154` reports `:model_selected` before `send_request` writes, and `send_request` can leave `phase: :failed` — the prompt never goes out but the stamp is already on the row. Same class as above and genuinely unavoidable, since the host cannot retract bytes it has not written yet (which is exactly why `:prompt_sent` is reported *after* the write, two lines down). Not asking for a change, but it belongs in the comment at `turn_machine.ex:767` so the next reader knows the stamp means "a prompt was about to be written", not "a prompt was written".

2. The exclusion closes the instance, not the class. A platform turn with numeric tokens that rounds to zero cents is still admitted and still writes nothing, so a full page of those stalls the pass exactly the same way (`price_turns/3` at `credit_pricer.ex:146` has the identical shape). There are none in production today — all 83 platform-stamped turns are priced — so this is a note, not a request. A cursor on `ended_at` rather than pure anti-join paging is the real fix whenever someone touches this next.

3. `Turn.inference_stamp_only?/1` returns `true` for `%{}` (`turn.ex:88`). Unreachable today, because `Usage.normalize/1` never returns an empty map and `record_usage/2` is the only writer. But if some future writer ever records `%{}`, the "already recorded" refusal silently turns back into a merge and the double-count door reopens. A `map_size(usage) > 0` guard would cost nothing.

4. `schemas.ex:787` still describes `usage` null as "while the turn runs, when the runtime reported none, or on turns that predate the field". The stamp-only case is arguably covered by "reported none", but naming it would save a reader the trip.

Everything else here is careful work: one derivation for both writes, the `enabled?()` gate keeping self-hosted rows byte-identical, the merge ordered so the late write wins, and tests that actually fail when the fix is removed. Fix the failure-clause stamp and I am happy.

jhgaylor:
Fixed in 35338219. You were right, and the contradiction was mine: the comment claimed the stamp marks "the moment tokens are about to be spent" while the code also stamped the one report that means the opposite.

**What changed**

- `{:failed, {:model_selection_failed, …}}` no longer stamps. It records the model selection and nothing else.
- `record_model_selection/3` now takes the *extra attributes to write* rather than the ctx, so the stamp travels in the same update as the selection that earned it and only the `{:model_selected, …}` clause supplies one. `stamp_inference/3` became `inference_stamp/2`, returning `%{usage: stamp}` or `%{}`.
- The comment at the helper now says what the code does, and names why: `:model_selection_failed` is raised in `phase: :setting_model`, strictly before `send_prompt/1`, and a stamp there would be indistinguishable afterwards from a turn that died mid-inference — the exact question #1916 has to answer.
- PR body updated the same way: the "a turn that dies before the prompt went out is not stamped" claim now describes the two clauses explicitly instead of asserting an invariant the code did not hold.

**Test**

`a failed model selection stamps nothing: no prompt went out` in `conversation_server_platform_inference_test.exs` — drives the clause alone with a platform ctx and asserts `is_nil(turn.usage)` while `model_selection["status"] == "failed"`. Verified by putting the stamp back on that clause (both the `_ctx` → `ctx` rename and the `inference_stamp(turn.row, ctx)` call): it is the only test that fails, and it fails on the `usage` assertion. Restored, and `turn_usage_test`, `turn_machine_test`, `conversation_server_platform_inference_test` and `credits_inference_test` are 90/90 green, with `mix format --check-formatted`, `mix compile --warnings-as-errors` and `mix credo --strict` on the file clean.

`Refs #1685` is unchanged — agreed, the symptom in the title is still true after this.

BinaryBourbon (approved):
`35338219` fixes it. Re-verified rather than read.

## The blocking finding is closed

`turn_machine.ex:293` now passes `%{}`, and `record_model_selection/3` takes the attributes to write rather than the ctx, so the stamp can only travel with the selection that earned it. Grepping every caller of `record_model_selection/3` and `inference_stamp/2` in `apps/` and `ee/` finds exactly two call sites: `:260` (the `:model_selected` clause, stamped) and `:293` (the failure clause, `%{}`). `{:model_rejected, …}` at `:277` delegates into the failure clause, so it inherits the empty extras. There is no third path, and the clause head takes `_ctx` now, so a future edit cannot quietly reach for one.

My own revert, not the author's: I put the stamp back on the failure clause (`ctx` in the head, `inference_stamp(turn.row, ctx)` as the extras) and re-ran the file. Exactly one test failed, on the assertion that matters:

```
1) test the mark at turn start (#1685) a failed model selection stamps nothing: no prompt went out
   code: assert is_nil(stored(row).usage)
```

The test is a real guard, and it is the only one that moves — which is the right blast radius for this change.

## The success path is still byte-identical

You asked me to re-confirm this against the new shape rather than carry the old verification forward, so I did it independently: drove `{:model_selected, …}` then `{:done, "end_turn", %{"input" => 5, "output" => 3, "cache_read" => 7}}` through a real row, and compared the stored map against `with_inference(reported, ctx)`, which *is* the pre-PR write verbatim. Three ctx shapes:

```
platform: got  = %{"cache_read" => 7, "inference" => "platform", "input" => 5, "model" => "anthropic/claude-opus-5", "output" => 3}
          want = %{"cache_read" => 7, "inference" => "platform", "input" => 5, "model" => "anthropic/claude-opus-5", "output" => 3}
own:      got  = %{"cache_read" => 7, "inference" => "own", "input" => 5, "output" => 3}
          want = %{"cache_read" => 7, "inference" => "own", "input" => 5, "output" => 3}
no source: got = %{"cache_read" => 7, "input" => 5, "output" => 3}
          want = %{"cache_read" => 7, "input" => 5, "output" => 3}
```

Equal in all three, and the conversation's sums moved once each (`usage_input_tokens: 5, usage_output_tokens: 3`). The `own` row is worth calling out: the stamp for an own turn is `%{"inference" => "own"}` with no `"model"` key, so the merge cannot leave a stale model id behind on a turn nothing prices. I also checked the case the PR exists for — `{:done, "end_turn", nil}` leaves `%{"inference" => "platform", "model" => …}` on the row and moves no counter, which is the whole point.

`Map.put(extra, :model_selection, selection)` puts the selection last, so extras can never shadow it. `inference_stamp(nil, ctx)` on a `row: nil` machine falls to the second clause and its result is discarded by `record_model_selection/3`'s own `row: nil` clause, so no crash and no write.

The comment at `:766-790` now says what the code does, including the `phase: :setting_model` reason, and it answers the note I left last time about the stamp meaning "a prompt is about to be written" rather than "was written".

## Unchanged halves

`git show --stat` confirms this commit touches only `turn_machine.ex` and the one test file, so last round's verification of the pricer selector, the merge in `_unsafe_record_turn_usage/2` and the API renderers still stands. Briefly re-confirmed the conclusion that mattered most: the `jsonb_typeof` requirement loses no revenue — `cost_cents/1` needs `is_integer`, `Usage.normalize/1` can only emit integers, and in production every token key in the whole `turns` table is a JSON number, with 0 of the 83 turns ever priced as `burn_inference` excluded by the new predicate.

Green here: `mix format --check-formatted`, and 67 + 133 tests across turn_machine, turn_usage, platform-inference, credits_inference and the conversation and team controllers, 0 failures.

`Refs #1685` reads correctly to me — the symptom the issue describes is still true after this lands, and the PR body is explicit that the pricing decision is #1916's.

## Approving. Three notes carried forward, none blocking

1. The exclusion closes the instance, not the class: a platform turn with numeric tokens that rounds to zero cents is still admitted and still writes nothing, so a full page of those stalls the pass the same way (`credit_pricer.ex:146` and `:234` share the shape). None exist in production today — all 83 platform-stamped turns are priced. A real cursor on `ended_at` is the fix whenever someone is next in there.
2. `Turn.inference_stamp_only?/1` (`turn.ex:88`) returns `true` for `%{}`. Unreachable today, since `Usage.normalize/1` never returns an empty map and `record_usage/2` is the only writer, but a `map_size(usage) > 0` guard would stop a future writer from silently reopening the double-count door.
3. `schemas.ex:787` still lists the reasons `usage` can be null without naming the stamp. Covered by "reported none", so only worth a word if you are in the file anyway.

Good change, and the revised body is honest about the round trip.

jhgaylor:

PR #1918, opened 2026-09-11
fix(broker): start the listener before the endpoint, and fail readiness while it is down

### LINKED ISSUE #1726: A pod is Ready before its broker listener is up, so provisions during a rollout fail with :listener_down
## Symptom

During the 2026-09-07 05:44 UTC rollout, two provisions failed outright:

```
05:45:49 provision|failed  {"reason":"{:broker, :unreachable, :listener_down}"}
05:45:49 broker|failed     {"reason":"broker_unreachable","detail":":listener_down"}
05:45:46 provision|failed  {"reason":"{:broker, :unreachable, :listener_down}"}
05:45:46 broker|failed     {"reason":"broker_unreachable","detail":":listener_down"}
```

Pods at that moment:

```
fountain-56c6fc766d-48nsw   started 05:44:31    <- new
fountain-56c6fc766d-qghrj   started 05:44:21    <- new
fountain-779b5b95b8-78rn8   started 04:12:47    <- old, terminating
```

The failures land 75 seconds after the new pods start. This is not the old pod going away; it is the **new** pods taking work they cannot do yet.

## Cause

`/health/ready` checks the database and nothing else:

```elixir
def ready(conn, _params) do
  checks = %{"database" => Fountain.Health.database()}
  ...
end
```

The broker listener runs **in-process** on `BROKER_LISTEN_PORT`, started as a supervised child. Readiness does not wait for it, so kubelet puts the pod in rotation as soon as Postgres answers. Every provision routed there in the gap fails, because with `BROKER_TENANTS=*` every tenant is brokered — there is no unbrokered path to fall back to.

## This belongs in the probe by the module's own rule

`Fountain.Health`'s moduledoc sets the bar:

> The rule for what belongs here: a check earns its place only if a failure means this pod should stop receiving traffic. That is a higher bar than "is something wrong".

A pod whose broker listener is down cannot provision a sandbox. It should stop receiving traffic until it can.

It is also unlike the two documented exclusions. **Sprites** is excluded because it would put a third party's uptime on our serving path — the broker listener is our own in-process socket, so that argument does not apply. **Migrations** are excluded because a failure there means no listening socket to probe at all — the broker listener starting late is precisely the case where the socket *is* up and the pod is *not* ready.

Suggested check: the listener is accepting on `BROKER_LISTEN_PORT`, evaluated only when `:broker_listen_port` is configured, so a deployment with brokerage off is unaffected.

## Why it looked like noise

I saw `:listener_down` yesterday while diagnosing #1667 and discounted it: every failure named a pod IP that no longer existed, which reads as a shutdown race and therefore harmless. That was wrong in the part that matters. The same signal also fires on **startup**, and there it fails new work rather than tidying up old work.

## It is not theoretical

It just broke CI twice. The Review Loop bot runs on Fountain under `jhgaylor+review-loop@gmail.com`; its provision landed in the 05:44 rollout window and its check failed on both open PRs:

```
Review Loop · Human review needed
Agent task failed: Required independent turn did not complete successfully
```

So the failure mode is: deploy Fountain → Fountain's own PR review bot fails → the PRs that fix Fountain need a human. Frequency scales with deploy rate, which is exactly when it is least welcome.

## Related

- #1667 — `session_gone`, the other way a deploy costs a user their next turn
- #1671 — broker CA install failing while the stage reports `done`

Both are the same theme: broker setup problems that do not surface as broker problems.

jhgaylor:
A pod must keep its egress broker available while it serves requests and drains conversations. Previously the broker started after the endpoint and stopped before it, so provisioning or reattachment during a rollout could fail with `:listener_down`.

This change starts the broker request-log writer and listener after the repository/migrator and before their consumers. The endpoint remains last, and main's execution transport supervisor and deadline worker remain in the tree. A new ordering regression covers those bounded-execution children.

`/health/ready` also checks the configured broker listener using the same preflight predicate as provisioning. A missing listener returns 503 with a bare dependency result; deployments without `BROKER_LISTEN_PORT` remain ready when the database is healthy. `/health` remains independent of dependencies.

The readiness schema example, generated TypeScript API output and operational documentation describe the added check. Main has been merged into this branch with the supervision-order conflict resolved.

`sdk-no-release`: the only SDK diff is the generated OpenAPI JSDoc readiness example. Exported types, serialization, transport and package APIs are unchanged, so no SDK version bump is warranted.

Validation: the isolated compile, format, strict Credo, Dialyzer, Sobelow and production release assembly passed. Focused readiness/order coverage passed 29 tests. Regenerated contract/TypeScript checks passed (174 operations, 202 schemas); the contract JSON itself did not change. Final complete `mix precommit --max-cases 4 --seed 5820` passed on signed commit `2fa5cb5ce4bb7bd635a500777b48c9951a254af0` with four BEAM schedulers and a fresh dedicated database: 6 doctests and 5,615 core tests, 144 Buzz tests and 33 Support tests, all with zero failures (2 skipped, 7 excluded). A later controlled full run also hit the unchanged docs heading-index guard's 60-second timeout inside Lumis.highlight; all existing parser and compiled-module files predated that run. Fresh-VM native probes using warm disk and parser-only caches completed the same 111-page checks in under four seconds, and eight concurrent heading checks completed in 2.6 seconds. These probes did not reproduce the stall; its cause remains unresolved.

The first full run overlapped several other suites and failed the unchanged SSE heartbeat assertion tracked by #1934 (the interrupted run reported 1 doctest, 4,707 tests and one failure before shutdown; it did not complete the full suite). After stopping the contending runs and resetting the dedicated database, the unchanged SSE module passed 18/18 twice, including original seed 5820, with its original timing assertions. This records the failure rather than treating a later pass as proof of its cause. Production rollout behavior has not been exercised by this repair.

Closes #1726


BinaryBourbon (approved):
Approving. The check is correct, it is genuinely cheap, and reusing `preflight/0` rather than inventing a second notion of "listening" is the right call — I traced both halves and they cannot drift. Two things I could not close out are below; neither blocks the merge, but the first bears on whether #1726 should be closed by it.

## What I verified

**The predicate really is the one a provision fails on.** `Fountain.Health.broker_listener/0` → `Fountain.Broker.preflight/0` (`apps/fountain/lib/fountain/broker.ex:504`) → `impl(:native).preflight()`. `Fountain.Conversations.Provisioning.check_broker_support/4` (`apps/fountain/lib/fountain/conversations/provisioning.ex:357`) calls the same function and publishes the `broker_unreachable` / `:listener_down` stage from its `{:error, {:broker, :unreachable, detail}}`. Same call, same atom.

**Nothing else `impl/1` can reach does I/O.** `backend/0` returns `:native` or nil and `defp impl(:native), do: Native` is the only clause, so `@type backend :: :native` is the whole surface. `Native.preflight/0` is `Managoat.Broker.running?/0`, which is `Process.whereis(Managoat.Broker.Listener)` plus `Process.alive?/1` (`deps/managoat_broker/lib/managoat/broker.ex:139-144`) — no socket dial, no pool checkout, nothing that can block. The `@check_opts` argument holds; no finding here.

**The skip holds in every configuration I could find.** The important one is `Fountain.Application.broker_children/0` (`apps/fountain/lib/fountain/application.ex:126`): it gates on `Fountain.Broker.backend() == :native`, which *is* `configured?/0`. So `configured?/0` true ⟺ the listener child was started — there is no config shape where the probe waits for a listener nobody asked for. Then:

- CI's release boot check and the core-release boot check (`.github/workflows/ci.yml`, both `probe /health/ready` steps) set no `BROKER_LISTEN_PORT` in their `env:` blocks → skip → 200.
- Compose: `docker-compose.yml:191` passes `${BROKER_LISTEN_PORT:-}`, `.env.compose.example:148` ships it commented out, and `config/runtime.exs:206` maps `""` to nil. `scripts/compose-boot-check.sh` copies that file verbatim. The compose `healthcheck` is `/health`, not `/health/ready`, so it was never in scope anyway.
- Dev and test set nothing.
- Permanent-503: the listener sits in the same `:one_for_one` supervisor as the endpoint, so a bind failure fails `Supervisor.start_link` and the node never serves, and a crash loop exhausts restart intensity and stops the app. I could not construct a pod that serves HTTP and 503s readiness forever.

**App env and the port-0 claim.** `BrokerTestHelpers.enable_broker_for/1` captures all three keys before writing and restores them in `on_exit` (`delete_env` where the original was nil); `configure_broker/0`'s own `put_env(:broker_listen_port, 0)` lands after that callback is registered, so it is restored too. The file is `async: false`, and ExUnit runs sync modules after every async one, so nothing reads those keys concurrently. `listener_spec/0` passes the config value straight to ThousandIsland, so 0 is a real ephemeral bind — two worktrees cannot collide.

**Revert.** Removed the `configured?/0` skip and re-ran: **5 failures of 23**, three of them in `health_controller_test.exs` (`returns 200 when every dependency answers`, `is publicly accessible without authentication`, `reports no detail about the failure`). Restored; tree clean.

**Gates.** `mix format --check-formatted` from `apps/fountain` (exit 0); `mix credo --strict` on both changed source files (no issues); the three health files 23/0; `force_ssl_probe_test.exs` + `docs_test.exs` 34/0. `scripts/docs-style.py` reports 2 findings, both pre-existing and on other pages. `vale --config .vale-ste.yml` on the edited page: 129 findings against 128 on the base revision, 0 errors in both, the one added being an advisory `STE.Vocabulary` suggestion. The docs curl example's key order matches what Jason actually emits — I checked rather than assuming.

## 1. The incident timeline does not fit the window this closes

I measured the thing the fix is sized against. Starting `Native.listener_spec()` and getting `Managoat.Broker.running?/0` to true takes **~100ms** on this machine, and it happens inside the same synchronous `Supervisor.start_link` that started `FountainWeb.Endpoint` one child earlier. So the gap between "endpoint accepts" and "listener up" is on the order of a hundred milliseconds, not seconds.

The failures in #1726 are at 05:45:46 and 05:45:49 — **three seconds apart**. Two requests three seconds apart cannot both land in a shared ~100ms window, and neither is 75s from a pod start by way of child-start ordering.

That does not make the check wrong; a readiness probe that answers what a provision answers is right on its own merits. But the PR's causal story ("the new pods taking work they cannot do yet") is not established by the evidence in the issue, and the issue itself records the competing reading: *"every failure named a pod IP that no longer existed"*. If the real window was the draining pod, this merges a good check and #1726 recurs on the next rollout. I would not auto-close the issue on merge — watch one rollout, or better, pull the pod's own boot log lines and confirm the listener was actually down at 05:45:46.

## 2. The ordering is fixed at one end only

`apps/fountain/lib/fountain/application.ex:102`:

```elixir
      FountainWeb.Endpoint
    ] ++ broker_children()
```

Supervisors terminate children in **reverse** order, so on SIGTERM the broker listener stops *first* and the endpoint drains afterwards. A pod in `Terminating` therefore serves HTTP for its whole drain with the broker down — which is exactly the shape of the `:listener_down` failures whose source IP no longer existed. Readiness cannot help at that end: the endpoints controller already excludes Terminating pods, so a 503 there changes nothing about what still arrives on a kept-alive connection.

Moving `broker_children()` **before** `FountainWeb.Endpoint` in that list fixes both ends at once: the listener is up before anything can be routed in, and it is torn down after the endpoint has stopped accepting. It would make this probe belt-and-braces rather than the only defence. I looked for a reason the broker must come after the endpoint and did not find one — `Repo` starts far earlier, `RequestLog` is inside `broker_children()` and already ordered first, and a bind failure crashloops the node either way. Worth a follow-up, or a one-line change here if you would rather not leave the drain half open.

## 3. Smaller things, none blocking

- **The moduledoc argues the bar with `BROKER_TENANTS=*`, but gates on `BROKER_LISTEN_PORT`** (`apps/fountain/lib/fountain/health.ex:25-38`). During the ADR 0019 §9 ratchet — port set, `BROKER_TENANTS` blank or a partial list, which is the documented rollout path — a listener-down pod is now held out of rotation although every tenant it serves is unbrokered. That is the right trade (a probe has no tenant to ask about) and the cost is the ~100ms window, but the bullet currently reads as if the wildcard case were the only one. One clause saying the gate is the port and not the ratchet, because a probe has no tenant, would make it argue what the code does. The two exclusion rebuttals I checked and both hold.
- **The rescue is insurance against a raise, not against a slow backend.** The docstring offers it as cover "for a backend that ever does more" — a `rescue` would not bound a backend that dials a socket. The real guard is `@type backend :: :native` and the single `impl/1` clause. Fine today; if a second backend ever arrives, this check needs a timeout and the docstring should not suggest it is already covered.
- **`FountainWeb.Schemas.ReadinessResponse`** (`apps/fountain/lib/fountain_web/schemas.ex:2421`) still carries `example: %{"database" => "ok"}`. No contract consequence — `checks` is `additionalProperties`, so the OpenAPI document and all four SDKs are unchanged and nothing needed regenerating (I checked `sdk/contract/contract.json` and the generated TS) — but the example now understates the payload.
- **A hazard for later, not for this PR.** The three CI probe helpers (`scripts/compose-boot-check.sh:35`, and the two `probe()` functions in `ci.yml`) treat the first non-200 as a verdict rather than a retry. Today none of them set `BROKER_LISTEN_PORT`, so they are safe. The moment a fixture turns brokerage on, the boot-window 503 this PR deliberately introduces becomes a hard CI failure that will read as a flake. Teaching those loops to retry a 503 costs nothing now.

jhgaylor:
Thanks — pushed `ea609fbb`, and the review changed what this PR is.

**Your causal challenge was right, and I checked it against production.** The
startup window does not fit: the listener starts ~100ms in, inside the same
synchronous `Supervisor.start_link` as the endpoint, and #1726's failures are
three seconds apart, 75 seconds after pod start. The most recent occurrence,
`09:22:10` on `fountain_server@10.42.3.141`, came from a pod whose first event
was `09:13:37` and whose replacement ReplicaSet appeared at `09:21:29`. Nine
minutes into its life, 41 seconds into its own drain. Most of these events are
`reattach|failed`, not `provision|failed`.

The window is the **shutdown**, and it is structural: `broker_children()` was
appended after `FountainWeb.Endpoint`, and `:one_for_one` terminates in
reverse, so the listener stopped first on every rollout while the endpoint went
on serving and the conversation servers went on reattaching. It also vindicates
the one observation #1726 dismissed — every failure named a pod IP that no
longer existed.

So commit 2 moves `broker_children()` ahead of everything that can ask the
proxy to broker, and `Closes #1726` is now `Refs #1726`; readiness is not what
stops these.

I went one step further than "before the endpoint": the listener also has to
outlive `Fountain.ConversationSupervisor`, because that is where the dominant
`reattach|failed` case comes from. I looked for a constraint forbidding it and
found none — `RequestLog` touches only `Fountain.Repo`, the listener needs the
repo and `ca_seed/0`, and nothing in `broker_children()` reaches Oban, PubSub,
Horde or the endpoint. `application_children_test.exs` fails on the old order
(3 of 7) *and* on the weaker just-before-the-endpoint placement (1, the
reattach case), so both claims are pinned rather than asserted.

Your three findings are in:

- **Ratchet vs gate.** The moduledoc now says it outright: mid-ratchet the
  check is stricter than it has to be, because `BROKER_LISTEN_PORT` is what
  decides whether a listener exists and no load balancer can sort brokered
  traffic from unbrokered. It also now says the check closes only the starting
  end, and points at the ordering for the other one.
- **The `rescue`.** You are right that it bounds nothing slow. It now says it
  bounds a raise and an exit and nothing else, and that a backend which could
  block needs a real `@check_opts`-shaped bound added there.
- **The stale `example`.** Fixed, and I verified your contract finding rather
  than taking it: `scripts/sdk-contract/build.sh --check` passes with no diff,
  164 operations and 190 schemas.

Also took the `probe()` note — all three helpers treat the first non-200 as a
verdict with no retry, so there is now a comment saying that the reorder is
what makes a transient 503 unreachable at boot, and that no fixture sets
`BROKER_LISTEN_PORT` today.

Thanks for tracing `backend/0` and `running?/0` yourself, and for the
`broker_children/0` observation — `configured?` true ⟺ the listener child was
started is a cleaner statement of the skip's correctness than the one I wrote.

Pushing has dismissed the approval; this wants another look, since the ordering
change was not in what you reviewed.

jhgaylor:
Correction to the note at the end of my last comment: the push did **not**
dismiss the approval. `reviewDecision` is still `APPROVED` against
`ea609fbb`, so this PR is mergeable as it stands — stale-review dismissal is
not on for this branch.

That matters here, because the ordering change in commit 2 is not what was
reviewed. I dequeued the PR to push (it had reached position 3 while I was
working); please do not re-queue it until commit 2 has had a look.

jhgaylor:

BinaryBourbon (approved):
Re-confirming the approval against `ea609fbb`, which I had not seen when I approved `2ed4679a`.

The stronger placement is right, and the reasoning behind it is better than the one I offered. I said "before the endpoint"; you read the evidence again and found that most of these events are `reattach|failed` raised by a `ConversationServer` under `Fountain.ConversationSupervisor`, so merely preceding the endpoint leaves the dominant case open. That is correct, and the test proves it rather than asserting it — see the revert reproduction below.

## The dependency argument holds

I traced it independently, because this is the change that fails in production rather than in CI.

**What the broker children need at start.** `Native.listener_spec/0` reads three application-environment keys and nothing else (`:broker_listen_port`, `:master_secrets_key` via `ca_seed/0`, `:broker_allow_private_upstreams`). `Managoat.Broker.init/1` builds an ETS certificate cache, writes one `:persistent_term` entry and starts a ThousandIsland child — no repo, no PubSub, no Horde, no Oban, no endpoint. `RequestLog.init/1` returns a static map and reaches `Fountain.Repo` only when it flushes. The telemetry handler is a `Logger` line plus a `GenServer.cast`. `Fountain.Repo` really is the only dependency, and it is far ahead.

**What ahead of the broker needs the broker.** One reference only: `FountainWeb.Telemetry:681` registers the poller measurement `{Fountain.Broker.Native, :emit_telemetry, []}`, and that is guarded by `Fountain.TelemetryTick.run/2` and reports `up: 0` when the listener is absent (`native.ex:463-484`). This PR shrinks that window rather than opening one.

**Then I booted it.** `Application.ensure_all_started(:fountain)` against the new tree with `:broker_listen_port` set to 0:

```
ensure_all_started: :ok in 375.7 ms
broker running?: true
preflight: :ok
endpoint up?: true
conv supervisor up?: true
start order tail: [Fountain.Broker.Native.RequestLog, Managoat.Broker,
                   Fountain.ConversationRegistry, Fountain.ConversationSupervisor,
                   Fountain.RunnerRegistry, FountainWeb.Endpoint]
```

That tail is `Supervisor.which_children/1` reversed, so it is the real runtime order, not a reading of the source. Termination reverses it: endpoint, runner registry, conversation supervisor, conversation registry, then the broker. The tree is what `children/0` claims it is.

## `attach_telemetry/0` opens no window

The old comment's ordering claim survives. Previously the attach ran while the children *list* was being built; now it is an explicit statement in `start/2`. Both are before `Supervisor.start_link/2`, so both are before any child process exists and therefore before the listener can accept a connection — the new one is if anything marginally earlier. No request can find the handler missing.

## The revert that justifies the stronger placement — reproduced

Both rows of your table, run here:

| Change | Result |
|---|---|
| `broker_children()` back after `FountainWeb.Endpoint` | **3 of 7 fail** — `starts before the endpoint`, `starts before the conversation supervisor`, `the endpoint is last` |
| `broker_children()` moved to immediately before `FountainWeb.Endpoint` | **exactly 1 fails** — `the listener starts before the conversation supervisor, so it outlives it` |

So the placement is load-bearing and the test is what holds it there. Restored after each; `git status` clean.

## `terminationGracePeriodSeconds`: no manifest change needed before this ships

I looked at this as a possible blocker and it is not one, for a reason that is easy to miss: **the reorder does not change the sum.** The broker was terminated *first* before this PR and waited exactly the same amount for its connections then; now it waits later in the sequence. Total drain is unchanged, only its position.

It is also bounded rather than open-ended. ThousandIsland defaults `shutdown_timeout: 15_000` (`deps/thousand_island/lib/thousand_island/server_config.ex:40`), applied as each connection child's `shutdown` (`connection.ex:32`), so a hung CONNECT tunnel costs at most ~15s. Against `terminationGracePeriodSeconds: 120` and a 5s `preStop` (`deploy/k8s/deployment.yaml:38,51`) there is ample headroom. If that wait ever does need tightening, the knob is `shutdown_timeout` in `listener_spec/0`, not the manifest.

## Two things to fix, neither blocking

**1. The code does not sit where the prose says it sits** (`apps/fountain/lib/fountain/application.ex:143-146`, and the same sentence in the commit message). Both say the broker now sits "after `Fountain.Repo`, which is the only thing they need". It actually sits after the whole first block — after `Oban`, `Analytics.Sink`, `Team.Comms.Inbound.Seen` and the rest. That has a consequence at the other end: reverse termination stops the broker **before** Oban, so an Oban job still running in the tail of a drain can still meet `:listener_down`.

That residue is strictly smaller than what you started with (the broker used to stop first, exposing Oban for the entire shutdown), and I found no worker under `apps/fountain/lib/fountain/workers/` that launches or provisions a conversation, so it looks empty today. But the comment is what a future reader will trust. Either move the entry to match the prose — immediately after the `{Ecto.Migrator, ...}` child, which closes the Oban end too — or say where it actually is and why after-Oban is far enough.

**2. One decision, read in two places.** `start/2` attaches the handler under `if Fountain.Broker.backend() == :native`, and `broker_children/0` builds the children under the same test. The split buys `children/0` its purity and is worth keeping, but if the two ever diverge the failure is a listener with no request-log handler: `broker_requests` silently stops filling and `/api/conversations/:id/egress` goes quiet, with nothing failing anywhere. One comment in `start/2` naming `broker_children/0` as the other half of the same decision would be enough.

*(Marginal: the test's `before?/3` compares `Enum.find_index` first occurrences. Fine today — one `Horde.DynamicSupervisor`, one `Managoat.Broker` — but a second named listener would map to the same id through `id/1` and the assertion would silently pick one. Not worth changing now.)*

## The review fixes from round one

All three landed correctly. The moduledoc now says plainly that mid-ratchet the check is stricter than it has to be and why (`health.ex:40-48`), and that it closes only the starting end — which is exactly the right relationship to state now that ordering closes the other. The docstring no longer implies the `rescue` bounds a slow backend and names `@check_opts` as the shape a real bound would take. `ReadinessResponse`'s example names both checks, and the contract genuinely cannot move: `sdk/contract/contract.json` records no `example` keys at all (`grep -c '"example"'` → 0), so that is structural, not just observed.

## Gates run here

`mix format --check-formatted` and `MIX_ENV=test mix compile --warnings-as-errors --force`, both exit 0 from `apps/fountain`. `mix credo --strict` on all three changed source files: no issues. Tests: the four health and ordering files 30/0; `broker_native_test.exs`, `runtime_config_test.exs`, `config_reference_test.exs` and `force_ssl_probe_test.exs` 87/0. Worktree clean at `ea609fbb`.

jhgaylor:
Pushed `546be6a7`. **Commit 3 is unreviewed**, and stale-review dismissal is off
on this branch, so the approval will keep showing green against code nobody has
read — flagging that up front.

**You were right about the comment, and the fix was not cosmetic.** The broker
children sat at `:100`, after Oban at `:98`, while the comment claimed they sat
after `Fountain.Repo` "which is the only thing they need". Reverse termination
therefore stopped the listener *before* Oban, leaving the same trap one layer
down: a job in the tail of a drain could still meet `:listener_down`. I agree
with your reasoning about `lib/fountain/workers/` — empty today is not a
property, it is a coincidence that nothing pins.

They now sit **third**, straight after `Fountain.Repo` and `Ecto.Migrator`. I
verified your dependency argument rather than taking it, and it holds:
`listener_spec/0` reads three app-env keys; `Managoat.Broker.init/1` builds an
ETS certificate cache, a `:persistent_term` entry and a ThousandIsland child
spec; `RequestLog.init/1` returns `{:ok, %{buffer: [], size: 0, timer: nil}}`
and reaches the repo only on flush. I added two checks you did not name, since
starting *this* early widens what could be missing: `Sessions.lookup/1` is a
`Repo.get_by` plus a decrypt, and `Fountain.Crypto` has no process at all — it
holds DEKs in its caller's state. So the repo and the migration that created
its tables really are the whole dependency set.

The test now pins it: `the listener starts before Oban, so it outlives every
job`, plus `the migrator starts before the broker, so its tables exist`.
Reverting to commit 2's position fails **exactly** the Oban assertion and
nothing else, which is the check I wanted before claiming the move bought
anything. 32 tests, 0 failures; format, `--warnings-as-errors` and
`credo --strict` clean.

Took the second note too: `start/2` and `broker_children/0` each carry a
comment naming the other, spelling out that a divergence gives you a working
proxy that writes no `broker_requests` row, with `/api/conversations/:id/egress`
going quiet and nothing failing anywhere.

And thank you for closing the drain question with the actual numbers —
`shutdown_timeout: 15_000` per connection child against a 120s grace and a 5s
`preStop`, and the sum unchanged because the broker was terminated first
before this PR anyway. I confirmed the default in
`thousand_island/lib/thousand_island/server_config.ex:40` and the per-connection
application in `connection.ex:32`. That has replaced the open question in the PR
body, with the knob named as `shutdown_timeout` in `listener_spec/0` rather than
a manifest.

Staying a draft, not merging.

BinaryBourbon (approved):
Re-confirming against `546be6a7`. Both findings from the last pass were taken, and the one that moved code is the right move. Nothing new blocking; I ran the two experiments you asked for and one you did not.

## Booted it, on a warm database and a cold one

`Application.ensure_all_started(:fountain)` with `:broker_listen_port` set to 0, then `Supervisor.which_children/1` reversed — the real runtime order, not a reading of the source:

```
 1. FountainWeb.Telemetry          11. Fountain.Conversations.Redaction
 2. Fountain.Repo                  12. Fountain.FeatureFlags.Cache
 3. Ecto.Migrator                  13. Fountain.Analytics.Sink
 4. Fountain.Broker.Native.RequestLog   14. Fountain.Team.Comms.Inbound.Seen
 5. Managoat.Broker                15. Oban
 6. DNSCluster                     16. Fountain.ConversationRegistry
 7. Phoenix.PubSub.Supervisor      17. Fountain.ConversationSupervisor
 8. Fountain.TaskSupervisor        18. Fountain.RunnerRegistry
 9. Fountain.PlatformChatGPT.Refresher  19. FountainWeb.Endpoint
10. FountainWeb.Plugs.RateLimit.Sweeper
```

`:ok`, `Managoat.Broker.running?/0` true, `Fountain.Broker.preflight/0` `:ok`. The tree is what `children/0` claims, Oban is at 15 with the listener at 5, and reverse termination therefore stops Oban well before the listener.

## The migration dependency — proved, not assumed

Two independent ways, because you were right that this is the one that crash-loops rather than failing CI.

**Structurally:** `Ecto.Migrator` is a `GenServer` whose `init/1` runs the migrations **synchronously** and then returns `:ignore` (`deps/ecto_sql/lib/ecto/migrator.ex:509-523`). `GenServer.start_link` blocks until `init/1` returns, and a supervisor does not start child N+1 until child N's `start_link` has returned. So every migration is committed before `RequestLog` and the listener start. The failure path is the same shape: `{:ok, _, _} = with_repo(...)` raises inside `init/1`, `start_link` returns an error, the whole supervisor start fails and the node never listens — which is exactly what `Fountain.Health`'s moduledoc has always claimed about migrations, still true with the broker at 3.

**Empirically, on a genuinely cold database.** Dropped and recreated a fresh database, pointed the repo at it, booted:

```
cold database fountain_cold_1726_rev created, zero tables
ensure_all_started on a COLD db: :ok in 256.6 ms
broker running?: true
Sessions.lookup on a fresh token: :error
broker_sessions: exists, 0 rows
broker_requests: exists, 0 rows
```

`Sessions.lookup/1` has no rescue, so a missing `broker_sessions` would have raised `undefined_table` rather than returning `:error`. It returned `:error`, on a database that had no tables at all moments earlier. The store is usable the instant the listener is.

## The risk commit 3 adds that commit 2 did not — checked

Starting third means the listener binds while `Phoenix.PubSub`, `Task.Supervisor`, `Oban`, `FeatureFlags.Cache` and `Analytics.Sink` are still starting. That is not only a startup-dependency question, because **session rows are shared across replicas**: a sandbox already running against another pod can land a CONNECT on this listener the moment it binds, mid-boot. So I traced the *live request* path, not just the start path.

`Managoat.Broker.Proxy` → `Sessions.lookup/1` → `Repo.get_by` plus `:telemetry.execute` plus `decrypt/1`; `Fountain.Crypto.load_tenant_key/1` is a `Repo.get_by` on `UserDataKey` and an AES unwrap under the app-env master key (`crypto.ex:100-112`) — no process, as you said. Completion fires `[:managoat, :broker, :request]` into the handler `start/2` attached before any child existed, which is a `Logger.info` plus a cast to `RequestLog`, one position earlier. `Fountain.Conversations.Redaction` is a supervised GenServer called explicitly by conversation code, not a `:logger` filter (nothing in `config/` installs it as one), so the broker's log line does not reach it. Nothing on that path touches PubSub, Oban, Horde, the task supervisor or the endpoint. The repo and its migrations really are the whole set, at request time as well as at start time.

## Revert check — reproduced exactly

`broker_children()` back at commit 2's position, after Oban:

```
1) test children/0 with the broker configured the listener starts before Oban, so it outlives every job
9 tests, 1 failure
```

Exactly the Oban assertion and nothing else, including the new migrator assertion, which correctly still passes there. Restored; `git status` clean.

## The two findings from last pass

Both addressed properly rather than papered over. The `broker_children/0` comment now describes where the code actually is and enumerates the dependency set it rests on, so prose and list agree. The two `backend() == :native` reads now point at each other and each names the cost of divergence — a listener proxying correctly while `broker_requests` fills with nothing and `/api/conversations/:id/egress` goes quiet, which is the right thing to have written down, because it is a failure with no symptom. The new test's framing note ("each *starts before X* is really *outlives X*") is the sentence that makes the whole file legible.

## Gates run here

`mix format --check-formatted` and `MIX_ENV=test mix compile --warnings-as-errors --force`, both exit 0 from `apps/fountain`. `mix credo --strict` on `application.ex`: no issues. Tests: the ordering, health, broker and force-SSL files together, **88 tests, 0 failures**. Worktree clean at `546be6a7`.

Nothing further from me. The ordering argument is now written down where the next person will find it, and the test holds all four ends of it.

BinaryBourbon (commented):
Reviewed `546be6a7e73ae7626b75a638ad541a5b60f5f1fc`.

No actionable findings. Checked broker startup/shutdown order against its dependencies, the configured-off path, bounded preflight behavior, readiness response and tests. The listener now starts before and outlives the endpoint/conversation supervisor.

Validation: static review of application children, health checks and the pinned broker implementation; rollout and HTTP integration tests were not rerun.

BinaryBourbon (changes requested):
**Changes requested** at `546be6a7e73ae7626b75a638ad541a5b60f5f1fc`.

**[P2] Regenerate the checked-in TypeScript OpenAPI output** (`apps/fountain/lib/fountain_web/schemas.ex:2421`). The new readiness example adds `broker_listener`, but `sdk/typescript/src/generated/openapi.ts` still has the old `ReadinessResponse.checks` example containing only `database`. The current-head [SDK contract job](https://github.com/managoat/fountain/actions/runs/34586751630/job/103222953675) regenerates the file, shows the missing `broker_listener` line, and fails with `sdk/typescript/src/generated/openapi.ts is stale`; consequently `CI required` fails too.

Run `scripts/sdk-contract/build.sh`, then `npm run generate` in `sdk/typescript`, and commit the generated result. This is a deterministic consequence of the schema change, not an infrastructure failure.

Validation: independently fetched current head/CI metadata and failed logs, and compared the exact-head schema with the checked-in generated example. Re-reviewed listener dependency/start/stop ordering and health behavior; no additional code defect found there. No boot, provider or production operations were run. Preserve the owner's draft/queue hold.

BinaryBourbon (changes requested):
Burndown verdict at `546be6a7e73ae7626b75a638ad541a5b60f5f1fc`: changes requested; the previous P2 remains unresolved.

**[P2] Commit the regenerated TypeScript OpenAPI output.** The changed readiness example in `apps/fountain/lib/fountain_web/schemas.ex` includes `broker_listener`, while `sdk/typescript/src/generated/openapi.ts` still contains only `database`. The [current-head contract job](https://github.com/managoat/fountain/actions/runs/34586751630/job/103222953675) reports that exact generated diff and fails with `openapi.ts is stale`; `CI required` is also red. Run `scripts/sdk-contract/build.sh`, then `npm run generate` from `sdk/typescript`, and commit the result after updating the branch.

I independently rechecked listener dependencies and the start/stop ordering, configured-off readiness behavior, public response shape and regression tests. No additional runtime defect found. This was source/test and CI-state review; boot, rollout and HTTP integration tests were not rerun.

There is also a current-main conflict in `application.ex` (confirmed with isolated `git merge-tree` against `bf8f890a5408119b97247aebfca6169d9989528f`). Preserve main's `execution_deadline_children()` and `ExecutionTransportSupervisor` while retaining this PR's early broker children and final endpoint. Then rerun application-order/readiness tests, generation checks and required CI, and review the new head. No branch or draft state changed by this review.

BinaryBourbon (approved):
Approve `2fa5cb5ce4bb7bd635a500777b48c9951a254af0` after independent review of the broker startup/readiness repair and current-main integration.

The supervision order now starts the native request log/listener after Repo and migrations and before their consumers, including Oban, execution transports and the deadline worker. Endpoint remains last, so reverse shutdown keeps brokerage available while consumers drain. The telemetry handler is attached during application startup and the children-list function stays free of that side effect. The integration retains main's execution-deadline children and transport supervisor.

The generated TypeScript OpenAPI example includes broker_listener and matches the readiness schema. Its SDK diff is documentation-only, so the documented sdk-no-release classification is appropriate. No actionable source defect remains.

Validation attribution: I reviewed the complete source, ordering/readiness regressions, merge changes and generated contract diff. The coder reports 29 focused tests and contract verification passing, then a completed full precommit at this exact head: compile, formatting, strict Credo, Dialyzer, Sobelow, production release assembly, 6 doctests + 5,615 core tests, 144 Buzz and 33 Support tests, all zero failures. I did not independently rerun the full suite. Earlier SSE/native-docs timing failures and the unchanged controlled rerun are retained in the PR's validation evidence. Required current-head and merge-group CI must pass before landing.

Issue #1919, opened 2026-09-11
No unique index on sandboxes.sprite_name, so two live rows can still claim one machine within an account


jhgaylor:
Follow-up to #1632, which closed the cross-account half of that gap.

## What is left

`sprite_name` is now scoped to the account: every name carries
`fountain-<account>-`, so a name minted in one account cannot resolve to a
machine in another. What scoping does not give is uniqueness *within* an
account. Two live sandbox rows in one account can still carry the same
`sprite_name` and therefore point at one machine, because:

- there is no unique index on `sandboxes.sprite_name` or on
  `(provider, sprite_name)` (see the initial schema migration; the only index
  on the table is `[:status]`, plus the partial home-identity one);
- the Sprites adapter adopts on 409 rather than failing, so a second create
  with the same name silently lands on the existing machine.

The blast radius is one tenant's own data rather than a cross-tenant leak,
which is why it was not fixed alongside #1632. It still breaks ADR 0045's
machine-isolation claim from inside an account, and it makes "which row owns
this machine" unanswerable from the database.

## Why the index was not added in #1632

A migration that adds a unique index **fails at deploy time** if production
already holds duplicates. Nothing in this repo can tell whether it does. The
check that is owed before writing the migration:

```sql
SELECT provider, sprite_name, count(*), array_agg(id)
FROM sandboxes
WHERE status NOT IN ('terminated', 'failed')
GROUP BY provider, sprite_name
HAVING count(*) > 1;
```

If that returns nothing, a partial unique index scoped to live rows
(`WHERE status NOT IN ('terminated','failed')`) is the right shape — terminated
rows are history and a name may legitimately be reused after a machine is
gone. If it returns rows, they need resolving first, and the answer to "how did
they get there" belongs in this issue.

## Also still open from #1632

`Provisioning.create_sandbox_handle/2` calls `Managoat.Sandbox.create/2` with
the bare name and no tenant metadata, so the adapter's adopt-on-409 comment
("Fountain minted this one") remains an assumption rather than a check. That
fix lives in `managoat_sandbox` on hex, not here: give `create/2` tenant
metadata and have the adopt path refuse a sprite whose metadata names another
tenant. With account-scoped names it is now defence in depth rather than the
load-bearing control.

jhgaylor:
Two additions from the review of #1920.

## A second pre-deploy check: names outside their own namespace

The duplicate check above is not the only thing a migration has to survive.
[...]

PR #1920, opened 2026-09-11
fix(conversations): scope a caller-supplied sprite_name to the account (#1632)

### LINKED ISSUE #1632: A caller-supplied sprite_name is not tenant-scoped, so two sandbox rows can resolve to one machine
`POST /api/conversations` lets the caller name the machine, and nothing ties that name to the tenant. Machine identity and Fountain's idea of a sandbox can therefore come apart.

Found while reviewing #1628 / ADR 0045, which relies on the opposite being true. It is not caused by that PR — every part of this predates it.

## The mechanism

Three facts that are each reasonable alone:

1. `sprite_name` is a public property of `ConversationCreateRequest` — "Override the auto-generated sprite name" (`apps/fountain/lib/fountain_web/schemas.ex:505`). `FountainWeb.ConversationController.create/2` passes `params` through untouched.

2. `Conversations.mint_sprite_name/3` returns a caller-supplied name verbatim:

   ```elixir
   defp mint_sprite_name(_provider, _user_id, name) when is_binary(name), do: {:ok, name}

   defp mint_sprite_name(_provider, user_id, nil),
     do: {:ok, "fountain-#{tenant_prefix(user_id)}-#{short_id()}"}
   ```

   The generated branch carries a tenant prefix and a random suffix. The override branch carries neither, has no ownership check, and there is no unique index on `sandboxes.sprite_name`.

3. `Provisioning.create_sandbox_handle/2` calls `Managoat.Sandbox.create(provider, sandbox.sprite_name)` with the bare name and no tenant metadata, and the Sprites adapter treats a name collision as its own retry:

   ```elixir
   # The sprite already exists — adopt it. Names are unique per token and
   # Fountain minted this one, so "already there" means an earlier attempt
   # (or a retry after a lost response) won the race.
   {:error, {:api_error, 409, _body}} ->
     {:ok, wrap(Sprites.sprite(client, name))}
   ```

   "Fountain minted this one" is the assumption fact 2 breaks. Sprites names are unique per deployment token, not per tenant.

## Why it matters

Two sandbox rows — including two belonging to different accounts — can end up pointing at one machine, with no row, index or check recording that they do. A machine holds a tenant's decrypted environment and vault values on disk and its agent's processes in memory, so co-tenancy is not a cosmetic mix-up.

It also breaks a guarantee we are now writing down. ADR 0045's `check_sandbox_api_attach/2` refuses to put a credentialed conversation on a credential-free worker's machine by asking whether any conversation row points at `sandbox.id`. That question cannot see a second row that named the same sprite, so the guarantee is over Fountain's rows rather than over machines.

Practical reach is limited by having to name a sprite that already exists: a generated name ends in a random suffix, so the realistic collisions are between names a caller chose. That bounds it; it does not make the invariant true.

## Suggested fixes

Any of these closes the row-versus-machine gap; the first is the one that makes the invariant true everywhere:

- Scope the override the way the generated branch already does — keep `fountain-<tenant_prefix>-` and let the caller supply only the suffix. A name is then unusable across accounts by construction, and existing rows keep working.
- Add a unique index on `sandboxes.sprite_name` (or on `(provider, sprite_name)`) so two live rows cannot claim one machine even within an account.
- Pass tenant metadata to `Managoat.Sandbox.create/2` and have the adopt-on-409 path refuse a sprite whose metadata names another tenant, so the adapter's stated assumption is checked rather than assumed.
- Narrowly, and worth doing regardless: refuse a caller-supplied `sprite_name` together with `sandbox_api_access: "none"`, so ADR 0045's machine-isolation claim holds as written.

## Deciding what to keep

Worth asking first whether `sprite_name` needs to be a public request field at all. It is documented only as an override of a generated name; if its callers are internal or operational, removing it from `ConversationCreateRequest` is smaller than any of the above.

jhgaylor:
Caller-supplied `sprite_name` values previously reached the provider verbatim. The Sprites adapter adopts an existing name on conflict, so a caller could create a second sandbox row pointing at another account’s machine. This change constrains hosted names to the account’s namespace and rejects overrides that cannot preserve runner placement or a fresh `none` sandbox.

### Resulting behavior

- Hosted names use `fountain-<first-eight-account-id-characters>-<suffix>`. A supplied value becomes the suffix; a full name already carrying the caller’s prefix is accepted without adding the prefix twice. Suffixes must match `[A-Za-z0-9][A-Za-z0-9_-]{0,39}` or return `422 invalid_sprite_name`.
- A nonempty override on a self-hosted runner returns `422 sprite_name_not_supported`. Runner names encode placement, so the server must mint them through the caller’s runner selection. Missing or empty overrides retain normal placement.
- A nonempty override with `sandbox_api_access: "none"` returns `422 invalid_sandbox_api_access`. Missing or empty names retain normal name generation.
- The field remains available in all four SDKs. The schema description and generated TypeScript documentation explain the new semantics; the request shape is unchanged. `sdk-no-release` applies because the generated SDK diff only changes this JSDoc description, with no type, wire contract or runtime change.

### Compatibility and remaining limits

Existing sandbox rows are not renamed or migrated. Repeating an arbitrary legacy value such as `my-box` now names a different machine, `fountain-<account>-my-box`. To return to an existing machine, use its `sandbox_id` while that row remains attachable: ready or suspended, without a pending reset, and compatible with the requested agent, vault, environment and runtime. Previously generated hosted names carrying the caller’s prefix continue to round-trip. Runner overrides are explicitly refused.

This is a scoped mitigation, not a complete machine-ownership guarantee. **#1919 remains necessary:** two accounts can share the truncated prefix; legacy names may already occupy another account’s namespace; and duplicate rows within an account can still name one machine. In particular, an `owner` launch can name the generated machine of an existing `none` conversation, bypassing isolation through a second row. Rejecting `none` plus an override does not prevent that reverse alias.

#1919 tracks production duplicate/name-ownership checks before a uniqueness migration and provider-side ownership validation on adoption. This PR adds neither the unique index nor provider metadata verification.

### Validation

Validated head: `92b17105d79a102ebe6b94e67c56a00ff5d34d46`. The merge preserves main’s queue-poke helper, both fallback tests and both changelog entries. A final merge-tree check against main `03af4683` was clean.

- Compile with warnings as errors, unused-dependency check, format, strict Credo, Dialyzer, Sobelow and production release assembly passed.
- Focused naming, runner placement and API tests: **182 tests, 0 failures**.
- Final full umbrella run: **6 doctests and 5,608 core tests, 144 Buzz tests, 33 Support tests; zero failures**. Existing 2 skips and 7 exclusions are unchanged.

The initial full-suite attempt was stopped after an inherited Ecto preload checkout timeout matching #1524 and a 300 ms provisioning-deadline race tracked in #2040. Its evidence was retained. Unrelated host overload was identified and corrected during that attempt. Both unchanged failing modules then passed **30 tests, 0 failures**. The final full run used a fresh dedicated database, the same seed `527730`, `ERL_FLAGS='+S 4:4'`, and `--max-cases 4`. No assertions or project timeouts changed; CI retains its normal configuration and is required before merge.

Closes #1632.


BinaryBourbon (approved):
Reviewed the whole diff against the worktree, reran the gates, and reproduced both revert checks. **Approving** — the invariant holds where it is claimed to, the gates are real, and the deferral of the index is the right call. Four things below are worth an edit; none of them block.

## What I verified, and how

**The scoping actually holds.** Every name that comes out of `mint_sprite_name/3` is now `prefix <> suffix` with `prefix = sprite_name_prefix(user_id)` (`conversations.ex:3544`), so two accounts' names can only be equal if their 8-character tenant prefixes are equal. There is no path around it: `attrs["sprite_name"]` reaches exactly one mint call (`conversations.ex:2538`), the respawn call passes `nil` (`conversations.ex:4162`), and the attach clause (`conversations.ex:2508`) never mints at all.

**Nothing looks a sandbox up by name.** This was the backward-compat case I most expected to break. I grepped `sprite_name` across `apps/` and `ee/`: every read is off a row that was already fetched — `Managoat.Sandbox.build_handle(provider, sandbox.sprite_name)` in provisioning, the reaper and `home_checkpoint`, and the reattach identity compare in `sandbox_identity.ex:23,27,99`. Attach is by `sandbox_id`; persistent reuse goes through `home_or_new/5` on the agent/env/vault identity, not on a name. So nothing recomputes a name to find a row, and existing rows keep resolving to their machines.

**`mint_sprite_name(:runner, …)` is untouched.** It is still the first clause and still the only one that calls `Runners.mint_sandbox_name/1`, reached from both call sites with `nil`. Clause order is safe — the `""` literal sits before the general `is_binary` clause and `mix compile --warnings-as-errors` is clean, so no clause is shadowed.

**The 422s go through the established shape.** Nothing between `fallback_controller.ex:5` and the new clause at `:433` matches a bare `:invalid_sprite_name` (the two guarded clauses at `:378` and `:553` are `reason in [...]` lists that do not name it), so it lands on its own clause rather than the terminal safety net at `:577`. The body is `error` + `message`, which is what `Schemas.UnprocessableEntityError` declares (`required: [:error]`, optional `message`), and `operation(:create, …)` already declares `unprocessable_entity` — no new status code, as you say. I also checked the non-string case out of curiosity: `"sprite_name": 123` is caught upstream and returns `422 validation_failed` with `errors`, not a `FunctionClauseError`.

**Gates**, all from the worktree with the mise toolchain:

- `mix format --check-formatted` from `apps/fountain` — ok
- `mix compile --warnings-as-errors` — ok, no warnings
- `mix credo --strict` on the three changed lib files (from the umbrella root, where credo lives) — no issues
- `scripts/sdk-contract/build.sh --check` — `ok (164 operations, 190 schemas)`, tree still clean afterwards
- regenerated the TS types from the freshly built `dist/openapi.json` into a scratch file and diffed against `sdk/typescript/src/generated/openapi.ts` — **byte-identical**, so the committed one-line change is exactly what the generator produces
- `conversations_start_test.exs` + `runners/placement_test.exs` + `sandbox_api_access_test.exs` — 68 tests, 0 failures
- `conversation_controller_test.exs` + the four schema/spec guard files — 108 tests, 0 failures

**Revert checks reproduce exactly.** Scoping clause back to `{:ok, name}`: 4 failures, the same four you name. Removing the `check_sandbox_api_name("none", …)` clause: 2 failures, the context refusal and the API 422. Restored and confirmed `git status` clean.

## Findings

**1. The round-trip claim is false on the runner provider (`conversations.ex:3535`).** "A name read off an earlier launch can be handed straight back" does not hold for a runner-backed sandbox. `Managoat.Runner.Names.for_runner/1` mints `runner-<32 hex>-<8 hex>` — 48 characters — and `@sprite_name_suffix` caps the suffix at 40, so handing that name back is now a 422. I checked it end to end rather than by counting:

```
RUNNERNAME: {48, "runner-e491dcd5d66c4163a8a8dc08fae4aab6-9767d48f"}
RUNNER_ROUNDTRIP: {422, "{\"error\":\"invalid_sprite_name\",\"message\":\"sprite_name is the suffix of …\"}"}
```

Before this change that one override *worked*: the name was taken verbatim and `Runners.parse_sandbox_name/1` could read the runner back out of it. So it is not only the pre-existing "an override produces a name placement cannot read" case you describe in the PR body — it is the one override on that path that used to resolve, and it now refuses. A loud 422 is better than the silent mis-placement every other override on that path gets, so I am not asking you to restore it. I am asking that the claim not be stated without the exception: either scope the sentence in the PR body and the schema description to the platform providers, or refuse an override on `:runner` explicitly (the way `none` is refused) so the 422 is a decision rather than a length accident.

**2. "can never resolve to another account's machine" is stronger than the mechanism (`schemas.ex:593`).** The namespace is `binary_part(user_id, 0, 8)` — 32 bits of a v4 UUID. Two accounts sharing a tenant prefix share a namespace, and then a shared suffix is a shared machine. I do not think this weakens the fix in practice and it is not a reason to hold the PR: an attacker cannot choose their own `user_id`, so mounting it deliberately means registering until you hit a given prefix, and a natural collision needs the prefix *and* a caller-chosen suffix to match. But "never" is a claim a reader will rely on, and the honest version is "a name minted for one account cannot resolve to another account's machine unless the two accounts share their 8-character prefix". Worth one clause in the description, and worth a sentence in #1919, where the partial unique index is what would actually make it "never".

**3. Legacy names can already sit inside another account's namespace — worth a second production check in #1919.** Scoping constrains names minted from now on. A row minted under the old verbatim rules holds whatever string was passed, including a string shaped like `fountain-<some other account's prefix>-x`. If such a row exists, the account that owns that prefix can still reach that machine by supplying `x`. Contrived, and #1919's index closes it, but #1919 currently only asks production for duplicates. The companion query is names that are not in their own account's namespace:

```sql
SELECT id, user_id, provider, sprite_name FROM sandboxes
WHERE status NOT IN ('terminated','failed')
  AND provider <> 'runner'
  AND sprite_name NOT LIKE 'fountain-' || left(user_id::text, 8) || '-%';
```

**4. The new fallback clause is not covered (`fallback_controller.ex:433`).** Deleting it fails no test: the terminal safety net at `:577` still returns 422 with `error: "invalid_sprite_name"`, just without the `message` and with a `fallback: unmapped error atom` warning in the log. The context-level refusal is tested (`conversations_start_test.exs:130`) and the `none` refusal is tested at the HTTP level, but the shape of this one is not. One API-level assertion on the 422 body next to the `none` test would lock it — it does work; my throwaway probe above is the evidence, and that evidence should live in the suite rather than in this comment.

**Nit:** this changes the meaning of a public, SDK-exposed request field, which is a `### Changed` entry under `[Unreleased]` in `CHANGELOG.md`. The convention here is loose (2 of the last 25 commits touched it), so take it or leave it — but this is precisely the class of change that goes missing when a release is rolled.

## On deferring the index

Agreed, and for the reason you give: a migration that adds a unique index fails at deploy if production already holds duplicates, nothing in the repo can answer whether it does, and a failed migration is a worse outcome than a known follow-up. Scoping is sufficient for the cross-tenant case #1632 is actually about — for every name minted from now on. #1919 is what closes the within-account half and the legacy residual in finding 3, so it should not sit for long.

I did not run the full suite, sobelow, dialyzer or the prod release assemble; CI covers those.

jhgaylor:
Pushed `de6a7336` with the two code findings fixed, plus the wording and the CHANGELOG nit. Thanks for the runner catch — that one was a real regression and I had it filed under the wrong heading.

**Note: I had to dequeue this PR to push.** It was at position 4 in the merge queue, and a protected-branch hook refuses a push to a queued branch. It is out of the queue now and needs re-queueing once you are happy with this commit.

### 1. The runner provider — explicit refusal

You are right that this was a length accident. `Managoat.Runner.Names.for_runner/1` mints 48 characters (`runner-<32 hex>-<8 hex>`) against a 40-character suffix bound.

I went with the explicit refusal rather than widening the bound, and widening turns out not to be an option at all: a runner name carries no account prefix, so a wider bound would prefix it to `fountain-<account>-runner-…`, and `Runners.parse_sandbox_name/1` cannot read the runner out of that. Widening would convert a 422 into a silently unlocatable sandbox. On this provider the name *is* the placement (ADR 0022), so there is nothing to honor.

`mint_sprite_name(:runner, _user_id, name) when is_binary(name)` now returns `{:error, :sprite_name_not_supported}` → 422, with a comment saying why. Two tests in `placement_test.exs`: the refusal, and the round-trip case specifically (mint a runner sandbox, hand its own name back, assert the refusal).

I have scoped the round-trip claim in the PR body to the hosted providers.

### 2. The fallback clause now has a test — and so does the new one

Correct, and it is the sharper version of the problem than it looks: the terminal safety net answers 422 with the bare atom, so a test asserting only the status and `error` would also have passed with no clause. Both tests assert the `message`.

- `invalid_sprite_name`: end-to-end through `POST /api/conversations` in `conversation_controller_test.exs`.
- `sprite_name_not_supported`: driven directly in `fallback_controller_test.exs`, following the `sandbox_reset_pending` precedent already in that file. Reaching it through the route needs an agent on the runner provider, and `runners_enabled` is global application env that an `async: true` module must not write (#1214) — `placement_test.exs` is `async: false` for exactly that reason.

Verified by deleting each clause and watching only its own test fail (2 failures for the runner refusal, 1 each for the two fallback clauses), then restoring.

### 3. Wording softened

Fair. A caller cannot choose their own `user_id`, so it is not adversarially reachable, but a 32-bit prefix is not "never". Both the schema description and the code comment now say the name lands in the caller's own namespace, and name #1919 as where "never" becomes true.

### 4. #1919 updated

Added your namespace check alongside the duplicate check. Also flagged there that ADR 0051 (`provider_instance_id`, "the same provider and ID cannot belong to two retained rows") landed on main while this was open and partly overtakes the index idea — it is uniqueness over what the provider handed back rather than over what Fountain asked for, though it is optional and recorded after the fact, so it does not close the issue.

### Nit taken

CHANGELOG entry added under Unreleased → Changed, covering the suffix semantics, the two refusals and the fact that existing rows are untouched.

### Re-verified after the changes

`mix compile --warnings-as-errors`, `mix format --check-formatted` (from `apps/fountain`), `mix credo --strict` on the three lib files, `scripts/sdk-contract/build.sh --check` ok, `npm run generate` → the single description line, `python3 scripts/docs-style.py` clean on CHANGELOG (its two findings are pre-existing pages I did not touch), `docs_test.exs` 29/0, and 233 tests / 0 failures across the conversations, placement, sandbox-api-access, conversation-controller, fallback-controller and schema-guard files.

jhgaylor:

BinaryBourbon (approved):
Re-reviewed at `de6a7336`. **Re-confirming the approval.** Commit 2 answers all four things I raised, and the runner refusal turned out to be more than a wording fix — see the ruling below. One finding, on the CHANGELOG wording, which is the only place the upgrade story is now told wrong.

## Ruling: 422 is right, and no deprecation path is owed

You asked me to rule on whether refusing beats honouring, since this is a documented SDK-exposed field on a provider where it previously worked. I went and checked what "worked" meant, rather than assuming it.

**It never worked.** `Managoat.Runner.Adapter.create/2` routes through `rpc/3` → `runner_id(name)` → `Names.parse(name)` (`deps/managoat_runner/lib/managoat/runner/adapter.ex:314,322`). A name that is not runner-shaped returns `{:error, {:invalid, {:not_a_runner_sandbox_name, name}}}`, so provisioning cannot place the sandbox at all. I confirmed it from the other side: with the refusal clause deleted, `"pinned-name"` on a runner agent returns `{:ok, conv}` and writes a row named `fountain-<account>-pinned-name` on provider `runner` — a row nothing can provision and `Runners.parse_sandbox_name/1` cannot read. So the behaviour being removed produced a broken conversation, not a working one. There is nothing to deprecate.

**And the one shape that did route was a cross-tenant path.** A runner-*shaped* name parses, and the layer it routes through has no tenant concept: `Connection.call/3` is `Config.host!().whereis(runner_id)` (`connection.ex:80`), a registry lookup by runner id with no user argument, reached from `Provisioning.create_sandbox_handle/2`, which passes the bare name. Before this PR, an account naming another account's runner sandbox would have had its create routed to that account's runner. Commit 1 did not close that — a runner name carries no account prefix, so the tenant prefix cannot constrain it; it was closed by the 48-versus-40 length accident, exactly as your commit message says. **Commit 2 is what makes that a decision rather than an accident, so I read it as the second half of the security fix, not as polish.** That is worth saying in the PR body: it is a stronger claim than "the override was never useful here".

## Clause ordering

Correct, and I checked the two orderings that matter rather than only the one you named.

- `mint_sprite_name(:runner, user_id, nil)` (`conversations.ex:3531`) is still first and `nil` is not a binary, so the new clause at `:3547` cannot shadow it from either direction. Verified live, not only by reading: the round-trip test starts a runner conversation with no name and gets `{:ok, conv}` with a minted `runner-<32 hex>-<8 hex>`, which is the thing that would break if the refusal had swallowed the mint.
- The subtler one: the `""` clause at `:3538` sits **before** the refusal, so `sprite_name: ""` on a runner agent is still "no override" and mints normally rather than 422-ing. That matches how `""` is treated for `sandbox_mode` and in `check_sandbox_api_name/2`, so it is the right answer — but it is load-bearing on clause order and is the one edge with no test. Not asking for one; noting it so the next person who reorders these knows.

## Both new clauses bite

Deleted each and ran `fallback_controller_test.exs` + `conversation_controller_test.exs` + `placement_test.exs` (99 tests) each time:

- delete `{:error, :invalid_sprite_name}` (`fallback_controller.ex:433`) → **1 failure**, `ConversationControllerTest` "returns 422 invalid_sprite_name, with a message, for a malformed suffix". Only its own.
- delete `{:error, :sprite_name_not_supported}` (`fallback_controller.ex:444`) → **1 failure**, `FallbackControllerTest` "a sprite_name on the runner provider is refused with a reason". Only its own.
- delete the `mint_sprite_name(:runner, _user_id, name)` clause → **2 failures**, both new placement tests, and the diff output is itself the evidence for the ruling above: the round-trip one falls back to `invalid_sprite_name` (the length accident), and the other returns `{:ok, %Conversation{}}` (the unplaceable row).

Agreed on driving the runner clause directly rather than through the route — `runners_enabled` is global application env and `async: true` must not write it (#1214). The `message` assertion is the right thing to assert in both, for the reason you give.

## Gates, re-run at commit 2

`mix format --check-formatted` and `mix compile --warnings-as-errors` from `apps/fountain` — clean, no unreachable-clause warning. `mix credo --strict` on the three changed lib files — no issues. `scripts/sdk-contract/build.sh --check` — ok (164 operations, 190 schemas). Regenerated the TS types from the rebuilt `dist/openapi.json` into a scratch file and diffed against the committed `sdk/typescript/src/generated/openapi.ts` — byte-identical, so the second description change is again exactly what the generator produces. Tests: the nine relevant files, **189 tests, 0 failures**. `docs_test.exs` — 29 tests, 0 failures (CHANGELOG.md is an `@external_resource` and `docs/changelog.md` snippet-includes it, so it is published; the prose gates walk `docs/**`, so the new entry's own wording is not linted).

## One finding: the CHANGELOG understates the break

`CHANGELOG.md:34` — "callers that passed a full name get the same machine under an account-scoped name" is the one sentence an upgrading operator will read, and on the natural reading it is false. A caller who passed `my-box` before this release has a row literally named `my-box`; passing `my-box` after it mints `fountain-<account>-my-box`, which is a **different machine**. The old row is untouched, as you say, but it is no longer reachable by the name that created it, and the caller is not told — they get a fresh machine and a 201.

I think you mean "the same machine across launches" (passing `my-box` repeatedly is still stable), which is true. But this is the entry's only upgrade-relevant fact and it currently reads as reassurance that nothing moves. Suggest something like: *a name used before this release now resolves to a new machine in the account's namespace; the machine it named before is untouched but is no longer reachable by that name.* Same change to the PR body, which carries the same claim.

## ADR 0051 and the deferred index

It does not change my view; it sharpens it in one place.

0051's uniqueness is over `provider_instance_id` — what the provider handed back — and it is recorded *after* creation by a trusted caller from provider control metadata. So when a second row adopts an existing machine, what fails is the identity **binding**, not the **create**: the conversation has already landed on that machine. That is detection, not prevention, which is why it does not close #1919 — you have that right. It is, from the other side, the "check the adapter's assumption" bullet the original issue asked for, so it is real defence in depth; it is also still **Proposed**, with the worker integration that would populate the column in separate work (#1754), so today it constrains a column nothing writes.

The one thing it changes: 0051 shows the shape that makes a uniqueness migration deploy-safe — existing rows get NULL and are therefore unconstrained. #1919 cannot borrow that, because a unique index on `sprite_name` constrains the existing rows by definition, which is exactly why your deferral was right. But once #1754 populates `provider_instance_id`, the production question #1919 owes is better asked of that column than of `sprite_name`: duplicates over the machine the provider actually gave back, rather than over the name Fountain asked for. Worth a line in #1919 alongside the two SQL checks.

Not merging, not marking ready. Worktree restored and clean at `de6a7336` after the four delete-and-check experiments.

jhgaylor:
Pushed `6e4aa5d3`. **Heads-up for the reviewer: stale-review dismissal is off on this branch, so the approval on `de6a7336` will keep showing green against this commit — it has not been reviewed.** It is comment, CHANGELOG and test only; no behaviour change from `de6a7336`.

### 1. The routing hole — claimed, and scoped to the right commit

I confirmed both call sites: `Adapter.rpc/3` (`adapter.ex:322`) reads the runner id out of the name and hands it to `Connection.call/3`, which is `Config.host!().whereis(runner_id)` (`connection.ex:80`) with no tenant argument. So a `sprite_name` shaped like another account's runner sandbox routed the launch to their runner. That is a bigger finding than the length accident and it is now in the PR body, the CHANGELOG and the code comment.

I did **not** adopt "the prefix scoping in commit 1 cannot close that", because it is not what the code does. Commit 1 prefixes a runner-shaped name to `fountain-<account>-runner-…`, `Names.parse` rejects it, and nothing routes — so the route is closed by commit 1. What commit 1 leaves is a 201 over a row nothing can place, which is precisely what deleting my refusal clause produces. Commit 2 makes that an explicit 422. All three places now say it that way.

Worth stating plainly given what this batch was about: the difference between "cannot close that" and "closes it into a 201 nobody can place" is the same class of error as the four prose findings here, so I would rather be pedantic about it than inherit the framing.

### 2. `CHANGELOG.md` upgrade note rewritten

You are right, and on re-reading it read as reassurance that nothing moves. It now says what happens: a caller who passed `my-box` has a row named `my-box`, passing `my-box` today provisions `fountain-<account>-my-box` and answers 201, and the old row is untouched and no longer reachable by the name that created it. A server-minted name still round-trips; an arbitrary legacy name does not, and `sandbox_id` attaches by row rather than by name and is the way back. The same paragraph is now in the PR body, and the round-trip claim there is scoped to the hosted providers *and* to server-minted names.

### 3. The `""` edge is tested

Good catch — it was the only branch with no coverage. `placement_test.exs` now pins that `sprite_name: ""` on a runner agent mints `runner-<32 hex>-<8 hex>` and the name still parses back to the registered runner. Verified by swapping the `""` clause below the runner refusal: only that test fails.

### 4. #1919 corrected

Agreed, and I have corrected my earlier comment there. 0051's uniqueness is over `provider_instance_id` recorded after creation, so a second adoption fails the identity binding rather than the create, and the conversation has already landed. Detection, not prevention, and not a reason to defer differently. Added your point that once #1754 populates the column the production duplicate question is better asked of the identity than of the name — over the machine the provider handed back rather than the name Fountain asked for.

### Re-verified at `6e4aa5d3`

`mix compile --warnings-as-errors`; `mix format --check-formatted` from `apps/fountain`; `mix credo --strict` on `conversations.ex` (no issues); `python3 scripts/docs-style.py` clean on CHANGELOG (its two findings are pre-existing pages); **193 tests, 0 failures** across conversations-start, placement, sandbox-api-access, conversation-controller, fallback-controller and docs. No schema change this round, so the contract and TS artifacts are untouched and still byte-identical to `de6a7336`.

Still a draft, still not queued.

BinaryBourbon (approved):
Re-reviewed at `6e4aa5d3`. **Re-confirming the approval.** Final pass from me.

## 1. You are right and I was wrong

My round-2 framing — "commit 1 did not close that; it was closed by the 48-versus-40 length accident" — was wrong about the mechanism, and wrong in the direction that matters: it made commit 1 sound fragile. Commit 1 closes the route **structurally, twice over**:

- Every value the general clause returns is `prefix <> suffix`, and `sprite_name_prefix/1` always begins `fountain-`. `Managoat.Runner.Names.parse/1` matches on a literal `"runner-" <> rest`. So no caller-supplied name can produce a string that parses as a runner name — including the verbatim branch, since a name carrying this account's prefix also begins `fountain-`. That is not an accident, it is the shape of the output.
- Independently: the shortest parseable runner name is 41 characters (`"runner-"` + 32 hex + `"-"` + at least one), and the suffix bound is 40. No parseable runner name can clear the regex either.

So what commit 1 leaves behind is exactly what you say — a 201 over a row nothing can place — and commit 2 is what turns that into a plain 422. The stronger claim is the right one; the code comment at `conversations.ex:3545-3553` and the CHANGELOG now state it correctly. My "second half of the security fix" line from last round overstated commit 2's value and should not be carried into the PR body; what is true is that the **pre-branch** routing hole was real (`Adapter.rpc/3` → `runner_id(name)` → `Connection.call/3` → `whereis(runner_id)` with no tenant argument), and that is what both texts now say.

## 2. The recovery advice holds — verified end to end

I did not take this one on reading, because an operator will follow it. Built a legacy row the way the old rules would have left one (`sprite_name: "my-box"`, no prefix) and ran both halves of the note:

```
A: {"my-box", "fountain-16ddbfd4-my-box", true}     # same value in -> different machine, 201
B: {:attached, true, "my-box"}                       # sandbox_id -> the legacy row, name intact
```

Both claims are accurate. The attach path never touches the name: `attach_conversation/3` fetches with `get_sandbox(sandbox_id, user_id)` (`conversations.ex:3238`) and everything after that is status and identity. Good catch on the first draft — "pass the row's current `sprite_name`" would have been false for precisely the rows the note exists for, and it would have read as authoritative.

**Optional, take it or leave it:** `sandbox_id` carries the standing ADR 0023 preconditions — `check_attachable/4` requires `status in ["ready", "suspended"]`, no pending reset, and the same agent, vault and environment. True of every attach, not of this note, so I would not restate ADR 0023 in a CHANGELOG entry; but an operator whose legacy machine was reaped, or who has since changed the agent's environment, gets `sandbox_identity_mismatch` rather than their machine back. A clause like "if it is still live and the launch declares the same agent, vault and environment" would close that off.

## 3. The `""` test bites

Verified the way the commit message says it was: moved the refusal clause above the `""` clause and reran `placement_test.exs` + `conversations_start_test.exs` — **1 failure**, "an empty sprite_name on the runner provider mints rather than refusing", and nothing else. It also asserts the right thing, `parse_sandbox_name/1` returning the registered runner id rather than just a 201, so it pins placement and not only the status code.

## 4. #1919

Read the update. The distinction you added is the useful half and is better than how I put it: a name index finds rows that **asked** for the same machine, the identity finds rows that **got** one. With adopt-on-409 in the middle, those are different sets, and only the second is the question #1919 actually needs answered.

## Gates at commit 3

Docs and tests only, as billed: `git diff de6a7336 6e4aa5d3 -- sdk/ dist/` is empty, so the contract and TS artifacts are untouched and still the ones I diffed byte-identical last round. `mix format --check-formatted` and `mix compile --warnings-as-errors` clean; `mix credo --strict` on `conversations.ex` no issues; the five conversation/runner/fallback/controller files plus `docs_test.exs` — **193 tests, 0 failures**. `docs_test.exs` passing matters here because `docs/changelog.md` snippet-includes `CHANGELOG.md`, so the new note is published at `/docs`.

Still a draft; not merging, not marking ready. Worktree restored and clean at `6e4aa5d3`.

BinaryBourbon (commented):
Reviewed `b940660c23e4baaad93f030d19fe7be2f4a7dfcb` for stack #1632.

No actionable defect introduced by this PR found. Reviewed account-prefix and suffix validation, runner naming/refusal, the sandbox API-access guard, error responses, schema changes, and regression tests. The within-account duplicate-machine and truncated-prefix risks remain explicitly deferred to #1919.

Validation: static review at the exact head; application/provider tests were not rerun.

BinaryBourbon (approved):
Approve the code at `b940660c23e4baaad93f030d19fe7be2f4a7dfcb`.

Re-reviewed the full diff and earlier review findings: caller-supplied hosted names retain the account prefix, runner overrides are explicitly refused, empty runner names still invoke placement, and the none-access guard and error responses are consistent. The previous runner, response-coverage and upgrade-wording findings are addressed. The final commit changes only the sandbox_id recovery wording relative to `6e4aa5d3`; duplicate-machine hardening remains explicitly deferred to #1919.

Independently exercised the exact-head naming and guard functions extracted from the AST, including cross-account names, suffix validation, runner empty/nil/refusal paths and none-access refusal; checks passed, with runner placement stubbed. Full application/provider tests were not rerun.

GitHub currently reports merge conflicts and no checks on this head. This is a code-review approval, not a claim that the current-main integration is validated: resolve the conflicts and run CI before merging. The owner’s existing draft/queue restriction remains in place.

BinaryBourbon (approved):
Re-reviewed `b940660c23e4baaad93f030d19fe7be2f4a7dfcb`. Approve this scoped security improvement; no new blocking defect found.

The account prefix, suffix validation, runner refusal, empty-name placement, error responses, and legacy-name upgrade note are consistent. Existing arbitrary names now identify a different machine when submitted again; `sandbox_id` is the documented route to an attachable old row. No SDK shape change or release is needed just to retain this field.

Landing work: resolve conflicts against current main in `CHANGELOG.md`, `apps/fountain/lib/fountain/conversations.ex`, and `apps/fountain/test/fountain_web/controllers/fallback_controller_test.exs`. Preserve main's queue-poke helper alongside the new suffix attribute, and preserve both fallback tests. GitHub currently reports a draft, merge conflicts, and no check runs on this head. Fresh CI is required after the update; this approval is not integration validation.

Residual security scope, already tracked by #1919: this does not fully establish ADR 0045 machine isolation. I reproduced a fresh `none` conversation followed by an `owner` launch using its generated full name: the second launch succeeds, with two distinct sandbox rows naming the same Sprites machine. The new guard rejects `none` plus an override but cannot reject that reverse alias. This is pre-existing within-account duplicate-machine behavior explicitly deferred in #1919, not a regression introduced here. Keep #1919 prioritized; do not advertise this PR as complete isolation. Tenant-prefix collisions and legacy names outside their owner's namespace remain relevant to the production audit too.

Validation: the five affected naming/placement/fallback/controller/API-access files passed at the exact head: **164 tests, 0 failures** (159 + 5). Application compilation also succeeded. A separate adversarial integration probe intentionally failed the expected-refusal assertion and demonstrated the documented residual above; provider provisioning was stubbed. Reviewed the adapter's adopt-on-409 code, but did not run a live provider or full CI.

The remaining human choice is accepting the documented legacy-name behavior and scheduling its rollout once current-main CI passes. Production duplicate/name-ownership inspection is needed before a uniqueness migration in #1919; it is not a prerequisite for this scoped fix.

BinaryBourbon (approved):
Approve the tenant-scoped naming repair at `92b17105d79a102ebe6b94e67c56a00ff5d34d46` after independent source and integration review.

Hosted-provider names retain the caller’s tenant prefix; supplied suffixes and previously returned same-tenant names are validated consistently. Runner overrides are refused before placement, and isolated `sandbox_api_access: none` launches cannot select an existing machine name. The API errors, schema/generated contract, legacy attachment behavior and regression cases agree with those rules. The merge preserves main’s queue helper and both fallback-controller tests.

No actionable defect found within this PR’s scope. The documented same-account aliasing / namespace uniqueness follow-up remains #1919; this approval does not claim that follow-up is solved.

Validation attribution: I reviewed the complete final source and test diff, including the conflict resolutions. The coder reports 182 focused tests and all precommit static and production-release gates passing. Complete unchanged-tree rerun results are recorded in the PR body; the initial timing failures and unchanged diagnostic reruns are retained there rather than discarded. Required current-head CI must pass before merge.

PR #1921, opened 2026-09-11
fix(connections): gate creation, not revocation (#1693)

### LINKED ISSUE #1693: Connections is gated by a feature flag at the doors: self-hosts lose it on upgrade, and revocation 404s while brokering continues
## Symptom

Two consequences of #1620's "gate Connections":

1. Every self-hosted deployment without PostHog loses the Connections UI and API on upgrade unless the operator sets `FEATURE_FLAGS_ON=connections`. Documented at `.env.example:199` and `docs/configuration.md:376`, but not as an upgrade step in the CHANGELOG.
2. A tenant with the flag off keeps having connection tokens brokered into sandboxes, while every door to revoke or delete a connection or binding returns 404. They cannot turn off a credential still in use.

## Mechanism

- Gate: `Broker.enabled_for?(user) and FeatureFlags.enabled?(:connections, user)` at `apps/fountain/lib/fountain/connections.ex:56-58`, enforced by controller plugs and LiveView mounts, not in the context.
- `feature_flags.ex:130-136`: `remote_enabled?` returns false for every flag when PostHog is not configured.
- Runtime paths still gate on the broker alone: `egress.ex:139,161,173,271`.
- Doors that 404: `connection_controller.ex:112`, `connections_live/index.ex:92`, `secret_binding_controller.ex:144`, `secret_bindings_live/index.ex:23`.

## Fix

Gate creation only; leave read, revoke and delete open whenever the tenant has rows. Default the flag on when PostHog is not configured, or make the runtime brokering respect the same gate. Add the upgrade note to CHANGELOG.

---

jhgaylor:
Closes #1693.

## The two consequences of #1620

**A self-host loses Connections on upgrade.** `FeatureFlags.remote_enabled?/2` returns false for every flag when PostHog is not configured, so a deployment with no flag service lost the Connections page and the `/api/connections`, `/api/connection-providers` and `/api/secret-bindings` routes on the upgrade that added the flag, unless the operator set `FEATURE_FLAGS_ON=connections`.

**An account with the flag off cannot revoke a credential that is still in use.** The gate stood in front of every door, the ones that take a credential away included, while the runtime paths kept gating on the broker alone and kept brokering those tokens into sandboxes. Revoking a connection, deleting a provider and unbinding a secret each answered 404. That is the security half.

## The gate split

`Fountain.Connections.enabled_for?/1` keeps its name and now answers one question only: **may this account add a credential path?** Broker plus the `connections` flag. It gates connecting an account, defining or editing a provider, running MCP discovery, binding a secret, and pointing a binding at a different host (retargeting a binding sends a credential somewhere it was not going, so it belongs on this side).

`Fountain.Connections.manageable_for?/1` is the broker alone — which is what `Conversations.Egress` and the rest of the runtime already gate on — and answers: **do this account's connections work, and may it list, revoke and delete them?** It gates every read and every removal. A brokered account with no rows gets an empty list rather than a 404, which is the price of never having to ask "do they still hold one?" before answering.

Call sites moved to the right predicate:

| Door | Gate |
|---|---|
| `GET/DELETE /api/connections*` | manageable (no action there adds one) |
| `/api/connection-providers` index, show, delete | manageable |
| `/api/connection-providers` create, update, discover | enabled |
| `/api/secret-bindings` index, presets, delete | manageable |
| `/api/secret-bindings` create, update | enabled |
| `/connections/:provider/start` and `/callback` | enabled |
| `POST /api/mcp/gmail/...` | broker, like every other runtime path |
| `/account/connections`, `/account/bindings` mount, nav links, agent-form connection list | manageable |
| The LiveView events that add (provider form, discovery, bind) | enabled, refused in the handler as well as hidden in the template |

`auth_me_controller`'s `connections_enabled` field is deliberately left on `enabled_for?` — see "Left out" below.

## The self-host default

`@on_without_posthog` in `Fountain.FeatureFlags` names the flags that gate a **shipped** feature, and those read on when `POSTHOG_PROJECT_API_KEY` is unset. "Off because we asked and were told no" and "off because there is nobody to ask" are different answers, and only the first is a decision. There was no prior notion of a default-on flag in the repo. Nothing changes where PostHog is configured (it answers the flag as before, including for a flag it does not mention), and `FEATURE_FLAGS_ON` still wins over both. The switch that turns Connections off on a self-host is `BROKER_TENANTS`.

`CHANGELOG.md` carries this as an **Upgrade notes** entry plus a Fixed entry; `.env.example` and `docs/configuration.md` are updated to match.

## How it was verified

`connections_rollout_test.exs` is rewritten around the split (10 tests), plus two in `feature_flags_test.exs` and one rewritten in `gmail_mcp_controller_test.exs`. The load-bearing test: an account with the flag off and existing connection, provider and binding rows lists and deletes all three over the API (204s), and does the same in both LiveViews.

Every fix was then reverted, the test watched to fail, and restored:

| Reverted | Test that failed |
|---|---|
| `@on_without_posthog` lookup → `false` | "a flag that gates a shipped feature reads on"; "a deployment with no PostHog keeps Connections" |
| All three controller plugs → `enabled_for?` | "an account that holds rows can still list and delete every one of them" |
| The creating half of the plugs → `manageable_for?` (guard weakened) | "adding one is still refused" |
| The LiveView mounts and the nav → `enabled_for?` | "both pages open, the nav still links them…"; "an event that would add something is refused" |
| The LiveView event guards alone | "an event that would add something is refused even when pushed at the page" |
| `agents_live/form` → `enabled_for?` | "an agent form still offers the connections the account holds" |
| The Gmail MCP gate → `enabled_for?` | "the rollout flag going off leaves a connection the agent names working" |

Also run: `mix compile --warnings-as-errors`, `mix credo --strict` on the ten changed lib files (no issues), `mix format` from `apps/fountain`, and the three prose gates (`docs-style.py`, `vale lint docs apps/*/docs`, `destink.mjs` — all clean for the pages touched; the two `docs-style` findings are pre-existing pages I did not touch). 143 tests across the connections, bindings, gmail, auth/me, agents-live, docs and egress suites pass.

## Left out, deliberately

- **`connections_enabled` on `GET /api/auth/me`** still reports `enabled_for?` (may connect). Splitting it for API clients means a second field, which drags in the OpenAPI spec, the SDK contract and four SDK regenerations — out of proportion to this fix. A client that hides its whole Connections UI on that field would still hide revoke; worth a follow-up issue.
- **Diff size.** 482 insertions / 70 deletions, over the ~400 mentioned when this was handed over. 234 of the insertions are the rewritten rollout test and 27 are the CHANGELOG; the rest is the split and its call sites, comment-heavy. Say the word and it can be cut into "feature flag default" and "gate split".

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -190,13 +190,15 @@ EMAIL_FROM=noreply@fountain.example.com
 
 # ── Feature flags ────────────────────────────────────────────────────────────
 # Per-user flags via PostHog (project API key, not a personal key). Unset, every
-# flag reads off unless forced on below. When PostHog is unreachable the last
-# answer it gave is reused; with none, flags read off.
+# flag reads off unless forced on below, except the flags over a shipped feature
+# (`connections`), which read on where there is no PostHog to ask. When PostHog
+# is unreachable the last answer it gave is reused; with none, flags read off.
 # POSTHOG_PROJECT_API_KEY=
 # POSTHOG_HOST=https://us.i.posthog.com
 # Force flags on for every user without PostHog (comma-separated keys).
-# Connections also requires the credential broker and configured provider apps.
-# FEATURE_FLAGS_ON=team_comms,connections
+# Connections also requires the credential broker and configured provider apps,
+# and needs no entry here: BROKER_TENANTS is what turns it on and off.
+# FEATURE_FLAGS_ON=team_comms
 
 # ── Product analytics ────────────────────────────────────────────────────────
 # The same project key above sends product events to PostHog, from the server.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,21 @@ upgrade, is in
 
 ## [Unreleased]
 
+### Upgrade notes
+
+- **Connections needs no `FEATURE_FLAGS_ON` entry on a deployment without
+  PostHog** (#1693). Gating Connections behind the `connections` flag (#1620)
+  took the feature away from every deployment that configures no flag service,
+  because a flag nobody can answer reads off. The flag now reads **on** where
+  `POSTHOG_PROJECT_API_KEY` is unset, so an upgrade keeps the Connections page
+  and the `/api/connections`, `/api/connection-providers` and
+  `/api/secret-bindings` routes. An operator who set
+  `FEATURE_FLAGS_ON=connections` to get the feature back can drop it, and
+  nothing changes where PostHog is configured: it answers the flag as before,
+  and `FEATURE_FLAGS_ON` still wins over both. The switch that turns
+  Connections off is the broker: an account is offered the feature only while
+  `BROKER_TENANTS` names it.
+
 ### Changed
 
 - **The project moved to `github.com/managoat/fountain`** and every coordinate
@@ -108,6 +123,18 @@ upgrade, is in
   and the console's conversation lists render them as chips.
 
 ### Fixed
+
+- **An account whose `connections` flag is off can revoke what it already
+  holds** (#1693). The flag stood in front of every door, the ones that take a
+  credential away included, while the runtime kept brokering those same tokens
+  into sandboxes: revoking a connection, deleting a provider and unbinding a
+  secret each answered 404 for a credential that was still in use. The flag
+  now gates only the doors that add one, which are connecting an account,
+  defining or editing a provider, binding a secret and pointing a binding at a
+  different host. Listing, revoking, unbinding and deleting are open to every
+  account the egress broker is on for, in the console and over the API, and
+  the Gmail MCP endpoint serves a connection that already exists the way the
+  rest of the runtime does.
 
 - A teammate can be moved to a different environment or vault. Fountain retires
   the computer the old binding named, so the teammate's next message builds one
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,11 +130,13 @@ upgrade, is in
   into sandboxes: revoking a connection, deleting a provider and unbinding a
   secret each answered 404 for a credential that was still in use. The flag
   now gates only the doors that add one, which are connecting an account,
-  defining or editing a provider, binding a secret and pointing a binding at a
-  different host. Listing, revoking, unbinding and deleting are open to every
-  account the egress broker is on for, in the console and over the API, and
-  the Gmail MCP endpoint serves a connection that already exists the way the
-  rest of the runtime does.
+  defining or editing a provider, binding a secret, pointing a binding at a
+  different host and enabling a binding that is disabled. Listing, revoking,
+  unbinding, deleting and disabling are open to every account the egress
+  broker is on for, in the console and over the API: `DELETE` unbinds and
+  `PATCH` with `enabled: false` and nothing else disables. The Gmail MCP
+  endpoint serves a connection that already exists the way the rest of the
+  runtime does.
 
 - A teammate can be moved to a different environment or vault. Fountain retires
   the computer the old binding named, so the teammate's next message builds one
--- a/apps/fountain/lib/fountain/connections.ex
+++ b/apps/fountain/lib/fountain/connections.ex
@@ -51,12 +51,39 @@ defmodule Fountain.Connections do
   alias Fountain.Connections.{Connection, OAuth, Platform, Provider}
   alias Managoat.McpAuth
 
-  @doc "Whether this account may use Connections and manage credential bindings."
+  @doc """
+  Whether this account may **add** a connection, a provider or a credential
+  binding.
+
+  Two things have to be true. The egress broker is on for the tenant
+  (ADR 0019) — without it a token would have to enter a sandbox in the clear.
+  And the `connections` rollout flag is on for them. Gate every door that
+  creates something on this one, and nothing else: see
+  `manageable_for?/1` for the rest.
+  """
   @spec enabled_for?(String.t()) :: boolean()
   def enabled_for?(user_id) do
     Fountain.Broker.enabled_for?(user_id) and
       Fountain.FeatureFlags.enabled?(:connections, user_id)
   end
+
+  @doc """
+  Whether this account's existing connections and bindings may be listed,
+  revoked and deleted.
+
+  The broker alone, deliberately, because that is what the runtime paths which
+  attach a token gate on (`Fountain.Conversations.Egress`). A tenant whose
+  rollout flag goes off keeps every credential that is already brokered into
+  their sandboxes, so the doors that take one away have to stay open. Behind
+  the creation gate, revocation answered 404 while the token kept flowing, and
+  the account had no way to turn off a credential that was still in use
+  (#1693).
+
+  A tenant with no rows gets an empty list rather than a 404, which is the
+  price of never having to ask "do they still hold one?" before answering.
+  """
+  @spec manageable_for?(String.t()) :: boolean()
+  def manageable_for?(user_id), do: Fountain.Broker.enabled_for?(user_id)
 
   # How close to expiry a token is considered stale. A turn may run for a
   # while on the token it started with, so refresh well ahead.
--- a/apps/fountain/lib/fountain/feature_flags.ex
+++ b/apps/fountain/lib/fountain/feature_flags.ex
@@ -16,7 +16,9 @@ defmodule Fountain.FeatureFlags do
      and a flag that was off stays off; the outage does not flip anything.
   4. **Off.** No override, no PostHog (or PostHog unreachable with nothing
      cached) → `false`. Fail closed: an unreachable flag service must never
-     turn a feature on.
+     turn a feature on. The one exception is `@on_without_posthog`, the flags
+     that gate a shipped feature rather than an unfinished one: those read on
+     where the deployment configured no PostHog at all.
 
   The lookup is bounded — `@timeout_ms` — so a slow PostHog costs a request
   at most that much once per user per `@fresh_ms`, not on every call. The
@@ -45,12 +47,27 @@ defmodule Fountain.FeatureFlags do
     openai_compat: "openai_compat"
   }
 
+  # The flags that read **on** where there is no PostHog to ask.
+  #
+  # "Off because we asked and were told no" and "off because there is nobody
+  # to ask" are different answers, and only the first one is a decision. Rule
+  # 4 above fails closed on both, which is right for a flag that gates an
+  # unfinished feature: nobody loses anything they had. It is wrong for a flag
+  # that gates a **shipped** one, because a self-host configures no PostHog —
+  # so the flag reads off there forever and the feature disappears on the
+  # upgrade that introduced the flag (#1620, #1693). A flag listed here is on
+  # for a deployment with no `POSTHOG_PROJECT_API_KEY`, and is answered by
+  # PostHog like any other wherever one is configured. `FEATURE_FLAGS_ON`
+  # still wins over both, and so does a PostHog answer of "off".
+  @on_without_posthog Map.new([:connections], &{Map.fetch!(@flags, &1), true})
+
   @doc "The PostHog key for a known flag atom."
   def key!(flag) when is_atom(flag), do: Map.fetch!(@flags, flag)
 
   @doc """
   Whether `flag` is on for `user` — a `%Fountain.Accounts.User{}`, a user id
-  string, or `nil` (no user: only static overrides apply).
+  string, or `nil` (no user: only the answers that are the same for everyone
+  apply, the static overrides and `@on_without_posthog`).
   """
   @spec enabled?(atom | String.t(), term) :: boolean
   def enabled?(flag, user) when is_atom(flag), do: enabled?(key!(flag), user)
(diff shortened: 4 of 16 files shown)


BinaryBourbon (changes requested):
The split is the right shape and the reasoning in the docstrings is the best part of this PR — "off because we asked" versus "off because there is nobody to ask" is the distinction the whole bug turned on. I enumerated the doors myself, spot-checked three of your reverts including the inverse one, and probed the two gates against each other. One door is on the wrong side, and a second one disagrees with itself between the console and the API. Both are in `secret_bindings`.

## What I verified

**Every call site, enumerated independently.** `grep -rn "Connections.enabled_for?\|Connections.manageable_for?\|Broker.enabled_for?" apps ee`, then each controller's action list against its `@creating` list:

| Controller | Actions | `@creating` | Verdict |
|---|---|---|---|
| `ConnectionController` | index, providers, show, delete | — (manage for all) | correct; none adds |
| `ConnectionProviderController` | index, create, show, update, delete, discover | create, update, discover | correct |
| `SecretBindingController` | index, presets, create, update, delete | create, update | correct |
| `ConnectionsController` (start/callback) | start, callback | plug-wide `enabled_for?` | correct; both create |

The per-action plug does fail safe in the direction that matters: `plug :require_connections` is module-wide with no `when action in`, so a new action gets `manageable_for?`, never nothing. Note the residual risk in the other direction — a *creating* action added later silently lands on the weaker gate with no test to catch it. That is the right trade (defaulting to `enabled_for?` reinstates the exact 404 you are fixing), but it is worth a line in the module docs so the next person adding a write action knows to touch `@creating`.

**LiveView events, enumerated against `handle_event/3` clauses and the template's `phx-*` attributes, not against the list in the source.** `connections_live/index.ex` has 13 clauses; `@adding` names 9; the 4 left out are `revoke`, `delete`, `delete_provider` (all one-way removal) and `cancel_provider` (closes a form). Complete. The refusal clause really is first (`:128`, before `revoke` at `:140`), and it carries the `@impl true`. No hole there.

**Revert checks, three of seven.**
- `not configured?() -> false` → 2 failures: "a flag that gates a shipped feature reads on" and "a deployment with no PostHog keeps Connections".
- The inverse one: both plugs' creating half weakened to `manageable_for?` → "adding one is still refused" fails on `|> json_response(404)`. The guard is load-bearing.
- Both LiveView guard clauses deleted → "an event that would add something is refused even when pushed at the page" fails.

**`@on_without_posthog`.** Reads correct. `configured?()` is the only condition, overrides are consulted first in `enabled?/2` so `FEATURE_FLAGS_ON` still wins, and an override of `false` still wins too. Hosted is unaffected: `configured?()` is true there, so PostHog decides, including "off". The PostHog-unreachable-with-empty-cache path is untouched — `configured?()` is true, so it still falls through to the cache and reads off. The one behaviour change nobody will notice: `enabled?(:connections, nil)` now answers true on a PostHog-less deployment where the old `remote_enabled?(_, nil)` clause short-circuited to false; every caller pairs it with `Broker.enabled_for?/1`, which is false for `nil`, so it is inert.

**Gates.** `mix format --check-formatted` and `mix compile --warnings-as-errors` from `apps/fountain` clean; `mix credo --strict` on the six changed lib files, no issues. 119 tests across the rollout, feature-flags, gmail, connection, provider, binding, both LiveViews, auth/me, agents-live and egress suites, 0 failures; `docs_test.exs` 29/29. `destink.mjs` 112 pages clean; `docs-style.py` reports only `docs/catalog/runtimes/codex.md:62` and `docs/concepts/environment.md:73`, neither touched here; vale's 8 SentenceLength errors are all in pages you did not touch and `docs/configuration.md` contributes none. All three are advisory via `prose-advice.sh` anyway.

## Findings

### 1. `toggle` re-enables a disabled binding with the flag off, and the API refuses the same change

`apps/fountain/lib/fountain_web/live/secret_bindings_live/index.ex:54` — `@binding_new` is `~w(draft pick_key preset save)`, and the comment at `:52` says "`toggle` and `delete` are not here: they take a binding away". `toggle` is bidirectional. The handler at `:148` writes `%{"enabled" => not binding.enabled}` unconditionally, and the button at `:322` renders "Enable" for a disabled binding regardless of `@may_bind?`.

Measured on the branch, flag off, broker on, against a binding that is already disabled:

```
API PATCH enabled=true  -> 404   (still disabled)
LiveView render_click "toggle" -> no flash, binding.enabled == true
```

So the console does the thing the API refuses, and it is the thing the rollout flag exists to prevent: a secret starts flowing to a host it was not flowing to a moment ago. By your own standard for retargeting ("it sends a credential somewhere it was not going"), enabling is creation — which is exactly why you put `:update` on the creating side for the API, since `enabled` is in `@fields`. The two doors need the same answer.

Smallest fix: refuse `toggle` in the same first clause when the binding is currently disabled, and hide or disable the Enable button when `!@may_bind?`. Disabling must stay open. A test in the shape of "an event that would add something is refused" — disabled binding, flag off, push `toggle`, assert it stays disabled — would have caught it, and is the one adding path the new LiveView test does not push.

### 2. The API has no off switch short of destroying the binding

The mirror of the above, same gate. `PATCH /api/secret-bindings/:id` with `{"enabled": false}` answers 404 for a flag-off tenant (measured), because `:update` is wholly on the creating side. The PR's thesis is that the doors which take a credential away stay open; in the console "Disable" is such a door and it works, over the API it does not. `DELETE` is open, so the tenant is not stuck — but their only remedy is to destroy a binding they may want back, and an API client cannot offer the softer switch the console offers.

I do not think this blocks on its own, and I would not want `require_brokerage` growing a body-inspecting special case for it. But pick one and write it down: either `update` whose only changed field is `enabled: false` falls to the manage gate, or the module doc says plainly that `DELETE` is the API's off switch and disabling needs the flag. Right now the docstring at `:9-17` implies the first and the code does the second.

## The two things you asked about

**Retargeting.** Treating `PATCH` with a new `host` as creation is right, and `assert %{host: "api.stripe.com"} = ...` after the refused patch is the correct way to prove it. `DELETE` as the off switch is right. The boundary itself is sound; it is only `enabled` that straddles it, which is finding 1.

**Size.** Keep it as one PR. The two halves are not two changes, they are one cause read in two directions: a flag that nobody can answer should not have been able to take a shipped feature away, and a flag that is off should not have been able to take away the ability to *stop* one. Splitting them would land a "feature flag default" PR whose justification is entirely in the PR that follows it, and the rollout test would have to be written twice. 234 of the 482 insertions are that test and 27 are the CHANGELOG; the actual gate change is small and reads in one sitting.

`connections_enabled` staying on `enabled_for?` with #1922 filed is the right call — it is a wire-contract change and it does not belong here.

Requesting changes on finding 1 only; everything else above is verification or a question. Re-approve is a small diff away.

jhgaylor:
Fixed in `90b18899`. You were right, and the comment was the reason it was invisible — I classified `toggle` by its name and wrote down the half of the answer that suited me.

**The blocking one.** `toggle` writes `enabled: not enabled`, so the direction is the only thing that says which gate it belongs under. It is now gated on the direction rather than the name, in both doors:

- `secret_bindings_live/index.ex` — the handler reads the binding first and refuses when `not binding.enabled and not may_bind?`, naming the key and host in the flash. The Enable button carries `:if={b.enabled or @may_bind?}`, so the control is gone as well as the event refused. The body of the flip moved to a private `flip/2` so the `cond` stays readable.
- The `@binding_new` comment now says why `toggle` is absent from the list and where it is gated instead.

**The mirror, which I have fixed rather than documented.** `PATCH {"enabled": false}` was refused too, so the API had no soft off switch while the console's Disable worked. `:update` is now creating *unless* the params are exactly `enabled: false` — the same act as `DELETE`, and not a new credential path. Anything else, including `enabled: true` or a disabling that also carries a host, needs the flag. The two doors agree now, and the module docstring says exactly that instead of implying it.

**Residual risk.** Both `@creating` lists gained a line: an action left off the list gets the weaker gate rather than none, which is the right default for one that reads or removes and the wrong one for a new action that creates — so it belongs on the list in the commit that adds it.

**Tests** (`connections_rollout_test.exs`, all flag off / broker on):

- `a binding can be disabled but not enabled again, over the API` — `PATCH {enabled: false}` → 200 and disabled; `PATCH {enabled: true}` → 404 and still disabled; `PATCH {enabled: false, host: "api.attacker.example"}` → 404, host unchanged.
- `the console cannot enable a disabled binding either` — the button is absent, `render_click "toggle"` is refused, the row stays disabled.
- `the console still disables an enabled binding` — the door that must stay open, still open.

Verified by reverting, three ways: dropping the LiveView direction check fails the console test; making `:update` wholly creating again fails the API test; making every `:update` management fails both that test and `adding one is still refused`, so the guard bites in the permissive direction too.

`mix format`, `mix compile --warnings-as-errors` and `mix credo --strict` on the changed files are clean; the binding, provider, rollout and bindings-LiveView suites pass (42 tests).

Kept as one PR, as agreed. Not merging.

BinaryBourbon (approved):
Approving. `90b18899` closes the hole and its mirror, and I could not find a params shape that gets a creating effect past the new predicate.

## Ruling on the body-inspecting plug

I said "decide and document, rather than a body-inspecting plug", and you built the plug. You were right and I was being squeamish. Two properties make this one safe in a way a general body-inspecting gate is not, and both are load-bearing, so I want them named where the next person will read them:

1. **The predicate and the write read the same map through the same `@fields`.** `disabling_only?/1` inspects `Map.take(conn.params, @fields)`; `binding_attrs/1` writes `Map.take(params, @fields)`; Phoenix hands the action the same `conn.params` the plug saw, path params merged before controller plugs run. So the gate cannot be shown one thing and the write another. That is the whole argument, and it holds only while `update/2` has exactly one write path. A second one that bypasses `binding_attrs/1` would break it silently — worth a sentence in the `creating?/1` comment.
2. **`@fields` drift is fail-safe by construction.** Add a writable field and a request carrying it fails `map_size(attrs) == 1`, so it lands on the creating gate. Remove `enabled` and the predicate always answers false, so `update` is always creating. Both directions close, neither opens.

I also like that `:update` stays in `@creating` even though `creating?/1` intercepts it first. Delete the `:update ->` branch and it falls through to `action in @creating` and is still gated. Please don't let anyone tidy that away as dead.

## What I probed

Thirteen params shapes against a `PATCH`, flag off and broker on, run twice — once against an enabled binding and once against a disabled one — asserting the status *and* re-reading the row afterwards:

| Params | Status | Row after |
|---|---|---|
| `enabled: false` (bool), `enabled: "false"` | 200 | disabled |
| `enabled: true` | 404 | unchanged |
| `enabled: 0`, `enabled: "FALSE"` | 404 | unchanged |
| `enabled: ["false"]`, `enabled: {"a": "false"}` | 404 | unchanged |
| `enabled: false` + `host` | 404 | host still `api.stripe.com` |
| `enabled: false` + `key` | 404 | unchanged |
| `enabled: false` + `user_id` | 200 | disabled, `user_id` unchanged |
| `enabled: false` + a key outside `@fields` | 200 | disabled, nothing else written |
| `host` alone, `{}` | 404 | unchanged |

The encodings all fail closed, which is the direction that matters: nothing the predicate accepts casts to `true` in `Ecto.Type.cast(:boolean, _)`, so there is no value that reads as disabling and writes as enabling. The `user_id` row is worth calling out — it passes because `user_id` is not in `@fields`, so `Map.take` drops it from both the predicate and the write, and `update_binding/3` deletes it again for good measure. No tenant hop.

One shape you did not test and I did: **query string and body disagreeing.** `PATCH /api/secret-bindings/:id?enabled=false` with `{"host": "api.attacker.example"}` in the body → 404, host unchanged. `Plug.Parsers` merges both into `conn.params` before the router, so the plug sees `map_size == 2` like everything else. Consider adding it; it is the shape someone reaches for when they think a gate reads only the body.

**Side effects of the now-open write.** `update_binding/3` casts `%{"enabled" => false}` and nothing else; `Binding.changeset/2`'s trims and validations touch only fields that are present. Disabling has no effect beyond `enabled`. The 200 renders the same body `GET /api/secret-bindings/:id` already returns to a flag-off account, so it discloses nothing new.

**The console.** Against a disabled binding with the flag off: the Enable button is absent, `render_click(view, "toggle", ...)` is refused with the flash, and the row stays disabled. Against an enabled one: the Disable button is present and the flip works. Both halves verified live, not read.

## Reverts, four of my own

| Broken | Failed |
|---|---|
| `map_size(attrs) == 1` guard dropped from `disabling_only?/1` | "a binding can be disabled but not enabled again" (the disable-plus-host half) |
| `:update -> false` (guard broken open) | that test **and** "adding one is still refused" |
| The `not binding.enabled and not may_bind?` branch deleted | "the console cannot enable a disabled binding either" |
| `:if={b.enabled or @may_bind?}` deleted from the button | same test, the `refute has_element?` half |

The last two matter together: the button guard and the handler guard are independently load-bearing and the test asserts both, so neither can rot into decoration.

## Gates

`mix format --check-formatted`, `mix compile --warnings-as-errors` and `mix credo --strict` on the three changed lib files are all clean from `apps/fountain`. 151 tests across the rollout, feature-flags, gmail, connection, provider, binding, both LiveViews, auth/me, agents-live, egress and docs suites, 0 failures. `destink` 112 pages clean; `docs-style` still reports only `docs/catalog/runtimes/codex.md:62` and `docs/concepts/environment.md:73`, neither touched here.

## Size

Unchanged: keep it as one PR. The second commit is +139 and it is the same thought finishing itself — `enabled` is the one field whose two values sit on opposite sides of the gate, and both doors now say so. Splitting that out would produce a PR whose diff is a boolean predicate and whose justification is the PR before it.

The CHANGELOG now names enabling as a gated door and disabling as an open one, and the controller docstring says which verb does which. Someone reading this in a year will be able to reconstruct the rule without reading the code, which is the bar.

Approved.

jhgaylor:

PR #1926, opened 2026-09-11
fix(cli): preserve literal apply variable values

### LINKED ISSUE #1896: fountain apply --var splits values on commas, silently defining an extra variable
`fountain apply --var` is declared as cobra `StringSlice`, which parses **each occurrence as CSV**. A value containing a comma is split into extra flag values, and `buildApplyVars/1` then reads the fragments as if they were separate `KEY=VAL` pairs.

Found while reviewing #1884, which hit the same trap on `--redirect-uri` and fixed it by switching to `StringArray`.

## What happens

Measured against `pflag` directly (`cli/internal/cmd`, Go pinned from `cli/go.mod`):

```
--var=HOSTS=a,b      StringSlice -> ["HOSTS=a", "b"]
--var=PAIRS=a=1,b=2  StringSlice -> ["PAIRS=a=1", "b=2"]
--var=Q=a\,b         StringSlice -> ["Q=a\\", "b"]        # backslash is not an escape
--arr=HOSTS=a,b      StringArray -> ["HOSTS=a,b"]         # what we want
```

Two failure modes fall out of that, and the second is the bad one:

**1. A confusing fatal.** `buildApplyVars` (`cli/internal/cmd/apply.go:236-242`) rejects the fragment that has no `=`:

```
$ fountain apply -f stack.yaml --var 'HOSTS=web1,web2'
--var must be KEY=VALUE, got: "web2"
```

The user never typed `web2` as a `--var`. Nothing in the message points at the comma.

**2. A silently wrong substitution.** When the fragment happens to contain an `=`, there is no error at all:

```
$ fountain apply -f stack.yaml --var 'PAIRS=a=1,b=2'
```

sets `PAIRS=a=1` **and** defines a second variable `b=2` that the user did not ask for. `${PAIRS}` then substitutes a truncated value into an environment or vault secret, and `${b}` resolves where it should have been reported as missing by `Managoat.Substitution`. No error, no warning.

There is no quoting escape hatch: `pflag`'s CSV reader does not honour a backslash, and shell quotes are consumed before `pflag` sees the value.

## Fix

`StringArray` instead of `StringSlice`, the same one-word change #1884 made in two places:

- `cli/internal/cmd/apply.go:24` — the declaration
- `cli/internal/cmd/apply.go:37` — `GetStringSlice` → `GetStringArray`

`StringArray` appends each occurrence verbatim and is the right type for anything that carries arbitrary text. `--var` is repeatable already, so no documented usage changes.

Two knock-on edits, both of which #1884 shows the shape of:

- Cobra prints the flag's type name in help, so `--var strings` becomes `--var stringArray` and `docs/cli/commands.md` must be regenerated. `cli/internal/cmd/docs_test.go` fails until it is.
- A regression test is worth having here even though #1884 did not add one, because this flag has a silent failure mode rather than only a loud one: assert that `--var 'PAIRS=a=1,b=2'` yields exactly one variable.

## Also worth a look, lower priority

`fountain conv prompt --image` is `StringSliceP` (`cli/internal/cmd/conv.go:39`, read at `:187`). A comma in a filename is legal and rare, and the failure is a loud "file not found" rather than a silent one — so it is the same bug with a much better error message. `fountain webhooks create --event` (`cli/internal/cmd/webhooks.go:71`) is fine as-is: event names carry no commas, and CSV is arguably the wanted behaviour there.

jhgaylor:
`fountain apply --var 'PAIRS=a=1,b=2'` now keeps the complete value instead of silently introducing `b=2` as another variable. The flag uses literal repeatable values, with a regression covering commas, quotes, empty values, and the existing environment merge. The generated CLI reference is refreshed.

Closes #1896.

Three files, +39/-5. Independent PR against main.

Validation: the regression fails with the old CSV flag parser; `go test ./...` and `go vet ./...` pass. Full local `mix precommit` passes: core 5,224 tests + 6 doctests, Buzz 144 tests, Support 33 tests; zero failures. Docs-style and Vale reports were reviewed (existing advisory findings); destink is clean.

### DIFF
--- a/cli/internal/cmd/apply.go
+++ b/cli/internal/cmd/apply.go
@@ -15,14 +15,18 @@ import (
 )
 
 func init() {
+	rootCmd.AddCommand(newApplyCmd())
+}
+
+func newApplyCmd() *cobra.Command {
 	applyCmd := &cobra.Command{
 		Use:   "apply",
 		Short: "Apply resource definitions from a YAML file or directory",
 		RunE:  runApply,
 	}
 	applyCmd.Flags().StringP("file", "f", "", "path to YAML file or directory")
-	applyCmd.Flags().StringSlice("var", nil, "extra variable for ${VAR} substitution (KEY=VAL, repeatable)")
-	rootCmd.AddCommand(applyCmd)
+	applyCmd.Flags().StringArray("var", nil, "extra variable for ${VAR} substitution (KEY=VAL, repeatable)")
+	return applyCmd
 }
 
 func runApply(cmd *cobra.Command, args []string) error {
@@ -34,7 +38,7 @@ func runApply(cmd *cobra.Command, args []string) error {
 		Fatal("usage: fountain apply -f <path-to-yaml> [--var KEY=VAL ...]")
 	}
 
-	varFlags, _ := cmd.Flags().GetStringSlice("var")
+	varFlags, _ := cmd.Flags().GetStringArray("var")
 	applyVars := buildApplyVars(varFlags)
 
 	docs, err := manifest.Read(path)
--- a/cli/internal/cmd/apply_test.go
+++ b/cli/internal/cmd/apply_test.go
@@ -138,3 +138,33 @@ func TestFormatResultErrors(t *testing.T) {
 		t.Errorf("nil errors should fall back to generic message")
 	}
 }
+
+func TestApplyVarsPreserveLiteralValues(t *testing.T) {
+	cmd := newApplyCmd()
+	err := cmd.ParseFlags([]string{
+		"--var", "PAIRS=a=1,b=2",
+		"--var", "HOSTS=web1,web2",
+		"--var", `QUOTED="a,b"`,
+		"--var", "EMPTY=",
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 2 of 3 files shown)


BinaryBourbon (commented):
Reviewed `14c0c914dade12d4ea20c461a0b6dd557c7269fa` for #1896. No actionable findings identified.

Checked literal flag parsing, first-equals splitting and environment merging. Targeted apply tests and generated CLI-reference parity pass: go test -count=1 ./internal/cmd -run 'TestApply|TestBuildApply|TestFormatResultErrors|TestGeneratedCLIReferenceIsCurrent'.

Limitations: Did not execute apply against a server or rerun the entire Go suite.

BinaryBourbon (approved):
Approve `14c0c914dade12d4ea20c461a0b6dd557c7269fa`. The flag now preserves each repeatable KEY=VALUE literally. Rechecked constructor registration and downstream first-equals splitting/environment override semantics; commas and quotes cannot create unintended variables. Generated help matches the new flag type.

Validation: Local targeted apply/environment-merge and generated CLI-reference parity tests pass. Exact-head SDK/CLI and full CI pass. Full CI passed for this exact head: https://github.com/managoat/fountain/actions/runs/34592432724.

Limits: No apply request was sent to a server.

PR #1927, opened 2026-09-11
ci: reject conflict markers in tracked text files

### LINKED ISSUE #1898: No gate catches a conflict marker; one reached /docs/changelog
A `>>>>>>> theirs` conflict marker reached `main` in `CHANGELOG.md` and was served to readers at `/docs/changelog` until #1881 removed it. Nothing in the repo would have caught it, and nothing would catch the next one.

## What happened

`9da64a10` (the #1839 stack tip) added the `runtime_command` changelog block from a botched resolution: the resolver kept *theirs* wholesale, dropped the `<<<<<<<` and `=======` lines, and left the trailing `>>>>>>> theirs`. No content was lost, so nothing read as broken in the diff.

`CHANGELOG.md` is a published docs page — `Fountain.Docs` declares it on `extra_resources:` (`apps/fountain/lib/fountain/docs.ex:48`), `docs/changelog.md` snippet-includes it, and `docs/nav.yml:150` mounts it. `>>>>>>>` is seven nested blockquotes in markdown, so it rendered in production as an indented box holding the word "theirs" in the middle of the unreleased notes:

```
$ curl -s https://managoat.com/docs/changelog | grep -B2 -A2 theirs
<blockquote><blockquote><p>theirs</p></blockquote></blockquote>
```

Full CI was green on that merge, and full CI is green on #1881. Neither run had anything to say about it.

## Why no gate saw it

Two independent holes, and the marker fell through both.

**There is no conflict-marker check anywhere.** A grep for the three marker forms over `.ex`, `.exs`, `.sh`, `.yml`, `.py` and `.mjs` finds no test, no script and no CI step that looks for them. This is a class gap, not a changelog one: a marker in an Elixir source file fails to compile, but one in markdown, JSON fixtures, `.env.example`, or a docs page under `apps/*/docs/` is invisible to every gate we have.

**`CHANGELOG.md` sits in a seam of the docs gates.** The `changes` job keeps it off the docs-only allowlist deliberately (`ci.yml:115`) so the release-bump flow cannot skip the suite. That is the right call. The side effect is that `docs_touched` stays false for a changelog-only diff, so **Docs** and **Docs prose gates** both skip — as they did on #1881. And the prose linters take `docs/` and `apps/*/docs/` as their path arguments, not the repo root. So the one file that is both a published docs page and prose-gate-exempt is exactly where this landed.

## What to build

A tree-wide marker check, not a changelog one. Two candidate homes:

1. **A `scripts/` gate** beside `coverage-gate.exs` and `hex-audit-gate.exs`, run from the `elixir-static` job or `CI policy tests`, grepping the tracked tree. Catches every file type, including ones no Elixir test reads.
2. **`docs_test.exs`**, which already walks the manual and the declared `external_resources/0`. Narrower — it would have caught this instance, since `CHANGELOG.md` is a declared resource — but it leaves fixtures and config files uncovered.

(1) is the one that closes the class. Worth considering whether it also belongs in `mix precommit`, so the marker dies before the push rather than in CI.

Two things for whoever picks it up:

- **The backlog is empty.** The tree is clean today, so the guard can go in with no allowlist and no ratchet file — the shape the docs gates had to grow into over #911 and the destink work.
- **Watch the self-match.** The check's own source, and any doc that explains git conflict resolution, will contain the patterns it searches for. Anchor to line start and exclude the gate's own path rather than adding a general allowlist, so the list cannot start growing.

Found while reviewing #1881.

jhgaylor:
A lone trailing merge-conflict marker can render as valid Markdown and reach published docs. A tracked-text scan now rejects standard Git conflict markers, including diff3 ancestors, both before local precommit compilation and in the always-required CI policy job.

Closes #1898.

Four files, +100. Independent PR against main. The scan covers every tracked text-file type, preserves inline examples and NOTICE separators, and needs no path allowlist.

Validation: all 27 CI policy tests pass, including marker fixtures in Markdown, JSON, config, and filenames with spaces. A temporary trailing marker also makes `mix precommit` fail before compilation. Full local `mix precommit` passes: core 5,224 tests + 6 doctests, Buzz 144 tests, Support 33 tests; zero failures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1786,6 +1786,8 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v5
         with:
           persist-credentials: false
+      - name: Reject conflict markers
+        run: python3 scripts/conflict-markers.py
       - name: Test CI decisions
         run: python3 -m unittest discover -s scripts/ci -p 'test_*.py' -v
 
--- a/mix.exs
+++ b/mix.exs
@@ -208,6 +208,7 @@ defmodule Fountain.Umbrella.MixProject do
       "ecto.migrate": [&migrate_in_app/1],
       "ecto.rollback": [&rollback_in_app/1],
       precommit: [
+        &conflict_markers_absent/1,
         "compile --warnings-as-errors",
         &deps_unlock_unused_changes_nothing/1,
         "format --check-formatted",
@@ -218,6 +219,11 @@ defmodule Fountain.Umbrella.MixProject do
         "test"
       ]
     ]
+  end
+
+  defp conflict_markers_absent(_args) do
+    {_, status} = System.cmd("python3", ["scripts/conflict-markers.py"], into: IO.stream())
+    if status != 0, do: Mix.raise("Conflict-marker check failed")
   end
 
   # Parity with CI's `mix deps.unlock --unused && git diff --exit-code
--- a/scripts/ci/test_conflict_markers.py
+++ b/scripts/ci/test_conflict_markers.py
@@ -1,0 +1,66 @@
+from pathlib import Path
+import subprocess
+import sys
+import tempfile
+import unittest
+
+
+GATE = Path(__file__).resolve().parents[1] / "conflict-markers.py"
+
+
+class ConflictMarkersTest(unittest.TestCase):
+    def setUp(self):
+        self.temp = tempfile.TemporaryDirectory()
+        self.addCleanup(self.temp.cleanup)
+        self.root = Path(self.temp.name)
+        self.git("init", "--quiet")
+
+    def git(self, *args):
+        return subprocess.run(["git", *args], cwd=self.root, check=True,
+                              capture_output=True)
+
+    def scan(self):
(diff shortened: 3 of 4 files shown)


BinaryBourbon (commented):
Reviewed `ed6120c9b814fc76f98c955ab90327b1ef13e434` for #1898. No actionable findings identified.

Checked tracked-file scanning, exit-status handling and the always-required CI/local precommit integration. All 27 CI policy tests pass, and a scan of an isolated tracked copy of the PR tree is clean.

Limitations: Did not rerun the complete mix precommit suite.

BinaryBourbon (approved):
Approve `ed6120c9b814fc76f98c955ab90327b1ef13e434`. The scan rejects standard conflict-marker lines across tracked working-tree text, preserves inline examples/binary files, and fails closed on Git errors. It runs before local compilation and in the CI job required by every plan.

Validation: Local: all 27 CI policy tests pass and the isolated tracked PR tree scans clean. Reviewed failure-code handling and fixture coverage for all four marker forms. Full CI passed for this exact head: https://github.com/managoat/fountain/actions/runs/34592878137.

Limits: The full precommit suite was not rerun locally; exact-head full CI passes.

PR #1928, opened 2026-09-11
refactor(permissions): centralize reserved policy validation

### LINKED ISSUE #1900: A second reserved permission_policy key would be validated by nothing
Raised as a non-blocking nit during the `stack:1635` review (#1852) and left open there, because closing it properly needs a small design choice rather than a one-liner.

## The hazard

`Fountain.PermissionPolicy` introduced the idea of a **reserved** policy key — one that names no tool and so must be stripped before the map reaches `Managoat.ACP.Permissions`, whose every function reads a key as a tool and a value as a verdict:

```elixir
@reserved [@ask_timeout]

def verdicts(policy) when is_map(policy), do: Map.drop(policy, @reserved)
def reserved_keys, do: @reserved
```

`verdicts/1` correctly drops **every** reserved key. But the two validators that then check what was dropped each match the single literal:

- `Fountain.Agents.Agent.reserved_errors/1` — `case Map.fetch(policy, "ask_timeout")`
- `Fountain.Conversations.validate_reserved_keys/1` — the same

So a second entry added to `@reserved` would be stripped from verdict validation by `verdicts/1` and validated by **nothing**: accepted with any value at all, stored, and then read by whatever consumes it with no guarantee about its shape. `ask_timeout` is the only reserved key today, so behaviour is currently identical and there is no live bug.

## Why it is not a one-liner

Iterating `reserved_keys/0` needs a per-key validator, because the keys are not the same shape — `ask_timeout` is a bounded number of seconds and the next one need not be — and each needs its own error message. Something like a `validate_reserved/2` in `Fountain.PermissionPolicy` that both doors call, with the per-key rules beside the `@reserved` list that declares them, so adding a key means adding its rule in the same place.

## When

Worth doing at the moment a second reserved key is proposed, and cheap enough to do before that. A guardrail test that asserts every `reserved_keys/0` entry is refused for at least one bad value would make the hazard self-closing.

jhgaylor:
Agent changesets and conversation launch validation now share the reserved-key validators in `Fountain.PermissionPolicy`. Every declared reserved key is checked before it can disappear from tool-verdict validation; a missing validator fails the guardrail instead of silently accepting arbitrary values.

Closes #1900.

Four files, +57/-30. Independent PR against main. Existing ask_timeout bounds, agent error messages, and conversation error tuples are preserved. This adds no new policy key.

Validation: all 22 policy unit tests pass, including the guardrail for every reserved key. All 132 targeted policy/agent/conversation-launch tests pass. Full local `mix precommit` passes: core 5,227 tests + 6 doctests, Buzz 144 tests, Support 33 tests; zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/agents/agent.ex
+++ b/apps/fountain/lib/fountain/agents/agent.ex
@@ -280,28 +280,10 @@ defmodule Fountain.Agents.Agent do
     end)
   end
 
-  # `ask_timeout` names no tool, so it is validated on its own rather than as
-  # a verdict (#1635). Seconds, positive, and bounded above by
-  # `PermissionPolicy.max_ask_timeout_seconds/0` — a year, which is not the
-  # idle bound in disguise but the point past which the deadline no longer
-  # fits in a `timestamp`. The message names the bound, because a refusal
-  # that only says "not a positive number" about 1e15 reads as a lie.
   defp reserved_errors(policy) do
-    case Map.fetch(policy, "ask_timeout") do
-      {:ok, value} ->
-        if PermissionPolicy.valid_ask_timeout?(value) do
-          []
-        else
-          [
-            permission_policy:
-              "ask_timeout: #{inspect(value)} is not a positive number of seconds " <>
-                "no greater than #{PermissionPolicy.max_ask_timeout_seconds()}"
-          ]
-        end
-
-      :error ->
-        []
-    end
+    Enum.map(PermissionPolicy.reserved_errors(policy), fn {key, message} ->
+      {:permission_policy, "#{key}: #{message}"}
+    end)
   end
 
   # A policy the runtime will never consult is refused rather than stored. The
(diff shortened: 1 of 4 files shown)


BinaryBourbon (commented):
Reviewed `03a581dbd69421aa240a2060cdce6a5cc377945e` for #1900. No actionable findings identified.

Checked both agent and conversation validation callers, error translation, timeout bounds and the reserved-key guardrail. All 22 permission-policy tests pass using standalone ExUnit against the exact PR module/test sources and the local ACP dependency source.

Limitations: Did not rerun database-backed agent/conversation integration tests.

BinaryBourbon (approved):
Approve `03a581dbd69421aa240a2060cdce6a5cc377945e`. Both policy write paths now call one reserved-key validator after map validation. Timeout bounds and agent/conversation error shapes are preserved; the guardrail fails if a future reserved key lacks a validator.

Validation: Local: all 22 policy tests pass through standalone ExUnit against exact-head sources and the local ACP dependency. Exact-head integration test partitions and contract/static checks pass in CI. Full CI passed for this exact head: https://github.com/managoat/fountain/actions/runs/34592921071.

Limits: Database-backed agent/conversation tests were not rerun locally.

PR #1931, opened 2026-09-11
fix: preserve git diff failures through output encoding

### LINKED ISSUE #1904: /diff cannot distinguish a failed git diff from an empty one
`GET /api/sandboxes/:id/diff` cannot tell a failed `git diff` from an empty one. The same repository that now returns 422 from `/git-status` returns `200 {"diff": ""}` from `/diff`, so the two endpoints disagree about the same repository.

This is the last instance of the class fixed for `/git-status` in #1596. There, `status_script/0`'s exit status was always `head`'s (the last command in a pipeline), so any git failure after repository discovery produced a response byte-identical to a clean tree. #1596 fixed that by capturing `${PIPESTATUS[0]}`, passing 0 and 141, and failing anything else as `exit 8`.

`diff_script/0` still has the original shape.

## Reproduction

Measured against `feat/sandbox-git-status` at `0e93bb17` by extracting the script verbatim and running it under bash against a real repository:

1. A repository with a corrupt `.git/index` (`git rev-parse --show-toplevel` succeeds, `git diff` writes `fatal: .git/index: index file smaller than expected` and exits nonzero).
2. `/git-status` on it: **422**, `{:sandbox_command_failed, 8, "fatal: ..."}`.
3. `/diff` on it: **200**, `{"diff": ""}` — indistinguishable from a repository with no changes.

The realistic trigger is not a corrupt index but a clean/smudge filter that fails: an agent that clones an LFS repository into a sandbox without `git-lfs` on PATH hits exactly this, and the caller renders "no changes" over a tree the agent has rewritten.

## Fix

Apply #1596's treatment to `diff_script/0`: capture git's own exit status rather than the pipeline's, keep 141 passing (`head -c` closing the pipe gives git SIGPIPE, and truncation is intended behaviour), and fail the rest through the existing catch-all.

Note `/diff` shipped in SDK 1.15.0, so a repository that currently answers 200 will begin answering 422. That is the point of the change, but it is a visible behaviour change for existing callers and wants a release-note line.

Found during review of #1596. `/diff`'s unconfined repository discovery and its unredacted `repo_root` were both fixed there already; this is what is left.

jhgaylor:
Git failures in `/diff` could return a successful empty diff because the encoding pipeline hid Git's exit status. Preserve pipeline failures and return the existing 422 error, while accepting intentional SIGPIPE truncation. Hold encoded output until success so failures expose only bounded diagnostics to the existing redactor.

Closes #1904. Three files, +66/-4, including the changelog.

Validation: real Git regression reproduces the corrupt-index failure on the parent; tests also cover clean trees, encoding failure, and large truncated diffs. All 68 focused tests pass. Full local precommit passes: 5,285 core tests plus 6 doctests, 144 Buzz tests, and 33 Support tests, with no failures.


BinaryBourbon (commented):
Reviewed `891eb7736219d58582ed15115eae41b58d4d492b`.

No actionable findings. The script preserves pipeline failures, withholds encoded partial output on error and continues to allow intentional SIGPIPE truncation.

Validation: ran the exact-head `sandbox_files_script_test.exs` suite standalone with the exact-head `SandboxFiles` module and existing local dependency beams: **12 tests, 0 failures**. This executes real Bash/Git against temporary repositories, including corrupt-index, failed-encoder and byte-cap cases. Full app/controller integration was not run; standalone compilation emitted unrelated missing-module/stale-export warnings outside the exercised script path.

BinaryBourbon (approved):
**Approved** at `891eb7736219d58582ed15115eae41b58d4d492b`.

The diff script correctly preserves Git/encoder failure, withholds partial base64 output on error and permits the intentional SIGPIPE from the byte cap. Error handling remains bounded and flows through the existing redacted command-error path. No blocking findings.

Validation: the exact-head real Bash/Git shell regression suite passed locally (**12 tests, 0 failures**) using the exact-head module and existing local dependency beams, including corrupt-index, failed-encoder and byte-cap cases. Head/base are unchanged; [CI run 34595416179](https://github.com/managoat/fountain/actions/runs/34595416179) is verified at this head with required/static/contract and all six test partitions green. I did not run a real provider/controller smoke test or a fresh full dependency build locally.

BinaryBourbon (approved):
Approved at `4e574c8cfccfff7effe1b4f215f447417492f5df`. Independently re-reviewed this pure merge of the previously approved `891eb773` with main `95b58efc`: its diff against main preserves the exact original three-file patch (stable patch-id `50bdaeb0666e1574c409b740c71a2b2675ebf6e2`), with no conflict edits. Main's redaction-overlap changes remain intact. No blocking integration finding.

Validation: ran the real Bash/Git regression suite against the merged SandboxFiles module: **12 tests, 0 failures**. Verified the committed source/test bytes exactly match the tested tree. This was a focused standalone run with existing dependency beams; no new full app/provider smoke test. Fresh required CI must pass before merge.

BinaryBourbon (approved):
Approved at `0c34d8f5a24afb6a1a1b0a6284d5baea9ef467ab` after updating #1931 from main `d8a3e077cea41f7f4eef1bbe82181d20fd2b7e41`. The only merge conflict was adjacent changelog additions; both the `/diff` fix and main's `execution_fenced` fix are preserved verbatim. Against main, this still changes only the original three files (+66/-4); the code-and-tests patch-id is identical to the originally reviewed PR (`a5268a0163db9e6c991fd1c24c93e5de271ce0ad`). No blocking integration finding.

Validation: reran the real Bash/Git shell suite against the merged SandboxFiles module: **12 tests, 0 failures**. Source/test bytes also match the previously tested main-update version; diff whitespace checks pass and the worktree is clean. This was a focused standalone run with existing dependency beams, not a fresh full application/provider test. Fresh required CI must pass before merge.

PR #1933, opened 2026-09-11
fix: separate credential and agent forms for fresh accounts

### LINKED ISSUE #1660: Fresh-account agent form nests credential form and loses browser controls
The deployed browser profile (#1618, tracker #1606) cannot finish creating an agent on a fresh verified account with no inference credentials. The inline credential form is nested inside the main agent form, producing invalid HTML and unstable browser DOM updates.

Observed on production revision `c795fea17e6336c45740cb7c1a4c662cbd85bcf9`, digest `sha256:3f7db53fc4e8d799f3ce543ef5d3e54d56bb1eeea3cac048f5096966256c7449`, with pinned Playwright 1.63.0 / Chromium build 1243.

After waiting for LiveView connection, sign-in and empty credential validation pass. The agent form initially exposes Name, Description, System prompt and Model fields. Filling them and selecting Codex causes the provider selector to disappear in some observed DOM states, despite `/api/catalog` advertising four providers. A diagnostic saw zero provider selects; another saw the expected select. A full run changed the driver to require the catalog-implied selector and wait for it, and the create-agent step still timed out after 30 seconds. The retained trace does not distinguish the provider wait from the later Save/navigation step; finer step instrumentation is needed. Cleanup could not find the run-named agent and retained its intent. The nested markup is a likely cause of the unstable DOM, not a fully isolated causal verdict.

Source evidence in `apps/fountain/lib/fountain_web/live/agents_live/form.ex`:

- The main form begins around line 681 with `phx-change="validate"` and `phx-submit="submit"`.
- It renders `missing_credential_card` inside that form, before the provider selector.
- The card contains another `<form phx-submit="save_credential">` around line 1437.

`apps/fountain/test/fountain_web/live/agents_live_test.exs` around line 205 explicitly notes the nested form and seeds credentials to avoid it for certain form tests. That leaves the fresh-account browser path uncovered.

Fix the markup so credential and agent forms are independent while preserving both save paths. Add a fresh-account regression that checks valid form structure, fills runtime/model/provider fields, saves the agent and separately exercises credential validation. Verify with a real browser; do not seed a provider key merely to hide the missing-credential card.

The first browser run also found an independent harness hydration race: controls were visible before `[data-phx-main].phx-connected`; waiting for connection made credential validation pass. That driver fix is retained separately. It does not resolve this product defect.

Evidence retained locally under `/private/tmp/fountain-conformance-live-20260906/`: `browser-console-3/`, `browser-console-4/`, `browser-agent-diagnostic.json`, and `browser-agent-select-diagnostic.json`. Native traces, credentials and page dumps were not recorded. Original browser creation intents and failed results are retained. Both dedicated test accounts and their resources were removed afterward, and admin reads verified account absence. A zero-total-agents cleanup diagnostic was corrected: verification automatically creates a starter agent, so total count alone is not a valid run-ownership assertion.

jhgaylor:
Fresh accounts render a credential form inside the agent form, so HTML parsing separates the later provider and Save controls from their intended form. Render the credential card before the agent form, give the latter a stable ID, and remove credential-seeding workarounds from its tests.

Closes #1660. Two files, +81/-26.

Validation: the fresh-account regression fails on the parent and all 25 agent LiveView tests pass with the fix. A real Playwright 1.63.0 / Chromium 1243 run, after LiveView connected, filled runtime/model/provider fields, exercised empty credential validation, confirmed every field remained intact, saved the agent, and reached the agent list. A database read confirmed the saved configuration. No inference key was seeded; the disposable browser account was removed after verification.

Full local precommit passes: 5,297 core tests plus 6 doctests, 144 Buzz tests, and 33 Support tests, with no failures. The first full run was invalidated by the committed browser fixture affecting account-count assertions; it was removed before this clean run.

### DIFF
--- a/apps/fountain/lib/fountain_web/live/agents_live/form.ex
+++ b/apps/fountain/lib/fountain_web/live/agents_live/form.ex
@@ -729,7 +729,14 @@ defmodule FountainWeb.AgentsLive.Form do
         </.link>
       </div>
 
+      <.missing_credential_card
+        :if={@missing_credential}
+        missing={@missing_credential}
+        message={@credential_message}
+      />
+
       <form
+        id="agent-form"
         phx-change="validate"
         phx-submit="submit"
         class="space-y-4 bg-white rounded shadow p-6 border border-zinc-200"
@@ -803,11 +810,6 @@ defmodule FountainWeb.AgentsLive.Form do
         >
           Not one of the models Fountain lists — it will be passed to the runtime as-is.
         </p>
-        <.missing_credential_card
-          :if={@missing_credential}
-          missing={@missing_credential}
-          message={@credential_message}
-        />
 
         <div :if={length(@sandbox_providers) > 1} class="space-y-1">
           <label class="block text-sm font-medium text-zinc-700">Sandbox provider</label>
--- a/apps/fountain/test/fountain_web/live/agents_live_test.exs
+++ b/apps/fountain/test/fountain_web/live/agents_live_test.exs
@@ -257,7 +257,7 @@ defmodule FountainWeb.AgentsLive.IndexTest do
       refute html =~ "credential on this account yet"
 
       view
-      |> form("form", %{
+      |> form("#agent-form", %{
         "agent" => %{
           "name" => "converger",
           "runtime" => "acp",
@@ -273,7 +273,7 @@ defmodule FountainWeb.AgentsLive.IndexTest do
     end
 
     test "switching an acp agent onto a model runtime drops the command", %{conn: conn} do
-      user = with_credential(insert_verified_user())
+      user = insert_verified_user()
 
       agent =
         insert_agent(user_id: user.id, runtime: "acp", runtime_command: "chant acp")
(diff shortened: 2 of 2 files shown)


BinaryBourbon (commented):
Reviewed `81f8de00bb2e1dddf113cc1c5b0063b7a4abd8fc` for #1660. No actionable findings identified.

Inspected the sibling form markup, both submit handlers, validation state preservation, selector updates and the fresh-account regression. Moving the credential card before the stable-ID agent form removes the nested form without changing the save handlers.

Limitations: Static review only; did not rerun LiveView tests or browser validation. The browser evidence in the PR description was not independently reproduced.

BinaryBourbon (approved):
Approve `81f8de00bb2e1dddf113cc1c5b0063b7a4abd8fc`. The credential form is now a sibling before the stable-ID agent form. Rechecked both handlers and field bindings: agent state is retained during credential validation, and provider/save controls remain inside the intended form. The fresh-account regression removes the former credential-seeding workaround.

Validation: Verified the exact-head six test partitions, static analysis, release/API-contract and aggregate CI all pass. Inspected the regression assertions and CI test execution logs. Full CI passed for this exact head: https://github.com/managoat/fountain/actions/runs/34596886127.

Limits: I did not independently rerun a browser; browser evidence remains the author’s recorded validation. No browser-dependent defect was found in the source review.

Issue #1934, opened 2026-09-11
Flake: SSE heartbeat stream exits before the idle-window assertion


jhgaylor:
The SSE heartbeat test failed locally while validating the unrelated #1691 dashboard change, then passed without changing the stream or test code.

- Test: `FountainWeb.SseStreamTest`, `heartbeats and the idle exit heartbeats hold the connection open past the idle timeout`, `apps/fountain/test/fountain_web/controllers/sse_stream_test.exs:311`.
- Assertion at line 332: `assert Process.alive?(task.pid), "the loop exited despite heartbeats"`; the task was no longer alive.
- Observed rate: one failure in one full-suite run on tree `a1e62a37` (parent `9a8b4580`), seed `246102`. The complete SSE file then passed 18/18 tests with that seed and unchanged code.
- Environment: local macOS arm64, pinned Erlang 28.3 / Elixir 1.19.2, dedicated PostgreSQL database. Several independent worktrees were running full suites or compiling concurrently. The full run used `max_cases: 20`; the focused run used `ERL_FLAGS='+S 4:4'` and `max_cases: 8`. No stream/test changes or external-provider dependency changes between runs. The branch changes only a dashboard expression, its Prometheus fixtures, and workflow paths.
- Local evidence: `/private/tmp/fountain-p2-1691-precommit-loaded.log` (original failing run retained before the clean rerun), `/private/tmp/fountain-p2-heartbeat-focused.log`. No CI URL: this was a local run.

The test schedules a heartbeat every 50 ms, sets an idle timeout of 200 ms, sleeps 700 ms, then checks that the task remains alive. It does not synchronize on the stream having entered its receive loop or capture the completed task's response when it exits. Scheduler delay under load is a hypothesis, not an established cause; the response/exit reason needs recording to distinguish a timer gap from an early response or another exit.

Searched all existing `flake` issues before filing. #1909 also mentions a heartbeat but concerns runner-daemon SQL Sandbox teardown, not this SSE test. Recording separately per CLAUDE.md's flaky-test policy; no timing-test change is included in the dashboard PR.

Issue #1691, opened 2026-09-07
Finance dashboard and the email-failure alert read fountain_email_delivery_exception_count, which nothing emits


jhgaylor:
## Symptom

The email-failure half of the finance dashboard is a permanent zero, and the hosted `FountainEmailDeliveryFailures` alert can never fire.

## Mechanism

- Emitter: `counter("fountain.email.delivery.exception")` at `apps/fountain/lib/fountain_web/telemetry.ex:296`. `TelemetryMetricsPrometheus.Core` adds no suffix, so the series is `fountain_email_delivery_exception`. The sibling `fountain.email.delivery.count` renders as `fountain_email_delivery_count`, which the dashboard gets right.
- `deploy/grafana/fountain-finance.json:853` (#1521) references `fountain_email_delivery_exception_count`; the `or vector(0)` hides the miss.
- home-cloud `platform/fountain-site/prometheusrule.yaml:790-791` uses the same name and adds it to the error half without `or vector(0)`, so the empty right operand empties the whole expression.

## Fix

Rename in both places, and add a fixture under `scripts/test-alerts.py` that would have caught the absent series. Same family as the four traps in the billing-alert notes (#1169).

---
Found in the 2026-09-07 review of the week on `main` (cb93a7ea).

PR #1937, opened 2026-09-11
fix: count emitted email exceptions in the finance dashboard

jhgaylor:
The finance dashboard reads an email-exception counter that Fountain never emits, so exceptions display as zero. Point it at `fountain_email_delivery_exception` and evaluate the actual dashboard expression in the existing promtool fixture suite. Dashboard edits now trigger that CI job.

Part of #1691. The separate hosted-alert correction is jhgaylor/home-cloud#224; both parts are required to resolve the issue. Three files, +31/-2.

Validation: 49 Prometheus expression fixtures pass, including errors only, exceptions only, both, and no failure counters. The new cases fail against the parent dashboard. An unchanged SSE heartbeat test failed during a concurrent full run and passed all 18 focused tests with the same seed; evidence is tracked in #1934. Full local precommit rerun passes with four VM schedulers: 5,316 core tests plus 6 doctests, 144 Buzz tests, and 33 Support tests, with no failures.

### DIFF
--- a/.github/workflows/alert-rules.yml
+++ b/.github/workflows/alert-rules.yml
@@ -4,12 +4,14 @@ on:
   pull_request:
     paths:
       - deploy/k8s/prometheusrule.yaml
+      - deploy/grafana/fountain-finance.json
       - scripts/test-alerts.py
       - .github/workflows/alert-rules.yml
   push:
     branches: [main]
     paths:
       - deploy/k8s/prometheusrule.yaml
+      - deploy/grafana/fountain-finance.json
       - scripts/test-alerts.py
       - .github/workflows/alert-rules.yml
   workflow_dispatch:
--- a/deploy/grafana/fountain-finance.json
+++ b/deploy/grafana/fountain-finance.json
@@ -850,7 +850,7 @@
       "targets": [
         {
           "refId": "A",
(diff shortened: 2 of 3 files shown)


BinaryBourbon (commented):
Reviewed `a1e62a37087cae235d553cc677daee4ee10f1e83` for #1691. No actionable findings identified.

Checked the corrected metric against the telemetry emitter and reviewed the actual dashboard expression fixtures. All 49 Prometheus expression cases pass in an isolated checkout, including errors only, exceptions only, both and neither.

Limitations: The hosted alert correction is explicitly tracked separately in home-cloud #224 and was not reviewed here.

BinaryBourbon (approved):
Approve `a1e62a37087cae235d553cc677daee4ee10f1e83`. The dashboard now reads the emitter’s actual metric, while preserving zero defaults for absent error/exception series. Fixtures evaluate the dashboard expression itself in all four presence combinations.

Validation: Local: all 49 Prometheus cases pass. Independently exercised the installed exporter and verified it emits fountain_email_delivery_exception without _count. Exact-head alert workflow also passes. Full CI passed for this exact head: https://github.com/managoat/fountain/actions/runs/34598182067.

Limits: Hosted alert changes in home-cloud #224 are outside this review.

PR #1942, opened 2026-09-11
fix(api): separate shared-ingress and per-key rate limits

### LINKED ISSUE #1771: Prevent shared-ingress IP limits from defeating per-key API isolation
Found during review of #1722 at 7ceb07b600613e3650b8159897234a7e06313dac.

The API pipeline runs a 600/min per-IP limiter before authentication, then the new 600/min per-key limiter. When multiple clients share an ingress address, the first limiter still rejects all of them once the address reaches 600 requests. For example, key A makes 600 requests through that address, then key B's first request is rejected before authentication. The per-key limiter cannot isolate B from A in this scenario.

The new isolated plug test proves separate key buckets, but does not exercise both limiters in pipeline order. The shared-address ceiling is pre-existing; #1722's claim that the per-key addition resolves shared-ingress interference is not achieved by this arrangement.

Acceptance criteria:
- Define a policy separating coarse pre-auth abuse protection from authenticated per-key quotas, with an intentional aggregate ceiling.
- Verify trusted proxy/client address handling; do not trust arbitrary forwarded headers.
- Add a pipeline-level regression with two keys sharing an address, confirming a noisy key is limited while the quiet key retains its intended allowance.
- Preserve limits on unauthenticated requests and document per-replica behavior and any intentional aggregate limit.

Do not simply remove pre-auth protection. Scope this as the shared-ingress isolation follow-up to #1722.

jhgaylor:
The pre-auth address limit was the same size as one API key's quota, so one client behind a shared ingress could exhaust the address before another client's first request. Separate the budgets: 600 authenticated requests per key per minute, 600 authentication refusals per client address, and an intentional 6,000-attempt pre-auth address ceiling. The aggregate includes rejected attempts and can still block all keys at the address when exhausted. All counters remain per replica.

Closes #1771. Four files, +117/-27. Document both the aggregate tradeoff and trusted-proxy behavior in the API guide.

Validation: the actual router-pipeline regression fails on the parent with a quiet key receiving 429. All 53 focused pipeline, authentication, limiter, and client-IP tests pass with the fix, covering key quota boundaries, failed/missing/unverified authentication, aggregate exhaustion, and forwarding accepted only from a trusted proxy. All three prose reports reviewed: existing docs-style findings, advisory Vale findings, and clean destink. Full local precommit passed: 5,372 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain_web/plugs/tenant_api_auth.ex
+++ b/apps/fountain/lib/fountain_web/plugs/tenant_api_auth.ex
@@ -7,6 +7,10 @@ defmodule FountainWeb.Plugs.TenantAPIAuth do
   Updates `last_used_at` in an unlinked task under `Fountain.TaskSupervisor`,
   so the stamp never blocks the request and cannot take it down.
 
+  Authentication refusals share a 600/minute per-address budget, separately
+  from the pipeline's coarse pre-auth ceiling and authenticated per-key quota.
+  A full refusal budget returns 429 without blocking a valid key.
+
   Returns 401 JSON on failure. The response body includes a machine-readable
   `reason` so clients (especially in-sprite agents holding a rotated
   `$FOUNTAIN_TOKEN`) can tell a revoked key apart from one that never existed:
@@ -25,6 +29,7 @@ defmodule FountainWeb.Plugs.TenantAPIAuth do
   require Logger
 
   alias Fountain.Accounts
+  alias FountainWeb.Plugs.RateLimit
 
   def init(opts), do: opts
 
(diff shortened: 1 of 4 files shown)


BinaryBourbon (commented):
Reviewed `4c15a6eb5927f1ea525667ba2750bfc120004004` for stack #1771.

No actionable defect introduced by this PR found. Reviewed the pipeline order, refusal path, bucket keys, and regressions. The documented 6,000/address aggregate ceiling intentionally still bounds all traffic; the 600/address failure budget limits refusal responses after authentication, while valid keys keep their 600/key budget.

Validation: static review at the exact head, including proxy-handling assertions and existing limiter implementation; pipeline tests were not rerun.

BinaryBourbon (approved):
Approve `4c15a6eb5927f1ea525667ba2750bfc120004004`.

The revised pipeline implements the issue’s intended policy: 6,000 attempts/address before authentication, 600 requests/key after authentication, and a separate 600/address refusal budget. The tests exercise two keys through the actual pipeline and trusted versus untrusted forwarding. The aggregate ceiling remains an explicit limit, and the refusal budget is deliberately checked after authentication. No actionable defect found.

Rechecked current CI and review discussions. [CI run 34601367641](https://github.com/managoat/fountain/actions/runs/34601367641) is green, including all six test partitions, static analysis and the required gate. Logs identify tested merge `1cb62c71` of this head into `3f59a76a4f63b3d445a8b71e5e33209f8a47bdd4`. The log confirms PreAuthRateLimitTest ran in partition 1. No additional local pipeline run.

PR #1943, opened 2026-09-11
fix(conversations): clean up retired provisioning attempts

jhgaylor:
A sandbox retired while provisioning could reject the final ready write, raise a MatchError, and then mark the conversation failed—even after it had moved to a replacement sandbox. Handle that retirement rejection explicitly: destroy the handle this attempt created, revoke only its prepared broker session, and stop normally with its callback-key state. Leave the retired row and replacement conversation untouched, and emit no provision success or new turn.

First small unit of #1766: provisioning completion only. Wake completion follows in a separate stacked PR; admission before provisioning remains separate. Two files, +117/-11.

Validation: deterministic provider pauses reproduce both terminated and failed retirement on the parent (2 tests, 2 failures). The fix preserves replacement sandbox/session rows, revokes the stale callback key and broker token, destroys only the stale handle, and never starts the queued prompt. All 34 focused broker, fingerprint, provision deadline and callback-revocation tests pass, including an unrelated database rejection that still fails provisioning. Full local precommit passed: 5,378 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -963,17 +963,15 @@ defmodule Fountain.Conversations.ConversationServer do
                  state.runtime_module,
                  agent,
                  sprite_env
-               ) do
-          # `build_fingerprint` records what the disk was built from, and
-          # `applied_skills` what was mounted on it, so a later reapply can
-          # answer both questions from the row rather than guessing (#1565).
-          {:ok, _} =
-            Conversations.update_sandbox(sandbox, %{
-              status: "ready",
-              build_fingerprint: Reapply.fingerprint(env),
-              applied_skills: skills
-            })
-
+               ),
+             # Record what the disk was built from only if this attempt still
+             # owns a live row. Retirement can win while provider I/O runs.
+             {:ok, _} <-
+               Conversations.update_sandbox(sandbox, %{
+                 status: "ready",
+                 build_fingerprint: Reapply.fingerprint(env),
+                 applied_skills: skills
+               }) do
           Output.publish_stage(state.conversation_id, "provision", "done")
 
           # Best-effort: snapshot the fully-provisioned state so subsequent
@@ -997,6 +995,14 @@ defmodule Fountain.Conversations.ConversationServer do
           # queue_initial_prompt/3.
           {:noreply, new_state}
         else
+          {:error, %Ecto.Changeset{errors: [status: {"sandbox is retired", []}]}} ->
+            # This handle and token belong to this attempt. Do not fail the
+            # conversation or release every session: a replacement may own it.
+            _ = Managoat.Sandbox.destroy(handle)
+            Egress.release_prepared(prepared)
+            {:ok, prepared_state} = prepared
+            {:stop, :normal, prepared_state}
+
           {:error, reason} ->
             Logger.error("provision step failed: #{inspect(reason)}")
             _ = Managoat.Sandbox.destroy(handle)
--- a/apps/fountain/test/fountain/conversations/conversation_server_broker_test.exs
+++ b/apps/fountain/test/fountain/conversations/conversation_server_broker_test.exs
@@ -466,6 +466,106 @@ defmodule Fountain.Conversations.ConversationServerBrokerTest do
       end
     end
 
+    for terminal <- ["terminated", "failed"] do
+      @tag terminal: terminal
+      test "retirement during provision preserves the #{terminal} row and replacement", %{
+        user: user,
+        agent: agent,
+        terminal: terminal
+      } do
+        conv = insert_conversation(user_id: user.id, agent: agent, sandbox_api_access: "owner")
+        sandbox = Conversations._unsafe_get_sandbox!(conv.sandbox_id)
+        test = self()
+        handle = stub_happy_sprite(sandbox.sprite_name)
+        stub(Fountain.Broker, :preflight, fn -> :ok end)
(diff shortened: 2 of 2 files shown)


BinaryBourbon (approved):
APPROVE — the guard covers the write it claims, and it fixes a sprite leak on the way past.

**What this changes**

The final `status: "ready"` write moves from a bare `{:ok, _} =` match into the provisioning `with` chain, so a retirement that lands while provider I/O is in flight arrives as a value instead of a `MatchError`. The new `else` clause treats that one rejection as "this attempt lost": destroy the sprite this attempt created, release only this attempt's prepared broker session, and stop `:normal` carrying the *prepared* state so `terminate/2` revokes the callback key this attempt minted. No `provision/done`, no `sandbox` row write, no `conversation` status write.

Direction protected: **revival** — a retired row can no longer be flipped back to `ready` and then used, and the machine this attempt built is destroyed rather than left billing. The opposite direction (a losing attempt left attached, or a conversation wedged waiting on a machine that will never arrive) is not opened: the server stops, the conversation is left `idle` pointing at whatever replacement owns it, and `CallbackKey.revoke/2`'s row check keeps the dying attempt from revoking a successor's live key.

**Findings**

1. `conversation_server.ex:998` — non-blocking, brittleness. The clause matches the changeset's `errors` list *exactly*. If `Sandbox.changeset/2` ever adds a second error to this write (it carries `build_fingerprint` and `applied_skills`, not just `status`), the retirement falls through to the generic arm at :1006, which writes `failed` to the already-terminal row and then marks the **conversation** `failed` — clobbering the replacement the retirement handed it to. Prefer membership (`Keyword.get(cs.errors, :status)` / a shared predicate) over list equality. #1971 later hoists this into a `retired_or_resetting/1` defguard but keeps the exact-list comparison, so the fix belongs there or here, once.

2. `conversation_server.ex:877-` (stage stream) — non-blocking, worth an answer. `do_fresh_provision_inner` has already published `provision/started` by the time this clause runs, and the clause deliberately publishes neither `done` nor `failed`. The conversations app pairs a stage's `started` with its terminal state, so this leaves an open `provision` block on the transcript that only the *next* attempt's events will close, ambiguously. Publishing a terminal state for the retired attempt (or saying in the comment why silence is right for this stage) would settle it. Note #1945 avoids the question entirely by stopping before any `started` event, which is the cleaner shape.

3. The new "unrelated ready-write error" test at `conversation_server_broker_test.exs:544-` — non-blocking, test gap. It passes on the parent too: the old `MatchError` was caught by `do_fresh_provision`'s rescue, which also published `provision/failed` and set both rows `failed`, so every assertion in that test holds before the change. The thing that actually changed for an unrelated error is that the sprite and the broker session are now released (the rescue path destroyed neither, so an unrelated ready-write failure previously leaked a live sprite). Add `assert_receive {:destroyed, ^handle}` and a session-lookup assertion and it pins the improvement instead of the status quo.

4. `conversation_server.ex:1003` — nit. `{:ok, prepared_state} = prepared` is safe (this clause is only reachable after `prepared` matched `{:ok, _}` at the head of the `with`), but it reads like a second unguarded match in a patch about unguarded matches. One clause comment saying so would help the next reader.

5. Merge hazard, not a defect in the patch: this branch carries `@pin 2762` from its merge base, but `main` has since lowered the size pin to **2646** with zero headroom (`conversation_server_size_test.exs`). This branch's `conversation_server.ex` is 2691 lines; rebased onto current `main` it lands at ~2652 and the size test goes red. The same applies to #1944 and #1945 (and the #1970/#1971 pair above them). The green run on this PR predates that pin change. Somebody has to decide between raising the pin once for the programme or moving these clauses into a `Fountain.Conversations.*` module — the pin's own moduledoc predicted exactly this ("left zero headroom, and went red two rounds later").

6. No `CHANGELOG.md` entry anywhere in this programme. Mentioning once, not as a blocker — these are internal race repairs.

**Checked and clean**

- `Egress.release_prepared(prepared)` releases only this attempt's token (#1827 scoped it), and the test proves a concurrently-prepared replacement session survives.
- Stopping with `prepared_state` rather than `state` is the right choice: it is the state that holds `callback_api_key_id`, and `CallbackKey.revoke/2` re-reads the conversation row so a successor's key is never taken.
- Destroying the handle here is correct and is *not* contradicted by #1944 declining to destroy: this handle is a sprite this attempt just created, whereas the wake path's handle can be a home shared by co-tenants (ADR 0023).
- `record_sandbox_url/2` earlier on this path still writes to a retired row, but only `sandbox_url`, which the revival fence deliberately lets through; harmless.
- The retirement × {terminated, failed} tests bite if reverted (the rescue path destroys nothing and marks the conversation `failed`, so both the `{:destroyed, handle}` and the `conv.status == "idle"` assertions fail).
- `reject(Managoat.Sandbox.Sprites, :spawn, 4)` is the right fence for "the queued prompt never ran" — `create/2` is the machine, `spawn/4` is the process.
- No tenant-scoping, audit-in-transaction, `Task.async` or `async: false` issues; no new `_unsafe_` call without an established owner.

jhgaylor:
⚠️ **Before merging this chain:** the campaign it starts lands ~72 lines over the `conversation_server.ex` size pin, and no PR in it is red. See #2032.

`main` is now at `@pin 2646` with the file at exactly 2646 lines (zero headroom) after #1749 and #1751. Every PR in this stack is measured against a base where the pin is still 2762, so each is green on its own; merging the chain tip onto current `main` auto-merges cleanly with no conflict on the pin line, then fails at 2673 > 2646.

Nothing warns you at merge time, which is why this note is here rather than only on the tracker. #2032 has the per-branch arithmetic and the two options.

PR #1944, opened 2026-09-11
fix(conversations): detach wakes that lose to sandbox retirement

jhgaylor:
A wake that lost to sandbox retirement either raised on the suspended-to-ready write or skipped the guard entirely when its cached row was already ready. Validate the persisted row after provider preparation in both cases. On retirement rejection, release only this wake's broker token and stop normally with its callback-key state before inspecting or attaching runtime sessions. The wake never destroys the existing disk.

Second small unit of #1766, stacked on #1943 (provisioning completion). Two files, +149/-26 relative to that base. Starting and park transitions are being audited separately; this PR addresses wake completion.

Validation: all four deterministic ready/suspended × failed/terminated race cases fail on the parent and pass with the fix. Replacement sandbox/session rows survive, stale credentials are revoked, and no session listing, spawn, destroy, or successful wake event follows the rejection. Ordinary ready and suspended wakes still succeed; only suspended wakes reset last_resumed_at. An unrelated database rejection retains its error behavior. All 55 focused broker, lifetime, shared-sandbox and callback-revocation tests pass. Full local precommit passed: 5,385 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -1176,39 +1176,46 @@ defmodule Fountain.Conversations.ConversationServer do
         {:error, reason} -> Logger.warning("runtime prepare on wake: #{inspect(reason)}")
       end
 
-      # Normally the wake path already flipped suspended → ready under the
-      # quota reservation; this covers the reaper parking the row mid-wake.
-      # Without it the row would stay `suspended` under a live server —
-      # invisible to the quota and unreachable by any reaper pass.
-      sandbox =
+      # Validate even a cached ready row: retirement may have won while the
+      # provider was waking. Only a suspended wake resets the lifetime clock.
+      attrs =
         if sandbox.status == "suspended" do
-          {:ok, s} =
-            Conversations.update_sandbox(sandbox, %{
-              status: "ready",
-              last_resumed_at: DateTime.utc_now() |> DateTime.truncate(:second)
-            })
-
-          s
+          %{
+            status: "ready",
+            last_resumed_at: DateTime.utc_now() |> DateTime.truncate(:second)
+          }
         else
-          sandbox
+          %{status: "ready"}
         end
 
-      new_state = %{
-        state
-        | handle: handle,
-          sprite_env: sprite_env,
-          sandbox_started_at: Lifecycle.clock_start(sandbox)
-      }
+      case Conversations.update_sandbox(sandbox, attrs) do
+        {:ok, sandbox} ->
+          new_state = %{
+            state
+            | handle: handle,
+              sprite_env: sprite_env,
+              sandbox_started_at: Lifecycle.clock_start(sandbox)
+          }
 
-      new_state = reattach_running_turn(%{new_state | current_turn: nil})
+          new_state = reattach_running_turn(%{new_state | current_turn: nil})
 
-      new_state =
-        Reattachment.finish_runner_reconnect(
-          new_state,
-          if(new_state.current_turn, do: "reattached", else: "turn_ended")
-        )
+          new_state =
+            Reattachment.finish_runner_reconnect(
+              new_state,
+              if(new_state.current_turn, do: "reattached", else: "turn_ended")
+            )
 
-      {:noreply, new_state}
+          {:noreply, new_state}
+
+        {:error, %Ecto.Changeset{errors: [status: {"sandbox is retired", []}]}} ->
+          # Wake owns this connection's credentials, not the existing disk or
+          # another connection's session. Never destroy the machine here.
+          Egress.release_prepared({:ok, state})
+          {:stop, :normal, state}
+
+        error ->
+          raise MatchError, term: error
+      end
     else
       {:error, :not_found} ->
         # The provider says the sandbox is gone. That is the one answer that
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -963,17 +963,15 @@ defmodule Fountain.Conversations.ConversationServer do
                  state.runtime_module,
                  agent,
                  sprite_env
-               ) do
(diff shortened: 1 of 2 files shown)


BinaryBourbon (approved):
APPROVE — this closes the real hole in the stack (the cached-`ready` wake had no guard at all), with one regression that #1970 fixes two stages later. Read the merge-order note before landing it.

**What this changes**

`do_reattach` used to write `ready` only when its cached row said `suspended`, and matched `{:ok, s} =`. Now it always validates the persisted row: `%{status: "ready", last_resumed_at: now}` for a suspended wake, `%{status: "ready"}` otherwise, and it branches on the result. On the retirement rejection it releases only this wake's broker session and stops `:normal` **before** `reattach_running_turn/1` and `finish_runner_reconnect/2`, so no runtime session is adopted; it deliberately does not destroy the disk.

Both directions matter here, and they split by the row's cached status:

- **Cached `ready`**: the old code wrote nothing, so a wake that lost to retirement *succeeded* — it reattached to a retired machine, finished the reconnect, and ran the queued prompt. That is the leaked-paid-machine direction, and this patch is the only thing in the stack that closes it. The `initial: "ready"` variants of the new test are the ones that matter; they fail on the parent by actually running work (`reject(Sprites, :spawn, 4)` trips).
- **Cached `suspended`**: the old code raised `MatchError`. Now the wake detaches, which is the "don't leave a conversation waiting on a machine that will never arrive" direction — the server stops, the row stays terminal, the conversation stays `idle` on its replacement.

**Findings**

1. `conversation_server.ex:1210-1217` — **blocking if this PR reaches `main` ahead of #1970; non-blocking as a stacked stage.** Writing `ready` unconditionally means a *cached-ready* row now goes through `update_sandbox/2`'s **reset** fence as well as its revival fence. `Conversations.update_sandbox/2` rolls back with `{:error, :sandbox_reset_pending}` when `reset_requested_at` is set and the new status is non-terminal, and this patch routes that to `raise MatchError, term: error`. Concrete scenario: `reset_sandbox/2` commits the fence (row stays `ready`, `reset_requested_at` set) and the provider destroy then fails — the docstring is explicit that this leaves the fence in place with "no automatic reconciliation", and the API wake path refuses with `:sandbox_reset_pending` rather than clearing it. A Horde rebalance or a deploy restarts that conversation's server, `do_reattach` runs, and the write is refused: the actor crashes, the supervisor restarts it, and it crashes again — a restart loop on the exact failure mode #1766 exists to remove. Before this patch that path wrote nothing and did not crash. #1970 (`codex/p2-1767-wake-reset-fence`) replaces this clause with `{:error, reason} when retired_or_resetting(reason)`, which fixes it, so per the review rules I am not blocking; but #1944 must not land on `main` without #1970 behind it. If the reset PRs slip, fold that single extra clause in here. Note #1946 — the last stage of *this* PR's own stack — already handles `{:error, :sandbox_reset_pending}` explicitly in `park_row/1`, so the shape was known before the stack ended.

2. `conversation_server.ex:1210` — non-blocking. Stopping before `reattach_running_turn/1` means `state.current_turn` is still `nil`, so `terminate/2`'s `TurnMachine.orphan_on_normal_stop/3` orphans nothing and a turn row left `running` by the previous server life stays `running`. Unlike the `:noop` fences elsewhere in the programme this one *is* reconciled — `AutonomousTurnReaper` sweeps running turns with no live server (30 min stuck + 15 min quiet) and `_unsafe_orphan_turn/3` flips the conversation back to `idle` — but the user's transcript stays open for up to 45 minutes. `find_running_turn(state.conversation_id)` is already in scope on this path (the transient arm below uses it); orphaning it here would close the transcript at once. Worth a sentence in the clause comment either way, since "nothing sweeps this" was round 1's assumption (#2021 item 5) and it is not true for this exit.

3. `conversation_server.ex:1210` — non-blocking, same as #1943 finding 2: `reattach/started` was published before the provider work, and this clause publishes no terminal state for that stage, leaving an unpaired stage block on the transcript. The new test asserts `refute ... state == "done"`, which is right, but does not say what the client is supposed to render instead.

4. `conversation_server.ex:1210` — non-blocking, brittleness: the exact `errors: [status: {"sandbox is retired", []}]` list match, as on #1943. #1970 keeps the exact comparison inside the defguard; fix it once, there.

5. Non-blocking, worth stating in the comment: the "never destroy the machine here" decision is right (the wake's handle can be a home shared by co-tenants, ADR 0023) but it leans on the retiring writer owning the destroy. `_unsafe_reap_sandbox/1`, `Lifecycle.destroy/5` and the `{:error, :not_found}` arm all satisfy that; the comment currently justifies it by credential ownership rather than by "the winner destroys".

6. Merge hazard (also on #1943/#1945): this branch carries `@pin 2762`; `main` now pins **2646** with zero headroom and this file is 2698 lines. Rebasing onto current `main` turns the size test red. The green CI on this PR predates the pin change.

**Checked and clean**

- The lifetime rule is preserved exactly: `last_resumed_at` is only sent on the suspended branch, and the new "ordinary wake" tests assert both halves (unchanged for `ready`, advanced for `suspended`). That was the easy thing to get wrong here.
- `Lifecycle.clock_start(sandbox)` now reads the row `update_sandbox/2` returned rather than the cached struct — strictly fresher, and the absolute-lifetime ceiling still dates from the row.
- The no-op `ready → ready` write costs one `FOR UPDATE` and no `UPDATE` (no changes in the changeset), emits no usage event (`record_sandbox_usage/2` only fires on `pending|starting → ready` and `suspended → ready`) and pokes no queue (`was` and `updated` are both active). No double-billing.
- `Egress.release_prepared({:ok, state})` releases the rebound inner `state`, i.e. this wake's own token; the test proves a concurrently-prepared replacement session survives and that the stale callback key is revoked while the conversation row still names it.
- `error -> raise MatchError` preserves crash-on-unknown-error, and the third new test pins it with a rejection that is deliberately *not* the retirement shape.
- The four-way `initial × terminal` matrix all bite on the parent, for two different reasons depending on `initial` — that is the right coverage for this change.

PR #1945, opened 2026-09-11
fix(conversations): stop retired attempts before provisioning starts

jhgaylor:
If retirement won just before an actor marked its cached sandbox starting, the rejected write was caught as a provisioning failure and could overwrite the terminal state or fail a replacement conversation. Handle that specific rejection before entering provisioning: stop normally, create no resources and publish no start or failure event. Preserve the original snapshot for interrupted-attempt cleanup and the existing error path for other failures.

Third small unit of #1766, stacked on #1944 (wake completion). Two files, +66/-5. The park caller audit remains a separate unit.

Validation: two deterministic tests pause just before the real starting write, retire to terminated/failed, move the conversation to a replacement sandbox, and release the actor. Both fail on the parent and pass with the fix; no provider create/destroy or broker preparation occurs. All 40 focused retirement, provision-deadline, broker and interrupted-provision tests pass. Full local precommit passed: 5,387 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -1176,39 +1176,46 @@ defmodule Fountain.Conversations.ConversationServer do
         {:error, reason} -> Logger.warning("runtime prepare on wake: #{inspect(reason)}")
       end
 
-      # Normally the wake path already flipped suspended → ready under the
-      # quota reservation; this covers the reaper parking the row mid-wake.
-      # Without it the row would stay `suspended` under a live server —
-      # invisible to the quota and unreachable by any reaper pass.
-      sandbox =
+      # Validate even a cached ready row: retirement may have won while the
+      # provider was waking. Only a suspended wake resets the lifetime clock.
+      attrs =
         if sandbox.status == "suspended" do
-          {:ok, s} =
-            Conversations.update_sandbox(sandbox, %{
-              status: "ready",
-              last_resumed_at: DateTime.utc_now() |> DateTime.truncate(:second)
-            })
-
-          s
+          %{
+            status: "ready",
+            last_resumed_at: DateTime.utc_now() |> DateTime.truncate(:second)
+          }
         else
-          sandbox
+          %{status: "ready"}
         end
 
-      new_state = %{
-        state
-        | handle: handle,
-          sprite_env: sprite_env,
-          sandbox_started_at: Lifecycle.clock_start(sandbox)
-      }
+      case Conversations.update_sandbox(sandbox, attrs) do
+        {:ok, sandbox} ->
+          new_state = %{
+            state
+            | handle: handle,
+              sprite_env: sprite_env,
+              sandbox_started_at: Lifecycle.clock_start(sandbox)
+          }
 
-      new_state = reattach_running_turn(%{new_state | current_turn: nil})
+          new_state = reattach_running_turn(%{new_state | current_turn: nil})
 
-      new_state =
-        Reattachment.finish_runner_reconnect(
-          new_state,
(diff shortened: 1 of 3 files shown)


BinaryBourbon (approved):
APPROVE — the cleanest stage in the stack: the only one where losing the race costs nothing, because nothing has been built yet.

**What this changes**

The `pending → starting` write moves out of `do_fresh_provision_inner/6` and up into `do_fresh_provision/6`, where its result is inspected instead of matched. A retirement that has already won is answered by stopping `:normal` before `create_sandbox_handle`, before the callback-key rotation, before the broker `prepare`, and before the `provision/started` event. `interrupted?` still reads the pre-write snapshot, which is what it read before.

Direction protected: **the machine leak**, at the earliest point it can be protected. A retired row can no longer cause a sprite to be created at all, so there is nothing to destroy, no token to release and no stage to close. The other direction is not opened: no resources exist to strand, the retired row keeps its `terminated_at`, and the conversation is left `idle` on its replacement. This is also the stage that closes #1943's residual window where a losing attempt would still have minted a callback key and written its id onto a conversation row that now belongs to a replacement.

**Findings**

1. `conversation_server.ex:853` — non-blocking, and pre-existing rather than introduced: `{:error, :sandbox_reset_pending}` from the reset fence lands on `raise MatchError`. The old code's `{:ok, _} =` match raised too, so this is not a regression, and #1971 (`codex/p2-1767-provision-reset-fence`) widens this exact clause to `retired_or_resetting(reason)`. Flagging only so the stack's merge order is deliberate.

2. `conversation_server.ex:847` — non-blocking: exact-list match on the changeset errors, as on #1943 and #1944. One shared predicate for all three sites, please, rather than three copies of a literal that a future validation on the same write would silently defeat.

3. `conversation_server.ex:873-878` — nit. The comment rewrite ("Finding the original snapshot already `starting`") gives up a true invariant. The row still only becomes `starting` in this one place; the write just moved one frame up the call stack. "The row only ever becomes `starting` in `do_fresh_provision/6` above, so finding the snapshot already `starting` means…" keeps the fact and the new location.

4. Merge hazard (as on #1943/#1944): this branch carries `@pin 2762`, `main` now pins **2646** with zero headroom, and this file is 2707 lines. Rebased onto current `main` the size test goes red. The green run predates the pin change.

**Checked and clean**

- The `case` is inside the existing `try`, so `do_fresh_provision_inner/6` keeps its rescue and an unexpected raise still publishes `provision/failed` and fails both rows.
- `interrupted? = sandbox.status == "starting"` is unchanged in value: it read the pre-write struct before and reads the same pre-write struct now. The interrupted-attempt discard path is not weakened.
- The struct passed into `do_fresh_provision_inner/6` is now one generation stale relative to the row, but it was equally stale before (the old code discarded the updated row), and every later write re-reads under `update_sandbox/2`'s row lock.
- Both callers of `fresh_provision/6` (the `pending|starting` clause and the catch-all in the status `case`) funnel through `do_fresh_provision/6`, so the fence is at the single choke point; no second path reaches the `starting` write.
- `reject(Managoat.Sandbox.Sprites, :create, 2)` is the right fence for "no machine was created" — `create/2` is the sandbox, `spawn/4` is a process inside one — and `reject(Fountain.Broker, :prepare, 4)` covers the credential half.
- `refute Enum.any?(..., &(&1.stage == "provision"))` is the strongest available assertion for this stage, and the tests bite on the parent: reverted, the `{:ok, _} =` match raises, the rescue publishes `provision/failed`, overwrites the retired row's status and marks the conversation `failed`, so four assertions fail.
- The new test file is a `Fountain.ConversationServerCase` module with `async: true` inherited, no `_unsafe_` misuse (`_unsafe_get_sandbox!` follows the scoped conversation insert), and no audit or `Task.async` concerns.

PR #1946, opened 2026-09-11
fix(conversations): skip park completion after retirement wins

### LINKED ISSUE #1766: Handle retired-sandbox update rejection without crashing conversation actors
## Problem

After #1761, `Conversations.update_sandbox/2` correctly rejects terminal-to-active transitions using the persisted row under a lock. Some callers still match `{:ok, _}` or `{:ok, s}` unconditionally. If retirement wins while provisioning or wake is in progress, the rejection raises a MatchError in the conversation actor instead of handling the stale work explicitly.

Static evidence at #1761 revision e37eb6973c053ceda37eb6e5457a4ce3fcba023c:

- `apps/fountain/lib/fountain/conversations/conversation_server.ex:909`: provisioning completion matches success when setting `ready` after provider work.
- `apps/fountain/lib/fountain/conversations/conversation_server.ex:1190`: wake recovery matches success when moving a previously observed suspended row to `ready`.
- The same assumption appears in the starting transition and in park callers; audit these for retirement races too.

The database guard is correct and should remain. This follow-up concerns actor error handling and cleanup; it is not a reason to remove or bypass the guard. The failure follows from static inspection; no concurrent runtime reproduction was performed during triage.

## Scope and overlap

#1764 (reviewed for overlap at 9f236ef0400065dfbd61a5afa14e1615dfe874c7) guards new user/autonomous turn admission and drops the connection on admission refusal. It does not change the provisioning/wake `update_sandbox/2` success matches above. #1732 is the broader execution-limits work; keep this repair focused.

## Acceptance criteria

- Handle retirement rejection explicitly at provisioning and wake completion, leaving the persisted sandbox terminal and stopping or detaching the stale actor cleanly.
- Do not publish successful provisioning/wake events or start new work after rejection.
- Clean up resources owned by the stale operation without affecting a replacement sandbox or another conversation's connection; preserve the conversation's ability to resume on a valid sandbox.
- Add deterministic regressions that pause provisioning/wake, retire the row, then release the callback. Assert no unhandled MatchError, no revival, and correct cleanup.
- Retain coverage that ordinary provisioning and suspended-to-ready wake still succeed; preserve handling for unrelated database errors.

jhgaylor:
A home checkpoint can finish after another operation retires the sandbox. The subsequent suspended write then rejects retirement revival, but Lifecycle.park/4 assumed success and raised. Handle the rejection explicitly and skip park completion: preserve the terminal row and replacement conversation, emit no suspension event, and notify no co-tenants. A row already terminal or reset-fenced also skips completion. Other database errors still raise. The conversation actor already drops its connection before calling park and stops normally afterward.

Final small unit of #1766, stacked on #1945 (the starting transition). Two files, +91/-17. The stack handles provisioning completion, ready/suspended wake completion, the starting transition and the audited park caller. Closes #1766 when the stack lands.

Validation: two deterministic checkpoint-provider pauses reproduce MatchError on the parent after failed/terminated retirement. With the fix, the park returns normally and preserves the replacement's running conversation and ready sandbox without a suspension event. All 78 focused lifecycle, reset, checkpoint, lifetime and shared-sandbox tests pass, including ordinary parking and unrelated database errors. Full local precommit passed: 5,390 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -1176,39 +1176,46 @@ defmodule Fountain.Conversations.ConversationServer do
         {:error, reason} -> Logger.warning("runtime prepare on wake: #{inspect(reason)}")
       end
 
-      # Normally the wake path already flipped suspended → ready under the
-      # quota reservation; this covers the reaper parking the row mid-wake.
-      # Without it the row would stay `suspended` under a live server —
-      # invisible to the quota and unreachable by any reaper pass.
-      sandbox =
+      # Validate even a cached ready row: retirement may have won while the
+      # provider was waking. Only a suspended wake resets the lifetime clock.
+      attrs =
         if sandbox.status == "suspended" do
-          {:ok, s} =
-            Conversations.update_sandbox(sandbox, %{
-              status: "ready",
-              last_resumed_at: DateTime.utc_now() |> DateTime.truncate(:second)
-            })
-
-          s
+          %{
+            status: "ready",
+            last_resumed_at: DateTime.utc_now() |> DateTime.truncate(:second)
+          }
         else
-          sandbox
+          %{status: "ready"}
         end
 
-      new_state = %{
-        state
-        | handle: handle,
-          sprite_env: sprite_env,
-          sandbox_started_at: Lifecycle.clock_start(sandbox)
-      }
+      case Conversations.update_sandbox(sandbox, attrs) do
+        {:ok, sandbox} ->
+          new_state = %{
+            state
+            | handle: handle,
+              sprite_env: sprite_env,
+              sandbox_started_at: Lifecycle.clock_start(sandbox)
+          }
 
-      new_state = reattach_running_turn(%{new_state | current_turn: nil})
+          new_state = reattach_running_turn(%{new_state | current_turn: nil})
 
-      new_state =
-        Reattachment.finish_runner_reconnect(
-          new_state,
-          if(new_state.current_turn, do: "reattached", else: "turn_ended")
-        )
+          new_state =
+            Reattachment.finish_runner_reconnect(
+              new_state,
+              if(new_state.current_turn, do: "reattached", else: "turn_ended")
+            )
 
-      {:noreply, new_state}
+          {:noreply, new_state}
+
+        {:error, %Ecto.Changeset{errors: [status: {"sandbox is retired", []}]}} ->
+          # Wake owns this connection's credentials, not the existing disk or
+          # another connection's session. Never destroy the machine here.
+          Egress.release_prepared({:ok, state})
+          {:stop, :normal, state}
+
+        error ->
+          raise MatchError, term: error
+      end
     else
       {:error, :not_found} ->
         # The provider says the sandbox is gone. That is the one answer that
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -963,17 +963,15 @@ defmodule Fountain.Conversations.ConversationServer do
                  state.runtime_module,
                  agent,
                  sprite_env
-               ) do
-          # `build_fingerprint` records what the disk was built from, and
-          # `applied_skills` what was mounted on it, so a later reapply can
-          # answer both questions from the row rather than guessing (#1565).
-          {:ok, _} =
-            Conversations.update_sandbox(sandbox, %{
-              status: "ready",
-              build_fingerprint: Reapply.fingerprint(env),
-              applied_skills: skills
-            })
-
+               ),
+             # Record what the disk was built from only if this attempt still
+             # owns a live row. Retirement can win while provider I/O runs.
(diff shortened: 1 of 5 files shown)


BinaryBourbon (approved):
APPROVE — the guard is in the right place, but this stage widens an existing skip from "don't write the row" to "announce nothing at all", and that half deserves to be stated in the `@doc` rather than only in the rewritten test.

**What this changes**

`Lifecycle.park/4` splits into `park_row/1` (the machine half) and `finish_park/4` (the conversation, stream, telemetry and co-tenant half). `park_row/1` keeps the pre-checkpoint filter, then inspects the `suspended` write instead of matching `{:ok, _} =`, so a retirement (or a reset fence) that wins while `HomeCheckpoint.on_park/1` is in provider I/O returns `:skipped`. `:skipped` now short-circuits the whole function.

Direction protected: **revival** — a retired home cannot be written back to `suspended`, which would have put a dead machine back into the quota's parked set, and the actor no longer raises. The wedge direction is largely covered elsewhere rather than here: the caller (`park_sandbox/2`) stops the server regardless, and `terminate/2`'s `orphan_on_normal_stop/3` → `_unsafe_orphan_turn/3` is what flips a conversation with a running turn back to `idle`, which is the realistic `max_lifetime` case.

**Findings**

1. `lifecycle.ex:362-367` — non-blocking, doc accuracy. The `@doc` still promises "the sandbox row to `suspended`, the conversation back to `idle`, the stage event, the telemetry and the co-tenants" unconditionally. After this patch a terminal, retired or reset-fenced row gets none of those five. The PR body says so plainly; the docstring is what the next caller reads. Add the sentence, and say which party is then responsible for the co-tenants.

2. `lifecycle.ex:370-373` — non-blocking, but the one behavior I would want a second opinion on. The pre-check case (row already terminal on entry) previously skipped only the checkpoint and the row write and still ran the tail; the rewritten assertion in `lifecycle_actions_test.exs:204-` records that change, dropping a comment that said "The conversation and the stream still get the news." Two things go with it:
   - **No `sandbox/done` event.** On the `:max_lifetime` park the user's stream gets `drop_connection`'s event and nothing that carries `reason`/`message` — i.e. no explanation that a lifetime ceiling took the machine. `explain(reason, :suspend)` exists precisely to say that.
   - **No `stop_cotenants/5`.** Every other conversation on that home keeps a live server holding a handle to a retired machine until its own next turn fails. The retiring writer usually tells them (`Lifecycle.destroy/5` and `_unsafe_reap_sandbox/1` both do), but the one path that deliberately does not is `do_reattach`'s `{:error, :not_found}` arm, which marks the row `failed` and by design touches no conversation. That is the case where the co-tenants are now told by nobody.
   - A narrower shape gets both directions right: skip when the conversation has moved on (`conv.sandbox_id != sandbox_id` — the replacement case this patch is protecting), and otherwise run `finish_park/4` even though the row write was refused, because the machine really is gone and the conversation and its co-tenants should hear it. `CallbackKey.revoke/2` uses exactly that row-still-names-me test for the same class of problem.
   - Residual, smaller: with `finish_park/4` skipped, a conversation sitting at `running` with no turn in the actor's state (the drift the `if conv.status == "running"` line was a backstop for) now stays `running` with no server to fix it. The turn-orphan path covers the case where a turn exists; nothing covers the drift.

3. `lifecycle.ex:377-383` — nit. The new comment says "Recheck after the provider checkpoint", which reads as if `park_row/1` re-reads the row after `HomeCheckpoint.on_park/1`. It does not; the recheck is `update_sandbox/2`'s own fence under the row lock. Wording it that way ("the write itself rechecks under the row lock") avoids sending the next reader looking for a second `_unsafe_get_sandbox!`.

**Checked and clean**

- `{:error, :sandbox_reset_pending}` is handled explicitly here, which is more than the wake site in #1944 does — the right call, and the reason #1970 exists.
- `park_row(nil) -> :ok` preserves the old behavior for a conversation with no sandbox id: the tail still runs, exactly as the old `if sandbox_id` wrapper allowed.
- The pre-check condition is the old one, negated correctly: terminal-or-reset-fenced skips, everything else checkpoints then writes. No inverted comparison.
- `HomeCheckpoint.on_park/1` stays best-effort and stays before the write, so an unparked disk still gets its checkpoint attempt.
- `raise MatchError, term: error` keeps unrelated write failures loud, and the third new test pins it with a non-retirement `status` error.
- The checkpoint-pause test bites on the parent: the pre-check passes while the row is still `ready`, the retirement lands during the stubbed checkpoint, and the `{:ok, _} =` match then raises, so `assert_receive {:park_result, :ok}` fails. The spawned-process plumbing (`SQL.Sandbox.allow/3` + `Mimic.allow/2`, killed in `on_exit`) is the right way to pause a non-actor code path.
- No tenant-scoping problem: both `_unsafe_` reads keep the "as `home?/1` above" ownership note, and `finish_park/4` reads the conversation by id as before.
- Unlike the other three stages this patch does not touch `conversation_server.ex`, so it is the only one in the stack that the lowered size pin on `main` (2646, zero headroom) does not put at risk.

Issue #1894, opened 2026-09-11
A failed provider delete strands a sandbox and its quota slot with no way back


jhgaylor:
Raised while reviewing #1768, which introduces the reset fence. The fence itself is right; this is about what happens when the provider call after it fails.

## What happens

`finish_sandbox_reset/1` returns `{:error, :sandbox_reset_pending}` on a destroy error, leaving `reset_requested_at` set and the status still `ready`/`suspended`. From there:

- `Quotas.active_sandboxes/0` keeps counting the row — deliberately, and correctly: freeing capacity for a machine that may still be running at the provider is the bug #1768 exists to fix.
- `SandboxReaper` skips it in **both** queries (`is_nil(s.reset_requested_at)`).
- `reset_sandbox/2` refuses a retry, `check_attachable/4` refuses attach, `maybe_reuse_sandbox/1` refuses reuse.
- **Nothing anywhere sets `reset_requested_at` back to `nil`.** Grepped the whole branch.

So one transient provider error — a 500, a network blip — permanently consumes one of the tenant's sandbox slots. With the default clamp (`SANDBOX_RESERVE_CENTS` $2, `SANDBOX_CAP_FLOOR` 2) a small account can lose half its concurrency to one blip, forever.

## Why it needs something

Fail-closed is the right direction and I am not asking to reverse it. The gap is the exit:

- the API tells the user `"reset is pending confirmation; contact the operator before retrying"`
- `docs/concepts/sandboxes.md` says `"The operator must reconcile the provider outcome"` and `"Automatic reconciliation is not implemented"`
- the operator has no lever

The only way to clear a fence today is a hand-written `UPDATE` against the production database, which is the one thing this project has committed to never doing. The documented recovery procedure points at nothing.

## Options, roughly in order of preference

1. **A retry sweep** over fenced rows with backoff — re-attempt the delete, retire on success. This is the "automatic reconciliation" ADR 0042 and the #1768 docs defer, and it makes the common case (transient error) self-healing.
2. **An admin-only unfence** behind `require_admin` that re-probes the provider and either clears the fence or retires the row, so the documented procedure exists and is auditable.
3. **At minimum**, docs that say what the operator actually does, so the sentence is not pointing at a lever nobody has.

(1) and (2) are complementary: the sweep handles blips, the admin path handles a provider that has genuinely lost the machine.

## Also worth deciding separately

[...]

PR #1947, opened 2026-09-11
fix(conversations): support confirmed retries of pending resets

jhgaylor:
A fenced sandbox reset can now be retried after provider failure or caller loss. Original requests and retries elect one finalizer under the sandbox row lock after provider deletion is confirmed; losing callers return `{:ok, :skipped}` and emit no completion event. Capacity remains reserved until confirmed deletion, and provider calls run outside database locks.

Completion events carry the retired sandbox identity. The receiver never interrupts an active turn and atomically checks the conversation’s tenant and current sandbox binding against the retired, fenced sandbox before changing status. A completion delayed behind a rebind cannot idle the replacement conversation or interrupt its turn. Matching completion still updates the sidebar and closes the old connection after the write.

Includes the prerequisite [deadline-wake fix #2047](https://github.com/managoat/fountain/pull/2047) at signed commit `e627aaf91c8a7abfea57376b6b592eee937c9a65`; this PR must land after #2047. Latest main is integrated through `d5e6f476` (including reset/retirement admission ordering).

Refs #1894. The durable retry sweep is stacked in #1948; the accepted admin-only provider reprobe/unfence path remains separate work.

Validation at `80ced6e95a5fa4f8948a6d86b3dce439200feb15`: focused reset/actor/real PostgreSQL concurrency/admission/transport/size coverage passed 68 tests. Tests force original-vs-retry and retry-vs-retry finalizers, provider-response delay after rebind, and a real blocked UPDATE while the holder rebinds and starts a new turn. Compile, formatter, strict Credo, Dialyzer (zero errors/skips), Sobelow and production release passed. Full final precommit passed: core 6 doctests + 5,629 tests, Buzz 144 tests, Support 33 tests, all with zero failures (2 skipped, 7 excluded). Seed 772406, max_cases 4. The earlier frozen repair tree also completed the full suite with zero failures before integration.

Retained initial full-run evidence: three inherited timing failures (AdminBrokerLiveTest timer redirects twice at 100 ms; ExecutionDeadlineWorkerTest 200 ms job timeout killed DB work before its expected terminator message). Those source/test files are unchanged from merged main. Unchanged focused reruns passed 18 admin and 13 deadline tests. Twenty unrelated orphan stress loops were independently verified and stopped; heavy suites are now scheduled at most two at once. No assertions/timeouts were changed. The failed first run was intentionally stopped after 4,724 tests and preserved, then the dedicated test DB was recreated before the constrained full rerun.


BinaryBourbon (commented):
Reviewed `4036d4b12dde3daa2cdddcd7a3bd8fb5c92146cf`.

**[P2] Make reset completion single-winner when a retry overlaps the original delete** — `apps/fountain/lib/fountain/conversations.ex:3560–3566` (and `finish_sandbox_reset/1:3647–3649`).

The pending-row check happens before provider I/O, while `update_sandbox/2` also returns success for an already-terminated row. If the original reset is still awaiting its provider response when a reconciliation retry runs, both deletes can return `:ok` (Sprites deliberately treats an already-deleted/404 machine as success), and both callers run `record_reset_completed/3`. This duplicates the completion audit and transcript notifications. More seriously, if the retry completes first and a tenant wakes the conversation onto its replacement machine before the original request returns, the original request's saved conversation IDs resolve to the new server; its unscoped `{:machine_gone, ...}` cast interrupts that healthy turn and stops the server.

The old reset door refused another fenced reset, so the newly introduced retry path creates this overlap. The stacked cron sweep makes it reachable without an operator calling the helper. Return a distinct already-completed result from an atomic finalization check, and only let the caller that changes pending to terminated publish completion; include the affected sandbox identity in delayed notifications. Add a barrier test where the original destroy is held while retry completes and the conversation wakes.

Validation: source review of both stack PRs, locked retirement, notification consumer and the pinned Sprites deletion adapter; existing sequential retry tests inspected. This concurrency schedule was derived from code, not reproduced against a database. No provider calls or production mutations.

BinaryBourbon (changes requested):
**Changes requested** at `4036d4b12dde3daa2cdddcd7a3bd8fb5c92146cf`.

**[P2] Make reset completion single-winner when retry overlaps the original delete** (`apps/fountain/lib/fountain/conversations.ex:3560–3566`, `3647–3649`). The pending-row check happens before provider I/O, and `update_sandbox/2` returns success even when another caller already terminated the row. If the original reset is awaiting its provider response when a retry completes, both callers can publish completion. Sprites deliberately treats an already-deleted/404 machine as a successful delete.

This can do more than duplicate the audit: after the retry retires the old machine, the tenant can wake onto a replacement before the original request returns. The original request then sends its saved conversation IDs an unscoped `{:machine_gone, ...}` cast; the replacement's server handles it by interrupting its healthy turn and stopping. Previously a second fenced reset was refused; the new retry door introduces the competing finalizer, and #1948's sweep schedules it automatically.

Return an already-completed result from an atomic finalization check and publish only for the caller that wins the pending-to-terminated transition; guard delayed machine notifications with their sandbox identity. Add a barrier regression that holds the original destroy while retry finishes and the conversation wakes.

Validation: re-reviewed unchanged current head and both stack layers, provider deletion semantics and notification handling. A database-free check using the exact-head Sandbox schema and local Ecto confirmed terminal-to-terminal update is a valid empty changeset returning `{:ok, row}`. The full overlap schedule is source-derived, not a database reproduction. [Exact-head CI](https://github.com/managoat/fountain/actions/runs/34607046645) is green but its reset tests cover sequential retries, not this race.

BinaryBourbon (changes requested):
Changes requested after independently re-reviewing unchanged head `4036d4b12dde3daa2cdddcd7a3bd8fb5c92146cf`.

**[P2] Make pending-reset finalization single-winner and scope delayed notifications to the retired sandbox.** In `apps/fountain/lib/fountain/conversations.ex:3560-3566`, the retry checks pending state before provider I/O. `finish_sandbox_reset/1:3644-3649` then uses `update_sandbox/2`, whose locked terminal-to-terminal update still returns `{:ok, row}`. Both an original reset and a retry can therefore call `record_reset_completed/3` for the same deletion.

Concrete failure schedule: hold the original provider response after its fence commits; let a retry confirm deletion and retire the row; wake a holder onto its replacement machine; then release the original successful response. The original caller retains the old holder IDs and resolves their current servers. The unscoped `{:machine_gone, ...}` consumer (`conversation_server.ex:1692-1706`) interrupts the replacement's current turn and stops its server. Duplicate completion audit/transcript events are another consequence.

Required fix: atomically distinguish the caller that changes the pending row to terminated from callers that find it already retired; only that winner should publish completion. Carry the affected sandbox ID through the notification and ignore it when the receiver is running a replacement. Add barrier regressions for original-reset/retry overlap, two retries, and delivery after a holder wakes; assert one completion audit/event and that the replacement turn survives.

Validation: reviewed the exact-head diff, reset fencing/retirement, holder lookup, notification consumer, and sequential tests, alongside #1948. Current CI required and Detect secrets pass. The overlap is established from code paths; I did not rerun PostgreSQL tests or make provider calls. The existing changes request remains unresolved; no separate product decision is needed.

BinaryBourbon (approved):
Approve `80ced6e95a5fa4f8948a6d86b3dce439200feb15` after independent review of the repaired reset finalization and final integration.

Original requests and retries now elect a single completion winner under the persisted sandbox lock after provider confirmation. The losing caller skips audit, quota-release and notification side effects. Completion carries the retired sandbox identity; the receiver checks its local actor state and uses an atomic tenant/binding/status predicate before idling the persisted conversation. A stale reset cannot idle a rebound conversation or interrupt its replacement turn. Matching delivery still updates the sidebar and closes the old connection outside the database lock.

I reviewed the real independent-connection finalizer races and blocked receiver UPDATE regression, plus the final merge resolution. The integration preserves both MachineEvents and ProvisionWatchdog aliases, main's admission FOR SHARE lock, and the independently reviewed deadline fix in #2047. No actionable source defect remains.

Validation attribution: the coder reports 43 original focused tests and 68 final integration tests passing, including admission-order and transport regressions. Full precommit passed on the exact head: compile, format, strict Credo, Dialyzer, Sobelow, production release assembly, 6 doctests + 5,629 core tests, 144 Buzz and 33 Support tests, zero failures. I reviewed source and test coverage rather than rerunning that full suite independently. The PR preserves earlier timing failures and their controlled rerun evidence.

Land after prerequisite #2047 with required current-head and merge-group checks passing. #1948 supplies durable scheduling; the accepted admin recovery work in #1894 remains separate.

PR #1930, opened 2026-09-11
fix: serialize principal credential replay with claims

jhgaylor:
Replaying a claimable principal's create could issue application credentials after a concurrent claim revoked the old ones. Re-read the persisted grant under the claim lock and hold that lock through both credential writes, refusing replay after claim, release or expiry.

Key construction is separated from persistence and audit recording so the credential transaction contains only database writes. Record creation after commit: a failed best-effort audit cannot abort issuance. Ordinary key creation uses the same builder and recorder. Three files, +243/-48, including deterministic concurrency and audit-boundary regressions.

Closes #1689. The independent PostgreSQL-connection race fails on the parent and passes with the lock. The audit regression fails on the earlier draft's in-transaction mint and passes after correction; an unavailable audit recorder leaves committed credentials usable. All 125 focused principal, replay-race and audit-guardrail tests pass. Full local precommit after the audit correction and rebase onto current main passed: 5,377 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

Issue #1688, opened 2026-09-07
Claimable principals: a claimed principal's API key never expires and its owner cannot revoke it


jhgaylor:
## Symptom

Once a visitor claims a principal, the key issued at claim time is unrevocable by anyone but account deletion.

## Mechanism (`apps/fountain/lib/fountain/principals.ex`)

- `:429-438`: the claimed key is minted with `expires_at: nil`.
- It belongs to the principal user, so the owner's `DELETE /api/auth/api-keys` cannot see it; the principal key itself is refused by `require_key_management`.
- `release/2` (`:578`) returns `:already_claimed`; the console shows no principals (#1566).
- Idempotent replay of the create (`:506-514`) mints another key without revoking the previous one.

## Fix

Give the claimed key a bounded expiry (or tie it to the claimer's session key lifetime), list principal keys under the owner's key management, and revoke the previous key on replay.

---
Found in the 2026-09-07 review of the week on `main` (cb93a7ea).

PR #1949, opened 2026-09-11
fix(accounts): let owners manage claimed principal credentials

jhgaylor:
A claimed principal's credential belonged to the principal user, so its owner could neither find nor revoke it. Add explicit management queries for the owner's own keys and principal-scoped credentials of its claimed principals, then use them in the existing API and console. Revocation records an owner-visible audit event identifying the principal. The console marks principal credentials.

Owner-management unit of #1688, stacked on #1930 so subsequent rotation and renewal changes share the credential transaction boundary. Six files, +152/-7, including the audit guardrail. Expiry and claim replay rotation remain separate. Existing tenant-only helpers and principal key-management scope restrictions remain; callback credentials are excluded from the owner's additional key view.

Validation: API and console regressions fail on the parent (27 tests, 2 failures). All 127 focused API, LiveView, scope and audit-guardrail tests pass, covering foreign-account and principal self-management refusals, callback exclusion, metadata-only listing, effective revocation and owner auditing. Full local precommit passed: 5,381 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/accounts.ex
+++ b/apps/fountain/lib/fountain/accounts.ex
@@ -921,6 +921,56 @@ defmodule Fountain.Accounts do
     |> Repo.all()
   end
 
+  @doc "Active keys the account manages: its own keys and its claimed principals' credentials."
+  def list_managed_api_keys(user_id) when is_binary(user_id) do
+    user_id
+    |> managed_api_keys()
+    |> where([k], is_nil(k.revoked_at))
+    |> order_by([k], desc: k.inserted_at)
+    |> Repo.all()
+  end
+
+  @doc "Revoke an owned or claimed-principal credential, with a trail the managing owner can read."
+  def revoke_managed_api_key(user_id, key_id, opts \\ []) do
+    case user_id |> managed_api_keys() |> where([k], k.id == ^key_id) |> Repo.one() do
+      nil ->
+        {:error, :not_found}
+
+      key ->
+        with {:ok, revoked} <- revoke_api_key(key.user_id, key.id, opts) do
+          if key.user_id != user_id do
+            Audit.record(%{
+              user_id: user_id,
+              action: "api_key.revoked",
+              resource_type: "api_key",
+              resource_id: key.id,
+              actor: Keyword.get(opts, :actor, "self"),
+              request_ip: Keyword.get(opts, :request_ip),
+              metadata: %{
+                "name" => key.name,
+                "key_prefix" => key.key_prefix,
+                "principal_user_id" => key.user_id
+              }
+            })
+          end
+
+          {:ok, revoked}
+        end
+    end
+  end
+
+  defp managed_api_keys(user_id) do
+    principals =
+      from o in Fountain.Principals.Owner,
+        where: o.owner_user_id == ^user_id,
+        select: o.principal_user_id
+
+    from k in ApiKey,
+      where:
+        k.user_id == ^user_id or
+          ("principal" in k.scopes and k.user_id in subquery(principals))
+  end
+
   @doc "List all users, ordered by insertion date."
   def list_users do
     from(u in User, order_by: [asc: u.inserted_at]) |> Repo.all()
--- a/apps/fountain/lib/fountain_web/controllers/api_key_controller.ex
+++ b/apps/fountain/lib/fountain_web/controllers/api_key_controller.ex
@@ -33,7 +33,7 @@ defmodule FountainWeb.ApiKeyController do
 
   def index(conn, _params) do
     user = conn.assigns.current_user
-    render(conn, :index, keys: Accounts.list_api_keys(user.id))
+    render(conn, :index, keys: Accounts.list_managed_api_keys(user.id))
   end
 
   operation(:create,
@@ -95,7 +95,7 @@ defmodule FountainWeb.ApiKeyController do
   def delete(conn, %{"id" => id}) do
     user = conn.assigns.current_user
 
-    case Accounts.revoke_api_key(user.id, id, Audited.attribution(conn)) do
+    case Accounts.revoke_managed_api_key(user.id, id, Audited.attribution(conn)) do
       {:ok, _key} ->
         send_resp(conn, :no_content, "")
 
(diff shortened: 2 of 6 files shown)


BinaryBourbon (approved):
APPROVE — the patch is correct. The red `CI required` is an infrastructure artifact, not a code failure, but it still blocks the merge; see the first finding.

**What this changes**

`Accounts.list_api_keys/1` and `revoke_api_key/3` are tenant-scoped to one `users` row, so a claimed principal's `principal`-scoped credential — which lives under the *principal's* `user_id`, not the owner's — was invisible and unrevokable from the owner's own key surface. This adds one private query, `managed_api_keys/1`, whose `where` is `k.user_id == ^user_id or ("principal" in k.scopes and k.user_id in subquery(principal_owners))`, and two public functions over it; the API index/delete and the console LiveView switch to them. Revoking somebody else's row additionally writes an `api_key.revoked` row under the *owner's* `user_id` carrying `principal_user_id`, because the row `revoke_api_key/3` already writes lands under the principal and no owner can read it.

**Findings**

1. **`CI required` is FAILURE for an infrastructure reason, and the PR cannot merge until it is re-run.** Non-blocking as a code review; blocking as a merge gate. Two CI runs were created for head `e648f99` at the same second — `34608000639` (success, every job green) and `34607999528`, which the concurrency group cancelled 13s in. The `gate` job runs `if: always()`, so in the cancelled run it executed against cancelled dependencies and `scripts/ci/gate.py:54` raised `ValueError: change classification is missing or invalid` (the `Decide whether this is a docs-only change` job was cancelled and emitted no classification). GitHub surfaces that job as the latest result for the check name, so the rollup is red while the identical tree is green in the sibling run. Nothing to fix here — re-run `34607999528`, or push to re-trigger. Worth a separate look though: a duplicate `pull_request` run whose cancellation manufactures a red required check will do this to any PR in this programme.

2. **`apps/fountain/lib/fountain_web/controllers/api_key_json.ex` — the API index now returns another tenant's key with nothing that identifies it as one.** Non-blocking. The console got a `Principal` badge (`live/api_keys_live/index.ex:147`), but `summary/1` emits only `id, name, prefix, created_at, last_used_at, scopes, expires_at` — no `user_id`, no principal marker. An API client reading `GET /api/auth/api-keys` sees a row it cannot attribute except by inferring from `scopes == ["principal"]` and the `principal:<application_id>` name convention. The `operation(:index, ...)` description still says "Metadata only — key material is returned once…" and does not mention that claimed principals' credentials now appear in this list at all. Either add a discriminator field or say it in the description; right now the console and the API disagree about how legible the list is.

3. **`accounts.ex:929` — `revoke_managed_api_key/3` does not filter `is_nil(k.revoked_at)`, so re-revoking is a no-op that records as a change.** Non-blocking, and pre-existing in `revoke_api_key/3` (`Repo.get_by(ApiKey, id:, user_id:)` has the same gap). Scenario: `DELETE /api/auth/api-keys/<id>` twice. The second call rewrites `revoked_at` to now, losing the original revocation instant, returns 204, and writes a second `api_key.revoked` pair. CLAUDE.md's "only record what happened" says a no-op records nothing. Since `list_managed_api_keys/1` *does* filter revoked, the read and the write disagree about what is in scope; adding the same filter to the write query would make a repeat a clean 404.

4. **`accounts.ex:964` — `Fountain.Accounts` now reads `Fountain.Principals.Owner` while `Fountain.Principals` calls `Fountain.Accounts`.** Non-blocking style note. The contexts are now mutually referential. A `Principals.owned_principal_ids(user_id)` returning the subquery (it already exposes `owner_id/1`) would keep the ownership predicate on the side that owns the table.

5. Nit: `list_managed_api_keys/1` has a `when is_binary(user_id)` guard and `revoke_managed_api_key/3` has none, and neither has a `@spec` while their single-tenant neighbours do.

**Checked and clean**

- Tenant scoping: `revoke_managed_api_key/3` establishes ownership with the scoped `managed_api_keys/1` fetch *before* delegating to `revoke_api_key(key.user_id, key.id, opts)`, which is itself scoped. No `_unsafe_` call, and no fetch-by-id-then-check.
- `principal_owners` has `unique_index(:principal_user_id)` and no release/retire column, so there is no former-owner who keeps the ownership row and therefore keeps read access to a rebound principal's credential.
- Audit: recorded in the context, outside any transaction (`revoke_api_key/3` is not transactional), metadata is `name` / `key_prefix` / `principal_user_id` — no key material, no hash. Actor defaults to `"self"` and comes from `Audited.attribution/2` at both call sites, so no bare `"system"`.
- The guardrail row is a real test, not a formality: `list_recent_for_user(user.id, 200)` is scoped to the *owner*, and only the new inline `Audit.record` puts an `api_key.revoked` row there — delete it and the case fails. Nothing was added to `@deliberately_silent`.
- Callback keys are excluded by the `"principal" in k.scopes` conjunct, and the controller test asserts it (`refute callback.id in ids`).
- Grepped the other `list_api_keys/1` callers: `admin_live/user_detail.ex:69` (per-user admin view) and `start_live.ex:60` (`has_key?` for the onboarding checklist) both genuinely want own-keys-only and were correctly left alone.
- `refute inspect(response) =~ raw` in the controller test, and the console diff adds no new render of key material.
- `@user_id` is assigned in `mount/3`, so the `k.user_id != @user_id` badge condition is sound on first render and after both `assign(:keys, …)` refreshes.
- No `CHANGELOG.md` `[Unreleased]` entry — consistent with the rest of this programme, noted once, not a blocker.

BinaryBourbon (approved):
Reviewed the rebased change against main. No blocking findings. The final head adds an empty CI refresh commit; its tree is identical to independently reviewed ac182701.

The management query requires the caller’s ownership and limits additional access to principal-scoped keys. Foreign principals and callback credentials remain excluded. API and console use that query; revocation adds the owner-visible audit event without exposing credentials. Regression coverage exercises these boundaries and effective revocation.

APPROVE. Merge remains contingent on passing CI for this head.

Reviewed commit: `becd11a40d3576610c5f19baebe2738deb376c66`. Code and test coverage were inspected; fresh CI/local execution is tracked separately.

BinaryBourbon (approved):
APPROVE after rebasing onto main d5e6f476 (#1969). Re-reviewed the resulting diff and the new base: git range-diff confirms both commits are patch-identical to the previously approved revision, and the main change concerns sandbox admission locking. Ownership scoping, callback-key exclusion and audit behavior remain unchanged. No blocking findings; waiting for this head’s fresh CI before merge.

PR #1950, opened 2026-09-11
fix(auth): rotate principal credentials atomically on claim replay

jhgaylor:
Replaying a successful claim left both the previous and new principal credentials active. Serialize the revoke and mint under the claim-row lock, so concurrent claim replays leave one usable principal credential. A failed first mint now rolls back ownership attachment and anonymous-key revocation together. Runtime callback credentials survive replay; credential audit records are written after commit.

Rotation unit of #1688, stacked on #1949. Four files, +176/-18. Owner renewal and bounded expiry follow as separate small units.

Validation: three regressions fail on the parent: retained old credentials, ownership committed despite a failed mint, and multiple active credentials after concurrent replay. The race uses independent PostgreSQL connections and verifies both operations wait on the claim lock. All 175 focused principal, race, API, console, scope and audit-guardrail tests pass; the final update-returning refinement also passes all 48 principal/race tests. Full local precommit passed: 5,383 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/accounts.ex
+++ b/apps/fountain/lib/fountain/accounts.ex
@@ -921,6 +921,56 @@ defmodule Fountain.Accounts do
     |> Repo.all()
   end
 
+  @doc "Active keys the account manages: its own keys and its claimed principals' credentials."
+  def list_managed_api_keys(user_id) when is_binary(user_id) do
+    user_id
+    |> managed_api_keys()
+    |> where([k], is_nil(k.revoked_at))
+    |> order_by([k], desc: k.inserted_at)
+    |> Repo.all()
+  end
+
+  @doc "Revoke an owned or claimed-principal credential, with a trail the managing owner can read."
+  def revoke_managed_api_key(user_id, key_id, opts \\ []) do
+    case user_id |> managed_api_keys() |> where([k], k.id == ^key_id) |> Repo.one() do
+      nil ->
+        {:error, :not_found}
+
+      key ->
+        with {:ok, revoked} <- revoke_api_key(key.user_id, key.id, opts) do
+          if key.user_id != user_id do
+            Audit.record(%{
+              user_id: user_id,
+              action: "api_key.revoked",
+              resource_type: "api_key",
+              resource_id: key.id,
+              actor: Keyword.get(opts, :actor, "self"),
+              request_ip: Keyword.get(opts, :request_ip),
+              metadata: %{
+                "name" => key.name,
+                "key_prefix" => key.key_prefix,
+                "principal_user_id" => key.user_id
+              }
+            })
+          end
+
+          {:ok, revoked}
+        end
+    end
+  end
+
+  defp managed_api_keys(user_id) do
+    principals =
+      from o in Fountain.Principals.Owner,
+        where: o.owner_user_id == ^user_id,
+        select: o.principal_user_id
+
+    from k in ApiKey,
+      where:
+        k.user_id == ^user_id or
+          ("principal" in k.scopes and k.user_id in subquery(principals))
+  end
+
   @doc "List all users, ordered by insertion date."
   def list_users do
     from(u in User, order_by: [asc: u.inserted_at]) |> Repo.all()
--- a/apps/fountain/lib/fountain/accounts.ex
+++ b/apps/fountain/lib/fountain/accounts.ex
@@ -730,10 +730,18 @@ defmodule Fountain.Accounts do
         key
         |> Ecto.Changeset.change(revoked_at: DateTime.utc_now() |> DateTime.truncate(:second))
         |> Repo.update()
-        |> audited_account("api_key.revoked", "api_key", opts, fn revoked ->
-          %{"name" => revoked.name, "key_prefix" => revoked.key_prefix}
-        end)
+        |> case do
+          {:ok, revoked} -> record_api_key_revoked(revoked, opts)
+          error -> error
+        end
     end
+  end
+
+  @doc "Audit a committed key revocation outside the transaction that changed it."
+  def record_api_key_revoked(%ApiKey{} = key, opts \\ []) do
+    audited_account({:ok, key}, "api_key.revoked", "api_key", opts, fn revoked ->
+      %{"name" => revoked.name, "key_prefix" => revoked.key_prefix}
+    end)
   end
 
   @doc """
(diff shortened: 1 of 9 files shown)


BinaryBourbon (approved):
APPROVE — the atomicity claim holds, and there is no window in which a caller sees either two live credentials or none.

**What this changes**

`claim/4` used to be two commits: `attach_owner/4` committed the ownership row and the status flip, then `mint_principal_key/2` inserted the key in its own transaction. A failed mint therefore left a claimed principal with no credential and an already-revoked anonymous key, and a replay minted a second credential without touching the first. `attach_owner/4` becomes `claim_with_credential/4`, which does the decide, the replay revoke (`revoke_claimed_credentials/1`, an `update_all … select: k` over `"principal" in k.scopes and is_nil(k.revoked_at)`) and the key insert inside the one transaction already serialized by `lock_claimable/1`'s `FOR UPDATE`. `Accounts.record_api_key_revoked/2` is extracted so the audit for the revoked rows can be written after commit alongside `record_api_key_created/2`.

**On the window the title claims to close**

There is none, and the `FOR UPDATE` is not what closes it — the shared snapshot is. Both the `UPDATE api_keys SET revoked_at` and the `INSERT` are uncommitted until the same `COMMIT`, so under READ COMMITTED a concurrent `authenticate_api_key/1` on another connection reads `revoked_at IS NULL` on the old row and cannot see the new row at all. The old credential keeps working right up to the commit instant and the new one is valid from it. A caller holding the old credential sees no gap in either direction, and a crash mid-transaction leaves the old credential live rather than the principal locked out. `revoke_keys/1` on the first-claim path is inside the same transaction for the same reason.

**Findings**

1. **`principals.ex:536-539` — the rotation's revoke audit lands only under the principal's `user_id`, so the owner surface #1949 just built shows no trail for it.** Non-blocking; fix belongs here rather than in #1949. `record_api_key_revoked/2` goes through `audited_account/5`, which sets `user_id` from `Map.get(record, :user_id)` — the principal. `principals_test.exs:358` asserts exactly that (`audit.user_id == first.claimable.user_id`). So: an owner replays a claim, their previous principal credential is revoked, and `Audit.list_recent_for_user(owner_id, …)` has nothing. #1949 added an owner-visible row specifically because that trail is unreachable, and #1951 adds an owner-visible `api_key.created` row for renewal — this path is the one of the three with no owner row. One `Audit.record` with `principal_user_id` and the replaced key ids, mirroring `renew_owned_credential/3`, would make the three consistent.

2. **`principals_claim_replay_test.exs:178-186` — the test asserts on a backend pid it may not have measured.** Non-blocking, and the pattern is inherited from #1930, but this test doubles the exposure. The task does `Repo.query!("SELECT pg_backend_pid()")`, which checks a connection out and back in, then calls `Principals.claim/4`, which checks out again for `check_claimer_eligible/1`'s `Billing.check_spend/1` and again for `Repo.transaction`. Nothing pins the transaction to the connection whose pid was sampled. If the pool hands the transaction a different connection, `waits_for_lock?/2` polls `pg_stat_activity` for an idle backend, never sees `wait_event_type = 'Lock'`, and fails on the 5s deadline — a red build with nothing wrong in the code under test. Two replay tasks means two independent chances of that per run. Sampling `pg_backend_pid()` *inside* the transaction (or looking the waiter up by `query LIKE`/`relation` rather than by pid) removes it. This is the "waits on a signal other than the one it asserts" shape that has produced flakes here before.

3. Nit: `principal_key_opts/2` is now used for three different jobs — the key's own `scopes`/`expires_at`, and the audit `:actor`/`:request_ip` opts for both `record_api_key_revoked/2` and `record_api_key_created/2`. It works because `audited_account/5` ignores the key-shaped entries, but passing a list named for key construction into an audit recorder invites a future `expires_at`-reading audit to silently read a value recomputed outside the transaction. Splitting the attribution pair out would make both call sites honest.

**Checked and clean**

- `decide_claim/4` reaches `{claimable, replay?}` on exactly two paths (replay, and `write_claim/3`); every other clause is a non-local `Repo.rollback`, so the new bare `{claimable, replay?} =` match cannot raise on an error case. `current = lock_claimable(id) || Repo.rollback(:not_found)` is the same non-local return.
- `revoke_claimed_credentials/1` is `is_nil(k.revoked_at)`-filtered, so a replay does not re-stamp the anonymous keys `write_claim/3` already revoked, and `"principal" in k.scopes` leaves `sprite` callback keys alone — asserted both in `principals_test.exs` (`is_nil(Repo.reload!(callback).revoked_at)`) and in the race test's count.
- `update_all(… select: k)` returns post-update rows, so the audit metadata reads the revoked `name`/`key_prefix` rather than a stale struct, and the audit never sees key material or the hash.
- Audit ordering: both `record_api_key_revoked/2` and `record_api_key_created/2` run after `Repo.transaction/1` returns. `Repo.in_transaction?` is false there, which the existing "credential audit runs after commit" test in this file already pins.
- `revoke_api_key/3`'s rewrite from `|> audited_account(…)` to `|> case do {:ok, revoked} -> record_api_key_revoked(…)` preserves both the `{:ok, key}` and the `{:error, changeset}` return shapes, so #1949's `revoke_managed_api_key/3` `with` still matches.
- Both new tests fail on the parent for the right reason: the parent mints without revoking, so the replay test's active-principal-key count is 3 rather than 1, and the mint-failure test's `ClaimableUser.status` is `"claimed"` rather than `"unclaimed"`. The Mimic stub in the second one wraps `call_original` and only adds a changeset error, so it does not stub the thing it asserts on.
- No `_unsafe_` call, no pool_size change, no new `async: false` beyond the file's pre-existing and justified one, no `Task.async` for fire-and-forget (the `Task.async` here is a test harness that awaits).

BinaryBourbon (approved):
Re-reviewed the rebased patch. No blocking findings. Ownership attachment, prior-key revocation and replacement mint commit together under the claim-row lock. Failed minting rolls back the whole change; concurrent replays serialize and preserve callback credentials. Credential audits occur after the transaction returns.

Correction to the earlier review: Sandbox.unboxed_run/2 checks out one connection around its callback, so the backend PID query and claim transaction are pinned to the same connection. That concern requires no fix.

Reviewed commit: `3131f6312655c90d884f8c11e0d4541593242c8c`. Code and test coverage were inspected; fresh CI/local execution is tracked separately.

BinaryBourbon (approved):
APPROVE after rebasing onto current main cb7f94eb5cade26b27137c881e69b9ad3b8f34be.

Re-reviewed the final patch and its integration with the preceding merged stack units. git range-diff confirms patch equivalence to the independently approved stack revision. The replay transaction still serializes revocation and minting under the claim-row lock, rolls back failed minting, preserves callback credentials and audits after commit. No blocking findings.

Reviewed commit: `06fff152c7da2785148be7a226b39cade2eace10`.

PR #1951, opened 2026-09-11
feat(auth): let owners renew claimed principal credentials

jhgaylor:
Owners who claimed without an idempotency key had no way to renew a lost or expired principal credential. Add an owner-scoped renewal operation addressed by principal id. It reads persisted ownership, takes the same claim-row lock as replay, and commits revocation plus replacement together. Renewal preserves runtime callback credentials, permits recovery without a spending balance, and records the new key and replaced key ids in the owner's audit trail after commit.

Context unit of #1688, stacked on #1950. Five files, +210/-14, including the ADR 0044 amendment and required audit guardrail. The console entry point and bounded expiry are separate small units.

Validation: all 135 focused principal, independent PostgreSQL race and audit-guardrail tests pass. Coverage includes no-token recovery after expiry/revocation, foreign-account/application/principal refusals, mint-failure rollback without a false audit, and concurrent owner renewal versus claim replay leaving one active credential. The decisions bundle validates and its regenerated index is unchanged; six existing trust-metadata warnings concern other ADRs. Full local precommit passed: 5,390 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/accounts.ex
+++ b/apps/fountain/lib/fountain/accounts.ex
@@ -730,10 +730,18 @@ defmodule Fountain.Accounts do
         key
         |> Ecto.Changeset.change(revoked_at: DateTime.utc_now() |> DateTime.truncate(:second))
         |> Repo.update()
-        |> audited_account("api_key.revoked", "api_key", opts, fn revoked ->
-          %{"name" => revoked.name, "key_prefix" => revoked.key_prefix}
-        end)
+        |> case do
+          {:ok, revoked} -> record_api_key_revoked(revoked, opts)
+          error -> error
+        end
     end
+  end
+
+  @doc "Audit a committed key revocation outside the transaction that changed it."
+  def record_api_key_revoked(%ApiKey{} = key, opts \\ []) do
+    audited_account({:ok, key}, "api_key.revoked", "api_key", opts, fn revoked ->
+      %{"name" => revoked.name, "key_prefix" => revoked.key_prefix}
+    end)
   end
 
   @doc """
--- a/apps/fountain/lib/fountain/principals.ex
+++ b/apps/fountain/lib/fountain/principals.ex
@@ -569,18 +569,78 @@ defmodule Fountain.Principals do
       {claimable, replay?} = decide_claim(current, claim_token, claimer, opts)
       revoked = if replay?, do: revoke_claimed_credentials(claimable.user_id), else: []
 
-      {changeset, raw} =
-        Accounts.build_api_key(
-          claimable.user_id,
-          key_name(claimable),
-          principal_key_opts(claimable, opts)
-        )
-
-      case Repo.insert(changeset) do
-        {:ok, key} -> {claimable, replay?, key, raw, revoked}
-        {:error, changeset} -> Repo.rollback(changeset)
-      end
+      {key, raw} = insert_principal_key(claimable, opts)
+      {claimable, replay?, key, raw, revoked}
     end)
+  end
+
+  @doc """
+  Replace a claimed principal's credential using its current owner's authority.
+
+  Works without the original claim token or idempotency key, including after
+  expiry or revocation. The claim lock serializes renewal with claim replay.
+  Revocation and minting commit together; runtime callback keys are untouched.
+  Returns `:not_found` for a principal the account does not own.
+  """
+  @spec renew_owned_credential(binary(), binary(), keyword()) ::
+          {:ok, {ApiKey.t(), String.t()}} | {:error, term()}
+  def renew_owned_credential(owner_user_id, principal_user_id, opts \\ []) do
+    result =
+      Repo.transaction(fn ->
+        owned =
+          from o in Owner,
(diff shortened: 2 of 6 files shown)


BinaryBourbon (approved):
## APPROVE

Reviewed on its own patch against `codex/p2-1688-claim-key-rotation`, plus the surrounding `Principals`/`Accounts` code the new function leans on. The security-relevant shape is right, and I verified the two guards that matter by breaking them rather than by trusting the tests.

### Verified

**Ownership scoping is real.** `renew_owned_credential/3` resolves the principal through a `principal_owners` subquery keyed on `owner_user_id`, plus `status == "claimed"` — no `_unsafe_` call, no unscoped first fetch. I removed the `c.user_id in subquery(owned)` clause and re-ran: `"the application, another account and the principal cannot renew"` and the console's forged-ownership test both fail. The guard guards at both layers.

**No plaintext in the trail.** I replaced `"key_prefix" => key.key_prefix` with `"key_prefix" => raw` and `"the owner renews without the claim token…"` fails on its `refute inspect(audit.metadata) =~ raw`. The raw key leaves the context exactly once, in the return tuple.

**Atomicity and the rotation window.** Revoke and mint sit in one `Repo.transaction` under `FOR UPDATE` on the same claimable row that `lock_claimable/1` takes, so renewal serialises against claim replay rather than racing it — and the new `:renewal` arm of `assert_one_credential_after_rotation/1` exercises that against a real second backend. There is no interval in which both credentials authenticate. On a mint failure `insert_principal_key/2` rolls the revocation back with it, and the Mimic test confirms the previous credential still works and no owner-side `api_key.created` row was written.

**Audit placement.** Both `Accounts.record_api_key_created/2` and the owner-side `Audit.record/1` are outside the transaction, in the `with` after commit — the rule that matters most here, since a rescued insert inside the transaction would take the mint down with it. Actor defaults to `"self"`, in the closed vocabulary; no bare `"system"` anywhere. The guardrail row was added.

**Route gates.** No new route: the diff touches no router, no operation declaration and no SDK surface, so the seven-gate checklist does not apply. ADR 0044 decision 5 is amended in the same PR and the index regenerates unchanged. All six partitions plus `okf validate` are green, and 148 focused tests pass locally.

### Non-blocking

1. **Suspension is not checked, and claim checks it.** `claim/4` goes through `check_claimer_eligible/1`, which refuses an unverified or `suspended_at` claimer. Renewal checks neither. Skipping `Billing.check_spend/1` is deliberate and I agree with it — `billing_subject_id/1` routes principal spend to the owner's ledger, so the gate still bites at every door that spends, which the balance test proves. But suspension is an operator action rather than a balance, and the key minted here authenticates as the *principal*, a user row that is not suspended. `require_authenticated_user` does not check suspension either, so a suspended owner with a live session reaches this. It is not a regression (the same page already lets them mint their own key, and a principal key they already hold survives suspension today), but the `suspended_at`/`email_verified_at` half of the eligibility check looks like it belongs here.

2. **The replacement credential never expires.** `principal_key_opts/2` returns `expires_at: nil` for a claimed principal, so every renewal mints a non-expiring key. That matches the existing claim path and the PR says bounded expiry is a separate unit — but the new ADR prose and the docstring now advertise renewal as the recovery path "after expiry" without saying the replacement has none. Worth a clause when the expiry unit lands.

3. **The owner's trail records no revocation row.** `revoke_managed_api_key/3` writes an owner-side `api_key.revoked` when the key belongs to a principal; renewal's revocations land only on the principal's trail, with the ids in the owner's `api_key.created` metadata. Readable either way, just inconsistent with the sibling path.

4. Cosmetic: with no opts, `principal_key_opts/2` defaults `:actor` to `"api"` while the new `Audit.record/1` defaults to `"self"`, so one renewal writes two rows under two actors. Both are in the vocabulary; the console passes `attribution/1` and gets `"ui"` on both, so it only shows up in background callers.

BinaryBourbon (approved):
Re-reviewed the rebased patch. No blocking findings. Ownership is enforced in the initial query; renewal and replay take the same claim-row lock; revocation and minting commit together; runtime callback credentials remain untouched. Audits run after commit and contain metadata rather than secrets. Tests cover unauthorized callers, recovery after expiry/revocation, failed-mint rollback and concurrent renewal versus replay through separate database connections.

Reviewed commit: `723b5de97fd1aa2da1ca748ef7d60a994a78efab`. Code and test coverage were inspected; fresh CI/local execution is tracked separately.

BinaryBourbon (approved):
APPROVE after rebasing onto current main bbe5c089ce2004f55800cbb56ab2d1fef4d66aee.

Re-reviewed the final patch and its integration with the preceding merged stack units. git range-diff confirms patch equivalence to the independently approved stack revision. The initial ownership-scoped query and shared claim-row lock preserve tenant isolation and atomic renewal; failed minting rolls back and audit writes happen after commit. No blocking findings.

Reviewed commit: `9d5caa7499a4d39f78738415e454a271390b074f`.

PR #1952, opened 2026-09-11
feat(console): recover claimed principal API keys

jhgaylor:
A principal owner can revoke a key, but cannot regain access from the console afterward. Add a replacement-key form populated from owned principal IDs, so it remains available when every previous key is revoked or expired. The context replaces the credential atomically; the console refreshes key metadata and reveals the new secret once in the existing modal.

Console unit of #1688, stacked on #1951. Two files, +98/-0. Bounded expiry follows separately.

Validation: all 13 focused console tests pass, including recovery after revocation, one-time secret dismissal, effective replacement, and forged-owner/malformed-ID refusals. A real Chromium session recovered a revoked credential, authenticated with its replacement, replaced it again, verified the previous credential returns HTTP 401 and the new one HTTP 200, and checked that the form fits a 390px viewport. Browser fixtures used a separate database and were deleted afterward. Full local precommit passed: 5,392 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/principals.ex
+++ b/apps/fountain/lib/fountain/principals.ex
@@ -569,18 +569,78 @@ defmodule Fountain.Principals do
       {claimable, replay?} = decide_claim(current, claim_token, claimer, opts)
       revoked = if replay?, do: revoke_claimed_credentials(claimable.user_id), else: []
 
-      {changeset, raw} =
-        Accounts.build_api_key(
-          claimable.user_id,
-          key_name(claimable),
-          principal_key_opts(claimable, opts)
-        )
-
-      case Repo.insert(changeset) do
-        {:ok, key} -> {claimable, replay?, key, raw, revoked}
-        {:error, changeset} -> Repo.rollback(changeset)
-      end
+      {key, raw} = insert_principal_key(claimable, opts)
+      {claimable, replay?, key, raw, revoked}
     end)
+  end
+
+  @doc """
+  Replace a claimed principal's credential using its current owner's authority.
+
+  Works without the original claim token or idempotency key, including after
+  expiry or revocation. The claim lock serializes renewal with claim replay.
+  Revocation and minting commit together; runtime callback keys are untouched.
+  Returns `:not_found` for a principal the account does not own.
+  """
+  @spec renew_owned_credential(binary(), binary(), keyword()) ::
+          {:ok, {ApiKey.t(), String.t()}} | {:error, term()}
+  def renew_owned_credential(owner_user_id, principal_user_id, opts \\ []) do
+    result =
+      Repo.transaction(fn ->
+        owned =
+          from o in Owner,
+            where: o.owner_user_id == ^owner_user_id,
+            select: o.principal_user_id
+
+        claimable =
+          Repo.one(
+            from c in ClaimableUser,
+              where:
+                c.user_id == ^principal_user_id and c.user_id in subquery(owned) and
+                  c.status == "claimed",
+              lock: "FOR UPDATE"
+          ) || Repo.rollback(:not_found)
+
+        revoked = revoke_claimed_credentials(claimable.user_id)
+        {key, raw} = insert_principal_key(claimable, opts)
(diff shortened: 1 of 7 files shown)


BinaryBourbon (approved):
## APPROVE

Reviewed on its own patch against `codex/p2-1688-owner-key-renewal`, with the full LiveView and the router's `live_session` around it.

### Verified

**The guard is right.** `/api-keys` sits in `live_session :authenticated` under `{FountainWeb.Live.Hooks, :require_authenticated_user}` — the correct hook for a console page, and unchanged by this PR.

**No cross-tenant exposure.** The owner id comes from `socket.assigns.user_id`, never from params, and `Principals.renew_owned_credential/3` does the scoping. The `principal_id` from the form is untrusted and treated as such: `@principals` is `Principals.list_owned(user.id)`, so the `<option>` set is owner-scoped, and submitting a foreign id anyway is refused by the context. I confirmed the refusal is load-bearing by deleting the ownership subquery in the context and re-running — this file's forged-ownership test fails.

**`Ecto.UUID.cast/1` is doing real work.** Without it a non-UUID `principal_id` reaches `Repo.one` and raises `Ecto.Query.CastError`, killing the LiveView. The `with` head catches it before the query, and the test covers `"not-a-uuid"`.

**No oracle in the error path.** The `else` collapses `:not_found` and a changeset failure into one flash, so a forged submitter cannot tell an unowned principal from a nonexistent one. The test asserts the same string for both arms and that the victim's credential still authenticates afterwards.

**One-time reveal.** The raw key reuses the existing `new-key-modal`, lives only in socket assigns, and `dismiss_new_key` clears it — the test asserts the plaintext is gone from the render after dismissal and that a second renewal revokes the first. Nothing persists or logs it. The key table still shows prefixes only.

No route, no schema, no SDK surface, so the API-route gates do not apply. CI is green across all six partitions.

### Non-blocking

1. **The destructive action has no confirmation, and its neighbour does.** The revoke button four elements down carries `data-confirm="Revoke this key? Any integrations using it will stop working."`. Submitting this form revokes a working credential just as irreversibly, with only prose warning about it. A `data-confirm` on the submit button would match the page.

2. **The select shows bare UUIDs.** An owner with more than one principal cannot tell them apart. `claimable.application_id` (or `application_id` plus a truncated id) would make the choice legible; `list_owned/1` currently selects only `principal_user_id`, so this needs a widened select.

3. **`@principals` is computed at mount only.** Claiming a principal in another tab leaves the select stale until reload. Cosmetic — `handle_event` refreshes `@keys` but not `@principals`.

For the record on the one thing that looked reachable and is not: a mint against a deleted principal would raise `Ecto.ConstraintError` past the `with/else` and crash the LiveView, but `release/2`, `expire_due/1` and `purge_closed/1` all act on `status == "unclaimed"` only, so a claimed principal's row can never outlive its user. Nothing to do here.

BinaryBourbon (approved):
Re-reviewed the rebased console change. No blocking findings. The authenticated socket supplies owner identity, UUID validation precedes the query and the context rejects forged ownership. Listing principals independently of active keys preserves recovery after revocation. Replacement uses the existing one-time secret display. Tests cover recovery, dismissal, rotation, malformed IDs and forged ownership.

Nonblocking followup: an application label alongside each principal UUID would make selection easier.

Reviewed commit: `e1215efb4137057fa5fd31ce8421a4a3c51d084d`. Code and test coverage were inspected; fresh CI/local execution is tracked separately.

BinaryBourbon (approved):
APPROVE after rebasing onto current main 1d91d4e8e43447469ea28ac5b421cc825bd5e6d0.

Re-reviewed the final patch and its integration with the preceding merged stack units. git range-diff confirms patch equivalence to the independently approved stack revision. The authenticated socket supplies ownership, malformed and foreign IDs are rejected, and recovery remains available after all earlier keys are revoked. The replacement secret uses the one-time display. No blocking findings.

Reviewed commit: `214feba72833295edc7c3fa614ba3f504bef8c8a`.

PR #1953, opened 2026-09-11
fix(auth): expire claimed principal credentials after 30 days

jhgaylor:
Claimed principal credentials had no expiry. Give each initial claim, claim replay and owner renewal a 30-day credential lifetime, independent of the anonymous grant deadline. Owners can recover from expiry through the console replacement form in #1952. Update ADR 0044 and the public API description to state that lifetime.

Issuance unit of #1688, stacked on #1952. Five files, +25/-12. A separate migration unit covers existing non-expiring credentials and older processes during rollout.

Validation: the parent fails all three new lifetime assertions (initial claim, replay, renewal). All 68 focused principal, independent-connection race and claim-controller tests pass; all 50 principal tests pass again after the parent check. The generated wire contract is unchanged: 173 operations and 201 schemas. The SDK change is only a regenerated description comment, so this PR uses sdk-no-release and does not change an SDK version. The decisions bundle validates and its regenerated index is unchanged. Full local precommit passed: 5,392 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/principals.ex
+++ b/apps/fountain/lib/fountain/principals.ex
@@ -491,15 +491,17 @@ defmodule Fountain.Principals do
     end
   end
 
-  # The anonymous credential expires with the grant, so a leaked one dies on
-  # the same schedule as the machine it reaches. The claimed one must not:
-  # after a claim the grant's deadline describes nothing — the principal is an
-  # ordinary tenant of the account that owns it — and a key that expired at it
-  # would take a live machine away from its new owner, usually within hours.
+  # Anonymous credentials follow the grant. Claimed credentials get 30 days
+  # from issuance, independently of that old deadline. The owner can replace
+  # them from API keys in the console, including after expiry.
   defp principal_key_opts(claimable, opts) do
     [
       scopes: ["principal"],
-      expires_at: if(claimable.status == "claimed", do: nil, else: claimable.expires_at),
+      expires_at:
+        if(claimable.status == "claimed",
+          do: DateTime.utc_now() |> DateTime.add(30, :day) |> truncate(),
+          else: claimable.expires_at
+        ),
       actor: Keyword.get(opts, :actor, "api"),
       request_ip: Keyword.get(opts, :request_ip)
     ]
--- a/apps/fountain/lib/fountain_web/live/api_keys_live/index.ex
+++ b/apps/fountain/lib/fountain_web/live/api_keys_live/index.ex
@@ -3,6 +3,7 @@ defmodule FountainWeb.ApiKeysLive.Index do
   use FountainWeb, :live_view
 
   alias Fountain.Accounts
+  alias Fountain.Principals
 
   @impl true
   def mount(_params, _session, socket) do
@@ -14,6 +15,7 @@ defmodule FountainWeb.ApiKeysLive.Index do
      |> assign(:page_title, "API keys")
      |> assign(:user_id, user.id)
      |> assign(:keys, keys)
+     |> assign(:principals, Principals.list_owned(user.id))
      |> assign(:new_key, nil)}
   end
 
@@ -35,6 +37,23 @@ defmodule FountainWeb.ApiKeysLive.Index do
 
       {:error, _cs} ->
         {:noreply, put_flash(socket, :error, "Failed to create API key")}
+    end
+  end
+
+  def handle_event("renew_principal_key", %{"principal_id" => id}, socket) do
+    with {:ok, principal_id} <- Ecto.UUID.cast(id),
+         {:ok, {key, raw_token}} <-
+           Principals.renew_owned_credential(
+             socket.assigns.user_id,
+             principal_id,
+             FountainWeb.Audited.attribution(socket)
+           ) do
+      {:noreply,
+       socket
+       |> assign(:keys, Accounts.list_managed_api_keys(socket.assigns.user_id))
+       |> assign(:new_key, %{key: key, raw_token: raw_token})}
+    else
+      _ -> {:noreply, put_flash(socket, :error, "Could not replace principal key")}
     end
   end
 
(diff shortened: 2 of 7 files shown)


BinaryBourbon (approved):
APPROVE — the change is small, correct, and the recovery path it depends on already landed in #1952. But the credential holder is given a deadline it has no way to read, and that is worth fixing before the rollout migration in #1954 makes the expiry real for existing keys.

**What this changes**

`principal_key_opts/2`'s `expires_at` was `nil` for a claimed principal and `claimable.expires_at` for an anonymous one. The claimed branch becomes `DateTime.utc_now() |> DateTime.add(30, :day) |> truncate()`. Because that helper feeds every principal-key mint — the initial claim, the replay rotation from #1950, and `renew_owned_credential/3` from #1951 — all three get a fresh 30-day lifetime from their own issuance instant, with no coupling to the grant deadline. The ADR decision and the `ClaimedPrincipal.api_key` description are amended to match; the only SDK movement is the regenerated TypeScript description comment.

**Findings**

1. **`schemas.ex:4045` — the claim response promises a 30-day lifetime in prose and returns no field the client can read it from, and the holder cannot look it up either.** Non-blocking, but it is the gap I would close first. `ClaimedPrincipal` has `user`, `principal_id`, `status`, `claimed_at`, `api_key` and nothing else; compare `ClaimableUserCreated`, which *requires* `expires_at` so the anonymous holder knows its own deadline. The obvious fallback is closed too: `/api/auth/api-keys` is behind `:require_key_management` (router.ex:350), so a `principal`-scoped key gets 403 there — #1949's controller test asserts exactly that. So a claimed principal cannot discover when its own credential dies by any route, and an integration that wants to renew early has to hardcode `30` from a description string. Adding `api_key_expires_at` to the response is two lines and makes the lifetime part of the contract rather than part of the docs.

2. **`principals.ex:499-505` — nothing warns before day 30, and the credential holder cannot renew itself.** Non-blocking; a consequence of the decision, not a defect in it, but say it out loud somewhere. Replay renewal needs `claimed_by_user_id == claimer.id` plus the original `:idempotency_key`, on a route behind `:require_full_scope`; `renew_owned_credential/3` needs the owner's account. Both are the *owner*, never the application holding the key. So the steady state this creates is: every claimed principal integration starts returning 401 on a 30-day cadence and stays broken until a human opens the console. No email, no `credential_expiring` webhook, no grace. The credits campaign got `credits-low`/`credits-exhausted` mail for the same shape of problem. Worth an issue at least.

3. **The comment this replaces argued the opposite, and the replacement does not say why 30 days wins.** Non-blocking, soft. The deleted lines were a specific argument — "a key that expired at it would take a live machine away from its new owner, usually within hours" — and the new comment asserts the conclusion without engaging it. ADR 0044 decision 4 has the same shape: it now states the 30-day lifetime but not why a bounded credential beats an unbounded one, which is the thing a reader six months from now will want. One sentence in the ADR ("an unbounded credential on a tenant with no human behind it is standing access nobody reviews; the owner can replace it in one click since #1952") turns a restatement into a decision.

4. **`30` is a hardcoded literal where every comparable window in this codebase is an env var.** Non-blocking nit. `CREDIT_OPENING_DAYS`, `CREDIT_NUMBER_CENTS`, the seven-day rent grace, `SANDBOX_RESERVE_CENTS` are all configurable; this one is not, and #1954 hardcodes the same `INTERVAL '30 days'` a second time in SQL, so the two can drift. A module attribute read by both (or a config key) would at least make it one number.

5. Nit: `assert_claimed_key_lifetime/1`'s `remaining in (30 * 86_400 - 5)..(30 * 86_400)` is fine here because the deadline is computed from `DateTime.utc_now()` in Elixir at insert time, so the 5s budget covers only the gap to the assertion. #1954's copy of this helper is *not* equivalent — see that review.

**Checked and clean**

- The anonymous branch is untouched: `ClaimableUser.changeset/2` has `validate_required([… :expires_at])`, so `claimable.expires_at` is never `nil` and an unclaimed key still dies with its grant. The "the anonymous credential expires with the grant" test is unchanged and still passes.
- All three issuance paths really do route through `principal_key_opts/2` — `issue_credentials/2`, `claim_with_credential/4` via `insert_principal_key/2`, and `renew_owned_credential/3` — and the diff adds a lifetime assertion at all three (`principals_test.exs:298`, `:352`, `:433`), so no path is left non-expiring.
- Each issuance recomputes `utc_now() + 30d`, so a replay or a console renewal is a genuine extension rather than a re-stamp of the original deadline. There is no backdating anywhere in this patch.
- `truncate()` after `DateTime.add/3` matches the `timestamp(0)` column, so no rounding surprise on read-back.
- `authenticate_api_key/1` already returns `{:error, :expired}` via `ApiKey.expired?/1` before the suspended and unverified checks, so an expired principal credential fails closed and is distinguishable from a revoked one by the client.
- `expires_at` is already exposed in `ApiKeyJSON.summary/1` and the console table, so the owner can see the deadline on the surface #1949 built — the visibility gap in finding 1 is the principal's, not the owner's.
- The regenerated `sdk/typescript/src/generated/openapi.ts` hunk is a description comment only, no type or required-field movement, so `sdk-no-release` with no version bump is the right call and matches the "OpenAPI is the contract" rule rather than working around it.
- ADR 0044's amendment is in the same PR as the behaviour, with no forward reference to anything unbuilt, and the `decisions/` index needed no regeneration.

BinaryBourbon (approved):
Re-reviewed the rebased expiry change. No blocking findings. The shared issuance helper gives initial claims, replays and owner renewals a fresh 30-day deadline; anonymous credentials retain their grant deadline. Each issuance path has a lifetime assertion, and authentication already rejects expired keys. ADR and generated descriptions agree with the code.

Nonblocking followup: expose the exact credential deadline in the claim response and console. Correction to the earlier review: the current console table shows creation and last-use dates, not expiry.

Reviewed commit: `2a7e35818232a4c0852f31d08a0b74f82f023b75`. Code and test coverage were inspected; fresh CI/local execution is tracked separately.

BinaryBourbon (approved):
APPROVE after rebasing onto current main 3dab11664b36527af2d6baa8fa60a30b7378d1f8.

Re-reviewed the final patch and its integration with the preceding merged stack units. git range-diff confirms patch equivalence to the independently approved stack revision. Initial claims, replay and owner renewal all use the shared 30-day expiry helper, while anonymous credentials retain their grant deadline. API descriptions and ADR agree. No blocking findings.

Reviewed commit: `b55febc590975c6473d1b64dff6300310fdf5c85`.

PR #1954, opened 2026-09-11
fix(auth): bound legacy principal credential expiry during rollout

### LINKED ISSUE #1688: Claimable principals: a claimed principal's API key never expires and its owner cannot revoke it

jhgaylor:
Existing principal credentials can remain non-expiring after the issuer is fixed, and an older process can still write a NULL expiry during deployment. A database backstop assigns a write-time 30-day deadline to principal-scoped writes that omit one. A separate migration gives active legacy credentials a full 30-day recovery window. Explicit deadlines, full/sprite keys and ordinary last-used updates retain their behavior. Rollback retains assigned deadlines.

Final unit of #1688, stacked on #1953. Owners have the replacement form from #1952 before expiry is introduced. Closes #1688 with the preceding owner management, atomic rotation, renewal and issuance units.

Trigger DDL commits before the backfill. Both migrations have 5-second lock and 30-second statement budgets, so contention fails and can be retried. ADR 0044 documents the permanent finite-expiry backstop.

Validation of the repaired migration on PostgreSQL 17: DDL and backfill each failed at the configured 5-second lock timeout under contention; reads remained available during trigger installation, and unrelated reads/writes continued during the separate backfill. A 90-day-old active key received migration-time grace; explicit/full/sprite/revoked exceptions were preserved. Old transactions and a non-UTC session used write-time deadlines. Rollback retained deadlines and a failed backfill succeeded after its blocker was removed. Focused ExUnit and fresh CI results will be recorded before merge.

### DIFF
--- a/apps/fountain/lib/fountain/principals.ex
+++ b/apps/fountain/lib/fountain/principals.ex
@@ -491,15 +491,17 @@ defmodule Fountain.Principals do
     end
   end
 
-  # The anonymous credential expires with the grant, so a leaked one dies on
-  # the same schedule as the machine it reaches. The claimed one must not:
-  # after a claim the grant's deadline describes nothing — the principal is an
-  # ordinary tenant of the account that owns it — and a key that expired at it
-  # would take a live machine away from its new owner, usually within hours.
+  # Anonymous credentials follow the grant. Claimed credentials get 30 days
+  # from issuance, independently of that old deadline. The owner can replace
+  # them from API keys in the console, including after expiry.
   defp principal_key_opts(claimable, opts) do
     [
       scopes: ["principal"],
-      expires_at: if(claimable.status == "claimed", do: nil, else: claimable.expires_at),
+      expires_at:
+        if(claimable.status == "claimed",
+          do: DateTime.utc_now() |> DateTime.add(30, :day) |> truncate(),
+          else: claimable.expires_at
+        ),
       actor: Keyword.get(opts, :actor, "api"),
       request_ip: Keyword.get(opts, :request_ip)
     ]
--- a/apps/fountain/lib/fountain_web/schemas.ex
+++ b/apps/fountain/lib/fountain_web/schemas.ex
@@ -4108,8 +4108,9 @@ defmodule FountainWeb.Schemas do
         api_key: %Schema{
           type: :string,
           description:
-            "A fresh `principal`-scoped key for the claimed principal, with no expiry. " <>
-              "The credential the application held is revoked by the claim."
+            "A fresh `principal`-scoped key that expires 30 days after issuance. " <>
+              "The claim revokes the application credential. Owners can replace the key " <>
+              "from API keys in the console, including after expiry."
         }
       },
       required: [:user, :principal_id, :status, :api_key]
--- a/apps/fountain/priv/repo/migrations/20260911142700_bound_legacy_principal_key_expiry.exs
+++ b/apps/fountain/priv/repo/migrations/20260911142700_bound_legacy_principal_key_expiry.exs
@@ -1,0 +1,43 @@
+defmodule Fountain.Repo.Migrations.BoundLegacyPrincipalKeyExpiry do
+  use Ecto.Migration
+
+  def up do
+    # An older process can still mint a NULL expiry during a rolling deploy.
+    # Give those writes the same deadline as the new issuer, without changing
+    # explicit grant deadlines or the lifetime of full/sprite credentials.
+    execute """
+    CREATE FUNCTION fountain_bound_principal_key_expiry() RETURNS trigger
+    LANGUAGE plpgsql AS $$
+    BEGIN
+      IF NEW.expires_at IS NULL AND 'principal' = ANY(NEW.scopes) THEN
+        NEW.expires_at := date_trunc('second', CURRENT_TIMESTAMP AT TIME ZONE 'UTC')
+                          + INTERVAL '30 days';
+      END IF;
+      RETURN NEW;
+    END;
+    $$
+    """
+
+    execute """
+    CREATE TRIGGER bound_principal_key_expiry
+    BEFORE INSERT OR UPDATE OF scopes, expires_at ON api_keys
+    FOR EACH ROW EXECUTE FUNCTION fountain_bound_principal_key_expiry()
+    """
+
+    # Existing active keys get a full renewal window from migration time,
+    # rather than expiring immediately because they were minted long ago.
+    execute """
+    UPDATE api_keys
+    SET expires_at = date_trunc('second', CURRENT_TIMESTAMP AT TIME ZONE 'UTC')
+                     + INTERVAL '30 days'
+    WHERE expires_at IS NULL AND revoked_at IS NULL AND 'principal' = ANY(scopes)
+    """
+  end
+
+  def down do
+    execute "DROP TRIGGER bound_principal_key_expiry ON api_keys"
+    execute "DROP FUNCTION fountain_bound_principal_key_expiry()"
+    # Retain assigned deadlines: clearing them would also immortalize keys
+    # deliberately issued with the same expiry after this migration.
+  end
+end
--- a/apps/fountain/priv/repo/migrations/20260911142700_bound_legacy_principal_key_expiry.exs
+++ b/apps/fountain/priv/repo/migrations/20260911142700_bound_legacy_principal_key_expiry.exs
@@ -2,6 +2,11 @@ defmodule Fountain.Repo.Migrations.BoundLegacyPrincipalKeyExpiry do
   use Ecto.Migration
 
   def up do
+    # CREATE TRIGGER takes SHARE ROW EXCLUSIVE: reads may continue, but writes
+    # wait. Bound contention and commit the DDL before the separate backfill.
+    execute("SET LOCAL lock_timeout = '5s'")
+    execute("SET LOCAL statement_timeout = '30s'")
+
     # An older process can still mint a NULL expiry during a rolling deploy.
     # Give those writes the same deadline as the new issuer, without changing
     # explicit grant deadlines or the lifetime of full/sprite credentials.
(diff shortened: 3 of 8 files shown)


BinaryBourbon (changes requested):
REQUEST_CHANGES — the expiry semantics are right (no backdating, nobody is locked out), but the migration takes an unguarded `ACCESS EXCLUSIVE` lock on `api_keys` and then does an unbounded backfill inside the same transaction. This repo already has the convention that prevents that, one migration earlier.

**What this changes**

Two things in one `up/0`. A `BEFORE INSERT OR UPDATE OF scopes, expires_at` trigger on `api_keys` rewrites `NEW.expires_at` to `now + 30 days` whenever it is `NULL` and `'principal' = ANY(NEW.scopes)`, so a pre-#1953 process that is still serving during a rolling deploy cannot mint an immortal principal credential. Then a one-shot `UPDATE` gives every currently active, non-expiring, principal-scoped key a deadline 30 days from **migration time** — not 30 days from its own issuance, and not immediate expiry. `down/0` drops the trigger and function and deliberately keeps the assigned deadlines.

**On the question of what happens to a key issued the day before the migration**

It gets a full 30 days measured from `CURRENT_TIMESTAMP` at migration, so the answer is neither "30 from issue" nor "immediate". There is no path here that shortens a key's life relative to what its holder already had, and nothing is backdated — the `WHERE expires_at IS NULL` predicate cannot touch a key that already carries a deadline. That is the right call and the migration comment says so.

**Findings**

1. **`20260911142700_bound_legacy_principal_key_expiry.exs:18` — `CREATE TRIGGER … ON api_keys` needs `ACCESS EXCLUSIVE` on the table every authenticated request reads, with no `lock_timeout`, and the backfill runs in the same transaction. Blocking.**

   Failure scenario, using this PR's own premise that the old release is serving traffic while the migration runs: any in-flight transaction that has touched `api_keys` holds `ACCESS SHARE` — that includes `Principals.claim/4`, which since #1950 holds the claimable `FOR UPDATE` lock *and* does `update_all` + `INSERT` on `api_keys` in one transaction, and includes `issue_credentials/2`. `CREATE TRIGGER` cannot get `ACCESS EXCLUSIVE` until that commits, and while it waits Postgres queues every *new* lock request behind it. So the next `SELECT … FROM api_keys WHERE key_hash = $1` blocks, and so does every one after it: authentication for the entire estate — API, console, sprite callbacks — stalls for as long as the blocker lives. With no `lock_timeout` that stall is unbounded rather than failing fast and being retried.

   Then, still inside the same transaction and still holding the lock, the backfill scans for `expires_at IS NULL AND revoked_at IS NULL AND 'principal' = ANY(scopes)`. There is `index(:api_keys, [:expires_at])` but none on `scopes`, and the `NULL` bucket contains every `full`-scoped key ever minted (the 20260801 migration backfilled `scopes` to `["full"]` and left `expires_at` NULL), so both the plan and the row count are unbounded by construction. The whole scan happens with auth locked out.

   This repo already solved this, in the migration three before it — `20260906020000_preserve_log_event_microseconds.exs`:

   ```elixir
   execute("SET LOCAL lock_timeout = '5s'")
   execute("SET LOCAL statement_timeout = '30s'")
   ```

   with the comment "fail promptly under contention and retry the migration once the blocker is gone, rather than queueing writes forever." That is exactly this situation on a hotter table. Please add the same two `SET LOCAL` lines, and consider splitting the backfill into its own migration after the DDL so the scan does not run under the exclusive lock at all (`@disable_ddl_transaction` + a batched loop is the other shape, and `20260803171945_index_log_events_inserted_at.exs` is the precedent for that).

2. **The trigger is framed as a rollout measure and nothing ever removes it.** Non-blocking, but decide it now rather than discovering it later. The title says "during rollout" and the body says "an older process… during a rolling deploy", yet there is no follow-up migration, no issue, and no `stale_after`-style marker. Once the deploy is done this is a permanent database-level policy with two consequences worth naming in the ADR (which currently describes the trigger as a fact but not as permanent):

   - A future deliberate decision to mint a non-expiring principal credential becomes impossible from the application layer, and the code will not say why. The changeset will carry `expires_at: nil`, the insert will succeed, and the row will quietly hold a deadline.
   - `Repo.insert/1` returns the changeset-applied struct, not the row, so any such path gets back `%ApiKey{expires_at: nil}` while the database holds a deadline. Your own test acknowledges this (it uses `Repo.reload!` and comments that the old issuer "did not… request that generated value back"), which is the tell that the divergence is real. It is harmless today only because #1953 sets the value explicitly on every live path.

   Either add the drop as a follow-up unit once the old release is gone, or keep it and say in ADR 0044 that the 30-day bound is enforced in the database as a permanent invariant, not a rollout patch.

3. **`principal_key_expiry_test.exs:62` — `assert_deadline/1`'s 5-second budget is measured from the sandbox transaction's `BEGIN`, not from the insert.** Non-blocking flake risk. The trigger uses `CURRENT_TIMESTAMP`, which in Postgres is the *transaction* timestamp, and under `Fountain.DataCase` the whole test body runs inside the one sandbox transaction opened at checkout. So `remaining` is `30 days - (everything this test did before the assertion, including `insert_verified_user/0` and its DEK wrap)`. This is not the same as #1953's identical-looking helper, where the deadline comes from `DateTime.utc_now()` in Elixir at insert time and the 5s covers only the gap to the assertion. On a loaded partition the preamble eating 5s is not absurd, and the failure would look like a wrong interval rather than a slow test. `clock_timestamp()` in the trigger removes the coupling entirely and is also the more accurate thing to store — a production insert inside `claim_with_credential/4` that waited on the claim lock currently gets a deadline short by the length of that wait.

4. Nit: `INTERVAL '30 days'` appears twice here and `DateTime.add(30, :day)` once in `principals.ex`, three copies of one policy number in two languages. See finding 4 on #1953.

**Checked and clean**

- `UPDATE OF scopes, expires_at` is the correct narrowing: it fires only when a column is named in the `SET` list, so `touch_api_key/1` (`set: [last_used_at: …]`), `revoke_api_key/3` and `revoke_claimed_credentials/1` (`set: [revoked_at: …]`) do not fire it. The "ordinary key use does not extend an expired principal credential" test pins that, and asserts the expired key still fails closed.
- `CURRENT_TIMESTAMP AT TIME ZONE 'UTC'` is the right incantation for a `:utc_datetime` column (`timestamp(0) without time zone`), and `date_trunc('second', …)` truncates before the column's precision-0 *rounding* could move it. No timezone or off-by-one-second hazard.
- The trigger cannot touch the anonymous credential: `ClaimableUser.changeset/2` has `validate_required([… :expires_at])`, so `principal_key_opts/2`'s unclaimed branch never passes `nil`, and "the anonymous credential expires with the grant" is unaffected. The explicit-deadline test asserts it directly.
- `'principal' = ANY(NEW.scopes)` is the only scope touched; the `for scope <- ["full", "sprite"]` case proves `full` and `sprite` keys stay unbounded, which matters because OAuth tokens are `api_keys` rows too.
- `principals.ex:499` is the sole `scopes: ["principal"]` site in the tree, so the trigger's blast radius is exactly the claimable-principal family and nothing else.
- The backfill's predicate excludes revoked keys, so released and expired claimables — whose keys `close/5` already revoked — are not resurrected with a deadline.
- `down/0` retaining assigned deadlines is correct and correctly reasoned: clearing them would also immortalize keys deliberately issued with that deadline *after* the migration, which is a worse outcome than a rollback that leaves some keys expiring. The comment says so.
- The trigger is a genuine backstop rather than the mechanism: every live path sets `expires_at` explicitly since #1953, so the two agree and reverting either one is still caught. The PR body reports the rolled-back-migration check, and the two `update_all` regressions are the right shape for it.
- ADR 0044's amendment lands with the behaviour and describes nothing unbuilt.
- No `CHANGELOG.md` `[Unreleased]` entry, consistent with the rest of the programme — noted, not counted against this PR.

BinaryBourbon (approved):
APPROVE — re-reviewed the rebased migration repair and resolved the earlier requested changes.

Trigger installation and legacy backfill now run in separate bounded transactions. Explicit deadlines, non-principal keys and revoked-key exceptions are preserved; rollback retains assigned deadlines. The trigger uses write-time clock_timestamp(), with database-clock bounds and an aged-transaction regression. ADR 0044 documents permanent enforcement and the returned-struct limitation.

Correction to the previous review: CREATE TRIGGER takes SHARE ROW EXCLUSIVE, not ACCESS EXCLUSIVE. Reads can continue; separating the backfill prevents its scan from extending the DDL lock that blocks writes.

Standalone PostgreSQL validation against the final SQL passed contention timeouts, actual lock modes/concurrent access, grace/exemptions, clock/timezone behavior, rollback preservation and backfill retry. No blocking code findings remain. Merge remains contingent on passing current-head CI and the focused application tests.

Reviewed commit: 61b02a14731a2ad4fd5a5a7a378a161a380edee2

BinaryBourbon (approved):
APPROVE after rebasing onto current main af1178dd2b3462eb7a26e9d655ff3fe09681a0ed.

Re-reviewed the final patch and its integration with the preceding merged stack units. git range-diff confirms patch equivalence to the independently approved stack revision. Separate bounded trigger/backfill transactions resolve the requested locking changes. Write-time expiry, legacy grace, scope exemptions and deadline-preserving rollback are intact. PostgreSQL contention/concurrency/rollback/retry validation passed. No blocking findings.

Reviewed commit: `5165e064d83db257fe1e18c8dcc79d12a3d77a86`.

BinaryBourbon:
Final validation note: the non-required SDK release guard initially failed while #1954 was prepared on #1953’s queued commit. At that point its main comparison included #1953’s generated API-description change, which is intentionally covered by sdk-no-release on #1953.

After the parent merged, #1954’s own diff contains zero SDK files. The unchanged guard now passes locally against origin/main (0 touched SDK files; no release needed). Rerunning the failed guard against that corrected base comparison; required CI and merge-group CI both passed before merge. No SDK version change is needed.

Issue #1710, opened 2026-09-07
Model enforcement: a runtime that advertises neither configOptions nor models now fails every turn, and opencode's adapter is unpinned


jhgaylor:
Two exposures left after #1640/#1663 (managoat_acp 0.2.2):

1. `peer.ex:1048-1053` fails the turn when the runtime advertises neither `configOptions` nor `models`; 0.1.1 wrote a stderr note and prompted on the default (`0.1.1 peer.ex:987-993`). opencode's adapter is unpinned (`managoat_runtimes` `acp.ex:102`), so an upstream change to its models advertisement takes every opencode agent down at once, with the same "one runtime kept working" disguise as the 09-06 outage. Pin it, or make "nothing advertised" a warning for runtimes that cannot select a model.
2. The production case (claude-opus-5 confirmed as `opus`, claude-sonnet-5 as `sonnet`) is pinned only in managoat_acp's `test/managoat/acp/peer_test.exs:1256-1294`. Fountain's own test (`conversation_server_acp_test.exs:299-333`) covers a verbatim echo only, so a library regression would reach `main` through a version bump with nothing here red. Add the alias case to Fountain's test.

`claude-fable-5-1` in the catalog is #1669; the `[1m]` suffix is #1668; the haiku alias is #1652.

---
Found in the 2026-09-07 review of the week on `main` (cb93a7ea).

PR #1955, opened 2026-09-11
test(acp): cover Claude model aliases through conversations

jhgaylor:
Fountain's conversation integration tests confirmed only a verbatim model echo, leaving Claude's production `opus` and `sonnet` aliases protected only inside the ACP dependency. Drive the real peer through Claude's `session/set_config_option` response and assert that both aliases complete a turn with the requested/effective model evidence preserved in the API. A `haiku` response to an Opus request must still fail before any prompt is sent.

Regression unit of #1710. One test file, +54/-0. The independent OpenCode pinning exposure remains separate.

Validation: all 105 ACP conversation integration tests pass. Replacing alias matching with strict normalized equality inside the isolated dependency checkout makes both positive regressions fail (3 tests, 2 failures); restoring and rebuilding the original dependency returns all three to green. The mismatched-family case stays rejected. No dependency changes are committed. Full local precommit passed: 5,378 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/test/fountain/conversations/conversation_server_acp_test.exs
+++ b/apps/fountain/test/fountain/conversations/conversation_server_acp_test.exs
@@ -143,6 +143,60 @@ defmodule Fountain.Conversations.ConversationServerACPTest do
     prompt_id
   end
 
+  describe "Claude model confirmation aliases (#1710)" do
+    for {model, confirmed, accepted?} <- [
+          {"claude-opus-5", "opus", true},
+          {"claude-sonnet-5", "sonnet", true},
+          {"claude-opus-5", "haiku", false}
+        ] do
+      test "#{model} confirmed as #{confirmed}", %{} do
+        model = unquote(model)
+        confirmed = unquote(confirmed)
+        user = insert_verified_user()
+        agent = insert_agent(user_id: user.id, runtime: "claude", model: "anthropic/" <> model)
+        conv = insert_conversation(agent: agent, user_id: user.id)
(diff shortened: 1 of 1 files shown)


BinaryBourbon (commented):
Reviewed `803c65dfee89bea471bc59a64f065e19ce36898a`.

No actionable findings. The new tests exercise Opus/Sonnet confirmation aliases, preserve requested/effective-model evidence through the API representation and reject a mismatched Haiku confirmation before prompting. The independent OpenCode pinning/nothing-advertised concern from #1710 is explicitly outside this PR's scope.

Validation: static review of all three new cases, the integration helper and model-selection behavior; ACP integration tests were not rerun.

BinaryBourbon (approved):
**Approved** at `803c65dfee89bea471bc59a64f065e19ce36898a`.

The three new cases cover production Opus/Sonnet aliases and a mismatched-family rejection through the real peer helper. The positive assertions preserve requested/effective model evidence through the API serializer; the negative case checks the turn fails without a prompt. Rechecked helper synchronization and the dependency's model-confirmation path. No blocking findings. The independent OpenCode pinning/nothing-advertised exposure is explicitly separate work.

Validation: [CI run 34612392572](https://github.com/managoat/fountain/actions/runs/34612392572) is verified at this exact head; six core partitions pass (5,385 tests in downloaded logs) and required/static/contract gates are green. I reviewed the integration source and logs, rather than running new local ACP integration tests.

Issue #1671, opened 2026-09-06
Investigate September 6 broker CA-install failure spike


jhgaylor:
## Resolution — September 18, 2026

Resolved by the investigation and focused fix in #2400. Production ran merge commit `1103fe84`; the post-rollout exercise recorded **7 successful broker setups, including 6 overlapping wakes on one shared Sprites sandbox, and zero broker failures**. The closure comment records rollout, monitoring, cleanup, and evidence limits. The separate later runtime exit 139 is tracked in #2402.

## Original investigation scope

Investigate the September 6 broker CA-install failure spike and establish whether it still occurs. The original success-reporting and alerting work had already landed; this issue owned the regression investigation and any focused fix supported by its findings.

The approximately 18% rate below is a historical observation from 2026-09-06, not a verified current failure rate.

## Completed work

- [x] Stop publishing `broker` / `done` before CA installation succeeds: #1956 (`d34e834e`).
- [x] Count conversation CA-install outcomes by provider, excluding probes and session refreshes: #1957 (`6ab30355`).
- [x] Add the portable failure-ratio alert: #1958 (`1a463d98`).
- [x] Add the alert to the hosted overlay: https://github.com/jhgaylor/home-cloud/pull/228 merged on 2026-09-17, closing #2041. Production metric scraping was confirmed on September 12 and 13. Flux reconciliation and the live alert were verified; the alert was healthy and inactive after the September 18 concurrent production exercise.

## Investigation

1. Correlate the September 6 failures with deployment/image SHA, sandbox provider, sandbox identity, and available command output. The original report identified eight deploys and a base-image-adjacent change set that day.
2. Compare failures before and after the **deployment** of #1683, merged September 7. That change fixed concurrent CA-bundle rebuilding on shared sandboxes and may explain the incident, but no causal link has been established for this issue.
3. Identify the failing command, not only the outer exit code. `install_broker_ca/2` chains CA installation, sudoers setup, and Git configuration; `ca_install_exit` with exit code 1 does not identify which command failed.
4. Distinguish the failure from the bundled-roots gap fixed in #1175 and the former Agent Vault CONNECT limiter in #1206. The latter was retired by the native-broker switch before this spike; do not treat its upstream issues as prerequisites without evidence that an affected deployment still used Agent Vault.
5. Check the current failure ratio and hosted alert, and document whether the failure persists. Keep probe/health traffic out of the denominator and distinguish installer exits from transport failures.

## Completion criteria and PR boundary

[...]

PR #1956, opened 2026-09-11
fix: complete broker stage after CA installation succeeds

jhgaylor:
Broker setup currently reports `done` as soon as its proxy session is minted, even if installing the CA then fails. Complete the stage only after sandbox trust installation succeeds; preserve the existing failure event and completion metadata.

First small unit of #1671: two files, +49/-17. The CA-installation metric, ratio alert and production regression investigation remain separate; this PR does not claim to resolve the September 6 root cause.

Validation: 72 focused tests passed (one Linux-only trust-store case skipped on macOS). Restoring the parent implementation makes two stage-order regressions fail; restoring this change returns all 21 egress tests to green. Full `mix precommit` passed: 5,376 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/egress.ex
+++ b/apps/fountain/lib/fountain/conversations/egress.ex
@@ -243,9 +243,10 @@ defmodule Fountain.Conversations.Egress do
   def sandbox_env(session), do: Broker.sandbox_env(session)
 
   @doc """
-  Mint (or re-mint) the conversation's proxy session, publishing the
-  `broker` stage around it. The caller decides whether the conversation is
-  brokered at all (`brokered?/1`) and holds the session that comes back.
+  Mint the conversation's proxy session and start the `broker` stage.
+  The stage completes only after `install_ca/3` establishes trust in the
+  sandbox. The caller decides whether the conversation is brokered at all
+  (`brokered?/1`) and holds the session that comes back.
   """
   @spec prepare(String.t(), map(), Broker.bindings(), keyword()) ::
           {:ok, map()} | {:error, term()}
@@ -256,11 +257,6 @@ defmodule Fountain.Conversations.Egress do
 
     case Broker.prepare(conversation_id, brokered, bindings, opts) do
       {:ok, session} ->
-        publish_stage(conversation_id, "broker", "done", %{
-          vault: session.vault,
-          expires_at: session.expires_at
-        })
-
         {:ok, session}
 
       {:error, reason} ->
@@ -485,8 +481,16 @@ defmodule Fountain.Conversations.Egress do
   @spec install_ca(session(), Managoat.Sandbox.Handle.t(), String.t()) :: :ok | {:error, term()}
   def install_ca(nil, _handle, _conversation_id), do: :ok
 
-  def install_ca(_session, handle, conversation_id),
-    do: Provisioning.install_broker_ca(handle, conversation_id)
+  def install_ca(session, handle, conversation_id) do
+    with :ok <- Provisioning.install_broker_ca(handle, conversation_id) do
+      publish_stage(conversation_id, "broker", "done", %{
+        vault: session.vault,
+        expires_at: session.expires_at
+      })
+
+      :ok
+    end
+  end
 
   # Every session of the conversation goes when its sandbox does. Still off
   # the caller's path: deleting rows is local and cannot fail the way a call
--- a/apps/fountain/test/fountain/conversations/egress_test.exs
+++ b/apps/fountain/test/fountain/conversations/egress_test.exs
@@ -297,7 +297,7 @@ defmodule Fountain.Conversations.EgressTest do
   end
 
   describe "the session" do
-    test "prepare/4 publishes the broker stage around the mint", %{user: user} do
+    test "the broker stage completes only after CA installation", %{user: user} do
       broker_on([user.id])
       conv = insert_conversation(user_id: user.id)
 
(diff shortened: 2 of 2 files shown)


BinaryBourbon (approved):
APPROVE — the completion moves to the one place that knows trust is actually installed, and both call sites reach it. Two non-blocking notes, one of which matters for #1958's alert.

**What this changes**

`Egress.prepare/4` no longer publishes `broker`/`done` the instant `Broker.prepare/4` mints a session; it publishes only `started`, and `install_ca/3` (`apps/fountain/lib/fountain/conversations/egress.ex:484`) publishes `done` inside a `with` that requires `Provisioning.install_broker_ca/2` to return `:ok` first. `done` therefore now asserts two things — a session exists *and* the sandbox trusts the broker's CA — where before it asserted only the first, which is why a conversation whose `update-ca-certificates` exited non-zero previously emitted `started, done, failed` for the same stage. Failure reporting is unchanged because `install_broker_ca/2` already publishes `broker`/`failed` from all three of its own failure arms.

**Findings**

1. **Non-blocking — `started` can now be the last broker event when provisioning fails between the mint and the install.** `prepare_state/1` publishes `started` at `conversation_server.ex:923` and `install_ca/3` is not reached until `:949`; in between sits `:ok <- Provisioning.write_runtime_config(...)`. Scenario: brokered tenant, sandbox created, the `.mcp.json` write returns `{:error, :timeout}` → the `else` arm logs, destroys the sandbox, `Egress.release_prepared/1` revokes the token, `provision`/`failed` is published — and the broker stage's last event is `started`, forever. Same window on the wake path: `prepare_state/1` at `:1107` precedes `Egress.reattach_policy/4` at `:1108`, whose failure lands in the transient arm without a terminal broker event. Before this patch those cases ended on `done` (wrong, but terminal); now they end on nothing. A webhook subscriber filtered to `conversation.broker.*` sees `started` with no pair, and a UI that pairs a stage's `started` with its terminal event shows the broker step running indefinitely on a conversation that has already failed. `release_prepared/1` is the natural place to publish `broker`/`failed` — it is already the function that knows a minted session is being thrown away. **Worth flagging into #1958:** if the alert there is a `done`/`started` ratio, these unterminated starts are now a source of ratio drift that is not a CA failure.

2. **Nit — `install_ca/3`'s own docstring does not say it completes the stage.** `egress.ex:481` still reads "Install the broker's CA into the sandbox; nothing to install without a session." The knowledge lives only in `prepare/4`'s `@doc` (`:246`). That matters because the wake call site treats `install_ca/3` as best effort ("the turn's own failure says more than a refused wake would", `conversation_server.ex:1136`), and it is now also the sole publisher of `broker`/`done`. One clause on `install_ca/3` would stop the next reader from moving or short-circuiting it.

3. **Soft — no CHANGELOG `[Unreleased]` entry, and `docs/reference/webhooks.md:65` now under-describes `done`.** The docs line says the stage covers Fountain preparing the egress broker; `done` now additionally means the CA is in the sandbox's trust store, and the event fires measurably later in provisioning. Neither statement is false, so this is not blocking, but the timing shift is visible to webhook subscribers and the `broker` stage is documented API surface.

4. **Test nit** — `egress_test.exs:321` puts the "no `done` yet" ordering assertion *inside* the Mimic `stub` for `install_broker_ca/2`. A stub that is never invoked asserts nothing; `expect` would make the call itself part of the contract. The guard still holds either way: with the fix reverted, the `assert [{"started", _}] = stages(...)` at `:318` fails before the stub ever runs.

**Checked and clean**

- Both new tests fail on a revert, for the right reason: `:300` asserts the stage set is exactly `[started]` after `prepare/4` returns (reverted: `[started, done]`), and `:331` asserts `[started, failed]` on a `bash` exit 1 (reverted: `[started, done, failed]`). Neither asserts on something it also stubs — the second drives the real `install_broker_ca/2` and stubs only the sandbox transport.
- `done` is published from exactly one place after this patch: grep for `"broker"` publishers gives `egress.ex` (`started`, `done`) and `provisioning.ex:464/469/474` plus `check_broker_support/4`'s two (`failed` only). No second path completes the stage.
- Both `install_ca/3` call sites (`conversation_server.ex:949` provision, `:1139` wake) are preceded by `prepare_state/1`, so `done` cannot appear without its `started`; and the `install_ca(nil, ...)` clause publishes nothing, matching `prepare_state/1`, which publishes nothing when `brokered?/1` is false. That is what `:421` now pins.
- A CA failure still terminates the stage on both call sites — including the best-effort wake one, where the error is only logged — because `install_broker_ca/2` publishes `failed` itself before returning. This is the reason finding 1 is limited to the *other* steps in the window.
- No duplicate `done` under a shared sandbox (ADR 0023): each conversation mints and publishes on its own stream. `reprepare/5` and `refresh_rules/4` still publish no stage, as their docstrings say.
- `stages/2` orders by `log_events.id`, which is bigserial (`create_initial_schema.exs:92`, no `primary_key: false`), and `publish_stage/4` inserts synchronously before broadcasting — so the two-element sequence assertions are deterministic rather than racing an async write. Not a flake source.
- `webhooks/events.ex:25` still needs no change: the stage keeps `started done failed`, and the source-scanning `webhook_events_test` guard follows the moved call site.
- No convention issues: no tenant scoping or `_unsafe_` involvement, no audit call moved into a transaction, no `Task.async`, pool size untouched, and `egress_test.exs` was already `async: false` for the documented application-env reason.
- Scope matches the body's claim: two files, no metric and no alert, and nothing here asserts the September 6 root cause is fixed.

PR #1957, opened 2026-09-11
feat(metrics): count conversation broker CA installation outcomes

jhgaylor:
Broker stage events cannot provide a reliable CA-installation failure denominator. Emit one `fountain_broker_ca_install_count` sample after each conversation trust-setup attempt, labeled only by provider and the bounded outcome (`ok`, `exit`, `unreachable`, `unavailable`). Session minting, refreshes, unbrokered conversations and proxy/health requests do not increment it.

Second small unit of #1671, stacked on #1956. Four files, +67/-1. The ratio alert is a separate child PR; identifying the September 6 production regression remains open.

Validation: 43 focused tests pass, covering all four outcomes, excluded paths, the exported metric subscription and producer inventory. Full `mix precommit` passed: 5,377 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/egress.ex
+++ b/apps/fountain/lib/fountain/conversations/egress.ex
@@ -482,7 +482,17 @@ defmodule Fountain.Conversations.Egress do
   def install_ca(nil, _handle, _conversation_id), do: :ok
 
   def install_ca(session, handle, conversation_id) do
-    with :ok <- Provisioning.install_broker_ca(handle, conversation_id) do
+    result = Provisioning.install_broker_ca(handle, conversation_id)
+
+    # Only a conversation's trust setup reaches this path. Broker health
+    # probes and session refreshes do not contribute to the failure ratio.
+    Fountain.Telemetry.event(
+      [:broker, :ca_install],
+      %{provider: handle.provider, outcome: ca_install_outcome(result)},
+      %{count: 1}
+    )
+
+    with :ok <- result do
       publish_stage(conversation_id, "broker", "done", %{
         vault: session.vault,
         expires_at: session.expires_at
@@ -491,6 +501,11 @@ defmodule Fountain.Conversations.Egress do
       :ok
     end
   end
+
+  defp ca_install_outcome(:ok), do: "ok"
+  defp ca_install_outcome({:error, {:broker, :ca_install_exit, _, _}}), do: "exit"
+  defp ca_install_outcome({:error, {:broker, :ca_install, _}}), do: "unreachable"
+  defp ca_install_outcome({:error, _}), do: "unavailable"
 
   # Every session of the conversation goes when its sandbox does. Still off
   # the caller's path: deleting rows is local and cannot fail the way a call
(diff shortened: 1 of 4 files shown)


BinaryBourbon (approved):
**APPROVE.** Reviewed against `codex/p2-1671-broker-stage` (the PR's own patch, four files). The counter is emitted at the one place a conversation's trust setup happens, its labels are bounded, and the emitter cannot raise on any shape `install_broker_ca/2` can return.

### What I verified

**The argument order is right, which is the easy one to get wrong here.** `Fountain.Telemetry.event/3` is `(name, metadata, measurements)` — not `:telemetry.execute/3`'s order — so `%{provider:, outcome:}` lands as metadata (the tags) and `%{count: 1}` as the measurement `counter("fountain.broker.ca_install.count", ...)` reads off the metric name's last segment. Both match.

**`ca_install_outcome/1` is total over the callee's real return set.** `Provisioning.install_broker_ca/2` (`provisioning.ex:485`) returns exactly `:ok`, `{:error, {:broker, :ca_install_exit, code, out}}`, `{:error, {:broker, :ca_install, reason}}`, and — through its `with/else` — `{:error, reason}` from `ca_pem/0` or the staging write. The three specific clauses plus the `{:error, _}` catch-all cover all four, and the two `:broker` clauses have different tuple arities so neither shadows the other. An unexpected shape would raise in `install_ca/3` on the provisioning path rather than inside a `:telemetry` handler, so I checked it can't happen: the callee's own `with/else` is what bounds it.

**No values reach the trail.** The `exit` clause discards the 500-char sandbox output the error tuple carries. Only `provider` — a bounded atom from `%Managoat.Sandbox.Handle{}`, the same vocabulary `Fountain.SandboxProviders.known_providers/0` enumerates — and one of four literal outcome strings become labels. Nothing keyed by conversation, user or sandbox id. Cardinality is 4 x providers.

**Placement gives the denominator the right population.** The emit is before the `with`, so a failure is counted, not just a success. `install_ca(nil, _, _)` short-circuits ahead of it, so an unbrokered conversation contributes nothing. I confirmed `prepare/4` and `reprepare/5` never route through `install_ca/3`, which is what the test's `refute_received` after both calls pins, and that the only two callers (`conversation_server.ex:949` and `:1139`) always hold a real `%Handle{}` from a provision or wake.

**The guard guards.** Deleting the `Fountain.Telemetry.event(...)` block and re-running fails `EgressTest`'s new test — it is not a test that passes whether or not the metric is emitted. The 72 tests across `egress_test.exs`, `metrics_test.exs` and `docs_test.exs` pass on the branch as shipped.

### Two things worth knowing, neither blocking

- **The wake path re-installs the CA on every reattach** (`conversation_server.ex:1139`, best-effort), so the denominator includes idempotent re-installs, not only first provisions. That is the conservative direction for a ratio alert — a healthy re-install dilutes toward quiet — but "attempts" in the child alert's description is broader than "new conversations".
- The `metrics_test.exs` producer-inventory row is a hand-maintained note, not a runtime assertion: removing the emit does not fail that file (only `EgressTest` catches it). True of every row in that list, not a regression here.

PR #1958, opened 2026-09-11
alert: detect broker CA installation failure ratios by provider

jhgaylor:
Add `FountainBrokerCAInstallFailureRate` to the portable rules: nonzero CA installer exits exceed 10% of actual conversation CA setups for a provider over one hour, with at least five attempts and ten minutes pending. A busy healthy provider cannot dilute another provider's failures, and proxy/health traffic never enters the denominator. Document the rule in the operations guide.

Third small unit of #1671, stacked on #1957. Three files, +46/-1. This supplies the portable alert; the hosted configuration and September 6 root-cause investigation remain separate.

Validation: promtool validates all 22 rules and passes all 54 expression cases, including nine new CA cases for missing/idle series, low volume, exactly 10%, the observed 18%, all failures, transport failures, counter resets and provider isolation. The documentation/metric focused suite passes all 50 tests. Full `mix precommit` passed: 5,377 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures. Prose checks find no new errors; the existing two repository style findings remain elsewhere, and destink passes the edited page.

### DIFF
--- a/deploy/k8s/prometheusrule.yaml
+++ b/deploy/k8s/prometheusrule.yaml
@@ -199,6 +199,26 @@ spec:
             summary: "Fountain conversation stages are failing"
             description: "Stage {{ $labels.stage }} failed more than twice in 30 minutes. Inspect conversation stages and provider health."
 
+        - alert: FountainBrokerCAInstallFailureRate
+          # Actual conversation trust setup only: proxy requests and health
+          # probes never emit this metric. Five attempts suppress idle noise.
+          expr: |
+            (
+              round(
+                sum by (provider) (increase(fountain_broker_ca_install_count{namespace="fountain", outcome="exit"}[1h]))
+                / sum by (provider) (increase(fountain_broker_ca_install_count{namespace="fountain"}[1h])),
+                0.000001
+              ) > 0.10
+            )
+            and
+            sum by (provider) (increase(fountain_broker_ca_install_count{namespace="fountain"}[1h])) >= 5
+          for: 10m
+          labels:
+            severity: warning
+          annotations:
+            summary: 'Broker CA installation failures exceed 10% for {{ $labels.provider }}'
+            description: 'More than 10% of at least five conversation CA setup attempts failed with a nonzero installer exit over the last hour. Inspect broker stage failures and sandbox provider/base-image changes.'
+
         - alert: FountainTurnFailureRate
           # Histogram counts are events, so sum across replicas. Require ten
           # terminal turns per provider to avoid paging on one failed attempt.
--- a/docs/guides/operate/observability.md
+++ b/docs/guides/operate/observability.md
@@ -60,6 +60,7 @@ thresholds or selectors in the overlay.
 | `FountainDatabasePoolSaturated` | Requests wait for database connections. |
 | `FountainProvisionFailures` | Sandbox provision attempts fail. |
 | `FountainStageFailures` | A stage after provisioning fails more than twice in 30 minutes. |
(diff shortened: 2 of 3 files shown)


BinaryBourbon (approved):
**APPROVE.** Reviewed against `codex/p2-1671-ca-metrics` (the PR's own patch, three files). The ratio has a real denominator guard, the fixtures cover the cases that usually make a ratio alert lie, and I re-ran them locally rather than taking the description's word for it.

### What I verified

**The denominator guard is the part that matters and it is present.** `and sum by (provider) (increase(...[1h])) >= 5` — a provider with no attempts produces no sample on the right-hand side, so the whole expression is empty and 0/0 never pages. `promtool` agrees: `python3 scripts/test-alerts.py` locally gives `22 rules found / SUCCESS / Passed 54 alert expression cases`, including the `no series`, `idle` and `too little traffic` fixtures.

**The bare `and` is correct here even though the rule above it spells out `on (provider)`.** Both sides reduce to `sum by (provider)`, so the default all-labels match is `{provider}` and the two forms are equivalent. Purely a cosmetic inconsistency with `FountainTurnFailureRate`.

**No zero-fill, correctly.** `FountainTurnFailureRate` needs `or 0 * sum by (provider)(...)` because an absent failure series there would silently drop a provider from the healthy comparison. Here an absent `outcome="exit"` series should yield no sample at all, which is exactly what the `healthy, no failure series` fixture asserts. The mirror case, `all fail, no success series`, still fires because numerator and denominator are the same series.

**`round(..., 0.000001)` follows the precedent set two rules down** for the float-noise-at-exactly-the-threshold case, and the fixtures pin both sides of it: `exactly 10 percent` does not fire, `18 percent fail` (the observed September 6 figure) does.

**Cardinality and cross-provider isolation.** Grouping is `by (provider)` only. The fixtures add a `provider="e2b"` series at 1000/min and a `fountain_broker_request_count` series, and neither moves the sprites verdict — so a busy healthy provider cannot dilute a failing one, and proxy traffic is not in the denominator. That last part is structural rather than incidental: #1957 emits this counter only from `Egress.install_ca/3`, never from session mint, refresh or a proxy request.

**The docs row is load-bearing.** `Fountain.DocsTest`'s "the portable alert table matches the shipped rule names" diffs `- alert:` names in the manifest against the `### Portable baseline` table. I deleted the new table row and re-ran: it fails. So the row is a real guard, not decoration.

`for: 10m` against the `fountain-sandboxes` group's `interval: 60s` is ten evaluations, and the 1h `increase()` window holds the value long enough to satisfy it.

### Three notes, none blocking

- **The hosted instance does not get this rule from this PR, and that is the deployment the September 6 regression happened on.** `deploy/k8s/prometheusrule.yaml` is commented out of `deploy/k8s/kustomization.yaml` and is the self-host baseline; the hosted rules live in the home-cloud overlay at `platform/fountain-site/prometheusrule.yaml`. The PR body and the page both say the hosted config is separate, so this is a flag for sequencing, not a change request: until a home-cloud change lands, this alert watches nobody's production.
- **One failure in five attempts fires it** (20% > 10%, 5 >= 5). That is materially more sensitive than `FountainTurnFailureRate`'s >25%-of-at-least-10. Defensible for an operation that should essentially never fail, at `severity: warning` — but if prod runs near five CA installs an hour in a quiet window, expect this to speak on a single flaky sandbox. Worth watching before anyone routes it anywhere noisy.
- **The numerator is `outcome="exit"` only**, so a provider failing 100% of installs on transport errors (`unreachable`/`unavailable`) scores 0 on this rule while inflating its denominator. `FountainStageFailures` covers that case — every one of those paths calls `publish_stage(conv_id, "broker", "failed", ...)` — and the doc row is honest about the scope ("fail with an installer exit"). Noting it so nobody later reads this alert as total CA-install health.

Issue #1052, opened 2026-08-23
Environment deletion changes agent config without writing a version


mdonigian:
Follow-up from #1049's review: `agents.environment_id` is `on_delete: :nilify_all`, so deleting an environment silently changes the live agent's config with no version row. The newest version then claims an `environment_id` that no longer exists, and the next edit's diff shows `environment_id: X → nil` attributed to the wrong change. Rollback to such a version is correctly refused by the FK re-validation, so nothing breaks — but the "newest version == live row" invariant the history page and the conversation stamp rely on has this one hole.

Same shape, no FK: `allowed_vault_ids` / `allowed_environment_ids` go stale on vault/environment deletion.

Options from the review: write a version from `Environments.delete_environment` (and the vault counterpart), or document the gap in ADR 0029.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1959, opened 2026-09-11
fix(agents): snapshot persisted fields after stale-struct edits

jhgaylor:
An edit made from a previously fetched agent struct can write a version that omits another edit already saved in the database. For example, saving new instructions and then changing the description through the original struct preserves both fields in the agent row but records the old instructions in its newest version. Request the complete persisted row with PostgreSQL `RETURNING` before inserting the version in the same transaction.

Small prerequisite for #1052: two files, +23/-1. Resource deletion and allowed-reference cleanup remain separate. This adds no query and preserves the transaction and post-commit audit boundaries.

Validation: the regression fails on the parent implementation and passes with this change. All 71 agent context, version API and version LiveView tests pass. Full `mix precommit` passed: 5,376 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/agents.ex
+++ b/apps/fountain/lib/fountain/agents.ex
@@ -164,7 +164,10 @@ defmodule Fountain.Agents do
     result =
       if snapshot_needed?(changeset) do
         Repo.transaction(fn ->
-          with {:ok, updated} <- Repo.update(changeset),
+          # The caller may hold an older struct. RETURNING snapshots the
+          # persisted row, including fields changed by another edit, while
+          # this update holds the row lock through the version insert.
+          with {:ok, updated} <- Repo.update(changeset, returning: true),
                {:ok, _} <-
                  Repo.insert(version_changeset(updated, next_version_number(updated.id))) do
             updated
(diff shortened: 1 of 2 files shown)


BinaryBourbon (commented):
Reviewed `156b5964ae6282993090c2435595aeaa26afe0db` for stack #1052.

No actionable defect introduced by this PR found. RETURNING supplies the persisted fields used by the version snapshot inside the existing transaction. This is explicitly a prerequisite for #1052: resource-deletion versioning and allowed-reference cleanup remain open.

Validation: static review of the exact-head update, snapshot fields, version numbering, and regression, plus the locally installed Ecto returning implementation. Database tests were not rerun.

BinaryBourbon (approved):
Approve `156b5964ae6282993090c2435595aeaa26afe0db`.

RETURNING supplies the persisted scalar configuration fields before the version snapshot is inserted in the same update transaction, closing the stale-struct history bug without an extra read. No actionable defect found. Approval covers this prerequisite only: #1052’s environment-deletion versioning and allowed-reference cleanup remain separate work.

Rechecked current CI and review discussions. [CI run 34616649357](https://github.com/managoat/fountain/actions/runs/34616649357) is green, including all six test partitions, static analysis and the required gate. Logs identify tested merge `2ae0bbe` of this head into `4fc851512f40e2326c4cf175af4d10d4933cfff1`. The log confirms AgentsTest ran in partition 5. The database regression was verified from CI rather than rerun locally.

PR #1960, opened 2026-09-11
fix(comms): surface unidentified provider-accepted sends

### LINKED ISSUE #1520: A comms send the provider accepts without a message id is unmetered and unaudited
Found while doing #1143. **Not introduced by it, and not closed by it** — same class of bug (a free message), different path.

## The gap

Every send handler in `Fountain.Team.Comms.Mcp` matches the provider response on the presence of a message id, and calls `audit/3` only in that branch:

```elixir
case phone(ctx).send_message(payload) do
  {:ok, %{"id" => mid} = resp} ->
    audit(ctx, "sms_send", %{"provider_message_id" => mid})   # audit + meter
    ok(id, %{message_id: mid, ...})

  {:ok, other} ->
    ok(id, other)                                             # nothing at all
  ...
```

`apps/fountain/lib/fountain/team/comms/mcp.ex` — `email_send` (~L237), `email_reply` (~L266), `sms_send` (the same shape).

`audit/3` is what the controller hangs metering off (`TeamCommsMcpController.audit_fn/2` → `meter/4`), so the `{:ok, other}` branch produces:

- no `comms_messages` row → **the customer is never charged**
- no `usage_events` row → invisible to the product mirror and the dashboard
- **no audit row either** → no record the teammate sent anything

The message still went out, and the provider still bills Fountain for it. The tool even returns success to the agent.

## Why #1143 did not fix it

#1143 moved billing onto a durable row that cannot be silently dropped, which fixes the case where the *write* fails. This is upstream of that: the write is never attempted, because the code decided there was nothing to record. The new path logs an explicit "returned no provider message id; it will not be billed" line, but only when it is reached — and this branch does not reach it.

## Severity, honestly

P2 rather than P1: `{:ok, other}` is not a documented AgentMail or AgentPhone response shape, so this may be unreachable today. It is a latent hole that opens the moment a provider changes its response, adds a field, or returns a partial success — and it opens silently, in the direction of giving product away. Prod currently has 2 contacts and 0 comms events in 7 days, so nothing is being lost right now.

## Suggested shape

Decide deliberately what an id-less success means, rather than leaving it as a fallthrough:

- If it is a real success, audit and meter it with whatever identifier exists (a thread id, a request id), or record it with a synthesised key so it is at least visible and reconcilable against the provider invoice.
- If it cannot be identified, arguably it should not report success to the agent — an unidentifiable send cannot be reconciled, retried, or charged.

Either is better than the current silence. At minimum it should log at `error` the way the id-less path in `record_message/6` already does.

## Refs

#1143, PR #1514 (which added the logging this branch bypasses), `Fountain.Team.Comms.Mcp`, `FountainWeb.TeamCommsMcpController`.

jhgaylor:
When a comms provider accepted a send without a message ID, the tool reported success and skipped its audit/metering callback. Route every accepted email send, email reply and SMS send through one completion path. A missing, blank or invalid ID now produces an explicit MCP tool error: the message may already have been sent, so check the provider records before retrying.

The chosen policy is to refuse an unidentifiable success, rather than fabricate a billable provider ID. The existing audit and usage paths still record the accepted request, labeled `outcome: unidentified`, and the existing error log makes the billing gap visible. These requests remain unbilled until they can be reconciled; ordinary identified successes keep their response and billing behavior. Provider response contents, recipients and message bodies stay out of the audit metadata.

Closes #1520. Four files, +143/-23.

Validation: all 42 focused tests pass. Three protocol regressions exercise eight unidentified response shapes each, with one send and one audit callback per call. The authenticated controller regression verifies the tool error, audit row, labeled usage event, explicit error log and absence of a fabricated comms billing row. All four regressions fail on the parent implementation. Full `mix precommit` passed: 5,379 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/team/comms/mcp.ex
+++ b/apps/fountain/lib/fountain/team/comms/mcp.ex
@@ -21,7 +21,7 @@ defmodule Fountain.Team.Comms.Mcp do
     * `:phone`   — the AgentPhone client module (default
                    `Fountain.Team.Comms.AgentPhone`)
     * `:audit`   — optional `(tool_name, summary_map) -> any` called after a
-                   successful send, for the audit trail
+                   provider-accepted send, including unidentified responses
   """
 
   alias Fountain.Team.Contact
@@ -231,12 +231,8 @@ defmodule Fountain.Team.Comms.Mcp do
         |> put_present("html", present(args["html"]))
 
       case mail(ctx).send_message(c.email_inbox_id, body) do
-        {:ok, %{"message_id" => mid} = resp} ->
-          audit(ctx, "email_send", %{"recipients" => length(to), "provider_message_id" => mid})
-          ok(id, %{message_id: mid, thread_id: resp["thread_id"]})
-
-        {:ok, other} ->
-          ok(id, other)
+        {:ok, resp} ->
+          finish_send(id, "email_send", ctx, resp, %{"recipients" => length(to)})
 
         {:error, reason} ->
           tool_error(id, "email_send failed: #{describe(reason)}")
@@ -256,16 +252,8 @@ defmodule Fountain.Team.Comms.Mcp do
         |> put_present("reply_all", if(args["reply_all"] == true, do: true))
 
       case mail(ctx).reply_to_message(c.email_inbox_id, message_id, body) do
-        {:ok, %{"message_id" => mid} = resp} ->
-          audit(ctx, "email_reply", %{
-            "reply_all" => args["reply_all"] == true,
-            "provider_message_id" => mid
-          })
-
-          ok(id, %{message_id: mid, thread_id: resp["thread_id"]})
-
-        {:ok, other} ->
-          ok(id, other)
+        {:ok, resp} ->
+          finish_send(id, "email_reply", ctx, resp, %{"reply_all" => args["reply_all"] == true})
 
         {:error, reason} ->
           tool_error(id, "email_reply failed: #{describe(reason)}")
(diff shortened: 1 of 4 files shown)


BinaryBourbon (commented):
Reviewed `a46471bcffb8fc0fc14c97c83469ca1c8f819510` for stack #1520.

No actionable defect introduced by this PR found. All three sends now report and audit unidentified acceptance without inventing a billable provider ID; reviewed the controller audit/usage path and privacy of the emitted metadata.

Validation: ran the exact-head MCP test file with exact-head Mcp and Contact modules and existing compiled dependencies: 20 tests, 0 failures, using fake providers only. Controller/database integration tests were not rerun.

BinaryBourbon (approved):
Approve `a46471bcffb8fc0fc14c97c83469ca1c8f819510`.

All three send paths audit unidentified acceptance and return an explicit ambiguity warning without exposing the provider payload or inventing a billable message identifier. The controller preserves the outcome on audit and usage records. No actionable defect found.

Rechecked current CI and review discussions. [CI run 34618111945](https://github.com/managoat/fountain/actions/runs/34618111945) is green, including all six test partitions, static analysis and the required gate. Logs identify tested merge `7df9bd14` of this head into `4fc851512f40e2326c4cf175af4d10d4933cfff1`. The log confirms TeamCommsControllerTest ran in partition 4. Also ran the exact-head MCP suite with fake providers and existing compiled dependencies: 20 tests, 0 failures. Controller integration was verified from CI rather than rerun locally.

Issue #1706, opened 2026-09-07
/api/events/stream can skip an event for good when two servers commit out of id order, and the cursor fix has no regression test


jhgaylor:
## Mechanism

After #1631 the live path of the account event stream no longer forwards the broadcast event. Each `{:log_event, id}` notification triggers a DB drain (`apps/fountain/lib/fountain_web/controllers/events_controller.ex:149-176`, `Conversations.list_user_log_events/2` at `conversations.ex:1633`) and advances the cursor to the highest id read.

Log-event ids come from a sequence assigned at insert, and two ConversationServers of one tenant commit independently (insert then broadcast, `conversations.ex:1494-1510`). If N+1 commits and broadcasts before N commits, the drain reads N+1, the cursor moves past N, and N is never delivered to that client. Frequency in production is inferred, not measured.

Also one DB round trip per event per connected client. An EXPLAIN on production shows the query itself is cheap (conversation index, ~3 ms over 267k rows), so this is a correctness note first.

## Test gap

The only filtered-row test (`events_stream_test.exs:180-226`) passed before the fix; nothing observes the cursor advancing past filtered rows.

## Fix

Deliver the notified event directly when it is the next id, and only drain on a gap; or drain with `id > cursor` but hold the cursor at the lowest committed id (a `pg_snapshot_xmin` style watermark). Add the regression test either way.

---
Found in the 2026-09-07 review of the week on `main` (cb93a7ea).

PR #1961, opened 2026-09-11
test(events): observe cursor advancement across filtered rows

jhgaylor:
The account-stream tests filtered out rows but did not observe the database cursor moving past them. Assert the next durable-feed read starts after the filtered row, and that a stale notification does not query or emit that row again. The request uses real database pages and the existing asynchronous SSE harness.

Regression-only unit of #1706: one test file, +35/-0. The independent transaction commit-order bug is addressed in a separate child PR.

Validation: all ten account-stream tests pass. Temporarily retaining the old cursor on filtered rows makes the new test fail at its next database read; restoring production code returns it to green. No production change is committed. Full `mix precommit` passed: 5,376 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/test/fountain_web/controllers/events_stream_test.exs
+++ b/apps/fountain/test/fountain_web/controllers/events_stream_test.exs
@@ -177,6 +177,41 @@ defmodule FountainWeb.EventsStreamTest do
       assert conn.resp_body =~ "from-new"
     end
 
+    @tag :filtered_cursor_regression
+    test "filtered rows advance the durable cursor and stale notifications do not rescan them", %{
+      user: user,
+      raw_key: key
+    } do
+      conv = insert_conversation(user_id: user.id, status: "idle")
+      old = insert_log_event(conv, %{kind: "output", stream: "stdout", data: "old"})
+      await_stream_start(user.id)
+      task = stream_async(key, "/api/events/stream?streams=stage")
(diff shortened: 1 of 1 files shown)


BinaryBourbon (commented):
Reviewed `d404c0e6286a888e0a878638e3a39e1b23b69ac1`.

Reviewed with stacked #1963. No actionable findings. The added regression observes the actual cursor passed to the durable-event query after a filtered row, rather than checking output alone.

Validation: static review of the test, stream implementation and test synchronization helpers; integration tests were not rerun.

BinaryBourbon (approved):
**Approved** at `d404c0e6286a888e0a878638e3a39e1b23b69ac1`.

The regression meaningfully verifies cursor advancement: it checks the durable query receives `old.id` first and `hidden.id` after the filtered row, then sends a stale notification. I rechecked this against `EventsController.replay/2`, the stale-ID receive guard and the test synchronization helper. No blocking findings.

Validation: reviewed unchanged current head/base and stacked #1963; verified [CI run 34618636703](https://github.com/managoat/fountain/actions/runs/34618636703) belongs to this exact head. All six core partitions pass (5,383 tests in downloaded logs); static, API contract and required gates pass. I inspected those logs and did not rerun the database suite locally.

Issue #1697, opened 2026-09-07
Deployed suite: recover fixtures after runner or cleanup-journal loss


jhgaylor:
Recover run-owned fixtures when cancellation, a hard job timeout, or loss of the GitHub Actions runner prevents cleanup and evidence upload, including when no usable cleanup manifest survives.

Scope split on 2026-09-17: scheduled canary/matrix enablement, environment approval policy, monitoring, spending readiness and the staging decision now belong to #2356. That work remains deferred. This issue owns the recovery path used by both manual verification and any future recurring runs.

## PR boundary and sequencing

Target one focused recovery PR against current `main`: an exercised operator-assisted procedure, with narrowly scoped tooling only where needed. A documented limitation alone does not close the issue.

No outstanding PR is a hard prerequisite. The existing preflight, diagnostics and cleanup replay fixes (#1962, #1964, #1965, #1966) are merged. Do not wait for #2356, a staging deployment, or the whole machine-owner tracker #2344 to start. #2386 concerns server-side machine destruction; coordinate with its final behavior only if an implementation touches that lifecycle. Reuse the existing public cleanup interfaces rather than adding another server lifecycle coordinator.

The GitHub Actions runner disappearing is distinct from the application rollout/runner-reconnect continuity tests in #2356 Part 2. This recovery task does not require restarting serving pods.

## Done when

- [ ] Document recovery from a separate process or machine after the original runner is gone. Cover both replay of surviving evidence and reconciliation when neither a usable manifest nor an uploaded artifact survives; another final step on the original runner is insufficient.
- [ ] Establish the target, suite account and exact fixture ownership before mutation. Explain how the operator reconstructs the necessary evidence without the original journal, including parent relationships and sandbox identity/mode where relevant. Ambiguous or foreign ownership must refuse deletion; a broad `suite-` prefix is not sufficient authorization.
- [ ] Cover interrupted creation, including a request that committed before its response or journal update was saved. Preserve unresolved intents until reconciled; one empty listing is not proof that an in-flight request cannot still commit.
- [ ] Stop run-owned execution and outbound sources where applicable, verify sandbox cleanup, and preserve parent ownership evidence until dependent cleanup succeeds. Repeating recovery must tolerate already-cleaned resources and visibly report anything still unresolved. State the supported fixture/profile coverage and escalation procedure for cases that cannot be safely reconciled.
[...]

PR #1962, opened 2026-09-11
ci(deployed): require a protected environment before verification

jhgaylor:
A deployed verification job can cause GitHub to create a missing target environment without reviewers. Add a read-only preflight in the existing validation job, before the verification job references its environment. Both manual and enabled scheduled runs must find the expected environment with at least one configured user or team reviewer; missing environments, API failures and incomplete protection metadata stop the run.

Small unit of #1697. This does not create environments, enable schedules, configure credentials or run the suite against a deployment. Cleanup replay, account funding and setup diagnostics remain separate. Current read-only inspection finds only `github-pages` and `pypi`, so deployed targets must still be configured by an operator.

The validation job receives only the additional `actions: read` permission required by GitHub's [Get an environment endpoint](https://docs.github.com/en/rest/deployments/environments#get-an-environment). The verification job's permissions stay unchanged.

Validation: All 204 deployed-suite tests pass, including 16 new preflight cases for staging/production, user/team reviewers, missing or malformed protection, API failures and invalid inputs. `actionlint` passes the changed workflow. Three files, +106/-0. Full `mix precommit` passed: 5,375 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/.github/workflows/deployed.yml
+++ b/.github/workflows/deployed.yml
@@ -41,6 +41,9 @@ jobs:
     if: github.event_name != 'schedule' || (github.event.schedule == '23 */6 * * *' && vars.DEPLOYED_CANARY_ENABLED == 'true') || (github.event.schedule == '43 4 * * 6' && vars.DEPLOYED_MATRIX_ENABLED == 'true')
     runs-on: ubuntu-latest
     timeout-minutes: 2
+    permissions:
+      contents: read
+      actions: read
     steps:
       - name: Require approved target and main checkout
         env:
@@ -51,6 +54,14 @@ jobs:
             staging|production) ;;
             *) echo 'Target is not approved'; exit 1 ;;
           esac
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v5
+        with:
+          persist-credentials: false
+      - name: Require an existing environment with reviewers
+        env:
+          REQUESTED_TARGET: ${{ inputs.target || 'production' }}
+          GH_TOKEN: ${{ github.token }}
+        run: node deployed/validate-environment.mjs
   verify:
     needs: validate
     runs-on: ubuntu-latest
--- a/deployed/test/environment-preflight.test.mjs
+++ b/deployed/test/environment-preflight.test.mjs
@@ -1,0 +1,57 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { requireProtectedEnvironment } from '../validate-environment.mjs';
+
+const env = { GITHUB_REF: 'refs/heads/main', REQUESTED_TARGET: 'staging', GITHUB_REPOSITORY: 'owner/repo', GH_TOKEN: 'fixture-token' };
+const approved = type => ({ name: 'deployed-staging', protection_rules: [
+  { type: 'required_reviewers', reviewers: [{ type, reviewer: { id: 123 } }] },
+] });
+const reply = data => async () => ({ ok: true, json: async () => data });
+
+for (const [type, target] of [['User', 'staging'], ['Team', 'staging'], ['User', 'production'], ['Team', 'production']]) {
+  test(`a configured ${type} reviewer permits a read-only ${target} preflight`, async () => {
+    const calls = [];
+    const request = async (url, options) => {
+      calls.push({ url, options });
+      return reply({ ...approved(type), name: `deployed-${target}` })();
+    };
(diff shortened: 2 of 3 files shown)


BinaryBourbon (approved):
**APPROVE** — adds a read-only preflight to `deployed.yml`'s `validate` job that refuses to proceed unless `deployed-<target>` already exists with at least one configured user or team reviewer, closing the window in which GitHub auto-creates an unprotected environment the moment `verify` first names it.

The fail-closed property holds, which was the thing worth checking:

- `verify` has `needs: validate` and no `if:`, so a failed preflight skips it. `validate`'s own `if:` can skip the job on a disabled schedule, and a skipped `needs` skips `verify` too — there is no path where `verify` runs without the preflight having succeeded.
- `deployed/validate-environment.mjs:29` sets `process.exitCode = 2` on every rejection, and the rejection branches are exhaustive: non-`main` ref, unapproved target, missing repo/token, transport failure, non-2xx, unparseable JSON, and missing/incomplete `required_reviewers`. Nothing degrades to a warning.
- Verified by running the CLI directly: empty environment, missing token, and an unreachable `GITHUB_API_URL` each print a one-line message and exit `2`.
- No secret exposure: the `catch` blocks at `validate-environment.mjs:21` and `:26` discard the caught error, so neither a response body nor `GH_TOKEN` reaches stderr; the success line prints only the environment name.

### Non-blocking

**`deployed/validate-environment.mjs:35-38` — the exit-code path, which is the whole gate, is not covered by the new tests.** `environment-preflight.test.mjs` imports and exercises `requireProtectedEnvironment` only; the CLI wrapper is never invoked. I confirmed the tests do guard the function — deleting the `if (!protectedTarget) throw` fails 6 of 16, and turning `if (!response.ok) throw` into a `console.warn` + `return name` fails 2 of 16 — but deleting `process.exitCode = 2` keeps **all 16 green**. The same blind spot covers the `resolve(process.argv[1]) === fileURLToPath(import.meta.url)` guard: if that comparison ever stops matching (invocation through a symlinked workspace path, or a future `--import`/wrapper), the step becomes a silent no-op that exits 0 and `verify` proceeds — exactly the "green while doing nothing" shape this repo has been bitten by. A single child-process test (`node deployed/validate-environment.mjs` with a fixture env, asserting a non-zero status and that stdout is empty) would pin both.

**`.github/workflows/deployed.yml:64` — `validate` runs `node` from the runner's preinstalled toolchain**, while every other `node --test` / `node deployed/*.mjs` site in the repo (`ci.yml:1000`, `deployed.yml:84`) pins `actions/setup-node@…v7` with `node-version: 24` first. `fetch` and `AbortSignal.timeout` are old enough that this is safe today; it is a consistency point, not a defect.

**`.github/workflows/deployed.yml:47` — `actions: read` is probably not required.** GitHub documents *Get an environment* as "anyone with read access to the repository can use this endpoint"; the workflow-level `contents: read` should already satisfy it. Harmless either way, and if it turns out to be needed the failure mode is a 403 that fails closed. Worth dropping only if you confirm it on the first real run.

### Checked

- Full `git diff BASE..HEAD`; the workflow parses (`yaml.safe_load`) and `actionlint` is clean on `deployed.yml`.
- `needs`/`if`/`environment` wiring for both jobs; the preflight's `deployed-${REQUESTED_TARGET}` matches `verify`'s `environment: deployed-${{ inputs.target || 'production' }}` for both values of the choice input.
- Ran the 16 new tests (`node --test deployed/test/*.test.mjs`): all pass. `ci.yml:1009` globs `deployed/test/*.test.mjs`, so the new file is picked up with no CI change.
- Mutation-tested the guard (results above).
- `REQUESTED_TARGET` reaches the script through `env:`, never interpolated into a shell command; the checkout is pinned to the same SHA as `verify`'s and keeps `persist-credentials: false`.
- Confirmed against the live repo that `deployed-staging` / `deployed-production` do not exist yet (only `github-pages` and `pypi`), and that `DEPLOYED_CANARY_ENABLED` / `DEPLOYED_MATRIX_ENABLED` are unset — so merging this turns no currently-running schedule red. An operator still has to create both environments before any dispatch can verify.
- CI-only change; no `[Unreleased]` CHANGELOG entry needed, and no new `runtime.exs` env var, so no `docs/configuration.md` row.

Not verified: I read YAML rather than running the workflow, so I did not confirm that `GITHUB_TOKEN` actually receives `protection_rules` (including the `reviewers` array) on this repo, nor the real response shape from the environments endpoint. The first dispatch after an environment is created is the test for that.

PR #1963, opened 2026-09-11
fix(events): serialize account log ids through commit

### LINKED ISSUE #1706: /api/events/stream can skip an event for good when two servers commit out of id order, and the cursor fix has no regression test

jhgaylor:
Account event streams advance an integer cursor to the newest committed row they read. Previously one conversation could allocate ID N and remain uncommitted while another conversation committed N+1, letting the cursor permanently skip N. Serialize the single production log writer by account before allocating an ID, and hold that advisory lock through commit, including any enclosing transaction.

Different accounts keep independent locks. The SSE wire form and cursor remain unchanged. This adds one account lookup/lock query and a transaction around each standalone log insert; the lock covers persistence, not sandbox work. The guarantee requires all log writers to run this version, so a mixed-version rollout can still contain old unsynchronized writers.

Commit-order unit of #1706, stacked on #1961, which supplies the separate filtered-cursor regression. Two files, +150/-1. Closes #1706 with that parent.

Validation: two independent-PostgreSQL-transaction regressions cover commit and rollback, assert the blocked backend, let a different account commit while the first is held, and verify the durable feed contains exactly the expected IDs. Both fail on the parent implementation; all 12 race and account-stream tests pass with this change. The production call sites use the single log writer. Full `mix precommit` passed on a clean isolated database: 5,378 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures. The focused race, stream and quota suite also passes all 39 tests and leaves zero fixture users or sandboxes.


BinaryBourbon (commented):
Reviewed `93a93964690499a0758abec4362f4a00055d3efe`.

Reviewed with parent #1961. No actionable findings. The account advisory lock is acquired before ID allocation and survives an enclosing transaction, covering the commit-order failure; inspected the single writer and the independent-transaction commit/rollback regressions. The mixed-version rollout limitation is correctly disclosed.

Validation: static review of production callers, transaction boundary, durable stream queries and tests; PostgreSQL integration tests were not rerun.

BinaryBourbon (approved):
**Approved** at `93a93964690499a0758abec4362f4a00055d3efe`.

The account advisory lock is acquired before sequence allocation and remains held through an enclosing transaction, preventing a later account event from committing ahead of an earlier allocated ID. Rechecked the single production writer, durable stream reads and independent PostgreSQL commit/rollback regressions, including the other-account progress check. No blocking findings. The documented mixed-version rollout caveat is accurate.

Validation: [CI run 34620311896](https://github.com/managoat/fountain/actions/runs/34620311896) is verified at this exact head; six core partitions pass (5,385 tests in downloaded logs), with static, API-contract and required gates green. Reviewed with parent #1961. No new local database run or production/provider calls.

BinaryBourbon (approved):
Approve `403165de684f12e1f59f1a000ce5b1f8b243ab9a` after merging main following parent #1961.

Independently verified the new head against main: the remaining diff is exactly the original two-file child change (+150/-1), with identical stable patch ID `1b289b5f5ad5f2efb8a7c67052bad5c22e9152e2`. Rechecked the account lock before log ID allocation, durable account cursor reads, and the newly integrated DeadlineEvents/ExecutionGuard path. The deadline event uses this same writer within the outer journal transaction, so the transaction-scoped account lock remains held through that commit. No blocking integration findings.

Additional validation: ran a disposable PostgreSQL 17 concurrency probe using the same advisory-lock query. In both commit and rollback cases a later event for the same account blocked, an unrelated account committed independently, and visible account IDs remained correctly ordered. The temporary server was stopped. This was a SQL-level check, not a full local Ecto suite. Original-head CI was previously green; fresh integration CI [34671589501](https://github.com/managoat/fountain/actions/runs/34671589501) has started and must pass before queue admission. The mixed-version rollout limitation remains: all writers need this lock protocol before ordering is guaranteed.

Issue #1439, opened 2026-09-03
Render: run the verification ladder and land the prod key


jhgaylor:
Depends on #1437. The verification ladder from `docs/integrations/adding-a-sandbox-provider.md`, run in order. Each rung has caught a real defect that the rung before it passed, and both previous adapters passed every local run before failing live.

1. **`mix test`**, conformance suite green against Req.Test stubs. Done in #1436.
2. **Live adapter smoke** against a real Render account. Exercise every behaviour callback: create, adopt an existing name, exec with a nonzero exit, a streaming spawn with multi-write stdin, attach with replay from byte zero, stdin EOF with the real exit code, destroy. Suspend and resume are expected to be absent rather than tested, and the absence itself needs confirming rather than assuming. Verify what `network_policy` actually enforces, and record it: a policy accepted with a `200` and never enforced is a false security claim we would pass on to an operator.
3. **A full conversation cycle** locally against Render. `start_conversation`, a real model turn, idle (which destroys on this provider rather than parks), a status check on the Render side confirming the sandbox is gone, a fresh conversation, a second turn, terminate.
4. **A prod smoke** after merge and deploy. Pin an agent to Render with `sandbox_provider` on `POST /api/agents`, run one conversation through the public API, poll the **turn status** rather than `turn_count` (a failed turn still counts), read the answer from `/events`, terminate, then confirm the Render account is clean.

Note that this repo has been burned by a green PR gate that shipped nothing: check `build.yml` on main before treating rung 4 as testable.

## After the ladder

- `RENDER_API_KEY` and the workspace id into the Infisical prod project, which syncs into `fountain-secrets` and reloads the deployment on its own.
- Record the live findings back into #1435 and into the `platform-requirements.md` Render column. Every correction the E2B and Daytona live runs produced was kept in ADR 0018 precisely because each one looks wrong without its context. Do the same here.

jhgaylor:
Folded into the tracker #1441 on 2026-09-16, under **Remaining work → 4. Verification ladder and the prod key**. The details were carried over and corrected against the 2026-09-13 research checkpoint (ADR 0040 identity, no 0.2.0 target). The work is still planned; it is tracked there now. Closing as a duplicate of #1441.

PR #1964, opened 2026-09-11
ci(deployed): report safe setup diagnostics

jhgaylor:
Every CI setup error previously printed the same generic message, so an operator could not tell whether enablement, target JSON, profile selection or rollout configuration was wrong. Report controlled configuration messages with the failing setting named. Wrap JSON and URL parser errors so their input excerpts never reach logs; unexpected failures retain the generic message.

Both target and matrix settings must be JSON objects. A matrix setting of `null` can no longer fall back to the probe profile. Setup validation still occurs before creating run files or making requests.

Separate diagnostics unit of #1697, stacked on #1962. Two files, +83/-12. Environment credentials, cleanup replay and operator enablement remain separate.

Validation: all 217 deployed-suite tests pass. Thirteen CLI subprocess cases verify actionable messages, unchanged exit code 2, no run files on configuration failures, hidden secret markers and generic handling of unexpected filesystem errors. Twelve cases fail on the parent implementation; its generic-error preservation case already passes. Full `mix precommit` passed: 5,375 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

## This closes a fail-open hole, not just a message

Flagged in review and worth stating plainly, because the title reads as
diagnostics-only. On the parent commit (`3ea371ce`), `SUITE_MATRIX_JSON='null'`
with `SUITE_PROFILE=matrix-canary` parsed to `null`, so `validateMatrix` was
skipped, no `matrix.json` was written, and the job silently ran the plain
single-target suite with `profiles: ["probe"]` — exiting 0 against a healthy
target. A matrix deployment verification that verified one probe and reported
success. Reproduced at that commit during review.

`settingObject()` now rejects it.

### DIFF
--- a/deployed/ci.mjs
+++ b/deployed/ci.mjs
@@ -7,25 +7,38 @@ import { runMatrix, validateMatrix } from './matrix.mjs';
 
 const matrixSubset = profile => ({ 'matrix-canary': 'canary', 'matrix-scheduled': 'scheduled', 'matrix-full': 'full' })[profile];
 
+// Only controlled configuration messages may reach CI stderr. Parser errors
+// can quote input values; unexpected errors can contain credentials or paths.
+class SetupError extends Error {}
+function settingObject(env, name) {
+  let value;
+  try { value = JSON.parse(env[name] || '{}'); }
+  catch { throw new SetupError(`${name} must contain valid JSON`); }
+  if (!value || typeof value !== 'object' || Array.isArray(value)) throw new SetupError(`${name} must be a JSON object`);
+  return value;
+}
+
 export function ciConfig(env) {
-  if (!['staging', 'production'].includes(env.SUITE_TARGET)) throw new Error('Target is not approved');
-  if (env.SUITE_ENABLED !== 'true') throw new Error('Target environment is not enabled');
-  if (!['probe', 'basic', 'execution', 'streaming', 'canary', 'secrets', 'mcp', 'webhooks', 'schedules'].includes(env.SUITE_PROFILE) && !matrixSubset(env.SUITE_PROFILE)) throw new Error('Profile is not approved');
-  const config = JSON.parse(env.SUITE_TARGET_JSON || '{}');
-  const url = new URL(config.base_url);
-  if (url.protocol !== 'https:') throw new Error('CI targets require HTTPS ingress');
+  if (!['staging', 'production'].includes(env.SUITE_TARGET)) throw new SetupError('Target is not approved');
+  if (env.SUITE_ENABLED !== 'true') throw new SetupError('Target environment is not enabled');
+  if (!['probe', 'basic', 'execution', 'streaming', 'canary', 'secrets', 'mcp', 'webhooks', 'schedules'].includes(env.SUITE_PROFILE) && !matrixSubset(env.SUITE_PROFILE)) throw new SetupError('Profile is not approved');
+  const config = settingObject(env, 'SUITE_TARGET_JSON');
+  let url;
+  try { url = new URL(config.base_url); }
+  catch { throw new SetupError('SUITE_TARGET_JSON must contain a valid base_url'); }
+  if (url.protocol !== 'https:') throw new SetupError('CI targets require HTTPS ingress');
   config.credentials = { primary: 'FOUNTAIN_SUITE_KEY', secondary: 'FOUNTAIN_SUITE_OTHER_KEY' };
   config.profiles = matrixSubset(env.SUITE_PROFILE) ? ['probe'] : env.SUITE_PROFILE === 'canary' ? ['basic', 'execution'] : [env.SUITE_PROFILE];
   // Bound exposure even if an environment variable accidentally requests a longer run.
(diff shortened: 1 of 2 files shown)


BinaryBourbon (approved):
**APPROVE**

Replaces the single catch-all `CI setup failed; check approved target configuration...` line with a `SetupError` class whose messages are constant literals, so a misconfigured `deployed.yml` run now says *which* setting is wrong, and adds `deployed/test/ci-diagnostics.test.mjs` to hold that line.

## Leakage review (the thing this PR is most at risk of)

No leak. Every `SetupError` message in `deployed/ci.mjs` is a fixed literal; the only interpolation is `${name}` in `settingObject` (ci.mjs:16-17), and both call sites pass a hardcoded variable name (`'SUITE_TARGET_JSON'` at ci.mjs:25, `'SUITE_MATRIX_JSON'` at ci.mjs:53). `SetupError` is module-local and unexported, so nothing thrown out of `run()` / `runMatrix()` can reach the message-printing branch at ci.mjs:76; those still get the generic string. Notably the PR resists three tempting leaks: the `JSON.parse` `SyntaxError` (which quotes input), the `new URL()` `TypeError` (which quotes `base_url`), and `validateMatrix`'s messages.

I verified the new tests actually guard rather than merely pass, by mutating `ci.mjs` in a scratch copy:
- `settingObject` message changed to include `${e.message}` and `${env[name]}` -> "target JSON" and "matrix JSON" cases fail.
- generic branch changed to `${error.stack}` -> "unexpected setup errors" case fails (it pre-creates `deployed-<marker>-1` under `RUNNER_TEMP` so `mkdirSync` at ci.mjs:59 raises a non-`SetupError` `EEXIST` carrying the path — a good way to reach that branch).

Also checked that the diagnostics cannot mask a failure: `process.exitCode = 2` on every path as before, no `continue-on-error` / `|| true`, and `.github/workflows/deployed.yml:109-111` runs `node deployed/ci.mjs` as a plain failing step. The new `try/catch` around `validateMatrix` (ci.mjs:55-56) rethrows rather than swallowing.

## Worth calling out: this is more than diagnostics

`settingObject` rejecting `null`, arrays and scalars closes a real fail-open hole that the title does not advertise. On base, `SUITE_MATRIX_JSON=null` with `SUITE_PROFILE=matrix-canary` parsed to `null`, so `if (matrix)` was false: `validateMatrix` was skipped, no `matrix.json` was written, and the job silently ran the plain single-target suite. Reproduced at `3ea371ce` — it wrote `target.json` with `"profiles":["probe"]` and no matrix file, i.e. a matrix run degraded to one probe and would have exited 0 against a healthy target. At `47fe584f` the same input is a `SetupError`. Please say so in the PR body; it is the kind of thing this repo has been bitten by and it deserves to be findable later.

## Nits (non-blocking)

1. `deployed/ci.mjs:55-56` collapses ~18 distinct `validateMatrix` failures into `'SUITE_MATRIX_JSON must describe a valid bounded matrix'`. Every `ensure()` message in `deployed/matrix.mjs` is itself a constant literal with no input interpolation (`'Duplicate matrix cell ID'`, `'Matrix leaves an undeclared combination'`, `'Render remains a declared gap until #1439...'`), so forwarding them would be safe and much more actionable for the operator who has to fix the environment variable. The one caveat, if you do it: `matrix.mjs:47` reaches `entries.map(e => e.name)` before element shape is checked, so `catalog_gaps.runtimes = [null]` throws a raw `TypeError` rather than an `ensure` failure — forward only `ensure`-raised errors (e.g. have `ensure` throw a shared error class), not everything.
2. The table in the new test covers every `SetupError` branch except `'Expected digest requires rollout mode'` (ci.mjs:39). One more row.

## Checked

- `git diff 3ea371ce 47fe584f` in full; 2 files, no workflow or config change.
- Ran `node --test deployed/test/*.test.mjs` on scratch copies of both `3ea371ce` and `47fe584f`: 204 -> 217 tests, and the 23 failures are byte-identical between the two (all network/Docker-dependent profiles that cannot run in my sandbox). All 13 new tests pass.
- Mutation-tested the three leak assertions as described above.
- The new file is picked up by CI: `.github/workflows/ci.yml:1009` globs `deployed/test/*.test.mjs` in the `sdk-clients` job, so no gate wiring is needed and no aggregate `if:`/`needs:` changes.
- Existing callers of `ciConfig` that assert on messages still match (`matrix.test.mjs:111` `/approved/`, `:112` `/adapter/`) — `SetupError extends Error` preserves both.
- `SUITE_TARGET_JSON` / `SUITE_MATRIX_JSON` come from `vars.*` (deployed.yml:76-77), not `secrets.*`, so GitHub would not mask them if they were echoed — the fact that they are not echoed is load-bearing, not belt-and-braces.
- CI-only change, so no CHANGELOG `[Unreleased]` entry and no `docs/configuration.md` row (no new `runtime.exs` env var).

Not verified: I read YAML and ran Node locally; I did not run the `deployed` workflow against staging, and I did not exercise the profiles whose tests need network or Docker.

PR #1965, opened 2026-09-11
fix(deployed): replay cleanup across run manifests

jhgaylor:
Adds a bounded batch cleanup command for normal runs and the sixteen-cell matrix plus preflight. Each manifest is checked by the existing runner for target, account and resource ownership before cleanup. A rejected manifest does not prevent attempts for the remaining journals, and no verification profile or inference starts.

Replay stops admitting manifests after two minutes. SIGINT, SIGTERM and deadline expiry immediately persist the stop reason and mark active work interrupted, then let its independently bounded cleanup finish. Later entries stay unattempted and the aggregate remains nonzero. Exceptions receive a fixed classification without exposing private errors.

The CLI remains strict about missing or empty evidence by default. Its explicit `--allow-empty` option lets CI record `no_manifests` without requiring a target file when setup failed before any journal was written. Unsafe evidence and actual failed manifests still fail. Original journals are updated in place, so retain their artifact before replay.

Based on main after #1964; #1966 wires this command into the workflow and adds operator documentation. Refs #1697. This does not enable the suite or recover evidence lost with a runner.

Validation:
- [Fresh CI](https://github.com/managoat/fountain/actions/runs/34688485256) passed on the merge checkout with main.
- All 232 deployed Node tests passed, including subprocess SIGINT/SIGTERM during real HTTP cleanup, deadline expiry, continued cleanup after owner rejection, idempotence, empty CI runs and invalid evidence rejection.
- `mix precommit` passed on the final stacked tree with the pinned Elixir/Erlang toolchain and an isolated test database: 5,598 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests; zero failures (core: 2 skipped, 7 excluded). The application code is identical in both PRs.

No deployed workflow was dispatched or live fixture modified.

### DIFF
--- a/deployed/cleanup-replay.mjs
+++ b/deployed/cleanup-replay.mjs
@@ -1,12 +1,14 @@
 #!/usr/bin/env node
 import { lstatSync, readdirSync, mkdirSync } from 'node:fs';
-import { resolve, join, relative, sep } from 'node:path';
+import { resolve, join, relative, sep, dirname } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { parseArgs } from 'node:util';
 import { run } from './lib/runner.mjs';
 import { atomicJson } from './lib/fixtures.mjs';
 
-function manifests(root) {
+export const REPLAY_START_MS = 120000;
+
+function manifests(root, allowEmpty) {
   const found = [];
   let visited = 0;
   function visit(path, depth) {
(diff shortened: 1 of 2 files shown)


BinaryBourbon (approved):
APPROVE — the selector is sound, the traversal is validated before the first API call, and the delete path is unchanged from the single-manifest cleanup it reuses.

**What this changes**

`deployed/cleanup-replay.mjs` walks a results tree for every file named exactly `cleanup.json`, validates the whole tree first (no symlinks, depth ≤ 4, ≤ 512 nodes visited, 1–17 manifests, output directory outside the source), and then calls `run({ configPath, manifestPath, ... })` once per manifest. Supplying `manifestPath` forces `runner.mjs` into cleanup mode, so no profile executes and no fixture or inference request is made. One manifest's rejection or failure no longer stops the rest: each becomes an entry in `replay.json` and the process exits 3 while any entry is `failed` or `not_run`.

**The selector, since that is the thing worth proving**

Replaying across run manifests does not widen what cleanup may delete. It widens only *which journals* are read; every journal is then re-validated against the live target:

- `Fixtures.load` (`deployed/lib/fixtures.mjs:26`) rejects a manifest whose `base_url` differs from the configured target or whose `owner_id` differs from the `id` returned by `GET /api/auth/me` on the suite key. A foreign-owner or foreign-target journal produces zero requests against its resources — `cleanup-replay.test.mjs:52` asserts exactly that, and the `run()` return is 2 (setup_failed), recorded as `failed`.
- Every resource must carry a name matching `suite-<manifest run_id>-<kind>-<n>` (or `SUITE_<hex>_BINDING_<n>`), with `run_id` a UUID. A coincidental match would require another object in the *same dedicated test account* named after the same random UUID.
- Deletion resolves by recorded UUID where one exists, and by exact unique name otherwise; `matches.length > 1` raises `Ambiguous cleanup intent` and deletes nothing, and every DELETE is preceded by a marker/host/url re-read (`fixtures.mjs:205`). The endpoints are the tenant-scoped `/api/...` collections, so the suite key cannot reach another tenant's row even if a name did collide.

Mid-creation state is handled without a blind delete: a `pending` resource with no id is resolved by exact name, and if it is not there the manifest fails with `Unresolved create intent; retry cleanup after the server settles` rather than being marked clean. Idempotency comes from `Fixtures` persisting `state: 'cleaned'` back into the source manifest, so a second replay skips those resources — `cleanup-replay.test.mjs:56` shows the second pass issuing no further DELETEs, and a manifest that failed stays `failed` until it actually succeeds.

**Findings**

1. `deployed/cleanup-replay.mjs:59` — no `SIGINT`/`SIGTERM` handling, unlike every sibling entry point (`cli.mjs:8`, `ci.mjs:52`, `matrix.mjs:189`), which install handlers precisely so an interrupt becomes an abort and the in-flight `run()` still reaches its `finally` cleanup. Concrete: the next PR caps this step at four minutes while the replay's own budget stops *starting* manifests at three, so a manifest can still be inside `fixtures.cleanup` at 3:00. On termination node exits immediately, and the uploaded `replay.json` is left at `status: "running"` with that entry `"running"` — an operator reading the artifact cannot tell whether it was attempted. Non-blocking (the exit code is still nonzero and the job is already red), but the interruption path is this feature's reason to exist. See the budget finding on #1966, which is the same arithmetic from the other side.
2. `deployed/cleanup-replay.mjs:21` — an empty tree throws, so the script exits 2 with a message pointing at the target configuration. That is right for an operator running it by hand, and wrong in CI, where a setup-failed suite legitimately produces no `cleanup.json` at all. Filed against #1966, which is where the workflow can guard it; noting here because the alternative fix (exit 0 on an empty tree) lives in this file.
3. Non-blocking nit, `deployed/cleanup-replay.mjs:29` — `lstatSync(configPath).isSymbolicLink()` is unreachable. `lstat` on a symlink already fails `isFile()`, so the `config symlink` scenario in the test passes on the first clause and never exercises the second. Harmless, but the test name implies coverage the code does not need.
4. Non-blocking nit, `deployed/cleanup-replay.mjs:45` — `catch { entry.status = 'failed'; }` records no classification at all, not even that `run()` threw rather than returning nonzero. In the normal case the detail is in `manifest-N/result.json`, redacted; in the one case where `run()` throws before writing it (`mkdirSync` on a pre-existing `manifest-N`) there is no evidence anywhere. A fixed string such as `entry.error = 'threw'` leaks nothing and costs nothing.
5. Non-blocking nit — the new entry point is not in `cli.mjs`'s `help` string and has no `--help` of its own, so `node deployed/cli.mjs --help` (which does document exit codes 0/1/2/3/130) does not lead a reader to the batch form. The README entry arrives in #1966, which is fine for a stack.

**Checked and clean**

- Cleanup mode cannot create or spend: `runner.mjs` gates the whole profile loop and the deployment observer on `!manifestPath`, so replay issues only identity, read and delete requests. The test asserts every request is a GET or a DELETE against the recorded collection.
- Abort cannot be reported as success: `if (signal.aborted) break` leaves later entries `not_run`, and `report.status` is `passed` only when *every* entry is `passed`. The aborted test asserts `not_run` survives into `replay.json` and the exit code is 3.
- Secrets: the replay passes `log() {}` so the runner's PASS/FAIL lines never reach the Actions log; the only stderr is a constant string; `replay.json` holds relative manifest paths, statuses and exit codes; per-manifest evidence goes through the runner's `Redactor`. The test asserts the key and a thrown error's message are absent from `replay.json`.
- Output containment: `out === resultsRoot || out.startsWith(resultsRoot + sep)` is checked after `resolve`, so replay evidence cannot be re-discovered as an input on a later pass; `mkdirSync(out)` runs only after the whole tree validates, and the rejection tests assert `existsSync(out) === false`.
- `endsWith(sep + 'cleanup.json')` does not match the sidecars in the same directory (`webhook-receiver.json`, `mcp-receiver.json`, `receiver.json`) — those are still picked up by the runner from `dirname(manifestPath)`, so a webhooks or MCP manifest still tears its receiver down.
- The 17-manifest bound is exactly the `matrix-full` shape: `matrix.mjs` writes `results/preflight/` plus up to 16 selected cells, one `cleanup.json` each, at depth 2.
- Manifests are updated in place in the source results tree rather than copied, which is what makes replay idempotent; the PR body says so and #1966's artifact naming preserves the pre-retry copy.
- No CHANGELOG entry, correctly — this is operator tooling, not user-visible behavior.

PR #1966, opened 2026-09-11
fix(deployed): retry fixture cleanup after failed verification

jhgaylor:
A failed or interrupted deployed verification step now gets a fresh cleanup process after the workflow uploads its original manifests. Updated journals and retry reports use a separate artifact name; a successful retry cannot turn failed verification green.

Verification retains a 52-minute step cap. Replay gets six minutes: two minutes admitting manifests, up to 90 seconds of fixture cleanup and 90 seconds of receiver cleanup, plus a minute for process/evidence overhead. Setup, monitors, both uploads and credential removal have explicit caps. The 75-minute job accommodates their combined 71-minute budget with four minutes of overhead; verification and inference budgets do not increase.

CI uses the replay command's explicit `--allow-empty` mode. Setup failures with no journals produce `no_manifests` evidence and no API calls, while unsafe evidence and failed manifests remain failures. The README explains interruption evidence, budgets and manual recovery.

#1965 is merged. This PR is based on main and contains only the workflow, operator instructions and workflow regression tests. Refs #1697.

Recovery remains best effort on the existing runner. Cancellation, hard timeout or runner loss can prevent finalizers and uploads. Protected environments, credentials, a comped canary account and recovery of lost journals remain enablement work under #1697.

Validation:
- [Fresh PR CI against main](https://github.com/managoat/fountain/actions/runs/34721489565) passed after rebasing onto the merged #1965. Checkout logs confirm head `9e666719` was merged into main `d7db94fb`. The deployed source and workflow are unchanged by the rebase.
- All 236 deployed Node tests passed again after rebasing onto main, including the workflow's actual cleanup shell command after setup failures before and after run-directory creation, failure preservation and timeout-budget checks.
- All 57 CI policy tests passed again after rebasing onto main.
- `actionlint .github/workflows/deployed.yml` passed.
- `mix precommit` passed on the final stacked tree with the pinned Elixir/Erlang toolchain and an isolated test database: 5,598 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests; zero failures (core: 2 skipped, 7 excluded). The application code is identical in both PRs.

No deployed workflow was dispatched, environment enabled or live fixture modified.

### DIFF
--- a/.github/workflows/deployed.yml
+++ b/.github/workflows/deployed.yml
@@ -65,7 +65,7 @@ jobs:
   verify:
     needs: validate
     runs-on: ubuntu-latest
-    timeout-minutes: 60
+    timeout-minutes: 75 # Step caps total 71 minutes, including setup and both uploads.
     environment: deployed-${{ inputs.target || 'production' }}
     env:
       SUITE_TARGET: ${{ inputs.target || 'production' }}
@@ -82,17 +82,21 @@ jobs:
       FOUNTAIN_RECEIVER_ADMIN_KEY: ${{ secrets.FOUNTAIN_RECEIVER_ADMIN_KEY }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v5
+        timeout-minutes: 2
         with:
           persist-credentials: false
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        timeout-minutes: 2
         with:
           node-version: 24
       - name: Start scheduled monitor
+        timeout-minutes: 1
         if: github.event_name == 'schedule'
         env:
           SUITE_MONITOR_URL: ${{ secrets[github.event.schedule == '43 4 * * 6' && 'SUITE_MATRIX_MONITOR_URL' || 'SUITE_MONITOR_URL'] }}
         run: node deployed/monitor.mjs in_progress
       - name: Configure deployment observer
+        timeout-minutes: 2
         if: inputs.mode == 'rollout'
         env:
           SUITE_KUBECONFIG: ${{ secrets.SUITE_KUBECONFIG }}
@@ -108,15 +112,17 @@ jobs:
           kubectl version --client
       - name: Verify public ingress and clean fixtures
         id: suite
-        timeout-minutes: 52 # Leave time for evidence and a fresh cleanup process.
+        timeout-minutes: 52
         run: node deployed/ci.mjs
       - name: Finish scheduled monitor
+        timeout-minutes: 1
         if: always() && github.event_name == 'schedule'
         env:
           SUITE_MONITOR_URL: ${{ secrets[github.event.schedule == '43 4 * * 6' && 'SUITE_MATRIX_MONITOR_URL' || 'SUITE_MONITOR_URL'] }}
           MONITOR_STATUS: ${{ steps.suite.outcome == 'success' && 'ok' || 'error' }}
         run: node deployed/monitor.mjs "$MONITOR_STATUS"
       - name: Retain redacted evidence and cleanup manifests
+        timeout-minutes: 2
         if: always()
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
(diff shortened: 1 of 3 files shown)


BinaryBourbon (approved):
APPROVE — the retry cannot turn a red verification green, and it runs after the original evidence is already uploaded. Two budget/noise findings, both non-blocking.

**What this changes**

The suite step gets `timeout-minutes: 52` inside the existing 60-minute job, reserving eight minutes for the finalizers. A new step runs `deployed/cleanup-replay.mjs` against `$RUNNER_TEMP/deployed-<run>-<attempt>` when the suite step's outcome is `failure` or `cancelled`, after the original `deployed-...` artifact has been uploaded, and a second `deployed-cleanup-...` artifact then captures the in-place-updated manifests plus the replay evidence. The README gains an operator section for both the manual and the CI form.

**Why it cannot launder a failure**

`steps.suite.outcome` is already fixed when the retry runs, and nothing writes to it, so a clean retry leaves the job red. The scheduled monitor step runs *before* the retry and reads the same `steps.suite.outcome`, so a successful retry does not report `ok` either. A failing retry surfaces on its own: `cleanup-replay.mjs` exits 3 while any manifest is `failed` or `not_run` and 2 on a setup problem, and the step has no `continue-on-error`. The retry also cannot delete something the failure left half-created: a `pending` resource with no recorded id is resolved by exact unique name, and if the server has nothing matching, the manifest fails with `Unresolved create intent` rather than being marked clean (`deployed/lib/fixtures.mjs:190`). Nothing is deleted outside the dedicated suite account — see the selector analysis on #1965.

**Findings**

1. `.github/workflows/deployed.yml:129` — the four-minute step cap is shorter than the replay's own documented worst case, so the slow case this step exists for is the case it truncates. `cleanup-replay.mjs` stops *starting* manifests at 180 s, but the manifest in flight at that moment is inside `run()`, whose `finally` opens a fresh `AbortSignal.timeout(config.limits.cleanup_ms)` — 90 s, from `deployed/ci.mjs:33` — that the replay's abort does not shorten (`deployed/lib/runner.mjs`, cleanup in `finally`). On the `secrets`, `mcp` and `webhooks` profiles a receiver finalizer opens another `cleanup_ms` after that. So a `matrix-full` replay where two manifests need real work can legitimately need until 4:30, GitHub terminates at 4:00, the in-flight resources stay, and `replay.json` uploads at `status: "running"` with that entry `"running"`. The README states both halves ("After three minutes, replay attempts no more manifests; the current runner retains its separate cleanup deadline", then "a four-minute step timeout") without reconciling them. Either raise the step to six minutes or lower the 180 s in `cleanup-replay.mjs` to 150 s. Non-blocking: the job is already red and the artifact still uploads, since a timed-out step's outcome is `failure`, not `skipped`.
2. `.github/workflows/deployed.yml:122` — the retry step goes red on every setup-failure run, including the ones where nothing was created and nothing needs cleaning. `manifests()` throws when it finds no `cleanup.json` (`deployed/cleanup-replay.mjs:21`). Concrete: `SUITE_TARGET_JSON` is malformed, or `configFrom` rejects the target, so `ci.mjs` exits 2 with `$root` either absent or holding only `results/result.json` and no manifest — and no `target.json` in the first case. `steps.suite.outcome == 'failure'` is true, the retry runs, and the step prints "Cleanup replay setup failed; retain the original manifests and inspect the target configuration and evidence paths" and exits 2. The misleading part is the message: it points at a replay configuration problem that does not exist. The lasting cost is that operators learn to ignore a red cleanup step, and a genuine `cleanup_failed` replay then reads identically. Guard the step (`ls "$root"/results/**/cleanup.json` or a `test -d`), or let the script exit 0 on an empty tree. Non-blocking.
3. Soft, `deployed/README.md:213` — "After three minutes, replay attempts no more manifests" is the same arithmetic as finding 1; whichever number changes, this line changes with it.

**Checked and clean**

- Credentials reach the retry: `FOUNTAIN_SUITE_KEY` and the receiver admin keys are job-level `env`, not step-level, so the replay's `configFrom` finds them. `KUBECONFIG` is still present too, but cleanup mode never reaches the deployment adapter (`config.deployment && !manifestPath` in `runner.mjs`), so the retry performs no cluster read and no rollout check, and `Remove deployment credentials` still runs last under `if: always()`.
- Artifact names do not collide: `deployed-cleanup-...` is distinct from `deployed-...`, which matters because upload-artifact v4+ rejects a duplicate name. Uploading `results/` twice is deliberate and stated — the first artifact is the pre-retry journal, the second the post-retry one, since replay updates manifests in place. Both contents are the runner's already-redacted evidence; the retry adds no new unredacted path.
- Path agreement with the producer: `ci.mjs:63` builds `resolve(RUNNER_TEMP, 'deployed-<GITHUB_RUN_ID>-<GITHUB_RUN_ATTEMPT>')`, writes `target.json` at its root and passes `results` as `out`, which is exactly what the step reconstructs. Including for the matrix path — per-cell configs live in a `mkdtemp` directory that `runMatrix` deletes, so `$root/target.json` is the only surviving config, and it is adequate: matrix cells run `execution`/`streaming` and leave no receiver sidecar, so cleanup mode never asks for `config.webhooks`/`mcp`/`secrets`. Those profiles run single-cell, where `target.json` does carry their receiver settings.
- No missed trigger: the retry is skipped on `success`, and `success` from `run()`/`runMatrix()` already implies zero remaining resources — a cleanup failure or `remaining > 0` produces `cleanup_failed` (exit 3), which is a step failure.
- `always()` is what lets the step run on cancellation; the README links GitHub's cancellation reference and states plainly that a hard job timeout or runner loss defeats both the upload and the replay, and that journals lost with the runner are unrecoverable. That is the right shape for CLAUDE.md's "no doc describing unbuilt behavior as existing".
- 52 minutes leaves only ~2 minutes of slack against the maximum `matrix.limits.run_ms` (3,000,000 ms, `matrix.mjs`), which is intentional here: hitting the step cap yields `failure`, which is precisely the outcome that triggers the retry, and the retry fits in the remaining eight minutes.
- No `tee` in either new step, so no last-writer stall; no `gh` call, so the missing `workflow` scope is not in play.
- No CHANGELOG entry, correctly — CI wiring and operator instructions, not user-visible behavior. `deployed/README.md` sits outside `docs/`, so the three prose gates and the nav guardrails do not apply to it.

PR #1967, opened 2026-09-11
fix(files): return git roots in the sandbox namespace

### LINKED ISSUE #1908: repo_root is a host path on self-hosted runners while entry paths are sandbox paths
`repo_root` and `path` in the same response are expressed in **different namespaces** on a self-hosted runner (ADR 0022), so the flow ADR 0039's amendment tells a caller to use does not work there.

On the runner's process backend, `run/3` maps the working directory through `Managoat.Sandbox.host_path/2`, so `repo_root` comes back as the **host** path. Every entry's `path` in the same payload is the **sandbox** path. The ADR amendment tells a caller to combine `repo_root` with an entry path and call `/file`; on a runner that combination produces `422 path_outside_sandbox`.

## Scope

Affects both `GET /api/sandboxes/:id/diff` (already on `main`, shipped in SDK 1.15.0) and `GET /api/sandboxes/:id/git-status` (#1596). It is not a regression from #1596 and not a leak — the confinement hole that *was* a leak is fixed there, and `repo_root` is now redacted. This is a shape problem: the field is correct for the platform-paid providers and wrong for runners.

## Why it was not fixed in #1596

Closing it needs an **inverse of `host_path/2`** — a way to map a host path back into the sandbox namespace — which the `Managoat.Sandbox` seam does not currently expose. That is a change to the sandbox behaviour and its adapters, not to the endpoint.

## Options worth weighing

- Add the inverse mapping to the sandbox seam and return `repo_root` in the sandbox namespace everywhere, so the two fields in one response always agree.
- Or drop `repo_root` from the response and return the repository root as a sandbox-relative path, which is what a caller actually needs to combine with an entry path.
- Or document the divergence explicitly in ADR 0039 and the API reference, and stop telling callers to combine the two fields.

Filing this because it was raised and consciously deferred twice during review of #1596, and deferred findings that live only in review comments become folklore.

jhgaylor:
On a self-hosted runner, `/diff` and `/git-status` returned a host filesystem path in `repo_root`, while `/file` accepts sandbox paths. A client combining that root with an entry path received `422 path_outside_sandbox`. Both Git endpoints now return the repository root in the sandbox namespace, including nested directories and homes reached through symlinks.

The scripts pair each allowed host root with its original sandbox spelling. Repository discovery still checks the execution namespace before converting the matching root and suffix. Arguments remain literal argv values, root headers stay NUL-framed, and existing redaction and Git failure handling remain in place. This uses the mappings already available at Fountain's file boundary without adding a provider behavior callback or changing the public response schema.

Stack: based on #1931 (Git diff error handling). Closes #1908.

Validation:
- Both new regressions fail against the parent implementation with host roots where sandbox roots were expected.
- All 70 file-context, real Bash/Git script and controller tests pass. New coverage checks symlinked homes, newline-containing directory names, nested requests, and a returned root/entry round-trip through `SandboxFiles.read/3` using the host-path adapter seam.
- `mix precommit` passed on this stack: 5,287 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests; zero failures.

No real runner or deployed environment was modified.


BinaryBourbon (commented):
Reviewed `e16133566fbc539346959365de221f51477e188d`.

Reviewed incrementally over #1931. No actionable findings. Checked tagged host/logical-root argument pairs, physical-root confinement and mapping back into the sandbox namespace.

Validation: ran the exact-head `sandbox_files_script_test.exs` suite standalone with the exact-head `SandboxFiles` module and existing local dependency beams: **13 tests, 0 failures**, including the symlinked-home mapping case. Full app/runner integration was not run; standalone compilation emitted unrelated missing-module/stale-export warnings outside the exercised script path.

BinaryBourbon (approved):
**Approved** at `e16133566fbc539346959365de221f51477e188d`.

Reviewed incrementally over #1931. Tagged host/logical-root pairs preserve argument mapping, and repository discovery is confined in the execution namespace before the matching path is returned in the sandbox namespace. Rechecked nested/symlinked home handling and the file-read round-trip test. No blocking findings.

Validation: the exact-head real Bash/Git shell suite passed locally (**13 tests, 0 failures**), including symlinked host-root mapping, using the exact-head module and existing local dependency beams. Head/base are unchanged; [CI run 34624396266](https://github.com/managoat/fountain/actions/runs/34624396266) is verified at this head with required/static/contract and all six test partitions green. A real runner/provider smoke test and fresh full dependency build were not run locally.

BinaryBourbon (approved):
Approved at `6dc36977135c73dc44ac7772a4e6148986e158aa` after merging main `94f879c43e683a8237d50d3861246ad1a034472d`, which includes parent #1931. Re-reviewed the resolved tree: the final diff against main is exactly the child’s original four-file scope (+149/-73), with every added/deleted line matching the original child diff. Main's changelog, #1931 Git-error handling and #1915 redact-before-cap behavior are preserved. No blocking integration finding.

Validation: **13 real Bash/Git shell tests passed**, including mapped/symlinked roots, confinement and parent error/truncation regressions. Two additional focused tests passed against the merged public SandboxFiles API using mocked database/provider boundaries and real local Git/files: returned roots compose into a file read, and a secret crossing the read/diff cap remains redacted. Verified the committed source/test bytes match those tested. These were standalone checks with existing dependency beams; a fresh full app/real runner test was not run. Required CI must pass before merge.

PR #1968, opened 2026-09-11
test(sandboxes): prove reset and admission commit order

jhgaylor:
Adds four deterministic, two-connection PostgreSQL regressions for reset versus turn admission: both winner orderings with capacity `1` and `:unbounded`. The tests pause a real transaction after it acquires the sandbox advisory lock, observe the other backend blocked on that connection, and then verify persisted turns, the reset fence, runtime sessions and provider side effects. The provider callback also proves deletion runs outside the database transaction.

Refs #1767. This is a test-only unit for the reset fence already on main. Retirement and reassignment writers still require their own consistency audit and fixes; this draft does not close the issue.

Validation:
- All 29 new and existing reset tests passed.
- Removing the durable reset fence makes both reset-wins regressions fail; restoring it returns them to green. Mutation evidence was retained.
- Explicit cleanup verified zero users, sandboxes, conversations and turns after the focused tests.
- `mix precommit` passed: 5,379 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests; zero failures.

No real provider was called.

### DIFF
--- a/apps/fountain/test/fountain/conversations/reset_admission_order_test.exs
+++ b/apps/fountain/test/fountain/conversations/reset_admission_order_test.exs
@@ -1,0 +1,183 @@
+defmodule Fountain.Conversations.ResetAdmissionOrderTest do
+  use Fountain.DataCase, async: false
+  use Mimic
+
+  alias Ecto.Adapters.SQL.Sandbox
+  alias Fountain.Conversations
+
+  for first <- [:reset, :admission], capacity <- [1, :unbounded] do
+    test "#{first} wins against #{inspect(capacity)} admission on independent connections" do
+      assert_order(unquote(first), unquote(capacity))
+    end
+  end
+
+  defp assert_order(first, capacity) do
+    Sandbox.unboxed_run(Repo, fn ->
+      user = insert_verified_user()
+      agent = insert_agent(user_id: user.id)
+
+      home =
+        insert_sandbox(user_id: user.id, agent_id: agent.id, mode: "persistent", status: "ready")
+
+      conv =
+        insert_conversation(
+          user_id: user.id,
+          agent: agent,
+          sandbox: home,
+          status: "idle",
+          runtime_session_id: "before-reset"
+        )
+
+      owner = self()
+
+      operation = fn
+        :reset ->
+          Conversations.reset_sandbox(home)
+
+        :admission ->
+          Conversations._unsafe_create_turn_on_sandbox(
+            %{
+              conversation_id: conv.id,
+              turn_number: 1,
+              status: "running",
+              prompt: "race"
+            },
+            home.id,
+            capacity
+          )
+      end
+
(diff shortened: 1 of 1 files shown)


BinaryBourbon (approved):
## APPROVE

Verified non-vacuous by reverting the guard, not just by watching it go green.

**What it proves.** Both `do_reset_sandbox/2` and `_unsafe_create_turn_on_sandbox/4` open with `pg_advisory_xact_lock(4316, phash2(sandbox_id))`. The test pauses the winner inside the telemetry event Ecto emits *after* that lock is granted, starts the loser on a second real connection, and then proves the wait with `pg_blocking_pids` rather than assuming it — so it waits on exactly the thing it asserts.

**Reversion check.** With the advisory lock deleted from `_unsafe_create_turn_on_sandbox/4` (nothing else changed), all 4 cases fail, three of them on `"no wait on the winning PostgreSQL connection observed"`:

```
4 tests, 4 failures
```

Restored, `4 tests, 0 failures`. It is measuring the fence.

**Two details worth calling out as good.** `assert_receive {:destroying, reset_pid, false}` pins the provider destroy to *outside* the transaction, which is the rule that actually matters here and is easy to regress silently. And the loser's outcome is asserted in both directions — `:sandbox_unavailable` when reset commits first, `:sandbox_mid_turn` when the turn does — so a fence that serialized but then decided wrongly would still be caught.

`async: false` and `Sandbox.unboxed_run/2` are genuinely required (two independent backends, real lock waits), and the `after` blocks delete the rows the unboxed run leaves behind. ExUnit runs sync modules after the async ones, so there is no interleaving with the sandboxed suite.

### Non-blocking nits

1. The generated names read wrong for half the matrix: `test "admission wins against 1 admission on independent connections"` — the opponent is the reset, not a second admission. `"#{first} wins against #{inspect(capacity)} admission"` wants to be something like `"#{first} commits first at #{inspect(capacity)} capacity"`.
2. `%{query: query, params: [4316, _]}` hardcodes `@sandbox_lock_namespace`. If that attribute ever moves, the handler stops matching — the failure is loud (`await_blocked` times out) rather than silent, so this is cosmetic, but a comment naming the attribute would save the next reader a grep.
3. No comment says *why* `async: false`; a one-liner pointing at `unboxed_run` would stop someone "fixing" it later.

jhgaylor:
⚠️ **Before merging this chain:** the campaign it starts lands ~72 lines over the `conversation_server.ex` size pin, and no PR in it is red. See #2032.

`main` is now at `@pin 2646` with the file at exactly 2646 lines (zero headroom) after #1749 and #1751. Every PR in this stack is measured against a base where the pin is still 2762, so each is green on its own; merging the chain tip onto current `main` auto-merges cleanly with no conflict on the pin line, then fails at 2673 > 2646.

Nothing warns you at merge time, which is why this note is here rather than only on the tracker. #2032 has the per-branch arithmetic and the two options.

PR #1969, opened 2026-09-11
fix(sandboxes): lock availability through turn admission

### LINKED ISSUE #1767: Make sandbox reset and retirement atomic with turn admission
## Problem

Follow-up to #1764: its admission checks reject a sandbox that is already retired or a conversation that has already moved. They do not make admission atomic with retirement/reset.

Confirmed by static inspection at #1764 revision 9f236ef0400065dfbd61a5afa14e1615dfe874c7; no concurrent runtime reproduction was performed during this review.

## Concrete reset interleaving

1. `do_reset_sandbox/2` takes the per-sandbox advisory lock, checks that no turns are running, and clears runtime session IDs.
2. That transaction commits and releases the advisory lock while the sandbox still has a nonterminal status.
3. A user or autonomous admission takes the same lock, sees the still-attached nonterminal sandbox, and inserts a running turn.
4. Reset then notifies actors and calls `_unsafe_retire_home/1`, destroying the provider machine and marking the sandbox terminated underneath the admitted turn.

Evidence: `apps/fountain/lib/fountain/conversations.ex`, `_unsafe_create_turn_on_sandbox/3` (around 1061), `do_reset_sandbox/2` (around 2356), and `_unsafe_retire_home/1` (around 2305). The reset docstring currently claims the running-turn check and row flip share the lock, but the row flip happens afterward.

Other retirement paths also need examination: #1761's row lock protects terminal-to-active updates, but does not coordinate with admission's advisory lock. A retirement may commit between admission's existence check and turn insertion. Reassignment needs the same consistency audit.

## Acceptance criteria

- Make reset's eligibility check and durable transition to a state that refuses admission atomic with admission, using a consistent lock order/protocol.
- If reset wins, subsequent user and autonomous admission must write no turn. If admission wins, reset must return `:sandbox_mid_turn` and leave the active machine intact.
- Keep provider destroy calls and actor notifications outside database locks; failed cleanup must remain recoverable by reconciliation.
- Audit retirement and reassignment writers for the corresponding check/insert race and document the guarantees for forced retirement.
- Add deterministic two-connection concurrency regressions for both orderings, including bounded and unbounded admission. Verify persisted turn/sandbox state and provider side effects.
- Correct the reset docstring to match the implemented guarantee.

This is the explicitly deferred race in #1764, separate from #1766 (handling retirement rejection in provisioning/wake callbacks). Related broader work: #1732.

jhgaylor:
A sandbox retirement could commit between turn admission's availability check and its insert, because retirement writes take a row lock while admission only held the sandbox advisory lock. Admission now also keeps the joined sandbox row share-locked through insertion. If retirement commits first, no turn is inserted; if admission wins, a forced row retirement waits until that turn commits.

Stack: based on #1968. Refs #1767. This unit coordinates the database availability check with row-based retirement writers. A forced retirement can still retire a previously admitted turn; its caller remains responsible for stopping the actor and cleaning up the provider. Paths that destroy the provider before persisting a refusal state and the remaining reassignment audit are separate follow-ups, so this does not close #1767.

Validation:
- All four new real PostgreSQL regressions fail on the parent because the competing backend does not wait; both winner orderings with capacity `1` and `:unbounded` pass with the row lock.
- All 44 retirement-order, reset-order, existing reset and saved-allowance admission tests passed.
- Explicit cleanup verified zero users, sandboxes, conversations and turns afterward.
- `mix precommit` passed: 5,383 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests; zero failures.

No real provider was called. The change adds a shared row lock to the existing transaction, with no provider I/O or audit writes under it.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1705,6 +1705,9 @@ defmodule Fountain.Conversations do
           Repo.rollback(:configuration_changed)
         end
 
+        # Row-only retirement writers do not take the admission advisory
+        # lock. Hold their row through insertion too: a committed retirement
+        # refuses admission, while a later forced retirement follows the turn.
         attached? =
           Repo.exists?(
             from c in Conversation,
@@ -1712,7 +1715,8 @@ defmodule Fountain.Conversations do
               on: s.id == c.sandbox_id,
               where:
                 c.id == ^conv_id and s.id == ^sandbox_id and c.user_id == s.user_id and
-                  s.status not in ["terminated", "failed"] and is_nil(s.reset_requested_at)
+                  s.status not in ["terminated", "failed"] and is_nil(s.reset_requested_at),
+              lock: "FOR SHARE"
           )
 
         unless attached?, do: Repo.rollback(:sandbox_unavailable)
--- a/apps/fountain/test/fountain/conversations/reset_admission_order_test.exs
+++ b/apps/fountain/test/fountain/conversations/reset_admission_order_test.exs
@@ -1,0 +1,183 @@
+defmodule Fountain.Conversations.ResetAdmissionOrderTest do
+  use Fountain.DataCase, async: false
+  use Mimic
+
+  alias Ecto.Adapters.SQL.Sandbox
+  alias Fountain.Conversations
+
+  for first <- [:reset, :admission], capacity <- [1, :unbounded] do
+    test "#{first} wins against #{inspect(capacity)} admission on independent connections" do
+      assert_order(unquote(first), unquote(capacity))
+    end
+  end
+
+  defp assert_order(first, capacity) do
+    Sandbox.unboxed_run(Repo, fn ->
+      user = insert_verified_user()
+      agent = insert_agent(user_id: user.id)
+
+      home =
+        insert_sandbox(user_id: user.id, agent_id: agent.id, mode: "persistent", status: "ready")
+
+      conv =
+        insert_conversation(
+          user_id: user.id,
+          agent: agent,
+          sandbox: home,
+          status: "idle",
+          runtime_session_id: "before-reset"
+        )
+
+      owner = self()
+
+      operation = fn
+        :reset ->
+          Conversations.reset_sandbox(home)
+
+        :admission ->
+          Conversations._unsafe_create_turn_on_sandbox(
+            %{
+              conversation_id: conv.id,
+              turn_number: 1,
+              status: "running",
+              prompt: "race"
+            },
+            home.id,
+            capacity
+          )
+      end
+
+      winner = independent(first, fn -> operation.(first) end, owner, true)
+
+      try do
+        assert_receive {:locked, winner_pid}, 5_000
+        assert winner_pid == winner.pid
+        second = if first == :reset, do: :admission, else: :reset
+        waiter = independent(second, fn -> operation.(second) end, owner, false)
+
+        try do
+          assert_receive {:backend, ^first, winner_backend}, 5_000
+          assert_receive {:backend, ^second, waiter_backend}, 5_000
+          refute winner_backend == waiter_backend
+
+          await_blocked(
+            waiter_backend,
+            winner_backend,
+            System.monotonic_time(:millisecond) + 5_000
+          )
+
+          assert Repo.aggregate(
+                   from(t in Conversations.Turn, where: t.conversation_id == ^conv.id),
+                   :count
+                 ) == 0
+
+          refute_received {:destroying, _, _}
+          send(winner.pid, :continue)
+
+          if first == :reset do
+            assert_receive {:destroying, reset_pid, false}, 5_000
+            assert reset_pid == winner.pid
+            assert {:error, :sandbox_unavailable} = Task.await(waiter, 5_000)
+            assert Repo.reload!(home).reset_requested_at
+            assert Repo.reload!(home).status == "ready"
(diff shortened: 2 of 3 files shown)


BinaryBourbon (approved):
APPROVE — the lock is at the right strength, in the right place, and cannot invert the established order.

**What this changes**

`_unsafe_create_turn_on_sandbox/4`'s availability probe (`apps/fountain/lib/fountain/conversations.ex:1710-1721`) now carries `lock: "FOR SHARE"`, so the joined `sandboxes` row is held from the check through the turn insert. That closes the window against writers that take only the sandbox row and not the per-sandbox advisory lock — `update_sandbox/2` being the one that matters. Under READ COMMITTED the locking read re-evaluates its quals against the row version the blocking writer committed (EvalPlanQual), so a retirement that commits first makes `attached?` false and the turn is never inserted; an admission that gets there first makes the retirement wait until the turn commits.

**Findings**

1. *Non-blocking, semantics worth stating in the docstring.* When admission wins, the retirement is **not** refused — `update_sandbox/2` deliberately lets terminal transitions through a fence (`conversations.ex:261-272`), and the test asserts exactly that (`retirement_admission_order_test.exs:113-116`: `{:ok, %{status: "terminated"}}` with the turn still `running`). So the invariant this patch buys is one-directional: *no turn is admitted after a committed retirement*. It leaves a `running` turn row on a `terminated` sandbox, and nothing in the programme reconciles that row (the #2021 item-5 shape). The PR body says this ("a forced retirement can still retire a previously admitted turn; its caller remains responsible"); the `@doc` on `_unsafe_create_turn_on_sandbox/4` does not. One sentence there would save the next reader the same derivation.

2. *Non-blocking, new coupling to flag.* The sandbox row is now share-locked by admission, and one holder of the **global fleet advisory lock** wants that row `FOR UPDATE`: `resume_and_wake/1` calls `update_sandbox/2` inside `Quotas.with_sandbox_reservation/3` (`conversations.ex:4756-4790`), which holds `pg_advisory_xact_lock(@fleet_lock_key)`. Admission on a `suspended` row is permitted by the probe's own predicate (only `terminated`/`failed` are excluded), so a co-tenant admitting on a shared home can now make a wake — and therefore every tenant's provisioning — wait. The wait is bounded because the admission transaction is local queries only, and it is the same category the `@rotation_lock_timeout_ms` comment at `conversations.ex:2787-2795` already reasons about ("Turn admission holds that row for a handful of local queries"), which still reads true after this patch. Worth knowing, not worth changing here.

3. *Nit.* `sandbox_identity.ex:50` hardcodes `4316` where `@sandbox_lock_namespace` is the same value. Pre-existing, but this stack is the one making that lock load-bearing, so it is a cheap tidy for whoever is next in the file.

**Checked and clean**

- **No lock-order inversion.** Every transaction that touches both a conversation row and a sandbox row takes the per-sandbox advisory lock first, and all of them order advisory → conversation → sandbox: admission (`1684`/`1697`/`1719`), `fail_initial_start/2` (`3099-3105`, whose comment already declares that order), and `reapply_conversation/3` (`with_sandbox_lock` → `Repo.one! ... FOR UPDATE` at `1220` → `update_sandbox` via `Reapply.update_identity/4`). The one path that takes the sandbox row *before* conversation rows — `do_reset_sandbox/2` (`3487-3568`, sandbox `FOR UPDATE` then `update_all` over conversations) — holds the same advisory lock, so it can never be interleaved with admission. The two writers that skip the advisory lock (`update_sandbox/2`, `SandboxIdentity.bind/2`) take the sandbox row and nothing else. No cycle exists.
- **No path takes the fleet lock after this row lock**, so the global-before-per-user order in `Quotas.with_sandbox_reservation/3` (`quotas.ex:241-246`) is untouched, and `:fleet_full` (503) still precedes `Credits.gate/1` (402) in the same `with`. Neither becomes reachable from admission.
- **Lock strength is right.** `FOR SHARE` conflicts with `FOR NO KEY UPDATE`/`FOR UPDATE` (so it blocks any status write, including a plain `Repo.update`) but not with the `FOR KEY SHARE` an FK reference takes — so a second conversation attaching to the same home is not blocked. `FOR KEY SHARE` would not have blocked a status change, and `FOR UPDATE` would have serialised concurrent admissions needlessly.
- **`Repo.exists?/2` keeps the lock**: `Ecto.Repo.Queryable.exists?/3` excludes `:select`, `:preload` and `:distinct` only, so `FOR SHARE` reaches the SQL. With no `OF` clause it locks both joined tables; the conversation row is already `FOR UPDATE` from `1697`, so that half is a no-op.
- **No I/O under the lock**: the transaction is advisory lock, two selects, the allowance read, the capacity count, one insert. `record_turn_usage/1` stays outside (`1744-1747`).
- **No second write path.** `_unsafe_create_turn_on_sandbox/4` has exactly two production callers (`conversations/connection.ex:279`, `conversations/turn_machine.ex:1042`); `_unsafe_create_turn/1` has none outside tests.
- **The test waits on the right signal.** `await_blocked/3` polls `pg_blocking_pids` rather than sleeping, the pause is a telemetry hook inside the winner's transaction (so the locks are genuinely held), the pause point is deliberately the row query rather than the advisory lock, and `refute_received {:destroying, _, _}` pins that no provider call happens inside. If the query-text match ever stops matching, `assert_receive {:locked, _}` times out — it fails loudly rather than passing vacuously. `async: false` plus `Sandbox.unboxed_run` is required for real backends and is the pattern #1968 established.
- All four orderings × capacities are covered, including `:unbounded`, which skips the capacity check but not the fence.

PR #1970, opened 2026-09-11
fix(conversations): detach wakes when a reset wins

jhgaylor:
When a reset fences a sandbox during wake preparation, the final ready write returns `:sandbox_reset_pending`. The callback treated that expected race as a MatchError. It now stops normally, releases only this wake's prepared credentials, and preserves the existing disk, reset fence and any replacement conversation/session. Unrelated write errors keep their existing failure behavior.

Stack: based on #1946, following the retirement callback fixes. Refs #1766 and #1767. This is the wake unit; provisioning and forced-teardown fencing remain separate work.

Validation:
- Both new ready/suspended reset-fence wake cases fail on the parent with MatchError.
- All 51 broker and wake tests passed. The new cases verify unchanged sandbox state/fence, replacement session survival, original token and callback-key revocation, and no runtime spawn or machine deletion.
- `mix precommit` passed: 5,392 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests; zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -26,6 +26,11 @@ defmodule Fountain.Conversations.ConversationServer do
   alias Fountain.Conversations.{ProvisionWatchdog, Reapply}
   alias Fountain.Conversations.{Reattachment, Redaction, SpriteEnv, TurnLaunch, TurnMachine}
 
+  defguardp retired_or_resetting(reason)
+            when reason == :sandbox_reset_pending or
+                   (is_struct(reason, Ecto.Changeset) and
+                      reason.errors == [status: {"sandbox is retired", []}])
+
   # ── public api ────────────────────────────────────────────────────────────
 
   def start_link(args) do
@@ -1166,7 +1171,7 @@ defmodule Fountain.Conversations.ConversationServer do
 
           {:noreply, new_state}
 
-        {:error, %Ecto.Changeset{errors: [status: {"sandbox is retired", []}]}} ->
+        {:error, reason} when retired_or_resetting(reason) ->
           # Wake owns this connection's credentials, not the existing disk or
           # another connection's session. Never destroy the machine here.
           Egress.release_prepared({:ok, state})
--- a/apps/fountain/test/fountain/conversations/conversation_server_broker_test.exs
+++ b/apps/fountain/test/fountain/conversations/conversation_server_broker_test.exs
@@ -178,7 +178,8 @@ defmodule Fountain.Conversations.ConversationServerBrokerTest do
       end
     end
 
-    for initial <- ["ready", "suspended"], terminal <- ["terminated", "failed"] do
+    for initial <- ["ready", "suspended"],
+        terminal <- ["terminated", "failed", "reset_pending"] do
       @tag initial: initial, terminal: terminal
       test "retirement during #{initial} wake preserves #{terminal} and the replacement", %{
         user: user,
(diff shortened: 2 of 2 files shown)


BinaryBourbon (approved):
APPROVE — the widened clause covers a reachable crash, and it is the narrowest widening that does.

**What this changes**

`update_sandbox/2` answers a fenced row two different ways depending on which fence won: a terminal row fails the changeset with `[status: {"sandbox is retired", []}]`, while a row carrying `reset_requested_at` rolls back with the bare atom `:sandbox_reset_pending` (`conversations.ex:261-272`). The reattach path's ready-write only knew the first shape, so a reset that fenced the home while the wake was talking to the provider reached `error -> raise MatchError`. A new `defguardp retired_or_resetting/1` (`conversation_server.ex:37-40`) accepts both, and the existing graceful arm — release this wake's prepared credentials, stop `:normal`, touch neither the disk nor another connection's session — now handles it.

This one is genuinely reachable today: `do_reset_sandbox/2` only fences rows in `ready` or `suspended`, which is exactly the state a reattach starts from.

**Findings**

1. *Blocking for the merge, not for this patch — and it belongs at the bottom of the chain, not here.* `apps/fountain/test/fountain/conversations/conversation_server_size_test.exs` carries `@pin 2762` on this branch, but `main` lowered it to **2646** (an exact record of the file, per that file's own moduledoc "never raise it") in `94f879c4`/`4f0c040c` after this chain forked at `3f59a76a` on 2026-09-11. `conversation_server.ex` is 2707 lines on your base and 2712 after this patch, so the pin passes here only because the base carries the stale, looser number. Rebased onto current `main` the file lands at roughly 2646 + 27 = 2673 against a pin of 2646 and `conversation_server_size_test` goes red — for the whole #1766/#1767 chain, not just this PR. Nothing to fix in this diff; whoever rebases the chain owns it, and the 5 lines this stage adds are the part that needs an offsetting move or an explicit justification.

2. *Non-blocking, pre-existing brittleness now inherited by the guard.* The changeset arm matches the error list **exactly**: `reason.errors == [status: {"sandbox is retired", []}]`. A changeset that ever carries a second error alongside the revival one falls through to `raise MatchError`. The guard did not introduce this — it copied it from the clause it replaced — but it is now written once in a guard used by three call sites (after #1971), which makes it the right place to relax to a membership test if anyone ever wants to.

3. *Non-blocking.* `update_sandbox/2`'s third answer, `{:error, :not_found}` (the row was deleted under the wake — account deletion cascades sandboxes), still reaches `raise MatchError`. Out of scope here; #1976/#1977 are the account-teardown stages.

4. *Non-blocking, observation about where the fence is read.* The wake checks the fence only at the tail write, so a reset that commits at the start of a reattach still pays a full provider round trip — resume, runtime config, CA install, `.env`, instructions, `prepare_runtime_sprite` — writing files onto a disk that is about to be destroyed. Harmless, and #1971 adds the pre-flight check on the provisioning side rather than this one, so this is a note rather than a request.

**Checked and clean**

- The guard is `defguardp`, defined before first use, and `is_struct/2` precedes the `reason.errors` map access, so a non-struct reason cannot raise inside the guard.
- Clause order is safe: the widened arm sits above `error -> raise MatchError`, and nothing below it matched `:sandbox_reset_pending` before.
- No other site in the server is left half-converted: `grep "sandbox is retired"` finds exactly three matches in `conversation_server.ex` (852, 1012, 1224 — the other two are #1971's) and `Lifecycle.park_row/1` (`lifecycle.ex:387-391`), which already handles **both** shapes with explicit clauses. Sites that write a terminal status (`621`, `787`, `827`, `855`, `1004`, `1016`) pass the fence by construction, so their `{:ok, _} =` matches are still sound.
- The test does fail on the parent: `update_sandbox(sandbox, %{status: "ready"})` on a row with `reset_requested_at` set rolls back `:sandbox_reset_pending`, which the old clause did not match. The new `reset_pending` case asserts the fence *and* `terminated_at` *and* the original status are all preserved, plus replacement-conversation survival and callback-key revocation.
- `Egress.release_prepared({:ok, state})` is the pre-existing call for "this connection's credentials only"; the machine is not destroyed on this path, which is what the comment promises.
- *Nit:* with `terminal <- [..., "reset_pending"]` the generated test name reads "retirement during ready wake preserves reset_pending", which is no longer a retirement. `fence <- [...]` would name it honestly.
- *Nit:* the `reset_pending` case sets `reset_requested_at` with a raw changeset rather than `reset_sandbox/2`, so it fabricates the fence instead of producing it. Right call here (the real path does provider I/O), but it means the test proves the *response* to a fence and not that a real reset produces one.
- No CHANGELOG entry — consistent with the rest of the programme, noted once across this batch and not held against this PR.

PR #1971, opened 2026-09-11
fix(conversations): preserve reset fences during provisioning

jhgaylor:
Provisioning now treats a winning reset fence like a winning retirement at both the starting write and the final ready write. Before startup, it exits without creating resources. After provisioning, it releases this attempt's handle and credentials without marking the fenced sandbox or replacement conversation failed.

Stack: based on #1970, reusing its reset/retirement rejection guard. Refs #1766 and #1767. This is the provisioning unit; forced-teardown fencing remains separate work.

Validation:
- Both new startup/completion fence cases fail on the parent.
- All 64 broker, provisioning-retirement and lifecycle tests passed, including original callback-key revocation and replacement-session survival. Unrelated write failures still follow their existing failure paths.
- A test fixture now rereads the persisted starting row before setting its fence, so state preservation is checked against the current database value.
- `mix precommit` passed: 5,394 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests; zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -795,7 +795,7 @@ defmodule Fountain.Conversations.ConversationServer do
         {:ok, _} ->
           do_fresh_provision_inner(state, conv, sandbox, agent, env, secrets)
 
-        {:error, %Ecto.Changeset{errors: [status: {"sandbox is retired", []}]}} ->
+        {:error, reason} when retired_or_resetting(reason) ->
           # No resources were created yet. Leave the winning retirement and
           # any replacement conversation alone, without announcing a start.
           {:stop, :normal, state}
@@ -955,7 +955,7 @@ defmodule Fountain.Conversations.ConversationServer do
           # queue_initial_prompt/3.
           {:noreply, new_state}
         else
-          {:error, %Ecto.Changeset{errors: [status: {"sandbox is retired", []}]}} ->
+          {:error, reason} when retired_or_resetting(reason) ->
             # This handle and token belong to this attempt. Do not fail the
             # conversation or release every session: a replacement may own it.
             _ = Managoat.Sandbox.destroy(handle)
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -26,6 +26,11 @@ defmodule Fountain.Conversations.ConversationServer do
   alias Fountain.Conversations.{ProvisionWatchdog, Reapply}
   alias Fountain.Conversations.{Reattachment, Redaction, SpriteEnv, TurnLaunch, TurnMachine}
 
+  defguardp retired_or_resetting(reason)
+            when reason == :sandbox_reset_pending or
+                   (is_struct(reason, Ecto.Changeset) and
+                      reason.errors == [status: {"sandbox is retired", []}])
+
   # ── public api ────────────────────────────────────────────────────────────
 
   def start_link(args) do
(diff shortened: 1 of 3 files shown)


BinaryBourbon (approved):
APPROVE — the two remaining sites get the same guard, and the ordering against #1972 is the right way round.

**What this changes**

The fresh-provision path had the same single-shape match #1970 just fixed on the reattach path, in both of its windows: the `starting` write before any resource exists (`conversation_server.ex:849-857`) and the `ready` write after the sprite is up (`1009-1016`). Both now use `retired_or_resetting/1`, so a reset fence lands the same way a retirement already did — exit before startup without creating anything, or destroy *this attempt's* handle and release *this attempt's* credentials without failing the conversation or the replacement.

**Findings**

1. *Non-blocking, and the reason it reads as dead code — worth one line in the PR body.* Neither of these two windows is reachable from the current reset path. `do_reset_sandbox/2` refuses anything that is not `ready` or `suspended` (`conversations.ex:3495-3497`), and its docstring gives the reason: "`pending` and `starting` as much as `terminated` and `failed` … A machine still being built has no disk to replace". A row in `pending`/`starting` therefore cannot carry `reset_requested_at` today, and `reset_requested_at` has exactly one production writer (`conversations.ex:3517`). What makes these guards live is **#1972**, whose `fence_home_destruction/1` stamps the fence on *every* non-terminal status, `starting` included. So this stage is hardening ahead of the stage that creates the state, which is the correct order for a stack — it just means "fails on the parent" (true, because the test fabricates the fence) is not the same as "fixes a reachable crash", and the body reads as though it were. Soft finding; nothing to change in the diff.

2. *Non-blocking, provider-side asymmetry worth confirming is deliberate.* At `1012` the fenced landing calls `Managoat.Sandbox.destroy(handle)`, while the agent-delete teardown goes on to call `_unsafe_retire_home/1`, which destroys `sandbox.sprite_name` too. If this attempt's handle and the row's `sprite_name` are the same machine, that is two destroys on one sprite. Both are best-effort (`_unsafe_retire_home/1` logs a destroy error and retires the row anyway; the `_ =` here discards the result), so the outcome is a warning in the log, not a stuck row. Fine, but if the second destroy ever stops being best-effort this is where it bites.

3. *Nit, same as on #1970.* The generated names now say "retirement during provision preserves the reset_pending row", which is a fence rather than a retirement, and the `reset_pending` branches fabricate `reset_requested_at` with a raw changeset instead of going through `reset_sandbox/2`.

**Checked and clean**

- `retired_or_resetting/1` is already in scope from #1970; both new clauses sit above their `error -> raise MatchError` / `else` fallthrough, and neither shadows an arm that previously handled the atom.
- With these two, all three `update_sandbox`-result matches in `conversation_server.ex` that can see a fence are converted (849, 1009, 1224) and `Lifecycle.park_row/1` already handled both shapes. I found no fourth site in the server or in `Conversations.*` that matches the retired changeset without also matching `:sandbox_reset_pending`.
- The `849` arm is correct to create nothing and announce nothing: it runs before `do_fresh_provision_inner/6`, so there is no handle, no token and no `provision` stage event to retract, and the `{:stop, :normal, state}` keeps the fenced row and any replacement conversation untouched.
- The `1012` arm's comment is accurate about ownership — it releases this attempt's handle and callback key only, and the broker test asserts the *original* callback key survives and the replacement conversation keeps its sandbox and `idle` status.
- Fixture change at `conversation_server_broker_test.exs:641-648` (`Repo.reload!` before stamping the fence) is a real improvement: the struct captured before the `starting` write would otherwise have stamped the fence onto a stale row and silently reverted the status, which is exactly the kind of fixture bug that makes a fence test assert on the wrong thing.
- No line-count change to `conversation_server.ex` (2712 both sides). The stale size pin problem I raised on #1970 is inherited here, not worsened.

PR #1972, opened 2026-09-11
fix(sandboxes): fence admission before forced home teardown

jhgaylor:
Deleting an agent could terminate its conversations and destroy its home while another connection still admitted a new turn. Forced home teardown now commits the existing admission fence under the sandbox advisory and row locks before actor shutdown or provider I/O. The fenced row keeps its quota slot until teardown completes. Already admitted turns can still be interrupted by this explicitly forced operation.

Teardown refuses an enclosing database transaction before side effects. Agent deletion propagates fencing errors instead of deleting the agent row after a refused teardown. Existing best-effort handling of provider deletion errors is unchanged.

Stack: based on #1971, including graceful wake/provision callback handling when a fence wins. Refs #1767. Other retirement/reassignment paths still need the remaining audit; this does not close the issue.

Validation:
- All five new regressions fail against the parent: missing fences before actor/provider work and unsafe teardown inside enclosing transactions.
- All 108 forced-home, agent, reset and orphaned-home tests passed. Coverage checks bounded/unbounded admission refusal, quota retention during provider deletion, actor shutdown ordering, completed row state and agent retention on refusal.
- Full `mix precommit` passed: 5,399 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures. The first full run exposed missing extension tables in the isolated local database; after applying the umbrella migrations, the complete gate passed with no code change.

### DIFF
--- a/apps/fountain/lib/fountain/agents.ex
+++ b/apps/fountain/lib/fountain/agents.ex
@@ -276,8 +276,13 @@ defmodule Fountain.Agents do
     # (ADR 0023 step 5). Torn down before the row goes, while `agent_id` still
     # names it — deletion would nilify the pointer and orphan the sprite.
     # Ownership: `agent` came from the caller's scoped fetch.
-    _ = Fountain.Conversations._unsafe_destroy_homes_for_agent(agent.id)
+    with count when is_integer(count) <-
+           Fountain.Conversations._unsafe_destroy_homes_for_agent(agent.id) do
+      delete_agent_row(agent, opts)
+    end
+  end
 
+  defp delete_agent_row(agent, opts) do
     # Unpin the agent's conversations from its versions before the row goes.
     # Deleting the agent cascades to `agent_versions`, and Postgres re-checks
     # `conversations.agent_version_id` while it is still setting
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -3649,28 +3649,68 @@ defmodule Fountain.Conversations do
   the sprite is destroyed and the row terminated. Best-effort per machine; a
   provider error is logged and the row still retires, so the reaper's sweep
   sees a terminal row rather than a live one nobody can find. Returns the
-  number of homes torn down. `_unsafe_`: the caller owns the agent.
+  number of homes torn down, or a fencing error. Refuses an enclosing database
+  transaction before any teardown. Admission is fenced before actor shutdown
+  and provider I/O; already admitted turns may be interrupted by this forced
+  operation. `_unsafe_`: the caller owns the agent.
   """
   def _unsafe_destroy_homes_for_agent(agent_id) when is_binary(agent_id) do
-    from(s in Sandbox,
-      where:
-        s.agent_id == ^agent_id and s.mode == "persistent" and
-          s.status not in ["terminated", "failed"]
-    )
-    |> Repo.all()
-    |> Enum.map(&_unsafe_destroy_home/1)
-    |> length()
+    if Repo.in_transaction?() do
+      {:error, :provider_transaction_open}
+    else
+      from(s in Sandbox,
+        where:
+          s.agent_id == ^agent_id and s.mode == "persistent" and
+            s.status not in ["terminated", "failed"]
+      )
+      |> Repo.all()
+      |> Enum.reduce_while(0, fn home, count ->
+        case _unsafe_destroy_home(home) do
+          :ok -> {:cont, count + 1}
+          {:error, _} = error -> {:halt, error}
+        end
+      end)
+    end
   end
 
   @doc false
   def _unsafe_destroy_home(%Sandbox{} = sandbox) do
-    sandbox = Repo.preload(sandbox, :conversations)
+    if Repo.in_transaction?() do
+      {:error, :provider_transaction_open}
+    else
+      with {:ok, fenced} <- fence_home_destruction(sandbox) do
+        fenced = Repo.preload(fenced, :conversations)
 
-    sandbox.conversations
-    |> Enum.reject(&(&1.status in ["terminated", "failed"]))
-    |> Enum.each(&ConversationServer.terminate_conversation(&1.id, actor: "system:home_reset"))
+        fenced.conversations
+        |> Enum.reject(&(&1.status in ["terminated", "failed"]))
+        |> Enum.each(
+          &ConversationServer.terminate_conversation(&1.id, actor: "system:home_reset")
+        )
 
-    _unsafe_retire_home(sandbox)
+        _unsafe_retire_home(fenced)
+      end
+    end
+  end
+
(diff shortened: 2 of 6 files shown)


BinaryBourbon (approved):
APPROVE — the fence commits in the right place, under the right locks, and the one leak it can create is the state `reset_sandbox/2` already documents as supported.

**What this changes**

Deleting an agent tore its homes down by terminating the conversations and destroying the sprite, with nothing stopping a concurrent connection admitting a turn in between. `fence_home_destruction/1` (`conversations.ex:3434-3453`) now stamps `reset_requested_at` under the per-sandbox advisory lock and the row's own `FOR UPDATE`, and commits that before any actor shutdown or provider I/O — so `_unsafe_create_turn_on_sandbox/4`'s `is_nil(s.reset_requested_at)` predicate refuses admission from the moment the fence lands. Both entry points refuse an enclosing transaction first, and `Agents.delete_agent/2` now propagates a fencing error instead of deleting the agent row over a teardown that did not happen.

**Findings**

1. *Non-blocking — a leaked fence is unreachable by the reaper and holds a fleet slot, and this door does not document the way out.* A fence that commits without the matching retirement leaves a non-terminal row with `reset_requested_at` set. Every reaper pass excludes exactly those rows (`sandbox_reaper.ex:125`, `:228`), and `Quotas` counts them as active (`quotas.ex:141-147`), so the row is invisible to automatic recovery while holding one tenant slot *and* one of the deployment-wide fleet slots. Concrete path: home fenced, then `ConversationServer.terminate_conversation/2` → `call_server/2` → the server crashes rather than times out (`call_server/2` catches `:timeout`/`:noproc`/`:normal`/`:shutdown`, not an abnormal exit), the exit propagates out of the `Enum.each` at `conversations.ex:3423-3427`, and `_unsafe_retire_home/1` never runs. A node going down between the fence commit and the retire does the same. This is the same accepted state `reset_sandbox/2`'s docstring spells out at length ("A provider error or lost caller leaves the fence and capacity in place … Reaping is the supported way out … There is no automatic reconciliation"), and I verified the escape hatch still works through this new door: `fence_home_destruction/1` returns the row unchanged when `reset_requested_at` is already set, so `/admin/sandboxes` reaping and a retried `delete_agent` both still complete. Two asks, neither blocking: say this on `_unsafe_destroy_home/1` the way `reset_sandbox/2` says it, and note that this door fences statuses `reset_sandbox/2` deliberately refuses (`pending`, `starting`), which is a slightly wider class of stranded row than the reset path can produce.

2. *Non-blocking — the new error return has no handler at either door.* `Agents.delete_agent/2` can now answer `{:error, :provider_transaction_open}` or `{:error, :not_found}` (the home row vanished between `Repo.all` and the fence), and both call sites match `{:ok, _} = Agents.delete_agent(...)` (`agent_controller.ex:181`, `agents_live/index.ex:83`). So the propagation the body advertises surfaces as a `MatchError` → 500, not as a message. The old code crashed on the same races too (`{:ok, _} = update_sandbox(...)` inside `_unsafe_retire_home/1`), so this is not a regression, and both reachable reasons are either a programming error or a rare race — but it does mean "propagates fencing errors" currently means "500s". Worth a `case` at the two doors, in this PR or a follow-up.

3. *Non-blocking — partial teardown on halt.* `Enum.reduce_while/3` (`conversations.ex:3406-3411`) halts on the first error, so an agent with several homes can end up with some destroyed, some untouched, and its row still present. Retry-safe (retired homes are filtered out by the `status not in` clause on the next attempt), and preferable to pressing on, but it is worth one sentence in the `@doc` that the count is "homes torn down before the first failure".
4. *Non-blocking, reuse.* `Conversations` already has `require_provider_commit_boundary/0` (`conversations.ex:3830-3834`) with this exact body, a comment explaining why it exists, and three `with :ok <- …` call sites (`2914`, `3707`, `4523`). This patch inlines the check twice more (`3397`, `3417`), bringing the module to four inline copies including `reset_sandbox/2`'s. Reusing the helper would read the same and keep the reasoning in one place.
5. *Non-blocking, and #1973's job.* `fence_home_destruction/1` mutates tenant-owned state with no audit row, where `reset_sandbox/2` records `sandbox.reset_requested` for the equivalent commit. #1973 ("audit committed teardown admission fences") is the stage that supplies it, so this is noted, not requested. Also noted once for the batch: no `CHANGELOG.md` entry, and agent deletion failing where it previously always proceeded is the most user-visible change in this group.

**Checked and clean**

- **Lock order.** `fence_home_destruction/1` takes `pg_advisory_xact_lock(@sandbox_lock_namespace, phash2(sandbox.id))` then the row `FOR UPDATE` and touches nothing else — the same prefix as `do_reset_sandbox/2` and `SandboxIdentity.bind/2`, and it can never invert against admission (advisory → conversation → sandbox), because the advisory lock serialises the two before either reaches a row. The `Repo.in_transaction?()` refusal is load-bearing beyond its stated purpose: it makes it *impossible* for this fence to be taken while a caller holds `Quotas.with_sandbox_reservation/3`'s global fleet lock, so the global-before-per-user order in `quotas.ex:241-246` is untouched and `:fleet_full` (503) versus `Credits.gate/1` (402) stay exactly where they were.
- **No slow work inside the lock.** The fence transaction is one advisory lock, one select, one update. Provider destroy and the `terminate_conversation` fan-out are both outside it, and the test asserts `refute Repo.in_transaction?()` from inside the mocked `Sprites.destroy/1` and from inside the mocked `terminate_conversation/2` — so ordering *and* the absence of an enclosing transaction are both pinned by assertion rather than by reading.
- **The fence is not cleared on an error path — correctly.** Nothing in the codebase sets `reset_requested_at` back to `nil`; a fence ends by the row going terminal, and `update_sandbox/2` deliberately lets terminal writes through (`conversations.ex:261-272`), which is why `_unsafe_retire_home/1`'s `{:ok, _} =` still holds on a fenced row. Idempotent re-entry verified (already-fenced and already-terminal rows both fall to the `current` branch without a second stamp).
- **Quota retention is asserted, not assumed**: `Quotas.active_sandbox_count/1 == 1` inside the destroy callback and `== 0` after, matching the body's claim that the fenced row keeps its slot until teardown finishes.
- **Admission refusal is proved through the real function** at both capacities (`1` and `:unbounded`), so the refusal comes from the fence predicate and not from the capacity check.
- **The enclosing-transaction tests prove no side effects**, using `reject/3` on both `Sprites.destroy` and `terminate_conversation` and then asserting the agent row, the fence, the sandbox status and the conversation status are all untouched — including through `Agents.delete_agent/2`, which pins the propagation.
- `@billable_terminal` is defined (`conversations.ex:242`) well before its use at `3446`; `count` in the new `with` is used by its own guard, so no unused-variable warning. No audit call inside a transaction was added. `_unsafe_destroy_home/1` has no production caller other than `_unsafe_destroy_homes_for_agent/1`, and account deletion reaches sandboxes through the FK cascade rather than either function, so nothing else inherits the new transaction refusal.

PR #1973, opened 2026-09-11
feat(sandboxes): audit committed teardown admission fences

jhgaylor:
Forced home teardown now records its durable admission fence before actor or provider cleanup. The shared context API commits under the existing sandbox locks and records `sandbox.teardown_requested` afterward, so interrupted cleanup has a request trail. Agent deletion forwards actor and request IP to that event.

Repeated calls preserve an existing fence without duplicate events; terminal rows need no new fence. Missing rows and enclosing transactions return errors before audit or provider work. The API performs no provider I/O and preserves the existing quota and forced-retirement behavior. Other retirement callers can adopt it in separate small PRs.

Stack: based on #1972. Refs #1767; lifecycle and account cleanup adoption and the remaining writer audit are separate work.

Validation:
- 150 focused teardown, agent and audit-guardrail tests passed. New cases cover the post-commit audit boundary, caller attribution, stale retries, existing reset fences, terminal/missing rows and enclosing transactions.
- The first focused run used an unsupported audit-query filter in the new tests; correcting it to the documented `action_prefix` filter resolved those five failures.
- Full `mix precommit` passed: 5,407 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

Auditing uses the existing best-effort post-commit convention; it is not an atomic outbox guarantee. No deployed environment or real provider was modified.

### DIFF
--- a/apps/fountain/lib/fountain/agents.ex
+++ b/apps/fountain/lib/fountain/agents.ex
@@ -277,7 +277,7 @@ defmodule Fountain.Agents do
     # names it — deletion would nilify the pointer and orphan the sprite.
     # Ownership: `agent` came from the caller's scoped fetch.
     with count when is_integer(count) <-
-           Fountain.Conversations._unsafe_destroy_homes_for_agent(agent.id) do
+           Fountain.Conversations._unsafe_destroy_homes_for_agent(agent.id, opts) do
       delete_agent_row(agent, opts)
     end
   end
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -3690,7 +3690,7 @@ defmodule Fountain.Conversations do
   and provider I/O; already admitted turns may be interrupted by this forced
   operation. `_unsafe_`: the caller owns the agent.
   """
-  def _unsafe_destroy_homes_for_agent(agent_id) when is_binary(agent_id) do
+  def _unsafe_destroy_homes_for_agent(agent_id, opts \\ []) when is_binary(agent_id) do
     if Repo.in_transaction?() do
       {:error, :provider_transaction_open}
     else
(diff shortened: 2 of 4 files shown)


BinaryBourbon (approved):
## APPROVE

Read the patch and verified both new guards by breaking them.

**What changed.** `fence_home_destruction/1` becomes the public, documented `_unsafe_fence_sandbox_for_teardown/2`, returns `{fenced, first_fence?}` so the audit row is written once, and `delete_agent/2`'s attribution now reaches it through `_unsafe_destroy_homes_for_agent/2`.

**Audit rules, all three respected.** `Audit.record/1` is called *after* `Repo.transaction/1` returns, never inside it — and the test enforces that with `refute Repo.in_transaction?()` inside the Mimic callback rather than trusting the reading. The metadata is `reason` plus `provider`; no values. And it records only what happened: a terminal sandbox or an existing reset fence produces no row, which is asserted with `reject(Audit, :record, 1)`.

**Reversion check 1 — the guardrail.** Stubbing out the `Audit.record/1` call:

```
1) test sandbox teardown fence leaves an audit event (Fountain.AuditGuardrailTest)
   sandbox teardown fence produced no sandbox.teardown_requested event.
```

The new guardrail row bites.

**Reversion check 2 — idempotence.** Changing the fence to re-stamp and re-announce on an already-fenced row fails 4 of the file's tests, including `"an existing reset fence keeps its timestamp without another request event"`. The `{fenced, true|false}` tuple is doing real work.

Full file green as submitted: `15 tests, 0 failures` across this and the two dependent files.

**On reusing `reset_requested_at` as the teardown fence.** The docstring says so plainly (`"Reuses the reset fence so every existing reuse path refuses the machine"`), which is the right call for a stack — every existing `is_nil(reset_requested_at)` check starts refusing the machine for free, and capacity is deliberately retained until retirement finishes. Noting for the record that a later sibling (`0d3fd544`, `persist forced teardown intent separately`) splits this into its own column plus a migration; nothing here has to change first, and the semantics stay honest in the meantime.

`_unsafe_` prefix is correct (no tenant scoping; the caller owns the row), the new `opts` default keeps the one existing `_unsafe_destroy_home/1` test call site working, and `{Conversations, :_unsafe_fence_sandbox_for_teardown, 2}` is correctly added to the "must accept attribution opts" list rather than to the exclusion list.

PR #1974, opened 2026-09-11
test(conversations): prove reassignment and admission commit order

jhgaylor:
Admission already locks the conversation row, but the reassignment side of #1767 had no concurrent regression. Eight PostgreSQL tests now prove that both context updates and the bulk update shape used for co-tenants serialize with admission, for bounded and unbounded capacity and both commit orderings.

If reassignment commits first, admission against the old sandbox writes no turn. If admission wins, reassignment waits through the insert and then moves the conversation. The docstring states that a later forced reassignment can still move an already-admitted conversation; these locks do not add a running-turn check to arbitrary writers.

Stack: based on #1969. Refs #1767. This covers database attachment ordering; actor notifications and provider cleanup remain separate lifecycle responsibilities.

Validation:
- All eight cases pass using distinct PostgreSQL connections, telemetry pauses after actual row queries, and `pg_blocking_pids` evidence of the waiter/holder relationship.
- Weakening the conversation locks makes all eight fail at the expected blocking assertion. Restoring the locks passes all 27 reassignment, retirement, reset-order and saved-allowance tests.
- Tests verify persisted turn and sandbox state and absence of provider deletion. Committed fixtures are explicitly cleaned up; database counts for users, sandboxes, conversations and turns were all zero after the mutation run.
- Full `mix precommit` passed: 5,391 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1668,8 +1668,12 @@ defmodule Fountain.Conversations do
 
   `capacity` is `Managoat.Runtimes.ACP.concurrency/1`. All runtimes take the
   per-sandbox advisory lock and verify that the conversation still belongs
-  to this nonterminal sandbox. An integer capacity also limits concurrent
-  turns; `:unbounded` skips only that capacity check. Saved execution allowances
+  to this nonterminal sandbox. The conversation row stays locked through the
+  insert: an earlier reassignment refuses this sandbox, while a later forced
+  reassignment can move a conversation whose turn was already admitted. This
+  ordering does not make arbitrary reassignment writers check for running turns.
+  An integer capacity also limits concurrent turns; `:unbounded` skips only that
+  capacity check. Saved execution allowances
   are checked under row locks; no runtime control is supported yet, so any
   nonempty allowance refuses the turn. Refusal writes no turn.
   Usage is recorded after the transaction commits, never inside it.
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1705,6 +1705,9 @@ defmodule Fountain.Conversations do
           Repo.rollback(:configuration_changed)
         end
 
+        # Row-only retirement writers do not take the admission advisory
+        # lock. Hold their row through insertion too: a committed retirement
+        # refuses admission, while a later forced retirement follows the turn.
         attached? =
           Repo.exists?(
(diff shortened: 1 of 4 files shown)


BinaryBourbon (approved):
## APPROVE

Verified non-vacuous by reverting the guard. This one took two attempts to break, and the reason is worth recording.

**What it proves.** That admission holds the *conversation* row from the attachment check through the turn insert, so a reassignment committed earlier refuses the turn (`:sandbox_unavailable`) and one arriving later waits for the insert. The test pauses admission after the `INNER JOIN "sandboxes"` attachment query — which is after both conversation locks are taken — and proves the wait with `pg_blocking_pids`, in both directions, for both `update_conversation/2` and a raw `Repo.update_all`.

**Reversion check.** Removing `lock: "FOR UPDATE"` from the first conversation read alone leaves all 8 green. That is not vacuity: the join's `lock: "FOR SHARE"` on the attachment query locks rows from *both* tables in the `FROM` list, so the conversation row is still held. The two locks overlap on that row (`FOR UPDATE` is there for the allowance FK, `FOR SHARE` also covers the sandbox row for the retirement race). Remove **both** and the test collapses:

```
8 tests, 8 failures — every one on "no wait on the winning PostgreSQL connection observed"
```

So the property is real and the test measures it. Worth knowing when someone later tries to simplify one of those two locks away.

**Lock order.** `advisory -> conversation FOR UPDATE -> (conversation, sandbox) FOR SHARE` matches `fail_initial_start/2` and the teardown fence in #1978, and `do_reset_sandbox/2`'s opposite order (`advisory -> sandbox -> conversations`) cannot cycle with it, because both take the same per-sandbox advisory lock first and a conversation reaches only its own `sandbox_id`. No deadlock introduced.

**Docstring.** The added paragraph is honest about what the ordering does *not* buy — `"This ordering does not make arbitrary reassignment writers check for running turns"` — and the `:admission`-first/`:bulk` case asserts exactly that permissiveness rather than papering over it. Good; a docstring claiming the stronger property is how this stack has gone wrong before.

### Non-blocking nits

Same naming problem as #1968 — `"admission wins context reassignment with 1 capacity"` reads as though admission is the reassignment. And the `:admission` pause matcher `query =~ ~s(INNER JOIN "sandboxes")` is broad; it happens to be unique in this transaction today, and a future second join would make the pause land early and fail loudly rather than silently, so it is fine, just fragile.

PR #1975, opened 2026-09-11
fix(conversations): fence lifecycle reclaim before teardown

jhgaylor:
Lifecycle reclaim could close its adapter and destroy the provider machine while another connection admitted a turn. Reclaim now commits the shared teardown fence before adapter shutdown and provider deletion, retaining the quota slot until retirement finishes. Its request event names `system:conversation_server` and the reclaim reason.

The server prepares the fence before dropping its connection. `Lifecycle.destroy/5` also checks the fence so direct callers obey the same boundary; the repeated check preserves the timestamp and emits no duplicate request event. A refused fence keeps the server available for the next lifecycle check and does not publish a successful reclaim. Enclosing transactions are refused before destruction, including the legacy path without a sandbox row.

Stack: based on #1973. Refs #1767. Already-admitted turns may still be interrupted by forced reclaim. Existing best-effort provider deletion behavior remains in place; account cleanup and park behavior are separate paths.

Validation:
- All six initial regressions fail against the parent: missing fences at provider/adapter boundaries and destruction inside enclosing transactions.
- All 55 focused lifecycle action, server lifetime and shared-sandbox tests passed. Coverage includes bounded/unbounded refusal, quota retention, one attributed request event, missing rows, enclosing transactions, and retry state when a fence check fails after adapter shutdown.
- Full `mix precommit` passed: 5,415 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/agents.ex
+++ b/apps/fountain/lib/fountain/agents.ex
@@ -277,7 +277,7 @@ defmodule Fountain.Agents do
     # names it — deletion would nilify the pointer and orphan the sprite.
     # Ownership: `agent` came from the caller's scoped fetch.
     with count when is_integer(count) <-
-           Fountain.Conversations._unsafe_destroy_homes_for_agent(agent.id) do
+           Fountain.Conversations._unsafe_destroy_homes_for_agent(agent.id, opts) do
       delete_agent_row(agent, opts)
     end
   end
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -3690,7 +3690,7 @@ defmodule Fountain.Conversations do
   and provider I/O; already admitted turns may be interrupted by this forced
   operation. `_unsafe_`: the caller owns the agent.
   """
-  def _unsafe_destroy_homes_for_agent(agent_id) when is_binary(agent_id) do
+  def _unsafe_destroy_homes_for_agent(agent_id, opts \\ []) when is_binary(agent_id) do
     if Repo.in_transaction?() do
       {:error, :provider_transaction_open}
     else
@@ -3701,7 +3701,7 @@ defmodule Fountain.Conversations do
       )
       |> Repo.all()
       |> Enum.reduce_while(0, fn home, count ->
-        case _unsafe_destroy_home(home) do
+        case _unsafe_destroy_home(home, Keyword.put_new(opts, :reason, "agent_deleted")) do
           :ok -> {:cont, count + 1}
           {:error, _} = error -> {:halt, error}
         end
@@ -3710,11 +3710,11 @@ defmodule Fountain.Conversations do
   end
 
   @doc false
-  def _unsafe_destroy_home(%Sandbox{} = sandbox) do
+  def _unsafe_destroy_home(%Sandbox{} = sandbox, opts \\ []) do
     if Repo.in_transaction?() do
       {:error, :provider_transaction_open}
     else
-      with {:ok, fenced} <- fence_home_destruction(sandbox) do
+      with {:ok, fenced} <- _unsafe_fence_sandbox_for_teardown(sandbox, opts) do
         fenced = Repo.preload(fenced, :conversations)
 
         fenced.conversations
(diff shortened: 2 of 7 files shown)


BinaryBourbon (approved):
APPROVE — the fence covers the window it claims, on both the server path and direct callers, and the tests would fail if it were reverted. Two things to answer before merge, plus a size-pin rebase trap that will turn this red on current `main`.

**What this changes**

`ConversationServer.destroy_sandbox/2` no longer closes its adapter and calls `Lifecycle.destroy/5` unconditionally. It first commits `Lifecycle.prepare_destroy/2`, which reuses the base PR's `Conversations._unsafe_fence_sandbox_for_teardown/2` to stamp `reset_requested_at` under the per-sandbox advisory lock, so no new turn can be admitted on the machine between the adapter shutdown and the provider destroy. `Lifecycle.destroy/5` re-runs the same preparation for direct callers (idempotent: an existing fence records no second `sandbox.teardown_requested`), and a refusal on either side returns `{:noreply, state}` instead of stopping, keeping the handle for the next tick.

**Findings**

1. **Non-blocking, but answer it: a fence committed on the automatic reclaim path is invisible to both `SandboxReaper` sweeps.** `sandbox_reaper.ex:125` (`release_stuck_sandboxes`) and `:228` (`sweep_abandoned_sandboxes`) both filter `is_nil(s.reset_requested_at)`, while `quotas.ex:145` deliberately keeps counting a fenced non-terminal row. In `lifecycle.ex:472-486` the fence is committed, then `Managoat.Sandbox.destroy/1` (473) and `Egress.release/2` (474) run *before* the row is terminated (482). If either raises — not returns `{:error, _}`, which 473 swallows, but raises or exits — the GenServer dies with the row still `ready` and fenced: the tenant permanently loses a slot from `Quotas.sandbox_limit/1` (floor 2, so two of these lock the account out of launching anything) and no sweep will ever reclaim it. On `main` the same crash left an *unfenced* `ready` row, which `sweep_abandoned_sandboxes` reclaimed after its grace window. Before this PR the fence was only ever set by an explicit user reset; this PR puts it on the once-a-minute reclaim tick for every sandbox in the fleet, so the exposure is qualitatively different. This is the unreconciled-fence gap round 1 filed as #2021 item 5 — please say in the PR body whether anything is expected to sweep it, or add the fenced-and-stale case to one of the two passes.

2. **Non-blocking: `reclaim_refused/2` retries forever and, on a persistent refusal, never destroys the sprite.** `conversation_server.ex:2143` returns `{:noreply, state}`, and the tick is re-armed at the top of the handler (`:2037`), so a refusal repeats every `lifecycle_check_ms` (default 1 minute) with a `Logger.warning` and no termination. The only reachable refusal is `{:error, :not_found}` from `prepare_destroy/2` when the sandbox row has gone; the new test `"a missing sandbox refuses preparation without touching the adapter"` locks that in with `reject(Managoat.Sandbox, :destroy, 1)`. On `main` the missing-row case destroyed the sprite first and *then* crashed on `_unsafe_get_conversation!`, so the machine died; now the machine is retained by design. I could not find anything in `apps/fountain/lib` that hard-deletes a `sandboxes` row (the user cascade nilifies `user_id`, it does not delete), so I am calling this defensive rather than live — but the loop has no bound and the sprite is exactly the thing that costs money. Consider stopping after N refusals, or falling through to the old destroy-the-handle behaviour when the row is gone.

3. **Non-blocking, will block the merge: the `conversation_server.ex` size pin.** This branch is based on a merge-base where `@pin` was 2762 and the file was 2685; the base PR takes it to 2712 and this one to 2722, so CI is green. Current `origin/main` has shrunk the file to 2646 and tightened `@pin` to exactly 2646. Neither branch touches `conversation_server_size_test.exs`, so a rebase silently takes main's 2646 against a 2683-line file and `Fountain.Conversations.ConversationServerSizeTest` fails — an identical patch plus a stale absolute pin. Raise the pin to the file's actual length in the rebase commit, per the pin's own rule that it records the real count (and squeeze out the 40 lines of slack the base PR left while you are there).

4. Nit: the success path now runs `prepare_destroy/2` twice — once at `conversation_server.ex:2124`, once inside `destroy/5` at `lifecycle.ex:466` — so every reclaim pays two extra queries and a second advisory-lock transaction. Deliberate and documented in the comment; noting it only because the comment says "for direct callers" and the server is not one.

**Checked and clean**

- The fence actually blocks admission: `_unsafe_create_turn_on_sandbox/3` filters `is_nil(s.reset_requested_at)` (`conversations.ex:1715`) and `check_attachable/4` returns `:sandbox_reset_pending` (`:3902`), and the new test asserts `{:error, :sandbox_unavailable}` *inside* the `Managoat.Sandbox.destroy/1` callback for both `capacity: 1` and `:unbounded` — it asserts on the window, not on a stub of it.
- Terminating a fenced row still works: `update_sandbox/2` rolls back `:sandbox_reset_pending` only when the target status is not in `@billable_terminal` (`conversations.ex:268-272`), so `do_destroy/5`'s `{:ok, _} = ` match at `lifecycle.ex:481` cannot become a `MatchError` because of the fence this PR adds. That was my first worry and it is correctly handled upstream.
- Variable scoping in the rewritten `destroy_sandbox/2`: the `state` rebound by `drop_connection/2` inside the `with` body is the one passed to `reclaim_refused/2` from the inner `case`, so a post-adapter refusal reports `current_command: nil` and keeps `handle` — asserted by `"a refusal after adapter shutdown keeps the fenced machine for retry"`.
- The `:noop`-equivalent path adds no abandoned `running` turn row: the fence is taken before any turn state changes and admission is refused rather than half-recorded.
- No enclosing-transaction hazard: `prepare_destroy/2` refuses `Repo.in_transaction?()` before touching the row, so the advisory lock and the `Audit.record/1` that follows the commit both stay outside a caller transaction — consistent with the never-audit-in-a-transaction rule.
- `_unsafe_get_sandbox/1` at `lifecycle.ex:439` carries the ownership comment by reference to `home?/1`; the server established ownership at `init/1`.
- `async: false` on the new test file is justified and explained (it writes application-wide lifecycle bounds) and restores the previous values in `on_exit`.
- Reverting the fence fails the suite: three of the seven new cases assert the refusal or the committed `reset_requested_at` from inside a provider callback.
- No `CHANGELOG.md` entry — consistent with the rest of the programme, noting once, not as a blocker.

PR #1976, opened 2026-09-11
fix(accounts): refuse teardown inside enclosing transactions

jhgaylor:
Account deletion and tenant compute cleanup could stop actors and delete provider machines inside a caller's database transaction. They now return {:error, :provider_transaction_open} before teardown when an enclosing transaction is open. The guard covers account deletion, including its owned principals, and both user/id forms of destroy_sprites/1.

Calls outside a transaction keep their existing deletion and best-effort provider behavior. The documented return contract now includes refusal. This is the transaction-boundary prerequisite for committing admission fences before account cleanup in a separate small PR.

Stack: based on #1973. Refs #1767; admission-fence adoption is separate.

Validation:
- All three new regressions fail on the parent. Provider and actor calls are rejected by the local test stubs.
- All 68 focused transaction-boundary, account deletion, principal and console deletion tests passed. Refusal preserves user, agent, sandbox, conversation and audit state.
- Full `mix precommit` passed: 5,410 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/accounts/deletion.ex
+++ b/apps/fountain/lib/fountain/accounts/deletion.ex
@@ -71,9 +71,10 @@ defmodule Fountain.Accounts.Deletion do
       an admin path should pass something identifying.
     * `:request_ip` — passed through to the audit event.
 
-  Returns `{:ok, summary}`. Nothing aborts before destruction: there is no
-  subscription to cancel first (ADR 0031), and the Stripe customer stays for
-  the refund trail.
+  Returns `{:ok, summary}` or an error. An enclosing database transaction is
+  refused before any teardown: actor and provider calls must run outside its
+  locks. There is no subscription to cancel first (ADR 0031), and the Stripe
+  customer stays for the refund trail.
 
   A claimable principal this account owns (ADR 0044) goes with it. The
   ownership row cascades either way, so the alternative is not "keep it" but
@@ -83,6 +84,14 @@ defmodule Fountain.Accounts.Deletion do
   """
   @spec delete_user(User.t(), keyword()) :: {:ok, map()} | {:error, term()}
   def delete_user(%User{} = user, opts \\ []) do
+    if Repo.in_transaction?() do
+      {:error, :provider_transaction_open}
+    else
+      do_delete_user(user, opts)
+    end
+  end
+
+  defp do_delete_user(user, opts) do
     delete_owned_principals(user, opts)
     sprites = destroy_sprites(user)
 
@@ -144,7 +153,8 @@ defmodule Fountain.Accounts.Deletion do
 
   @doc """
   Stop every sandbox a tenant is running, and return how many sprites were
-  destroyed.
+  destroyed. Refuses an enclosing database transaction before stopping actors
+  or calling a provider.
 
   Ask a live ConversationServer to tear itself down where one exists, so the
   sprite goes through the same path as a user-initiated terminate. Otherwise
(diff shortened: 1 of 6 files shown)


BinaryBourbon (approved):
APPROVE — a real runtime guard at both public entry points, not a docstring, and it fires before anything irreversible. One soft finding about the new error value being discarded by the only non-deletion caller until #1977 lands.

**What this changes**

`Deletion.delete_user/2` and `Deletion.destroy_sprites/1` now check `Repo.in_transaction?()` and return `{:error, :provider_transaction_open}` before delegating to the extracted `do_delete_user/2` / `do_destroy_sprites/1`. The refusal is placed ahead of `delete_owned_principals/2`, the `ConversationServer.terminate_conversation/2` calls and every provider call, so a caller that wrapped deletion in a transaction gets nothing done rather than half of it.

**On the CLAUDE.md rule this resembles: it is a different hazard.**

The audit rule is about `Audit.record/1` being best-effort *by rescuing*: a rescue does not survive a transaction, so a failed audit insert aborts the enclosing one and takes the mutation with it. That is a rollback-contagion hazard. What this PR refuses is a liveness hazard on the other side: `do_destroy_sprites/1` makes a synchronous `GenServer` call into each live `ConversationServer` and then a provider HTTP call per sandbox, and `_unsafe_fence_sandbox_for_teardown/2` (the base PR) takes `pg_advisory_xact_lock` plus a `FOR UPDATE` row lock. Doing that inside a caller's transaction holds Postgres locks for the duration of a provider timeout and gives the actor shutdown a connection it may block on. Same remedy shape, different failure. Both are relevant here and both are covered, because the refusal is checked at runtime — `Repo.in_transaction?()` at `deletion.ex:87` and `:172` — and asserted by the new test for all three entry shapes, not merely written down.

**Findings**

1. **Non-blocking, soft, stack hygiene.** `destroy_sprites/1` changes from returning `non_neg_integer()` to `non_neg_integer() | {:error, term()}`, and its only caller outside this module discards the return: `Principals.stop_compute/1` is `Deletion.destroy_sprites(user_id); :ok` (`principals.ex:816` on this head). As merged on its own base, a refusal there would silently mean "this principal's compute was not stopped" with no log line and no error — and the `rescue` clause below it does not catch a returned tuple. In practice `stop_compute/1` is called from `close/5` *after* its `Repo.transaction` has committed (`principals.ex:674`), so the guard cannot fire on that path today, and #1977 adds the `{:error, reason} -> Logger.warning(...)` branch that closes it properly. Flagging it because if #1976 merges and #1977 slips, the window is real; not a reason to hold this PR.
2. **Non-blocking, doc.** `delete_user/2`'s docstring now says "Returns `{:ok, summary}` or an error", but the moduledoc's `## Order matters` step 1 still says sprite-destruction "Failures here are logged rather than fatal: `SandboxReaper` reconciles anything missed on its next run." That statement is still true of *this* PR (the refusal happens before any teardown, so there is no partial state to reconcile). It stops being true in #1977, where a fence error aborts the whole deletion — I have put the finding there.
3. Nit: the guard is duplicated verbatim in two functions and `delete_user/2` reaches `destroy_sprites/1`, which checks again. Harmless, and defensible as defence in depth for a public function; mentioning only so nobody "simplifies" the inner one away later.

**Checked and clean**

- Every caller of both functions reached from a non-transaction context, so the refusal cannot break a legitimate call: `account_data_controller.ex:179`, `account_live.ex:82`, `admin_controller.ex:520`, `admin_live/users.ex:263`, `workers/unverified_account_pruner.ex:88`, `principals.ex:810` (`purge_one/1`, outside `close/5`'s transaction) and `principals.ex:817`. None of them wraps the call in `Repo.transaction`.
- The recursive call in `delete_owned_principals/2` (`deletion.ex:150`) re-enters the public `delete_user/2` and so re-checks the guard — correct, because `do_delete_user/2` opens no transaction of its own, so the nested call is not refused. A refusal firing on that legitimate nested path would have broken account deletion for any account owning a claimable principal; it does not.
- The test asserts the refusal is *before* teardown, not merely that an error came back: `reject(ConversationServer, :terminate_conversation, 2)`, `reject(Managoat.Sandbox.Sprites, :destroy, 1)`, the user/agent/sandbox/conversation rows all unchanged, and `Audit.list_for_user(user.id) == events` — so no audit row either. It covers all three arities/shapes via the `for` loop, and it fails if the guard is reverted.
- `async: true` with no global state; no `pool_size` or config touched.
- The `@spec` widening matches the new return on both clauses, and `destroy_sprites(%User{})` delegates to the guarded binary clause rather than bypassing it.
- No file overlap with its sibling #1975 (`conversation_server.ex` / `lifecycle.ex` there, `accounts/deletion.ex` here), so the two do not conflict textually — they only need rebasing onto whichever lands first.
- No `CHANGELOG.md` entry, consistent with the rest of the programme; noted once, not a blocker.

PR #1977, opened 2026-09-11
fix(accounts): fence admission before tenant compute teardown

jhgaylor:
Account and principal cleanup could stop actors and delete machines while new turns were still admitted. Cleanup now commits admission fences for the known machines before actor shutdown, then re-reads the remaining rows and fences each before provider deletion. Account deletion propagates fencing errors and retains the account instead of reporting a completed deletion.

Teardown requests preserve caller actor/request IP and distinguish account deletion from principal closure. Repeated fences produce no duplicate request event. The existing provider count remains the number of confirmed direct deletions: actor-retired rows are skipped, and provider failures still retire their rows for the existing reaper to reconcile. Principal closure keeps its existing post-commit best-effort cleanup and now logs fencing refusals.

Stack: based on #1976. Refs #1767. Already-admitted turns may be interrupted by forced cleanup. This does not introduce an atomic account-wide barrier against concurrently creating new resources.

Validation:
- All seven initial regressions fail on the parent.
- All 149 focused account, principal and audit-guardrail tests passed. New coverage includes bounded/unbounded admission, actor/provider ordering, newly discovered machines, initial/late fence refusal, attribution, quota retention and provider-count behavior.
- Moving the initial fence after teardown makes the two actor-boundary regressions fail. Restoring the implementation passes all nine new cases.
- Full `mix precommit` passed: 5,420 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/accounts/deletion.ex
+++ b/apps/fountain/lib/fountain/accounts/deletion.ex
@@ -93,8 +93,14 @@ defmodule Fountain.Accounts.Deletion do
 
   defp do_delete_user(user, opts) do
     delete_owned_principals(user, opts)
-    sprites = destroy_sprites(user)
 
+    with sprites when is_integer(sprites) <-
+           destroy_sprites(user, Keyword.put_new(opts, :reason, "account_deleted")) do
+      delete_user_row(user, sprites, opts)
+    end
+  end
+
+  defp delete_user_row(user, sprites, opts) do
     Audit.record(%{
       user_id: user.id,
       action: "account.deleted",
@@ -154,7 +160,10 @@ defmodule Fountain.Accounts.Deletion do
   @doc """
   Stop every sandbox a tenant is running, and return how many sprites were
   destroyed. Refuses an enclosing database transaction before stopping actors
-  or calling a provider.
+  or calling a provider. Known machines are admission-fenced before actor
+  shutdown; machines found afterward are fenced before provider deletion.
+  Forced cleanup may interrupt already-admitted turns. Options carry actor,
+  request_ip and a reason for the committed teardown request.
 
   Ask a live ConversationServer to tear itself down where one exists, so the
   sprite goes through the same path as a user-initiated terminate. Otherwise
@@ -165,18 +174,35 @@ defmodule Fountain.Accounts.Deletion do
   keeps the rows. Duplicating this would be duplicating the part that costs
   money when it is wrong.
   """
-  @spec destroy_sprites(User.t() | binary()) :: non_neg_integer() | {:error, term()}
-  def destroy_sprites(%User{id: user_id}), do: destroy_sprites(user_id)
+  @spec destroy_sprites(User.t() | binary(), keyword()) :: non_neg_integer() | {:error, term()}
+  def destroy_sprites(user, opts \\ [])
+  def destroy_sprites(%User{id: user_id}, opts), do: destroy_sprites(user_id, opts)
 
-  def destroy_sprites(user_id) when is_binary(user_id) do
+  def destroy_sprites(user_id, opts) when is_binary(user_id) do
     if Repo.in_transaction?() do
       {:error, :provider_transaction_open}
     else
-      do_destroy_sprites(user_id)
+      opts = Keyword.put_new(opts, :reason, "compute_stopped")
+
+      with :ok <- fence_sprites(user_id, opts) do
+        do_destroy_sprites(user_id, opts)
+      end
     end
   end
 
-  defp do_destroy_sprites(user_id) do
+  defp fence_sprites(user_id, opts) do
+    # ownership: live_sandboxes/1 scopes every row to this caller-owned user_id.
+    user_id
+    |> live_sandboxes()
+    |> Enum.reduce_while(:ok, fn sandbox, :ok ->
+      case Conversations._unsafe_fence_sandbox_for_teardown(sandbox, opts) do
+        {:ok, _} -> {:cont, :ok}
+        {:error, _} = error -> {:halt, error}
+      end
+    end)
+  end
+
+  defp do_destroy_sprites(user_id, opts) do
     conv_ids =
       Conversations.list_conversations(user_id)
       |> Enum.filter(&(ConversationServer.whereis(&1.id) != nil))
@@ -194,10 +220,29 @@ defmodule Fountain.Accounts.Deletion do
       end
     end)
 
+    # Actors may have retired a row or created another machine while stopping.
+    # Re-read and fence each remaining row before touching its provider.
+    # ownership: live_sandboxes/1 scopes these fresh rows to the same user_id.
+    user_id
+    |> live_sandboxes()
+    |> Enum.reduce_while(0, fn sandbox, count ->
+      case Conversations._unsafe_fence_sandbox_for_teardown(sandbox, opts) do
+        {:ok, %{status: status} = fenced} when status in @non_terminal ->
+          {:cont, count + if(destroy_sprite(fenced), do: 1, else: 0)}
+
+        {:ok, _retired} ->
+          {:cont, count}
+
+        {:error, _} = error ->
+          {:halt, error}
+      end
+    end)
+  end
+
+  defp live_sandboxes(user_id) do
     Sandbox
     |> where([s], s.user_id == ^user_id and s.status in ^@non_terminal)
     |> Repo.all()
-    |> Enum.count(&destroy_sprite/1)
   end
 
   defp destroy_sprite(%Sandbox{sprite_name: name} = sandbox) when is_binary(name) do
--- a/apps/fountain/lib/fountain/accounts/deletion.ex
+++ b/apps/fountain/lib/fountain/accounts/deletion.ex
@@ -16,7 +16,10 @@ defmodule Fountain.Accounts.Deletion do
      cascade takes `conversations` with it, and after that nothing links a
      sandbox to the user who was paying for it. Failures here are logged rather
      than fatal: `SandboxReaper` reconciles anything missed on its next run,
-     which is exactly the case it exists for.
+     which is exactly the case it exists for. An enclosing transaction is the
+     one thing that refuses, and it refuses before any teardown starts
+     (`delete_user/2`); a single machine's fence being refused mid-run is
+     logged and the run carries on.
 
   2. **Record the audit event.** Before the delete, and carrying the email and
      user id in `metadata` — `audit_events.user_id` is `SET NULL` on delete, so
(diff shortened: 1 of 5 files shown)


BinaryBourbon (changes requested):
REQUEST_CHANGES — the fencing itself is well built and the two-pass structure is the right shape, but it turns sprite teardown into an aborting step in account deletion, which reverses a recorded ADR 0009 decision and contradicts the moduledoc that this very module says to change in the ADR first; and the compensating mechanism that moduledoc names cannot see the rows this patch leaves behind. Separately, every `sandbox.teardown_requested` event it writes on the deletion path becomes an anonymous row seconds later.

**What this changes**

`destroy_sprites/2` now runs in two fenced passes: `fence_sprites/2` commits `Conversations._unsafe_fence_sandbox_for_teardown/2` on every live row of the tenant *before* the first `ConversationServer.terminate_conversation/2`, and then, after actor shutdown, re-reads the tenant's live rows and fences each again before its provider delete — so a machine created or revived while actors were stopping is fenced too, and a row an actor already retired is skipped rather than destroyed twice. `Principals.stop_compute/2` and `do_delete_user/2` forward actor, request_ip and a per-caller `reason`, and `do_delete_user/2` now only proceeds to the audit event and the row delete when `destroy_sprites/2` returns an integer.

**Findings**

1. **Blocking. Sprite teardown becomes an aborting step, against ADR 0009, and the safety net the moduledoc names does not cover the state it leaves.**
   `deletion.ex:97` — `with sprites when is_integer(sprites) <- destroy_sprites(user, ...)` — means any fence error skips `delete_user_row/3` entirely: no `account.deleted` event, no `Repo.delete(user)`. ADR 0009's Decision says the opposite in two places: "**Stripe cancellation first, and it is the only aborting step**" and "2. Sprite destruction, best-effort (`SandboxReaper` reconciles stragglers)". The moduledoc at `deletion.ex:14-18` still says "Failures here are logged rather than fatal: `SandboxReaper` reconciles anything missed on its next run, which is exactly the case it exists for", and the moduledoc's own opening tells you the rule: "The decisions here — ordering, ... — are recorded in ADR 0009. **Change them there first.**" Neither the ADR nor that paragraph is touched by this patch, so the module now documents behaviour it does not have — the same class of defect the 2026-07 audit was about, read in the other direction.
   It matters beyond the paperwork, because the named reconciler is blind to the rows this creates. Concrete scenario: a tenant has sandboxes A and B; `fence_sprites/2` commits A's fence (`reset_requested_at` plus a `sandbox.teardown_requested` row) and then halts on B; `destroy_sprites/2` returns the error and deletion aborts. A is left `status: "ready"` with `reset_requested_at` set and its sprite still running, and **neither** `SandboxReaper.release_stuck_sandboxes/0` (`sandbox_reaper.ex:125`) **nor** `sweep_abandoned_sandboxes/0` (`:228`) will ever see it — both filter `is_nil(s.reset_requested_at)` — while `quotas.ex:145` keeps counting it against the tenant's cap. The PR's own test `"refusing a newly discovered machine's fence retains the account"` asserts exactly this half-torn state as intended, with the user row retained and no `account.deleted` event. I could find nothing in `apps/fountain/lib` that hard-deletes a `sandboxes` row, so the only refusal I can actually reach is `{:error, :not_found}` from the fence's locked re-read and I am not claiming this fires in production today — which is why the ask is small: either amend ADR 0009 and the step-1 paragraph to say sprite teardown now aborts and why, or keep the ADR's semantics by logging the fence error and proceeding. Please do not leave the code and the ADR disagreeing.

2. **Blocking, and cheap to fix. The teardown events this writes on the deletion path lose the tenant seconds after they are written.** `audit_events.user_id` and `sandboxes.user_id` are both `SET NULL` on the user delete (moduledoc `deletion.ex:27-30`). `_unsafe_fence_sandbox_for_teardown/2` records `user_id`, `resource_id: sandbox.id` and metadata of `reason` + `provider` only. So after `Repo.delete(user)` at `deletion.ex:119`, every `sandbox.teardown_requested` row this patch caused is an anonymous event pointing at an anonymous sandbox row: the trail says a machine was torn down for `"account_deleted"` and cannot say whose, and `account.deleted`'s metadata (`email`, `user_id`, `sprites_destroyed`) names no sandbox to join back to. This is the exact failure mode the moduledoc's step 2 exists to prevent, and it explains why that event denormalises. One key fixes it — either add the sandbox ids to `account.deleted`'s metadata, or give `_unsafe_fence_sandbox_for_teardown/2` a metadata passthrough and put `"user_id"` in it on this path.

3. **Non-blocking, answering the question the fence raises: a turn admitted a microsecond before the fence is fine; a *machine* created after the second pass is not.** An in-flight turn keeps running and is then interrupted by `terminate_conversation/2`, and it cannot be extended, because every admission path re-reads `reset_requested_at` under the row lock (`conversations.ex:1715`, `:3902`) — so a fenced machine cannot resurrect a partially torn-down tenant, which is what the title claims and what the tests demonstrate. But the fence is per-row: nothing here fences the *tenant*, so a conversation start or wake landing between the second pass and `Repo.delete(user)` mints a fresh, unfenced sandbox whose row is then orphaned by the nilify — the permanent, unfindable leak the `@non_terminal` comment exists to prevent. The second pass narrows that window a great deal versus `main`'s single pass; it does not close it. Closing it needs something at the tenant level that provisioning consults, which is out of scope here — worth adding to #2021 rather than to this PR.

4. **Non-blocking: `admin:<operator_id>` now appears on a non-deletion event.** `admin_controller.ex:521` and `admin_live/users.ex:263` pass `actor: "admin:#{admin.id}"`, and `opts` is forwarded verbatim into the fence, so that string lands on `sandbox.teardown_requested` rows (the new test asserts the forwarding, with `"ui"`). CLAUDE.md's closed vocabulary allows `admin:<operator_id>` for **account deletion only**. Defensible — these events are part of one account deletion — but ADR 0013 owns the vocabulary, so either widen it there or map to plain `"admin"` for the sandbox event.

5. **Non-blocking: `delete_owned_principals/2` discards the new error value.** `deletion.ex:147-152` iterates `delete_user(principal, ...)` in a `for` with no match, and the `rescue` below it only catches raises. So an owner's deletion can now complete while an owned principal is left with fenced, undestroyed machines and its row intact, and nothing logs a plain `{:error, _}`. A `case`/`Logger.warning` here would match what you correctly added to `Principals.stop_compute/2`.

6. **Non-blocking: the partial count is thrown away.** The second pass's `Enum.reduce_while/3` (`deletion.ex:226-239`) returns `{:halt, error}` and discards the accumulator, so sprites it *did* destroy before halting are missing from both the log line and `sprites_destroyed`.

**Checked and clean**

- The fence is genuinely before the actor: `"all known machines are fenced before the first actor is stopped"` captures `reset_requested_at` for **both** sandboxes from inside the `terminate_conversation/2` stub, not afterwards, and asserts `Repo.in_transaction?() == false` there. `"a machine found after actor shutdown is also fenced before provider deletion"` creates the late sandbox from inside that stub and proves the second pass catches it. Reverting either pass fails these.
- Idempotence holds where it must: `"an actor-retired machine is not destroyed or counted again"` asserts exactly one `sandbox.teardown_requested` event and no provider call, and the fence's existing-fence branch returns `{:ok, current}` with a non-terminal status so a retried deletion still destroys a previously fenced row.
- Admission really is refused, at both capacities, asserted from inside the provider callback (`{:error, :sandbox_unavailable}`).
- `"provider failure still retires the fenced row"` confirms the important invariant for finding 1: a row that *reaches* `destroy_sprite/1` is always terminated, provider error or not, so the reaper's terminal-row reconciliation still applies to it. Only rows stranded by a `{:halt, _}` are invisible.
- Tenant scoping: `live_sandboxes/1` scopes by `user_id`, and both `_unsafe_fence_sandbox_for_teardown/2` call sites carry the adjacent ownership comment naming that fetch, as the contract requires.
- Audit convention: the fence commits and then records, outside any transaction; nothing records values; `{Fountain.Accounts.Deletion, :destroy_sprites, 2}` is correctly added to the "every audited context function takes an attribution opts list" list, and the new `do_account_compute_teardown/1` case makes the guardrail fail if this path stops auditing.
- `stop_compute/2` now logs a refusal instead of discarding it, which closes the silent-discard window #1976 opened; it still runs after `close/5`'s transaction commits (`principals.ex:674`), so the `in_transaction?` guard cannot fire there.
- `destroy_sprite/1`'s `is_binary(name)` guard is never the deciding factor: `sprite_name` is `validate_required` on the schema, so `pending`/`starting` rows are fenced *and* terminated like any other. I checked this specifically because a fenced row left non-terminal would be permanently invisible to `release_stuck_sandboxes/0`.
- Does not touch `conversation_server.ex`, so the size-pin rebase trap I raised on #1975 does not apply here.
- No `CHANGELOG.md` entry — consistent with the rest of the programme, noted once, not part of the request.

BinaryBourbon (approved):
**Approve.** Both blocking findings are genuinely fixed, and I checked each by breaking it rather than by reading the description.

### Finding 1 — the deletion path no longer aborts on a fence refusal

Verified in both passes, and verified the guard above them is now unreachable rather than merely unlikely.

- Pass 1: `fence_sprites/2` (`apps/fountain/lib/fountain/accounts/deletion.ex:218-231`) is an `Enum.each/2` — the refusal branch at :228 logs and the pipeline still yields `:ok`. So `with :ok <- fence_sprites(...)` at :202 cannot short-circuit.
- Pass 2: the `Enum.reduce_while/3` at :256-268 emits `{:cont, _}` on every branch including `{:error, reason}` at :264-266. No `{:halt, ...}` remains.
- Therefore the only non-integer `destroy_sprites/2` can return is `{:error, :provider_transaction_open}` from :191, which `delete_user/2` already refuses at :88 before `do_delete_user/2` runs. `delete_owned_principals/2` opens no transaction between the two, so the `with sprites when is_integer(sprites)` guard at :100 is a belt-and-braces structural guard, exactly as claimed — I could not construct a path that escapes it.
- Mutation check: rewriting `fence_sprites/2` back to a halting `reduce_while` makes `delete_user/2` return `{:error, :fixture_refusal}` and fails `deletion_fence_test.exs:84`. The test is not vacuous.

I also confirmed the justification in the comment at :210-217 rather than taking it on trust: `release_stuck_sandboxes/0` (`workers/sandbox_reaper.ex:125`) and `sweep_abandoned_sandboxes/0` (:228) both require `is_nil(s.reset_requested_at)`, and `quotas.ex:145` counts a fenced non-terminal row as active. A halt mid-pass really would strand the rows fenced before the refusal, invisible to every reaper pass and still burning a slot. Best-effort is the correct exit, and it matches ADR 0009 decision 2 (`decisions/0009-account-deletion-and-export.md:38`) without amending it.

Moduledoc vs code: `deletion.ex:18-24` now says failures are logged not fatal, that an enclosing transaction is the one thing that refuses and that it refuses before any teardown starts, and that a mid-run fence refusal is logged and the run carries on. That is what the code does, clause for clause. No new mismatch substituted for the old one.

### Finding 2 — the teardown event still names the tenant

`destroy_sprites/2:200` denormalises `%{"user_id" => user_id}` and `conversations.ex:3722-3730` merges it into the event metadata. `deletion_fence_test.exs:194-211` deletes the user, asserts `event.user_id` is nil (proving the column really is nilified, so the assertion has teeth) and asserts `event.metadata["user_id"]` survives. Removing the `Map.merge` fails that test on `left: nil`.

On the merge safety: caller metadata wins on collision, so in principle a caller could overwrite `reason` or `provider`. I enumerated every caller — `deletion.ex:223` and `:257` are the only ones that pass `:metadata` at all, via a `put_new` of a single `"user_id"` key, and `Principals.stop_compute/2` forwards opts that never carry `:metadata`. Nothing can smuggle anything today. On the audit rule: a user id is an identifier and `account.deleted` already denormalises the same pair (`deletion.ex:117-121`), so "never record values" is not crossed.

Non-blocking: if a future caller does pass `:metadata`, the merge direction lets it silently displace `provider`, which is derived from the row rather than supplied. Merging the other way round, or merging under a nested key, would close that off.

### Findings 4 and 5

- `teardown_actor/1` (`conversations.ex:3744-3745`) collapses `admin:<id>` to `admin` on `sandbox.teardown_requested`, which is what ADR 0013 requires (`decisions/0013-audit-trail.md:123-127` reserves the operator-id form for account deletion). Covered by `deletion_fence_test.exs:213-219`; reverting the collapse fails it on `left: "admin:operator-7"`.
- The comprehension at `deletion.ex:153-160` now logs a failed nested `delete_user/2` instead of discarding it.

Non-blocking on the comment at `conversations.ex:3741-3743`: it cites "an operator reaping a machine from /admin/sandboxes", but the only `admin:<id>` producers in the tree are the two account-deletion call sites (`admin_controller.ex:521`, `admin_live/users.ex:264`) and neither is a sandbox reap. The behaviour is right; the example is of a call site that does not exist yet.

### Tests and gates

`deletion_fence_test.exs` is not weakened. `events/1` uses `Audit.list_for_user/2` only where the user survives; the two post-delete assertions correctly query by action (`:194-211`, `:216-224`) because the column is nil by then — that is a necessary change of method, not a softened assertion. The second-pass refusal test at `:150-192` proves the account is still deleted, the refused machine is left untouched (`refute late.reset_requested_at`, `assert late.status == "ready"`) and the machine that did fence is still destroyed.

`audit_guardrail_test.exs:74-75` adds a positive mutation row for the deletion path, and the `:231-232` addition is to the "must accept attribution opts" list, not to the exclusion list — it tightens the guard rather than exempting anything.

Ran locally against a clean database: 158 tests across `deletion_fence_test.exs`, `deletion_test.exs`, `deletion_transaction_test.exs`, `audit_guardrail_test.exs`, `teardown_fence_test.exs` and `principals_test.exs`, 0 failures. `mix format --check-formatted` and `mix credo --strict` clean. Full CI is green on 6b3307c.

Non-blocking nits, neither worth another round:
- The `with :ok <- fence_sprites(...)` at :202 can no longer fail, so it reads as a live guard that is actually inert. A future change that gives `fence_sprites/2` an error return would silently reinstate the abort this PR removed. A plain call plus a comment would say what is true.
- `fence_sprites/2` has no `rescue`, unlike `destroy_sprite/1` and `delete_owned_principals/2`. A raise from the fence (a DB error inside `Repo.query!`/`Repo.update!`) still aborts the deletion. Low stakes, since the delete would likely fail too.
- The pass-2 log at :265 omits the sandbox id that pass 1 includes at :228.

PR #1978, opened 2026-09-11
feat(conversations): preserve held machines in termination fences

jhgaylor:
Explicit conversation termination must preserve persistent homes and machines another live conversation still holds. The shared teardown fence now accepts a terminating_conversation_id to enforce that rule under the existing admission lock protocol.

The conditional path takes the sandbox advisory lock, verifies and locks the conversation's current owner/attachment, then locks the sandbox before checking the retention rule and writing the fence. Persistent or shared machines return {:error, :sandbox_kept} without mutation or an audit event; stale, missing or foreign conversation bindings return {:error, :sandbox_unavailable}. Calls without this option keep their existing forced-teardown behavior.

Stack: based on #1973. Refs #1767. This is the context prerequisite; conversation actor integration and concurrent attachment coverage are separate small units.

Validation:
- Five of the six new cases fail on the parent; its existing unconditional fence already passes the last-conversation case.
- All 98 conditional fence, teardown fence, forced-home and audit-guardrail tests passed. Tests check stale row input, retained machines, refused bindings, attributed audit events and admission refusal after a successful fence.
- Full `mix precommit` passed: 5,413 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

No actor or provider I/O runs in the fence transaction. No real provider or deployed environment was modified.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -3737,12 +3737,17 @@ defmodule Fountain.Conversations do
   retaining capacity until retirement completes. A new fence records
   `sandbox.teardown_requested` after commit; repeats preserve its timestamp.
   Refuses an enclosing transaction. `opts` carries actor, request_ip and reason.
+
+  With a terminating_conversation_id, first lock and verify that conversation's
+  current attachment and owner. A persistent home or another live conversation
+  returns {:error, :sandbox_kept} without a new fence. The sandbox row stays
+  locked through this decision and the fence, serializing supported attachments.
   """
   def _unsafe_fence_sandbox_for_teardown(%Sandbox{} = sandbox, opts \\ []) do
     if Repo.in_transaction?() do
       {:error, :provider_transaction_open}
     else
-      case do_fence_sandbox_for_teardown(sandbox) do
+      case do_fence_sandbox_for_teardown(sandbox, Keyword.get(opts, :terminating_conversation_id)) do
         {:ok, {fenced, true}} ->
           Audit.record(%{
             user_id: fenced.user_id,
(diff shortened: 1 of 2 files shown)


BinaryBourbon (approved):
## APPROVE

Read the patch and broke both halves of the new guard to confirm the tests hold it.

**What changed.** `terminating_conversation_id` makes the teardown fence conditional: lock and verify that conversation's *current* attachment and owner, then refuse with `:sandbox_kept` if the machine is a persistent home or still carries another live conversation.

**Reversion check 1 — the policy.** Deleting the `:sandbox_kept` rollback fails both policy cases plus the `:attachment`-first case in #1979:

```
8 tests, 3 failures
```

**Reversion check 2 — the ownership predicate.** Relaxing `lock_terminating_conversation/2`'s `where` to `c.id == ^ending_id` (dropping `c.sandbox_id == ^sandbox.id and c.user_id == ^sandbox.user_id`) fails exactly the two tests that exist for it:

```
1) a different tenant's conversation cannot authorize this conditional fence
2) a stale actor cannot fence a conversation that moved to another machine
```

That second one is the important one. A stale actor holding a `%Sandbox{}` struct from before a reassignment cannot fence the machine its conversation has left — and the check is made on the row re-read under the lock, not on the passed-in struct. That is precisely the "a fence that can be silently disabled" failure mode this campaign keeps producing, and it is closed here.

**Lock order is right, and the comment is accurate.** `advisory -> conversation FOR UPDATE -> sandbox FOR UPDATE` now matches admission's `advisory -> conversation FOR UPDATE -> (conversation, sandbox) FOR SHARE` and `fail_initial_start/2`. I checked the one path that runs the opposite order — `do_reset_sandbox/2` goes `advisory -> sandbox -> conversations` — and it cannot cycle with this: both take the same per-sandbox advisory lock first, and `lock_terminating_conversation/2` only ever reaches a conversation whose `sandbox_id` is this sandbox, so the row sets are disjoint across sandboxes.

**Nil handling.** `lock_terminating_conversation(_sandbox, nil)` returns `:ok` and the `sandbox_kept` branch is guarded by `not is_nil(ending_id)`, so the unconditional callers (`_unsafe_destroy_home/2`) keep their old behavior exactly. A nil id widens the fence rather than silently disabling it, which is the safe direction.

`reject(Managoat.Sandbox.Sprites, :destroy, 1)` in `setup` covers every case in the file, so no test can pass by quietly reaching the provider.

### One note, not a blocker

The `terminating_conversation_id` branch has no production caller in this PR — `_unsafe_destroy_home/2` still calls it unconditionally. The wiring lands downstream in `a19bc6ad` (`fence explicit actor termination before teardown`), which replaces the currently unfenced `_unsafe_sandbox_kept_on_terminate?(state.sandbox_id, ...)` reads in `conversation_server.ex`. Fine for a stack; flagging so this does not merge and then sit unwired if the tail of the stack stalls.

PR #1979, opened 2026-09-11
test(conversations): prove termination and attachment commit order

jhgaylor:
Conditional termination must decide whether the machine is still shared at the same database boundary that fences admission. Two PostgreSQL regressions now race that decision against the real public start_conversation attachment path, in both commit orderings.

If termination wins, attachment waits and then returns sandbox_reset_pending without inserting a conversation. If attachment wins, termination waits for its commit and returns sandbox_kept without a fence or teardown request event. The existing conversation remains idle and the provider is not touched.

Stack: based on #1978. Refs #1767. This verifies the context protocol; conversation actor integration is separate.

Validation:
- Both cases pass with independent PostgreSQL connections, telemetry pauses after the actual sandbox row queries, and pg_blocking_pids evidence of the waiter/holder relationship.
- Weakening the sandbox lock to FOR KEY SHARE makes both fail: the winning fence no longer blocks attachment, and a winning attachment is followed by an incorrect fence.
- Restoring the lock passes all 29 concurrent attachment, conditional fence, attachment API and teardown-fence tests.
- Committed fixture counts for users, sandboxes, conversations and turns were all zero after the mutation run.
- Full `mix precommit` passed: 5,415 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -3737,12 +3737,17 @@ defmodule Fountain.Conversations do
   retaining capacity until retirement completes. A new fence records
   `sandbox.teardown_requested` after commit; repeats preserve its timestamp.
   Refuses an enclosing transaction. `opts` carries actor, request_ip and reason.
+
+  With a terminating_conversation_id, first lock and verify that conversation's
+  current attachment and owner. A persistent home or another live conversation
+  returns {:error, :sandbox_kept} without a new fence. The sandbox row stays
+  locked through this decision and the fence, serializing supported attachments.
   """
   def _unsafe_fence_sandbox_for_teardown(%Sandbox{} = sandbox, opts \\ []) do
     if Repo.in_transaction?() do
       {:error, :provider_transaction_open}
     else
-      case do_fence_sandbox_for_teardown(sandbox) do
+      case do_fence_sandbox_for_teardown(sandbox, Keyword.get(opts, :terminating_conversation_id)) do
         {:ok, {fenced, true}} ->
           Audit.record(%{
             user_id: fenced.user_id,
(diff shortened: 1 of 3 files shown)


BinaryBourbon (approved):
## APPROVE

Verified non-vacuous by reverting the guard — twice, and the "no wait" assertion is what caught it.

**What it proves.** That `_unsafe_fence_sandbox_for_teardown/2`'s conditional fence and `start_conversation/2`'s attachment cannot interleave: the fence takes `sandboxes FOR UPDATE` and `create_attached_conversation` takes `sandboxes FOR SHARE`, so whichever commits first decides. Termination first gives the newcomer `:sandbox_reset_pending`; attachment first gives termination `:sandbox_kept` and leaves the machine unfenced with no audit row.

**Reversion check.** Dropping `lock: "FOR UPDATE"` to a plain read in `do_fence_sandbox_for_teardown/2`:

```
1) termination wins the race ... — no wait on the winning PostgreSQL connection observed
2) attachment wins the race ... — left: {:error, :sandbox_kept}  right: {:ok, ...}
2 tests, 2 failures
```

Both directions fail, and the first fails on the *wait* rather than on the outcome — which is the check that matters, because an outcome-only assertion would have stayed green on a fence that merely happened to win the schedule. Removing the `:sandbox_kept` policy from #1978 also fails the attachment case. The test is holding the real property.

**It waits on what it asserts.** `await_blocked/3` polls `pg_blocking_pids` on two genuinely separate backends (`refute holder == blocked` first, so a single-connection degradation cannot pass), with a deadline that fails loudly. The pre-release state is pinned too — `conversation_count == 1` and `refute reset_requested_at` before `:continue` — so a winner that had already committed would be caught rather than absorbed.

**Audit is asserted both ways**: exactly one `sandbox.teardown_requested` when the fence wins, and `== []` when it loses. A fence that refused but still announced would not slip through.

`async: false` plus `Sandbox.unboxed_run/2` is required here, and the `after` blocks clean up the rows the unboxed run commits.

### Non-blocking nits

1. The `:termination` pause matcher is `query =~ ~s(FROM "sandboxes")` with no lock-clause requirement, unlike the `:attachment` arm which does check `FOR SHARE`. It is unambiguous today (the advisory lock and the conversation lock queries do not match), and if the fence ever gained an earlier unlocked sandbox read the pause would land too early and fail on "no wait" rather than pass — so the failure direction is safe. Tightening it to `and query =~ "FOR UPDATE"` would make the intent self-evident and match the sibling test in `retirement_admission_order_test.exs`.
2. This file `reject`s `Managoat.Sandbox` while #1968 and #1974 stub `Managoat.Sandbox.Sprites`. Both work; one convention across the three would be easier to read.

PR #1980, opened 2026-09-11
fix(conversations): fence explicit actor termination before teardown

jhgaylor:
Explicit actor termination could close its adapter and destroy a machine while another connection admitted a turn. The actor now prepares the conditional admission fence first. An unshared ephemeral machine is fenced before adapter/provider work; persistent or shared machines keep the existing keep-machine path. A stale, missing or refused binding returns an error with the actor state intact.

The actor accepts both the legacy :terminate_conv message and an attributed {:terminate_conv, opts} form. Public callers still send the legacy request in this PR, so they retain the context's default fence attribution. Deploy this handler support before a separate caller migration starts forwarding attribution across nodes. Conversation termination's existing outer audit remains unchanged.

Stack: based on #1978. Refs #1767. Already-admitted work can be interrupted by this forced operation. Provider deletion remains best-effort, with terminal rows left for reconciliation when deletion fails.

Validation:
- Nine new cases produce seven failures on the parent; its two existing keep-machine cases pass.
- All 58 focused actor-fence, shared-sandbox, reprovision, broker and forced-home tests passed. Coverage includes bounded/unbounded admission, adapter/provider ordering, caller attribution, retention, stale/missing bindings, enclosing transactions and provider failure.
- Full `mix precommit` passed: 5,422 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests, all with zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -3737,12 +3737,17 @@ defmodule Fountain.Conversations do
   retaining capacity until retirement completes. A new fence records
   `sandbox.teardown_requested` after commit; repeats preserve its timestamp.
   Refuses an enclosing transaction. `opts` carries actor, request_ip and reason.
+
+  With a terminating_conversation_id, first lock and verify that conversation's
+  current attachment and owner. A persistent home or another live conversation
+  returns {:error, :sandbox_kept} without a new fence. The sandbox row stays
+  locked through this decision and the fence, serializing supported attachments.
   """
   def _unsafe_fence_sandbox_for_teardown(%Sandbox{} = sandbox, opts \\ []) do
     if Repo.in_transaction?() do
       {:error, :provider_transaction_open}
     else
-      case do_fence_sandbox_for_teardown(sandbox) do
+      case do_fence_sandbox_for_teardown(sandbox, Keyword.get(opts, :terminating_conversation_id)) do
         {:ok, {fenced, true}} ->
           Audit.record(%{
             user_id: fenced.user_id,
@@ -3768,16 +3773,24 @@ defmodule Fountain.Conversations do
     end
   end
 
-  defp do_fence_sandbox_for_teardown(sandbox) do
+  defp do_fence_sandbox_for_teardown(sandbox, ending_id) do
     Repo.transaction(fn ->
       Repo.query!("SELECT pg_advisory_xact_lock($1, $2)", [
         @sandbox_lock_namespace,
         :erlang.phash2(sandbox.id)
       ])
 
+      # Match admission's advisory -> conversation -> sandbox lock order.
+      lock_terminating_conversation(sandbox, ending_id)
+
       current =
         Repo.one(from s in Sandbox, where: s.id == ^sandbox.id, lock: "FOR UPDATE") ||
           Repo.rollback(:not_found)
+
+      if not is_nil(ending_id) and
+           (current.mode == "persistent" or _unsafe_sandbox_held_by_other?(current.id, ending_id)) do
+        Repo.rollback(:sandbox_kept)
+      end
 
       # Forced teardown may stop an admitted turn, but cannot admit a new one
       # after this commit. Keep capacity until the existing teardown finishes.
@@ -3792,6 +3805,18 @@ defmodule Fountain.Conversations do
         {fenced, true}
       end
     end)
+  end
+
+  defp lock_terminating_conversation(_sandbox, nil), do: :ok
+
+  defp lock_terminating_conversation(sandbox, ending_id) when is_binary(ending_id) do
+    Repo.one(
+      from c in Conversation,
+        where:
+          c.id == ^ending_id and c.sandbox_id == ^sandbox.id and c.user_id == ^sandbox.user_id,
+        select: c.id,
+        lock: "FOR UPDATE"
+    ) || Repo.rollback(:sandbox_unavailable)
   end
 
   # Destroy the sprite behind a home and retire its row. Best-effort on the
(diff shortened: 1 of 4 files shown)


BinaryBourbon (approved):
APPROVE — the fence covers the window the title claims, and the keep-machine decision moves under the sandbox lock, which is a genuine improvement over the unlocked `_unsafe_sandbox_kept_on_terminate?/2` it replaces.

**What this changes**

`handle_call(:terminate_conv, ...)` no longer decides "keep or destroy" from an unlocked read. `prepare_termination/2` re-reads the sandbox and calls `_unsafe_fence_sandbox_for_teardown/2` with `terminating_conversation_id`, so the home/co-tenant decision, the `FOR UPDATE` on the terminating conversation and the `reset_requested_at` fence all commit inside one advisory-locked transaction *before* `drop_connection/2` and `Managoat.Sandbox.destroy/1` run. The two outcomes become named helpers (`terminate_kept_machine/1`, `terminate_machine/2`), and a third outcome appears that did not exist before: the call can now reply `{:error, _}` and leave the server running. A legacy `:terminate_conv` clause delegates to `{:terminate_conv, []}` so no caller changes in this PR.

**Findings**

1. **Non-blocking — the fenced sandbox struct replaces a post-destroy re-read, and that is the right call, but the semantics changed for an already-terminal row.** `conversation_server.ex:2059` (`prepare_termination/2`): when the sandbox is already `terminated`/`failed` the fence returns `{:ok, {current, false}}`, so `terminate_machine/2` still calls `update_sandbox(sandbox, %{status: "terminated", terminated_at: now()})` and re-stamps `terminated_at`. The old code did the same thing unconditionally, so this is not a regression — but `Lifecycle.retire_terminated_sandbox/2` in #1982 grew an explicit `when status in ["terminated", "failed"] -> :ok` guard for exactly this case. The two termination paths now disagree about whether a terminal row gets re-stamped. Worth making them agree, on this stage or #1982's.

2. **Non-blocking, pre-existing — two co-tenants terminating concurrently can both take the kept path and strand the machine.** The `:sandbox_kept` decision is a `Repo.rollback`, so the advisory lock is released before `terminate_kept_machine/1` writes `conversations.status`. Conversations A and B both idle on ephemeral sandbox S: A fences, sees B live, rolls back `:sandbox_kept`; B fences before A's conversation write lands, sees A live, also rolls back; both conversations go `terminated` and S is left `ready` with nothing on it. The old unlocked code produced the same outcome, and because the fence rolled back, `reset_requested_at` is nil and `SandboxReaper.sweep_abandoned_sandboxes/0` does reclaim it after the abandoned grace — so this is bounded, not a leak. Recording it because the fence docstring's "serializing supported attachments" reads as if it also serialises this, and it does not.

3. **Non-blocking — a refusal is now reachable from the API with no mapping.** `{:error, :sandbox_unavailable}` and `{:error, :provider_transaction_open}` reach `FountainWeb.FallbackController` with no clause of their own, so they land on the terminal safety net: 422 with the bare atom plus a `Logger.warning("fallback: unmapped error atom …")` on every occurrence. The reachable case is narrow — `move_cotenants/3` repoints `conversations.sandbox_id` by `update_all` and casts `{:machine_gone, …}`, so a terminate that arrives between the repoint and the server processing that cast fences against a sandbox the row no longer names — and it is self-correcting (the server stops moments later; the retry takes the no-server path). But the operator-visible shape of that race is a warning log and an unexplained 422. A clause naming `sandbox_unavailable` as a retryable 409/503 would be cheap.

4. **Nit — the test drives `handle_call/3` directly against a hand-built state map.** `termination_actor_fence_test.exs:143` calls the callback with a nine-key literal rather than a started server, so the test cannot catch a regression in how `state.sandbox_id` is bound (`init/1`), which is the premise the whole fence rests on. The assertions inside the `expect(Managoat.Sandbox, :destroy, ...)` closure are the good part of this file — checking `reset_requested_at` is committed and that `_unsafe_create_turn_on_sandbox/3` already refuses at both capacities, from inside the provider call, is exactly the ordering proof the title claims, and it is proof the old code could not have passed.

**Checked and clean**

- `_unsafe_sandbox_kept_on_terminate?/2` and `mode == "persistent" or _unsafe_sandbox_held_by_other?/2` are semantically identical, including the `status not in ["terminated", "failed"]` filter on the co-tenant — no drift in the keep decision.
- Ordering: fence commits, then `drop_connection`, then `destroy`, then the terminal row writes. `Egress.release/2` is retained on the destroy path and correctly absent from the kept path; `handle: nil` on the kept stop is preserved.
- `interrupt_turn/1` still runs before `drop_connection/2` on the kept path only, as before.
- Audit: the only new record is `sandbox.teardown_requested`, written by `Fountain.Conversations` (the context), after the fence transaction commits, with `reason` + `provider` as metadata and no values. The legacy clause's `[]` yields the context default `"self"`, which the first test asserts.
- Attribution forwarded through `{:terminate_conv, opts}` is `"ui"` in the second test — inside the closed vocabulary.
- No `Task.async`, no `async: false`, pool size untouched.
- **The size pin is not this PR's doing.** `@pin 2762` is inherited from the branch point (3f59a76a, file 2685 lines) and no commit in the 13-commit stack touches `conversation_server_size_test.exs`. See the cross-stack note below.

**Cross-stack: the pin will be red after a rebase.** `main` is now exact at 2646/2646 (tightened by #1749 to 2652 and #1751 to 2646). This stack branched before those and still carries `@pin 2762`. The stack adds ~52 net lines to `conversation_server.ex`, so once these branches rebase onto current `main` the file is ~2698 against a pin of 2646 and the size test fails. Nothing in the stack can be raised to fix that — the pin only shrinks — so the added lines need to move into a `Fountain.Conversations.*` module before merge. #1982 already does this correctly by putting its helper in `Lifecycle`; this PR's ~56 new lines of helpers are the largest single contribution and are the obvious candidates.

No `CHANGELOG.md` `[Unreleased]` entry anywhere in the programme. Noting once, not per PR.

PR #1981, opened 2026-09-11
fix(conversations): guard and attribute termination requests

jhgaylor:
The termination client now refuses an enclosing database transaction before contacting an actor or updating the no-server fallback rows. It also forwards caller options to the actor, so the committed teardown fence receives the same actor and request IP as the existing conversation-termination audit. Actor refusals remain errors and produce no completed-termination audit.

Rollout prerequisite: deploy #1980 across the actor nodes before rolling out this caller change. This PR sends the attributed tuple that #1980 introduces; older handlers do not accept it. The existing public function signature is unchanged.

Stack: based on #1980. Refs #1767.

Validation:
- The corrected parent test uses a real local registry and a supervised stand-in actor: three of four cases fail, demonstrating dropped attribution and transaction leakage with or without an actor.
- All 149 focused client, actor, shared-sandbox, controller, audit and call-timeout tests passed.
- `mix precommit` passed: 5,426 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -301,6 +301,9 @@ defmodule Fountain.Conversations.ConversationServer do
   sprite. If not, just mark the DB rows terminated so the user can still
   clean up dead conversations after a server restart.
 
+  An enclosing database transaction is refused before contacting the actor or
+  updating rows, so teardown cannot escape a caller's rollback.
+
   Named `terminate_conversation` rather than `terminate`: taking `opts` for
   audit attribution (#545) would have made this `terminate/2`, which is the
   OTP callback below. Two different meanings under one name in one module was
@@ -309,12 +312,16 @@ defmodule Fountain.Conversations.ConversationServer do
   the client half gets the unambiguous name.
   """
   def terminate_conversation(conv_id, opts \\ []) do
-    # ownership: callers established this conversation's tenant. Cleanup must
-    # survive a blocked actor, failed provider teardown, or subsequent deletion.
-    case Fountain.Conversations.ExecutionGuard._unsafe_interrupt(conv_id) do
-      {:ok, _} -> terminate_after_retirement(conv_id, opts)
-      {:error, :not_found} -> {:error, :not_running}
-      {:error, _} = error -> error
+    if Fountain.Repo.in_transaction?() do
+      {:error, :provider_transaction_open}
+    else
+      # ownership: callers established this conversation's tenant. Cleanup must
+      # survive a blocked actor, failed provider teardown, or subsequent deletion.
+      case Fountain.Conversations.ExecutionGuard._unsafe_interrupt(conv_id) do
+        {:ok, _} -> terminate_after_retirement(conv_id, opts)
+        {:error, :not_found} -> {:error, :not_running}
+        {:error, _} = error -> error
+      end
     end
   end
 
@@ -336,6 +343,7 @@ defmodule Fountain.Conversations.ConversationServer do
               # must not take either down.
               sandbox_id = conv.sandbox_id
 
+              # ownership: sandbox_id comes from that same authorized conversation.
               if is_binary(sandbox_id) and
                    not Conversations._unsafe_sandbox_kept_on_terminate?(sandbox_id, conv.id) do
                 sb = Conversations._unsafe_get_sandbox!(sandbox_id)
@@ -349,7 +357,16 @@ defmodule Fountain.Conversations.ConversationServer do
           end
 
         pid ->
-          call_server(pid, :terminate_conv)
+          # This pid is routinely on another pod, and mid-deploy some pods
+          # predate the tuple clause: they answer the catch-all with
+          # `:unknown_call`, which would leave the sprite running and billing.
+          # Retry the bare atom they understand. Only that catch-all produces
+          # `:unknown_call` here and it has no side effects, so no double
+          # terminate; #1980 keeps the mirror clause for old-pod callers.
+          case call_server(pid, {:terminate_conv, Keyword.take(opts, [:actor, :request_ip])}) do
+            {:error, :unknown_call} -> call_server(pid, :terminate_conv)
+            other -> other
+          end
       end
 
     audit_lifecycle(conv_id, "conversation.terminated", result, opts)
--- a/apps/fountain/test/fountain/conversations/termination_client_test.exs
+++ b/apps/fountain/test/fountain/conversations/termination_client_test.exs
@@ -1,0 +1,133 @@
+defmodule Fountain.Conversations.TerminationClientTest do
+  use Fountain.DataCase, async: true
+
+  alias Fountain.Audit
+  alias Fountain.Conversations.ConversationServer
+
+  defmodule Probe do
+    use GenServer
+
+    def start_link(args), do: GenServer.start_link(__MODULE__, args)
+
+    def init({owner, reply, conv_id}) do
+      {:ok, _} = Horde.Registry.register(Fountain.ConversationRegistry, conv_id, nil)
+      {:ok, {owner, reply}}
+    end
+
+    def handle_call(message, _from, {owner, reply} = state) do
+      send(owner, {:termination_request, self(), message})
+
+      # A function reply lets a test answer by message shape, which is how a
+      # pod predating the tuple clause behaves.
+      case reply do
+        fun when is_function(fun, 1) -> {:reply, fun.(message), state}
+        term -> {:reply, term, state}
+      end
+    end
+  end
+
+  setup do
+    user = insert_verified_user()
+    sandbox = insert_sandbox(user_id: user.id, status: "ready")
+    conv = insert_conversation(user_id: user.id, sandbox: sandbox, status: "idle")
+    %{user: user, sandbox: sandbox, conv: conv}
+  end
+
+  test "forwards attribution to the actor and keeps the outer lifecycle audit", ctx do
+    pid = probe(ctx.conv.id, :ok)
+    opts = [actor: "ui", request_ip: "192.0.2.4"]
+    assert :ok = ConversationServer.terminate_conversation(ctx.conv.id, opts)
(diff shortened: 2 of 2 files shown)


BinaryBourbon (changes requested):
REQUEST_CHANGES — the attribution half is correct, but sending `{:terminate_conv, opts}` unconditionally breaks terminate for the duration of every rolling deploy, and the stack's own merge order cannot satisfy the rollout prerequisite this PR's description states.

**What this changes**

Two things. `terminate_conversation/2` refuses an enclosing `Repo.transaction`, before either the actor call or the no-server row writes, by splitting the body into `do_terminate_conversation/2`. And the client stops sending the bare `:terminate_conv` atom, sending `{:terminate_conv, opts}` instead so the fence #1980 commits inside the actor records the same actor and request IP that the outer `conversation.terminated` row already carried.

**Findings**

1. **Blocking — new node → old node terminate returns `{:error, :unknown_call}` and leaves the sprite running.** `conversation_server.ex:351`. `call_server/2` is `GenServer.call` on a pid from the cluster-wide `Fountain.ConversationRegistry`, so the target server is routinely on another pod. A pod that predates #1980 has no `{:terminate_conv, opts}` clause; the tuple falls through to the `handle_call/3` catch-all at `:1662`, which replies `{:error, :unknown_call}`. `call_server/2` (`:152`) catches only the five `:exit` shapes, so that error is returned verbatim: `terminate_conversation/2` returns `{:error, :unknown_call}`, `audit_lifecycle/5` records nothing because `result != :ok`, the conversation is not terminated, the sandbox is not retired, **the sprite keeps running and keeps burning turn/provider time**, and the API renders 422 `unknown_call` through the fallback safety net with a `Logger.warning`.

   The description says "deploy #1980 across the actor nodes before rolling out this caller change", and #1980's says the same in reverse. That sequencing is exactly what a stack cannot deliver: #1980 and #1981 are consecutive squashes onto `main` minutes apart and will be in one image, or at best two images deployed back to back — either way the pods that are *old* during #1981's rollout are pre-#1980 pods that do not understand the tuple. The window is the rollout, and every terminate routed to an old pod inside it fails.

   Cheap fix, entirely inside this PR:

   ```elixir
   pid ->
     case call_server(pid, {:terminate_conv, opts}) do
       {:error, :unknown_call} -> call_server(pid, :terminate_conv)
       other -> other
     end
   ```

   That loses attribution against an old node — which is the correct degradation, since the fence there defaults to `"self"` — and costs nothing once every pod is new. Alternatively, hold the client change until #1980 has actually been deployed, which is a release-ordering commitment somebody has to own rather than a sentence in a PR body.

2. **Non-blocking — the transaction guard is right and does not strand any existing caller.** I checked all seven: `Deletion.destroy_sprites/1` (`deletion.ex:172`) runs outside `delete_user/2`'s writes and destroys sprites directly afterwards regardless; `Principals.stop_compute/1` is called after `close/5`'s `Repo.transaction` returns (`principals.ex:630`); `_unsafe_reap_sandbox/1` (`conversations.ex:177`), `Team.remove_teammate/3` (`team.ex:314`) and the controller are all transaction-free; `_unsafe_destroy_home/2` already has the same guard above it. So no caller silently loses its termination. Worth keeping this list in the PR body, because a future caller added inside a transaction will get a silent no-op with `audit: false` swallowing the evidence.

3. **Non-blocking — `opts` is forwarded whole, including keys the actor has no business seeing.** `audit: false` from `Deletion` and `Team.remove_teammate/3` now travels into `handle_call({:terminate_conv, opts}, ...)` and on into `_unsafe_fence_sandbox_for_teardown/2`, which ignores it. Harmless today; it becomes a trap the moment the fence grows an option whose name collides. `Keyword.take(opts, [:actor, :request_ip])` at the send site would say what is actually part of the wire contract.

**Checked and clean**

- Actor vocabulary: every path that reaches the fence supplies a legal value — `"ui"`/`"api"` from `Audited.attribution/2` (the route is authenticated, so it never derives a bare `"system"`), `"system:sandbox_reaper"`, `"system:home_reset"`, and the context default `"self"` where `opts` carries only `audit: false`.
- The fence's `Audit.record/1` is in the context, is called after `do_fence_sandbox_for_teardown/2`'s transaction has committed (not inside it), and records `reason` + `provider` only — no values.
- `audit_lifecycle/5` still gates on `result == :ok`, so the refusal paths record nothing. Correct per "only record what happened".
- Tests have teeth: `Probe` registers in the real Horde registry and asserts the exact message shape, so a revert to the bare atom fails the pattern match; the transaction cases assert `refute_received` plus unchanged `conversations.status` and `sandboxes.status`, and fail on a revert with a live probe (`{:ok, :ok}` vs `{:ok, {:error, :provider_transaction_open}}`). Both run `async: true` with no shared global state.
- The added docstring paragraph describes only behavior this PR supplies.

**Cross-stack** (same note on all four): `main` tightened `@pin` to an exact 2646 via #1749 and #1751 after this stack branched; the stack still carries the inherited `@pin 2762` and no commit in it touches the size test. Rebased onto current `main` the file is ~2698 against a pin of 2646 and `conversation_server_size_test.exs` fails. The pin only shrinks, so the fix is moving code out, not raising it. No `CHANGELOG.md` `[Unreleased]` entry anywhere in the programme.

BinaryBourbon (approved):
**Approve.** The rolling-deploy hole is closed, the conflict resolution is faithful to both sides, and I verified the retry's safety premise independently rather than accepting it.

### The conflict resolution

Diffed the whole file against `origin/main` and against the pre-rebase tip `e08ec8b8`.

- `main` had `terminate_conversation/2` wrapping `ExecutionGuard._unsafe_interrupt/1` and calling `terminate_after_retirement/2`. The pre-rebase branch had the `Repo.in_transaction?()` refusal delegating to a locally-renamed `do_terminate_conversation/2` and no execution guard at all (that landed on main after the branch was cut).
- The merged form (`conversation_server.ex:315-325`) keeps the transaction refusal outermost and the execution-guard case inside it, and drops the branch's private rename in favour of main's `terminate_after_retirement/2`. Nothing from either side is lost: all three `_unsafe_interrupt` result arms survive, and the body of the old `do_terminate_conversation/2` is byte-identical to main's `terminate_after_retirement/2` apart from the pid clause this PR changes.
- The ordering is the right way round. `_unsafe_interrupt/1` writes rows, so putting the refusal inside the case would let an interrupt run inside a caller's transaction — exactly what the guard exists to prevent. The two `live_actor` variants at `termination_client_test.exs:56-67` pin this: inside `Repo.transaction/1` the call returns `{:error, :provider_transaction_open}`, no message reaches the actor (`refute_received`), and the conversation and sandbox rows are untouched. Removing the guard fails both.

### Is the retry safe?

I traced every producer of `{:error, :unknown_call}` in the tree rather than relying on the claim.

- `conversation_server.ex:1627-1630` — the `handle_call/3` catch-all. Logs and replies; no side effects.
- `pending.ex:523` — `Pending.await/3` for a missing MCP call id. Reaches a caller only through the caller-tools await path (`caller_tools.ex:316`), never as the reply to a `{:terminate_conv, _}` call.

The tuple clause itself (`:1568-1573`) cannot produce it: `prepare_termination/2` returns `{:error, :sandbox_unavailable}` or a fence error, and `_unsafe_fence_sandbox_for_teardown/2` (`conversations.ex:3706-3735`) returns only `:provider_transaction_open` or the transaction's `:not_found`. `call_server/2`'s `catch` clauses (`:146-151`) map exits to `:provisioning` and `:not_running`. So on this reply path the catch-all is the sole source, and the retry cannot double-terminate. The `refute_received` in `termination_client_test.exs:97` and `:106` proves no retry fires on either a normal `:ok` or an ordinary refusal.

I also confirmed the other half of the rollout is covered: #1980's mirror clause at `:1565-1566` maps a bare `:terminate_conv` from an old-pod caller onto the tuple handler, so both directions of a mixed fleet work.

### `Keyword.take`

Enumerated the callers of `terminate_conversation/2`: `team.ex:314`, `conversations.ex:194`, `:1589`, `:3687`, `accounts/deletion.ex:172`, `conversation_controller.ex:876`. Between them they pass `audit: false`, `actor: "system:..."` and `Audited.attribution(conn)` — no caller passes `:reason` or anything else the remote reads. On the remote side `prepare_termination/2` uses only `:actor`, `:request_ip` and `:reason` (which it `put_new`s to `"conversation_terminated"` anyway), and `:audit` is consumed locally in `audit_lifecycle/5` after the call returns, so it was never the actor's business. Nothing a caller needs is dropped, and the allowlist additionally stops a caller injecting `:terminating_conversation_id`, which `prepare_termination/2` sets itself and which steers the fence's kept-machine decision. Net improvement.

### The tests

Not vacuous — I broke each fix and watched them fail:

- dropping the fallback fails "retries the bare atom when the actor pod does not know the tuple";
- dropping `Keyword.take` fails "forwards only attribution keys to the actor";
- dropping the transaction guard fails both `live_actor` variants.

On the audit claim specifically: `audit_lifecycle/5` gates on `Keyword.get(opts, :audit, true) and result == :ok` (`:459`), and `terminate_after_retirement/2` passes the retry's return value straight through as `result` (`:365-372`). So `assert [event] = events(ctx)` at `termination_client_test.exs:87` really does depend on the fallback having succeeded — without the retry, `result` is `{:error, :unknown_call}` and no event is written. The Probe's function-reply form answers by message shape, which is a fair model of a pod that has the catch-all but not the tuple clause.

`conversation_server.ex` is 2637 lines against the pin of 2641 (`conversation_server_size_test.exs:40`), so this stays under it.

Ran locally against a clean database: 73 tests across `termination_client_test.exs`, `conversation_server_test.exs` and `teardown_fence_test.exs`, 0 failures. `mix format --check-formatted` and `mix credo --strict` clean. Full CI green on cb503aa.

One thing worth knowing rather than changing: on the fallback path the old pod runs its pre-#1980 `:terminate_conv`, so no `sandbox.teardown_requested` fence event is recorded for that machine and no attribution reaches the actor. That is correct — an old pod cannot do either — and the outer `conversation.terminated` event still carries the actor. It just means the fence's coverage is not complete until the rollout is.

PR #1982, opened 2026-09-11
fix(conversations): fence termination without a running actor

jhgaylor:
When termination finds no running actor, the fallback now commits the shared conditional teardown fence before retiring the sandbox row. A home or an attachment that wins the fence decision keeps its machine; attachments after the fence are refused. The teardown intent carries caller attribution. Fence and retirement-write errors now propagate instead of producing a completed-termination audit.

The conversation row still becomes terminal first. If the later sandbox operation fails, that thread stays ended, and the error tells the caller machine retirement did not complete. This preserves the existing ordering; it does not claim an atomic multi-row termination. No provider I/O runs in the fallback: terminal rows remain the reaper's cleanup responsibility.

The helper lives in the existing lifecycle module so ConversationServer stays under its fixed size guard.

Stack: based on #1981. Refs #1767. The conditional fence's real two-connection attachment ordering is separately covered by #1979.

Validation:
- Four new regression cases fail on the parent code (missing fence, conditional retention, and error propagation).
- 100 focused fallback, client, actor, audit, server and attachment tests passed before the lifecycle extraction.
- 103 lifecycle, fallback, size and audit-guard tests passed after extraction.
- Full `mix precommit` passed: 5,432 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all zero failures.

No deployed environment or real provider was modified.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -334,26 +334,10 @@ defmodule Fountain.Conversations.ConversationServer do
               {:error, :not_running}
 
             conv ->
-              now = DateTime.utc_now() |> DateTime.truncate(:second)
-              {:ok, _} = Conversations.update_conversation(conv, %{status: "terminated"})
-
-              # A sandbox handed on to a successor conversation
-              # (`release_conversation/2`) is that conversation's computer now,
-              # and a home is the agent's (ADR 0023); terminating this thread
-              # must not take either down.
-              sandbox_id = conv.sandbox_id
-
-              # ownership: sandbox_id comes from that same authorized conversation.
-              if is_binary(sandbox_id) and
-                   not Conversations._unsafe_sandbox_kept_on_terminate?(sandbox_id, conv.id) do
-                sb = Conversations._unsafe_get_sandbox!(sandbox_id)
-
-                if sb.status not in ["terminated", "failed"] do
-                  Conversations.update_sandbox(sb, %{status: "terminated", terminated_at: now})
-                end
+              with {:ok, terminated} <-
+                     Conversations.update_conversation(conv, %{status: "terminated"}) do
+                Lifecycle.retire_terminated_sandbox(terminated, opts)
               end
-
-              :ok
           end
 
         pid ->
--- a/apps/fountain/lib/fountain/conversations/lifecycle.ex
+++ b/apps/fountain/lib/fountain/conversations/lifecycle.ex
@@ -334,6 +334,52 @@ defmodule Fountain.Conversations.Lifecycle do
   end
 
   @doc """
+  Retire the machine of an authorized, terminated conversation with no actor.
+  The conditional fence preserves homes and other live co-tenants, and blocks
+  new attachments before the terminal write. No provider I/O runs here; the
+  reaper handles terminal rows. The caller owns the conversation lifecycle
+  audit; the fence records teardown intent using the supplied attribution.
+  """
+  def retire_terminated_sandbox(%{sandbox_id: nil}, _opts), do: :ok
+
+  def retire_terminated_sandbox(conv, opts) do
+    opts =
+      opts
+      |> Keyword.put(:terminating_conversation_id, conv.id)
+      |> Keyword.put_new(:reason, "conversation_terminated")
+
+    # ownership: this sandbox belongs to the conversation authorized by the caller.
+    case Conversations._unsafe_get_sandbox(conv.sandbox_id) do
+      nil ->
+        {:error, :sandbox_unavailable}
+
+      sandbox ->
+        # ownership: the authorized conversation supplies this sandbox; the fence rechecks binding.
+        case Conversations._unsafe_fence_sandbox_for_teardown(sandbox, opts) do
+          {:ok, %{status: status}} when status in ["terminated", "failed"] ->
+            :ok
+
+          {:ok, fenced} ->
+            now = DateTime.utc_now() |> DateTime.truncate(:second)
+
+            with {:ok, _} <-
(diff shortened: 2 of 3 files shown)


BinaryBourbon (approved):
APPROVE — the fallback now takes the same locked decision the actor path takes, and putting the helper in `Lifecycle` rather than the server is the right call. Three findings, none blocking.

**What this changes**

The no-server branch of `terminate_conversation/2` stops doing its own unlocked `_unsafe_sandbox_kept_on_terminate?/2` check followed by a bare `update_sandbox/2`. It marks the conversation terminated and hands the machine to `Lifecycle.retire_terminated_sandbox/2`, which re-reads the sandbox, commits the shared conditional teardown fence (advisory lock, `FOR UPDATE` on the terminating conversation, home/co-tenant check, `reset_requested_at`) and only then writes the terminal row. No provider I/O — the reaper still owns the sprite. Fence and write errors now propagate to the caller instead of being swallowed.

**Findings**

1. **Non-blocking — a completed termination can return an error and record no `conversation.terminated` row.** `conversation_server.ex:328` and `lifecycle.ex:344`. The conversation row is written terminal *before* the fence, so on `{:error, :sandbox_unavailable}` the conversation is terminated and the caller is told the operation failed; `audit_lifecycle/5` gates on `result == :ok`, so nothing records the termination that did happen. `termination_fallback_test.exs:63` asserts exactly this (`status == "terminated"` alongside `events(ctx, "conversation.terminated") == []`), so it is a decision rather than an accident, and the description is honest about the ordering. But CLAUDE.md's audit contract is that a mutation records its own event — a change with no row is the same hole the #540 campaign closed from the other side, and the caller retrying gets the same error forever while the UI shows the conversation gone.

   Reachable, if narrowly: `move_cotenants/3` repoints `conversations.sandbox_id` with `update_all`, so a conversation with no live server that a sibling's wake is moving fails `lock_terminating_conversation/2`'s `c.sandbox_id == ^sandbox.id` and the fence rolls back `:sandbox_unavailable`. Either record the lifecycle audit before the sandbox half, or return `:ok` with the sandbox failure logged — the reaper does reclaim the row in this case, because the rolled-back fence leaves `reset_requested_at` nil.

2. **Non-blocking — `{:error, :write_refused}` leaves a row nothing can ever reclaim, and the test enshrines it.** `termination_fallback_test.exs:73` and `lifecycle.ex:360`. When `update_sandbox/2` fails after the fence has committed, the sandbox stays `ready` with `reset_requested_at` set. From there: `SandboxReaper` skips it in **both** passes (`is_nil(s.reset_requested_at)` at `sandbox_reaper.ex:125` and `:228`), `Quotas.active_sandboxes/0` keeps counting it (`quotas.ex:145`), `check_attachable/4` refuses attach, and nothing anywhere clears the field. One of the tenant's 2–20 slots and one fleet slot are gone permanently and the sprite is never destroyed. This is #1894 verbatim, reached through a second door, and the same thing happens on a node crash between the fence commit and the terminal write — a window this PR widens by putting a fence commit ahead of a row write with no journal. Not this PR's to fix (#1983 records the intent, #1894 owes the sweep), but the test currently reads as "this outcome is fine", and it should say which issue owns the exit.

3. **Non-blocking — `_unsafe_sandbox_kept_on_terminate?/2` is now dead, with a docstring that says otherwise.** `conversations.ex:3291`: "Both `ConversationServer` terminate paths ask this." #1980 removed the first caller and this PR removes the second, so the only remaining references are two assertions in `sandbox_mode_test.exs`. Either delete it with its test moved onto `_unsafe_sandbox_held_by_other?/2`, or correct the docstring — a public `_unsafe_` function claiming callers it does not have is the kind of thing the next reader trusts.

4. **Nit — the deleted comment was the only place `release_conversation/2` explained this check.** The removed block named the two reasons the kept decision exists ("a sandbox handed on to a successor conversation … and a home is the agent's"). The new docstring compresses that to "preserves homes and other live co-tenants", which is accurate but drops the pointer to the mechanism that creates the second case. `_unsafe_sandbox_held_by_other?/2`'s own docstring still carries it, so this is recoverable, just one hop further away.

**Checked and clean**

- The fence's kept decision is semantically identical to the `_unsafe_sandbox_kept_on_terminate?/2` it replaces, co-tenant status filter included, so no behavior drift in what survives a terminate.
- `retire_terminated_sandbox/2`'s early `:ok` for `status in ["terminated", "failed"]` matches the old `sb.status not in [...]` guard. (#1980's `terminate_machine/2` has no equivalent and re-stamps `terminated_at`; the two paths disagree, noted there.)
- `%{sandbox_id: nil}` clause is dead given the `NOT NULL` column, but mirrors the old `is_binary(sandbox_id)` guard — harmless.
- `with {:ok, terminated} <- update_conversation(...)` replaces a `{:ok, _} =` match, so a changeset error is now a 422 instead of a crash. Improvement.
- Audit: the fence records in the context, after its transaction commits, with `reason`/`provider` metadata and no values; `"ui"` + request IP are asserted end to end.
- Tests would fail on a revert: the first drives `attach/1` to `{:error, :sandbox_reset_pending}` from inside the `update_sandbox` closure, which is only true once the fence is committed, and the second wins an attachment *before* the fence and asserts the machine survives. Neither waits on a signal other than the one it asserts. `async: true`, no global state.
- Helper placed in `Lifecycle` to stay under the size pin — correct instinct, and the only stage in the stack that does it.

**Cross-stack** (same note on all four): `main` tightened `@pin` to an exact 2646 via #1749 and #1751 after this stack branched; the stack still carries the inherited `@pin 2762` and no commit in it touches the size test. Rebased onto current `main` the file is ~2698 against a pin of 2646 and `conversation_server_size_test.exs` fails; the pin only shrinks, so the fix is moving more code out. No `CHANGELOG.md` `[Unreleased]` entry anywhere in the programme.

Issue #1864, opened 2026-09-11
Track remaining lifecycle extractions from #1754


jhgaylor:
Finish replacing frozen #1754 with focused, independently reviewable PRs. The small-extraction audit is complete; the remaining changes need working lifecycle paths with their persistence and recovery rules.

Reference: [frozen source](https://github.com/managoat/fountain/tree/ae6f8b2bb86276fca5e812fe7455edeb5a4cdc0a), audited against main `bed264af` on 2026-09-10. Parent: #1732; consumer: [Review Loop #30](https://github.com/managoat/review-loop/issues/30).

Already extracted: #1796, #1797, #1818, #1820, #1826, #1827, #1831 and #1840. Reset fencing and its caller-transaction guard belong to #1768. Preserve the earlier #1761–#1764 fixes and #1801's same-identity co-tenant rule. Recheck current main before each extraction.

Suggested sequence; replace each checkbox with its implementing PR links:

- [ ] **Provider create intent and result.** Integrate `SandboxOperations` with one real provisioning caller. Commit intent before provider I/O; retain physical identity, original owner and capacity through timeout, caller loss and parent deletion. Test confirmed success, definite refusal and uncertain completion. Pair dependency upgrades with their tested consumer.
- [ ] **Confirmed deletion and reconciliation.** Extend that journal through teardown, reaping and account deletion. Release reservations only with sufficient provider evidence; retain uncertain obligations and their accounting. Test crash/retry and name reuse. GET 404 alone must not release the existing uncertain create `53c090d1-fc9e-46e9-b88a-478a4c984f4a`.
- [ ] **Committed holder transfer.** Extract the relevant `SandboxHolders` behavior with conditional binding updates, attachment-clock writes and post-commit notifications. Preserve tenant/environment/vault identity, reject stale transfers and account for unused reservations. Test concurrent wake/rebind. Current notifications precede binding updates; the frozen destination-checked callback requires a committed producer.
- [ ] **Actor launch ownership and recovery.** Integrate claims, launch/startup outcomes and one worker startup/teardown path. Fence stale callbacks by actor incarnation; registry absence must not authorize takeover of unresolved work. Test caller loss, restart, reconnect and delayed callbacks. Add dispatch/watchdog recovery in follow-up PRs with their callers.
[...]

PR #1983, opened 2026-09-11
fix(conversations): persist forced teardown intent separately

jhgaylor:
Forced teardown now records `teardown_requested_at` alongside the shared admission fence. An ordinary home reset leaves that field empty. Escalating an existing reset records a forced-teardown intent and its audit event while preserving the original `reset_requested_at`; repeating the forced request preserves both timestamps and emits no duplicate intent.

This gives recovery code a durable distinction between the two operations, including ephemeral and provisioning sandboxes. The field is internal and cannot be set through general sandbox attributes. Terminal rows and refused conditional teardowns gain no new intent.

Stack: based on #1982. Refs #1767 and #1894. This PR records the distinction; the forced-teardown reconciliation worker is a separate follow-up. Existing untyped fences are not guessed or backfilled from best-effort audit records. It does not implement the provider-operation journal tracked by #1864.

Validation:
- Two regression cases fail against the existing fence implementation.
- 115 intent, retention, actor/fallback, forced-home and audit-guard tests passed.
- Migration rollback preserved a pending row and its original admission fence; reapply allowed the row to acquire forced intent without replacing that fence. The disposable fixture was explicitly removed and both row counts checked at zero.
- Full `mix precommit` passed: 5,434 core tests + 6 doctests, 144 Buzz tests and 33 Support tests, all zero failures.

No provider calls or deployed environments were used for the migration checks.


BinaryBourbon (approved):
APPROVE — the column is well-scoped, the escalation case it fixes was silently losing an audit row, and the "cannot be forged through general attributes" property is real and tested. The answer to "who clears it" is nobody, which is correct for evidence but needs saying out loud.

**What this changes**

`sandboxes` gains `teardown_requested_at`, written only by `do_fence_sandbox_for_teardown/2`. The fence's idempotency key moves from `reset_requested_at` to `teardown_requested_at`, which changes one case: a row that already carries an ordinary reset fence used to return `{current, false}` — no write, **and no `sandbox.teardown_requested` audit row** — and now gets the forced intent stamped, keeps its original `reset_requested_at`, and records the audit. Admission is untouched: the shared `reset_requested_at` is still the only thing `check_attachable/4`, `maybe_reuse_sandbox/1`, `Quotas.active_sandboxes/0` and `HomeCheckpoint.on_park/1` read.

**Findings**

1. **Non-blocking, but the headline — the intent has exactly one reader, and it is this fence's own idempotency check.** Nothing clears `teardown_requested_at` and nothing acts on it; no sweep, no admin surface, no reaper query. That is defensible for evidence (`reset_requested_at` carries the same "retained after completion" comment), and the description is straight about the worker being a follow-up. What the description does not say is how bad the state it now labels actually is, so for the record: a fenced non-terminal row is skipped by `SandboxReaper` in **both** passes (`is_nil(s.reset_requested_at)` at `sandbox_reaper.ex:125` and `:228`), is still counted by `Quotas.active_sandboxes/0` (`quotas.ex:145`) against both the tenant cap and `SANDBOX_FLEET_CEILING`, and refuses every reuse path. So an interrupted forced teardown — `update_sandbox/2` failing in `Lifecycle.retire_terminated_sandbox/2` (#1982), or a pod dying between the fence commit and the terminal write — costs the tenant one of its 2–20 slots permanently and leaks the sprite, with no lever short of a hand-written production `UPDATE`.

   The good news, and the difference from what round 1 found in the upper chain: the deferral target here is real. #1894 is open and describes this exact dead end, option 1 in it is the sweep this column enables, and #1983 gives it the predicate it was missing:

   ```sql
   where teardown_requested_at is not null and status not in ('terminated','failed')
   ```

   Please put that query (or the equivalent Ecto) in #1894 as the acceptance criterion while the reasoning is fresh, and link this PR from it. A column whose only purpose is to feed a worker that does not exist yet is one refactor away from being deleted as unused.

2. **Non-blocking — escalation is one-way and a retry of a failed forced teardown records nothing.** `conversations.ex:3500`. Because the key is `is_nil(current.teardown_requested_at)`, the second and later forced requests against the same row are silent: no write, no audit. That matches the documented "repeats preserve both timestamps" and matches the old behavior for repeats, so it is not a regression — but combined with finding 1 it means the operator trail for a stranded row shows one intent and no evidence of how many times someone tried to clear it. Worth a `metadata["attempt"]`-style distinction if #1894's admin path ever lands.

3. **Nit — the schema comment undersells the new field relative to its neighbour.** `sandbox.ex:47` says "Forced teardown intent; admission still uses the shared reset fence", which is exactly right about the mechanism, but the neighbouring `reset_requested_at` comment explains *why* it survives completion ("retained after completion as operation evidence"). Given finding 1, the one thing a future reader most needs from this line is that nothing clears it and nothing reads it yet. One clause.

**Checked and clean**

- The `cond` is correct on every branch: terminal rows get no intent (asserted for both `terminated` and `failed`), a clean row gets both stamps at the same `now` (asserted equal), an escalated row keeps its original `reset_requested_at` and gets a strictly later `teardown_requested_at` (asserted `:gt`), and repeats no-op. Dropping the old `not is_nil(current.reset_requested_at)` short-circuit is the intended behavior change, not an oversight.
- `reset_requested_at: current.reset_requested_at || now` writes no change when the field is already set, so the escalation genuinely preserves the original admission fence rather than refreshing it. The migration-rollback check in the description covers the same property from the other side.
- Not castable: neither timestamp is in `Sandbox.changeset/2`'s `cast/3` list, so `update_sandbox/2` cannot forge or clear the intent. The test asserts both directions (`nil` and a future timestamp) produce no change and leave `get_field/2` at the committed value — a genuinely good test, and one that fails the moment somebody adds the field to the cast list for convenience.
- `reset_sandbox/2` is untouched and its own fence still stamps only `reset_requested_at`; the "an ordinary failed reset does not become a forced teardown" case pins that boundary, and it is the case that would silently break if the two fences were ever merged.
- Migration: nullable column, no default, no backfill — metadata-only on Postgres, reversible, and the comment is right that reversing keeps the admission fence intact. Type `:utc_datetime_usec` matches both the schema field and `reset_requested_at`.
- Audit: still recorded in the context, after the fence transaction commits, `reason` + `provider` only, no timestamp values in metadata, `"ui"` inside the closed actor vocabulary. The `Audit.record` closure asserting `refute Repo.in_transaction?()` *and* that both timestamps are already committed is the right shape for this check.
- No new `:noop` path, no provider I/O, no `Task.async`, no `async: false`, pool size untouched. `ee/` untouched.

**Cross-stack** (same note on all four): `main` tightened `@pin` to an exact 2646 via #1749 and #1751 after this stack branched; the stack still carries the inherited `@pin 2762` and no commit in it touches the size test. Rebased onto current `main` the file is ~2698 against a pin of 2646 and `conversation_server_size_test.exs` fails; the pin only shrinks, so the fix is moving code out of the server, not raising the number. No `CHANGELOG.md` `[Unreleased]` entry anywhere in the programme — `teardown_requested_at` is internal, so this stage probably owes none, but #1980–#1982 change what the terminate endpoint can return.

Issue #1413, opened 2026-09-02
CI: run affected SDK jobs and fan out on public API changes


jhgaylor:
## Problem

The main CI workflow currently installs and tests the SDK projects on the
general code path. That was inexpensive with one or two clients, but the suite
now contains TypeScript, Elixir and Swift, with Python nearly merged. Running
all language toolchains for an unrelated UI, docs or server-internal change
will add fixed setup time and runner cost as the SDK count grows.

We cannot solve this with SDK-directory path filters alone: an API/OpenAPI
change must fan out to every SDK even when no `sdk/<language>` path was touched.
Branch protection also needs a stable check result when individual matrix legs
are skipped.

## Proposed routing

Classify changes once near the start of CI, then use the result to select SDK
jobs:

| Change | SDK checks |
|---|---|
| `sdk/<language>/**` | That SDK |
| API schema, router, public controllers or OpenAPI generation | Every SDK |
| `sdk/conformance/**` | Every SDK |
| One SDK's release/test workflow or tooling | That SDK |
| Unrelated server internals, UI or docs | None, unless a checked SDK snippet is affected |
| Classifier uncertainty | Every SDK (fail safe) |

## Acceptance criteria

- [ ] SDK checks are separate jobs or matrix legs rather than serial steps on
      the general server CI critical path.
- [ ] A change under one `sdk/<language>` runs that SDK's format, minimum-runtime
      compatibility, tests, docs/package dry run and contract checks.
- [ ] A public API or OpenAPI schema change runs all four SDK legs.
- [ ] A shared conformance-fixture change runs all four SDK legs.
- [ ] Unrelated changes do not install all SDK toolchains.
- [ ] A stable aggregate `SDK checks` result is always reported for branch
      protection, including when every language leg is skipped.
- [ ] Pushes to `main`, workflow dispatches and pull requests preserve the
      current fail-closed behavior.
- [ ] The classifier has tests or fixture cases for each row in the routing
      table.
- [ ] CI documentation says where a new SDK must register its paths and
      contract triggers.
- [ ] Before/after workflow timing and runner-job counts are recorded in the PR.

## Non-goals

- Skipping SDKs when the public contract may have changed.
- Combining SDK release versions or publish workflows.
- Replacing each language's native test runner.

jhgaylor:
[...]

PR #1984, opened 2026-09-11
ci: run the Elixir SDK checks in a separate job

jhgaylor:
The Elixir SDK's ten setup/check steps now run in their own `elixir-sdk` job, parallel with the remaining Python/TypeScript, CLI and plugin checks. The toolchain versions, cache key and commands are unchanged. Fixture lint runs before the Elixir tests. The existing `CI required` aggregate requires the new job on every full plan and rejects failures, cancellations and unexpected skips.

This is the first job-extraction unit for #1413. Per-language path routing, the remaining SDK split, minimum-runtime expansion, dispatch support and the dedicated SDK aggregate remain follow-ups. All SDK checks still run. Base: main.

Validation:
- All 23 CI policy tests passed, including supported PR, merge-group and main plans and failures/skips of every required job.
- A structured YAML comparison confirms all ten Elixir setup/check steps are preserved exactly and removed from the shared job.
- Workflow validation passes with shellcheck disabled. The five shellcheck findings from the complete linter are identical on main and this change; none is in the moved steps.
- [GitHub CI required](https://github.com/managoat/fountain/actions/runs/34639880793) passed, including the extracted Elixir job.

Timing and runner jobs:
- A full mixed PR plan grows from 19 to 20 jobs; the new job can run in parallel, but this unit does not claim lower total runner usage.
- Baseline [run 34637353550](https://github.com/managoat/fountain/actions/runs/34637353550/job/103388442744): the combined SDK/CLI job took 78 seconds. That application-only PR executed 18 jobs; this CI-file change also selects the docs-prose job.
- After extraction, [Elixir SDK](https://github.com/managoat/fountain/actions/runs/34639880793/job/103396732250) passed in 22 seconds and the [remaining SDK/CLI job](https://github.com/managoat/fountain/actions/runs/34639880793/job/103396732309) passed in 68 seconds. These are single-run observations with different runners/cache conditions, not a guaranteed speedup. The two job durations sum to 90 seconds versus 78 seconds before; the split prepares selective routing rather than reducing total runner time by itself.

- Overall observed workflow time, from run start to the last job completion: 266 seconds before, 252 seconds after. Executed jobs were 18 before and 20 after: the new SDK job plus the docs-prose job selected by this CI-file change.

Queue settings are unchanged; their comments now describe the 20-job full plan.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -815,8 +815,8 @@ jobs:
             exit 1
           fi
 
-  sdk-clients:
-    name: SDK clients, CLI and plugins
+  elixir-sdk:
+    name: Elixir SDK
     runs-on: ubuntu-24.04
     timeout-minutes: 15
     needs: [already-tested, changes]
@@ -826,18 +826,15 @@ jobs:
     permissions:
       contents: read
 
-    # No umbrella deps/_build cache and no Postgres. Everything here either is
-    # a standalone project (sdk/elixir has its own cache below, cli/ and
-    # apps/fountain_buzz/cli are Go modules) or reads the COMMITTED
-    # sdk/contract/contract.json rather than rebuilding it -- which is the
-    # whole reason that projection is committed. The one job that does rebuild
-    # it is `release-and-contract`, and it fails there if it is stale.
+    # Independent from the umbrella, Postgres and the other SDK toolchains.
     env:
       MIX_ENV: test
 
     steps:
-
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v5
+
+      - name: SDK conformance scenarios are well formed
+        run: python3 sdk/conformance/lint.py
 
       - name: Set up Elixir
         id: beam
@@ -847,10 +844,7 @@ jobs:
           otp-version: "28.3"
 
       # The Elixir SDK is a standalone Mix project. Keep its dependency graph
-      # and build artifacts separate from this umbrella's -- which also puts
-      # them outside the two caches above, so they need their own entry or
-      # every PR recompiles finch, mint, hpax, ex_doc and makeup from scratch
-      # on the critical path.
+      # and build artifacts separate from the umbrella's cache.
       - name: Restore Elixir SDK cache
         uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v5
         with:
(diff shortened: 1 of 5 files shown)


BinaryBourbon (approved):
**APPROVE** — the Elixir SDK steps move out of `sdk-clients` into a new `elixir-sdk` job, and the split is lossless: the union of steps across the two jobs is identical to the old job's, and the new job is registered in both halves of the gate.

Nothing blocking. Four documentation points, all worth a follow-up commit on this PR.

### 1. ADR 0050 still says a full run is 19 jobs

`decisions/0050-a-merge-queue-tests-what-it-merges.md:42` — "One full CI run is 19 of them."

This PR bumps that same fact from 19 to 20 in three places (`scripts/ci/require-checks.py:22` and `:29`, `scripts/ci/test_require_checks.py:78`, `scripts/ci/README.md:74`) and leaves the ADR, which is where the reasoning originates and which the `require-checks.py` comment restates. After this PR the ADR is the one copy that disagrees, and it is the copy `okf` tracks for staleness. Same edit, one file short.

### 2. `CLAUDE.md`'s CI table no longer describes the pipeline

`CLAUDE.md:424` still reads: `**sdk-clients** | The four SDKs (Elixir, TypeScript, Python, and Swift in its own swift-sdk job) ...`. After this PR `sdk-clients` runs two SDKs, not four, and there is a job in the table's grouping ("the job name in a red build tells you which toolchain to go and look at") that the table does not list. A red `Elixir SDK` build now points at a row that does not exist. CLAUDE.md's own opening line asks for this to be kept current.

### 3. The new comment credits the Swift job with a lint it does not run

`.github/workflows/ci.yml:908`: "The Elixir and Swift jobs validate them independently."

`elixir-sdk` does run `python3 sdk/conformance/lint.py` (line 837). `swift-sdk` does not — its steps are version check, `swift format lint`, the two `--filter` test runs, the full debug test and the release build; nothing invokes `sdk/conformance/lint.py`. The fixture-schema validation the lint performs is therefore not duplicated on the Swift side. Either drop "and Swift" or add the step there.

### 4. Stack note: the run now sits exactly on the free-plan ceiling

`scripts/ci/README.md:74` now says the run "has 20 jobs, matching the documented concurrent runner limit" — which is accurate under the existing convention (22 job instances minus `coverage` and `gate`, which run strictly after the fan-out), but it means the next PR in this stack that adds a job puts a single full run over 20. ADR 0050 records what that costs (a burst where jobs waited 29 minutes to start). Not this PR's problem, but the sizing note and `max_entries_to_build` reasoning should be revisited at the point in #1413 where the job count actually crosses it, rather than left as a line that was true for one commit.

### Checked

- Step-for-step diff of the old `sdk-clients` against the new `elixir-sdk` + `sdk-clients`, comparing parsed YAML step names and `run:` bodies. Every one of the 32 old steps is present exactly once, except `actions/checkout` and `SDK conformance scenarios are well formed`, which each appear in both jobs by design. No step, `working-directory` or flag changed.
- The dropped `env: MIX_ENV: test` on `sdk-clients` affects nothing: the steps that remain there are Go, Python and Node only, and `elixir-sdk` keeps it.
- `if:` on `elixir-sdk` is identical to `sdk-clients`, so both skip on `docs_only` and on a reused tree together — which is what `gate.py` expects, since `elixir-sdk` is in `FULL_JOBS` and `validate()` asserts exact result strings (`success`/`skipped`), not `!= 'failure'`. A skipped-but-expected-success `elixir-sdk` fails the gate.
- `gate.py`'s `set(needs) != JOBS` check plus `test_gate.py::test_workflow_and_gate_cover_the_same_jobs` bind the workflow `needs:` list (line 1798) and `FULL_JOBS` together; both were updated. Ran `python3 -m unittest discover -s scripts/ci -p 'test_*.py'` against the head tree: 23 tests, OK.
- Grepped the whole head tree for `sdk-clients` / "SDK clients": only ci.yml, `gate.py`, `scripts/ci/README.md` and the stale `CLAUDE.md` row. No ruleset context change is needed — `require-checks.py` publishes `["CI required", "Detect secrets"]`, not per-job contexts.
- `permissions: contents: read`, action pins, `runs-on` and `timeout-minutes` on the new job all match the job it was split from.

Not verified: I read YAML and ran the `scripts/ci` unit tests; I did not run the workflow, so the new job's actual runtime and cache hit rate are unmeasured.

jhgaylor:

PR #1985, opened 2026-09-11
ci: run the Python SDK checks in a separate job

jhgaylor:
The Python SDK's four checks now run in a separate `python-sdk` job, parallel with the TypeScript, CLI and plugin job. It keeps the same runner and commands, uses the runner's Python as before, and validates conformance fixtures before the tests. `CI required` requires the new job on every full plan and rejects failures, cancellations and unexpected skips.

Stack: based on #1984. Refs #1413. This is the Python job-extraction unit; path routing, minimum-runtime expansion and the dedicated SDK aggregate remain follow-ups. All SDK checks still run.

Validation:
- All 23 CI policy tests pass. A structured YAML comparison confirms all four Python check steps are preserved exactly and removed from the shared job.
- All 43 Python SDK tests pass, including conformance. Fixture validation covers 24 scenarios; the wire-contract check covers 67 operations, 41 schemas and 4 enums. Compilation passes.
- Workflow validation passes with shellcheck disabled; changed documentation has no new repository-style findings and passes destink.

Timing and runner jobs:
- A full mixed PR plan grows from 20 to 21 jobs. This split prepares independent routing; it does not claim lower total runner usage.
- Parent [run 34639880793](https://github.com/managoat/fountain/actions/runs/34639880793) completed in 252 seconds with 20 executed jobs; its remaining SDK/CLI job took 68 seconds. After extraction, [Python SDK](https://github.com/managoat/fountain/actions/runs/34640688693/job/103399363194) passed in 22 seconds and the [remaining TypeScript/CLI job](https://github.com/managoat/fountain/actions/runs/34640688693/job/103399363415) passed in 43 seconds. The successful workflow took 267 seconds with 21 executed jobs. These are single-run observations with different runner/cache conditions.

Queue settings are unchanged; their comments now describe the 21-job full plan.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -887,8 +887,41 @@ jobs:
         working-directory: sdk/elixir
         run: mix test test/conformance_test.exs
 
+  python-sdk:
+    name: Python SDK
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
+    needs: [already-tested, changes]
+    if: ${{ !cancelled() && needs.already-tested.outputs.skip != 'true'
+         && needs.changes.outputs.docs_only != 'true' }}
+
+    permissions:
+      contents: read
+
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v5
+
+      - name: SDK conformance scenarios are well formed
+        run: python3 sdk/conformance/lint.py
+
+      # The Python SDK is deliberately stdlib-only, so its source tests run
+      # on the same interpreter without an install step.
+      - name: Python SDK tests
+        run: python3 -m unittest discover -s sdk/python/tests -v
+
+      - name: Python SDK compiles
+        run: python3 -m compileall -q sdk/python/src sdk/python/scripts
+
+      - name: Python SDK contract
+        working-directory: sdk/python
+        run: python3 scripts/verify_contract.py
+
+      - name: Python SDK conformance
+        working-directory: sdk/python
+        run: python3 -m unittest discover -s tests -p "test_conformance.py" -v
+
   sdk-clients:
-    name: Python/TypeScript SDKs, CLI and plugins
+    name: TypeScript SDK, CLI and plugins
     runs-on: ubuntu-24.04
     timeout-minutes: 15
     needs: [already-tested, changes]
@@ -904,8 +937,8 @@ jobs:
 
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v5
 
-      # Validate fixture bodies against the committed contract before either
-      # SDK adapter runs. The Elixir and Swift jobs validate them independently.
+      # Validate fixtures before the TypeScript adapter runs. The other SDK
+      # jobs validate them independently.
       - name: SDK conformance scenarios are well formed
         run: python3 sdk/conformance/lint.py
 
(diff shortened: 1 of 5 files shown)


BinaryBourbon (approved):
**APPROVE** — extracts the four Python SDK steps out of `sdk-clients` into a new `python-sdk` job, registers it in the `CI required` gate, and updates the job-count prose.

No blocking defect found.

The extraction is complete and the gate knows about it:

- All four Python steps in the base `sdk-clients` job (`.github/workflows/ci.yml:973-991` at base) reappear verbatim in `python-sdk` (`.github/workflows/ci.yml:890-919`), and `sdk/conformance/lint.py` is duplicated rather than moved, so both jobs still validate fixtures before their adapters run. Nothing in the removed block was left without a home; `python3` usage that stays behind (`lint.py`, Hermes) is untouched.
- `python-sdk` copies `runs-on`, `timeout-minutes`, `needs: [already-tested, changes]`, the `!cancelled() && skip != 'true' && docs_only != 'true'` guard, `permissions: contents: read` and the pinned `actions/checkout@3d3c42e` from `sdk-clients`, so it runs on exactly the plans the Python steps used to run on. No `setup-python`, matching the base job's reliance on the runner's `python3` — behaviour is unchanged.
- The gate cannot pass while this job does nothing: `python-sdk` is in both `gate`'s `needs:` (`ci.yml:1815`) and `FULL_JOBS` (`scripts/ci/gate.py:12`), and `gate.py:validate` asserts `set(needs) == JOBS` plus an exact per-job expected `result`, so a skipped or cancelled `python-sdk` on a full plan fails `CI required` rather than sliding through a `!= 'failure'` test.
- `test_gate.py::test_workflow_and_gate_cover_the_same_jobs` pins the workflow's job set and the gate's `needs:` list against `JOBS`; I ran the full `scripts/ci` suite against this tree's files (23 tests, OK).

Two non-blocking notes:

1. `CLAUDE.md:424` still describes **sdk-clients** as running "The four SDKs (Elixir, TypeScript, Python…)". That row went stale in #1984 and is now stale twice over; CLAUDE.md's own header says stale guidance misleads every downstream dispatch. Fixing it once at the end of the stack (when the final job layout is known) is fine, but it should not be forgotten.
2. `scripts/ci/README.md:74-87` — the honest "21 jobs against a documented limit of 20 concurrent runners" is the right call to surface, but the edited paragraph leaves two over-long reflowed lines next to otherwise hard-wrapped prose. Cosmetic.

Checked:
- Full base-vs-head diff; every removed step accounted for in the new job.
- `yaml.safe_load` of the head `ci.yml`; job set = 17 definitions, 23 expanded legs, and 21 on a full non-docs plan (docs + docs-prose skipped) — which is consistent with the 20 the README claimed before this job existed, so the arithmetic in `require-checks.py` and its test docstring is right.
- `python3 -m unittest discover -s scripts/ci` against head files: 23 tests, OK.
- `actionlint` on the head `ci.yml`: the only findings are pre-existing SC2010/SC2012 warnings in the compose jobs (lines 1327/1359/1479), none in the new job.
- Grepped the tree at head for the old job display name and for other registries of CI job names (`.github/`, `scripts/`, docs): the merge-queue required contexts are `["CI required", "Detect secrets"]`, so no branch-protection context needs adding.

NOT verified: I read YAML, I did not run CI. I did not execute the Python SDK steps themselves, so I am asserting that they run in the new job's environment only from the fact that the environment (`ubuntu-24.04`, no `setup-python`, same checkout) is identical to the one they ran in before.

jhgaylor:

PR #1986, opened 2026-09-11
ci: run the TypeScript SDK checks in a separate job

jhgaylor:
The TypeScript SDK's eight setup/check steps now run in `typescript-sdk`, parallel with the other SDKs. Its Node version, npm cache and commands are preserved. The remaining `cli-plugins` job keeps the Go CLIs, Hermes plugin and deployed-runner tests; it still installs Node for the deployed runner but no longer restores an SDK npm cache. The three extracted SDK jobs lint conformance fixtures independently; Swift retains its existing conformance tests.

`CI required` requires the new job and the renamed CLI job on every full plan, rejecting failures, cancellations and unexpected skips. Required branch checks are the stable aggregate and Detect secrets, so the old shared job name is not a branch-protection requirement.

Stack: based on #1985. Refs #1413. This completes the separation into one job per SDK; path routing, minimum-runtime expansion, dispatch support and the dedicated SDK aggregate remain follow-ups. All SDK checks still run.

Validation:
- All 23 CI policy tests pass. Structured YAML comparison confirms all eight TypeScript setup/check steps are preserved exactly; deployed-runner tests retain their Node setup in the CLI job.
- On Node 24.21.0: 109 tests passed; type checking, build, wire-contract verification, all 24 conformance scenarios and browser bundling passed.
- Workflow validation passes with shellcheck disabled. Changed documentation has no new repository-style findings and passes destink.

Timing and runner jobs:
- A full mixed PR plan grows from 21 to 22 jobs. The split prepares selective routing; it does not claim lower total runner usage.
- Parent [run 34640688693](https://github.com/managoat/fountain/actions/runs/34640688693) completed in 267 seconds with 21 executed jobs; its remaining TypeScript/CLI job took 43 seconds. After extraction, [TypeScript SDK](https://github.com/managoat/fountain/actions/runs/34641315620/job/103401470332) passed in 18 seconds and [CLI/plugins](https://github.com/managoat/fountain/actions/runs/34641315620/job/103401469970) passed in 39 seconds. The successful workflow took 255 seconds with 22 executed jobs. These are single-run observations with different runner/cache conditions.

Queue settings are unchanged; their comments now describe the 22-job full plan.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -792,7 +792,7 @@ jobs:
           cache: npm
           cache-dependency-path: sdk/typescript/package-lock.json
 
-      # `--no-audit --no-fund`: see the sdk-clients job, which carries the
+      # `--no-audit --no-fund`: see the typescript-sdk job, which carries the
       # measurement. Here it is only to make `npm run generate` runnable.
       - name: SDK install
         working-directory: sdk/typescript
@@ -920,8 +920,8 @@ jobs:
         working-directory: sdk/python
         run: python3 -m unittest discover -s tests -p "test_conformance.py" -v
 
-  sdk-clients:
-    name: TypeScript SDK, CLI and plugins
+  typescript-sdk:
+    name: TypeScript SDK
     runs-on: ubuntu-24.04
     timeout-minutes: 15
     needs: [already-tested, changes]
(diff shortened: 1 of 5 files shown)


BinaryBourbon (approved):
**APPROVE** — splits `sdk-clients` into `typescript-sdk` (Node/SDK legs) and `cli-plugins` (Go modules, gofmt, Hermes, deployed-runner tests), and registers both in the gate.

No step is dropped. I parsed both files and diffed the step lists by (name/uses, working-directory, run body, with-block): the union of `typescript-sdk` + `cli-plugins` at HEAD contains every step of `sdk-clients` at BASE, with exactly one difference — `cli-plugins`' `Set up Node` drops `cache: npm` / `cache-dependency-path`. That is correct: the only Node work left in that job is `node --test deployed/test/*.test.mjs`, and every file under `deployed/test/` imports `node:*` or a relative `../lib/*.mjs`, so nothing is installed there. In BASE the `npm ci` ran *after* the deployed tests anyway, so those tests never had dependencies available either.

Both jobs inherit BASE's `runs-on`, `timeout-minutes: 15`, `needs: [already-tested, changes]`, `if:` and `permissions: {contents: read}` unchanged.

The two checks the extraction could plausibly have orphaned both survive, and neither was ever in this job:
- `npm run generate` + `git diff --exit-code src/generated/openapi.ts` lives in `release-and-contract` (.github/workflows/ci.yml:800-816 at HEAD) and is untouched; the comment reference there was correctly repointed to `typescript-sdk`.
- the SDK release guard is its own workflow (`.github/workflows/sdk-release-gate.yml`), not a step of this job.

Gate wiring is right, and it fails closed rather than silently: `scripts/ci/gate.py:11` adds both names to `FULL_JOBS`, and the gate job's `needs` (.github/workflows/ci.yml:1822-1838) adds both. `validate()` asserts `set(needs) != JOBS` as an exact set — 17 names on both sides — so a name added in one place and not the other raises rather than passing, and on a full plan each job must be literally `"success"`, not merely not-failure. A skipped `typescript-sdk` or `cli-plugins` fails `CI required`.

Nits, neither blocking:

1. `CLAUDE.md:424` still documents a **sdk-clients** row ("The four SDKs ... both Go modules ... and the Hermes plugin") for a job that no longer exists after this commit. It was already stale at BASE (#1984/#1985 added `elixir-sdk`/`python-sdk` without touching it), so this is the stack's debt rather than this PR's, but the file's own header says stale guidance misleads every downstream dispatch — worth one commit somewhere in #1984..#1995 rewriting that row into the per-SDK job names.

2. `scripts/ci/README.md:88` now generalises BASE's accurate "Both jobs validate fixtures before their tests" (Elixir + Python) to "Each SDK job validates fixtures before its tests". `swift-sdk` runs `Swift SDK conformance` but no `sdk/conformance/lint.py` step, so the sentence overstates by one job. "The Elixir, Python and TypeScript jobs validate fixtures before their tests" would be exact.

Checked:
- step-by-step set diff BASE `sdk-clients` vs HEAD `typescript-sdk` + `cli-plugins` (parsed YAML, not eyeballed diff)
- job-level keys (`runs-on`, `if`, `needs`, `permissions`, `timeout-minutes`) preserved on both halves
- `gate.py` `FULL_JOBS`/`JOBS` vs the gate job's `needs` list — 17 == 17, exact-set assertion holds
- no other reference to `sdk-clients` in the tree at HEAD except `CLAUDE.md:424`
- `actionlint` on BASE and HEAD `ci.yml`: byte-identical finding sets (5 pre-existing shellcheck notes, all in `core-distribution`); no new shell issues
- job-count bump 21 -> 22 in `scripts/ci/README.md`, `require-checks.py` and the `test_require_checks.py` docstring is a consistent +1; the docstring carries no numeric assertion, so nothing there can go stale silently
- `deployed/test/*.test.mjs` imports, to confirm the cacheless Node setup is safe

Not checked: I read YAML, I did not run CI. I did not time the two new jobs against the 20-concurrent-runner ceiling the README sizing section describes, beyond confirming the documented count was incremented.

jhgaylor:

PR #1987, opened 2026-09-11
ci: report a stable SDK aggregate check

jhgaylor:
CI now reports a stable `SDK checks` aggregate after the four SDK jobs, including when docs-only classification or verified tree reuse skips every SDK leg. It validates the selected job plan and refuses missing results, failures, cancellations and unexpected skips. `CI required` depends on the aggregate and continues validating the complete plan independently.

Both gates share probe validation. The SDK-only CLI mode never writes `tested-tree.txt`; only a complete successful CI plan can publish evidence for reuse. Null checkout-tree outputs now produce the existing invalid-tree diagnostic.

Stack: based on #1986. Refs #1413. Path routing, minimum-runtime expansion and workflow dispatch remain follow-ups. Required branch rules are unchanged.

Validation:
- All 30 CI policy tests passed, covering PR, merge-group and main plans; docs-only and reused trees; invalid probes; failed/cancelled/skipped jobs; exact workflow dependencies; and the actual CLI exit/evidence behavior.
- A mutation that ignored SDK results produced 169 failing subcases; restoring validation returned the full policy suite to green.
- Workflow validation passes with shellcheck disabled. Changed documentation has no new repository-style findings and passes destink.

Timing and runner jobs:
- A full mixed PR plan grows from 22 to 23 jobs. Docs-only and reused plans also run this small aggregate so the stable result is always reported.
- Parent #1986 completed CI in 255 seconds with 22 executed jobs. [This draft’s successful run](https://github.com/managoat/fountain/actions/runs/34642601251) took 273 seconds with 23 executed jobs; [SDK checks](https://github.com/managoat/fountain/actions/runs/34642601251/job/103406486608) took 6 seconds. These are single-run observations with different cache conditions, not a guaranteed speedup.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1814,6 +1814,27 @@ jobs:
       - name: Test CI decisions
         run: python3 -m unittest discover -s scripts/ci -p 'test_*.py' -v
 
+  sdk-checks:
+    name: SDK checks
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    if: ${{ always() }}
+    needs:
+      - already-tested
+      - changes
+      - elixir-sdk
+      - python-sdk
+      - typescript-sdk
+      - swift-sdk
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v5
+        with:
+          persist-credentials: false
+      - name: Every selected SDK check passed
+        env:
+          CI_NEEDS: ${{ toJSON(needs) }}
+        run: python3 scripts/ci/gate.py --sdk
+
   gate:
     name: CI required
     runs-on: ubuntu-24.04
@@ -1823,6 +1844,7 @@ jobs:
       - already-tested
       - changes
       - workflow-checks
+      - sdk-checks
       - test
       - coverage
       - elixir-static
--- a/scripts/ci/README.md
+++ b/scripts/ci/README.md
@@ -71,7 +71,7 @@ tested it.
 ### Sizing
 
 `MERGE_QUEUE` in `require-checks.py` explains the queue settings. A full CI
-run now has 22 jobs, compared with the documented limit of 20 concurrent
+run now has 23 jobs, compared with the documented limit of 20 concurrent
 runners. The queue builds one group at a time and batches up to five PRs. Revisit
 `max_entries_to_build` when the concurrency limit changes.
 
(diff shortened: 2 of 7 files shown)


BinaryBourbon (approved):
**APPROVE** — adds an `sdk-checks` aggregate job that runs `gate.py --sdk` over the four per-SDK jobs plus both probes, and makes `CI required` depend on it.

No blocking defect. I went looking specifically for the "gate green while gating nothing" failure and could not construct it.

## Why the aggregate is not a hollow gate

`validate_sdks` (`scripts/ci/gate.py:77`) builds its expectation from `_expected_plan`, which starts every job at `"skipped"` and promotes only the probes the event runs, then promotes the four SDK jobs to `"success"` when `full` (`scripts/ci/gate.py:79`). `_check_results` (`scripts/ci/gate.py:84`) compares with `!=`, so `skipped`, `cancelled`, `failure` and a missing `result` are all rejected where `success` is owed — the `!= 'failure'` trap the repo has been bitten by is not present. Two further properties hold:

- `full` is derived from the same expressions the SDK jobs gate on. The four jobs run under `!cancelled() && needs.already-tested.outputs.skip != 'true' && needs.changes.outputs.docs_only != 'true'` (`.github/workflows/ci.yml:823`, `:895`, `:928`, `:1192`), and `full = not reuse and not docs_only`. A leg that skips when it should not run, or runs when it should have skipped, is a mismatch either way.
- A probe that produced no decision fails closed. `_reuse` rejects a missing/invalid `skip`; `_classification` rejects missing `docs_only`/`docs_touched`/`cli_docs` and a non-40-hex `tree`. So a cancelled or failed `changes`/`already-tested` job cannot silently authorize skipping the SDKs. The `isinstance(tree, str)` addition at `scripts/ci/gate.py:41` is what keeps a `None` output a `ValueError` rather than a `TypeError`.

`if: ${{ always() }}` (`.github/workflows/ci.yml:1821`) is required for the job to report at all when every leg skipped, and `validate` at `scripts/ci/gate.py:65` then demands `sdk-checks == "success"` in *every* plan — including reuse and docs-only — so the aggregate cannot itself go missing without failing `CI required`. `tested-tree.txt` stays behind the non-`--sdk` branch (`scripts/ci/gate.py:97`), so the SDK gate cannot mint reuse evidence.

## Observations (non-blocking)

1. **The aggregate is redundant with `gate` today.** `FULL_JOBS` already contains all four SDK jobs and `gate` already `needs` them directly, so nothing this job rejects would have passed `CI required` anyway. Its value is a stable *reported check name* for when the later PRs in #1413 add per-language path routing — worth stating in the README, because as written ("reports the SDK result even when ... skips every SDK leg") a reader can mistake green for "the SDKs were tested". It is not: on a reuse or docs-only plan, green means "the four legs were correctly skipped". The cost is one more concurrent job against a ceiling this same PR documents as 23 vs 20 (`scripts/ci/require-checks.py:22`).

2. **`scripts/ci/test_sdk_gate.py:32` parses the workflow positionally.** `split("\n  sdk-checks:\n")[1].split("\n  gate:\n")[0]` assumes `sdk-checks` is declared immediately above `gate`. Insert a job between them and `assertIn("    if: ${{ always() }}\n", job)` can be satisfied by the *inserted* job's `if:`. It fails loudly rather than silently in the cases that matter (the `needs:`/`SDK_GATE_JOBS` equality still keys off `sdk-checks`' own block), so this is a robustness nit, not a hole.

3. **`CLAUDE.md:424`** still describes a single **sdk-clients** job and has no row for `SDK checks`. That is debt this stack inherited from #1984–#1986 rather than something this PR introduced, but the table is now wrong in two ways and something in the stack should fix it.

No CHANGELOG entry is owed — this is CI-only and reads no new runtime env var.

## Checked

- Ran `python3 -m unittest discover -s scripts/ci -p 'test_*.py'` against a `git show`-extracted copy of the PR tree: 30 tests, OK.
- Mutation-tested the new guards against that copy, each caught:
  - `_check_results` relaxed to flag only `"failure"` → 321 failures;
  - `validate_sdks` made a no-op → 221 failures;
  - `sdk-checks` removed from `gate`'s `needs` → 1 failure;
  - `swift-sdk` removed from `sdk-checks`' `needs` → 1 failure;
  - `if: ${{ always() }}` removed from `sdk-checks` → 1 failure;
  - `expected["sdk-checks"] = "success"` dropped from `validate` → 16 failures/errors.
- `yaml.safe_load` on the HEAD `ci.yml`.
- Job-count arithmetic: a full PR run is 21 single jobs + `test`×6 + `swift-sdk`×2 = 23, so the 22→23 edits in `README.md`, `require-checks.py` and `test_require_checks.py` are right.
- `require-checks.py` binds only `CI required` and `Detect secrets` as required contexts, so the new job's name is not load-bearing in branch protection and cannot block merges by never reporting.
- README claim that three jobs lint fixtures and Swift does not: matches — `sdk/conformance/lint.py` runs in elixir/python/typescript only, Swift has its own `ConformanceTests` filter step.
- `CI_NEEDS: ${{ toJSON(needs) }}` is passed through `env:`, not interpolated into the shell; checkout is the same pinned SHA with `persist-credentials: false` as the rest of the workflow; the job inherits the workflow-level `contents: read`.

Not verified: I read YAML and ran the Python unit tests locally. I did not observe a real GitHub run, so GitHub's exact `needs.*.result` value for a cancelled `always()` job, and matrix-job result aggregation for `swift-sdk`, are taken from documented behaviour rather than measured here.

jhgaylor:

PR #1988, opened 2026-09-11
ci: classify SDK paths before enabling selective jobs

jhgaylor:
CI now reports a separate selection for each SDK using the existing PR or merge-group diff base. An SDK directory, its documentation page or registered release tooling selects that language. Shared contract/conformance, API implementation, build configuration and unknown paths select all four. An explicit allowlist selects none for unrelated docs, console UI, server tests and telemetry.

This prerequisite reports decisions without changing SDK job execution. The following draft will apply them in job conditions and both aggregate gates. Keeping classification separate makes the allowlist and failure behavior reviewable on their own.

The Git diff uses NUL-delimited paths and disables rename detection so moving a file out of an SDK still selects its former owner. Missing/invalid bases, unreadable or empty diffs, malformed paths and undecodable names select all SDKs. Only fixed boolean outputs are emitted.

Stack: based on #1987. Refs #1413. Selective execution, minimum-runtime expansion and workflow dispatch remain follow-ups. No release or branch-rule changes.

Validation:
- All 37 CI policy tests passed, including real Git renames, deletions, invalid bases and filenames containing an attempted output injection.
- Enabling rename detection in an isolated test copy made the rename regression fail.

Timing and runner jobs:
- Job selection and the maximum full mixed PR plan remain unchanged at 23 jobs; this adds one small classification step to the existing changes job.
- Parent #1987 completed CI in 273 seconds with 23 executed jobs. [This draft’s successful run](https://github.com/managoat/fountain/actions/runs/34643756044) took 242 seconds with 23 executed jobs. The SDK classification step completed within the same reported second in [the changes job](https://github.com/managoat/fountain/actions/runs/34643756044/job/103409482813). Single-run comparisons have different cache conditions and do not establish a speedup.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
         run: bash scripts/ci/already-tested.sh
 
   changes:
-    name: Decide whether this is a docs-only change
+    name: Classify documentation and SDK changes
     # Never on push, and deliberately so. There, `already-tested` above
     # already collapses a rebased squash-merge to a few seconds, and a main
     # push that is NOT already-tested is precisely the case that must run
@@ -90,6 +90,10 @@ jobs:
       docs_touched: ${{ steps.filter.outputs.docs_touched }}
       cli_docs: ${{ steps.filter.outputs.cli_docs }}
       tree: ${{ steps.tree.outputs.tree }}
+      sdk_elixir: ${{ steps.sdks.outputs.sdk_elixir }}
+      sdk_python: ${{ steps.sdks.outputs.sdk_python }}
+      sdk_typescript: ${{ steps.sdks.outputs.sdk_typescript }}
+      sdk_swift: ${{ steps.sdks.outputs.sdk_swift }}
 
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v5
@@ -120,8 +124,10 @@ jobs:
           MERGE_GROUP_BASE_SHA: ${{ github.event.merge_group.base_sha }}
         run: |
           set -u
+          base=""
           decide() {
             {
+              echo "diff_base=${base:-}"
               echo "docs_only=$1"
               echo "cli_docs=$2"
               echo "docs_touched=$3"
@@ -192,6 +198,12 @@ jobs:
 
           echo "docs-only — running the docs job instead of the Elixir suite"
           decide true "${cli_docs}" false
+
+      - name: Select SDKs from the same diff
+        id: sdks
+        env:
+          DIFF_BASE: ${{ steps.filter.outputs.diff_base }}
+        run: python3 scripts/ci/sdk_changes.py "$DIFF_BASE" >> "$GITHUB_OUTPUT"
 
   test:
     name: Build and Test (partition ${{ matrix.partition }})
--- a/scripts/ci/README.md
+++ b/scripts/ci/README.md
@@ -103,6 +103,28 @@ SDK job, also update `SDK_JOBS` and the `sdk-checks` dependencies. Run
 `test_gate.py` and `test_sdk_gate.py` to verify their agreement and every
 supported event plan.
 
+## SDK path classification
+
+`changes` reports four `sdk_<language>` outputs from `sdk_changes.py`, using
+its existing PR merge base or merge-group base. Job conditions do not yet
+consume these outputs; SDK execution stays unchanged in this prerequisite.
+
+The classifier selects an SDK for its directory, documentation page or
+registered release tooling. Shared contract and conformance files select all
+SDKs. So do API implementation, build configuration and unregistered paths.
+An explicit allowlist selects none for unrelated docs, console UI, server
+tests and telemetry. Mixed changes select the union of their SDKs.
+
+Invalid bases, failed or empty diffs, malformed paths and undecodable names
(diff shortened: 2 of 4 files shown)


BinaryBourbon (approved):
**APPROVE** — adds `scripts/ci/sdk_changes.py`, wires it into `changes` as a fourth output group, and leaves every SDK job's `if:` untouched, so this lands inert as advertised.

I went at this specifically looking for a classifier that can answer "no SDKs" when it does not actually know. I could not construct one. Every branch I traced is fail-open:

- `scripts/ci/sdk_changes.py:53` — a base that is not a 40-char hex OID returns every SDK. That covers the two workflow branches that `decide` with `base` still empty (`.github/workflows/ci.yml:150` fetch failure, `:156` no merge base), since `diff_base` is written as `${base:-}`.
- `scripts/ci/sdk_changes.py:61,64` — an empty diff, a diff whose `-z` stream is truncated, undecodable bytes, and a `git diff` that exits nonzero all return every SDK. That is the merge-group case where `MERGE_GROUP_BASE_SHA` passes the hex check but is not in the checkout: `git cat-file -e` already `decide`s there, and the SHA it hands on makes `git diff` fail, so both halves land on "run everything".
- `scripts/ci/sdk_changes.py:42-46` — anything not under `sdk/<lang>/`, not in `OWNED_FILES`, and not on the explicit unrelated allowlist returns every SDK. `sdk/contract/`, `sdk/conformance/`, `scripts/sdk-contract/`, `mix.exs`, `mix.lock`, `config/`, the router, the schemas and `.github/workflows/ci.yml` all hit that branch, which is the direction the brief asked for. I confirmed the only repo-root files the four SDK jobs read outside their own tree are `sdk/contract/`, `sdk/conformance/` and `mix.exs` (the swift job's version-lockstep at `.github/workflows/ci.yml:1225`), and all three fan out to all four.
- `workflow_dispatch` and `push` skip the `changes` job entirely, so the outputs are empty strings rather than `false`. Inert today; see the note below for #1989.

Shallow checkout is not reachable here — the `changes` job already checks out with `fetch-depth: 0` — and there is no API pagination to get wrong, since the classifier shells out to local `git` rather than the files endpoint, so the >300-file case is a non-issue.

The tests assert on values, not exit status: `classify` is compared against explicit sets, and `test_actual_git_diff_handles_deletions_renames_and_output_safety` builds a real repo, runs the script as a subprocess, parses `GITHUB_OUTPUT` into a set and asserts the key set and the value domain. I ran the file (7 tests, all pass) and confirmed `workflow-checks` (`.github/workflows/ci.yml:1827`) discovers it via `unittest discover -s scripts/ci -p 'test_*.py'` and that `workflow-checks` is in the `gate` job's `needs`, so the new tests are actually gated rather than decorative.

The output-injection fixture is the right test to have written: a changed file literally named `docs/file\nsdk_python=false` is rejected by the control-character check at `scripts/ci/sdk_changes.py:43` and fans out, and the assertion on the parsed key set proves no attacker-chosen line reaches `GITHUB_OUTPUT`. `DIFF_BASE` is passed through `env:` rather than interpolated into the shell, consistent with the rest of the file.

Non-blocking notes for the PR that turns routing on (#1989), not for this one:

1. `scripts/ci/gate.py::_classification` validates `docs_only`, `cli_docs` and `docs_touched` are exactly `'true'`/`'false'` but does not validate the four `sdk_*` outputs. Once a job condition reads them, the condition must be `needs.changes.outputs.sdk_python != 'false'`, not `== 'true'` — on `push`, and on any run where `changes` failed after the `filter` step, the output is the empty string, and `== 'true'` would skip every SDK job while `sdk-checks` expected them skipped too. Extending `_classification` to the new keys is the cheap way to make that impossible.
2. `scripts/ci/sdk_changes.py:20` registers `Package.resolved` and `.swift-format`, neither of which exists at this SHA. Harmless forward registration (if they appear they select swift alone), but worth knowing they are covering nothing today.
3. `UNRELATED_PREFIXES` (`scripts/ci/sdk_changes.py:22`) lists `apps/fountain/test/`, `apps/fountain_buzz/test/` and `apps/fountain_support/test/` but not `ee/test/`, so an ee-test-only change fans out to all four SDKs. Fail-open and therefore fine; just an asymmetry if the savings matter later.

Checked: the full `changes` job as of this SHA, including every `decide` branch and what `diff_base` holds at each; `git diff <base> HEAD` semantics against the `pull_request` merge-commit checkout (merge-base is the base tip, so the diff is the PR's own changes, same base the docs classifier already trusts); YAML parses and the job declares all four outputs; no reference anywhere to the old job display name `Decide whether this is a docs-only change`, and the ruleset's required check is the `gate` job, so the rename is safe; `gate.py`'s `JOBS`/`SDK_JOBS` need no change because nothing was added or renamed at job-id level; the four SDK jobs' `if:` are byte-identical to base. Ran `test_sdk_changes.py` out of a scratch copy of the tree — 7 passed.

Not checked: I did not run CI, so I have not observed the step against a real `pull_request` or `merge_group` payload; the merge-group reasoning above is read off the YAML. I also did not audit every path in the repo against `UNRELATED_PREFIXES` exhaustively — I checked the ones the SDK jobs actually consume plus the OpenAPI spec's inputs, and confirmed no OpenApiSpex schema lives under the allowlisted `live/`, `components/`, `assets/` or test trees.

jhgaylor:

PR #1989, opened 2026-09-11
ci: run only SDKs selected by the change classifier

jhgaylor:
CI now runs only the SDKs selected by the change classifier. For example, a Python SDK edit runs Python, a console UI edit skips all SDK legs, and a public API or shared conformance change runs all four. SDK documentation selects its language even when the server uses the docs-only plan.

Both `SDK checks` and `CI required` independently validate every selection and exact job result. Missing or malformed decisions fail the gates; only explicit `false` skips an SDK job. Main still runs all SDKs unless the existing tested-tree proof authorizes reuse.

The release job also skips TypeScript dependency installation and type generation when TypeScript is unselected. OpenAPI validation and wire-contract generation/freshness checks remain required on every full server plan.

Stack: based on #1988. Refs #1413. Minimum-runtime expansion and workflow dispatch remain follow-ups. No release or branch-rule changes.

Validation:
- All 42 CI policy tests passed. Every SDK subset is checked on PR and merge-group code/docs/reuse plans, including invalid outputs and failed/cancelled/unexpectedly skipped jobs.
- Workflow assertions cover all four job conditions and the three TypeScript-only release steps; server contract generation remains unconditional within its job.
- A mutation that required every SDK regardless of selection caused 61 failing cases in the routing tests.

Timing and runner jobs:
- CI policy changes conservatively select all SDKs, so this draft retains the maximum full mixed PR plan of 23 jobs. Parent #1988 completed CI in 242 seconds with 23 executed jobs. [This draft’s successful run](https://github.com/managoat/fountain/actions/runs/34644486691) took 260 seconds with 23 executed jobs. Single-run comparisons have different cache conditions and do not establish a speedup.
- Replaying historical diffs through the classifier selects 0 SDK runner legs for the credential-form UI change (`81f8de00`, previously 5), 1 for a TypeScript dependency change (`7ed0d7fb`, previously 5), and all 5 for a public API change (`c730774e`). Swift has two OS legs. These are classifier replays, not measured GitHub skip runs; no runtime savings are inferred from them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -798,6 +798,7 @@ jobs:
       # which is also what enforces `erasableSyntaxOnly`, the setting that
       # keeps those sources runnable without a build step.
       - name: Set up Node
+        if: ${{ needs.changes.outputs.sdk_typescript != 'false' }}
         uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version: 24
@@ -807,6 +808,7 @@ jobs:
       # `--no-audit --no-fund`: see the typescript-sdk job, which carries the
       # measurement. Here it is only to make `npm run generate` runnable.
       - name: SDK install
+        if: ${{ needs.changes.outputs.sdk_typescript != 'false' }}
         working-directory: sdk/typescript
         run: npm ci --no-audit --no-fund
 
@@ -817,6 +819,7 @@ jobs:
       # API that drift apart. This is the arm that covers the WHOLE document;
       # the manifest checks below cover what each client reaches for by name.
       - name: SDK types match the spec
+        if: ${{ needs.changes.outputs.sdk_typescript != 'false' }}
         working-directory: sdk/typescript
         run: |
           npm run generate
@@ -833,7 +836,7 @@ jobs:
     timeout-minutes: 15
     needs: [already-tested, changes]
     if: ${{ !cancelled() && needs.already-tested.outputs.skip != 'true'
-         && needs.changes.outputs.docs_only != 'true' }}
+         && needs.changes.outputs.sdk_elixir != 'false' }}
 
     permissions:
       contents: read
@@ -905,7 +908,7 @@ jobs:
     timeout-minutes: 15
     needs: [already-tested, changes]
     if: ${{ !cancelled() && needs.already-tested.outputs.skip != 'true'
-         && needs.changes.outputs.docs_only != 'true' }}
+         && needs.changes.outputs.sdk_python != 'false' }}
 
     permissions:
       contents: read
@@ -938,7 +941,7 @@ jobs:
     timeout-minutes: 15
     needs: [already-tested, changes]
     if: ${{ !cancelled() && needs.already-tested.outputs.skip != 'true'
-         && needs.changes.outputs.docs_only != 'true' }}
+         && needs.changes.outputs.sdk_typescript != 'false' }}
 
     permissions:
       contents: read
@@ -1202,7 +1205,7 @@ jobs:
     timeout-minutes: 10
     needs: [already-tested, changes]
     if: ${{ !cancelled() && needs.already-tested.outputs.skip != 'true'
-         && needs.changes.outputs.docs_only != 'true' }}
+         && needs.changes.outputs.sdk_swift != 'false' }}
 
     permissions:
       contents: read
--- a/scripts/ci/README.md
+++ b/scripts/ci/README.md
@@ -88,8 +88,9 @@ tests, compilation, contract and conformance checks. The TypeScript job owns
 installation, type checks, tests, builds, browser bundling, contract and
 conformance checks. These three extracted jobs lint fixtures before their
 tests. Swift retains its separate conformance test step.
-`CI required` requires all SDK jobs on every full plan, including main and
-merge groups. A failed, cancelled or unexpectedly skipped job fails the gate.
+`CI required` requires every selected SDK job. A failed, cancelled or
+unexpectedly skipped job fails the gate. Main runs all SDKs unless a verified
+tested tree authorizes reuse.
 
 `SDK checks` reports the SDK result even when docs-only classification or a
 previously tested tree skips every SDK leg. It validates the same probe
@@ -97,7 +98,7 @@ outputs as the full gate and rejects missing, failed or unexpectedly skipped
 jobs. `CI required` depends on this aggregate. Only the full gate publishes
 `tested-tree` evidence; passing the SDK gate cannot authorize reuse of a tree.
 
-Per-language path routing remains tracked in #1413. Register a new job in
+Register a new job in
 `gate.py`'s `FULL_JOBS` and the workflow gate's `needs` list together. For an
 SDK job, also update `SDK_JOBS` and the `sdk-checks` dependencies. Run
 `test_gate.py` and `test_sdk_gate.py` to verify their agreement and every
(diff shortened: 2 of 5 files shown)


BinaryBourbon (approved):
**APPROVE.** Turns on selective SDK running: each SDK job's `if:` now reads its own `needs.changes.outputs.sdk_<language>`, `release-and-contract`'s three TypeScript steps become conditional, and both `gate.py` entry points require exactly the set the classifier selected.

I looked hard for the two failure shapes this repo keeps getting bitten by — a gate that passes while running nothing, and a skipped job read as a success — and did not find either. Details in "Checked". No blocking findings; three non-blocking notes.

## Non-blocking

**1. `.github/workflows/ci.yml:1202-1209` — a docs-only edit to `docs/swift-sdk.md` now starts a macOS runner that never reads the page.**

`swift-sdk` is a two-way matrix (`ubuntu-24.04`, `macos-15`). Before this PR its `if:` carried `docs_only != 'true'`, so a PR touching only `docs/swift-sdk.md` skipped it. After this PR, `docs/swift-sdk.md` is in `sdk_changes.OWNED_FILES["swift"]`, so `sdk_swift=true` and both runners start — and none of the job's steps (version check, `swift format lint`, `swift test`, `swift build`) reads `docs/`. Same shape for `docs/sdk.md` → `typescript-sdk` and `docs/elixir-sdk.md` → `elixir-sdk`. The doc page's own correctness is already covered by the `docs` job, which runs on exactly that path.

This is the one direction in which the PR *increases* spend, and it is a deliberate, documented choice (the `_expected_plan` comment, the README line "SDK docs select their language even on the server docs-only path", and the `["docs/swift-sdk.md"] -> {"swift-sdk"}` case in the new test). Flagging it only to confirm it is intended rather than a side effect of the doc-page entries in `OWNED_FILES`: since `docs/` is already in `UNRELATED_PREFIXES`, those entries can only ever add jobs, never prevent one from being skipped.

**2. `scripts/ci/test_gate.py:12-16` — `plan()` silently discards an explicit `sdks=` on `push`.**

```python
if sdks is None:
    sdks = set() if docs else SDK_JOBS
if event == "push":
    sdks = SDK_JOBS
```

Correct for every current caller (`test_sdk_routing` only ever calls `plan("push")` with no `sdks`). But a future `plan("push", sdks={"python-sdk"})` written to assert that main *cannot* select a subset would get all four and pass vacuously — the "gate green while checking nothing" shape, one level up. An `assert sdks is None` in that branch, or raising, would make it loud.

**3. `scripts/ci/README.md:110-114` — "Missing or malformed outputs fail both gates" is true only for `pull_request` and `merge_group`.**

On `push` the `changes` job is deliberately skipped, `"changes" not in PROBES["push"]`, `_classification` is never called and the absent `sdk_*` outputs are correctly *not* an error (`gate.py:64-66` falls back to `SDK_JOBS`). The sentence reads as unconditional. The next paragraph's "Main runs all SDKs unless a verified tested tree authorizes reuse" carries the real rule, so this is wording only. Also `scripts/ci/README.md:110` now wraps as a two-word line ("Register a new job in") after the `#1413` sentence was removed — worth a rewrap.

## Checked

- **Output plumbing (the "classifier yields empty" trap).** `changes.outputs` declares all four `sdk_elixir`/`sdk_python`/`sdk_typescript`/`sdk_swift` from `steps.sdks`, step id `sdks` matches, and `sdk_changes.py` prints exactly those four keys with `str(...).lower()`. Parsed the YAML and printed the four job `if:` strings plus the three step `if:` strings to confirm no typo'd key. `actionlint` is clean on the file (only pre-existing SC2010/SC2012 warnings at lines 1358/1390/1510, untouched here).
- **String-vs-boolean.** Every condition is `!= 'false'`, i.e. fail-*open*: an empty output (job skipped, job failed, output typo'd) runs the SDK, it does not skip it. That is the safe direction, and `gate.py:_classification` independently rejects anything that is not literally `"true"`/`"false"`, so a typo'd key cannot pass the gate quietly — it runs everything *and* fails `CI required`.
- **Skipped-vs-success.** `_check_results` is exact equality against a per-job expected result, not a `!= 'failure'` test. A selected job that fails, is cancelled, is skipped, or is missing all produce an error; an unselected job that ran successfully also errors. `test_sdk_routing.test_every_sdk_subset_on_code_docs_and_reused_plans` flips each of the four jobs through all five results across all 16 selections × {PR, merge_group} × {docs, code} × {reuse, fresh} and asserts *both* gates reject. That is the strongest form of this assertion I could ask for.
- **`main` push bypasses selection.** `PROBES["push"] = {"already-tested"}`, so `_classification` is not consulted and `sdks = SDK_JOBS` (`gate.py:64-66`); in the workflow, `changes` never runs on `push`, so `needs.changes.outputs.sdk_* == ''`, `'' != 'false'`, and all four run. `test_main_still_requires_all_sdks_unless_the_tree_was_tested` pins both halves. `ci.yml` has no `tags:` trigger, so there is no release-run path through this workflow; the three `*-sdk-release-gate.yml` workflows are independent and unaffected.
- **Reuse interacts correctly.** `if not reuse: expected.update(...)` matches `needs.already-tested.outputs.skip != 'true'` in the job conditions, so `push`/`merge_group` with a verified tested tree expects all four skipped. `_classification` is still called and still validated on a reused merge-group tree (`test_missing_or_invalid_sdk_decisions_fail_even_for_reused_trees`).
- **`validate` vs `validate_sdks` agree.** `FULL_JOBS - SDK_JOBS` in `validate` no longer double-books the SDK jobs `_expected_plan` already set; `SDK_JOBS ⊂ FULL_JOBS`, so nothing else was dropped from the full plan. `SDK_GATE_JOBS` still equals the `sdk-checks` job's six `needs`, so the `set(needs) != jobs` guard holds.
- **Contract generation stays unconditional.** `Validate OpenAPI spec` and `SDK wire contract is current` carry no `if:`, and the new test asserts that plus `release.count(condition) == 3`. I walked `sdk_changes.classify`'s no-SDK paths (`docs/`, `decisions/`, `assets/`, `apps/fountain/assets/`, `.../live/`, `.../components/`, the three `test/` trees, the three telemetry files) and none of them can move `FountainWeb.ApiSpec`'s output, so gating the `openapi.ts` freshness check on `sdk_typescript` cannot let the generated file go stale. Anything that *can* move the spec (`apps/fountain/lib/fountain_web/**` outside those two subtrees, `ee/`, `config/`, `mix.exs`, `sdk/contract/`, `sdk/conformance/`) falls through to `LANGUAGES`.
- **Step ordering in `release-and-contract`.** `scripts/sdk-contract/build.sh --check` already ran before `Set up Node` on the base, so making Node conditional does not strand it. Skipping `Set up Node` leaves no later step needing Node 24.
- **Tests actually run.** `workflow-checks` uses `python3 -m unittest discover -s scripts/ci -p 'test_*.py'`, so `test_sdk_routing.py` is picked up without registration. Extracted `scripts/ci` + `.github/workflows/ci.yml` at `fa038bc` into a scratch tree and ran the full discovery: 42 tests, OK — including the pre-existing `test_sdk_gate.py` from #1987, which this PR changes `validate_sdks` underneath without editing.
- **Required checks.** `require-checks.py` requires only `CI required` and `Detect secrets`, so no per-SDK context can be stranded "expected, never reported" by a skip.

Not verified: I read YAML and ran the Python gate tests locally; I did not run CI, so the real `needs` context shape (particularly `toJSON(needs)` for the `swift-sdk` matrix job) is taken on the basis that #1987 already ships against it. I did not re-litigate #1988's `OWNED_FILES`/`UNRELATED_PREFIXES` allowlist beyond the spec-staleness question above.

jhgaylor:

PR #1990, opened 2026-09-11
ci: validate the full plan on manual workflow dispatch

jhgaylor:
Manual CI dispatch now selects the full server and SDK plan, including prose checks. It records the checkout tree but bypasses diff-based selection and tested-tree reuse. The classifier makes this decision before attempting a base fetch.

Both aggregate gates reject a manual plan that skips server tests, prose checks or an SDK. The complete gate may write tested-tree evidence after all required jobs pass; the SDK-only gate still cannot publish whole-tree evidence.

Stack: based on #1989. Refs #1413. Minimum-runtime and native SDK check coverage remain follow-ups. No releases, merges or branch-rule changes.

Validation:
- All 45 CI policy tests passed, including the actual embedded shell classifier, SDK classifier and complete gate CLI.
- Manual dispatch participates in failed/cancelled/skipped-job and missing-probe coverage. Partial manual plans fail even when job results match the attempted reduced selection.
- The embedded-shell test verifies dispatch never invokes Git to fetch or classify a base.

Timing and runner jobs:
- Manual dispatch uses the existing maximum full mixed PR plan of 23 executed jobs, with changes active and the reuse probe skipped. Existing PR/main/merge-group selection is unchanged.
- Parent #1989 completed CI in 260 seconds with 23 executed jobs. [This draft’s successful PR run](https://github.com/managoat/fountain/actions/runs/34644866883) took 248 seconds with 23 executed jobs. Single-run comparisons have different cache conditions.
- The manual event was exercised locally through the workflow shell and gate CLI; no GitHub manual run has been triggered.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches: [main]
   pull_request:
+  workflow_dispatch:
   # The merge queue. Every required check must name this event or the queue
   # waits for a check that never arrives, times out and ejects the PR.
   merge_group:
@@ -78,7 +79,8 @@ jobs:
     # The queue classifies its group for the same reason a PR does: without
     # this, every docs-only merge would run the full Elixir suite on the way
     # in, which is 11% of what lands here.
-    if: ${{ github.event_name == 'pull_request' || github.event_name == 'merge_group' }}
+    if: ${{ github.event_name == 'pull_request' || github.event_name == 'merge_group'
+         || github.event_name == 'workflow_dispatch' }}
     runs-on: ubuntu-24.04
     timeout-minutes: 5
 
@@ -134,6 +136,12 @@ jobs:
             } >> "$GITHUB_OUTPUT"
             exit 0
           }
+
+          # A manual run requests complete validation, without a diff or
+          # tree-reuse shortcut. An empty base selects every SDK too.
+          if [ "${GITHUB_EVENT_NAME}" = "workflow_dispatch" ]; then
+            decide false false true
+          fi
 
           # A merge group already knows the commit it was stacked onto, and
           # there is no base *ref* to fetch: the group branch is transient and
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -491,7 +491,7 @@ retarget is now GitHub's job rather than yours, which removes the trap where a
 stale base merged cleanly into a branch that no longer existed. Land a feature
 as a stack of small chained PRs exactly as before.
 
-`scripts/ci/README.md` documents the three CI events and the queue's sizing;
+`scripts/ci/README.md` documents the four CI events and the queue's sizing;
 `gate.py`'s `PROBES` table is the authority on what each event owes.
 
 ### Coverage
--- a/scripts/ci/README.md
+++ b/scripts/ci/README.md
@@ -52,7 +52,7 @@ out `check_response_timeout_minutes` and ejects it, with no red job anywhere to
 explain why. `test_every_event_the_workflow_triggers_on_is_a_plan` in
 `test_gate.py` keeps the workflow and `gate.py` from drifting apart later.
 
-Three CI events now exist, and `gate.py`'s `PROBES` table is the authority on
+Four CI events now exist, and `gate.py`'s `PROBES` table is the authority on
 what each one owes:
 
 | Event | Probes that run | What the plan means |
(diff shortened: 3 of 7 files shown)


BinaryBourbon (approved):
**APPROVE** — adds `workflow_dispatch` as a fourth CI event whose classifier short-circuits to the complete plan (no diff, no tree reuse), and makes both gates refuse a manual run that classified anything less.

No blocking defect found. The trap I went looking for — a `workflow_dispatch` input compared as `inputs.run_all == true` when dispatch inputs are strings — does not exist here: the trigger declares no inputs at all (`ci.yml:7` parses to `workflow_dispatch: null`), and the override is keyed on `GITHUB_EVENT_NAME` in the shell. That is the stronger design; nothing to default safely.

What I traced end to end for "does the override reach every job":

- `ci.yml:142` fires `decide false false true` before any `git` call, so `docs_only=false`, `cli_docs=false`, `docs_touched=true`, `diff_base=""`.
- `diff_base=""` fails `sdk_changes.py`'s `re.fullmatch(r"[0-9a-f]{40}")` in `from_git`, which returns `LANGUAGES` — all four SDK jobs, not a subset.
- `already-tested` stays skipped on dispatch (`ci.yml:49` is unchanged), so `needs.already-tested.outputs.skip` is empty and every `!= 'true'` workload job runs. No tree-reuse shortcut is reachable.
- `docs-prose` (`ci.yml:1121-1123`) needs `docs_touched == 'true'`, which the forced classification supplies, so prose runs even on a manual run that touched no docs. `docs` (`ci.yml:1633`) stays skipped because `docs_only` is false. Both match what `gate.py` expects.
- `gate.py:68` then rejects any manual plan that is docs-only, not docs-touched, or missing an SDK, in both `validate` and `validate_sdks`.

Non-blocking notes:

1. `ci.yml:1898` — a dispatch run uploads `tested-tree`, but nothing ever reads it. `already-tested.sh` builds its candidate list from `...?event=pull_request&...` and, for push, `event=merge_group`; a `workflow_dispatch` run is in neither set. The artifact is dead weight, and `test_dispatch.py:68` (`test_full_manual_gate_can_publish_tested_tree_evidence`) reads as though it buys something. Harmless, and it fails in the safe direction, but the test name overstates it. If it is meant to be usable evidence, `already-tested.sh` needs the event added; if not, `ci.yml:1898` could say `github.event_name != 'push' && github.event_name != 'workflow_dispatch'`.

2. Dispatching on a PR's head branch posts a second `CI required` check run on the same head SHA, and GitHub resolves a duplicate context name to the most recent check run. So a manual run's verdict can supersede the `pull_request` run's for branch protection, and it validated the branch tree rather than the synthetic merge tree. This does not open a path to merging untested code — the queue's `already-tested` probe ignores dispatch runs (note 1), so `merge_group` still runs the full suite — but it is a behaviour worth a line in `scripts/ci/README.md` next to the new table row, since the escape hatch's whole point is to be used on a branch someone is trying to land.

3. `test_dispatch.py:22` asserts the literal `"|| github.event_name == 'workflow_dispatch' }}"`, which encodes the current line wrapping of `ci.yml:82-83`. Unwrapping that `if:` to a single line fails the test with no behaviour change. The `_expected_plan` guard already covers the semantics; the string assertion is the brittle half.

Checked:
- Extracted the PR tree's `scripts/ci/` and `.github/workflows/ci.yml` into a scratch dir and ran `python3 -m unittest discover -s scripts/ci -p 'test_*.py'`: 45 tests, all pass — including the three new `test_dispatch.py` cases and the `workflow_dispatch` rows added to `test_gate.py` / `test_sdk_gate.py`.
- `yaml.safe_load` on the PR's `ci.yml`: the wrapped `if:` folds to the intended single expression, and the `on:` block is `push, pull_request, workflow_dispatch, merge_group` — which is what `test_every_event_the_workflow_triggers_on_is_a_plan`'s `^  ([a-z_]+):$` regex sees, so the new trigger is genuinely covered by `PROBES`.
- `test_unsupported_event_is_rejected` correctly moved from `workflow_dispatch` to `unknown_event`; the "unsupported event" guard in `_expected_plan` still has a case.
- Read `already-tested.sh`, `sdk_changes.py` and every `if:`/`needs:` in `ci.yml` for `pull_request`/`push`/`merge_group` regressions — the change is additive on all three paths (one disjunct in the `changes` gate, one early branch keyed on `GITHUB_EVENT_NAME`, one `PROBES` entry, one event-scoped guard).
- CI-only change; no `[Unreleased]` CHANGELOG entry needed, and no new env var, so no `docs/configuration.md` row.

Not checked: I read YAML and ran the Python gate tests only. I did not dispatch the workflow, so the end-to-end claim that GitHub populates `GITHUB_EVENT_NAME=workflow_dispatch` with an empty `github.event.merge_group.base_sha` and an empty `GITHUB_BASE_REF` is reasoned from the documented contract, not observed. Note 2's duplicate-check-run behaviour is likewise from GitHub's documented resolution rule, not measured on this repo's ruleset.

jhgaylor:

PR #1991, opened 2026-09-11
ci: test Python SDK on declared runtime boundaries

jhgaylor:
The Python SDK CI job now tests Python 3.9 and 3.13, the package minimum and newest declared runtime. Each leg runs fixture lint, tests, compilation, contract and conformance checks. Fail-fast is disabled so both results remain visible, and the existing SDK aggregate requires the whole matrix to pass.

The interpreter is selected before fixture lint, using the setup-python action already pinned in the Python publish workflow. SDK source, dependencies, release behavior and path selection are unchanged. Queue sizing comments now account for the extra leg; queue settings are unchanged.

Stack: based on #1990. Refs #1413. Packaging/format coverage and the other SDKs’ runtime gaps remain separate follow-ups.

Validation:
- Python 3.9.6 and 3.13.15 each passed all 43 SDK tests, shared fixture lint (24 scenarios), compilation and the contract check (67 operations, 41 schemas, 4 enums).
- All 46 CI policy tests passed, including a check that the minimum and newest runtime declared in pyproject.toml remain covered and both run every existing check.

Timing and runner jobs:
- A full mixed PR or manual plan grows from 23 to 24 executed jobs. Changes that do not select Python still skip the entire matrix.
- Parent #1990 completed CI in 248 seconds with 23 executed jobs. [This draft’s successful CI run](https://github.com/managoat/fountain/actions/runs/34645529559) took 254 seconds with 24 executed jobs. [Python 3.9](https://github.com/managoat/fountain/actions/runs/34645529559/job/103415319751) took 29 seconds and [Python 3.13](https://github.com/managoat/fountain/actions/runs/34645529559/job/103415319740) took 23 seconds. Single-run comparisons have different cache conditions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -911,7 +911,7 @@ jobs:
         run: mix test test/conformance_test.exs
 
   python-sdk:
-    name: Python SDK
+    name: Python SDK (${{ matrix.python }})
     runs-on: ubuntu-24.04
     timeout-minutes: 15
     needs: [already-tested, changes]
@@ -921,27 +921,37 @@ jobs:
     permissions:
       contents: read
 
+    strategy:
+      fail-fast: false
+      matrix:
+        python: ["3.9", "3.13"]
+
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v5
 
+      - name: Set up Python
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: ${{ matrix.python }}
+
       - name: SDK conformance scenarios are well formed
-        run: python3 sdk/conformance/lint.py
+        run: python sdk/conformance/lint.py
 
       # The Python SDK is deliberately stdlib-only, so its source tests run
-      # on the same interpreter without an install step.
+      # on the minimum and newest declared runtimes without a package install.
       - name: Python SDK tests
-        run: python3 -m unittest discover -s sdk/python/tests -v
+        run: python -m unittest discover -s sdk/python/tests -v
 
       - name: Python SDK compiles
-        run: python3 -m compileall -q sdk/python/src sdk/python/scripts
+        run: python -m compileall -q sdk/python/src sdk/python/scripts
 
       - name: Python SDK contract
         working-directory: sdk/python
-        run: python3 scripts/verify_contract.py
+        run: python scripts/verify_contract.py
 
       - name: Python SDK conformance
         working-directory: sdk/python
-        run: python3 -m unittest discover -s tests -p "test_conformance.py" -v
+        run: python -m unittest discover -s tests -p "test_conformance.py" -v
 
   typescript-sdk:
     name: TypeScript SDK
--- a/scripts/ci/README.md
+++ b/scripts/ci/README.md
@@ -72,7 +72,7 @@ tested it.
 ### Sizing
 
 `MERGE_QUEUE` in `require-checks.py` explains the queue settings. A full CI
-run now has 23 jobs, compared with the documented limit of 20 concurrent
+run now has 24 jobs, compared with the documented limit of 20 concurrent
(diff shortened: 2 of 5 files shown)


BinaryBourbon (approved):
**APPROVE** — the `python-sdk` job becomes a two-leg matrix (`3.9`, `3.13`) with an explicit `actions/setup-python`, switches its steps from `python3` to `python`, and adds `scripts/ci/test_python_sdk_matrix.py` to hold the matrix to what `sdk/python/pyproject.toml` declares.

## The boundary check

`sdk/python/pyproject.toml` at this HEAD declares `requires-python = ">=3.9"` and classifiers `3.9, 3.10, 3.11, 3.12, 3.13`. The matrix at `.github/workflows/ci.yml:926` is exactly `["3.9", "3.13"]` — the declared minimum and the newest declared version, no version the package does not claim. Confirmed that `3.9.25` is published for `linux-24.04-x64` in the `actions/python-versions` manifest, so the `3.9` leg resolves on `ubuntu-24.04` rather than falling back to whatever the image ships.

I also checked that the 3.9 leg is plausibly green rather than an immediate red: no `X | Y` annotations, `match`/`case`, `tomllib`, `assertNoLogs`, `enterContext` or other 3.10+ stdlib/unittest APIs anywhere under `sdk/python/` or `sdk/conformance/lint.py`.

## The guard actually guards

`scripts/ci/test_python_sdk_matrix.py` is picked up by `workflow-checks` (`ci.yml:1848`, `unittest discover -s scripts/ci -p 'test_*.py'`). I extracted the tree at HEAD and ran it: 46 tests, OK. Mutating the matrix to `["3.10", "3.13"]` fails it with `'3.9' not found in ['3.10', '3.13']`, so it is not a guard that passes on anything. `ROOT = parents[2]` resolves to the repo root correctly.

## Gate and naming

- Renaming the job to `Python SDK (${{ matrix.python }})` does not orphan a required check: the only required contexts are `CI required` and `Detect secrets` (`scripts/ci/test_require_checks.py:75`), and nothing else in `.github/` or `scripts/` refers to the old display name.
- `gate.py` and `gate.py --sdk` read `toJSON(needs)`, which carries one aggregated `python-sdk.result` for the whole matrix — with `fail-fast: false`, a failed `3.9` leg still makes that aggregate `failure`, so the gate cannot go green on a half-failed matrix.
- `sdk_changes.py` classifies `.github/workflows/ci.yml` and `scripts/ci/*` as unregistered paths and therefore selects every SDK, so this PR exercises its own new matrix rather than skipping the job it changed.

## Nits (non-blocking)

- `scripts/ci/README.md:75` and the `require-checks.py` comments move 23 → 24. The delta is right (one job becomes two), but the absolute figure was already stale before this PR: expanding matrices, `ci.yml` is 25 concurrent jobs at the base SHA and 26 here (`test` ×6, `swift-sdk` ×2, now `python-sdk` ×2). Worth reconciling in whichever PR owns that prose, not here.
- `assertNotIn("        if:", job)` in the new test forbids *any* step-level condition in this job forever. That is the intent stated in the comment, but a future legitimate conditional step (say, an artifact upload on failure) will trip it with a message that does not explain itself.
- `sdk/conformance/lint.py` now runs twice, once per leg. Harmless, a few seconds.

## Checked

- Matrix versions against `pyproject.toml` `requires-python` + classifiers at HEAD.
- `fail-fast: false`, per-leg `timeout-minutes`, job-count effect on the merge-queue sizing prose.
- `needs`/`if` shape of `sdk-checks` and `gate`; skipped-vs-success semantics for the aggregated matrix result.
- `actions/setup-python` SHA pin matches the one already used in `python-sdk-publish.yml` (`5fda3b9…` / v7.0.0).
- Ran `scripts/ci` tests and a mutation of the matrix, from an extracted copy of HEAD.

**Not verified:** I did not run the SDK suite on a real CPython 3.9 (none available locally) — the 3.9-compatibility conclusion is from a source scan, not execution. I read YAML; I did not run CI.

jhgaylor:

PR #1992, opened 2026-09-11
ci: test TypeScript SDK on the minimum Node runtime

jhgaylor:
The TypeScript SDK now runs on Node 20.19.0, its declared minimum, as well as Node 24. The minimum leg compiles the existing test sources to JavaScript in a temporary fixture tree. Node 24 retains native TypeScript tests. Both legs run typechecking, builds, browser bundling, contract checks and conformance.

The compiled-test helper preserves relative fixture paths and source/package inspections, forwards a failing test exit status and removes its temporary tree in a finally block. It adds no package dependency or shipped SDK code. Fail-fast is disabled so both runtime results remain visible.

Stack: based on #1991. Refs #1413. Elixir minimum-runtime and remaining native formatting/docs/package coverage stay separate follow-ups. No release or queue-setting changes.

Validation:
- Node 20.19.0 and Node 24.21.0 each passed 109 SDK tests, the separate 24-scenario conformance check, typechecking, build, contract verification (73 operations, 49 schemas, 5 enums) and browser bundling.
- An injected failing test in an isolated SDK copy returned exit 1 and left no temporary fixture tree.
- All 47 CI policy tests passed, including minimum-version agreement with package.json, retained Node 24 native checks and unconditional build/contract checks on both legs.
- Actionlint passes with shellcheck disabled. Changed documentation passes destink and has no new repository-style findings. The SDK release guard confirms no release is required.

Timing and runner jobs:
- Full mixed PR/manual plans grow from 24 to 25 jobs. Full merge groups grow from 25 to 26 because both probes run; queue-sizing comments now distinguish these counts. Unselected TypeScript changes still skip both matrix legs.
- Parent #1991 completed CI in 254 seconds with 24 executed jobs. [This draft’s successful CI run](https://github.com/managoat/fountain/actions/runs/34646356505) took 264 seconds with 25 executed jobs. [TypeScript SDK (24)](https://github.com/managoat/fountain/actions/runs/34646356505/job/103418013146) took 23 seconds; [TypeScript SDK (20.19.0)](https://github.com/managoat/fountain/actions/runs/34646356505/job/103418013253) took 25 seconds. Single-run comparisons have different cache conditions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -954,7 +954,7 @@ jobs:
         run: python -m unittest discover -s tests -p "test_conformance.py" -v
 
   typescript-sdk:
-    name: TypeScript SDK
+    name: TypeScript SDK (${{ matrix.node }})
     runs-on: ubuntu-24.04
     timeout-minutes: 15
     needs: [already-tested, changes]
@@ -964,20 +964,24 @@ jobs:
     permissions:
       contents: read
 
+    strategy:
+      fail-fast: false
+      matrix:
+        node: ["20.19.0", "24"]
+
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v5
 
       - name: SDK conformance scenarios are well formed
         run: python3 sdk/conformance/lint.py
 
-      # The TypeScript SDK (sdk/typescript). Node runs the .ts sources
-      # directly, so `npm test` needs nothing but the dev dependency on tsc —
-      # which is also what enforces `erasableSyntaxOnly`, the setting that
-      # keeps those sources runnable without a build step.
+      # Node 24 runs the TypeScript test sources directly. The minimum
+      # runtime runs JavaScript compiled from those same tests. Both legs
+      # typecheck, build and verify the SDK contract and browser bundle.
       - name: Set up Node
         uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
-          node-version: 24
+          node-version: ${{ matrix.node }}
           cache: npm
           cache-dependency-path: sdk/typescript/package-lock.json
 
@@ -1006,7 +1010,13 @@ jobs:
         working-directory: sdk/typescript
         run: npm run typecheck
 
+      - name: SDK tests on the minimum runtime
+        if: ${{ matrix.node == '20.19.0' }}
+        working-directory: sdk/typescript
+        run: node scripts/test-compiled.mjs
+
       - name: SDK tests
+        if: ${{ matrix.node == '24' }}
         working-directory: sdk/typescript
         run: npm test
 
@@ -1024,7 +1034,13 @@ jobs:
         working-directory: sdk/typescript
         run: npm run verify-contract
 
+      - name: TypeScript SDK conformance on the minimum runtime
+        if: ${{ matrix.node == '20.19.0' }}
+        working-directory: sdk/typescript
+        run: node scripts/test-compiled.mjs --conformance
+
       - name: TypeScript SDK conformance
+        if: ${{ matrix.node == '24' }}
         working-directory: sdk/typescript
         run: npm run conformance
 
--- a/scripts/ci/README.md
+++ b/scripts/ci/README.md
@@ -71,9 +71,10 @@ tested it.
 
 ### Sizing
 
-`MERGE_QUEUE` in `require-checks.py` explains the queue settings. A full CI
-run now has 24 jobs, compared with the documented limit of 20 concurrent
-runners. The queue builds one group at a time and batches up to five PRs. Revisit
+`MERGE_QUEUE` in `require-checks.py` explains the queue settings. A full mixed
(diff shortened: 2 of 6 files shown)


BinaryBourbon (approved):
**APPROVE.** Splits `typescript-sdk` into a two-leg matrix — Node 24 keeps running the `.ts` sources natively, Node 20.19.0 runs the same tests compiled to JavaScript by `sdk/typescript/scripts/test-compiled.mjs` — and pins the arrangement with a new CI policy test.

No blocking findings.

The version claim holds in both directions. `sdk/typescript/package.json:11` declares `"node": ">=20.19"` and `.github/workflows/ci.yml:970` tests `"20.19.0"`; `scripts/ci/test_typescript_sdk_matrix.py:14-17` derives the matrix entry from `engines.node` rather than restating it, so the two cannot drift apart silently.

The minimum leg is not a green no-op. I extracted `sdk/typescript`, `sdk/conformance` and `sdk/contract` at this head into a scratch tree and ran the real thing on a real Node 20.19.0 (`npm ci --no-audit --no-fund`, then the two new commands):

- `node scripts/test-compiled.mjs` — 109 tests / 21 suites, 0 fail, exit 0. All eight `test/*.test.ts` files compile and run, matching what `npm test` globs on the 24 leg.
- `node scripts/test-compiled.mjs --conformance` — 24 scenarios, 0 fail, exit 0.
- `npm run typecheck`, `npm run build`, the esbuild browser bundle and `npm run verify-contract` (73 operations, 49 schemas, 5 enums) all pass on 20.19.0, which is what the unconditional steps need since they now run on both legs.

The two things most likely to have broken this both work: `tsc -p tsconfig.json --outDir <tmp>` emits despite `allowImportingTsExtensions`, because `rewriteRelativeImportExtensions` is on (TS 5.7+ relaxes the `noEmit` requirement), and the `.ts` specifiers in the tests are rewritten to `.js`. On the runtime side, the only post-18 APIs in `src/` are `AbortSignal.any` (Node 20.3) and global `ReadableStream` (Node 18) — nothing that needs 21+.

Failure propagation in `test-compiled.mjs` is right: `execFileSync` throws on a tsc failure, `spawnSync`'s `result.status ?? 1` turns both a nonzero exit and a signal kill into a nonzero exit code, and `if (!tests.length) throw` means an empty compile can't pass as a green run.

Gate wiring is unaffected. `scripts/ci/gate.py` keys on job **ids**, not display names, and `_check_results` asserts exact equality against `'success'`, so a skipped or failed leg fails both `sdk-checks` and `CI required`; `fail-fast: false` makes the aggregate `failure` if either leg fails rather than cancelling the sibling. The name change to `TypeScript SDK (${{ matrix.node }})` touches no required status context — `require-checks.py` publishes only `CI required` and `Detect secrets`.

The job-count edits are accurate, which I did not expect to have to check but the diff asserts them. Counting legs at this head against each job's `if:`: a PR that is neither docs-only nor untouched-docs runs 25 (`already-tested` and `docs` skipped), and a merge group runs 26 because `PROBES["merge_group"]` in `gate.py` is the only plan containing both probes — exactly what `scripts/ci/README.md:74-77` and `require-checks.py:22-23` now say.

Two nits, neither blocking:

- `test_typescript_sdk_matrix.py:14` builds the expected matrix entry as `minimum + ".0"`. That is only correct while `engines.node` stays two-component. If it ever becomes `">=20.19.1"` the test asserts `"20.19.1.0"` and fails — loudly and on the right file, so it fails closed, but the error will read as a matrix problem rather than a format one.
- The minimum leg compiles `tsconfig.json`, which `include`s `examples/**`, so every example is emitted into the temp tree and never run. Harmless, a second or two.

### Checked

- `engines.node` at head vs. the matrix, and that the guard derives rather than duplicates it.
- Mutation-tested the new guard three ways (matrix minimum → `20.18.0`; drop the `if:` from the minimum-runtime test step; `engines.node` → `">=22"`). All three fail the test, so it is guarding in both directions rather than merely existing.
- `workflow-checks` runs `unittest discover -p 'test_*.py'` over `scripts/ci` with no `if:` and no path filter, so the new file is picked up on every event, and `workflow-checks` is in the `gate` job's `needs`.
- Ran the new script's real commands on a real Node 20.19.0 (results above), plus `npm ci`, typecheck, build, browser bundle and contract on that same runtime.
- `gate.py` / `sdk-checks` / `CI required` semantics for a matrix job, and `fail-fast: false`.
- Grepped `.github`, `scripts` and `docs` for the old literal job name `TypeScript SDK` — nothing depends on it as a check context.
- Job-leg count vs. the README and `require-checks.py` prose.

Not verified: I did not run this in GitHub Actions, so `setup-node`'s behaviour with two legs sharing one npm cache key (the lockfile hash does not include the Node version) is reasoned about, not observed — at worst it is the benign "unable to reserve cache" warning, and `python-sdk` and `swift-sdk` already have the same shape. I also did not run the Node 24 leg; it is unchanged apart from the `if:`. CI-only change, so no CHANGELOG entry is owed.

jhgaylor:

PR #1993, opened 2026-09-11
ci: test Elixir SDK on its minimum supported runtime

jhgaylor:
The Elixir SDK now checks its minimum supported minor version alongside the current toolchain: Elixir 1.15.8 with OTP 26.2.5.21, and Elixir 1.19.2 with OTP 28.3. Both pairs compile with warnings as errors, run tests, build docs and packages, and verify contract/conformance behavior. Their caches already include both toolchain versions, and fail-fast is disabled.

Formatting stays on Elixir 1.19.2 because the 1.15 formatter produces different layout for existing guards and other expressions. Minimum-runtime compatibility therefore does not impose conflicting formatting rules.

Stack: based on #1992. Refs #1413. Remaining native formatting/docs/package coverage for the other SDKs stays separate. SDK source, package versions, dependency versions and queue settings are unchanged.

Validation:
- Both toolchains passed 52 SDK tests, the separate contract test and 24 conformance scenarios, warnings-as-errors compilation, docs generation and package creation. Current-toolchain formatting also passed.
- Minimum checks ran in a disposable official Elixir container pinned by digest, with CA certificates installed and read-only source mounts. The exact OTP patch version and its Ubuntu CI build availability were verified.
- The initial minimum-formatter failure is retained as evidence for keeping formatting on the current toolchain. No source was reformatted.
- All 48 CI policy tests passed, including minimum-version agreement, toolchain-specific cache keys and both pairs retaining all non-format checks.

Timing and runner jobs:
- Full mixed PR/manual plans grow from 25 to 26 jobs; full merge groups grow from 26 to 27 because both probes run. An unselected Elixir change still skips both legs.
- Parent #1992 completed CI in 264 seconds with 25 executed jobs. [This draft’s successful CI run](https://github.com/managoat/fountain/actions/runs/34647386372) took 244 seconds with 26 executed jobs. [Elixir SDK (1.19.2 / OTP 28.3)](https://github.com/managoat/fountain/actions/runs/34647386372/job/103421362138) took 24 seconds; [Elixir SDK (1.15.8 / OTP 26.2.5.21)](https://github.com/managoat/fountain/actions/runs/34647386372/job/103421362161) took 50 seconds. Single-run comparisons have different cache conditions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -839,7 +839,7 @@ jobs:
           fi
 
   elixir-sdk:
-    name: Elixir SDK
+    name: Elixir SDK (${{ matrix.elixir }} / OTP ${{ matrix.otp }})
     runs-on: ubuntu-24.04
     timeout-minutes: 15
     needs: [already-tested, changes]
@@ -853,6 +853,15 @@ jobs:
     env:
       MIX_ENV: test
 
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - elixir: "1.15.8"
+            otp: "26.2.5.21"
+          - elixir: "1.19.2"
+            otp: "28.3"
+
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v5
 
@@ -863,8 +872,8 @@ jobs:
         id: beam
         uses: erlef/setup-beam@54075bcc5e249e4758d363f27d099f55d843f124 # v1.24.1
         with:
-          elixir-version: "1.19.2"
-          otp-version: "28.3"
+          elixir-version: ${{ matrix.elixir }}
+          otp-version: ${{ matrix.otp }}
 
       # The Elixir SDK is a standalone Mix project. Keep its dependency graph
       # and build artifacts separate from the umbrella's cache.
@@ -882,7 +891,10 @@ jobs:
         working-directory: sdk/elixir
         run: MIX_ENV=dev mix deps.get
 
+      # Formatter output changes across Elixir versions. The current
+      # toolchain owns style; both pairs check runtime compatibility.
       - name: Elixir SDK formatting
+        if: ${{ matrix.elixir == '1.19.2' }}
         working-directory: sdk/elixir
         run: mix format --check-formatted
 
--- a/scripts/ci/README.md
+++ b/scripts/ci/README.md
@@ -72,7 +72,7 @@ tested it.
 ### Sizing
 
 `MERGE_QUEUE` in `require-checks.py` explains the queue settings. A full mixed
-PR runs 25 jobs; a full merge group runs 26 because both probes run. The
+PR runs 26 jobs; a full merge group runs 27 because both probes run. The
 documented limit is 20 concurrent runners. The queue builds one group at a
 time and batches up to five PRs. Revisit
 `max_entries_to_build` when the concurrency limit changes.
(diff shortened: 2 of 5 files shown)


BinaryBourbon (approved):
**APPROVE** — the `elixir-sdk` job becomes a two-leg matrix (1.15.8/OTP 26.2.5.21 and 1.19.2/OTP 28.3), formatting is pinned to the current toolchain, and a new policy test locks the pairing to what `sdk/elixir/mix.exs` declares.

No blocking findings.

The central claim holds. `sdk/elixir/mix.exs:10` declares `elixir: "~> 1.15"`, `sdk/elixir/README.md:15` and `docs/elixir-sdk.md:21` both say "Elixir 1.15 and newer", and the minimum leg is 1.15.8 — the job tests the floor the package actually publishes. The lower floor than the umbrella's `.tool-versions` (28.3 / 1.19.2-otp-28) is the right call for a Hex package and mirrors what `python-sdk` already does with 3.9/3.13.

Both pairs exist and are installable on `ubuntu-24.04`: `builds.hex.pm/builds/otp/ubuntu-24.04/builds.txt` lists `OTP-26.2.5.21`, and the Elixir build list has `v1.15.8-otp-26` and `v1.19.2-otp-28`. Elixir 1.15 supports OTP 24–26, so 1.15.8 + OTP 26 is a valid combination rather than one that silently resolves elsewhere.

The minimum leg should also resolve and compile: every entry in `sdk/elixir/mix.lock` declares an Elixir requirement 1.15 or lower (`ex_doc` 0.40.3, `finch` 0.23.0, `mint` 1.10.0, `earmark_parser` 1.4.46 and `hpax` 1.0.4 are all `~> 1.15`; the rest are older), and the SDK source uses no post-1.15 stdlib.

### Non-blocking

- `scripts/ci/test_elixir_sdk_matrix.py:19,24` hardcodes `1.19.2` / `28.3` in two places, while the minimum is derived from `mix.exs`. A `.tool-versions` bump therefore leaves the "current" leg on a stale pair with every gate green — the failure this repo keeps hitting in a different costume. Deriving the current pair from `.tool-versions` the way the minimum is derived from `mix.exs` would close it. The version-skew *within* this PR is already covered: dropping the min leg, bumping the declared minimum, or adding an `if:` to a non-formatting step each fail the test (verified by mutation, below).
- `scripts/ci/test_elixir_sdk_matrix.py:15` reads `elixir: "~> ([0-9.]+)"` and then matches `minimum + "."`, so a future `~> 1.15.8` in `mix.exs` would make the assertion unsatisfiable (`"1.15.8".startswith("1.15.8.")` is false). Only a trap for a later edit, not a problem today.
- `.github/workflows/elixir-sdk-release-gate.yml` and `.github/workflows/elixir-sdk-publish.yml` still pin a single toolchain. Out of scope here, but the release gate is the other place a 1.15-incompatible change could land unnoticed.

### Checked

- Job/leg arithmetic in `scripts/ci/README.md:75` and `scripts/ci/require-checks.py:22,30`: parsed `ci.yml` and counted matrix legs. Base = 27 total legs, HEAD = 28. A `pull_request` skips `already-tested` and `docs` (which needs `docs_only == 'true'`) → 25 → 26; a `merge_group` skips only `docs` → 26 → 27. The 25→26 / 26→27 edits are exactly right, not an off-by-one carried forward.
- The gate: `scripts/ci/gate.py` asserts on `needs["elixir-sdk"]["result"]`, the matrix *aggregate*, so a failing 1.15 leg fails `sdk-checks` and `CI required` despite `fail-fast: false`. Both legs must pass.
- The `name:` change to `Elixir SDK (${{ matrix.elixir }} / OTP ${{ matrix.otp }})` breaks no required context: `require-checks.py` requires only `CI required` and `Detect secrets` (asserted in `test_require_checks.py:74`), and no other workflow or doc refers to the old check name.
- Cache isolation: the key at `ci.yml:887` already keys on `steps.beam.outputs.otp-version` / `elixir-version`, so the two legs cannot share a `_build`. The new test asserts both stay in the key.
- The skip is narrow: only `Elixir SDK formatting` carries an `if:`; dependencies, compile, tests, docs, `hex.build`, contract and conformance run on both legs, and the test enforces that no `if:` creeps onto them.
- Ran `python3 -m unittest discover -s scripts/ci -p 'test_*.py'` on the HEAD tree — 48 tests, OK. Mutation-tested the new guard three ways (remove the 1.15.8 matrix entry; add `if:` to `Elixir SDK tests`; bump `mix.exs` to `~> 1.17`) — each fails it, so it is guarding rather than merely passing.
- `actionlint` on the HEAD `ci.yml`: nothing in or near this job (the only output is pre-existing SC2010/SC2012 notes around lines 1404–1556).

### Not verified

I read YAML and queried the build/Hex indexes; I did not install Elixir 1.15.8 or run the SDK suite on it. Whether `mix compile --warnings-as-errors` and `mix docs --warnings-as-errors` are clean on 1.15.8 is what the first run of this job will tell you — and if it is not, that is the job doing its job.

jhgaylor:

PR #1994, opened 2026-09-11
ci: test the built Python SDK wheel in isolation

jhgaylor:
Python SDK CI now builds an sdist and wheel on both supported runtime legs, installs the wheel into a fresh virtual environment, and runs all existing SDK regression tests against the installed package. This replaces the full source-tree run. Separate contract and conformance steps retain their focused diagnostics.

The verifier imports the installed package before discovering tests, then checks every loaded SDK module stays inside that package. It also verifies the type marker and distribution version. Installation uses only the local wheel, with no dependency resolution or registry access, and temporary environments are removed on success or failure.

Stack: based on #1993. Refs #1413. No SDK source, version or publishing changes. Remaining formatting and other SDK package/docs checks stay separate.

Validation:
- Python 3.9.6 and 3.13.15 each built an sdist/wheel and passed all 43 SDK tests against an isolated wheel install.
- Damaged wheels missing py.typed or a runtime module were rejected even when source files were supplied through PYTHONPATH; temporary environments were removed.
- A fixture without any checkout SDK sources ran all 43 real tests, propagated one deliberately failing extra test and cleaned its environment.
- All 48 CI policy tests passed. Actionlint passes with shellcheck disabled; changed docs pass destink and have no new repository-style findings.

Timing and runner jobs:
- Job selection and counts are unchanged: a full mixed PR/manual plan has 26 jobs, and a full merge group has 27.
- Parent #1993 completed CI in 244 seconds with 26 executed jobs. This draft completed CI successfully in 273 seconds with 26 executed jobs. Python 3.9 took 41 seconds; Python 3.13 took 30 seconds (run 34648421953). Single-run comparisons have different cache conditions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -949,11 +949,8 @@ jobs:
       - name: SDK conformance scenarios are well formed
         run: python sdk/conformance/lint.py
 
-      # The Python SDK is deliberately stdlib-only, so its source tests run
-      # on the minimum and newest declared runtimes without a package install.
-      - name: Python SDK tests
-        run: python -m unittest discover -s sdk/python/tests -v
-
+      # The SDK has no runtime dependencies. Check source syntax and wire
+      # assumptions, then exercise the built distribution below.
       - name: Python SDK compiles
         run: python -m compileall -q sdk/python/src sdk/python/scripts
 
@@ -964,6 +961,15 @@ jobs:
       - name: Python SDK conformance
         working-directory: sdk/python
         run: python -m unittest discover -s tests -p "test_conformance.py" -v
+
+      - name: Python SDK build tool
+        run: python -m pip install --disable-pip-version-check build
+
+      - name: Python SDK package dry run
+        run: python -m build --outdir "$RUNNER_TEMP/python-sdk-dist" sdk/python
+
+      - name: Python SDK installed wheel
+        run: python sdk/python/scripts/verify_wheel.py "$RUNNER_TEMP"/python-sdk-dist/*.whl
 
   typescript-sdk:
     name: TypeScript SDK (${{ matrix.node }})
--- a/scripts/ci/README.md
+++ b/scripts/ci/README.md
@@ -90,6 +90,11 @@ with OTP 26.2.5.21 and Elixir 1.19.2 with OTP 28.3. Formatting uses 1.19.2
 because formatter output differs between versions; all remaining checks run
 on both pairs. The Python job owns its
 tests, compilation, contract and conformance checks on Python 3.9 and 3.13.
+Both legs also build an sdist and wheel, install the wheel in a fresh virtual
+environment and run the SDK regression suite against that installed package.
+This replaces the full source-tree test run; separate contract and conformance
+steps retain their focused diagnostics.
+The verifier checks the type marker, package version and every SDK import path.
 These cover the package minimum and newest declared runtime. Both legs must
 pass; a failure does not cancel the other leg. The TypeScript job owns
 installation, type checks, tests, builds, browser bundling, contract and
--- a/scripts/ci/test_python_sdk_matrix.py
+++ b/scripts/ci/test_python_sdk_matrix.py
@@ -7,7 +7,7 @@ ROOT = Path(__file__).resolve().parents[2]
 
 
 class PythonSDKMatrixTest(unittest.TestCase):
-    def test_supported_runtime_boundaries_run_all_existing_checks(self):
+    def test_supported_runtime_boundaries_run_complete_sdk_checks(self):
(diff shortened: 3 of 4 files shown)


BinaryBourbon (approved):
**APPROVE.** Replaces the Python SDK's source-tree test run in `ci.yml` with a build-then-install-then-test pass: `python -m build` into `$RUNNER_TEMP`, and `sdk/python/scripts/verify_wheel.py` installs the wheel into a throwaway venv and runs the regression suite against the installed package.

I extracted the tree at `aa1013ff` into a scratch directory and actually ran the new path, including a negative case. The four properties that matter here hold:

- **Isolation is real.** The venv install is `--no-index --no-deps`, `cwd` is the temp directory, and `python -I` drops the cwd from `sys.path`. `verify_wheel.py:20` imports `fountain` *before* `unittest.discover`, so the `sys.path.insert(0, .../src)` at `sdk/python/tests/test_sdk.py:15` and `test_conformance.py:18` cannot win — submodules resolve through the already-bound `fountain.__path__`. The post-run sweep at `verify_wheel.py:32-35` then proves it per module.
  I checked this by building a wheel whose `__version__`/`version` were `9.9.9`, reverting the source to `0.1.1`, and running the verifier: `test_version_tokens_match_package_metadata` failed (wheel `9.9.9` vs source `pyproject.toml` `0.1.1`) and the step exited non-zero. If the source tree were satisfying the import it would have passed silently.
- **It exercises the package, not an empty install.** 43 tests ran inside the venv (`test_sdk`, `test_contract`, `test_conformance`), plus the `py.typed` marker check and the `__version__` vs `importlib.metadata.version("fountain-agent-sdk")` check. `countTestCases()` guarding against a zero-test discovery (`verify_wheel.py:27-28`) is the right instinct — an empty discover is exactly the "green while doing nothing" shape this repo keeps hitting.
- **The sdist is covered, transitively.** `python -m build` with no selector builds the sdist and then builds the wheel *from that sdist* — I confirmed the log line `* Building wheel from sdist` on this project. So the wheel under test is what an sdist consumer would get, and a packaging omission cannot be masked by the working tree.
- **Nothing stale.** `--outdir "$RUNNER_TEMP/python-sdk-dist"` is per-job and empty at job start, and `sdk/python/dist` is never consulted. A zero-match glob passes the literal path through and `pip install` fails closed; a multi-match glob trips argparse. Both fail loud.

Coverage is not reduced: `compileall`, contract and conformance still run against `src/`, and the guard at `scripts/ci/test_python_sdk_matrix.py:24-31` was updated in the same commit, so neither leg of the 3.9/3.13 matrix can quietly lose a step. No new job, so `gate.py --sdk` needs no change. `sdk_changes.py` selects `python` for anything under `sdk/python/`, so `verify_wheel.py` edits do re-run the job.

Non-blocking notes:

1. `.github/workflows/ci.yml:966` — `pip install build` is unpinned, and build's isolation then fetches `setuptools>=68` from PyPI. This job previously needed no network beyond the checkout; it now has two unpinned PyPI resolutions in a repo that SHA-pins every action. Risk is low (`contents: read`, no secrets, test-only), but a PyPI hiccup now reds a job that used to be offline-capable. Pinning `build==<version>` would keep it consistent with the rest of `.github/workflows/`.
2. `sdk/python/scripts/verify_wheel.py:51-52` — `subprocess.run(..., check=True)` on the test run means a plain test failure prints the real `FAILED (failures=1)` and then a `CalledProcessError` traceback that dumps the entire `RUN_TESTS` source as one escaped line. In a CI log that buries the signal. `sys.exit(subprocess.run(...).returncode)` for the second call would keep the exit semantics without the noise.
3. `.github/workflows/python-sdk-publish.yml:56` still runs `unittest discover -s sdk/python/tests` against the source tree before publishing. Out of scope for this PR, but that is the one workflow where "does the artifact we are about to upload actually import" pays for itself.
4. The conformance suite now runs twice per leg (the dedicated step and again inside the wheel run), ~15s. `scripts/ci/README.md` says this is deliberate for focused diagnostics; noting it only so the duplication is not mistaken for an oversight later.
5. CI-only change, no CHANGELOG entry needed, no new env var, so no `docs/configuration.md` row.

Checked: ran `python -m build` and `verify_wheel.py` on the extracted tree (macOS, 3.11) — 43 tests pass; ran the stale-wheel negative and confirmed it fails closed; confirmed `build` does sdist-then-wheel-from-sdist; read `sdk_changes.py`, the `python-sdk` job's `if:`/`needs:`, and the updated matrix guard. Not checked: behaviour on the 3.9 leg specifically (`Path.is_relative_to` is 3.9+, so it should hold), actual runner wall-clock against the 15-minute timeout, and whether `build`/`setuptools` resolve cleanly on 3.9 under the current PyPI index.

jhgaylor:

PR #1940, opened 2026-09-11
ci: check public GitHub links in documentation

jhgaylor:
The app-repository links from #1698 were repaired by #1829, but documentation CI still could not detect another public link becoming private or disappearing. Check GitHub links anonymously in the manual, extension manuals, and CLAUDE.md; deduplicate URLs, exclude code examples, bound requests and retries, and reject redirects outside the host allowlist. Run this blocking check on both docs-only and mixed docs/code CI paths, including edits to the checker or contributor guide.

The first live check found one remaining 404: the observability guide linked a private home-cloud file. Point readers to the hosted-rule explanation already on that page.

Closes #1698. Four files, +148/-3.

Validation: all 27 CI policy tests pass, including four checker regressions. The first live run reported the private link; after the edit, all 76 public GitHub URLs passed. Both CI paths and source-path classification were verified. All three prose reports reviewed: two existing docs-style findings elsewhere, eight existing Vale errors with advisory wording findings, and a clean destink report. Full local precommit passes with four VM schedulers: 5,342 core tests plus 6 doctests, 144 Buzz tests, and 33 Support tests, with no failures.

PR #1939, opened 2026-09-11
ci: require shipped alert fixtures on every CI plan

jhgaylor:
Alert fixtures ran only in a path-filtered workflow that the merge gate did not require. Run them in `workflow-checks`, which `CI required` already requires for every supported plan, including docs-only changes and reused trees. Allow ten minutes for the evaluator package installation and checks.

Portable CI slice of #1692: one file, +8/-2. Hosted credits fixtures and the missing-worker fix are a separate small stack in jhgaylor/home-cloud#225, on #224's email fixtures.

Validation: all 23 CI policy tests pass, including failure/cancellation/skip rejection for required jobs on each plan; all 45 shipped Prometheus expression cases pass. Full local precommit passes with four VM schedulers: 5,342 core tests plus 6 doctests, 144 Buzz tests, and 33 Support tests, with no failures.

PR #1995, opened 2026-09-11
ci: split SDK checks into per-SDK jobs with selective routing (#1413)

jhgaylor:
Collapsed stack for #1413: splits the monolithic `sdk-clients` CI job into
per-SDK jobs, adds a stable aggregate check, a path classifier with selective
routing, runtime-boundary matrices and packaging tests.

Originally cut as twelve chained PRs (#1984-#1995). Each was reviewed and
approved individually; this PR carries all twelve commits so the combination
gets one full CI run against current `main` instead of twelve serial
merge-queue builds. The individual PRs stay open until this merges, then close
as superseded.

| PR | Commit |
|---|---|
| #1984 | run the Elixir SDK checks in a separate job |
| #1985 | run the Python SDK checks in a separate job |
| #1986 | run the TypeScript SDK checks in a separate job |
| #1987 | report a stable SDK aggregate check |
| #1988 | classify SDK paths before enabling selective jobs |
| #1989 | run only SDKs selected by the change classifier |
| #1990 | validate the full plan on manual workflow dispatch |
| #1991 | test Python SDK on declared runtime boundaries |
| #1992 | test TypeScript SDK on the minimum Node runtime |
| #1993 | test Elixir SDK on its minimum supported runtime |
| #1994 | test the built Python SDK wheel in isolation |
| #1995 | validate the packed TypeScript SDK in a consumer project |

`main` moved 36 commits since the stack was cut, including three CI changes
(#1927, #1939, #1940). The merge is clean and the CI policy suite passes on the
merged tree (56 tests), with both sides' workflow changes intact.

Known follow-ups, none blocking:

- A docs-only edit to `docs/swift-sdk.md` now starts the two-runner `swift-sdk`
  matrix including `macos-15`; no step in that job reads the page.
- Full-run concurrency goes from ~21 to 26-27 legs against the free plan's
  20-concurrent ceiling. `require-checks.py`'s `max_entries_to_build: 1`
  reasoning and `decisions/0050` still cite the old count.
- `CLAUDE.md:424` still documents a single `sdk-clients` job;
  `scripts/ci/README.md:88` and `ci.yml:908` claim `swift-sdk` validates
  conformance fixtures, which it does not.
- `ee/test/` is missing from `sdk_changes.py`'s `UNRELATED_PREFIXES` (fail-open).
  Any fix must be `ee/test/` specifically, never `ee/` — `ee/lib/fountain_web/**`
  must keep fanning out because it can move the OpenAPI document.

https://claude.ai/code/session_01HD7jxhmpwc8fAbWAHby4PE

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
         run: bash scripts/ci/already-tested.sh
 
   changes:
-    name: Decide whether this is a docs-only change
+    name: Classify documentation and SDK changes
     # Never on push, and deliberately so. There, `already-tested` above
     # already collapses a rebased squash-merge to a few seconds, and a main
     # push that is NOT already-tested is precisely the case that must run
@@ -90,6 +90,10 @@ jobs:
       docs_touched: ${{ steps.filter.outputs.docs_touched }}
       cli_docs: ${{ steps.filter.outputs.cli_docs }}
       tree: ${{ steps.tree.outputs.tree }}
+      sdk_elixir: ${{ steps.sdks.outputs.sdk_elixir }}
+      sdk_python: ${{ steps.sdks.outputs.sdk_python }}
+      sdk_typescript: ${{ steps.sdks.outputs.sdk_typescript }}
+      sdk_swift: ${{ steps.sdks.outputs.sdk_swift }}
 
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v5
@@ -120,8 +124,10 @@ jobs:
           MERGE_GROUP_BASE_SHA: ${{ github.event.merge_group.base_sha }}
         run: |
           set -u
+          base=""
           decide() {
             {
+              echo "diff_base=${base:-}"
               echo "docs_only=$1"
               echo "cli_docs=$2"
               echo "docs_touched=$3"
@@ -192,6 +198,12 @@ jobs:
 
           echo "docs-only — running the docs job instead of the Elixir suite"
           decide true "${cli_docs}" false
+
+      - name: Select SDKs from the same diff
+        id: sdks
+        env:
+          DIFF_BASE: ${{ steps.filter.outputs.diff_base }}
+        run: python3 scripts/ci/sdk_changes.py "$DIFF_BASE" >> "$GITHUB_OUTPUT"
 
   test:
     name: Build and Test (partition ${{ matrix.partition }})
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1048,6 +1048,10 @@ jobs:
         working-directory: sdk/typescript
         run: npx --yes esbuild dist/index.js --bundle --platform=browser --format=esm --outfile=/dev/null
 
+      - name: SDK packed artifact
+        working-directory: sdk/typescript
+        run: node scripts/verify-package.mjs
+
       - name: TypeScript SDK contract
         working-directory: sdk/typescript
         run: npm run verify-contract
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -815,8 +815,8 @@ jobs:
             exit 1
           fi
 
-  sdk-clients:
-    name: SDK clients, CLI and plugins
+  elixir-sdk:
+    name: Elixir SDK
     runs-on: ubuntu-24.04
     timeout-minutes: 15
     needs: [already-tested, changes]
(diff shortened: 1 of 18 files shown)


BinaryBourbon (approved):
**APPROVE**

Adds `sdk/typescript/scripts/verify-package.mjs` and a `SDK packed artifact` step to the `typescript-sdk` job: it packs the just-built SDK, installs the tarball into a throwaway consumer project under `$TMPDIR`, typechecks that consumer against the shipped declarations, runs it, and bundles it for a browser.

This is the shape a packaging test has to have, and I checked the five ways these usually turn out to be no-ops.

**Verified by running it**

I copied `verify-package.mjs` at this head into the scratchpad, repointed only the `sdk` constant at the checkout's `sdk/typescript`, and ran it (Node 26 / npm 11, against the checkout's older-but-structurally-identical SDK). It passed end to end: `npm pack --json --offline` parses, `npm install --offline` into a `{private, type:"module"}` package.json with no `name`/`version` works, `tsc` from the SDK's `node_modules` typechecks the consumer, `node consumer.js` runs, `esbuild --platform=browser` produces a 56 kB bundle.

Then the negative cases, on the installed tarball in `node_modules`:

- delete `dist/index.d.ts` -> `TS7016 ... implicitly has an 'any' type` (exit 2). Missing declarations are caught, because `--strict` is passed and no `skipLibCheck`.
- delete `dist/` -> `TS2307: Cannot find module '@…/fountain-sdk'` (exit 2). A `files:` list that drops `dist` is caught.

So (c) holds. `execFileSync` throws on non-zero and the `finally` only does `rmSync`, so every one of those exits the script non-zero and fails the step.

**The other four checks**

- (a) Isolation. The consumer is `mkdtempSync(join(tmpdir(), "fountain-package-"))` + `/consumer`, so Node's upward resolution walk hits `$TMPDIR/node_modules` and `/node_modules`, never `sdk/typescript`. No workspace root exists to link against — the repo has no root `package.json`. `tsc` is invoked by path out of the SDK's `node_modules` but with `cwd` at the consumer, so `@types` resolution walks from the consumer too (confirmed: the baseline typecheck passes with no `@types/node` present, which it would not if the SDK's tree were leaking in).
- (b) It calls something real. `consumer.ts` constructs both entry points, calls `client.agents.list()`, and asserts inside the injected `fetch` that the request is `GET https://package.invalid/api/agents` with `authorization: Bearer package-test`, then asserts `calls === 1`. I traced that against `src/config.ts` (option `apiKey`/`baseUrl` win over env and the credentials file), `src/http.ts` (`Authorization: Bearer …`, `list` unwraps `{data}`) and `Agents`' `/api/agents` path — the assertion is the real wire shape, not a tautology. An empty install cannot pass this.
- (d) ESM/CJS. The package is `"type": "module"` with no `require` condition in `exports`, so there is no CJS resolution to exercise; correctly out of scope. Both conditions that *do* exist are covered: `import Fountain from "<name>"` under Node takes the `node` condition (`dist/node.js`), `"<name>/browser"` and the esbuild `--platform=browser` pass take the `browser`/`default` condition (`dist/index.js`).
- (e) Freshness. `ci.yml:1051` sits after `SDK build` (`ci.yml:1041`), and `--pack-destination` is the fresh mkdtemp, so nothing stale can satisfy it.

`scripts/ci/test_typescript_sdk_matrix.py` guards the step (KeyError if it is deleted, and asserts no `if:` so it runs on both matrix legs); I ran that test against this head's `ci.yml` + `package.json` in an isolated tree and it passes. `.github/workflows/ci.yml` still parses. No new job, so `CI required` needs nothing. `scripts/` is not in `files:`, so the new script is not shipped; a CI-only change like this needs no CHANGELOG entry.

**Non-blocking**

1. `verify-package.mjs:20` — `npm install --offline` doubles as an unstated assertion that the SDK has zero runtime dependencies (the comment above it says as much). The day someone adds one, this step fails with a registry-cache error that reads like CI flake rather than "the SDK took a dependency". A one-line explicit check on the tarball's `dependencies` would name the failure.
2. `verify-package.mjs:48` — both import specifiers resolve `types` to `dist/index.d.ts`, so `dist/node.d.ts` is never typechecked by the consumer. A *missing* `dist/node.js` is still caught (by `node consumer.js` at line 51), but a broken declaration for the Node entry would not be.
3. `verify-package.mjs:52` — `npx --yes esbuild` unpinned, same as the existing `SDK bundles for the browser` step. Pre-existing pattern, not this PR's to fix.

**Checked / not checked**

Checked: the script's real behaviour and its two failure modes, as above; the `ci.yml` step placement and matrix applicability; the `package.json` `exports`/`files`/`type`; the request assertion against `config.ts`/`http.ts`/`resources.ts`; the Python guard test; YAML parse.

Not checked: I did not build this PR's own SDK sources — the local run used the checkout's older `sdk/typescript` with an identical `exports`/`files`/`type` shape, so it validates the harness, not that this stack's `dist` typechecks in a consumer. I also ran npm 11 / Node 26, not the npm bundled with Node 20.19.0, so `npm pack --json` stdout purity on that leg is inferred rather than observed. And I read CI, I did not run it.

PR #1996, opened 2026-09-11
fix: preserve rebound turns when an old actor stops

jhgaylor:
A normally stopping conversation actor could orphan its running turn and mark the conversation idle after another process moved that conversation to a different sandbox. Normal shutdown now supplies the actor’s sandbox ID. Orphan reconciliation locks the conversation row and checks that binding before writing either the turn or the conversation; a mismatch produces no turn, status, stage or audit writes.

The parent lock remains held through both conditional writes. Existing system-reaper calls retain their unbound behavior, and completed turns remain a no-op. Audit and event publication remain outside the transaction.

Stack: based on #1983. Refs #1767. This unit protects normal-stop bookkeeping across sandbox reassignment. Same-sandbox replacement actor ownership and durable forced-teardown recovery remain separate follow-ups.

Validation:
- 146 focused tests passed, including normal shutdown, existing reaper/turn-machine behavior and audit guardrails.
- Two real PostgreSQL connection tests observe row-lock waits in both race orders. Reassignment first preserves the running turn; orphaning first finishes before reassignment commits.
- Running the new tests against the parent implementation produced the three expected regression failures; restoring this change passed all focused tests.
- Full `mix precommit` passed: Fountain 5,437 tests and 6 doctests, Buzz 144 tests, Support 33 tests; zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2224,33 +2224,44 @@ defmodule Fountain.Conversations do
 
   This function is unscoped because it is called by a conversation's own
   server and by the system reaper. Callers may supply audit attribution.
+
+  An actor recovering its own turn passes `:expected_sandbox_id`; the locked
+  parent having been rebound to another sandbox answers
+  `{:error, :ownership_changed}` and writes nothing. The reaper omits the key,
+  because it recovers on nobody's behalf.
   """
   def _unsafe_orphan_turn(%Turn{} = turn, why, opts \\ []) do
     # ownership: this is the existing actor's turn or a system recovery candidate.
+    recover_opts = Keyword.take(opts, [:expected_sandbox_id])
+
     result =
-      ExecutionGuard._unsafe_recover_turn(turn, fn current, conv, latest?, bounded? ->
-        now = DateTime.utc_now() |> DateTime.truncate(:second)
-        reply_text = current.reply_text || _unsafe_turn_reply_text(current)
+      ExecutionGuard._unsafe_recover_turn(
+        turn,
+        fn current, conv, latest?, bounded? ->
+          now = DateTime.utc_now() |> DateTime.truncate(:second)
+          reply_text = current.reply_text || _unsafe_turn_reply_text(current)
 
-        updates =
-          if current.status == "running",
-            do: [status: "interrupted", ended_at: now, orphaned_at: now],
-            else: [orphaned_at: now]
+          updates =
+            if current.status == "running",
+              do: [status: "interrupted", ended_at: now, orphaned_at: now],
+              else: [orphaned_at: now]
 
-        updated =
-          current
-          |> Turn.changeset(Map.new(maybe_set_reply_text(updates, reply_text)))
-          |> Repo.update!()
+          updated =
+            current
+            |> Turn.changeset(Map.new(maybe_set_reply_text(updates, reply_text)))
+            |> Repo.update!()
 
-        conversation_changed? = latest? and conv.status == "running"
+          conversation_changed? = latest? and conv.status == "running"
 
-        conv =
-          if conversation_changed?,
-            do: conv |> Conversation.changeset(%{status: "idle"}) |> Repo.update!(),
-            else: conv
+          conv =
+            if conversation_changed?,
+              do: conv |> Conversation.changeset(%{status: "idle"}) |> Repo.update!(),
+              else: conv
 
-        {updated, conv, conversation_changed?, bounded?}
-      end)
+          {updated, conv, conversation_changed?, bounded?}
+        end,
+        recover_opts
+      )
 
     case result do
       {:ok, :noop} ->
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -2021,7 +2021,11 @@ defmodule Fountain.Conversations.ConversationServer do
     _ = CallbackKey.revoke(state.conversation_id, state.callback_api_key_id)
 
     # Last and best-effort, so it cannot skip the revocation above.
-    _ = TurnMachine.orphan_on_normal_stop(reason, state.current_turn, state.conversation_id)
+    _ =
+      TurnMachine.orphan_on_normal_stop(reason, state.current_turn, state.conversation_id,
+        expected_sandbox_id: state.sandbox_id
+      )
+
     :ok
   end
 
--- a/apps/fountain/lib/fountain/conversations/execution_guard.ex
+++ b/apps/fountain/lib/fountain/conversations/execution_guard.ex
@@ -387,14 +387,27 @@ defmodule Fountain.Conversations.ExecutionGuard do
     end)
   end
 
-  @doc "Retire an orphan's execution before recovery writes, in parent/journal/turn lock order."
-  def _unsafe_recover_turn(%Turn{} = observed, writer) do
+  @doc """
+  Retire an orphan's execution before recovery writes, in parent/journal/turn lock order.
+
+  `:expected_sandbox_id` is the recovering actor's own binding. `cleanup_binding?/1`
+  already refuses a bounded turn whose journal names a sandbox the parent no
+  longer points at, but an unbounded turn has no journal row and nothing else
+  records which machine was driving it. An actor that supplies the option and
+  finds the locked parent reassigned recovers nothing: the rollback happens
+  before `retire_orphan/2`, so a stale actor writes neither the turn nor the
+  journal. Supplying `nil` is an expectation of "no sandbox", not an absent one;
+  omitting the key entirely is what the system reaper does, because it is
+  recovering on nobody's behalf.
+  """
+  def _unsafe_recover_turn(%Turn{} = observed, writer, opts \\ []) do
     transaction(fn ->
       conv = lock_parent(observed.conversation_id) || Repo.rollback(:not_found)
       execution = lock_execution_by_turn(observed.id)
       turn = lock_turn(observed.id) || Repo.rollback(:turn_missing)
       if turn.conversation_id != conv.id, do: Repo.rollback(:ownership_changed)
 
+      if rebound?(opts, conv), do: Repo.rollback(:ownership_changed)
       if execution && not cleanup_binding?(execution), do: Repo.rollback(:ownership_changed)
       running? = turn.status == "running"
       {turn, changed, event} = retire_orphan(execution, turn)
@@ -958,6 +971,13 @@ defmodule Fountain.Conversations.ExecutionGuard do
   # must still prove its tenant/name/provider binding; a surviving conversation
   # must also remain bound to it. Missing or changed sandbox identity stays
   # uncertain. Reset cannot reuse this row while its journal remains open.
+  defp rebound?(opts, conv) do
+    case Keyword.fetch(opts, :expected_sandbox_id) do
+      {:ok, expected} -> expected != conv.sandbox_id
+      :error -> false
+    end
+  end
+
   defp cleanup_binding?(execution) do
     sandbox_matches =
       Repo.exists?(
--- a/apps/fountain/lib/fountain/conversations/execution_guard.ex
+++ b/apps/fountain/lib/fountain/conversations/execution_guard.ex
@@ -971,13 +971,6 @@ defmodule Fountain.Conversations.ExecutionGuard do
   # must still prove its tenant/name/provider binding; a surviving conversation
   # must also remain bound to it. Missing or changed sandbox identity stays
   # uncertain. Reset cannot reuse this row while its journal remains open.
-  defp rebound?(opts, conv) do
-    case Keyword.fetch(opts, :expected_sandbox_id) do
-      {:ok, expected} -> expected != conv.sandbox_id
-      :error -> false
-    end
-  end
-
   defp cleanup_binding?(execution) do
     sandbox_matches =
       Repo.exists?(
@@ -994,6 +987,28 @@ defmodule Fountain.Conversations.ExecutionGuard do
       end
 
     sandbox_matches and parent_matches
+  end
+
+  # The other half of the same question, for a turn that has no journal row to
+  # ask it of. `cleanup_binding?/1` above compares the *journal's* recorded
+  # sandbox identity; this compares the *actor's* — the only record that exists
+  # when there is no `TurnExecution`.
+  #
+  # That is not a corner case. Whether a journal row exists is decided by
+  # execution limits (`resolve_turn_limits/1`), not by turn capacity, and limits
+  # ship inert: `host_ceiling/0` is `%{}` and `enforced_controls/1` is `[]`
+  # (#1773-#1793). So today every production turn takes the no-journal path and
+  # this guard is the only thing standing between a stale actor and another
+  # machine's turn. Do not delete it as exotic.
+  #
+  # An explicit `nil` is an expectation of "no sandbox" and fences; only an
+  # absent key means the caller is recovering on nobody's behalf, which is why
+  # this is `Keyword.fetch/2` and not `Keyword.get/2`.
+  defp rebound?(opts, conv) do
+    case Keyword.fetch(opts, :expected_sandbox_id) do
+      {:ok, expected} -> expected != conv.sandbox_id
+      :error -> false
+    end
   end
 
   defp current_binding?(execution) do
--- a/apps/fountain/lib/fountain/conversations/turn_machine.ex
+++ b/apps/fountain/lib/fountain/conversations/turn_machine.ex
@@ -1693,10 +1693,16 @@ defmodule Fountain.Conversations.TurnMachine do
   reattach. Best-effort by design: the caller runs this on the way out of
   `terminate/2`, and a raise here must not take the rest of that callback
   with it.
+
+  The stopping server is an actor, so it passes the sandbox it was bound to:
+  a parent rebound underneath it belongs to a successor and this recovery
+  writes nothing. `opts` is required rather than defaulted, because a caller
+  that forgets it silently recovers a turn it no longer owns.
   """
-  @spec orphan_on_normal_stop(term(), Conversations.Turn.t() | nil, String.t() | nil) :: :ok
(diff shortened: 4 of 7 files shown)


BinaryBourbon (approved):
APPROVE — the binding is on the right side of the comparison, covers every exit of the orphan transaction, and the two race-order tests fail without it. Findings below are non-blocking; the first is a campaign gap the parent should route somewhere.

## What this changes

`_unsafe_orphan_turn/3` grows a transaction that takes `SELECT … FOR UPDATE` on the conversation row *before* touching the turn, and folds the whole reconciliation into a `with` so a caller-supplied `:expected_sandbox_id` that no longer matches `conversations.sandbox_id` yields `:noop` — no turn write, no `running → idle`, no `reattach/interrupted` stage event, no audit row. `ConversationServer.terminate/2` now supplies `state.sandbox_id` through `TurnMachine.orphan_on_normal_stop/4`, so a stale actor's normal shutdown can no longer close a turn that a newer actor picked up on a different sandbox. Callers that pass no `:expected_sandbox_id` (the reaper) default the expectation to the row's own value and keep their old unbound behaviour.

## Findings

**1. `mark_orphan/3` reaches the same write unbound, and it is an actor path too — non-blocking here, but nothing in the nine-PR stack closes it.**
`apps/fountain/lib/fountain/conversations/conversation_server.ex:1443`

```elixir
```

Four call sites (`:1330` `no_active_session`, `:1339` `list_sessions_failed`, `:1390` `acp_prompt_not_sent`, `:1438` `attach_failed`) hand it a `state` that it discards, and the turn comes from `find_running_turn/1` (`:1445`), which selects the newest `running` turn on the conversation with no reference to the actor's sandbox at all. Scenario: node A is partitioned, its actor for conversation C is inside `reattach_running_turn/1`, and `Managoat.Sandbox.list_sessions/1` is working through `Retry.with_backoff`. Meanwhile C is rebound to sandbox S2 and actor A2 is running turn T. A's retry budget expires → `mark_orphan(state, T, "list_sessions_failed")` → T goes `interrupted` with `orphaned_at` set, C goes `idle`, a `reattach/interrupted` event is published to the live client, and A2 later writes a completion onto an interrupted turn. That is the same failure #1767 describes, on the same function this PR just taught to guard itself.

I checked `origin/codex/p2-1767-machine-gone-senders` (stack tip): `mark_orphan` is still `_state` and still unbound there, so this is not something a later PR in the chain supplies. Not blocking this PR — its stated scope is normal-stop and the gap predates the patch — but the `_state` parameter sitting there unused makes it look handled when it is not, and the campaign should either bind it or say in a comment why the reattach path is deliberately unbound.

**2. The guard's authority lags the new actor by design, so the window is narrowed rather than closed.** `apps/fountain/lib/fountain/conversations.ex:2098`
The expectation is compared against `conversations.sandbox_id`, and `create_fresh_sandbox_and_start/4` (`conversations.ex:4998`) deliberately repoints that column *after* `start_conversation_server/4` returns — the comment there (#717) calls the gap "transient and self-correcting". During it, a stale actor's `expected_sandbox_id` still equals the column, the `with` passes, and the old write lands. Steady state is genuinely fixed; the start-to-repoint window is not. Non-blocking (the fresh-sandbox path orphans the old turn on purpose anyway, so I could not construct a loss from it), but the docstring reads as a total guarantee and every later PR in this stack inherits the same authority — worth one sentence saying what the binding does *not* cover.

**3. An explicit `nil` reads as "expect unbound", not "skip the check".** `apps/fountain/lib/fountain/conversations.ex:2103`
`Keyword.get(opts, :expected_sandbox_id, conv.sandbox_id)` only falls back when the key is *absent*. `state.sandbox_id` comes from `Keyword.fetch!(args, :sandbox_id)` (`conversation_server.ex:463`) and callers pass a nullable `conv.sandbox_id` (`rehydrator.ex:143`), so `expected_sandbox_id: nil` against a bound row would silently make every reconciliation a no-op. I convinced myself it is unreachable today — `_unsafe_list_resumable_conversations/0` inner-joins the sandbox, and a server whose `sandbox_id` is nil stops in `handle_continue(:provision)` before a turn exists — so this is a foot-gun, not a bug. `Keyword.get(opts, :expected_sandbox_id) || conv.sandbox_id` would remove it, and the next six PRs in this stack pass the same field.

**4. No CHANGELOG `[Unreleased]` entry anywhere in the stack.** `git diff origin/main...origin/codex/p2-1767-machine-gone-senders -- CHANGELOG.md` is empty across all nine PRs. This is user-visible: a turn that used to flip to `interrupted` after a sandbox move now stays running, and the conversation stays `running` instead of `idle`. The Unreleased *Changed* section already documents comparable fencing work ("A sandbox reset is fenced until the provider confirms the deletion"). One campaign entry on the last PR is the right shape — per-PR entries for nine units would be noise. Parent's call which PR carries it.

**5. Soft, stacked-PR hygiene.** The new docstring line at `conversations.ex:2083` — "An actor supplies `:expected_sandbox_id`" — is true of exactly one of the five actor call sites as of this PR (finding 1 is the other four). "The conversation server's normal stop supplies…" would be accurate now and stay accurate as the stack lands.

## Checked and clean

- **Comparison direction, both ways.** Stale actor: expected `S1` vs row `S2` → `false` → `:noop`, turn untouched. Current actor: `S1 == S1` → writes. Reaper: key absent → default is the row's own value → always `true`. Right way round.
- **Every exit of the `with`.** Conversation gone → `Repo.one` returns `nil` → `:noop`; binding changed → `:noop`; turn already ended or gone → `{0, _}` → `:noop`, which is exactly the old `count == 0` branch. `{1, _}` is the only success and the `id` predicate makes a count above 1 impossible. One behaviour change: a conversation deleted mid-flight used to raise out of `Repo.get!` and roll back, and is now a clean `:noop` — turns cascade with the conversation, so nothing is lost.
- **Lock ordering, no deadlock inversion.** The new order is conversation-then-turn, the same order as the turn-admission transaction at `conversations.ex:1693`. I grepped every `lock: "FOR UPDATE"` in `apps/fountain/lib` and `ee/lib`: no site locks a turn first and then writes the conversation, so nothing can hold the inverse pair.
- **Unchanged callers.** `autonomous_turn_reaper.ex:69` passes only `:actor`; `_unsafe_orphan_turn/3`'s default arg keeps its arity-2 call shape. `orphan_on_normal_stop/4` keeps the `:normal`-only + `not is_nil(turn)` gate, its `rescue`, and the catch-all clause. Audit and `publish_stage` are still outside the transaction (CLAUDE.md's never-audit-in-a-transaction rule), and the audit records keys and reasons, no values.
- **The size pin.** `conversation_server.ex` goes 2737 → 2741 against `@pin 2762`; the stack tip is 2743 against a lowered `@pin 2759`. Clean on both.
- **The reaper test is the stale actor, not a same-sandbox actor.** `autonomous_turn_reaper_test.exs:150` passes `conv.sandbox_id` from the struct captured *before* `update_conversation`, so the expectation really is the old sandbox. `refute_receive {:log_event, %{stage: "reattach"}}` matches the shape the positive assertion at `:57` uses, so it is a real refutation rather than a mismatched pattern, and `Audit.record/1` is synchronous (`audit.ex:69`, plain `Repo.insert`), so the audit refutation is not racing a supervised task.
- **Revert check on the new isolation test.** With the `FOR UPDATE` and the binding removed, the reassignment-first order never blocks and `await_blocked/2` raises "no PostgreSQL row-lock wait observed"; if it did get past that, `Repo.reload(turn).status == "running"` fails. The orphan-first order fails the same way. Both orders assert on the thing they wait for, and the pause is taken from real Ecto query telemetry inside the real transaction rather than from an outer wrapper — the failure mode called out in the comment at `orphan_binding_isolation_test.exs:19`.
- **The unboxed test harness is the house pattern, not new risk.** `unboxed_run` + `independent/1` + `await_blocked/2` + `hold/1` + manual row cleanup in `after` is lifted from `admission_lock_isolation_test.exs`; the `async: false` is earned (unboxed connections and a globally attached telemetry handler), and the handler is fenced by a process-dictionary flag that it deletes on first match. Audit rows are deleted by `user_id` before the user, which is right given deletion nilifies attribution rather than cascading.

jhgaylor:
**This PR has been reworked, not rebased — the approval above predates the change.**

#1751 ("Fence parent state and orphan recovery by turn generation") landed on `main` after this branch was cut and refactored the same block into `ExecutionGuard._unsafe_recover_turn/2`. The original implementation here rewrote the older inline `Repo.transaction`, so the two are the same function reimagined two ways and a mechanical conflict resolution would have dropped one of them.

**What changed, and why it is now much smaller.**

#1751 did not make this redundant, but it made most of it redundant. `_unsafe_recover_turn/2` already rolls back `:ownership_changed` when `execution && not cleanup_binding?(execution)`, and `cleanup_binding?/1` compares the journal row's `sandbox_id` against the locked parent. For a **bounded** turn that is exactly this PR's check arriving by another route, and `turn_parent_test.exs:312` already asserts it.

The gap is the **unbounded** turn: `_unsafe_create_turn_on_sandbox(attrs, sandbox_id, :unbounded)` creates no `TurnExecution`, and `turns` has no `sandbox_id` column, so nothing records which machine was driving it and `cleanup_binding?/1` has nothing to compare. The stopping actor's `state.sandbox_id` is the only record that exists. That is what this now guards, and only that.

**No structural change was needed.** `_unsafe_recover_turn` already takes `lock_parent(conversation_id)` `FOR UPDATE` as its first statement, so the guard is a comparison against an already-locked row, placed beside #1751's rollback and before `retire_orphan/2` — so a stale actor writes neither the turn nor the journal.

**Two deliberate deviations from the original:**

- Returns `{:error, :ownership_changed}` rather than `:noop` — #1751 established that vocabulary for a stale binding, and a reassigned `conv.sandbox_id` is the same sentence. All three production callers are unaffected.
- `orphan_on_normal_stop/4` takes `opts` with no default, so a caller that forgets which sandbox it held fails to compile rather than silently recovering a turn it no longer owns. `Keyword.fetch/2` rather than `Keyword.get/2` preserves the distinction: an explicit `nil` means "expect no sandbox" and fences; only an absent key means "recovering on nobody's behalf".

**Verification.** Four mutations, each against all three test files (40 tests): guard deleted → 4 failures; explicit `nil` stops fencing → 1; `Keyword.take` → `[]` → 4; server stops passing its binding → 1; restored → 0. Two tests were added because the original pair left the `nil`-expectation semantics asserted only in a docstring with no test behind it. #1751's own fence is untouched — its test files are unmodified apart from those two additions, and 115 of its tests pass.

Re-review requested rather than assumed: the approval above was for a different implementation.

BinaryBourbon (approved):
APPROVE — the central claim holds in both directions, and I confirmed it by execution rather than by reading. The rework is smaller than what it replaces and the part it kept is the part that was actually missing. Findings are non-blocking; the first is a two-line move I would like made here.

## The central claim, verified

I built three probes against this branch and ran them, rather than reasoning from the code:

| Probe | Setup | Result |
|---|---|---|
| **A** | journal row present, parent rebound, **no** `:expected_sandbox_id` | `{:error, :ownership_changed}`; turn stays `running`, conv stays `running` |
| **B** | **no** journal row, parent rebound, **no** `:expected_sandbox_id` | `:ok` — turn went `interrupted`. The hole is real. |
| **C** | `_unsafe_create_turn_on_sandbox(attrs, sandbox.id, 1)` under default config | **no** `TurnExecution` row |

Probe A settles the first half: #1751 genuinely covers the journal case on its own, through `cleanup_binding?/1`'s `parent_matches` half (`conv.sandbox_id == execution.sandbox_id`). Nothing has been narrowed away.

Probe B settles the second: with the journal row absent there is nothing to compare, `turns` carries no `sandbox_id` field (checked the schema and the migrations), and the recovery lands. The stopping actor's `state.sandbox_id` is the only surviving record of which machine was driving that turn, which is exactly what this now supplies.

And the two guards are not one invariant twice. I removed **#1751's** `cleanup_binding?` line and left this PR's `rebound?` in place: `turn_parent_test.exs:338` "a changed binding cannot expire or recover another machine's parent" fails. That test rebinds `sprite_name`, not `conversations.sandbox_id`, so `rebound?` cannot see it. Two comparisons, two different facts — the journal's record of the machine versus the actor's belief about it — that mostly coincide and each catch cases the other cannot. Not a redundant second fence.

## The description undersells this, and misnames the cause

The write-up frames the gap as *"the unbounded turn: `_unsafe_create_turn_on_sandbox(attrs, sandbox_id, :unbounded)` creates no `TurnExecution`"*. That reads as the `capacity` argument, and the capacity argument has nothing to do with it. Whether a journal row exists is decided in `ExecutionGuard._unsafe_register_bounded/3` by `resolve_turn_limits/1` — execution limits, not concurrency. Probe C admits a turn with **integer** capacity `1` and still gets no journal row.

That matters, because `ExecutionLimits.host_ceiling/0` is `Application.get_env(:fountain, :execution_limit_ceiling, %{})` and `enforced_controls/1` returns `[]` — execution limits shipped inert (#1773-#1793). Absent a per-user `execution_limits` or a saved allowance, `resolve_turn_limits/1` returns `%{}` and **every turn in production today takes the no-journal path**. This is not the corner case the docstring makes it sound like; on current configuration it is the only case, and #1751's fence is the one that rarely fires. Please say that in `_unsafe_recover_turn/3`'s docstring rather than "an unbounded turn has no journal row" — someone reading it in six months should not conclude this guard is exotic and delete it.

## Findings

**1. `rebound?/2` is spliced in underneath the comment that documents `cleanup_binding?/1`.** `apps/fountain/lib/fountain/conversations/execution_guard.ex:970`

```elixir
  # A persisted retirement survives parent deletion. The original sandbox row
  # must still prove its tenant/name/provider binding; a surviving conversation
  # must also remain bound to it. Missing or changed sandbox identity stays
  # uncertain. Reset cannot reuse this row while its journal remains open.
  defp rebound?(opts, conv) do
    ...
  end

  defp cleanup_binding?(execution) do
```

Every sentence of that comment is about `cleanup_binding?/1` — tenant/name/provider, the journal row, reset reuse. `rebound?/2` does none of it. On a PR whose whole argument is that these two functions check different things, this is the one comment that must not be misfiled. Move `rebound?/2` above the comment (or below `cleanup_binding?/1`) before merge.

**2. `terminate/2` dereferences `state.sandbox_id` outside the rescue.** `conversation_server.ex:2025`
`orphan_on_normal_stop/4` keeps its `rescue`, and the comment above the call still says "Last and best-effort, so it cannot skip the revocation above" — but `state.sandbox_id` is evaluated at the call site, so a state map without the key raises out of `terminate/2` rather than being swallowed. Only tests build partial states and all three in `autonomous_turn_reaper_test.exs` are updated, so nothing is broken. Noting it because the no-default `opts` decision (which I agree with) pushed a new required field into a callback whose whole design is "must not raise".

**3. Still open from the previous round, both campaign-level, neither this PR's job:**
- `mark_orphan/3` (`reattachment.ex:344`) reaches `_unsafe_orphan_turn/3` with `_state` discarded and no binding. Same failure this PR fixes, four call sites, still unfenced at the stack tip.
- No `[Unreleased]` CHANGELOG entry anywhere in the campaign. `git diff origin/main...HEAD -- CHANGELOG.md` is empty here too. Parent's call which PR carries the one entry.

## Mutations

Reproduced all four counts the description claims, against the three touched test files (40 tests):

| Mutation | Failures | Which |
|---|---|---|
| delete `if rebound?(opts, conv), do: Repo.rollback(...)` | **4** | both new `turn_parent_test` cases, the reaper case, and `reassignment serializes…` |
| `Keyword.fetch/2` → `Keyword.get(opts, :k) \|\| conv.sandbox_id` | **1** | "an actor expecting no sandbox cannot recover a bound parent" |
| `recover_opts = Keyword.take(...)` → `[]` | **4** | as above |
| server passes `[]` instead of its binding | **1** | "an old actor leaves a rebound conversation and its turn untouched" |
| *(mine)* delete #1751's `cleanup_binding?` rollback | **1** | `turn_parent_test.exs:338` — proves the two fences are independent |

The `nil`-versus-absent distinction really is behind a test now, not just a docstring. The previous round's finding 3 is closed.

## Checked and clean

- **The lock is genuinely held at the guard.** `lock_parent/1` is `Repo.one(… lock: "FOR UPDATE")` and is the first statement in the transaction; `lock_execution_by_turn/1` and `lock_turn/1` are locking reads, not writes; `retire_orphan/2` is the first write and both rollbacks precede it. The isolation test proves it against a real backend with `pg_blocking_pids`, not against a same-connection sandbox.
- **The reaper is still the backstop, so a self-fenced actor cannot strand a turn.** `autonomous_turn_reaper.ex:69` omits the key, `rebound?/2` returns `false` on `:error`, and `reap_if_unowned/1` only runs when `whereis/1` is nil — so a turn a successor actually owns is left alone and an unowned one is still reaped. Its `{:error, reason}` clause already existed for #1751 and covers the new rollback.
- **#1751 is untouched.** Of its four test files, only `turn_parent_test.exs` changed, +34/-0, two new tests appended. `turn_parent_actor_test.exs`, `turn_machine_test.exs` and `conversation_server_size_test.exs` are byte-identical to `origin/main`.
- **Callers.** One production caller of `orphan_on_normal_stop/4`, passing `state.sandbox_id`, which is set once at `conversation_server.ex:478` from `Keyword.fetch!` and never rebound in state (no `%{state | sandbox_id: …}` anywhere), so it cannot go stale under its own server. `MachineEvents.reset/6` requires `current_turn: nil`, so it never reaches this path. `Keyword.take/2` in `_unsafe_orphan_turn/3` correctly filters `:actor` out while preserving an explicit `nil`.
- **Size pin.** `conversation_server.ex` is 2509 against `@pin 2549`.

## Runs

Own worktree, own database (`fountain_review_1996`, created and migrated from this tree, dropped after).

- Touched three files ×6 consecutive runs: 40 tests, 0 failures each time. The `async: false` unboxed telemetry test did not flake once.
- Those plus `turn_parent_actor_test.exs` and `turn_machine_test.exs`: 92 tests, 0 failures.
- `test/fountain/conversations` + `test/fountain/workers`: 1,507 tests, 0 failures, 2 skipped.
- Full core suite: 5,792 tests + 6 doctests, 0 failures, 2 skipped.
- `mix format --check-formatted` clean; `mix credo --strict` from the umbrella root, 871 files, no issues; `mix compile --warnings-as-errors --force` clean.

PR #1998, opened 2026-09-11
fix: preserve moved conversations when teardown finishes

jhgaylor:
A conversation reassigned while its old actor cleans up a provider or keeps a shared machine could still be marked terminated when that cleanup finished. The actor now uses a single conditional database update matching both conversation and sandbox IDs. If the binding changed or the conversation was deleted, it stops with `:sandbox_unavailable` and publishes no completed-termination stage. The public client therefore also skips its successful-termination audit.

Both destroy and keep-machine paths use the same final write. Cleanup and retirement of the old sandbox retain their existing behavior.

Stack: based on #1996. Refs #1767. This unit protects the final conversation status. Turn completion/interruption bookkeeping and durable forced-teardown recovery remain separate follow-ups.

Validation:
- 116 focused tests passed, including the existing actor/client/fallback lifecycle behavior and audit guardrails.
- Reassignment during destroy and keep-machine cleanup preserves the replacement binding, its running turn and its ready sandbox. Deletion during provider cleanup returns a refusal instead of crashing.
- A real PostgreSQL race test observes the termination statement waiting on an uncommitted reassignment, then refusing after that reassignment commits.
- The three actor regressions fail with the parent actor implementation. Removing the SQL binding predicate makes the database race test fail; restored code passes.
- Full `mix precommit` passed: Fountain 5,441 tests and 6 doctests, Buzz 144 tests, Support 33 tests; zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -1243,6 +1243,34 @@ defmodule Fountain.Conversations do
   end
 
   @doc """
+  Finish an actor's termination only while the conversation is still bound to
+  its sandbox. The binding check and status write are one database statement,
+  so a reassignment during provider cleanup cannot terminate the new binding.
+
+  The actor owns both IDs. This is internal lifecycle bookkeeping; the public
+  `ConversationServer.terminate_conversation/2` records the action's audit once
+  after a successful reply. A missing or moved conversation returns a refusal.
+  """
+  def _unsafe_finish_conversation_termination(conversation_id, sandbox_id) do
+    query =
+      from(c in Conversation,
+        where: c.id == ^conversation_id and c.sandbox_id == ^sandbox_id,
+        select: c
+      )
+
+    now = DateTime.utc_now() |> DateTime.truncate(:second)
+
+    case Repo.update_all(query, set: [status: "terminated", updated_at: now]) do
+      {1, [conv]} ->
+        broadcast_sidebar_update(conv.user_id)
+        {:ok, conv}
+
+      {0, _} ->
+        {:error, :sandbox_unavailable}
+    end
+  end
+
+  @doc """
   Re-resolve the Agent, Environment and Vault for an existing conversation,
   on the machine it is already running (#1565).
 
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -1904,11 +1904,8 @@ defmodule Fountain.Conversations.ConversationServer do
     # no stop path touches the sprite.
     state = if state.current_turn, do: interrupt_turn(state), else: state
     state = drop_connection(state, "terminated")
-    # ownership: the conditional fence verified this server-owned conversation.
-    conv = Conversations._unsafe_get_conversation!(state.conversation_id)
-    {:ok, _} = Conversations.update_conversation(conv, %{status: "terminated"})
 
-    Output.publish_stage(state.conversation_id, "terminate", "done", %{
+    finish_termination(%{state | handle: nil}, %{
       sandbox: "kept",
       reason:
         if(Lifecycle.home?(state.sandbox_id),
@@ -1916,8 +1913,6 @@ defmodule Fountain.Conversations.ConversationServer do
           else: "held_by_another_conversation"
         )
     })
-
-    {:stop, :normal, :ok, %{state | handle: nil}}
   end
 
   defp terminate_machine(state, sandbox) do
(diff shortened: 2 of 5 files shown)


BinaryBourbon (approved):
APPROVE — the final status write is correctly bound, and the SQL race test proves the property it claims. Four non-blocking notes, three of them about writes in the same path that this unit deliberately leaves alone.

**What this changes**

The two-step `_unsafe_get_conversation!` + `update_conversation(%{status: "terminated"})` at the end of both teardown paths becomes one conditional statement, `Conversations._unsafe_finish_conversation_termination/2`, whose `WHERE id = ? AND sandbox_id = ?` makes the binding recheck and the status write the same UPDATE. `terminate_machine/2` and `terminate_kept_machine/1` now converge on `finish_termination/2`, which publishes the `terminate`/`done` stage and replies `:ok` only when that UPDATE matched a row, and otherwise replies `{:error, :sandbox_unavailable}` — so `do_terminate_conversation/2`'s `audit_lifecycle/5` (which records only on `:ok`) skips the `conversation.terminated` event too. The window closed is the one between the admission fence and the final write: the provider-facing `close_stdin` / `destroy` calls in between, which is precisely where the new tests put the reassignment.

**Findings**

1. *Non-blocking, supplied by #2000 — but nothing here pins it.* On the kept path, `interrupt_turn/1` runs before `finish_termination/2`, and it reaches `TurnMachine.close_interrupted/1` (`apps/fountain/lib/fountain/conversations/turn_machine.ex:693`), which still does an unguarded `_unsafe_get_conversation!` + `update_conversation(%{status: "idle"})`. That write happens *after* `Managoat.Sandbox.close_stdin/1` and `stop_command/1` (conversation_server.ex:2232-2233), i.e. inside the same window this PR closes. Scenario: conversation X on a shared sandbox A with a running turn; `:terminate_conv` arrives; X is rebound to B and starts turn T2 while A's adapter calls are in flight; `close_interrupted/1` writes X `idle` while T2 runs on B, and then `finish_termination/2` correctly refuses — so the row ends up wrong in a different way. `drop_connection/2` → `close_autonomous_turn/2` → `finish_acp_turn/4` has the same shape and is #1999's. Both are named as follow-ups in the body, so this is scope, not a defect. Worth saying explicitly: the new `for cleanup <- [:destroy, :keep]` tests run with `current_turn: nil` in the fixture state, so the running-turn kept path is exercised by nothing in this PR — a case with `current_turn` set belongs in #2000, otherwise the interaction between the two guards is never asserted anywhere.

2. *Non-blocking, and not claimed by any PR in the chain.* `handle_call(:release_conv, ...)` (conversation_server.ex:1609-1615) is the third path that ends a conversation, and it still does the old two-step: `drop_connection(state, "released")` (adapter `close_stdin`/`stop_command`, the same blocking window), then `_unsafe_get_conversation!` + `update_conversation(%{status: "terminated"})`, then an unconditional `terminate`/`done` stage. Same failure as the one you just fixed: a conversation rebound during the adapter close gets marked `terminated` while it is live on a new sandbox, and the SPA is told it ended. `release_conversation/2` is how `Team.open_fresh_conversation/3` hands a machine over, so this is the closest sibling of the path in the patch, and `_unsafe_finish_conversation_termination/2` drops straight in — no new mechanism needed. None of #1999-#2007's titles cover it; worth a tenth unit under #1767 rather than leaving it.

3. *Non-blocking, pre-existing.* Two conversation-scoped side effects in the teardown-finished path are keyed by conversation only, and neither is rechecked: `Egress.release(state.user_id, state.conversation_id)` (conversation_server.ex:2081) → `Broker.release(conv_id)`, which deletes *every* broker session of the conversation, and `Redaction.delete(state.conversation_id)` in `terminate/2` (line 2179). Under the same reassignment a stale actor deletes the replacement's egress credentials (its in-flight turn starts failing outbound until the next prepare re-mints, with `BrokerReaper` covering strays) and, on the same node, the replacement's redaction patterns — after which output is persisted unscrubbed until the next `Redaction.put/2`. Small blast radius and untouched by this patch, but the review question was "every write in the teardown-finished path", and these are the two that are addressed by conversation id rather than by binding.

4. *Soft.* `:sandbox_unavailable` has no `FallbackController` clause, so the refusal leaves `DELETE /conversations/:id/terminate` through the terminal safety net: 422, plus a `fallback: unmapped error atom` warning that reads like a server fault, while the operation declares `service_unavailable: "Sandbox or fleet unavailable"`. The atom was already reachable from the fence commit below this PR, so this is not new — but this PR makes it the common case, and a retry *would* now reach the live server, which is the definition of 503 + `Retry-After` in this controller's vocabulary (`:provisioning`, `:fleet_full`).

**Checked and clean**

- The guard is one statement, so the recheck cannot straddle a commit, and the isolation test asserts the right property: Postgres re-evaluates the `WHERE` against the committed row version after the lock wait, so the UPDATE matches 0 rows. It waits on `pg_blocking_pids` — the actual row-lock wait it depends on, not a proxy — and `async: false` is required by `unboxed_run`.
- `state.sandbox_id` is assigned once in `init/1` (line 463) and never rebound, and `conversations.sandbox_id` is `null: false`, so the equality guard cannot false-refuse a live actor or degenerate into `= NULL`: any rebind means a different server, which is exactly the actor the refusal is aimed at.
- `updated_at` truncated to `:second` matches `timestamps(type: :utc_datetime)`; `select: c` returns the post-update row, so `broadcast_sidebar_update/1` gets the same `user_id` the replaced `update_conversation/2` would have broadcast. Nothing else in that changeset path is lost.
- Refusal is silent in the right places: no `terminate`/`done` stage (asserted as `termination_stages(ctx) == []`), no `conversation.terminated` audit, and the old sandbox's `sandbox.teardown_requested` fence event is untouched — "only record what happened".
- The deleted-during-cleanup test is a real fix, not just a nicer refusal: the old `_unsafe_get_conversation!` raised inside `handle_call` *after* the sprite had been destroyed, taking the reply with it.
- Revert check: all three new actor tests assert on `{:error, :sandbox_unavailable}`, the replacement's `running` status and the absence of stage rows, none of which the old two-step can produce; the SQL test fails with the `c.sandbox_id == ^sandbox_id` predicate removed, as the body claims.
- The isolation test's `after` block cleans its committed rows, and `conversations.sandbox_id`'s `on_delete: :delete_all` makes the delete order safe whichever sandbox the conversation ends up on.
- Audit guardrail: the exclusion entry gives a reason and is accurate — `terminate_conversation/2` records the action once. (Nit: `Fountain.Audit`'s "Deliberately not audited" moduledoc is not extended, though `@deliberately_silent` already carries entries the moduledoc lacks, so "mirrors the moduledoc" was already aspirational.)
- conversation_server.ex goes 2741 → 2750 lines, under the 2762 pin, and the new DB work landed in the context rather than the server.

PR #1999, opened 2026-09-11
fix: preserve turn results after sandbox reassignment or teardown

jhgaylor:
A late actor completion could overwrite an already-ended turn or set a moved/terminated conversation back to idle. Turn completion now carries the actor's sandbox binding and locks the conversation before the current turn. It only writes while the binding still matches, the conversation is nonterminal and the turn is running; the turn result and idle status commit together.

A stale completion clears the actor's local turn state and span without another completion stage or metric. Successful completion still materializes reply text and activates the account after commit. The context reads the current turn under lock so persisted permission and reply fields survive.

Stack: based on #1998. Refs #1767. This unit covers `TurnMachine.finish/4`. The split interrupt path, earlier permission/usage updates and forced-teardown recovery remain separate follow-ups.

Validation:
- All 287 focused tests passed, including turn-machine, real ACP, turn metrics, platform-inference, activation, search, audit guardrails and PostgreSQL race tests.
- Six stale-completion regressions cover reassignment, terminated/failed conversations, completed/interrupted turns with a newer running turn, and deletion. They all fail with the parent finish implementation.
- Real PostgreSQL tests observe row-lock waits in both orders against reassignment. Removing the binding check makes the reassignment-first test fail; restored code passes.
- Reply materialization and account activation remain covered, alongside the server's ACP completion and turn-metric tests.
- Full `mix precommit` passed: Fountain 5,450 tests and 6 doctests, Buzz 144 tests, Support 33 tests; zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/activation.ex
+++ b/apps/fountain/lib/fountain/activation.ex
@@ -19,10 +19,9 @@ defmodule Fountain.Activation do
 
   ## The seam
 
-  `turn_replied/1` is called from `Conversations._unsafe_update_turn/2` —
-  every turn ending in the system goes through there, from the
-  ConversationServer's six endings to the orphan sweep, so a new way for a
-  turn to end is instrumented by construction rather than by remembering.
+  `turn_replied/1` is called after general turn updates, conditional
+  completion and orphan reconciliation materialize a reply. The conditional
+  paths call it after their transaction commits and skip it for stale writes.
   It is the *account's first* reply that this module cares about, so the
   function is a no-op for every later turn.
 
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2278,8 +2278,18 @@ defmodule Fountain.Conversations do
             |> maybe_put_reply_text(current)
 
           updated = Repo.update!(changeset)
-          idle = conv |> Conversation.changeset(%{status: "idle"}) |> Repo.update!()
-          {updated, idle, is_binary(Ecto.Changeset.get_change(changeset, :reply_text))}
+
+          # A successor admitted while this actor was ending its turn keeps the
+          # conversation running. `_unsafe_idle_after_turn/1` refused this
+          # through `latest_turn?`, and that refusal has to survive the move to
+          # a writer of our own; the parent lock above is the same one turn
+          # admission takes, so nothing can be admitted between the two reads.
+          conv =
+            if Repo.exists?(newer_running_turn_query(turn.conversation_id, turn.id)),
+              do: conv,
+              else: conv |> Conversation.changeset(%{status: "idle"}) |> Repo.update!()
+
+          {updated, conv, is_binary(Ecto.Changeset.get_change(changeset, :reply_text))}
         else
           _ -> :noop
         end
@@ -2294,6 +2304,12 @@ defmodule Fountain.Conversations do
         broadcast_sidebar_update(conv.user_id)
         {:ok, updated}
     end
+  end
+
+  defp newer_running_turn_query(conversation_id, turn_id) do
+    from(t in Turn,
+      where: t.conversation_id == ^conversation_id and t.id != ^turn_id and t.status == "running"
+    )
   end
 
   @doc """
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2247,6 +2247,11 @@ defmodule Fountain.Conversations do
   another actor, is a no-op. Reply materialization shares that transaction;
   activation and sidebar publication run after it commits.
 
+  The parent only idles under the two conditions `_unsafe_idle_after_turn/1`
+  idled under: `ExecutionGuard.latest_turn?/2` and a `running` parent. The
+  journal's own condition is not carried here — see the known gap on
+  `Fountain.Conversations.TurnMachine.finish/4`.
+
   `"interrupted"` is a terminal status this writer accepts, because a bounded
   turn that its journal retires comes back through the same ending
   (`BoundedTurn.retire/2` hands `finish/4` the persisted status, ADR 0046).
@@ -2279,15 +2284,18 @@ defmodule Fountain.Conversations do
 
           updated = Repo.update!(changeset)
 
-          # A successor admitted while this actor was ending its turn keeps the
-          # conversation running. `_unsafe_idle_after_turn/1` refused this
-          # through `latest_turn?`, and that refusal has to survive the move to
-          # a writer of our own; the parent lock above is the same one turn
-          # admission takes, so nothing can be admitted between the two reads.
+          # The two preconditions `_unsafe_idle_after_turn/1` idled under, kept
+          # exactly (`ExecutionGuard.parent_write_allowed?/4`, mode `:idle`):
+          # this turn is the conversation's newest generation, and the parent
+          # is `running`. A successor admitted while this actor was ending its
+          # turn therefore keeps the conversation running, and an abandoned
+          # older turn cannot stop this one from idling it. The parent lock
+          # above is the same one turn admission takes, so nothing can be
+          # admitted between this read and the write.
           conv =
-            if Repo.exists?(newer_running_turn_query(turn.conversation_id, turn.id)),
-              do: conv,
-              else: conv |> Conversation.changeset(%{status: "idle"}) |> Repo.update!()
+            if conv.status == "running" and ExecutionGuard.latest_turn?(conv.id, turn.id),
+              do: conv |> Conversation.changeset(%{status: "idle"}) |> Repo.update!(),
+              else: conv
 
           {updated, conv, is_binary(Ecto.Changeset.get_change(changeset, :reply_text))}
         else
@@ -2304,12 +2312,6 @@ defmodule Fountain.Conversations do
         broadcast_sidebar_update(conv.user_id)
         {:ok, updated}
     end
-  end
-
-  defp newer_running_turn_query(conversation_id, turn_id) do
-    from(t in Turn,
-      where: t.conversation_id == ^conversation_id and t.id != ^turn_id and t.status == "running"
-    )
   end
 
   @doc """
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2205,11 +2205,9 @@ defmodule Fountain.Conversations do
   @doc """
   Update a turn's row. When the update ends the turn — its status becomes
   `completed`, `failed` or `interrupted` — the assistant's text for the
-  turn is materialised into `reply_text` in the same write (#826): every
-  turn ending goes through here, from the ConversationServer's six endings
-  to the orphan sweep, so search coverage is by construction rather than
-  by each ending remembering. A turn that already carries a `reply_text`
-  keeps it.
+  turn is materialised into `reply_text` in the same write (#826). Conditional
+  completion and orphan reconciliation also materialize the reply in their
+  transactions. A turn that already carries a `reply_text` keeps it.
 
   The same write is where activation is decided (ADR 0038): a turn that ends
   carrying a reply is handed to `Fountain.Activation.turn_replied/1`, which
@@ -2238,6 +2236,63 @@ defmodule Fountain.Conversations do
 
       {:error, reason} ->
         {:error, reason}
+    end
+  end
+
+  @doc """
+  Complete a running turn only on the actor's current sandbox binding.
+
+  Lock the conversation before the turn, and commit its idle status with the
+  turn's result. A moved or terminal conversation, or a turn already ended by
+  another actor, is a no-op. Reply materialization shares that transaction;
+  activation and sidebar publication run after it commits.
+
+  `"interrupted"` is a terminal status this writer accepts, because a bounded
+  turn that its journal retires comes back through the same ending
+  (`BoundedTurn.retire/2` hands `finish/4` the persisted status, ADR 0046).
+  The two-phase interrupt the server drives itself is a different write, and
+  it keeps the conversation running until the peer has stopped.
+  """
+  def _unsafe_complete_turn(%Turn{} = turn, sandbox_id, status)
+      when status in ["completed", "failed", "interrupted"] do
+    {:ok, result} =
+      Repo.transaction(fn ->
+        conversation_query =
+          from(c in Conversation, where: c.id == ^turn.conversation_id, lock: "FOR UPDATE")
+
+        turn_query =
+          from(t in Turn,
+            where: t.id == ^turn.id and t.conversation_id == ^turn.conversation_id,
+            lock: "FOR UPDATE"
+          )
+
+        with %Conversation{} = conv <- Repo.one(conversation_query),
+             true <- conv.sandbox_id == sandbox_id and conv.status not in ["terminated", "failed"],
+             %Turn{status: "running"} = current <- Repo.one(turn_query) do
+          changeset =
+            current
+            |> Turn.changeset(%{
+              status: status,
+              ended_at: DateTime.utc_now() |> DateTime.truncate(:second)
+            })
+            |> maybe_put_reply_text(current)
+
+          updated = Repo.update!(changeset)
+          idle = conv |> Conversation.changeset(%{status: "idle"}) |> Repo.update!()
+          {updated, idle, is_binary(Ecto.Changeset.get_change(changeset, :reply_text))}
+        else
+          _ -> :noop
+        end
+      end)
+
+    case result do
+      :noop ->
+        :noop
+
+      {updated, conv, reply_materialized?} ->
+        if reply_materialized?, do: Fountain.Activation.turn_replied(updated)
+        broadcast_sidebar_update(conv.user_id)
+        {:ok, updated}
     end
   end
 
--- a/apps/fountain/lib/fountain/conversations/execution_guard.ex
+++ b/apps/fountain/lib/fountain/conversations/execution_guard.ex
@@ -429,7 +429,19 @@ defmodule Fountain.Conversations.ExecutionGuard do
     {decision.turn, changed, event}
   end
 
-  defp latest_turn?(conv_id, turn_id) do
+  @doc """
+  Is this turn the conversation's newest generation?
+
+  The predicate every parent write is gated on: a turn that is not the highest
+  `turn_number` on its conversation has been superseded, so its ending says
+  nothing about whether the conversation is still working. It is deliberately
+  ordered on `turn_number` rather than asking whether some other turn is
+  `running` — an abandoned older turn left `running` must not stop the current
+  generation from idling its parent, and a superseded turn must not idle it
+  even when nothing else is running. Callers hold the parent lock, so admission
+  cannot land between this read and the write it guards.
+  """
+  def latest_turn?(conv_id, turn_id) do
     Repo.one(
       from t in Turn,
         where: t.conversation_id == ^conv_id,
--- a/apps/fountain/lib/fountain/conversations/turn_machine.ex
+++ b/apps/fountain/lib/fountain/conversations/turn_machine.ex
@@ -86,6 +86,7 @@ defmodule Fountain.Conversations.TurnMachine do
 
   @type t :: %__MODULE__{
           conversation_id: String.t() | nil,
+          sandbox_id: String.t() | nil,
           row: Conversations.Turn.t() | nil,
           span: term() | nil,
           metrics: map() | nil,
@@ -95,6 +96,7 @@ defmodule Fountain.Conversations.TurnMachine do
         }
 
   defstruct conversation_id: nil,
+            sandbox_id: nil,
             row: nil,
             span: nil,
             metrics: nil,
@@ -113,6 +115,7 @@ defmodule Fountain.Conversations.TurnMachine do
   def from_state(state) do
     %__MODULE__{
       conversation_id: state.conversation_id,
+      sandbox_id: Map.get(state, :sandbox_id),
       row: state.current_turn,
       span: state.current_turn_span,
       metrics: state.turn_metrics,
@@ -642,54 +645,49 @@ defmodule Fountain.Conversations.TurnMachine do
   the completion metric and the conversation back to idle. What the server
   resolves first (a held permission, parked caller tools, the quiet timer)
   is the server's; what it clears after (activity) is too. Returns the turn
-  with its bookkeeping cleared.
+  with its bookkeeping cleared. A stale completion clears local bookkeeping
+  without overwriting the persisted result or emitting another completion.
   """
   @spec finish(t(), String.t(), map(), map()) :: t()
   def finish(%__MODULE__{} = turn, status, span_attrs, stage_meta) do
     # Before the turn span ends: totals land on it, abandoned tool spans close.
     finalize_tracer(turn.tracer)
 
-    {:ok, row} =
-      Conversations._unsafe_update_turn(turn.row, %{
-        status: status,
-        ended_at: now()
-      })
+    # Ownership: the server supplies its binding, and the context locks and rechecks it.
+    case Conversations._unsafe_complete_turn(turn.row, turn.sandbox_id, status) do
+      {:ok, row} ->
+        stage_meta = Map.merge(stage_meta, %{turn_id: row.id, turn_number: row.turn_number})
 
-    stage_meta = Map.merge(stage_meta, %{turn_id: row.id, turn_number: row.turn_number})
+        # A service-enforced limit sets both: `limit_reason` names which limit,
+        # and `stop_reason` is the field every existing transcript reader
+        # already switches on. The conversations app and the team app live
+        # outside this repo (ADR 0034), so a new field alone would render as an
+        # unexplained failure in both until each one shipped.
+        stage_meta =
+          if row.limit_reason,
+            do:
+              stage_meta
+              |> Map.put(:limit_reason, row.limit_reason)
+              |> Map.put(:stop_reason, row.limit_reason),
+            else: stage_meta
 
-    # A service-enforced limit sets both: `limit_reason` names which limit, and
-    # `stop_reason` is the field every existing transcript reader already
-    # switches on. The conversations app and the team app live outside this
-    # repo (ADR 0034), so a new field alone would render as an unexplained
-    # failure in both until each one shipped.
-    stage_meta =
-      if row.limit_reason,
-        do:
-          stage_meta
-          |> Map.put(:limit_reason, row.limit_reason)
-          |> Map.put(:stop_reason, row.limit_reason),
-        else: stage_meta
+        publish_stage(
+          turn.conversation_id,
+          "turn",
+          # `row.status`, not `status`: a turn the execution journal fenced has
+          # had its requested status dropped, so the persisted row is the only
+          # honest source of what actually happened (ADR 0046). `stage_meta`
+          # already carries turn_id/turn_number from the merge above.
+          if(row.status == "completed", do: "done", else: "failed"),
+          Map.merge(stage_meta, waiting_meta(row))
+        )
 
-    publish_stage(
-      turn.conversation_id,
-      "turn",
-      # `row.status`, not `status`: a turn the execution journal fenced has had
-      # its requested status dropped, so the persisted row is the only honest
-      # source of what actually happened (ADR 0046). `stage_meta` already
-      # carries turn_id/turn_number from the merge above.
-      if(row.status == "completed", do: "done", else: "failed"),
-      Map.merge(stage_meta, waiting_meta(row))
-    )
+        end_span(turn.span, if(row.status == "completed", do: :ok, else: :error), span_attrs)
+        emit_completed(turn, row.status)
 
-    end_span(
-      turn.span,
-      if(row.status == "completed", do: :ok, else: :error),
-      span_attrs
-    )
-
-    emit_completed(turn, row.status)
-
-    {:ok, _} = Conversations._unsafe_idle_after_turn(row)
+      :noop ->
+        end_span(turn.span, :error, %{"outcome" => "completion_ignored"})
+    end
 
     %{turn | row: nil, span: nil, metrics: nil, tracer: nil, session_retry: nil}
   end
--- a/apps/fountain/lib/fountain/conversations/turn_machine.ex
+++ b/apps/fountain/lib/fountain/conversations/turn_machine.ex
@@ -115,7 +115,7 @@ defmodule Fountain.Conversations.TurnMachine do
   def from_state(state) do
     %__MODULE__{
       conversation_id: state.conversation_id,
-      sandbox_id: Map.get(state, :sandbox_id),
+      sandbox_id: state.sandbox_id,
       row: state.current_turn,
       span: state.current_turn_span,
       metrics: state.turn_metrics,
@@ -647,6 +647,16 @@ defmodule Fountain.Conversations.TurnMachine do
   is the server's; what it clears after (activity) is too. Returns the turn
   with its bookkeeping cleared. A stale completion clears local bookkeeping
   without overwriting the persisted result or emitting another completion.
+
+  Known gap: the write goes to `Conversations._unsafe_complete_turn/3`, which
+  takes the parent lock itself rather than going through
+  `ExecutionGuard._unsafe_write_turn/3`, so completion no longer consults the
+  execution journal (#1732). The journal's expiry, `uncertain_spawn` and
+  `:execution_not_started` conditions do not apply at this ending, and a
+  fenced turn does not idle its parent. This is inert while
+  `ExecutionLimits.enforced_controls/1` returns `[]`, because no
+  `turn_executions` row is written at all; re-plumbing completion through the
+  guard is tracked separately.
   """
   @spec finish(t(), String.t(), map(), map()) :: t()
   def finish(%__MODULE__{} = turn, status, span_attrs, stage_meta) do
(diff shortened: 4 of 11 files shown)


BinaryBourbon (approved):
APPROVE — the binding comparison is right, no legitimate completion is dropped, and the one gap it leaves (an abandoned `running` row) is the stack's declared follow-up rather than a defect in this patch. Two non-blocking findings, one of which I'd fix in this PR because it is two characters.

**What this changes**

`TurnMachine.finish/4` stops doing its two writes separately (`_unsafe_update_turn`, then a re-read plus `update_conversation(%{status: "idle"})`) and calls one new context function, `Conversations._unsafe_complete_turn/3`. That function locks the conversation `FOR UPDATE`, then the turn, and writes only while the conversation still names the actor's sandbox, is not `terminated`/`failed`, and the turn is still `running`; the turn result and the `idle` flip commit together, with `reply_text` materialised from the re-read row instead of the machine's in-memory copy. Activation and the sidebar broadcast move after the commit. The actor's binding reaches the machine through a new `sandbox_id` field on the struct, filled by `from_state/1`.

**Findings**

1. *Non-blocking, but please fix here:* `turn_machine.ex:109` reads the binding with `Map.get(state, :sandbox_id)` while every other field on the same struct is a direct access (`state.conversation_id`, `state.current_turn`). `conversations.sandbox_id` is `validate_required` (conversation.ex:136), so a `nil` actor binding matches nothing: if that server field is ever renamed or a caller hands `from_state/1` a partial map, **every** completion becomes a silent `:noop` — no turn ever closes again, and nothing crashes to tell you. That is not hypothetical shape-wise: `reattachment.ex:151` already calls `from_state/1` on a state map built elsewhere, and #2000/#2001/#2003 route three more endings through this same field. `state.sandbox_id` turns that failure into a `KeyError` at the first turn instead.

2. *Non-blocking, and I read the body as already owning it — flagging the cost so it does not stay deferred.* On `:noop` the patch clears `row`/`span` locally and writes nothing. The turn row therefore stays `status: "running"`, `ended_at: nil`, `orphaned_at: nil` — and because `finish/4` nils `row`, `terminate/2`'s safety net cannot see it either: `TurnMachine.orphan_on_normal_stop/4` takes `state.current_turn`, and its own docstring (turn_machine.ex:1549) is "a turn still marked `running` would never be closed by anything else". Concretely, conversation C is rebound from S1 to S2 (wake path repoint, or `follow_cotenants`) while server A on S1 is still alive; A's peer answers, A calls `finish("completed")`; the write no-ops, A clears its row, A later stops and orphans nothing. Turn T is then abandoned `running` for good, and two money paths read it:
   - `Conversations._unsafe_sandbox_busy_elsewhere?/4` (conversations.ex:2036) is true for any co-tenant with `t.status == "running"`, so a live co-tenant server on **S2** never reaches its idle verdict and that machine bills to `SANDBOX_MAX_LIFETIME_HOURS` — the #413 shape, and `SandboxReaper.sweep_abandoned_sandboxes/0` cannot help because it rejects sandboxes with a live server.
   - `SandboxUsage` treats `is_nil(t.ended_at)` as still running and clips the interval to the sandbox's effective end (sandbox_usage.ex:445-467), and the interval is attributed through `c.sandbox_id` — i.e. to **S2**. So `busy_seconds` and `turn_seconds` inflate on the new sandbox while `CreditPricer` charges nothing (it requires `ended_at`, credit_pricer.ex:163): the finance panel over-reports the sandbox the turn never ran on, and the turn time is permanently unbilled.
   Refusing the stale write is right, and `_unsafe_orphan_turn/3` would no-op on the same changed binding, so the recovery genuinely belongs to the "durable forced-teardown recovery" follow-up every PR in the stack names. Two things would make the wait cheap: a `Logger.warning` on the `:noop` arm (right now the only trace is an OTel span attribute, and only when a span exists — you cannot tell in production whether the guard has ever fired, nor find the row), and a note in the follow-up issue that the reconciler has to sweep abandoned `running` rows, not just fence writes.

3. *Soft.* The body says a stale completion "clears the actor's local turn state and span without another completion stage or metric", and the new tests assert `Repo.get(Turn, row.id) == persisted_turn`. On the real `{:done, …}` path `record_usage/2` (turn_machine.ex:410) still writes `usage` to that row immediately before the now-guarded finish, so the persisted row is not untouched — the tests hold only because they call `finish/4` directly. The body does list "earlier permission/usage updates" as a follow-up; the claim is just slightly stronger than the code. It writes the stale actor's own row, not the new binding's turn, so nothing cross-contaminates.

**Checked and clean**

- The comparison is `conv.sandbox_id == sandbox_id` against the actor's `init/1` value, and nothing in `conversation_server.ex` ever reassigns `state.sandbox_id` (only `Keyword.fetch!` at line 463). A new sandbox always means a new server — the reapply/reset (#1565) and persistent-home reset paths keep the row's sandbox — so a matching binding is genuinely "still mine" and a mismatch is genuinely stale. No inverted side.
- The terminal set `["terminated", "failed"]` is exactly the non-live half of `Conversation.@statuses`; `pending` is deliberately still completable, which matters because the reassignment window leaves the row `pending`.
- Nothing the old path did is lost: `update_conversation/2` is a changeset update plus `broadcast_sidebar_update/1`, and both halves are reproduced (the broadcast after commit, not inside it); activation is still the `reply_text` *materialising* write only, and still outside the transaction.
- Which turns get closed and priced is unchanged on every live path — status values are still only `completed`/`failed` (the `when` guard is exhaustive for both `{:done, …}` arms and all six `{:finish, …}` sites), `ended_at` is still set, so `CreditPricer`'s `ended_at`/`orphaned_at` predicates see the same rows. No turn is newly *excluded* while it still closes.
- Lock order is conversation-then-turn, matching `_unsafe_orphan_turn/3`; the only other path that locks a conversation on this flow (`open_turn`, conversations.ex:1720) takes the sandbox advisory lock first and never takes a turn lock, so there is no cycle.
- Re-reading the turn under the lock is a small real improvement: `maybe_put_reply_text/2` now tests the persisted `reply_text`, not the machine's stale `nil`.
- `_unsafe_` prefix, the ownership comment at the call site naming what established it, no audit inside the transaction, and the `audit_guardrail_test` exclusion with a reason — all per CLAUDE.md. `async: false` on the isolation test is justified (`unboxed_run` plus a global telemetry handler), and it matches the two base isolation tests.
- The six stale cases do fail if the guard is reverted (the parent writes status/`ended_at`, and the struct comparison catches it), and the happy-path pair (`reply_text == "hi"`, `onboarding_completed_at`) proves the fixture's binding matches, so the suite is not just asserting no-ops. The Postgres test observes a real `pg_blocking_pids` wait in both orders rather than waiting on a proxy signal.
- Nit on that test: the `try/after` cleanup of the unboxed rows does not run if ExUnit kills the test on timeout, which would leave a user, two sandboxes and a conversation in the shared test DB. Same exposure as the base isolation tests, so not this PR's to fix.
- No CHANGELOG entry — consistent with #1996/#1998, and the fix is invisible in the API surface. Worth one `### Fixed` line under `[Unreleased]` on the last PR of the stack.
- Docstring softening in `activation.ex` and the `_unsafe_update_turn/2` doc describes what is true at this stage of the stack (orphan reconciliation landed in the base PRs), not what a later one adds.

jhgaylor:
**Reworked, not rebased — the approval above predates substantial changes, and three of them are corrections rather than merge text.**

#1751 refactored `finish/4` underneath this branch: `main` now merges `turn_id`/`turn_number` into `stage_meta`, adds `limit_reason`/`stop_reason`, switches on **`row.status` rather than `status`** (ADR 0046 — the journal can drop the requested status), calls `_unsafe_idle_after_turn/1` instead of `update_conversation`, and clears `session_retry`. This PR replaced that whole body with a `case Conversations._unsafe_complete_turn/3`. **Taking either side wholesale silently reverted the other**, so the resolution keeps this PR's `case` and re-homes every one of main's pieces inside its `{:ok, row}` branch.

Three things surfaced that the original review could not have seen, because they only appear against today's `main`:

**1. It was red.** `finish/4` receives `"interrupted"` via `BoundedTurn.retire/2` → `_unsafe_complete/3`. The guard here was `status in ["completed", "failed"]`, so that path raised `FunctionClauseError` — two failures in `bounded_lifecycle_test.exs`. Widened to include `"interrupted"`, matching what `main` does today.

**2. It disarmed one of #1751's guard tests.** `turn_parent_actor_test.exs:45` barriers on `_unsafe_update_turn/2`, which this PR routes around, so its `assert_receive` never fired and the test passed while guarding nothing. The barrier now points at `_unsafe_complete_turn/3` and fails under mutations M3 and M4 — it guards again.

**3. A hole, fixed in its own commit (`689cc8ee`) so it can be reverted alone.** `_unsafe_idle_after_turn/1` refuses to idle a conversation whose newest turn is not the one ending (`latest_turn?/2`). The inline idle here had no equivalent. Measured on this tree:

```
_unsafe_idle_after_turn/1  -> conversation stays "running"
_unsafe_complete_turn/3    -> conversation goes "idle"   (successor still running)
```

Scope, stated precisely: this is **not** every completion. It needs a successor admitted mid-completion — #1751's cancellation-then-admission window. With the refusal removed, exactly one test in the core suite goes red. The refusal is restored, reading under the same parent `FOR UPDATE` lock, and #2000's `_unsafe_idle_interrupted_turn/2` now shares one helper rather than carrying a second copy of the predicate.

Also fixed: `conversation_server_platform_inference_test.exs:318` built a `%TurnMachine{}` with no `sandbox_id`, so its completion silently no-oped.

**Relationship to #1996.** Complementary, not overlapping — #1996 fences orphan *recovery*, this fences turn *completion*, and neither shadows the other. Both close the same underlying gap: `ExecutionGuard._unsafe_write_turn/3` is a pass-through when there is no journal row.

Re-review requested rather than assumed.

BinaryBourbon (changes requested):
**REQUEST CHANGES** — the re-homing kept the stage/limit/`row.status`/`session_retry` pieces, but the `latest_turn?` restoration is not `latest_turn?`. It is a different predicate with a different failure mode, and the direction it gets wrong is the one this PR's own `:noop` arm creates. Everything below is measured on `689cc8ee` against `a130ce33` (`origin/main`), same tree, own database.

## Claim 1 — no, three things were lost

Present and reachable in the `{:ok, row}` branch, confirmed by reading: the `stage_meta` merge of `turn_id`/`turn_number`, `limit_reason` + `stop_reason`, the `if(row.status == "completed", …)` switch in both `publish_stage` and `end_span`, `emit_completed(turn, row.status)`, and the `session_retry: nil` clear on the returned struct. Good.

What is not:

**1a. `_unsafe_idle_after_turn/1`'s `latest_turn?` was replaced, not restored.** Main's predicate is "this turn has the highest `turn_number`". The new one is "no *other* turn in this conversation is `running`". They disagree in both directions. Both probes run `TurnMachine.finish(machine, "completed", …)` on a plain unbounded turn, no journal:

| fixture | `main` | `689cc8ee` |
|---|---|---|
| an **older** turn abandoned `running`, the newest one completes | conv → `idle` | conv → **`running`** |
| a **newer** turn already `completed`, the older one completes | conv → `running` | conv → `idle` |

The first row is the problem, because this PR manufactures its input. The `:noop` arm leaves the stale actor's turn `status: "running"`, `ended_at: nil` forever — the gap the earlier review's finding 2 documents and the stack defers to "durable forced-teardown recovery". Once one such row exists on a conversation, **no later completion can ever idle that conversation again**: the successor turn on the new binding completes, `newer_running_turn_query` finds the abandoned row, and the parent stays `running` permanently. That is the input to `_unsafe_sandbox_busy_elsewhere?/4` (conversations.ex:2036), so the co-tenant machine bills to `SANDBOX_MAX_LIFETIME_HOURS`, and `SandboxReaper` will not touch a sandbox with a live server. Before this PR the deferred gap cost one abandoned row; now it also permanently pins the parent.

`ExecutionGuard.latest_turn?/2` (execution_guard.ex:432) already exists and is what `_unsafe_write_parent` uses. Use that predicate.

**1b. `_unsafe_idle_after_turn/1`'s `conv.status == "running"` precondition is gone.** `parent_write_allowed?(:idle)` requires it (execution_guard.ex:485). Measured: a `pending` conversation completing a turn stays `pending` on main and goes `idle` here. Reaching it needs a `pending` conversation still on the actor's own sandbox, which the reassignment path (conversations.ex:5586 sets `sandbox_id` and `pending` together) does not produce, so I read this as low — but it is a dropped guard, not a no-op.

**1c. `finish/4` no longer consults the execution journal at all.** Main's write went through `_unsafe_update_turn/2` → `ExecutionGuard._unsafe_write_turn/3`; `_unsafe_complete_turn/3` calls `Repo.update!` directly. Consequences, measured at the `finish/4` seam:

- Normal bounded completion leaves the journal `active` where main left it `ready`. On the real actor path `close_bounded_connection` still gets it to `ready`, so `bounded_lifecycle_test`'s `state == "ready"` assertions hold — the end state is the same, the window is new.
- **A journal-fenced turn no longer idles its parent.** `bounded_lifecycle_test.exs:362` ("deadline retirement fences queued actor output"), real actor, only added `IO.puts`: main `conv=idle`, `689cc8ee` `conv=running`. Same `["started", "failed"]` stages either way (the expiry publishes its own), so nothing catches it — the suite is green with the conversation stuck.
- `uncertain_spawn` and the `:execution_not_started` rollback no longer apply at completion.

This half is **inert today**: `ExecutionLimits.enforced_controls/1` returns `[]` (execution_limits.ex:58), so `_unsafe_register_bounded/3` returns `:unbounded` and no `turn_executions` row exists in production. It is a latent revert of #1732/#1773-#1793's machinery, not a live incident. Worth saying out loud in the PR body either way.

**1d. Consequent:** the "`row.status`, not `status`" comment now describes a mechanism this call site no longer has. Nothing between `Turn.changeset` and `Repo.update!` can make them differ, so the ADR 0046 branch is unreachable here. Not a behaviour bug — a comment that will mislead the next reader into thinking the fence is still in the path.

## Claim 2 — the lock is right, the scope claim is not

Right: the `Repo.exists?` runs inside the same transaction, after `Repo.one(conversation_query)` with `lock: "FOR UPDATE"`. Admission takes `FOR SHARE` on the same row (conversations.ex:1819), which `FOR UPDATE` excludes, so nothing can be admitted between the two reads. Lock order conversation→turn matches `_unsafe_write_parent`; no advisory lock is taken here, so there is no cycle with admission's. Ordinary single-turn completions are not refused (`turn.id != ^turn_id`).

Wrong: the blast radius is not "only a successor admitted mid-completion". An *older* abandoned `running` turn triggers it too — see 1a. I could not check the `_unsafe_idle_interrupted_turn/2` sharing; that function does not exist on this branch.

## Claim 3 — correct

`["completed", "failed", "interrupted"]` is exactly `ExecutionGuard.@terminal_turns` (execution_guard.ex:25) and exactly `Turn.@statuses` minus `pending`/`running`. Not wider than what main accepts.

One dead edge: `BoundedTurn.retire/2` hands `finish/4` `decision.turn.status`, and by then `stop/4` has already written the turn `interrupted`, so `%Turn{status: "running"}` fails to match and the call no-ops anyway. Widening the guard turned a `FunctionClauseError` into a no-op; it did not make that path do anything.

## Claim 4 — armed, but weakly, and the story is wrong

It guards. It does not guard what the comment says.

- Removing the successor refusal **alone**: 1 failure, `TurnMachineTest` only. `TurnParentActorTest` stays green, because its old turn is already `interrupted` by the time `finish/4` runs, so the write no-ops before the idle decision is reached.
- Removing the `running`-status requirement **alone**: 2 failures, again not this test.
- Both together: it fails.

And "it passed while guarding nothing" is not what the tree shows. I put the old `stub(Conversations, :_unsafe_update_turn, …)` back on this head unchanged: the test goes **red**, `assert_receive :after_dispatch_before_write` timing out after 2s at line 65. It was a hard failure, not a silent pass. The re-point is the right fix; the reason given for it is not the one the code has.

## Stale `%TurnMachine{}` fixtures — yes, two more

`conversation_server_platform_inference_test.exs:320` is fixed. Also missing `sandbox_id`, and both call `finish/4`:

- `apps/fountain/test/fountain/conversations/turn_parent_test.exs:31` — `defp machine(c)`. The test at :54, "stale completion cannot idle a successor admitted after cancellation", is satisfied by the binding no-op, not by the refusal it names.
- `apps/fountain/test/fountain/conversations/deadline_events_test.exs:150`.

Both stay green when I fill in `sandbox_id: c.conv.sandbox_id`, so no assertion is wrong — but neither is exercising its stated path. `turn_parent_test.exs:54` still does not fail under the successor-refusal mutation even with the binding filled in, for the same already-`interrupted` reason as above.

`conversation_server_platform_inference_test.exs:188` also omits it; that describe block never calls `finish/4`, so it is fine.

## Mutation table

Files: `turn_machine_test`, `turn_completion_binding_isolation_test`, `turn_parent_actor_test`, `turn_parent_test`, `bounded_lifecycle_test`, `conversation_server_platform_inference_test`, `conversation_server_turn_metrics_test` (137 tests, seed 1111).

| mutation | failures |
|---|---|
| baseline | 0 |
| drop `conv.sandbox_id == sandbox_id` | 2 — `TurnMachineTest` *a reassigned completion…*; `TurnCompletionBindingIsolationTest` *reassignment serializes…* |
| drop binding **and** terminal check | 4 — above + `TurnMachineTest` *a terminated completion…*, *a failed completion…* |
| `%Turn{status: "running"}` → `%Turn{}` | 2 — `TurnMachineTest` *a completed completion…*, *a interrupted completion…* |
| successor refusal → `if false` | 1 — `TurnMachineTest` *a successor admitted mid-completion keeps the conversation running* |
| successor refusal **and** running-status both dropped | 4 — above + `TurnParentActorTest` *cancellation and successor admission…* |
| barrier reverted to `_unsafe_update_turn` (lib untouched) | 1 — `TurnParentActorTest`, `assert_receive` timeout |

## Repeat from the last review, still open

`turn_machine.ex:119` reads `Map.get(state, :sandbox_id)` while every neighbouring field is a direct access. With #2000/#2001/#2003 routing three more endings through this field, a rename or a partial map turns every completion in the stack into a silent `:noop` and nothing crashes. `state.sandbox_id` is two characters and makes it a `KeyError` on the first turn.

## Checked and clean

- Size pin: `conversation_server.ex` is 2518 against `@pin 2549`; the PR does not touch `conversation_server_size_test.exs`.
- `mix credo --strict` from the umbrella root: 873 files, no issues. `mix format --check-formatted` in `apps/fountain`: clean.
- Full core suite on this head, own database: 6 doctests, 5806 tests, 0 failures, 2 skipped. Touched files green on three further seeds (1, 424242, 987654). No flakes seen.
- Audit guardrail exclusion carries a reason; no audit inside the transaction; activation and the sidebar broadcast are outside it; `maybe_put_reply_text` reads the row re-read under the lock.

## What would flip this to approve

1a is the only one I would block on: use `latest_turn?`'s predicate (or `ExecutionGuard.latest_turn?/2` directly) so an abandoned older `running` row cannot pin the parent. 1d and the `Map.get` are cheap and belong here. 1b, 1c and the two stale fixtures I would take as a follow-up issue with the measurements above, provided 1c is written into the PR body rather than left implicit — the bounded path being inert is what makes it survivable, and that fact is not recorded anywhere in this PR.

jhgaylor:
All CHANGES_REQUESTED findings addressed. Head is `29ba8ac8`; three commits on top of the previous head.

**1a — the blocker.** The reviewer is right that the restoration was not `latest_turn?`. `main`'s predicate is "this turn has the highest `turn_number`"; what shipped was "no other turn is `running`", and they disagree in both directions. Replaced with `ExecutionGuard.latest_turn?/2` — the predicate `_unsafe_write_parent` already uses, promoted from `defp` with a docstring rather than a third implementation — and `newer_running_turn_query/2` deleted. All three probes now match `main`:

| fixture | `main` | before | after |
|---|---|---|---|
| an **older** turn abandoned `running`, the newest completes | `idle` | **`running`** | `idle` |
| a **newer** `completed` turn, the older one completes | `running` | **`idle`** | `running` |
| a `pending` conversation completes a turn | `pending` | **`idle`** | `pending` |

**1b.** `conv.status == "running"` restored. The idle arm is now exactly `parent_write_allowed?(:idle)`'s two non-journal conditions.

**1d.** The `row.status` comment rewritten to describe the mechanism that is actually there, and to point at the known gap, rather than claiming a journal fence this call site no longer has.

**Stale fixtures.** `turn_parent_test.exs:31` and `deadline_events_test.exs:150` bound. Also `from_state/1` now uses `state.sandbox_id` rather than `Map.get/2`, so a missing binding is a `KeyError` on the first turn instead of a silent no-op on every completion.

**`turn_parent_test.exs:54`.** The reviewer is right that filling in the binding does not save it — `successor/1` interrupts before admitting, so the old turn is already `interrupted` when `finish/4` runs and the successor refusal is unreachable from that fixture. Renamed to *"a completion whose turn another actor already ended writes nothing"*, with an explicit assertion making the mechanism visible, and it now dies under the turn-status mutation alone. The refusal it used to claim is covered by three new tests, one per condition.

**The strongest evidence is the reverse mutation.** Reverting to the predicate this PR originally shipped now fails three tests where it previously failed none.

**Barrier test, individually — partly, stated plainly.** It now dies under the turn-status mutation alone (it did not before; it never asserted the old actor's own turn result, which is the thing this PR is named for). It still needs the combined mutation for `latest_turn?`, and that was left deliberately: forcing a single-guard death means dropping the cancellation from the setup, at which point the test is no longer "cancellation and successor admission after dispatch" and becomes a slower duplicate of `TurnMachineTest`'s *a successor admitted mid-completion*, which already covers that shape and dies under that exact mutation. Every guard has at least one test that dies when it alone is removed.

**Runs:** 17 clean runs of the touched set across six seeds, plus two full core suite runs (5,809 tests) on an isolated database. No flakes.

---

> **Known gap: completion no longer consults the execution journal.** On `main` the completion write went through `_unsafe_update_turn/2` → `ExecutionGuard._unsafe_write_turn/3`. `_unsafe_complete_turn/3` takes the parent lock itself and calls `Repo.update!` directly, so the journal's conditions no longer apply at this ending: deadline expiry, `uncertain_spawn` and the `:execution_not_started` rollback are not evaluated, a normal bounded completion leaves the journal `active` where main left it `ready` (the real actor path still reaches `ready` via `close_bounded_connection`, so the end state matches and only the window is new), and a journal-fenced turn no longer idles its parent — measured at `bounded_lifecycle_test.exs:362`, where `main` leaves `conv=idle` and this branch leaves `conv=running`, with the same `["started", "failed"]` stages and the suite green either way. This is a latent revert of #1732/#1773-#1793's machinery rather than a live incident: `ExecutionLimits.enforced_controls/1` returns `[]` (`execution_limits.ex:58`), so `_unsafe_register_bounded/3` returns `:unbounded` and no `turn_executions` row is written in production at all. Re-plumbing completion through the guard is a separate change and is not attempted here; the gap is recorded in the docstring on `TurnMachine.finish/4` so the next reader is not misled about which fences are in the path.

That gap is deliberately out of scope for this PR and will be filed as a follow-up issue.

BinaryBourbon (approved):
**APPROVE.** The blocker is fixed and I verified it independently rather than from the summary. All three 1a probes now match `main`, the reverse mutation costs three tests, and the two conditions are separately covered. Nothing the three commits added is broken, and the one remaining divergence class is the one you documented and scoped out — I agree with that call, with one line of caution for #2000.

Measured on `29ba8ac8` against `a130ce33` (`origin/main`), same probe file run against both trees, own database.

## 1. The 1a probes — confirmed, all three match `main`

| fixture | `main` | `689cc8ee` | `29ba8ac8` |
|---|---|---|---|
| **older** turn abandoned `running`, newest completes | `idle` | **`running`** | `idle` ✔ |
| **newer** turn already `completed`, older completes | `running` | **`idle`** | `running` ✔ |
| `pending` parent completes a turn | `pending` | **`idle`** | `pending` ✔ |
| successor `running` at a higher `turn_number` | `running` | `running` | `running` ✔ |
| plain single-turn completion | `idle` | `idle` | `idle` ✔ |
| already-`idle` parent | `idle` | `idle` | `idle` ✔ |

The dangerous direction — an abandoned older `running` row permanently pinning the parent, which the `:noop` arm manufactures — is gone. Using `ExecutionGuard.latest_turn?/2` rather than a third copy is the right shape; the idle arm is now literally `parent_write_allowed?(:idle)`'s two non-journal conditions, which is easy to check by eye against execution_guard.ex:485.

## 2. The reverse mutation — confirmed, and it decomposes cleanly

Restoring exactly the expression `689cc8ee` shipped:

| mutation | failures |
|---|---|
| revert to the originally-shipped predicate verbatim | **3** — `TurnParentTest` *an abandoned older turn cannot stop the newest one idling the parent*, *a superseded turn cannot idle the parent once its successor has ended too*, *a completion cannot idle a parent that is not running* |
| drop only `ExecutionGuard.latest_turn?/2` | 2 — *a superseded turn…*; `TurnMachineTest` *a successor admitted mid-completion…* |
| drop only `conv.status == "running"` | 1 — *a completion cannot idle a parent that is not running* |
| swap `latest_turn?` for the `newer_running_turn` predicate, keeping the `running` check | 2 — *an abandoned older turn…*, *a superseded turn…* |

Three failures for the full revert, and each half dies on its own. That is the evidence I wanted.

Full table on this head, 192 tests, seed 1111, baseline 0:

| mutation | failures |
|---|---|
| drop `conv.sandbox_id == sandbox_id` | 2 — `TurnMachineTest` *a reassigned completion…*; `TurnCompletionBindingIsolationTest` *reassignment serializes…* |
| drop binding **and** terminal check | 4 — above + *a terminated completion…*, *a failed completion…* |
| `%Turn{status: "running"}` → `%Turn{}` | **4** — `TurnMachineTest` *a completed completion…*, *a interrupted completion…*; `TurnParentTest` *a completion whose turn another actor already ended writes nothing*; **`TurnParentActorTest`** *cancellation and successor admission…* |

That last row is the change I asked about: on `689cc8ee` the turn-status mutation cost 2 tests and `TurnParentActorTest` survived. The added `assert Repo.get!(Turn, turn.id).status == "interrupted"` puts it on the hook by itself. Every guard in `_unsafe_complete_turn/3` now has at least one test that dies when it alone is removed.

## 3. The barrier test trade — accepted

Take it as it stands. `latest_turn?` is not unguarded: `TurnMachineTest`'s *a successor admitted mid-completion keeps the conversation running* dies under that mutation alone, and `TurnParentTest`'s *a superseded turn…* dies under both revert shapes. Forcing `TurnParentActorTest` to die on `latest_turn?` alone means removing the cancellation from its setup, and the cancellation is the thing that test is named for — you would be trading a real concurrency shape for a slower duplicate of coverage that already exists. The reasoning in the comment is the right one and it is written down where the next reader will find it, which is what makes a partial guard acceptable rather than a hole.

## 4. Nothing broken by the three commits

- Full core suite, isolated database, seed 424242: **6 doctests, 5809 tests, 0 failures, 2 skipped** (5806 before; the three new `TurnParentTest` cases account for the delta).
- Touched set (`turn_machine`, `turn_completion_binding_isolation`, `turn_parent_actor`, `turn_parent`, `bounded_lifecycle`, `deadline_events`, `execution_guard`, `conversation_server_platform_inference`, `audit_guardrail`) green on seeds 5, 90210, 31337 — 265 tests each. No flakes.
- `mix credo --strict` from the umbrella root: 873 files, no issues. `mix format --check-formatted` in `apps/fountain`: clean.
- Size pin unchanged at 2518 against `@pin 2549`; `conversation_server_size_test.exs` still untouched.
- `from_state/1`'s switch to `state.sandbox_id` introduces no new crash: the one caller outside the machine, `reattachment.ex:151`, is handed the real server state, and `sandbox_id` is set from `Keyword.fetch!(args, :sandbox_id)` at conversation_server.ex:478, so the key is always present. Promoting `latest_turn?/2` changed no existing call site.
- The unbounded interrupt path is untouched — `close_interrupted/1` still writes through `_unsafe_idle_after_turn/1` (turn_machine.ex:755), so nothing but `finish/4` moved off the guard.

## 1c — I agree it is out of scope, with one caveat for #2000

The gap is real, correctly measured, correctly scoped, and now written in two places where a reader will hit it. The inertness argument holds: `ExecutionLimits.enforced_controls/1` returns `[]`, `_unsafe_register_bounded/3` returns `:unbounded`, no `turn_executions` row exists in production. Filing it rather than re-plumbing completion inside a binding PR is the right split.

One thing to carry into that issue, because it is *not* journal-dependent: the `%Turn{status: "running"}` requirement means **any** turn already ended by another writer leaves the parent `running` where `main` idled it. I probed it with no journal at all — main `conv=idle`, this branch `conv=running`. Today the only producers of that precondition are the journal (inert) and the split interrupt, and the split interrupt still has its own writer, so nothing is stuck in production. But #2000 introduces `_unsafe_idle_interrupted_turn/2` on exactly this shape, and if it inherits the same "already terminal, therefore write nothing" reading without inheriting an idle decision, that becomes a live stuck-`running` conversation. Worth a sentence in the follow-up so it is not rediscovered from a support ticket.

## Nit, non-blocking

`ExecutionGuard.latest_turn?/2` is now public and unscoped while every other public unscoped predicate in the same module carries the prefix — `_unsafe_open_execution?/1` (:75), `_unsafe_sandbox_open?/1` (:865), `_unsafe_fenced?/1` (:872). CLAUDE.md's rule is "the prefix goes on *every* unscoped function, including ones whose callers happen to check ownership first". It reveals nothing and both callers hold the parent lock, so this is cosmetic — but `_unsafe_latest_turn?/2` would keep the module reading uniformly, and the docstring already says the caller must hold the lock.

Good rework. The corrections in the summary comment are accurate, including the two on my own earlier framing.

BinaryBourbon (commented):
**Addendum to my approval — activation checked. Verdict unchanged: APPROVE.** Activation is correct on every path this PR is for, and the moduledoc weakening is honest about drift that started elsewhere. One finding to fold into the follow-up, and a concrete guardrail suggestion.

Measured on `29ba8ac8` against `a130ce33`, same probe file run against both trees, own database.

## 1. Activation is correct

The new call site's predicate is byte-identical to `_unsafe_update_turn/2`'s — `is_binary(Ecto.Changeset.get_change(changeset, :reply_text))` — computed inside the transaction on the row re-read under the lock, applied after commit. `maybe_put_reply_text/2` is the same function with the same `reply_text: nil` guard. No divergence to find by reading; measured anyway, on a fresh unactivated account with one assistant chunk in `log_events`:

| scenario | `main` | `29ba8ac8` |
|---|---|---|
| first reply, happy path | activated | activated ✔ |
| first reply on a **superseded** turn (parent deliberately not idled) | activated | activated ✔ |
| turn already carries `reply_text` | not activated | not activated ✔ |
| no assistant text at all | not activated | not activated ✔ |
| conversation **reassigned** to another sandbox | activated | not activated ✔ |
| conversation **terminated** | activated | not activated ✔ |
| **turn already ended by another writer** | **activated** | **not activated** ⚠ |

The two "stale" rows are the fix working, not a loss. In the reassignment probe `main` does not merely activate — it writes `status: "completed"`, `reply_text: "hi"` onto a turn whose conversation has moved to another sandbox. That is the defect #1999 exists to stop, and dropping the activation with it is right.

Row 2 matters more than it looks: activation and the idle decision are now independent, so a completion that correctly refuses to idle the parent still activates the account. That is the behaviour you want and it is not obvious from the code.

**Mutation:** delete the `if reply_materialized?, do: Fountain.Activation.turn_replied(updated)` line — exactly one test dies, `TurnMachineTest` *completion materializes a reply and activates the account*. The new site is guarded.

## 2. The ⚠ row — one more cost on the gap you already scoped out

A turn already ended by another writer, same binding, conversation not moved: `main` materialises the reply and activates; this branch does neither. Same precondition as the stuck-`running` finding, same reachability (journal expiry, inert; and #2000's split interrupt), so nothing is lost in production today.

The reason to record it rather than wave it through: **it does not self-heal.** `_unsafe_backfill_reply_texts/0` writes `reply_text` and never calls `turn_replied/1`, so a reply recovered later by the backfill leaves the account unactivated for good. And `maybe_put_reply_text/2` only re-derives on a *later write to that turn*, which for an already-terminal row may never come. If that turn was the account's first reply, a real user silently never counts as activated.

Put it in the same follow-up as the idle decision, not a separate one — it is the same `%Turn{status: "running"}` condition producing a second consequence, and whoever fixes one should see both.

## 3. The doc weakening — acceptable, because the choke point was already gone

I would not re-plumb activation in this PR. The old claim was the right architecture, but restoring it means putting completion back through `_unsafe_update_turn/2`, which is exactly the re-plumbing you scoped out as the 1c follow-up. Two changes of that size inside a binding fix is how the last two rounds happened.

**When the drift started:** `86a0a255` (#1416) introduced activation with one call site and the by-construction moduledoc. `93cbee9a` — *fix(conversations): recover orphaned running turns (#1197)*, #1473, 2026-09-03 — added the orphan-reconciliation site at `conversations.ex:2300` and did not touch `activation.ex`. So the moduledoc has claimed a single choke point while two existed for nine days. #1999 is the PR that noticed and wrote down what is actually true. Your framing is right on this one.

## 4. A guardrail — yes, and I would make it stricter than a count

Nothing pins this today. `activation_test.exs` unit-tests `turn_replied/1` and `first_reply_at/1`; nothing walks the endings, which is why #1473's drift went unnoticed.

A call-site count would have caught #1473, but it would not catch the hole in §2, because the backfill's problem is a writer that *should* call `turn_replied/1` and does not. The `audit_guardrail_test.exs` shape generalises better: enumerate the functions that can set `reply_text` from `nil` — today `_unsafe_update_turn/2`, `_unsafe_complete_turn/3`, `_unsafe_orphan_turn/3` and `_unsafe_backfill_reply_texts/0` — and fail unless each either calls `turn_replied/1` or is on a documented exclusion list with a reason. The backfill has a legitimate reason (a one-time release task over historical rows, where activation timestamps would be wrong anyway) and would go on the list explicitly rather than by omission. That is one small test file and it makes the next `_unsafe_*_turn` writer in this campaign — #2000, #2001, #2003 all add one — fail until someone decides.

Out of scope here. Worth its own issue alongside the 1c follow-up.

## Unchanged

Everything in my approving review stands: the 1a probes match `main`, the reverse mutation costs three tests, every guard has a test that dies alone, full core suite 5809/0, credo and format clean, size pin 2518/2549.

PR #2000, opened 2026-09-11
fix: preserve newer work across both interrupt bookkeeping phases

jhgaylor:
A stale interrupt could overwrite a turn after reassignment, and its later idle write could overwrite a conversation whose next turn had already started. Both interrupt bookkeeping phases now check the actor's sandbox binding. The first phase conditionally marks a running turn interrupted while retaining the conversation status; the second locks the parent and checks the interrupted turn plus the absence of newer running work before idling it.

The existing adapter/peer shutdown order is preserved. A stale mark publishes no interruption stage or completion metric. A successful mark records its completion metric once even if reassignment or new work subsequently prevents the idle write. Reply materialization and activation use the same guarded transaction helper as normal completion.

Stack: based on #1999. Refs #1767. Transport-failure bookkeeping and durable forced-teardown recovery remain separate follow-ups.

Validation:
- All 309 focused tests passed, covering the turn machine, real ACP flows, metrics, platform inference, activation, search, termination actors and audit guardrails.
- Eight regressions cover stale marks and changes between interrupt halves, including reassignment, terminal/deleted conversations and newer running work.
- Two real PostgreSQL tests exercise actual bounded/unbounded turn admission while final interrupt bookkeeping waits on the parent lock. The newly admitted turn and conversation both remain running.
- Substituting the original unconditional writes produced the ten expected regression failures. Restoring the guards passed the tests.
- Full `mix precommit` passed: Fountain 5,460 tests and 6 doctests, Buzz 144 tests, Support 33 tests; zero failures.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2341,9 +2341,19 @@ defmodule Fountain.Conversations do
   and `ExecutionGuard._unsafe_recover_turn/3` returns `:noop` for one, so the
   recovery backstop every other fence in #1767 leans on does not cover this
   state. Refusing to write here is therefore not "leave it to the next
-  writer"; it is a conversation stuck `running` with nothing running under it,
-  and `_unsafe_sandbox_busy_elsewhere?/4` reads that parent, so a shared
-  machine bills to `SANDBOX_MAX_LIFETIME_HOURS`.
+  writer"; it is a conversation left `running` with nothing running under it,
+  and no sweep that will ever notice.
+
+  That is the whole harm, and it is worth being exact about its edges, because
+  two larger claims about it are false. `_unsafe_sandbox_busy_elsewhere?/4`
+  does **not** read the parent's status — it queries co-tenant `turns` rows and
+  conversation `updated_at` — so a stuck parent does not by itself hold a
+  shared machine to `SANDBOX_MAX_LIFETIME_HOURS`. And the obvious producer,
+  `follow_cotenants/2`, largely rescues itself: it casts `:machine_gone`
+  *before* its `update_all`, and `MachineEvents.gone/4` idles a `running`
+  parent unconditionally. The state is still reachable — a cross-pod co-tenant
+  that Horde's `whereis` misses gets the rebind and no cast — but that is a
+  narrower window than "any sprite replacement".
 
   Two conditions gate the write, the same two `_unsafe_complete_turn/3`
   applies: `ExecutionGuard.latest_turn?/2` and a `running` parent. A turn
@@ -2351,6 +2361,17 @@ defmodule Fountain.Conversations do
   abandoned older turn left `running` does not stop this one idling it. The
   lock matches turn admission, so a concurrent admission cannot slip between
   this read and the write.
+
+  Deliberately **no sandbox-binding condition**, unlike every other #1767
+  fence. The binding is still checked — but by the process rather than the
+  query, and more tightly: `interrupted?` appears in neither `from_state/1`
+  nor `into_state/2`, so it cannot survive a mailbox round-trip, and the two
+  halves are one synchronous body separated only by `stop_acp_peer/1`, a
+  bounded one-second `GenServer.stop`. `state.sandbox_id` is written once at
+  init. A stale actor therefore cannot reach here at all, whereas an SQL
+  binding check *would* refuse the legitimate rebind-between-halves case and
+  strand the parent. Note the cost of that reasoning: it holds because of this
+  one caller's shape, and nothing enforces that a second caller keeps it.
 
   The actor's sandbox binding is deliberately **not** a third condition, which
   is why this takes no `sandbox_id`. `_unsafe_complete_turn/3` and
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2334,57 +2334,43 @@ defmodule Fountain.Conversations do
   @doc """
   Release the conversation `TurnMachine.mark_interrupted/1` left running.
 
-  The second half of the two-phase interrupt. The first half retired the turn
-  and deliberately left the parent `running` so the peer could stop under it,
-  which makes this the **only** writer that will ever idle it: a turn that is
-  no longer `running` is invisible to `AutonomousTurnReaper.sweep_stuck_turns/0`,
-  and `ExecutionGuard._unsafe_recover_turn/3` returns `:noop` for one, so the
-  recovery backstop every other fence in #1767 leans on does not cover this
-  state. Refusing to write here is therefore not "leave it to the next
-  writer"; it is a conversation left `running` with nothing running under it,
-  and no sweep that will ever notice.
+  Caller requirement: call only from `TurnMachine.close_interrupted/1` after
+  its matching `mark_interrupted/1` successfully retired the turn. The public
+  `_unsafe_` helper does not enforce this requirement for another caller.
 
-  That is the whole harm, and it is worth being exact about its edges, because
-  two larger claims about it are false. `_unsafe_sandbox_busy_elsewhere?/4`
-  does **not** read the parent's status — it queries co-tenant `turns` rows and
-  conversation `updated_at` — so a stuck parent does not by itself hold a
-  shared machine to `SANDBOX_MAX_LIFETIME_HOURS`. And the obvious producer,
-  `follow_cotenants/2`, largely rescues itself: it casts `:machine_gone`
-  *before* its `update_all`, and `MachineEvents.gone/4` idles a `running`
-  parent unconditionally. The state is still reachable — a cross-pod co-tenant
-  that Horde's `whereis` misses gets the rebind and no cast — but that is a
-  narrower window than "any sprite replacement".
+  The binding check belongs to that successful mark: `_unsafe_interrupt_turn/2`
+  checks the actor's sandbox binding under the parent lock, and only success
+  sets `interrupted?`. `close_interrupted/1` calls this helper only when that
+  flag is true. Neither `from_state/1` nor `into_state/2` carries the flag, so
+  it cannot survive a mailbox round-trip. `ConversationServer.interrupt_turn/1`
+  runs both halves synchronously, separated only by `stop_acp_peer/1`, whose
+  `GenServer.stop/3` has a one-second timeout. The server's `sandbox_id` is set
+  at init and never changed. An actor already stale at the mark cannot reach
+  this write; a rebind after a successful mark can still reach it.
 
-  Two conditions gate the write, the same two `_unsafe_complete_turn/3`
-  applies: `ExecutionGuard.latest_turn?/2` and a `running` parent. A turn
-  superseded while the peer was stopping keeps the conversation running; an
-  abandoned older turn left `running` does not stop this one idling it. The
-  lock matches turn admission, so a concurrent admission cannot slip between
-  this read and the write.
+  This half deliberately takes no `sandbox_id` and does not repeat the binding
+  check. It writes no turn result. Under the parent and turn locks, it requires
+  a `running` parent, an `interrupted` turn and `ExecutionGuard.latest_turn?/2`.
+  A successor admitted while the peer was stopping prevents the idle write;
+  an older turn left `running` does not. Admission inserts its turn and sets
+  the parent `running` in the same transaction under the same parent lock, so
+  it cannot slip between this check and the write.
 
-  Deliberately **no sandbox-binding condition**, unlike every other #1767
-  fence. The binding is still checked — but by the process rather than the
-  query, and more tightly: `interrupted?` appears in neither `from_state/1`
-  nor `into_state/2`, so it cannot survive a mailbox round-trip, and the two
-  halves are one synchronous body separated only by `stop_acp_peer/1`, a
-  bounded one-second `GenServer.stop`. `state.sandbox_id` is written once at
-  init. A stale actor therefore cannot reach here at all, whereas an SQL
-  binding check *would* refuse the legitimate rebind-between-halves case and
-  strand the parent. Note the cost of that reasoning: it holds because of this
-  one caller's shape, and nothing enforces that a second caller keeps it.
+  Rechecking the binding here could leave the parent `running` after the
+  first half retired its last running turn. `AutonomousTurnReaper.sweep_stuck_turns/0`
+  selects running turns, and `ExecutionGuard._unsafe_recover_turn/3` returns
+  `:noop` for a retired turn, so those recovery paths cannot repair that state.
+  Other paths can idle the parent: `MachineEvents.gone/4` and the lifecycle's
+  park and reclaim paths do so when they run.
 
-  The actor's sandbox binding is deliberately **not** a third condition, which
-  is why this takes no `sandbox_id`. `_unsafe_complete_turn/3` and
-  `_unsafe_fence_sandbox_for_teardown/2` fence on the binding because they
-  write a *result*, and a stale actor must not overwrite the owner's. This
-  writes no turn row at all. Its only write releases a `running` status that
-  this actor itself set moments ago and that nothing else can clear, and it
-  writes it only when the locked parent proves there is no work: the turn is
-  terminal and `latest_turn?/2` says no successor has been admitted. A
-  conversation can be rebound under a live server without any actor
-  changing — `follow_cotenants/2` moves every co-tenant of a replaced sprite
-  with one unlocked `update_all` (ADR 0023 gate 5) — so a binding condition
-  here would strand exactly the conversations that shared a machine.
+  `follow_cotenants/2` sends `:machine_gone` before rebinding with `update_all`,
+  so a server that receives the cast has `gone/4`'s cleanup as a backstop.
+  `tell_cotenants/3` skips the cast when `ConversationServer.whereis/1` misses,
+  including a cross-pod registry miss, while the rebind still applies. The
+  parent can then remain `running` until another cleanup path runs.
+  `_unsafe_sandbox_busy_elsewhere?/4` reads co-tenant turns and `updated_at`
+  for conversations without turns, not the parent's status; a stuck parent
+  alone does not keep the shared sandbox busy or extend its billing lifetime.
 
   Known gap: a turn no longer `interrupted` writes nothing, and nothing
   overwrites a terminal turn today, so that arm has no live producer (#2054
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2306,7 +2306,7 @@ defmodule Fountain.Conversations do
           # admitted between this read and the write.
           #
           # The interrupt path passes `idle?: false` and idles later, through
-          # `_unsafe_idle_interrupted_turn/2`, which asks the same two
+          # `_unsafe_idle_interrupted_turn/1`, which asks the same two
           # questions once the peer has stopped.
           updated_conv =
             if idle? and conv.status == "running" and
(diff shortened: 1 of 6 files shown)


BinaryBourbon (approved):
APPROVE — both persisted phases are genuinely guarded, the window between them is closed by the right lock, and the tests fail on the pre-fix writes. Three non-blocking notes below, one of which I'd like the author to consider before the stack lands.

## What this changes

`mark_interrupted/1` stops writing the turn row through `_unsafe_update_turn/2` and goes through a new `_unsafe_interrupt_turn/2`, which is `_unsafe_complete_turn/3`'s transaction refactored into a shared `end_running_turn/4` with an `idle?` flag: conversation `FOR UPDATE`, then turn `FOR UPDATE`, binding equality, non-terminal conversation, turn still `running` — and for the interrupt case it deliberately leaves the conversation's status alone so the peer can still be stopped. The result is carried forward on a new transient `interrupted?` field, and `close_interrupted/1` only emits the completion metric and calls the new `_unsafe_idle_interrupted_turn/2` when the mark actually applied. Phase two re-locks the parent, re-checks the binding and terminal status, requires the turn to still be `interrupted`, and refuses to idle when a `running` turn exists.

I verified there are exactly two persisted phases. Between them `interrupt_turn/1` runs only `stop_acp_peer/1` (`Connection.stop_peer/1`: demonitor, `GenServer.stop(peer, :normal, 1_000)` — no row write, but up to a second of real wall clock, so the window is not theoretical). `finish_runner_reconnect/2` ahead of phase one only publishes a stage. And the window cannot be used to rebind and then write against the new binding: `_unsafe_create_turn_on_sandbox/4` takes `lock: "FOR UPDATE"` on the same conversation row (`apps/fountain/lib/fountain/conversations.ex:1724`), so phase two's parent lock really does serialize against admission rather than merely reading before it — which is what `interruption_admission_isolation_test.exs` proves with `pg_blocking_pids`.

## Findings

**1. Non-blocking, but worth a line of code: phase two's "newer work" check is not scoped to newer work.** `apps/fountain/lib/fountain/conversations.ex:2186-2189` — `running_query` matches *any* `running` turn on the conversation, not one newer than the turn being closed. Failure scenario: conversation C is rebound from sandbox A to B while turn 4 is running on A. A's server later interrupts; phase one no-ops on the binding check, so **turn 4's row stays `running` forever** from this actor. B's server runs turn 5; the user interrupts it. Phase one marks turn 5 `interrupted`; phase two finds turn 4 still `running`, no-ops, and C stays `running` with nothing running on it. It self-heals — `_unsafe_orphan_turn/3` ends turn 4 and its own `where: c.status == "running"` update_all then idles C — but the window is the reaper's interval, not milliseconds, where before this patch it was zero. The precondition (a `running` row a stale actor never ended) is precisely the state this stack newly creates, so it is not a hypothetical. `t.turn_number > ^turn.turn_number` is an exact fence for "newer" here (`_unsafe_next_turn_number/1` makes it monotonic per conversation) and matches the PR title. Your call whether to tighten it or just say so in the docstring; either is fine by me.

**2. Non-blocking (observability): a stale interrupt's span claims success.** `turn_machine.ex:693` — `end_span(turn.span, :error, %{"outcome" => "interrupted"})` runs before the `if turn.interrupted?`, so a mark that was refused still closes its trace as though it had interrupted the turn. `finish/4` distinguishes this case (`%{"outcome" => "completion_ignored"}` at `turn_machine.ex:639`). One expression — `if(turn.interrupted?, do: "interrupted", else: "interruption_ignored")` — makes the two ends of the machine read the same in Honeycomb.

**3. Non-blocking (test hygiene in a shared database): the unboxed test leaves rows and can mask its own failure.** `interruption_admission_isolation_test.exs` — the `after` block cleans audit events, the user and the sandbox (correctly ordered: `sandboxes.user_id` is `nilify_all` so the sandbox survives the user, and `conversations.user_id` is `delete_all` so the conversation and turns cascade). But `_unsafe_create_turn_on_sandbox/4` calls `record_turn_usage/1` after its transaction commits, and `usage_events.user_id` is `nilify_all` (`20260801160000_nilify_usage_events_user_on_delete.exs`), so each run leaves a `turn_started` usage row with a nil tenant in `fountain_test` — which several workspaces share on this machine. Separately, on an *already failing* run, `Task.shutdown(admission, :brutal_kill)` kills a task holding the conversation's `FOR UPDATE` plus the per-sandbox advisory lock, and the `Repo.delete!(user)` two lines later has to take that same conversation row to cascade; if the dying connection has not been reaped yet that delete blocks and the test reports a checkout/timeout rather than the assertion that actually failed. Neither costs you a green build today; both cost someone an hour later.

**Nit:** the `@deliberately_silent` entry (`audit_guardrail_test.exs:172`) is added under a comment that says the map "Mirrors the `Fountain.Audit` moduledoc", and the moduledoc was not touched. Its existing "per-turn state-machine writes — turn started, output chunk, turn ended" bullet arguably already covers this, and #1999 did the same for `_unsafe_complete_turn/3`, so I'd leave it — flagging only so the drift is on the record.

**Stack-wide, not yours to fix:** `handle_cast({:machine_gone, ...})` at `conversation_server.ex:1689` still does an unguarded `Conversations.update_conversation(conv, %{status: "idle"})` right after `interrupt_turn/1`, which re-opens exactly the clobber this PR closes. #2006 is titled for that path, so I am treating it as supplied later rather than a finding here.

## Checked and clean

- **Every caller.** `mark_interrupted/1` and `close_interrupted/1` have exactly one production caller each, chained inside `interrupt_turn/1`, so the new `interrupted?` flag is never observed across a `from_state/into_state` round trip (and `into_state/2` correctly does not carry it — it is reset to `false` on the way out anyway). All three `interrupt_turn/1` call sites either pattern-match `current_turn: nil` or guard on `state.current_turn`, so the `%Turn{}` clause on `_unsafe_interrupt_turn/2` cannot be reached with nil.
- **No dropped side effect in the swap from `update_conversation/2` to a raw write.** `update_conversation/2` is changeset + `Repo.update` + `broadcast_sidebar_update/1` and nothing else; phase two reproduces all three with the broadcast after the commit, and it is strictly more conservative than the old code, which idled even a `terminated` conversation. Activation moved from inside `_unsafe_update_turn/2` to after the transaction commits, which is the right side of the never-audit/never-write-inside-a-transaction rule.
- **Guard polarity.** `conv.sandbox_id == sandbox_id` (not the reverse), `status not in ["terminated", "failed"]`, `%Turn{status: "running"}` in phase one and `"interrupted"` in phase two, `false <- Repo.exists?(...)`. A deleted conversation falls out on `Repo.one/1` returning nil. All five `else` paths collapse to `:noop`.
- **Lock ordering.** Phase one, phase two, `_unsafe_complete_turn/3` and `_unsafe_orphan_turn/3` all take the conversation before the turn; admission takes the sandbox advisory lock then the conversation, and neither interrupt phase takes the advisory lock. No cycle.
- **Metric accounting.** `emit_completed/2` is pure telemetry with no billing path, so skipping it on a stale mark loses no credit. A successful mark followed by a refused idle still emits exactly once, and the between-halves tests assert that plus a `refute_receive` for the second.
- **Tests bite.** The four stale-mark cases compare full reloaded structs against pre-call snapshots and assert no `turn` stage and no metric; the four between-halves cases (reassigned / terminated / new_turn / deleted) assert the conversation row is byte-identical afterwards — all of which fail against the old unconditional `_unsafe_update_turn` + `update_conversation(idle)`. The isolation test waits on the signal it asserts (a real Postgres lock wait via `pg_blocking_pids`, then the admitted turn and the conversation both still `running`, and `:noop` from the ending half), not on a sleep. The telemetry pause handler is keyed on a process-dictionary flag set only in the admission task and on the `INSERT INTO "turns"` query text, so it is inert for every other process, and it is detached in `after`.
- **`conversation_server.ex` is not touched at all**, so the hard line-count pin is not in play. No `Task.async` for fire-and-forget, no `async: false` beyond the one unboxed test that genuinely needs it, pool_size untouched, no docs/nav or ADR obligation, and no docstring in the patch describing behavior a later PR supplies.

jhgaylor:
**Reworked. The approval above predates a fix for a bug this PR would have shipped, and that bug is unrecoverable.**

## 1. It carried the predicate #1999's review rejected

This PR called `newer_running_turn_query` in three places — the definition and two call sites. That is the *"no other turn is `running`"* predicate that #1999's third review round identified as the blocker; `main`'s predicate is *"this turn has the highest `turn_number`"* (`ExecutionGuard.latest_turn?/2`), and #1999's fix deleted the wrong one. So this PR would either have failed to compile or reintroduced the bug four review rounds removed.

Both call sites now use `conv.status == "running" and ExecutionGuard.latest_turn?(conv.id, turn.id)`, matching what #1999 landed. `grep newer_running_turn_query` returns nothing.

## 2. The split interrupt created an unrecoverable stuck-`running` conversation

Ten pre-states driven through `mark_interrupted/1` + `close_interrupted/1`, measured on `main`, on this PR as it stood, and after the fix:

| pre-state | main | this PR | + fix | reaper sees it? |
|---|---|---|---|---|
| control | idle | idle | idle | — |
| turn already `completed` / `failed` / `interrupted` | idle | **running** | running | **no** |
| parent rebound *before* | idle | running | running | yes → rescued |
| **parent rebound *between* halves** | idle | **running** | **idle** | **no** |
| newer turn between halves | running | running | running | yes |
| older turn stuck `running` | idle | idle | idle | — |
| parent `pending` | pending | pending | pending | — |

**The producer is live and needs no actor change.** `follow_cotenants/2` (`conversations.ex:5748`) moves every co-tenant of a replaced sprite with one unlocked `Repo.update_all(set: [sandbox_id: ...])` (ADR 0023 gate 5), so a conversation's binding changes under its own live server — between the two halves, while the peer is stopping.

**And nothing recovers it.** `AutonomousTurnReaper.sweep_stuck_turns/0` selects `where t.status == "running"`, and `_unsafe_recover_turn/3` no-ops unless the turn is `running`. The interrupt's *first* half already retired the turn, so no sweep will ever look at that conversation again. The parent then feeds `_unsafe_sandbox_busy_elsewhere?/4` and a shared machine bills to `SANDBOX_MAX_LIFETIME_HOURS`.

The general point, which is why this is worth stating at length: **every other #1767 fence is safe precisely because refusing leaves the turn `running`**, and therefore visible to the reaper. The probe shows it directly — `parent_rebound_before` is orphaned and rescued; `parent_rebound_between` is not orphaned and stays stuck.

**Fix (`159ceb19`, a separate commit so it can be dropped independently):** `_unsafe_idle_interrupted_turn/1` drops the binding condition and keeps the two conditions provable under the parent lock — parent `running`, and `latest_turn?/2`. The reasoning: a binding fence belongs on a writer that persists a *result*; this one writes no turn row, and its single write releases a status this same actor set moments ago.

The four rows still `running` after the fix are `end_running_turn`'s `%Turn{status: "running"}` arm refusing because another writer moved the row. Every writer that could produce that while a server holds `current_turn` was enumerated and none can, so they are correct as shipped — recorded on the docstring as known gaps rather than dismissed. A crash between the halves reaches the same state, but does so on `main` too, so it is pre-existing.

## 3. Activation was correct by accident, and now is not

This PR's writer does call `Activation.turn_replied/1`. But the tail is two adjacent conditionals, and it passes `idle?: false`:

```elixir
if reply_materialized?, do: Fountain.Activation.turn_replied(updated)
if idle?, do: broadcast_sidebar_update(conv.user_id)
```

Merging them into `if idle? and reply_materialized?` — the obvious tidy-up — **survives 1,456 tests with zero failures** while permanently unactivating every interrupted first reply. Added *"an interrupted first reply activates the account"*; the mutation now dies on it, as does dropping the call entirely. This is the decision #2055's proposed guardrail would force.

Re-review requested rather than assumed — particularly on the fence-removal argument in §2, which is the load-bearing judgement here.

BinaryBourbon (changes requested):
## Re-review of the rework (`159ceb19`, two commits on `859f60c7`). Verdict: **request changes** — on the recorded justification, not on the code.

The code is right. I could not break it. But the docstring that exists to stop a future reader re-adding the fence contains two statements about this codebase that are false, and I checked both by running them rather than reading them. That docstring is the only record of why a campaign-wide invariant is deliberately absent here, so a reader who verifies it and finds it wrong has every reason to conclude the decision was unfounded — and re-add the fence.

Own worktree, own database, three trees built from source, several seeds.

---

## 1. The probe, from my own run

Ten pre-states through `mark_interrupted/1` + `close_interrupted/1`, on `main` (`859f60c7`), on the approved head (`6d47cbbf`) and on the tip (`159ceb19`). Cells are **conversation status / turn status**. "reaper" is whether the conversation still has a `running` turn, i.e. whether `AutonomousTurnReaper.sweep_stuck_turns/0` can ever see it again.

| pre-state | `main` | approved `6d47cbbf` | tip `159ceb19` | reaper |
|---|---|---|---|---|
| control | idle / interrupted | idle / interrupted | idle / interrupted | — |
| turn `completed` before | **idle / interrupted** | running / completed | running / completed | no |
| turn `failed` before | **idle / interrupted** | running / failed | running / failed | no |
| turn `interrupted` before | idle / interrupted | running / interrupted | running / interrupted | no |
| parent rebound *before* | **idle / interrupted** | running / running | running / running | yes |
| **parent rebound *between*** | idle / interrupted | **running** / interrupted | **idle** / interrupted | **no** |
| newer turn between halves | running / interrupted | running / interrupted | running / interrupted | yes |
| older turn stuck `running` | idle / interrupted | **running** / interrupted | idle / interrupted | yes |
| parent `pending` | pending / interrupted | **idle** / interrupted | pending / interrupted | no |
| terminated between halves | terminated | terminated | terminated | — |

Three rows move between the approved head and the tip, all in the right direction, and none regresses:

- **`rebound_between`** — the claimed bug. Reproduced. The approved head leaves the conversation `running` with no `running` turn, so the reaper is blind to it. Fixed at the tip.
- **`older_turn_running`** — the approved head's predicate really was the one #1999's third round rejected. It is not called `newer_running_turn_query` on this branch and never was (`git grep` at `6d47cbbf` finds nothing), so that part of the account is wrong on the name; the substance is right. It was an inline `running_query` — `Repo.exists?(any running turn on this conversation)` — and it regressed against `main`, which idles. The tip matches `main`.
- **`parent_pending`** — new, unmentioned: the approved head idled a `pending` conversation. The tip's `%Conversation{status: "running"}` head fixes it, and there is a test.

Bold on `main` is the harm #1767 exists to prevent, still visible for contrast: `main` overwrites a `completed` turn with `interrupted`, and lets a stale actor write after reassignment.

---

## 2. The fence removal. I attacked it; it holds — for a reason the docstring does not give

**A stale actor cannot reach the write, and the SQL guards are not what stops it.** I built the fixture: parent `running`, turn `interrupted`, our turn latest — every one of phase 2's three conditions satisfied — but phase 1 refused on the binding. Result: `running`. Nothing written.

The reason is structural, and stronger than "moments ago":

- `close_interrupted/1` writes only under `if turn.interrupted?`, set only by a `mark_interrupted/1` whose `_unsafe_interrupt_turn/2` passed the binding fence;
- `interrupted?` is in neither `from_state/1` nor `into_state/2`, so it cannot survive a mailbox round-trip — `from_state/1` always rebuilds it as the struct default `false`;
- both halves are one synchronous body in `interrupt_turn/1`, separated only by `stop_acp_peer/1`, which is `GenServer.stop(peer, :normal, 1_000)` — a **bounded 1 s window**;
- `state.sandbox_id` is written once at `init` (`conversation_server.ex:478`) and never again, so a rebind always makes a live server stale for phase 1.

So the fence did not go away. It moved out of SQL into the process, where it is tighter. That is the argument I would want in the docstring, because it is checkable and it is what actually holds. Worth stating the corollary too: **`_unsafe_idle_interrupted_turn/1` is safe only from a `close_interrupted/1` whose mark succeeded**, and nothing in the code enforces that — it is a public `_unsafe_` context function with one caller.

**Is `parent-running` + `latest_turn?` sufficient under the lock?** Yes. `_unsafe_create_turn_on_sandbox/4` takes the parent `FOR UPDATE` (`conversations.ex:1804`) and inserts the turn *and* sets the parent `running` (`:1873`) in that same transaction, so there is no window where a successor exists but `latest_turn?/2` still answers ours. `Connection`'s and `Reattachment`'s `running` writes both accompany a turn row. And `(conversation_id, turn_number)` is uniquely indexed, so `latest_turn?/2`'s `order_by … limit 1` cannot tie. The one hole — an *older* turn genuinely `running` under a reattached server — is `latest_turn?/2`'s documented, deliberate semantics, identical on `main` and on `_unsafe_complete_turn/3`. Not this PR's.

---

## 3. What I want changed: two false statements in the docstring

**(a) `_unsafe_sandbox_busy_elsewhere?/4` does not read the parent.**

> …and `_unsafe_sandbox_busy_elsewhere?/4` reads that parent, so a shared machine bills to `SANDBOX_MAX_LIFETIME_HOURS`.

It reads co-tenant *turn* rows (`t.status == "running" or t.inserted_at > cutoff or t.ended_at > cutoff`), plus `c.updated_at` for co-tenants that have no turn at all. It never reads `conversations.status`. Measured on exactly this state — a co-tenant `running` with one two-hour-old `interrupted` turn:

```
parent_running_terminal_turn -> false
control_running_turn         -> true
```

The stuck parent does not make the machine look busy, and the billing consequence does not follow.

**(b) The named producer rescues itself.**

`follow_cotenants/2` casts `{:machine_gone, …}` to every co-tenant **before** its unlocked `update_all` (`conversations.ex:5772` then `:5776`). That cast lands in `MachineEvents.gone/4`, which is:

```elixir
state = if state.current_turn, do: interrupt.(state), else: state
state = drop.(state, event)
conv = Conversations._unsafe_get_conversation!(state.conversation_id)
if conv.status == "running", do: Conversations.update_conversation(conv, %{status: "idle"})
```

The interrupt it runs *is* the two-phase pair, and the tail idles a `running` parent unconditionally, with no binding check. Driven through `gone/4` on both trees:

| | after phase 1 + rebind | after phase 2 | after `gone`'s tail |
|---|---|---|---|
| approved `6d47cbbf` | running | **running** | **idle** |
| tip `159ceb19` | running | idle | idle |

So on the producer the PR names, the stuck state lasts microseconds. `_unsafe_recover_turn/3` and the reaper are blind to it, as claimed, but they are not the only backstops: `Lifecycle.finish_park/4` (`:445`), the reclaim path (`:536`) and `gone/4` (`:67`) all clear a `running` parent, and the server's own idle tick reaches the first two without reading the conversation row at all. `SandboxReaper` is the one that genuinely never touches it ("The conversation is deliberately left alone", `sandbox_reaper.ex:339`).

The harm is still reachable — `tell_cotenants/3` skips a co-tenant whose `ConversationServer.whereis/1` misses on this node, which under Horde is exactly the cross-pod case ADR 0023 gate 5 is about; that server gets the rebind and no cast. But that is a narrower and different claim from the one written, and "**The producer is live and needs no actor change**" oversells it.

What I would like instead: keep the conclusion, replace the two supporting facts with the ones above and with §2's structural argument, which needs no appeal to how bad the alternative is.

---

## 4. Mutation coverage

Every guard in the new code dies on at least one test. Narrow set first (`turn_machine`, `interruption_admission_isolation`, `turn_completion_binding_isolation`, `execution_guard`, `activation` — 128 tests), widened to 272 for survivors. A no-op control mutation survived both, so the harness is not just failing to compile.

| mutation | |
|---|---|
| drop the binding fence in `end_running_turn/4` | **killed** (3) |
| drop the terminal-parent guard | **killed** (3) |
| accept a non-`running` turn | **killed** (4) |
| drop `idle? and` (phase 1 idles) | **killed** (3) |
| drop `latest_turn?` from the phase-1 idle | **killed** (1) |
| phase 2: accept any parent status | **killed** (2) |
| phase 2: accept any turn status | **killed** (1) |
| phase 2: drop `latest_turn?` | **killed** (3) |
| `close_interrupted` idles regardless of `interrupted?` | **killed** (4) |
| `mark_interrupted` always claims applied | **killed** (4) |
| **`if idle? and reply_materialized?`** | **killed** (1) |
| *reverse:* re-add the binding fence to phase 2 | **killed** (1) |

The two that matter most:

- The activation mutation is real. With `"an interrupted first reply activates the account"` deleted, `if idle? and reply_materialized?` **survives 145 tests**, including `activation_test.exs`, `turn_parent_actor_test.exs` and `conversation_server_test.exs`. With the test present it dies. Confirmed in both directions.
- The fence removal is itself guarded: putting the binding condition back (and passing `turn.sandbox_id` from `close_interrupted/1`) fails `"a rebind between the interrupt halves still releases the parent"`. Two other failures in that run were arity errors from my own patch, not evidence.

**Two survivors**, neither blocking:

- `conv.status == "running"` on the *completion* idle survives 272 tests. That guard is #1999's, unchanged here — a pre-existing gap, and the mirror of the `parent: "pending"` case this PR did add for the interrupt.
- `if idle?, do: broadcast_sidebar_update(conv.user_id)` survives. This PR introduced that conditional; nothing asserts the sidebar broadcast is suppressed in phase 1 and emitted in phase 2.

---

## 5. Smaller things

- `newer_running_turn_query` is gone (it never existed on this branch). Both former call sites read `conv.status == "running" and ExecutionGuard.latest_turn?(conv.id, turn.id)`, matching `859f60c7`. ✅
- Every function the new docstring names resolves: `_unsafe_fence_sandbox_for_teardown/2`, `_unsafe_recover_turn/3`, `follow_cotenants/2`, `sweep_stuck_turns/0`, `_unsafe_sandbox_busy_elsewhere?/4`. ✅
- Size pin: `conversation_server.ex` is **2518** lines at all three of `859f60c7`, `71db849a` and `159ceb19` — this PR does not touch that file — against `@pin 2549`. (The 2524/2525 figures in the account above are not what the file measures.) ✅
- `turn_machine_test.exs:610` — the `for change <- [:terminated, :new_turn, :deleted]` loop still carries a `:reassigned ->` branch that nothing generates. Dead since the tag list lost `:reassigned` in `159ceb19`. Delete it.
- The new `@deliberately_silent` entry is inert: removing it changes nothing, and so does removing the `_unsafe_complete_turn/3` entry beside it. That map is checked only for "every entry has a reason" — the guardrail's teeth are the explicit `@must_audit` list, which does not discover new context mutations. The entry is correct as documentation; just don't count it as coverage. (Repo-level, not this PR's.)
- `mix format --check-formatted` clean.

---

## 6. Suite

Own database, pristine, `apps/fountain`:

- `159ceb19`, seed `31337` — 5,824 tests, **1 failure**
- `159ceb19`, seed `424242` — 5,824 tests, **0 failures**
- `859f60c7`, seed `31337` — 5,810 tests, **0 failures**

The one failure is `FountainWeb.TeamStreamTest`, `"events from every teammate arrive on one connection, labelled"` — unrelated to this PR (it touches neither the interrupt path nor the sidebar broadcast), and it passes 3/3 in isolation at the same seed on both trees. It waits on `Process.sleep(300)` for the SSE subscription and then asserts on the events, which is the usual shape. Filed rather than dismissed.

A first run of both suites showed 41 failures on the tip and 0 on the base. That was my own contamination: `interruption_admission_isolation_test.exs` uses `Ecto.Adapters.SQL.Sandbox.unboxed_run`, and my mutation runs left one user and one conversation behind, which is enough to break every test that counts rows globally (`FunnelTest`, `AccountsAdminListingTest`, admin pagination, `OpsGauges`). Recreating the database took it to the numbers above. Worth knowing before anyone runs a mutation pass over that file.

---

## Summary

Change the two false statements in `_unsafe_idle_interrupted_turn/1`'s docstring and delete the dead `:reassigned` branch, and I will approve immediately. No code change is needed — I could not find a state where the removed fence buys anything, and the in-process gate that replaces it is tighter than the one it replaces.

BinaryBourbon (approved):
Approved at `078e1fdd5801578791f2c3f606841a8984655e96`.

The previous blocking review findings are resolved. The cleanup docstring now accurately describes the successful-mark caller requirement, the synchronous interrupt gate, the recovery paths that can still idle the parent, and the co-tenant notification/rebind behavior. It no longer attributes sandbox billing lifetime to the parent's status. The unreachable `:reassigned` test branch is removed.

I reviewed the full PR diff and traced both interrupt phases through their production caller, turn admission, generation checks, recovery, and reply activation. The first phase preserves the binding and terminal guards; the second preserves newer admitted work under the parent lock while allowing cleanup after a rebind between phases. I found no blocking issue.

Validation: independently ran 227 focused tests on an isolated snapshot and a separate PostgreSQL database, with zero failures. These cover the turn machine, interruption/admission and completion/admission isolation, execution guards, activation, platform inference, and audit guardrails. GitHub's `CI required` and coverage gate are also successful for the reviewed head.

PR #2001, opened 2026-09-11
fix: preserve newer turn results when a transport fails

jhgaylor:
Transport failure handling could overwrite a turn or idle its conversation after reassignment, termination or another actor's completion. It now uses `TurnMachine.finish/4` and its guarded transaction instead of separate unconditional updates.

The existing local peer-stop attempt runs before committing the turn result, so completion also rechecks changes made during that callback. Successful failures retain their stage reason, span cleanup, completion metric and cleared connection state. Stale failures clear local state without changing the persisted result or emitting another turn-completion event. The peer stop retains its existing best-effort behavior.

Stack: based on #2000. Refs #1767. Earlier pending-permission/usage writes, startup-failure bookkeeping and durable forced-teardown recovery remain separate follow-ups.

Validation:
- All 288 focused tests passed, including real ACP, turn metrics, platform inference, termination actors and audit guardrails.
- Five new tests use a real local GenServer peer. They verify ordering and change the binding/result/terminal state or delete the conversation during shutdown.
- All five tests fail against the parent failure handler and pass with this change.
- Full `mix precommit` passed: Fountain 5,465 tests and 6 doctests, Buzz 144 tests, Support 33 tests; zero failures.

Integration validation: this stack merged cleanly in a temporary worktree with #1948, #1974, #1975, #1977 and #1979, including main at `4fc85151`. All 182 targeted lifecycle tests passed. Full `mix precommit` on the combined tree passed 5,510 Fountain tests and 6 doctests, 144 Buzz tests and 33 Support tests with zero failures. The draft PRs remain separate; the integration branch was not published.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/reattachment.ex
+++ b/apps/fountain/lib/fountain/conversations/reattachment.ex
@@ -127,42 +127,25 @@ defmodule Fountain.Conversations.Reattachment do
     Logger.error("sprite command error mid-turn: #{inspect(reason)} — failing the turn")
     state = finish_runner_reconnect(state, "failed")
 
-    # ownership: current_turn belongs to the conversation this server owns.
-    {:ok, turn} =
-      Conversations._unsafe_update_turn(state.current_turn, %{
-        status: "failed",
-        ended_at: DateTime.utc_now() |> DateTime.truncate(:second)
-      })
-
-    Output.publish_stage(state.conversation_id, "turn", "failed", %{
-      turn_id: turn.id,
-      turn_number: turn.turn_number,
-      reason: "sprite connection lost: #{inspect(reason)}"
-    })
-
-    TurnMachine.finalize_tracer(state.stream_tracer)
-
-    # An ACP turn can also end here — the adapter exits, is interrupted, or its
-    # socket drops before it ever answers `session/prompt`. The peer has nothing
-    # left to drive and must not outlive the turn.
+    # Stop the failed connection's local peer before committing the turn result.
+    # Completion rechecks the actor's binding after this callback can yield.
     Connection.stop_peer(Connection.from_state(state))
-    TurnMachine.end_span(state.current_turn_span, :error, %{"error" => inspect(reason)})
 
-    TurnMachine.emit_completed(TurnMachine.from_state(state), turn.status)
+    turn =
+      TurnMachine.finish(
+        TurnMachine.from_state(state),
+        "failed",
+        %{"error" => inspect(reason)},
+        %{reason: "sprite connection lost: #{inspect(reason)}"}
+      )
 
-    # ownership: state.conversation_id is the server's bound conversation.
-    conv = Conversations._unsafe_get_conversation!(state.conversation_id)
-    {:ok, _} = Conversations.update_conversation(conv, %{status: "idle"})
+    state = TurnMachine.into_state(state, turn)
 
     {:noreply,
      %{
        %{state | last_activity_at: DateTime.utc_now()}
        | current_command: nil,
          current_command_ref: nil,
-         current_turn: nil,
-         current_turn_span: nil,
-         turn_metrics: nil,
-         stream_tracer: nil,
          runner_reconnect: nil,
          runner_replay: nil,
          acp_peer: nil,
--- a/apps/fountain/test/fountain/conversations/reattachment_failure_test.exs
+++ b/apps/fountain/test/fountain/conversations/reattachment_failure_test.exs
@@ -1,0 +1,154 @@
+defmodule Fountain.Conversations.ReattachmentFailureTest do
+  use Fountain.DataCase, async: false
+
+  alias Fountain.Conversations
+  alias Fountain.Conversations.{Reattachment, TurnMachine}
+
+  defmodule Peer do
+    use GenServer
+    def start_link(on_stop), do: GenServer.start_link(__MODULE__, on_stop)
+    def init(on_stop), do: {:ok, on_stop}
+    def terminate(:normal, on_stop), do: on_stop.()
+    def terminate(_reason, _on_stop), do: :ok
+  end
+
+  setup do
+    user = insert_verified_user()
+    conv = insert_conversation(user_id: user.id, status: "running")
+    turn = insert_turn(conv, status: "running")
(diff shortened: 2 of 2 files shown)


BinaryBourbon (approved):
APPROVE — the guard covers the window it claims, the happy path still writes and still publishes, and the ordering change is the load-bearing part of the patch rather than incidental.

## What this changes

`Reattachment.fail_transport/2` stops doing its own four unconditional writes (turn row, `turn`/`failed` stage, span, `conv → idle`) and hands the ending to `TurnMachine.finish/4`, whose `_unsafe_complete_turn/3` takes the conversation and turn locks, rechecks `conv.sandbox_id == state.sandbox_id` and the terminal-status/`running`-turn conditions, and commits the turn result and the idle status in one transaction. The second half of the patch is an ordering move: `Connection.stop_peer/1` now runs *before* the commit rather than between the turn write and the span close, so the up-to-1s `GenServer.stop(peer, :normal, 1_000)` window is inside the fence rather than straddling it. The five new tests drive the real function with a real local GenServer peer whose `terminate/2` mutates the database, which is the only way to exercise that window.

## Findings

None blocking.

**1. Non-blocking — the stale path drops the transport reason from the span.** `turn_machine.ex:639`: the `:noop` branch closes the span with `%{"outcome" => "completion_ignored"}`, so `%{"error" => inspect(reason)}` — the only structured record of *why* the transport died — is not stamped anywhere on a stale failure. The `Logger.error` at `reattachment.ex:127` still carries it, so nothing is invisible, but a trace of a stale transport failure now says "ignored" and not "closed". Merging the two attribute maps in the `:noop` branch would cost nothing. This is `finish/4`'s behavior from an earlier PR in the chain and applies to every caller, so it belongs upstream of this patch if you take it.

**2. Non-blocking, test nit — `refute Repo.in_transaction?()` (`reattachment_failure_test.exs:39`) is evaluated in the peer process, not the server's.** If it is meant to assert "the peer stop is not happening inside the completion transaction", it cannot see the calling process's transaction depth through a `Sandbox.allow`-shared connection. The assertion that actually proves the ordering is the `"running"` status it sends on the next line — which is exactly the right discriminator, since the parent handler writes `failed` before `stop_peer` and would send `"failed"` instead. Worth keeping the status assertion and either dropping the `refute` or moving the check to where the server process runs.

**3. Non-blocking — `async: false`.** CLAUDE.md asks for `async: false` only where the test genuinely needs it. The handler id is `{__MODULE__, make_ref()}` and the handler filters on `meta.conv_id`, so the global `:telemetry.attach` looks safe under `async: true`; the `refute_receive {[:fountain, :turn, :completed], _}, 50` is the only thing I can see arguing for serial, and the conv-id filter already covers it. If there is a reason, a one-line comment saves the next reader the same walk.

**4. Soft, chain-level — no `[Unreleased]` CHANGELOG entry anywhere in this chain yet** (`git diff origin/main...origin/codex/p2-1767-transport-completion -- CHANGELOG.md` is empty). A conversation that no longer idles itself out from under a newer sandbox is user-visible. Presumably one of #2002/#2003/#2006/#2007 carries it for the whole of #1767; flagging so it does not fall through the gap.

## Checked and clean

- **The guard cannot swallow the ordinary failure.** `state.sandbox_id` is set once from `init` args (`conversation_server.ex:463`) and never reassigned, so it is a genuine immutable actor binding; with nothing concurrent, `conv.sandbox_id == state.sandbox_id` holds and the turn is written `failed`, the conversation `idle`, the stage published and `turn.completed` emitted. The first test asserts all four.
- **The failure still surfaces on every path.** `Logger.error("sprite command error mid-turn: …")` is unconditional and precedes everything, and `finish_runner_reconnect/2` still publishes its `connection`/`done` stage before the fence, so a dropped write is never a silent drop.
- **A turn left `running` by the `:noop` branch is not left forever.** Wake-time reconciliation (`conversation_server.ex:1444` → `_unsafe_orphan_turn/3` with *no* `expected_sandbox_id`) and `AutonomousTurnReaper.sweep_stuck_turns/0` (`autonomous_turn_reaper.ex:69`, likewise unbound) both reconcile it. Note the reaper only fires when `ConversationServer.whereis/1` is nil, so the window is "until the old server stops, then the stuck bound" rather than immediate — acceptable, and better than the clobber it replaces.
- **Stage and webhook parity.** `TurnMachine.publish_stage/4` → `Conversations.publish_stage/4`, which is exactly what `Output.publish_stage/4` delegated to, so the log event, the `[:stage]` telemetry, the PubSub broadcast and the `conversation.turn.failed` webhook dispatch are unchanged. `finish/4`'s conditional third argument is invisible to `webhooks/events_test.exs`'s regex, but that file already documents the exception and two literal `"turn", "failed"` sites remain in `turn_machine.ex`, so the catalogue guard is not weakened into vacuity.
- **The idle write loses nothing.** `end_running_turn/4` does `Conversation.changeset` + `Repo.update!` + `broadcast_sidebar_update/1`, which is `update_conversation/2` (`conversations.ex:1168`) with the broadcast moved after the commit. No audit call is involved on either side, so nothing moved into a transaction.
- **The deleted-conversation case is strictly better.** The old code called `_unsafe_get_conversation!/1`, which would raise and take the server down if the row went away mid-flight; the locked `Repo.one` now returns nil and the whole ending becomes a no-op. The `:deleted` test covers it.
- **Moving `stop_peer` earlier is safe.** It demonitors with `[:flush]` and catches the exit, so no DOWN or EXIT survives it; the peer's own messages queue behind this callback either way, so the `{:lines, …}`-with-nil-row / phantom-autonomous-turn hazard is unchanged from the parent. `finalize_tracer/1` now runs after the stop rather than before, matching what `disconnect_runner/1` two functions down already did.
- **Tests fail against the parent.** The happy-path test's `"running"` reload inside `terminate/2` inverts under the old write-then-stop order, and each of the four stale cases is a write the parent would have overwritten (`completed` → `failed`, `terminated` → `idle`, a moved binding idled, a deleted row raising). No test stubs the thing it asserts, and each waits on the peer's own exit — the signal it then asserts on.
- **No aliases orphaned** (`Conversations` and `Output` both still used in the file), and `conversation_server.ex` is untouched, so the line-count pin is not in play.

PR #1948, opened 2026-09-11
fix(workers): reconcile pending sandbox resets with durable retries

jhgaylor:
Pending resets recover automatically after provider failures or caller loss. A five-minute sweep discovers fenced machines and enqueues one durable job per sandbox. Oban retries failed deletes with backoff, keeps incomplete jobs unique, and allows a later sweep to recover an exhausted job. Missing provider credentials snooze the job; stale jobs skip unfenced or retired machines. Capacity stays reserved until deletion is confirmed.

Builds on the landed #1947 atomic single-winner finalization and sandbox-scoped completion delivery, including the landed [deadline-wake fix #2047](https://github.com/managoat/fountain/pull/2047). The overlap regression drives the actual sweep and job while the original deletion response is held, rebinds the holder, then releases the original: the loser skips with one completion audit and no stale notification.

Refs #1894. The [September 11 decision](https://github.com/managoat/fountain/issues/1894#issuecomment-5632016801) also requires an admin-only provider reprobe/unfence path; this stack ships the retry sweep, so the issue remains open for that remaining path.

Validation at `5690c5652f0212b79c332808dc45963fd90180f7` (tree `4b3a398185e1a61a2a2c216b7aa984f95d95dba3`): focused worker and real PostgreSQL finalizer/blocked-notification coverage passed 7 tests. Full final precommit passed compile, format, strict Credo, Dialyzer (zero errors/skips), Sobelow, production release assembly, core 6 doctests + 5,689 tests, Buzz 144 tests and Support 33 tests, all with zero failures (2 skipped, 7 excluded). Seed 772406, max_cases 4. Destink passed 112 pages / 33 rules. Reviewed prose reports: 3 existing style findings and 11 Vale sentence-length findings outside the edited sandbox page; no unrelated prose edits.

CI retarget note: the retained [pre-retarget SDK guard](https://github.com/managoat/fountain/actions/runs/34692072253) used the obsolete parent base and counted SDK code already on main. The actual five-file main-based diff contains no SDK change. [Fresh main-based CI](https://github.com/managoat/fountain/actions/runs/34692208143) and [secrets](https://github.com/managoat/fountain/actions/runs/34692208121) passed on unchanged head `5690c565`; no version or release-label workaround was used.

Issue #608, opened 2026-08-07
conversations: a turn failed before start drops the runtime's exit code — turns.exit_code stays NULL and the reason names the mechanism, not the cause


jhgaylor:
Follow-up to #603, which is correctly closed — the crash is fixed and the turn
now ends `failed` instead of orphaned. This is about what that failure *says*.

`fail_turn_before_start/4` reports `reason: :command_exited`. That names the
mechanism — the command process was gone — but not the cause. The runtime's
exit code and whatever it wrote before dying are **already delivered to the
ConversationServer** and then dropped, and `turns.exit_code` has a column for
the exit code that stays `NULL`.

## Why the messages have nowhere to land

`Sprites.Command` sends the owner its exit before it stops
(`command.ex`, the binary and text exit frames):

```elixir
send(owner, {:exit, %{ref: ref}, code})
…
{:stop, :normal, %{state | exit_code: code}}
```

So by the time `SpriteStdin.write/2` catches the `:normal` exit and returns
`{:error, :command_exited}`, `{:exit, %{ref: ref}, code}` is already in the
ConversationServer's mailbox. Same for any `{:stdout, …}` / `{:stderr, …}` the
runtime produced first.

Every one of those handlers is guarded on the ref being the *current* command
(`conversation_server.ex:1117`, `:1151`, `:1155`):

```elixir
def handle_info({:exit, %{ref: ref}, code}, %{current_command_ref: ref} = state) do
```

`current_command_ref` is only assigned on the `:ok` branch of `stdin_result`.
The `{:error, reason}` branch calls `fail_turn_before_start/4`, which returns
`state` untouched — so on exactly this path the ref is never set, no guard
matches, and all three fall through to the catch-all at `:1278`:

```elixir
def handle_info(_msg, state), do: {:noreply, state}
```

Silently. The exit code that would say *why* the runtime quit is discarded a
few milliseconds after the turn is marked failed without it.

## What this costs

The production triggers for this path are the ones #603's `@moduledoc` lists —
a bad flag, a missing binary, an immediate non-zero exit, an OOM kill. Those
are precisely the cases where the exit code is the whole diagnosis:

| what happened | what the operator sees today |
|---|---|
| `claude` exited 1 on an expired key | `prompt write failed: :command_exited` |
| runtime binary renamed by a sprite image bump | `prompt write failed: :command_exited` |
| OOM kill at startup | `prompt write failed: :command_exited` |
[...]

PR #2002, opened 2026-09-11
fix: guard startup failure by actor sandbox binding

jhgaylor:
A delayed spawn failure could overwrite an already-ended turn or idle a conversation after reassignment. Startup failure now supplies the actor's sandbox binding to the guarded completion transaction. The transaction writes the exit code, result and idle status together; stale failures preserve the database state and emit no failure stage. The local span still closes on either outcome.

Stack: based on #2001. Refs #1767. This unit covers startup-failure bookkeeping; command-exit handling and durable forced-teardown recovery remain separate follow-ups.

Validation:
- 169 focused tests and 59 server tests passed, including audit guardrails, reply activation and inference attribution.
- Regressions cover reassignment, terminal conversations, ended turns followed by a new turn, deletion, and reassignment inside the actual server spawn callback.
- Two added PostgreSQL tests verify both lock orderings with the failure exit code committed atomically.
- Replacing the handler with its parent implementation makes seven regressions fail (13 selected tests); the fixed handler is restored.
- Full `mix precommit` passed: Fountain 5,475 tests and 6 doctests, Buzz 144 tests, Support 33 tests; zero failures. Static analysis, Dialyzer, Sobelow and production release assembly passed.

Integration validation: this change and the command-exit follow-up merged cleanly into the unpublished verification branch containing #1948, #1974, #1975, #1977 and #1979. All 223 targeted lifecycle tests passed on that combined tree.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2245,7 +2245,8 @@ defmodule Fountain.Conversations do
   Lock the conversation before the turn, and commit its idle status with the
   turn's result. A moved or terminal conversation, or a turn already ended by
   another actor, is a no-op. Reply materialization shares that transaction;
-  activation and sidebar publication run after it commits.
+  activation and sidebar publication run after it commits. The optional
+  `:exit_code` is persisted atomically with the result.
 
   The parent only idles under the two conditions `_unsafe_idle_after_turn/1`
   idled under: `ExecutionGuard.latest_turn?/2` and a `running` parent. The
@@ -2258,9 +2259,9 @@ defmodule Fountain.Conversations do
   The two-phase interrupt the server drives itself is a different write, and
   it keeps the conversation running until the peer has stopped.
   """
-  def _unsafe_complete_turn(%Turn{} = turn, sandbox_id, status)
+  def _unsafe_complete_turn(%Turn{} = turn, sandbox_id, status, opts \\ [])
       when status in ["completed", "failed", "interrupted"] do
-    end_running_turn(turn, sandbox_id, status, true)
+    end_running_turn(turn, sandbox_id, status, true, Map.new(Keyword.take(opts, [:exit_code])))
   end
 
   @doc """
@@ -2271,7 +2272,7 @@ defmodule Fountain.Conversations do
   def _unsafe_interrupt_turn(%Turn{} = turn, sandbox_id),
     do: end_running_turn(turn, sandbox_id, "interrupted", false)
 
-  defp end_running_turn(turn, sandbox_id, status, idle?) do
+  defp end_running_turn(turn, sandbox_id, status, idle?, attrs \\ %{}) do
     {:ok, result} =
       Repo.transaction(fn ->
         conversation_query =
@@ -2288,10 +2289,12 @@ defmodule Fountain.Conversations do
              %Turn{status: "running"} = current <- Repo.one(turn_query) do
           changeset =
             current
-            |> Turn.changeset(%{
-              status: status,
-              ended_at: DateTime.utc_now() |> DateTime.truncate(:second)
-            })
+            |> Turn.changeset(
+              Map.merge(attrs, %{
+                status: status,
+                ended_at: DateTime.utc_now() |> DateTime.truncate(:second)
+              })
+            )
             |> maybe_put_reply_text(current)
 
           updated = Repo.update!(changeset)
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -2232,7 +2232,14 @@ defmodule Fountain.Conversations.ConversationServer do
         log_output(acc, stream, data)
       end)
 
-    TurnMachine.fail_before_start(turn, state.conversation_id, what, detail, exit_code)
+    TurnMachine.fail_before_start(
+      turn,
+      state.conversation_id,
+      state.sandbox_id,
+      what,
+      detail,
+      exit_code
+    )
 
     state = %{state | current_turn: nil, turn_session_retry: nil}
 
--- a/apps/fountain/lib/fountain/conversations/turn_machine.ex
+++ b/apps/fountain/lib/fountain/conversations/turn_machine.ex
@@ -1406,7 +1406,8 @@ defmodule Fountain.Conversations.TurnMachine do
   or the runtime exited before the prompt reached its stdin (#603). Both
   leave nothing running, so both end the same way — the turn `failed` with
   the reason, a `turn`/`failed` stage event, and the conversation back to
-  "idle".
+  "idle". A reassigned or terminal conversation, or an already-ended turn,
+  preserves its persisted result and emits no failure stage.
 
   `exit_code` is what the runtime managed to say before it went (#608); the
   spawn-failure path has none, since there was never a process. `detail` is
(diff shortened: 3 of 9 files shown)


BinaryBourbon (approved):
APPROVE — the guard covers the window it claims, the dropped write is reconciled by two existing backstops rather than wedging the row, and the patch is one idea. Three non-blocking notes, one of which asks for a docstring line.

**What this changes**

`TurnMachine.fail_before_start/6` stops writing the turn row and the conversation status with two unguarded updates and instead routes both through `Conversations._unsafe_complete_turn/4`, which now carries an optional `:exit_code` so the failure code lands inside the same locked transaction as the `failed` status and the conversation's `idle`. The server supplies `state.sandbox_id`, so a spawn failure arriving after the conversation was rebound to another sandbox is a `:noop`: no row write, no `turn`/`failed` stage. The span still closes on both outcomes.

**Findings**

1. **Reply materialization and activation now ride along on a turn that never started** (non-blocking, docstring fix asked) — `apps/fountain/lib/fountain/conversations/turn_machine.ex:1325`. `_unsafe_complete_turn/4` runs `maybe_put_reply_text/2` for any terminal status, so a startup failure whose drained parting words (#608) parsed to assistant text now writes `turns.reply_text` and, post-commit, calls `Fountain.Activation.turn_replied/1` — which stamps `users.onboarding_completed_at` and emits `activation.first_reply`. `turn_machine_test.exs:713` asserts exactly that, so it is intended, but neither the PR body ("the exit code, result and idle status together") nor the `fail_before_start/6` docstring says it. Failure scenario for a reader, not for the code: the next person auditing "why did this account activate on a conversation that never spawned" finds nothing in the docstring to explain it. Please add a line to the `@doc` (and ideally the body) saying the runtime's parting words are now materialized as the turn's reply and can activate the account. I checked the leak risk: `_unsafe_turn_reply_text/1` scopes events by `turn_id`, so a startup-failed turn cannot inherit the previous turn's text.

2. **The idle write is now unconditional for any non-terminal conversation** (non-blocking) — `apps/fountain/lib/fountain/conversations.ex:2148`. The old code was `if failed_conv.status == "running"`; the new guard is only `status not in ["terminated", "failed"]`, and the statuses are `pending running idle failed terminated`, so the one reachable difference is a conversation sitting at `pending` when the spawn fails, which now becomes `idle`. `run_fresh_turn/7` writes `running` a few lines before the spawn (`conversation_server.ex:2331`), so the window needs a concurrent write to leave the row `pending`, and `idle` is the more accurate answer there anyway — I am not asking for a change. Worth noting that the deleted comment carried the *why* ("without this it stays running in the API and UI until some later turn completes... the `:exit` and `:interrupt` handlers both do the same reset"); the replacement docstring states the behavior but drops the pointer.

3. **What a dropped startup failure leaves behind, and the one thing it can cost a co-tenant** (non-blocking, chain-wide, not a change to make here). On `:noop` the turn row stays `running` with `started_at` set and `ended_at` nil, and the stale server nils `current_turn`, so `terminate/2` → `orphan_on_normal_stop/4` sees `nil` and never touches it. It is not wedged: the next `ConversationServer` start for that conversation takes `reattach_running_turn/1` → `find_running_turn/1` → `mark_orphan/3` (no `expected_sandbox_id`, so it applies), and `AutonomousTurnReaper` is the durable backstop after 30 min stuck + 15 min quiet once no server holds the conversation — `_unsafe_orphan_turn/3` has no terminal-conversation guard, so it closes the row even on a terminated conversation. The cost while it stands: `_unsafe_running_turns_elsewhere/2` joins through the conversation's *current* `sandbox_id`, so the phantom counts against the **replacement** sandbox's runtime capacity for every co-tenant conversation on it (ADR 0023). With `gemini` or `opencode` (`concurrency: 1` in the adapter table) a co-tenant's next turn is refused `:sandbox_at_capacity` until one of the two backstops fires. This is the same property `finish/4`'s `:noop` already has from the earlier stack PR, so it belongs on #1767 rather than in this diff — flagging it so it is on the tracker and not discovered in prod.

**Checked and clean**

- The guard covers the whole window: `_unsafe_create_turn_on_sandbox/4` already verifies the conversation is attached to the actor's sandbox under the per-sandbox advisory lock, so the only way to reach `fail_before_start/6` with a stale binding is a reassignment between admission and the spawn result — which is exactly what `conversation_server_test.exs:952` reproduces by rebinding inside the `Sprites.spawn` stub. Revert the binding argument and that test fails on `turn.status == "running"`.
- Both `:noop` sub-cases beyond reassignment are sound: a terminal conversation and an already-ended turn are the two arms of the same `with`, and the `for stale <- [...]` matrix in `turn_machine_test.exs` covers reassigned / terminated / failed / completed / interrupted / deleted by asserting the persisted turn and conversation structs are byte-identical afterwards (so an `updated_at`-only write would fail too) plus no `turn` stage and no activation.
- `state.sandbox_id` cannot be nil here: `init/1` takes it with `Keyword.fetch!` and both start paths pass a concrete sandbox. And the binding comparison is in Elixir (`conv.sandbox_id == sandbox_id`), not in the query, so there is no `= NULL` / Ecto nil-comparison trap.
- `Map.merge(attrs, %{status: ..., ended_at: ...})` puts the caller's attrs first, so `:exit_code` cannot override the status or the timestamp, and `Keyword.take(opts, [:exit_code])` means no other field can be smuggled in. `:exit_code` is in `Turn.changeset/2`'s cast list (`turn.ex:108`), so the value really lands; `exit_code: nil` is a no-op change on a fresh row.
- Only one non-test caller each for `fail_before_start/6` and the new `_unsafe_complete_turn/4` arity, so no second path reaches the same write with the old signature. `_unsafe_interrupt_turn/2` still gets `attrs \\ %{}` and is unchanged in behavior.
- `_unsafe_complete_turn/4` stays inside the `_unsafe_` convention (the server owns the conversation) and the audit guardrail exclusion key is renamed to the new arity rather than dropped; the audit position (outside the transaction) is untouched.
- The server file is 2,757 lines against the `@pin` of 2,813, so the +7 here is inside the pin.
- `turn_completion_binding_isolation_test.exs` now runs the startup-failure outcome under both real-Postgres lock orderings and asserts `exit_code` is 17 on the applied side and nil on the dropped side — it waits on the same rows it asserts on, no signal substitution.
- No CHANGELOG entry, consistent with the rest of the chain; the only user-visible delta is a stage event that no longer fires for a stale actor.
- Scope: 8 files, but 5 are tests and 2 of those are a one-line arity/name update. This is one idea, not two.

jhgaylor:
Noted while re-verifying this PR against the corrected #1999 — **not blocking, and this PR is otherwise clean** (auto-merged, no conflict, no reference to the predicate #1999 deleted, and it makes no idle decision of its own — it deletes a hand-rolled one and routes through `_unsafe_complete_turn`).

The arity bumps in this PR leave **six stale references in prose**. Docs only, no behaviour:

`_unsafe_complete_turn/3` → `/4`:
- `apps/fountain/lib/fountain/conversations.ex:2351`
- `apps/fountain/lib/fountain/conversations.ex:2359`
- `apps/fountain/lib/fountain/conversations/turn_machine.ex:653`
- `apps/fountain/lib/fountain/conversations/turn_machine.ex:692`
- `apps/fountain/test/fountain/conversations/turn_parent_actor_test.exs:47`

`fail_before_start/5` → `/6`:
- `apps/fountain/lib/fountain/conversations/turn_launch.ex:247`

Worth fixing here rather than leaving, because two of them (`turn_machine.ex:653` and `:692`) sit in the docstring and comment that #1999 rewrote specifically to stop misleading the next reader about which fences are in the path. A stale arity in that particular comment undoes part of what that change was for.

PR #2003, opened 2026-09-11
fix: guard command exit by actor sandbox binding

jhgaylor:
An adapter's exit could overwrite a turn or idle its conversation after reassignment, termination or another actor's completion. The command-exit handler now commits through the existing guarded completion transaction, including its exit code.

The existing local peer-stop attempt runs before that transaction, so changes made during shutdown are rechecked. Applied exits retain their existing `turn/done` event, exit-code metadata and completion metric. Stale exits clear local state and close the span without changing the persisted result or publishing duplicate completion events. Peer shutdown retains its existing best-effort behavior.

Stack: based on #2002. Refs #1767. Machine-gone notifications and durable forced-teardown recovery remain separate follow-ups.

Validation:
- All 80 focused tests passed, including the real server lifecycle and turn metrics suites.
- Ten new tests use a real local GenServer peer and cover both zero/nonzero exit codes, with reassignment, another completion, termination or deletion during peer shutdown.
- All ten new tests fail against the parent handler and pass with this change.
- The first test run exposed a missing `autonomous_quiet` field in the new hand-built fixture. Adding the field required by `Connection.from_state/1` fixed that setup error; the failed log was retained.
- Both this change and #2002 merged cleanly into the unpublished verification branch containing #1948, #1974, #1975, #1977 and #1979. All 223 targeted lifecycle tests passed on the combined tree.
- Full `mix precommit` passed: Fountain 5,485 tests and 6 doctests, Buzz 144 tests, Support 33 tests; zero failures. Static analysis, Dialyzer, Sobelow and production release assembly passed.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -1715,37 +1715,36 @@ defmodule Fountain.Conversations.ConversationServer do
   defp handle_execution_info({:exit, %{ref: ref}, code}, %{current_command_ref: ref} = state) do
     turn = state.current_turn
 
-    {:ok, turn} =
-      Conversations._unsafe_update_turn(turn, %{
-        status: if(code == 0, do: "completed", else: "failed"),
-        exit_code: code,
-        ended_at: now()
-      })
-
-    Output.publish_stage(state.conversation_id, "turn", "done", %{
-      turn_id: turn.id,
-      turn_number: turn.turn_number,
-      exit_code: code
-    })
-
     # Finalize stream tracer: close any tool spans still open (abandoned calls).
     TurnMachine.finalize_tracer(state.stream_tracer)
 
-    # An ACP turn can also end here — the adapter exits, is interrupted, or its
-    # socket drops before it ever answers `session/prompt`. The peer has nothing
-    # left to drive and must not outlive the turn.
+    # An ACP adapter can exit before answering session/prompt. Attempt local
+    # peer cleanup before rechecking ownership: its callback may change the
+    # conversation while this actor waits for it to stop.
     stop_acp_peer(state)
 
-    # Close the OTel turn span we opened in kick_turn.
-    TurnMachine.end_span(
-      state.current_turn_span,
-      if(code == 0, do: :ok, else: :error),
-      %{"exit_code" => code}
-    )
+    status = if(code == 0, do: "completed", else: "failed")
 
-    emit_turn_completed(state, turn.status)
+    # Ownership: this actor supplies the sandbox binding captured at startup.
+    case Conversations._unsafe_complete_turn(turn, state.sandbox_id, status, exit_code: code) do
+      {:ok, ended} ->
+        Output.publish_stage(state.conversation_id, "turn", "done", %{
+          turn_id: ended.id,
+          turn_number: ended.turn_number,
+          exit_code: code
+        })
 
-    {:ok, _} = Conversations._unsafe_idle_after_turn(turn)
+        TurnMachine.end_span(
+          state.current_turn_span,
+          if(code == 0, do: :ok, else: :error),
+          %{"exit_code" => code}
+        )
+
+        emit_turn_completed(state, ended.status)
+
+      :noop ->
+        TurnMachine.end_span(state.current_turn_span, :error, %{"outcome" => "completion_ignored"})
+    end
 
     {:noreply,
      %{
--- a/apps/fountain/test/fountain/conversations/command_exit_binding_test.exs
+++ b/apps/fountain/test/fountain/conversations/command_exit_binding_test.exs
@@ -1,0 +1,176 @@
+defmodule Fountain.Conversations.CommandExitBindingTest do
+  use Fountain.DataCase, async: false
+
+  alias Fountain.Conversations
+  alias Fountain.Conversations.{ConversationServer, TurnMachine}
+
+  defmodule Peer do
+    use GenServer
+    def start_link(on_stop), do: GenServer.start_link(__MODULE__, on_stop)
+    def init(on_stop), do: {:ok, on_stop}
+    def terminate(:normal, on_stop), do: on_stop.()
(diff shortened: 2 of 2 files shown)


BinaryBourbon (approved):
**APPROVE** — the guard covers the window it claims, on the only path that reaches this write, and the tests have teeth. Four non-blocking notes.

## What this changes

`handle_info({:exit, %{ref: ref}, code}, …)` stops writing the turn result with a bare `_unsafe_update_turn` + an unconditional `conv -> "idle"`, and instead commits through `Conversations._unsafe_complete_turn/4`, handing it `state.sandbox_id` (set once at `init/1`, line 463, never reassigned) and the exit code. The transaction locks the conversation and the turn, re-reads both, and refuses when the conversation has moved to another sandbox, is `terminated`/`failed`, or the turn is no longer `running`. `stop_acp_peer/1` is deliberately moved *ahead* of that transaction so a peer `terminate` callback that changes the conversation is visible to the recheck — and so the commit is not held open across a `GenServer.stop(peer, :normal, 1_000)`. On `:noop` the actor closes its span with `outcome: "completion_ignored"`, publishes no `turn`/`done`, emits no completion metric, and clears local bookkeeping as before.

A side benefit worth naming: `end_running_turn/5` re-reads the turn `FOR UPDATE` rather than updating through `state.current_turn`, so this handler no longer writes a possibly-stale struct over fields another writer set mid-turn.

## Findings

1. **`refute Repo.in_transaction?()` can never fail** — `command_exit_binding_test.exs:55`. That line runs inside the `Peer.terminate/2` callback, i.e. in the peer process, which never has a connection in its own process dictionary, so `in_transaction?/0` is `false` there whether or not the owner is mid-transaction. The assertion that actually proves the ordering is the line under it: the peer shares the owner's sandbox connection via `Sandbox.allow/3`, so a completion already committed (or merely open on that connection) would read back as `completed`, and `assert_received {:peer_stopped, "running"}` catches it. Suggest dropping the refute or demoting it to a comment, so the next reader does not trust it as the guard. Non-blocking.

2. **A stale exit leaves the turn row `running` with `ended_at: nil`, and nothing here closes it.** `AutonomousTurnReaper.reap_if_unowned/1` bails while `ConversationServer.whereis(conversation_id)` returns a pid, so the reaper is not the backstop for a conversation that still has a live server. On the reassignment path the new actor's `orphan_stale_turn` (`conversation_server.ex:1444`) reconciles it at wake, which is fine. On the "terminated while the peer was stopping" path there is no such actor: the turn sits `running` on a `terminated` conversation until this server stops and the reaper's 30+15 minute window elapses. Non-blocking, and strictly better than the behaviour it replaces (which resurrected a terminated conversation to `idle`); flagging it because the PR body scopes durable forced-teardown recovery as a follow-up and this is one of the shapes that follow-up owes.

3. **The stage-event asymmetry with `TurnMachine.finish/4` deserves a comment.** This handler publishes `turn`/`done` for *both* zero and nonzero exits with `exit_code` in the data, while `finish/4` publishes `turn`/`failed` for a failed status. That is inherited from the base and is the right call — unifying them would change the event clients already pair on — but the two code paths are now structurally identical (`case … {:ok, row} -> publish/end_span/emit; :noop -> end_span`), so the next reader will try to collapse them. One line saying why the stage stays `done` would prevent that. Non-blocking nit.

4. **No CHANGELOG `[Unreleased]` entry.** This is a user-visible correctness fix, and no branch in the #1767 programme touches `CHANGELOG.md` (checked all 25). Cross-stack rather than this PR's alone, so not a blocker here — but someone owes one entry for the programme.

## Checked and clean

- **Line-count pin**: `conversation_server.ex` goes 2757 → 2755. Ratchet respected.
- **Missing / ambiguous exits** (the history this repo has with them): the `integer()` in the `Managoat.Sandbox` command contract holds on every adapter, and the genuinely-missing exit is `{:error, ref, :closed_before_exit}` since `managoat_sandbox` 0.2.0 — that lands on the `{:error, …}` clause → `fail_transport/2` → `TurnMachine.finish/4`, which is already routed through the same guarded `_unsafe_complete_turn/4`. So the two terminal frames are both covered, not just this one. The one residual ambiguity is upstream and unchanged by this patch: `Managoat.Runner.Connection` reads an `exit` frame as `Map.get(frame, "code", 0)`, so a runner frame with no code still commits `completed`/`0` — same before and after.
- **Second path to the same write**: grepped every caller of `_unsafe_complete_turn/4` and every terminal turn write in the server. `fail_transport`, `fail_before_start/6` and the interrupt halves are all bound; `handle_cast({:machine_gone, …})` at line 1685 is still unguarded and is explicitly scoped out in the body.
- **Nothing lost by not calling `update_conversation/2`**: that function is `changeset |> Repo.update() |> tap(broadcast_sidebar_update)`, and `end_running_turn/5` does both (the broadcast after commit, correctly, not inside the transaction). No audit call is bypassed, and `Conversations._unsafe_complete_turn/4` is already on the `audit_guardrail_test.exs` exclusion list.
- **Terminal-status preservation is scoped right**: only `terminated`/`failed` block the idle write, so a conversation in any other status still idles exactly as before.
- **`emit_turn_completed(state, ended.status)`** reads the pre-clear `state`, so `turn_metrics` and the tracer are still populated when the metric is emitted; the `:noop` branch deliberately emits none, matching `TurnMachine.finish/4`.
- **The peer-stop assertion is not a flake**: `send(owner, …)` from `Peer.terminate/2` and the peer's own exit signal share one sender, so Erlang's pairwise ordering guarantees the message is in the mailbox before `GenServer.stop/3` returns. `assert_received` (no wait) is correct here, not a race.
- **Reverting the fix fails the tests**: each of the eight `during_stop` cases asserts on a value the parent handler overwrites (`exit_code == nil`, `exit_code == 23`, `status == "running"`, `conv.status == "terminated"`), and the `:deleted` case would raise on the parent's `_unsafe_get_conversation!`. The first two assert `"running"` at peer-stop time, which the parent's write ordering cannot satisfy.
- `async: false` matches the sibling `turn_completion_binding_isolation_test.exs`, so it is stack consistency rather than a new exemption.

PR #2004, opened 2026-09-11
deps: managoat_runtimes 0.3.4, for the 0.75.1 adapter pin and the model-list warm-up

jhgaylor:
## Why

Two `managoat_runtimes` releases from 2026-09-07 were published and never consumed here. `main` still locks **0.3.2**, and production runs `main`, so neither fix is live:

- **0.3.3** moves the `claude-agent-acp` pin from 0.66.0 to **0.75.1**. The adapter's accepted model list is whatever the Claude Code binary bundled in its SDK reports. 0.66.0 bundles SDK 0.3.220 (CLI 2.1.220); 0.75.1 bundles SDK 0.3.257 (CLI 2.1.257), which also resolves a full model id onto the alias row the CLI advertises rather than matching it exactly.
- **0.3.4** adds `Managoat.Runtimes.Claude.prepare_sandbox/3`, warming the CLI's model list at provisioning. That binary learns an org's "additional models" from a fetch it makes *after* a session has started, caching it for the *next* launch — so the first session in a fresh sandbox saw a shorter list than the second one in the same sandbox, which is what made a model refusal read as intermittent.

This is the missing middle step of the three-PR sequence in #1729. That PR is approved and green but deliberately a draft: it re-suggests `claude-fable-5-1`, and until this bump is deployed, doing so would re-suggest an id that fails every first turn.

## The requirement moves too, not just the lock

`~> 0.3.2` already tolerates 0.3.4, so this would work as a lockfile-only change. Moving the requirement is deliberate, the same call #1725 made: the two fixes are the *reason* for the upgrade, and a clean checkout should resolve to them rather than to the oldest version the range allows.

## The warm-up needs no call-site change

`Provisioning.prepare_runtime_sprite/5` already dispatches through a capability guard, so the new callback is picked up by the bump alone:

```elixir
Code.ensure_loaded(runtime_module)

with :ok <- prepare_acp_adapter(handle, runtime, sprite_env) do
  if function_exported?(runtime_module, :prepare_sandbox, 3) do
    runtime_module.prepare_sandbox(handle, agent, sprite_env)
```

The warm-up is best-effort upstream: a cache that stays cold is logged and provisioning continues, so a credential the org refuses costs the bounded 30s and fails nothing.

## Scope

One line of `mix.lock`. `mix deps.get` also volunteered `gun` 2.5.0 → 2.6.0 and `cowlib` 2.19.0 → 2.20.0; both are unrelated to this change and were reverted to `main`'s versions, and the tree still resolves.

`0.4.x` was **not** taken on purpose: 0.4.0 requires `managoat_acp ~> 0.4.0`, a bump from the `~> 0.3.1` pinned here with nothing to do with Fable. 0.3.4's own requirements (`managoat_acp ~> 0.3.0`, `managoat_sandbox ~> 0.2.0`) are already satisfied, so no other pin moves.

## Verification

Against the resolved dependency, not the library's own suite:

```
$ grep -o '"managoat_runtimes": {:hex, :managoat_runtimes, "[^"]*"' mix.lock
"managoat_runtimes": {:hex, :managoat_runtimes, "0.3.4"

$ grep -n 'version: "0.75.1"' deps/managoat_runtimes/lib/managoat/runtimes/acp.ex
146:      version: "0.75.1"

$ grep -n 'def prepare_sandbox' deps/managoat_runtimes/lib/managoat/runtimes/claude.ex
147:  def prepare_sandbox(handle, _agent, sprite_env) do
```

Local gates: `mix compile --warnings-as-errors` clean, `mix format --check-formatted` clean, `mix deps.unlock --unused` no-op. `provisioning_steps_test`, `conversation_server_test`, `acp_model_contract_test`, `model_catalog_test`: **84 tests, 0 failures**. No test asserts an adapter version number — the suite pins the binary name `claude-agent-acp`, which is unchanged.

## After this merges

1. Confirm the image actually deployed (`build.yml` on `main`, not this PR's gate).
2. Drive a cold-sandbox turn on `anthropic/claude-fable-5-1` in production.
3. Then promote #1729 out of draft.

https://claude.ai/code/session_01PkZYEpwKRZTwhTc8cXMtLP


BinaryBourbon (approved):
Approving. The bump is correct, minimal, and the non-obvious risk checks out.

## The claim I went and verified

A version bump whose whole purpose is a behavior change earns a check that the behavior actually arrives. Two ways this could have shipped inert:

**1. The warm-up is dispatched.** `Provisioning.prepare_runtime_sprite/5` guards on `function_exported?(runtime_module, :prepare_sandbox, 3)` after `Code.ensure_loaded/1`, so the new callback is picked up by the bump with no call-site edit. Confirmed in the resolved dep, not just the changelog:

```
deps/managoat_runtimes/lib/managoat/runtimes/claude.ex:147:  def prepare_sandbox(handle, _agent, sprite_env) do
deps/managoat_runtimes/lib/managoat/runtimes/acp.ex:146:      version: "0.75.1"
```

**2. The warm-up is gated on a credential the broker takes away.** This is the one that worried me. Upstream, `prepare_sandbox/3` runs only `if credential?(sprite_env)`, which wants a non-empty `ANTHROPIC_API_KEY` or `CLAUDE_CODE_OAUTH_TOKEN` **in the sandbox env** — and under ADR 0019 a brokered tenant's sandbox holds a placeholder while the broker holds the value. A silent `:ok` for every brokered tenant would have made this a no-op in production, which is precisely where the verification is supposed to happen.

It survives, for two independent reasons:

- `Broker.split_inference/2` substitutes a **vendor-shaped** placeholder, not an empty string or a sentinel, so `credential?/1` sees a non-empty binary and the warm-up runs. The CLI's own key-shape check passes for the same reason.
- Ordering at the call site puts `Egress.install_ca/3` *before* `prepare_runtime_sprite/5`, and `sprite_env` already carries `Egress.sandbox_env/1`, so the warm-up's fetch egresses through the proxy with the CA trusted and the real key swapped in at the header.

So the fetch it depends on can complete for a brokered tenant. Worth knowing that this is load-bearing: if `split_inference/2` ever moves to a non-vendor-shaped placeholder, this warm-up goes quietly inert rather than failing.

## Scope

Right call on both restraints, and I checked each:

- **`gun`/`cowlib` reverted.** `mix deps.get` volunteered 2.5.0 → 2.6.0 and 2.19.0 → 2.20.0 with nothing in this change asking for them. `mix.lock` is one line now.
- **0.4.x correctly refused.** 0.4.0 requires `managoat_acp ~> 0.4.0` against the `~> 0.3.1` pinned here. 0.3.4's own requirements (`managoat_acp ~> 0.3.0`, `managoat_sandbox ~> 0.2.0`) are satisfied, so no other pin moves. `~> 0.3.4` also correctly excludes 0.4.0.

Moving the requirement rather than riding `~> 0.3.2`'s tolerance matches what #1725 did and for the same stated reason.

## Two notes, neither blocking

1. **Provisioning gets slower on the claude path.** One extra sandbox `exec` with a 90s exec timeout over a 30s cache poll. Upstream measured under three seconds and it is idempotent (a warm sandbox exits before doing work, so a shared sandbox's second conversation pays nothing), but the worst case is a credential the org refuses, which pays the full poll. Nothing here is wrong; it is the first thing to look at if provisioning p99 moves after deploy.
2. **The CHANGELOG entry will sit next to #1729's.** This one is at the top of `### Changed`, #1729's at the bottom, ~45 lines apart, so they merge without conflict. Whoever lands #1729 should keep both: this entry is the adapter and warm-up, that one is the catalog re-add.

## On the sequencing

This is the step that was actually missing, not a nice-to-have. #1729 has been sitting approved and green since 2026-09-07 behind a precondition nobody had done: `main` still locked 0.3.2, so production ran the 0.66.0 adapter with no warm-up, and promoting #1729 would have re-suggested an id that fails every first turn — the exact regression #1669 removed it for.

Merging this does not by itself clear #1729. The remaining gates are deploy confirmation on `main` and a real cold-sandbox turn on `anthropic/claude-fable-5-1` in production, which the PR body correctly lists and does not claim.

PR #2006, opened 2026-09-11
fix: scope machine-gone handling to the actor sandbox

jhgaylor:
A delayed machine-gone notification could stop an actor on a replacement sandbox. Cleanup could also idle a reassigned conversation or one with a newer running turn. The receiver now accepts a sandbox-qualified message and ignores notifications for another sandbox. After local cleanup, a guarded transaction rechecks the persisted binding, terminal status and running turns before idling the parent or publishing the sandbox event.

The old message remains supported for rollout compatibility. Deploy this receiver before migrating senders; legacy messages cannot distinguish an obsolete sandbox from the current actor's sandbox. The guard identifies the sandbox, not a new actor generation on the same sandbox.

Stack: based on #2003. Refs #1767. The sender migration is a separate draft. Durable forced-teardown recovery remains open follow-up work.

Validation:
- All 99 focused actor, audit and PostgreSQL tests passed.
- Real actor tests cover active and idle cleanup, ignoring another sandbox, and reassignment, terminal states, a new turn or deletion during cleanup.
- Two added PostgreSQL tests hold real admission open at capacity 1 and unbounded, proving cleanup waits and preserves the new running turn.
- Restoring the old cleanup body while retaining the new message routing produces five regression failures (15 selected tests).
- The initial test run caught a preloaded-versus-unloaded association comparison in a new assertion. Both snapshots now use the same database read; the failure log was retained.

The notification handler lives in `Lifecycle.machine_gone/4`; the server supplies its local cleanup callbacks. This keeps the combined lifecycle stack within the server-size guardrail and lowers the cap from 2,762 to 2,759 lines. An initial full run caught the oversized server and a separate 60-second documentation-rendering timeout in `Lumis.highlight/2`. The size issue was corrected; the documentation test passed with the actor and guardrail checks in a 129-test recheck. Original logs were retained, and final full validation runs sequentially.

Final integration validation: all 299 targeted tests passed on the combined lifecycle tree, including documentation and the server-size guardrail. The unpublished verification branch includes #1948, #1974, #1975, #1977 and #1979, with the reset-helper conflict resolved by keeping the shared helper and adding the sandbox ID to its notification.

Full `mix precommit` passed on the final commit: Fountain 5,495 tests and 6 doctests, Buzz 144 tests, Support 33 tests; zero failures. Static analysis, Dialyzer, Sobelow and production release assembly passed. The documentation timeout did not recur in this full run.

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2363,11 +2363,16 @@ defmodule Fountain.Conversations do
   first half retired its last running turn. `AutonomousTurnReaper.sweep_stuck_turns/0`
   selects running turns, and `ExecutionGuard._unsafe_recover_turn/3` returns
   `:noop` for a retired turn, so those recovery paths cannot repair that state.
-  Other paths can idle the parent: `MachineEvents.gone/4` and the lifecycle's
+  Other paths can idle the parent: `MachineEvents.gone/5` and the lifecycle's
   park and reclaim paths do so when they run.
 
   `follow_cotenants/2` sends `:machine_gone` before rebinding with `update_all`,
-  so a server that receives the cast has `gone/4`'s cleanup as a backstop.
+  so a server that receives the cast has `gone/5`'s cleanup as a backstop.
+  `gone/5` ignores a notification naming another sandbox and
+  `_unsafe_finish_machine_gone/2` answers `:noop` for a moved conversation
+  (#2006), but its idle write is deliberately not conditioned on the binding,
+  so a parent left `running` with no running turn is still released. The
+  backstop this paragraph relies on therefore survives the rebind.
   `tell_cotenants/3` skips the cast when `ConversationServer.whereis/1` misses,
   including a cross-pod registry miss, while the rebind still applies. The
   parent can then remain `running` until another cleanup path runs.
@@ -2411,13 +2416,29 @@ defmodule Fountain.Conversations do
   end
 
   @doc """
-  Finish an actor's machine-gone notification on its current binding.
+  Finish an actor's machine-gone notification, and release a stranded parent.
 
-  Lock the parent through the running-turn check and optional idle write, as
-  admission does. A moved, terminal or deleted conversation, or a newer running
-  turn, makes this a no-op. Only a running conversation changes status;
-  already-idle actors can still record the sandbox event. Publication happens
-  after commit, and this function performs no provider or actor I/O.
+  Lock the parent through the running-turn check and any idle write, as
+  admission does. A deleted conversation or a newer running turn makes this a
+  no-op. `:ok` says the notification is this actor's to narrate: the binding
+  still names the sandbox it closed and the conversation is not terminal. A
+  moved or terminal conversation answers `:noop`, so an obsolete actor emits no
+  sandbox event. Publication happens after commit, and this function performs
+  no provider or actor I/O.
+
+  The idle write is not conditioned on the binding, and that is deliberate.
+  Under the parent lock a `running` conversation with no running turn is not a
+  legitimate transient: admission inserts the running turn and sets the parent
+  `running` in one transaction under this same lock (`_unsafe_create_turn_on_sandbox/3`),
+  and the two other writers of `running` (`Reattachment` and `Connection`) both
+  have their running turn in hand. So that pair is a stuck row, and the sweeps
+  cannot repair it — `AutonomousTurnReaper.sweep_stuck_turns/0` and
+  `ExecutionGuard._unsafe_recover_turn/3` both select a running turn, and there
+  is none. `_unsafe_idle_interrupted_turn/1` declines to recheck the binding for
+  exactly this reason and names this function as its backstop (#2000); a
+  refusal here would strand the parent with nothing left to release it.
+  Releasing it is a repair, not an act of the stale actor, so it still answers
+  `:noop` and publishes nothing.
   """
   def _unsafe_finish_machine_gone(conversation_id, sandbox_id) do
     {:ok, result} =
@@ -2429,12 +2450,21 @@ defmodule Fountain.Conversations do
           from(t in Turn, where: t.conversation_id == ^conversation_id and t.status == "running")
 
         with %Conversation{} = conv <- Repo.one(conversation_query),
-             true <- conv.sandbox_id == sandbox_id and conv.status not in ["terminated", "failed"],
              false <- Repo.exists?(running_query) do
-          if conv.status == "running" do
-            {:updated, conv |> Conversation.changeset(%{status: "idle"}) |> Repo.update!()}
-          else
-            :unchanged
+          # A terminal conversation is never `running`, so this decides the
+          # answer, never the write.
+          current? = conv.sandbox_id == sandbox_id and conv.status not in ["terminated", "failed"]
+
+          cond do
+            conv.status == "running" ->
+              idled = conv |> Conversation.changeset(%{status: "idle"}) |> Repo.update!()
+              if current?, do: {:updated, idled}, else: {:released, idled}
+
+            current? ->
+              :unchanged
+
+            true ->
+              :noop
           end
         else
           _ -> :noop
@@ -2445,6 +2475,10 @@ defmodule Fountain.Conversations do
       {:updated, conv} ->
         broadcast_sidebar_update(conv.user_id)
         :ok
+
+      {:released, conv} ->
+        broadcast_sidebar_update(conv.user_id)
+        :noop
 
       :unchanged ->
         :ok
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -2411,6 +2411,50 @@ defmodule Fountain.Conversations do
   end
 
   @doc """
+  Finish an actor's machine-gone notification on its current binding.
+
+  Lock the parent through the running-turn check and optional idle write, as
+  admission does. A moved, terminal or deleted conversation, or a newer running
+  turn, makes this a no-op. Only a running conversation changes status;
+  already-idle actors can still record the sandbox event. Publication happens
+  after commit, and this function performs no provider or actor I/O.
+  """
+  def _unsafe_finish_machine_gone(conversation_id, sandbox_id) do
+    {:ok, result} =
+      Repo.transaction(fn ->
+        conversation_query =
+          from(c in Conversation, where: c.id == ^conversation_id, lock: "FOR UPDATE")
+
+        running_query =
+          from(t in Turn, where: t.conversation_id == ^conversation_id and t.status == "running")
+
+        with %Conversation{} = conv <- Repo.one(conversation_query),
+             true <- conv.sandbox_id == sandbox_id and conv.status not in ["terminated", "failed"],
+             false <- Repo.exists?(running_query) do
+          if conv.status == "running" do
+            {:updated, conv |> Conversation.changeset(%{status: "idle"}) |> Repo.update!()}
+          else
+            :unchanged
+          end
+        else
+          _ -> :noop
+        end
+      end)
+
+    case result do
+      {:updated, conv} ->
+        broadcast_sidebar_update(conv.user_id)
+        :ok
+
+      :unchanged ->
+        :ok
+
+      :noop ->
+        :noop
+    end
+  end
+
+  @doc """
   Reconciles a turn left `running` after its server or runtime disappeared.
 
   The turn transition and the conversation's `running` to `idle` transition
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -1495,8 +1495,20 @@ defmodule Fountain.Conversations.ConversationServer do
     MachineEvents.reset(state, sandbox_id, reason, by, message, &drop_connection/2)
   end
 
-  def handle_cast({:machine_gone, event, reason, message}, state) do
-    MachineEvents.gone(state, {event, reason, message}, &interrupt_turn/1, &drop_connection/2)
+  # Compatibility for senders deployed before the sandbox-qualified message.
+  # Roll out this receiver before migrating senders; the old tuple cannot
+  # distinguish an obsolete sandbox notification from one for this actor.
+  def handle_cast({:machine_gone, event, reason, message}, state),
+    do: handle_cast({:machine_gone, state.sandbox_id, event, reason, message}, state)
+
+  def handle_cast({:machine_gone, sandbox_id, event, reason, message}, state) do
+    MachineEvents.gone(
+      state,
+      sandbox_id,
+      {event, reason, message},
+      &interrupt_turn/1,
+      &drop_connection/2
+    )
   end
 
   # Catch-all for the same reason as the handle_call one above (#315).
--- a/apps/fountain/lib/fountain/conversations/machine_events.ex
+++ b/apps/fountain/lib/fountain/conversations/machine_events.ex
@@ -56,23 +56,41 @@ defmodule Fountain.Conversations.MachineEvents do
 
   def reset(state, _sandbox_id, _reason, _by, _message, _drop), do: {:noreply, state}
 
+  # Handle a notification for this actor's sandbox; ignore another sandbox's.
   # The server supplies its connection/turn teardown callbacks. Keeping the
   # transcript handling here leaves the actor's mailbox clauses small.
-  def gone(state, {event, reason, message}, interrupt, drop) do
+  #
+  # Cleanup runs before the guarded context transition, outside its
+  # transaction. A moved or terminal conversation, or a newer running turn,
+  # keeps its state and emits no sandbox event. The obsolete actor still stops
+  # after cleanup, because its handle is already gone with the machine.
+  def gone(
+        %{sandbox_id: sandbox_id} = state,
+        sandbox_id,
+        {event, reason, message},
+        interrupt,
+        drop
+      )
+      when is_binary(sandbox_id) do
     state = if state.current_turn, do: interrupt.(state), else: state
     state = drop.(state, event)
 
-    # ownership: the calling ConversationServer established this holder at init.
-    conv = Conversations._unsafe_get_conversation!(state.conversation_id)
-    if conv.status == "running", do: Conversations.update_conversation(conv, %{status: "idle"})
+    # Ownership: the actor supplies the sandbox whose local handle it closed.
+    case Conversations._unsafe_finish_machine_gone(state.conversation_id, sandbox_id) do
+      :ok ->
+        Output.publish_stage(state.conversation_id, "sandbox", "done", %{
+          event: event,
+          reason: reason,
+          by: "another_conversation",
+          message: message
+        })
 
-    Output.publish_stage(state.conversation_id, "sandbox", "done", %{
-      event: event,
-      reason: reason,
-      by: "another_conversation",
-      message: message
-    })
+      :noop ->
+        :ok
+    end
 
     {:stop, :normal, %{state | handle: nil}}
   end
+
+  def gone(state, _sandbox_id, _notification, _interrupt, _drop), do: {:noreply, state}
 end
--- a/apps/fountain/lib/fountain/conversations/machine_events.ex
+++ b/apps/fountain/lib/fountain/conversations/machine_events.ex
@@ -62,8 +62,11 @@ defmodule Fountain.Conversations.MachineEvents do
   #
   # Cleanup runs before the guarded context transition, outside its
   # transaction. A moved or terminal conversation, or a newer running turn,
-  # keeps its state and emits no sandbox event. The obsolete actor still stops
-  # after cleanup, because its handle is already gone with the machine.
+  # emits no sandbox event. A moved one can still be released: this is the
+  # backstop `_unsafe_idle_interrupted_turn/1` names, so the write that idles a
+  # parent stranded `running` with no running turn survives the rebind, while
+  # the transcript event does not. The obsolete actor still stops after
+  # cleanup, because its handle is already gone with the machine.
   def gone(
         %{sandbox_id: sandbox_id} = state,
         sandbox_id,
--- a/apps/fountain/test/fountain/audit_guardrail_test.exs
+++ b/apps/fountain/test/fountain/audit_guardrail_test.exs
@@ -178,7 +178,8 @@ defmodule Fountain.AuditGuardrailTest do
     "Conversations._unsafe_interrupt_turn/2 and _unsafe_idle_interrupted_turn/1" =>
       "conditional interrupt bookkeeping; its turn stage and public lifecycle action carry the trail",
     "Conversations._unsafe_finish_machine_gone/2" =>
-      "conditional actor bookkeeping; its sandbox stage records the successful notification",
+      "conditional actor bookkeeping; the sandbox stage records a current notification, and " <>
+        "releasing a parent the rebind stranded repairs machine state rather than tenant state",
     "ConversationServer per-turn state" => "high-volume machine state; log_events covers it",
     "Accounts.touch_api_key/1" => "a last-used stamp on every authenticated request",
     "Runners.touch/1 and reconnects" =>
--- a/apps/fountain/test/fountain/audit_guardrail_test.exs
+++ b/apps/fountain/test/fountain/audit_guardrail_test.exs
@@ -177,6 +177,8 @@ defmodule Fountain.AuditGuardrailTest do
       "conditional per-turn bookkeeping; the turn stage records completion outside its transaction",
     "Conversations._unsafe_interrupt_turn/2 and _unsafe_idle_interrupted_turn/1" =>
       "conditional interrupt bookkeeping; its turn stage and public lifecycle action carry the trail",
+    "Conversations._unsafe_finish_machine_gone/2" =>
+      "conditional actor bookkeeping; its sandbox stage records the successful notification",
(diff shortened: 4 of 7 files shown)


BinaryBourbon (approved):
APPROVE — the receiver is correct on this tree (legacy senders are not turned into no-ops), with one behavioural gap in the `follow_cotenants` path that the sender PR above should close rather than only re-qualify.

## What this changes

`handle_cast({:machine_gone, …})` moves out of the server into `Lifecycle.machine_gone/4`, which takes the actor's `interrupt_turn/1` and `drop_connection/2` as callbacks, and the unconditional `update_conversation(conv, %{status: "idle"})` becomes `Conversations._unsafe_finish_machine_gone/2`: a `FOR UPDATE` on the parent, then a binding + terminal-status + running-turn recheck, and only then the idle write. The sandbox stage event is now published only when that recheck passes. A four-element message is accepted for rollout compatibility and is rewritten as `{:machine_gone, state.sandbox_id, …}`, so the notification's sandbox id is the actor's own until #2007 migrates the senders.

## Findings

**1. `follow_cotenants` rebinds the row with no actor to take over, so the guard leaves the conversation `running`. Non-blocking** (`apps/fountain/lib/fountain/conversations.ex:2223`, sender at `apps/fountain/lib/fountain/conversations.ex:5282-5303`).

The guard's premise is "the row moved, so a newer actor owns its status". That holds for a wake/reapply, but `follow_cotenants/2` rebinds a *co-tenant's* row from a sibling conversation's provisioning and starts no server for it:

1. Conversation X is mid-turn on sandbox S1 with a live server. A co-tenant wakes, finds the sprite gone, provisions S2.
2. `tell_cotenants/3` casts `{:machine_gone, "replaced", "sprite_gone", msg}` to X's server, and *then* `Repo.update_all` sets `X.sandbox_id = S2`.
3. X's server runs `interrupt_turn/1` (ACP cancel, `close_stdin`, `stop_command` against a sprite that is gone) and `drop_connection/2` before reaching `_unsafe_finish_machine_gone(X, S1)`. `end_running_turn/5` holds `FOR UPDATE` on X while it writes the turn, so the queued `update_all` either lands before it or blocks on that lock and lands immediately after it — both interleavings put the rebind before the finish call.
4. `conv.sandbox_id` is now S2 → `:noop`. X stops. Row: `status: "running"`, no server.

Outcomes: if the rebind beat the interrupt write too, the turn is also left `running` and only `AutonomousTurnReaper` heals it, after `@stuck_after_minutes 30` + `@quiet_grace_minutes 15`. If the interrupt won the lock, the turn is `interrupted` and nothing heals the conversation status at all — the rehydrator sweeps on boot only (`rehydrator.ex` moduledoc), so the console shows X running until its next prompt, and `_unsafe_resume_gate/1` (`conversations.ex:4485`) refuses a detached-permission answer with `{:error, :busy}` for that whole window. The old code idled X immediately, and idling was the right answer here.

Not blocking: the transcript is intact (`follow_cotenants` publishes the `"replaced"` stage for every id itself), X is still wakeable (`assert_resumable/1` only refuses terminal statuses), and the PR body already books durable forced-teardown recovery as follow-up. The fix belongs with the sender, which already owns the `update_all` and the stage publish: have `follow_cotenants`/`strand_cotenants` write the status they now leave behind.

**2. For #2007: qualify with the sandbox that was lost, not the replacement. Non-blocking, cross-PR.**

`Lifecycle.machine_gone/4`'s fallback clause is `{:noreply, state}` — it neither cleans up nor stops. `follow_cotenants` has `new_sandbox_id` in hand and already puts it in its own stage meta; if the sender migration passes that id, every notification on this path hits the ignore clause, the actor keeps a handle to a dead sprite and the conversation sits running forever — the failure this PR ordering was meant to avoid, arriving from the other side. The id has to be the actor's own (old) sandbox.

**3. Nit: the size pin keeps 16 lines of slack.** `conversation_server_size_test.exs:18` pins 2759; the file is 2743 lines. The moduledoc says the pin is "the file's line count … lower it when you move something out", so 2743 is the number. Context for the tracker, not for this PR: `origin/main` is 2646/2646 and #1996 raised the pin to 2762 with the file at 2741, so across this stack a file that "only shrinks" is +97 over main. This is the only PR in the chain that lowers it.

**4. Nit: no CHANGELOG `[Unreleased]` entry anywhere in the nine-PR stack**, though "a stale actor no longer idles a conversation that has moved to a new sandbox" is user-visible. Soft, and probably one entry for the tracker rather than per-PR.

## Checked and clean

- **The no-op risk the ordering creates.** All three senders on this tree still send the four-element tuple (`conversations.ex:3869`, `conversations.ex:5339` via `tell_cotenants/3`, `lifecycle.ex:577`); the compat clause substitutes `state.sandbox_id`, so the sandbox comparison is trivially true and no notification becomes a no-op before #2007. `state.sandbox_id` is always a binary for a live server (`Keyword.fetch!` at `init/1`; `start_conversation_server/4` passes `sandbox.id`, the rehydrator passes `conv.sandbox_id` from a query that joins the sandbox, and the schema rejects a nil), so the `when is_binary(sandbox_id)` guard cannot route a live actor to the ignore clause. The pre-existing legacy-tuple test at `conversation_server_shared_sandbox_test.exs:302` still covers that clause.
- `Conversations.publish_stage/4` is what `Output.publish_stage/4` delegated to, so the Prometheus stage counter, the `conv:` PubSub broadcast, webhook dispatch and the analytics mirror are unchanged by the move.
- `_unsafe_finish_machine_gone/2` reproduces `update_conversation/2`'s only side effect (`broadcast_sidebar_update/1`, after commit) and matches `_unsafe_idle_interrupted_turn/2`'s lock discipline — parent `FOR UPDATE` before the running-turn `exists?`, which is the lock admission takes, so an admission cannot slip between the check and the write.
- A conversation deleted during cleanup now returns `:noop` instead of raising out of `_unsafe_get_conversation!/1` inside a `handle_cast` — the old shape crashed a `restart: :transient` child and spent the shared Horde restart budget. Strict improvement.
- Guard direction: `conv.sandbox_id == sandbox_id` compares the persisted binding against the actor's, the same side as the sibling guards; `:unchanged` still publishes, so an already-idle actor's transcript keeps the event.
- Audit: no `Audit.record/1` inside the new transaction, no values recorded, and the guardrail exclusion is added with a reason in the same style as its neighbours.
- Tests have teeth: the five-element routing test falls to the catch-all `handle_cast` if the routing is reverted; the obsolete-notification test synchronises with `:sys.get_state/1` after the cast, so its `refute_received` is not racing, and `stub_turn_boundary/0` delivers no exit event so the state comparison is not racing a turn completion; the during-cleanup tests assert `refute Repo.in_transaction?()` inside the stub, which pins the "cleanup outside the transaction" claim rather than assuming it; the two `interruption_admission_isolation_test.exs` cases hold real admission open at capacity 1 and unbounded.

jhgaylor:
**Reworked, not rebased — and the naive rebase was actively dangerous.** The approval above predates this.

## Why it could not be rebased

#1947 landed on `main` after this branch was cut and moved machine-gone handling into `Fountain.Conversations.MachineEvents.gone/4`. This PR adds a competing sandbox-qualified implementation to `Lifecycle.machine_gone/4`.

Cherry-picking it produced **exactly one conflict — the three-line dispatch site.** Everything else applied silently: `Lifecycle.machine_gone/4` landed at lines 571-598 with no marker, `_unsafe_finish_machine_gone/2` landed silently, and `MachineEvents.gone/4` sat untouched at line 61 still running its unguarded `_unsafe_get_conversation!` + `update_conversation(idle)`.

**Both resolutions of that conflict lose something, and only one loss is detectable.** Both were built and measured:

- *Keep main's dispatch* — compiles with `--warnings-as-errors`, **exit 0, zero warnings**, suite **1,520 tests 0 failures**. The new guarded functions become dead code (`def`, not `defp`, so nothing complains) while the **unguarded** original keeps serving traffic. Invisible to every gate — and it is the resolution a careful reader would pick.
- *Keep this PR's dispatch* — silently deletes the `{:sandbox_reset, ...}` dispatch to `MachineEvents.reset/6`, which this branch predates. Caught, but by accident: it orphaned the `MachineEvents` alias, so the compiler warned. Had anything else in the file still referenced that alias, it would have been compile-silent too, leaving one failing test.

So it is re-expressed into `MachineEvents.gone/4` instead. `main`'s structure wins, `MachineEvents` already owns `reset/6` beside it, and the server's dispatch already points there.

## Carried over

`_unsafe_finish_machine_gone/2` verbatim (it mirrors `_unsafe_idle_interrupted_turn`'s locking shape); the sandbox-qualified clause plus its `{:noreply, state}` fallback, reshaped as `gone/5` with a **positional** `sandbox_id` to match the neighbouring `reset/6`; the `is_binary` guard; the rolling-deploy compatibility clause (the pattern #1980/#1981 established); the audit-guardrail exclusion; all of this PR's tests. Kept `Output.publish_stage` (main's alias) over `Conversations.publish_stage`.

## Dropped: the `@pin` hunk

This PR changed `@pin 2762` → `@pin 2759`. That reads as a lowering and is measured against a stale base — `main`'s pin is **2549**, so applying it would have been a **210-line raise** that silently un-protects the file, as a clean-looking one-line change. Dropped entirely. `git diff origin/main -- .../conversation_server_size_test.exs` is empty; the file is 2535 against 2549.

## Rebase against today's main

One conflict, in `interruption_admission_isolation_test.exs`: `main` now has `_unsafe_idle_interrupted_turn/1`, because #2000 moved that binding fence out of SQL and into the process. This PR's test called the arity-2 form. Resolved by keeping this PR's `case ctx.ending` branch structure with main's arity-1 call — taking either side whole would have failed to compile or dropped the `:machine_gone` branch.

## Verification

Six mutations, each killing only its intended tests: the guard clauses in `_unsafe_finish_machine_gone` (6 failures), `gone/5` accepting any sandbox id (1), deleting the 4-tuple compatibility clause (1), `:unchanged -> :noop` (2), and both sender reversions (1 and 2). Compile `--warnings-as-errors`, format, `credo --strict` (8,387 mods/funs) all clean; 134 tests across the touched files and 1,466 across `test/fountain/conversations/` twice, 0 failures.

Re-review requested rather than assumed.

BinaryBourbon (changes requested):
REQUEST CHANGES — the re-expression is correct and lost nothing I can find, but the guard this PR exists to add is **not covered by any test**, and it silently removes a backstop that a docstring merged on `main` explicitly relies on. Both are cheap to close.

Reviewed at `4a0903a7` and again with #2007 on top, in a clean worktree on its own database. `mix format --check-formatted`, `MIX_ENV=test mix compile --warnings-as-errors --force` and `mix credo --strict` from the umbrella root (8,387 mods/funs) are clean. `test/fountain/conversations` + `audit_guardrail_test.exs`: 1,548 tests 0 failures on this commit (seed 424242), 1,550 on #2007's head across seeds 987654 and 31337.

## The re-expression: all four load-bearing checks pass

1. **One implementation.** `grep -rn machine_gone --include="*.ex"` finds no `def machine_gone` anywhere. The only handler is `MachineEvents.gone/5` (`machine_events.ex:67`); `main`'s unguarded `_unsafe_get_conversation!` + `update_conversation(idle)` body is gone, not orphaned. No dead `Lifecycle` twin.
2. **`{:sandbox_reset, …}` survived.** `conversation_server.ex:1494` still dispatches to `MachineEvents.reset/6`, unchanged.
3. **Rolling-deploy compatibility holds on this tree.** Both senders still emit the 4-tuple here (`conversations.ex:5889`, `lifecycle.ex:578`), the compat clause rewrites it with `state.sandbox_id` (always a binary — `Keyword.fetch!(args, :sandbox_id)` at `init/1:478`), so the comparison is trivially true and no notification becomes a no-op before #2007. Deleting the clause fails `a co-tenant told the machine is gone records it and stops`.
4. **The pin.** `git diff origin/main -- …/conversation_server_size_test.exs` is empty. `@pin 2549`, file 2535 (`main` 2523).

Both hand resolutions are right. `lifecycle_actions_test.exs` keeps #2069's `await_registered(other.id, 2_000)` with its comment intact — the Horde flake is not reintroduced. `interruption_admission_isolation_test.exs` keeps `case ctx.ending` over `main`'s arity-1 `_unsafe_idle_interrupted_turn(turn)`, and both arms genuinely reach `:noop` by different routes (`latest_turn?/2` false for `:interrupt`; the running-turn clause for `:machine_gone`), so the parametrisation still buys two cases.

## Blocking 1 — two of the three guard clauses are untested

Mutated one clause at a time against the full `test/fountain/conversations` tree plus `audit_guardrail_test.exs`:

| mutation of `_unsafe_finish_machine_gone/2` | result |
|---|---|
| drop `conv.sandbox_id == sandbox_id` | **0 failures** (1,550 tests) |
| drop `conv.status not in ["terminated", "failed"]` | **0 failures** (1,552 tests, incl. 2 probes of my own) |
| drop `false <- Repo.exists?(running_query)` | 3 failures |
| all three together | exactly **6 failures** |

That last row is the PR body's "6 failures", so the reported verification mutated the whole `with` line. Clause by clause, only the running-turn check has teeth — and the binding recheck is the entire subject of this PR.

The masking is systematic, not luck. Every `#{change} during machine-gone cleanup` case opens with `send_prompt`, so the actor holds a running turn. When the stub reassigns or terminalises the row, `end_running_turn`'s own binding/terminal guard (`conversations.ex:2288`, from earlier in this stack) refuses to retire that turn — so the turn is still `running` when `_unsafe_finish_machine_gone/2` runs, and the running-turn clause returns `:noop` no matter what the other two clauses say. There is no idle-turn variant of those five cases, and `a qualified notification stops its idle actor` passes a matching sandbox id, so it never exercises a mismatch either. (`an obsolete notification leaves the replacement actor and its turn alone` does have teeth, but against `gone/5`'s head guard, not against the transaction.)

One test closes it. This fails on `main`-of-this-branch only when the binding clause is removed:

```elixir
test "a reassigned conversation with no running turn is left alone" do
  user = insert_verified_user()
  s1 = insert_sandbox(user_id: user.id, status: "terminated")
  s2 = insert_sandbox(user_id: user.id, status: "ready")
  conv = insert_conversation(user_id: user.id, sandbox: s1, status: "running")
  turn = insert_turn(conv, status: "running")
  assert {:ok, _} = Conversations._unsafe_interrupt_turn(turn, s1.id)   # the actor won the lock
  {1, _} = Repo.update_all(from(c in Conversation, where: c.id == ^conv.id), set: [sandbox_id: s2.id])

  assert :noop = Conversations._unsafe_finish_machine_gone(conv.id, s1.id)
  assert Repo.reload!(conv).status == "running"
end
```

A terminal-status equivalent (no running turn, `status: "terminated"`) closes the other one.

## Blocking 2 — the recheck voids the argument `_unsafe_idle_interrupted_turn/1` is built on

`conversations.ex:2362-2374`, on `main` today, from #2000:

> Rechecking the binding here could leave the parent `running` after the first half retired its last running turn. `AutonomousTurnReaper.sweep_stuck_turns/0` selects running turns, and `ExecutionGuard._unsafe_recover_turn/3` returns `:noop` for a retired turn, so those recovery paths cannot repair that state. Other paths can idle the parent: **`MachineEvents.gone/4`** and the lifecycle's park and reclaim paths do so when they run.
>
> `follow_cotenants/2` sends `:machine_gone` before rebinding with `update_all`, so a server that receives the cast has **`gone/4`'s cleanup as a backstop**.

#2000 declined to recheck the binding *because* `gone/4` idled unconditionally. This PR adds that recheck to the backstop, so on the `follow_cotenants` path the backstop no longer fires — and the paragraph now describes a mechanism that does not exist, which is the ADR/docstring rule in CLAUDE.md.

It is reachable, not theoretical. Run against this branch, all three assertions hold:

```
conv running on S1, turn already `interrupted` by the actor, row rebound to S2 by follow_cotenants
  _unsafe_finish_machine_gone(conv.id, S1) -> :noop
  conv.status == "running",  no running turn,  ConversationServer.whereis(conv.id) == nil
```

`follow_cotenants/3` starts no server for the row it rebinds (`conversations.ex:5830-5852`), and it casts before the `update_all`, so the actor's finish transaction routinely reads the new binding. `AutonomousTurnReaper` selects running *turns* and there is none; the rehydrator sweeps on boot only. On `main` this conversation is idled. This is the earlier review's finding 1, which booked the fix to the sender PR — #2007 re-qualifies the message but does not write the status, so as the stack stands the gap ships.

Either fix works: have `follow_cotenants/3` write `status: "idle"` in the same `update_all` that rebinds (it already owns the write and the stage publish), or scope the recheck so a rebind with no successor turn still idles. If it is deliberately deferred, the #2000 docstring needs to stop naming `gone/4` as the backstop and say what covers it instead.

## Nits

- Arity drift in that same docstring after this PR: `MachineEvents.gone/4` (2366, 2370), `follow_cotenants/2` (2369), `tell_cotenants/3` (2371) are now `/5`, `/3`, `/4`.
- The comment at `machine_events.ex:64` ("A moved or terminal conversation … keeps its state and emits no sandbox event") is accurate but is the natural place for the stranding caveat.
- No `[Unreleased]` CHANGELOG entry anywhere in the stack, though "a stale actor no longer idles a conversation that has moved" is user-visible. One entry for the tracker is probably right.

## Checked and clean

- `_unsafe_finish_machine_gone/2` reproduces `update_conversation/2`'s only side effect (`broadcast_sidebar_update/1`, after commit) and matches `_unsafe_idle_interrupted_turn/1`'s lock discipline: parent `FOR UPDATE` before the running-turn `exists?`, the same lock admission takes.
- No audit inside the transaction; the guardrail exclusion is added with a reason in its neighbours' style.
- A conversation deleted during cleanup returns `:noop` instead of raising out of a `handle_cast` and spending the Horde restart budget. Strict improvement.
- `:unchanged -> :ok` still publishes, so an already-idle actor keeps its transcript event; mutating it to `:noop` fails two tests.
- The during-cleanup stubs assert `refute Repo.in_transaction?()`, which pins the "cleanup outside the transaction" claim rather than assuming it.

jhgaylor:
Both CHANGES_REQUESTED findings addressed. Head is `664e3402`, two commits on current `main`.

## Finding 2 — the backstop, and why the fix takes this shape

You were right that this removed a backstop `main` declares. #2000's docstring says `_unsafe_idle_interrupted_turn/1` skips the binding recheck **because** `gone/4` is unconditional, and this PR added a recheck to exactly that backstop.

The fix **splits the two decisions the single `with` line was making at once**. The binding and terminal checks now decide only whether this actor may *narrate* the notification; the idle write stays unconditional:

```elixir
current? = conv.sandbox_id == sandbox_id and conv.status not in ["terminated", "failed"]

cond do
  conv.status == "running" ->
    idled = conv |> Conversation.changeset(%{status: "idle"}) |> Repo.update!()
    if current?, do: {:updated, idled}, else: {:released, idled}

  current? -> :unchanged
  true -> :noop
end
```

`{:released, _}` broadcasts the sidebar and returns `:noop`, so a stale actor repairs the row but publishes no sandbox stage.

**Why the unconditional write is safe.** Under the parent `FOR UPDATE` lock, "`running` with no running turn" is not a legitimate transient. Every writer of `status: "running"` was enumerated: `_unsafe_create_turn_on_sandbox/3` (`conversations.ex:1877`, inserts the running turn and sets the parent in one transaction under this same lock), `Reattachment` (`reattachment.ex:301`, already holding `running_turn`), and `Connection` (`connection.ex:348`, after creating a running turn). A turn admitted `pending` never sets the parent `running`. So that pair is always a stuck row, and idling it is a repair, not an act of the stale actor.

**Rejected alternatives, since the shape matters:**
- *Scope only the turn handling, leave the parent idle unconditional* — collapses to a full revert. This function writes no turn; the turn is retired by `interrupt.(state)` beforehand, which carries its own binding fence at `conversations.ex:2288`. What remains **is** the parent idle plus the publish decision.
- *Have `follow_cotenants/2` set `status: "idle"` in its rebinding `update_all`* — worse on three counts: it would idle co-tenants that still have a running turn (leaving turn and parent disagreeing, the exact pair `:1877`'s comment exists to prevent); it covers neither `strand_cotenants/1` nor the `tell_cotenants/3` registry-miss path the #2000 docstring also names; and it patches one producer rather than the invariant.

The invariant is restored: the only refusals that now leave a row unwritten are ones a sweep can reach (a running turn exists → `AutonomousTurnReaper`) or where there is nothing to repair (idle, terminal, deleted).

### Probe

| scenario | `main` | `4a0903a7` | `664e3402` |
|---|---|---|---|
| **rebound, turn `interrupted`, no running turn** | `:ok` / **idle** | `:noop` / **running** ← stranded | `:noop` / **idle** |
| rebound, running turn on the new binding | `:ok` / idle ← the bug this PR fixes | `:noop` / running | `:noop` / running |
| idle, rebound | `:ok` / idle | `:noop` / idle | `:noop` / idle |
| terminal, matching binding | `:ok` / terminated | `:noop` / terminated | `:noop` / terminated |
| matching binding, no running turn | `:ok` / idle | `:ok` / idle | `:ok` / idle |

Row 1 restored to `main`'s outcome; every row this PR improved is unchanged.

## Finding 1 — the guard had no test

Correct, and the masking was systematic: every existing `during machine-gone cleanup` case sends a prompt first, so the running-turn clause short-circuits before the other two are evaluated. The "6 failures" in my earlier comment came from mutating the whole `with` line at once, which is exactly what hid it. Your test is included, and each clause is now verified under its own single-clause mutation.

1,560 tests across `test/fountain/conversations/` + `audit_guardrail_test.exs`, 0 failures. Compile `--warnings-as-errors` and format clean. Pin untouched — `git diff origin/main` on the size test is empty; file 2535 against 2549.

BinaryBourbon (approved):
Adversarial review — transaction and concurrency correctness. **APPROVE** at `664e34021e8a47ce23148093c3efac2f052642ba`.

No actionable regression found. I traced the changed cleanup through turn interruption, admission, reattachment, co-tenant rebinding, actor teardown, and the added regression assertions.

- The parent `FOR UPDATE` lock spans the running-turn check and idle write ([conversations.ex:2443–2472](https://github.com/managoat/fountain/blob/664e34021e8a47ce23148093c3efac2f052642ba/apps/fountain/lib/fountain/conversations.ex#L2443-L2472)). Admission takes that same lock before inserting a turn and setting its parent running, including `:unbounded` capacity; the new helper does not introduce an inverse lock order.
- A rebound parent is released only when running with no running turn, while terminal parents and running successors are preserved. The rebound repair returns `:noop`, suppressing the obsolete sandbox event. Provider cleanup precedes the transaction and event publication follows commit ([machine_events.ex:70–95](https://github.com/managoat/fountain/blob/664e34021e8a47ce23148093c3efac2f052642ba/apps/fountain/lib/fountain/conversations/machine_events.ex#L70-L95)).
- The concurrency regression uses independent database sessions and observes a PostgreSQL lock wait before releasing admission, then checks both parent and successor remain running ([interruption_admission_isolation_test.exs:26–62](https://github.com/managoat/fountain/blob/664e34021e8a47ce23148093c3efac2f052642ba/apps/fountain/test/fountain/conversations/interruption_admission_isolation_test.exs#L26-L62)).

Evidence for this reviewer is static source and test inspection; I did not execute tests locally.

BinaryBourbon (approved):
Adversarial lifecycle/rollout review of `664e34021e8a47ce23148093c3efac2f052642ba`: **APPROVE**. No blocking defect found in the current changes.

Traced both cotenant sender paths through the receiver and teardown callbacks. The [legacy receiver](https://github.com/managoat/fountain/blob/664e34021e8a47ce23148093c3efac2f052642ba/apps/fountain/lib/fountain/conversations/conversation_server.ex#L1498) preserves the deployed tuple format; the qualified path ignores another sandbox before any cleanup. Sender qualification remains a separate rollout step, as documented.

The [guarded finish](https://github.com/managoat/fountain/blob/664e34021e8a47ce23148093c3efac2f052642ba/apps/fountain/lib/fountain/conversations.ex#L2443) shares admission's parent lock, preserves running successors and terminal/deleted conversations, and releases a rebound parent only when no running turn remains. This preserves the cotenant rebind recovery backstop while suppressing obsolete sandbox narration. Provider cleanup and event publication remain outside the transaction.

Evidence: independently inspected all seven changed files, lifecycle/cotenant callers, admission and interruption paths, and regression assertions for legacy/qualified messages, rebinding, terminal/deleted rows, and concurrent admission. This reviewer performed source review; no local test execution is claimed.

BinaryBourbon (approved):
Adversarial regression review of `664e34021e8a47ce23148093c3efac2f052642ba`: **APPROVE**. No blocking findings.

The binding and terminal guards now have independent, non-masking coverage. The [rebound-idle and same-binding terminal cases](https://github.com/managoat/fountain/blob/664e34021e8a47ce23148093c3efac2f052642ba/apps/fountain/test/fountain/conversations/machine_gone_binding_test.exs#L38-L55) reach the guard with no running turn, so the running-turn check cannot hide a missing binding or terminal check. The [rebound-parent cases](https://github.com/managoat/fountain/blob/664e34021e8a47ce23148093c3efac2f052642ba/apps/fountain/test/fountain/conversations/machine_gone_binding_test.exs#L76-L96) also distinguish releasing a stranded parent from preserving a busy replacement.

Validation in an isolated snapshot/database, using Elixir 1.19.5 / OTP 28 and seed 2006:

- `machine_gone_binding_test.exs`, `conversation_server_shared_sandbox_test.exs`, and `interruption_admission_isolation_test.exs`: **31 tests, 0 failures**, including real PostgreSQL admission interleavings for both capacities.
- Independently removing only the binding predicate: **8 tests, 2 expected failures**, specifically the two rebound refusals.
- Independently removing only the terminal predicate: **8 tests, 2 expected failures**, specifically terminated and failed parents.

Mutation edits were confined to the isolated copy and restored afterward. No full-suite rerun or mixed-version production rollout was performed.

PR #2007, opened 2026-09-11
fix: identify the lost sandbox in machine-gone notifications

jhgaylor:
Reset, reclamation and co-tenant wake notifications now carry the ID of the sandbox that was lost. A delayed notification can therefore be ignored by an actor on a replacement sandbox. Wake threads the captured old ID through both co-tenant paths, including conversations that follow onto the new sandbox and those that retain their separate identity.

Stack: based on #2006. Refs #1767. Deploy the receiver first: older actors do not understand the qualified message. The receiver keeps legacy-message support for existing senders during the transition.

This change migrates the three notification senders. It does not change the co-tenant reassignment writes or introduce durable forced-teardown recovery.

Validation:
- All 87 focused tests passed, covering reset, lifecycle actions, wake and the real receiver.
- Captured notifications verify the sandbox ID for reset and reclamation, and both following and stranded co-tenants after wake.
- Restoring the old senders produces four failures in 54 selected tests; the migration is restored.

Integration note: #1948's ancestry extracts reset notifications into `record_reset_completed/3`. The verification merge retained that helper and added `completed.id` to its message, resolving the single conflict without duplicating the reset audit or notification logic. Apply the same resolution when combining these stacks.

Final integration validation: all 299 targeted tests passed on the combined lifecycle tree, including documentation and the server-size guardrail. The unpublished verification branch includes #1948, #1974, #1975, #1977 and #1979, with the reset-helper conflict resolved as described.

Full `mix precommit` passed on the final commit: Fountain 5,497 tests and 6 doctests, Buzz 144 tests, Support 33 tests; zero failures. Static analysis, Dialyzer, Sobelow and production release assembly passed.


BinaryBourbon (approved):
APPROVE — all three senders are migrated, each passes the sandbox the receiver actually compares against, and nothing else in the tree constructs the message.

**What this changes**

The three places that cast `{:machine_gone, …}` now name the lost sandbox in the message, so #2006's receiver can drop a notification aimed at a binding the conversation no longer has. Reset passes `completed.id`, `Lifecycle.stop_cotenants/5` threads its own `sandbox_id` through to the cast, and the wake path threads the captured `old_sandbox_id` into both `follow_cotenants/3` and `strand_cotenants/2` by way of `tell_cotenants/4`. The receiver's legacy 4-tuple clause is left in place for the rollout, so nothing in-tree sends it any more.

**Findings**

1. *Non-blocking, pre-existing — a following co-tenant can still be left `running`.* `conversations.ex:5288-5296`: `follow_cotenants/3` casts with `old_sandbox_id` and then `Repo.update_all` moves those rows to `new_sandbox_id`. If the co-tenant's server is scheduled after that commit, `_unsafe_finish_machine_gone/2` (`conversations.ex:2231`) reads `conv.sandbox_id == new_sandbox_id`, mismatches the message's `old_sandbox_id`, and returns `:noop`. Scenario: B is mid-turn on the dead sprite when A wakes; B's server interrupts the turn, then the ownership check no-ops, so B stops with `status: "running"`, no server and no running turn, until the reaper's pass. This is not a regression from your patch — before it the legacy receiver substituted `state.sandbox_id`, which is the same old id, so the comparison and the race were identical — and your body explicitly scopes the reassignment writes out. Worth a follow-up: the cleanest fix is on the sender side (normalize `status` in `follow_cotenants/3`'s `update_all`), not by passing the new id, since the receiver's first clause requires the message id to equal the actor's own `state.sandbox_id`.

2. *Non-blocking — the rollout note is not achievable if #2006 and #2007 land in one release.* "Deploy the receiver first" needs two releases. In a single release, a rolling restart has old pods' actors hit the catch-all `handle_cast/2` (`conversation_server.ex:1758` on `main`), log `unexpected cast` and ignore the qualified message, so such an actor keeps a handle to a dead sprite and stays `running` until the reaper. Bounded and short-lived, but if the sequencing is meant to be real, #2006 has to ship on its own first.

3. *Nit* — `sandbox_reset_test.exs:348`: the new `assert_receive` takes the default timeout while the sibling assertion at `:103` uses `1_000`. The cast is already in the mailbox when `wake_conversation/1` returns, so this only needs the stand-in to be scheduled, but this suite has form for timeout flakes on loaded partitions.

4. *Nit, pre-existing* — `lifecycle.ex:569`: `stop_cotenants/5`'s spec still admits `sandbox_id :: String.t() | nil`, while `_unsafe_list_cotenant_ids/2` guards on `is_binary/1` and the new receiver clause does too. A `nil` would raise in the query before any cast, so the `nil` in the spec is cosmetic either way.

5. *Soft* — nothing in the #1767 stack adds a `CHANGELOG` `[Unreleased]` entry, and a conversation left `running` because a stale actor clobbered a rebound binding is reader-visible. The top of the stack is the natural place for one `Fixed` line covering the whole chain.

**Checked and clean**

- Every sender, by grep rather than by the diff: `git grep machine_gone` across `apps/fountain`, `ee`, `apps/fountain_buzz`, `apps/fountain_support` and `cli` finds exactly three `GenServer.cast` senders — `conversations.ex:3869` (reset), `conversations.ex:5339` (`tell_cotenants/4`, both wake paths) and `lifecycle.ex:577` (`stop_cotenants/5`, park and reclaim) — and all three are migrated. No worker, reaper, rehydrator, controller or test helper constructs the shape; `sandbox_reaper.ex:177`, `autonomous_turn_reaper.ex:66` and `deletion.ex:164` only call `whereis/1` for liveness.
- Each id is the lost sandbox *and* matches what the receiver compares against. `finish_sandbox_reset/1` updates the same row in place, so `completed.id == fenced.id == sandbox.id` and the conversations still name it (the reset only cleared `runtime_session_id`). `stop_cotenants/5`'s two call sites (`finish_park/4`, `destroy/5`) pass the actor's own parked or destroyed sandbox, and neither moves a co-tenant row. `move_cotenants/3` passes `old_sandbox_id`, captured before the waking conversation was rebound, and `strand_cotenants/2` deliberately leaves the rows naming it, so the stranded branch matches.
- Arity split on the receiver is unambiguous: the legacy clause at `conversation_server.ex:1687` matches only 4-tuples, the qualified one at `:1690` only 5-tuples, and `conversation_server_shared_sandbox_test.exs:302` still covers the legacy shape for the other rollout direction.
- The tests fail on revert rather than on a stub: all three updated assertions pin `^sandbox_id` positionally, so a 4-tuple cast never matches them. The new parametrized test asserts the per-branch discriminator (`"replaced"` vs `"reset"`) and the resulting row binding, and registers its stand-in *after* `reset_sandbox/2`, so the only cast in the mailbox is the wake's.
- `conversation_server.ex` is untouched, so the line-count pin is unaffected. No tenant-scoping, audit-in-context or `Task.async` rule is in play; the test stays `async: true`.

BinaryBourbon (approved):
APPROVE on its own diff — the sender migration is correct, complete and mutation-covered, and it qualifies with the **old** sandbox, which was the cross-PR hazard flagged on #2006. Two things the merger needs, neither a defect in this diff: it is stacked on #2006, where I have requested changes, and the "receiver first" rollout the body asks for is not achievable if both land in one release.

Reviewed at `db0f2a01`. `mix format --check-formatted`, `MIX_ENV=test mix compile --warnings-as-errors --force`, `mix credo --strict` from the umbrella root (8,387 mods/funs, no issues). `test/fountain/conversations` + `audit_guardrail_test.exs`: 1,550 tests 0 failures on seeds 987654 and 31337, and again under six other mutation runs. Clean worktree, own database.

## Verified

**Every sender migrated, none left behind.** `grep -rn "GenServer.cast(pid, {:machine_gone" --include="*.ex"` returns exactly two sites, both 5-tuple: `conversations.ex:5889` (`tell_cotenants/4`, reached by both `follow_cotenants/3` and `strand_cotenants/2`) and `lifecycle.ex:578` (`stop_cotenants/5`). The only 4-tuple send anywhere is the deliberate compat test at `conversation_server_shared_sandbox_test.exs:364`.

**The shapes match the receiver**, including the compatibility path: `{:machine_gone, sandbox_id, event, reason, message}` lands on `conversation_server.ex:1504`, and #2006's 4-tuple clause at 1501 is now exercised only by that one test — which still passes, so a pod running the old sender against this receiver is still handled.

**It qualifies with the lost sandbox, not the replacement.** This was the specific way this PR could have gone wrong: `follow_cotenants/3` has `new_sandbox_id` in hand, and passing it would have routed every notification on that path to `gone/5`'s ignore clause, leaving the actor holding a dead sprite forever. It passes `old_sandbox_id`. Mutation-checked:

| mutation | result |
|---|---|
| `tell_cotenants(ids, new_sandbox_id, "replaced", …)` | 1 failure (`a following co-tenant is told which old sandbox was lost`) |
| `tell_cotenants` reverted to the 4-tuple | 2 failures (both new `sandbox_reset_test` cases) |
| `lifecycle.ex:578` reverted to the 4-tuple | 1 failure (`stop_cotenants/5 casts :machine_gone to every other live server`) |

Each guard dies on its own mutation, which is the standard this campaign has been held to.

**The `lifecycle_actions_test.exs` conflict resolution is right.** #2069's `assert {:ok, ^stand_in} = ConversationServer.await_registered(other.id, 2_000)` and its "registration can precede lookup visibility in Horde" comment are intact above the changed assertion; only the tuple shape moved. The Horde registry flake fixed in #2069 is not reintroduced. The two new `sandbox_reset_test` cases follow the same discipline — register, `assert_receive {:registered, ^actor}`, then `await_registered` — before waking, so the `assert_receive` on the cast is not racing registration.

**Parametrising over `follows?` earns its keep**: the following case asserts the row rebinds to the new sandbox while the notification still names the old one, and the stranded case asserts the row keeps the retired id. Those are the two different rebinding behaviours of `follow_cotenants/3` and `strand_cotenants/2`, so the same assertion is not being run twice.

## Two things for the merger

**1. Do not merge before #2006 is resolved.** This is based on `codex/p2-1767-machine-gone-receiver`, and #2006 has two blocking findings from me: the binding recheck in `_unsafe_finish_machine_gone/2` has no test that dies when it alone is removed (verified against 1,550 tests), and it removes the `gone/4` backstop that #2000's `_unsafe_idle_interrupted_turn/1` docstring on `main` explicitly relies on. This PR does not make either worse — before it, #2006's compat clause substitutes `state.sandbox_id`, which is the same value this PR now sends — but it does not close them either, and the earlier review on #2006 booked the `follow_cotenants` status write to this PR. The body scopes it out ("It does not change the co-tenant reassignment writes"), which is a fair call to make explicitly; it just means the gap ships with the stack unless #2006 handles it.

**2. "Deploy the receiver first" is unachievable as a stack.** The compat clause covers new receiver / old sender. Nothing covers new sender / old receiver: a pre-#2006 pod has only the 4-tuple clause, so a 5-tuple cast falls to `handle_cast(msg, state)` at `conversation_server.ex:1514`, logs `unexpected cast`, returns `{:noreply, state}` — the actor never stops, keeps its handle on a dead sprite, and the conversation is not idled. Because this PR is based directly on #2006, both land in one image unless someone deliberately holds it for a release, and then every rolling deploy has a window where new pods cast across to old ones. `{:sandbox_reset, …}` had no such two-step, so there is no precedent in the tree to lean on.

Severity is limited — the old pods are being replaced during that same deploy, and the rehydrator sweeps on boot, so most of the stranding self-heals — but it is the exact failure this campaign is closing, and it is worth being a decision rather than an accident. Either hold this PR for the release after #2006, or say in the body that the window is accepted and why.

## Nit

No `[Unreleased]` CHANGELOG entry in this PR or anywhere else in the stack. One entry for the tracker rather than per-PR is probably the right shape.

PR #2010, opened 2026-09-12
docs: propose user-owned ChatGPT grants (ADR 0052)

jhgaylor:
Admins can already connect a ChatGPT account, but individual users cannot link their own subscription with Fountain managing token renewal. ADR 0052 proposes extending the existing grant storage and Codex broker path to tenant-owned accounts.

The proposal covers device-code linking in account settings and the API, tenant-key encryption, refresh coordination across servers, explicit account identity through sandbox lifecycle operations, and disconnect/reconnect fencing. Broker session creation and rule updates serialize against grant revocation, and every authenticated request—including requests inside existing tunnels—checks the active grant generation. Managed grant tokens are excluded from custom-template inputs and restricted to the protected Codex destination for both user and platform accounts. The broker rejects protocol upgrades before token injection, fails closed on unexpected `101` responses, and drains legacy upgraded connections before enabling the protected HTTP-only path.

User subscriptions count as BYO inference; the proposed default prefers the subscription, offers an explicit API-key choice, and never silently falls back to paid inference. Initial support is limited to Codex on broker-enabled backends.

Includes implementation sequencing, acceptance checks for stale issuance after disconnect, cached tunnels, raw-client upgrades, and direct/cross-key template token export, alternatives, and a refreshed ADR index. The ADR is **Proposed**, and explicitly identifies all user-level behavior as unbuilt. This PR changes documentation only.

Validation:
- `okf validate decisions`: passed; six pre-existing warnings, none in ADR 0052.
- ADR index regenerated and checked for stable output.
- `git diff --check`: passed.
- `mix precommit`: could not run because dependencies are not installed in the isolated documentation worktree.


decisions/0052-user-owned-chatgpt-grants.md
```diff
@@ -0,0 +1,286 @@
...
+Pin source/generation for a running peer. Normal token rotation updates
+the broker without rewriting its auth file. Source changes take effect on
+new peers; reconnecting to a different account invalidates old-generation
+peers and requires reauthentication before another turn. Do not swap only
+the bearer beneath an old account ID.
+
+Disconnect invalidates all broker sessions using that grant, removes cached
+credentials, and fences refresh/link completion across nodes. Do not report
+completion while known sessions can still start authenticated requests;
+retry invalidation failures and expose pending status. Already-forwarded
```
BinaryBourbon: **[P1] Fence broker issuance against disconnect, not only grant lifecycle writes**

Node A can select active generation G and retain its bearer, then node B can disconnect G, invalidate every existing session, and report success. If A subsequently calls `Broker.prepare` with its earlier selection, it creates a new session carrying G's still-valid token. That session was not among the "known sessions" invalidated by disconnect. Stale credential-rule writes after account replacement have the same problem.

`Broker.Native.Sessions.create/1`, `update_rules/4`, and `lookup/1` (`apps/fountain/lib/fountain/broker/native/sessions.ex`) currently operate on materialized encrypted rules and TTL without grant ID/generation/state. Section 3's grant-row fencing does not by itself fence those operations.

Please require source/generation to be durable broker authorization data and fence session creation and rule writes atomically against disconnect/replacement, or reject stale generations at request dispatch. Preserve invalidation of cached tunnel credentials as well: lookup occurs once per CONNECT tunnel, so row deletion alone is insufficient.

Add a two-node acceptance case that pauses provisioning after selection, completes disconnect, resumes provisioning, and proves no old-generation request can authenticate.

decisions/0052-user-owned-chatgpt-grants.md
```diff
@@ -0,0 +1,286 @@
...
+Environment/vault API-key overrides must resolve into the same explicit
+source decision; do not allow an env key to silently override a selected
+subscription later during provisioning.
+
+### 5. Pin sandbox identity and invalidate access
+
+Use the selected grant to build Codex's placeholder `auth.json`; eliminate
+the global `PlatformChatGPT.sandbox_auth/0` lookup from the user path.
+Refresh broker rules from that exact grant. Preserve the existing
+ChatGPT HTTP transport and ACP authentication behavior.
```
BinaryBourbon: **[P2] Decide whether user-defined bindings may export managed ChatGPT bearers**

Reusing the current broker path preserves two export routes. `Broker.split_inference/2` permits a tenant binding for `CODEX_CHATGPT_ACCESS_TOKEN` to override the implicit provider destination; a user can direct it to a controlled host that echoes the bearer. Separately, `Broker.Native.binding_rules/4` (`apps/fountain/lib/fountain/broker/native.ex:352–353`) supplies the entire `brokered` map to custom templates. A binding for another tenant secret can therefore use `{"Authorization":"Bearer {{ CODEX_CHATGPT_ACCESS_TOKEN }}"}` and extract the same bearer. Reserving only the grant's named binding would not close that route.

This is inherited behavior, not a new cross-tenant vulnerability introduced here. Creating bindings already requires full-scope authority (`router.ex:413–417`). It also affects the existing platform-grant path. For the new user path, however, it needs to be reconciled with the proposed managed-token custody boundary.

Please explicitly reserve managed grants to provider-controlled destinations, exclude their values from arbitrary custom-template credential maps, and reject or ignore conflicting bindings. Alternatively, explicitly document that full-scope users may export their bearer through brokerage. Add acceptance cases for both direct overrides and cross-key template references.

BinaryBourbon (changes requested):
Independent adversarial review by a separate Codex agent of `eac78a74019f31a0b50d7fb7bbf5bbd615ad473c`, posted as BinaryBourbon at the user's request.

Requesting changes to the proposed architecture:

1. **P1 — Disconnect must fence broker issuance and rule updates.** Invalidating known sessions and fencing grant-row writes still permits an in-flight provision with a previously selected bearer to create fresh broker access after unlink succeeds.
2. **P2 — Define the managed-bearer export boundary.** Existing tenant-defined bindings and cross-key custom templates can send managed ChatGPT bearers to tenant-controlled destinations. This is inherited transport behavior, not a new cross-tenant exploit introduced by the ADR, but the proposal needs an explicit custody decision.

Inline comments include concrete interleavings, code evidence, and acceptance cases. This was a read-only architecture/code review; no runtime exploit tests were performed. The draft already covers tenant encryption, refresh coalescing, stale terminal failures, connection cancellation, auth-file isolation, and billing origin; those planned mechanisms are not being treated as missing implementation.

decisions/0052-user-owned-chatgpt-grants.md
```diff
@@ -0,0 +1,389 @@
...
+The broker must also authorize each credential-bearing upstream request
+against the durable active generation, including every HTTP request inside
+an already-open CONNECT tunnel. Do not authorize the entire tunnel from its
+first lookup. Define request admission as the grant-state check serialized
+with the disconnect transaction: requests admitted before that fence are
+in flight and may complete; no request admitted after it may use G. A stale
+cache or unavailable authorization store must fail closed. Do not hold the
+database lock through the upstream response or stream. A broker incapable
+of enforcing this per-request gate cannot serve managed grants; include a
+`managoat_broker` change, release, and pin update if needed.
```
BinaryBourbon, line 243: **[P2] Reject protocol upgrades on managed-grant routes**

The per-HTTP-request gate does not cover new provider operations on an already authenticated WebSocket. A sandbox can complete an upgrade while G is active, then send new operations after disconnect on a broker node that missed cleanup notifications. Those frames never reach another HTTP admission check. Completing the upgrade handshake must not classify every future operation as an already-admitted request.

`CodexTransport` sets `supports_websockets: false`, but this configures Codex, not a raw sandbox client's access through the broker. The protected destination policy currently specifies the upstream host and routes without requiring rejection of protocol upgrades on an allowed route.

The pinned `managoat_broker` 0.11.0 injects the handshake and switches to a byte pipe (`lib/managoat/broker/proxy.ex:757–761,852–866`); its response observer treats the HTTP request as finished at `101` (`lib/managoat/broker/response.ex:279–282`). ADR 0047's G2 measurement (line 421) also records an authenticated WebSocket upgrade in the existing ChatGPT path. This scenario was identified from source and recorded evidence, not newly reproduced against OpenAI.

For the initial HTTP-only integration, please require the protected broker policy to reject protocol upgrades **before injecting a managed bearer**, including raw sandbox-client attempts. If upgraded sessions are required, specify authorization of subsequent operations independently of cleanup notifications. Add an acceptance case for the upgrade attempt and post-disconnect traffic on a node that missed invalidation.

BinaryBourbon (changes requested):
Fresh adversarial review by a separate Codex agent of `8f7beef30e561a02f6ec553b3c5babf799aef8e3`, posted as BinaryBourbon at the user's request.

The previous broker-issuance fencing and custom-binding token-export findings are addressed at the ADR level. One remaining P2 architecture gap needs clarification: the new per-HTTP-request admission rule does not cover subsequent provider operations over a previously authenticated WebSocket. The initial HTTP-only integration should require the broker to reject protocol upgrades before injecting a managed bearer, rather than relying on Codex's client configuration.

The inline finding includes the concrete scenario, pinned broker-source evidence, and requested acceptance case. I found no other material issues in the proposed owner scoping, encryption choice, connection-attempt fencing, refresh coordination, explicit source selection, or accounting requirements.

This was a documentation and source review. No new provider exchange or live reproduction against OpenAI was performed, and the proposed safeguards are not implemented by this documentation-only PR.

BinaryBourbon (commented):
Fresh adversarial review by a separate Codex agent of `0f2fc7072ff686a3d2b912a5d0616d254d1bb83e`, posted as BinaryBourbon at the user's request.

**No new material actionable findings.**

Reviewed the complete ADR 0052, its index entry, relevant lifecycle/broker/credential-selection code, and ADRs 0047 and 0019. The previous findings are addressed in the proposed architecture:

- **Broker issuance and revocation:** lines 216–243 require durable grant-generation checks for session issuance and every credential-bearing request, including requests inside existing CONNECT tunnels.
- **Managed-token export:** lines 277–313 isolate managed credentials from custom-template inputs and enforce protected provider destinations.
- **Protocol upgrades:** lines 245–262 require rejecting upgrades before bearer injection, checking effective outgoing headers, failing closed on unexpected `101` responses, and draining legacy upgraded connections.

No further architectural contradiction or concrete security failure emerged from checking ownership/encryption, connection-attempt races, refresh coordination, source pinning, account deletion, and billing. ADR validation also passes with six pre-existing warnings and none in ADR 0052.

This is a review of the proposed design, not an implementation verification or GitHub approval. It does not establish that the unbuilt implementation satisfies these requirements or that the pinned Codex/broker integration works; the ADR gates rollout on those checks. No application tests or live provider exchanges were run.

BinaryBourbon (approved):
Independent adversarial review by a fresh Codex agent of `0f2fc7072ff686a3d2b912a5d0616d254d1bb83e`. Posted as BinaryBourbon with the user's explicit authorization to approve if this review found no material issues.

**No material findings. Approving this documentation-only PR.**

Reviewed the complete ADR and index change against the existing credential-selection, Codex authentication, refresh, broker issuance/template, and account-deletion paths.

The earlier findings are addressed: grant fencing covers session creation, delayed updates, and each request inside existing tunnels; managed bearers are excluded from tenant templates and restricted to protected destinations; protocol upgrades are rejected before injection, unexpected `101` responses fail closed, and legacy connections require draining.

Ownership, encryption, source pinning, refresh races, paid fallback, and rollout requirements are consistent. This approval accepts the proposed architecture. It does not certify the unbuilt implementation or runtime compatibility; those remain subject to the ADR's implementation and rollout checks.

PR #2011, opened 2026-09-12
fix: fence ChatGPT lifecycle writes by grant generation

jhgaylor:
A refresh response can arrive after an admin reconnects or disconnects the ChatGPT account. The old terminal-error path could revoke a replacement, and a refresh that did not rotate its refresh token could be overwritten by an older response. Fence refresh success, revocation, and expiry writes with grant ID, generation, version, and active state. Reconnect changes generation; normal refresh advances only the write version. A stale caller cannot receive a replacement account's bearer.

Reconnect and disconnect serialize their existing-row mutations in short transactions. The additive migration backfills generation/version for the platform account and refuses unexpected tenant rows so their encryption cannot be silently reinterpreted.

First foundation slice for ADR 0052 (#2010). Deployment-wide refresh coordination, broker issuance/request revocation, and user linking remain follow-up work; this PR does not claim those guarantees.

Validation:
- 53 focused lifecycle, platform-account, admin LiveView, and Codex sandbox tests passed.
- Migration smoke check passed: platform ciphertext preservation, defaults, rollback, unexpected tenant-row rejection, and successful migration after correction.
- Full `mix precommit` passed: compile, formatting, Credo, Dialyzer, Sobelow, release assembly, 5,457 core tests plus 6 doctests, 144 Buzz tests, and 33 Support tests; zero failures (2 core tests skipped, 7 excluded).

## Behaviour change worth naming

`access_token/0` can now answer `{:error, :stale_grant}`: a reconnect that lands while a refresh is in flight no longer serves the replacement account's bearer, because the caller may already have pinned the old account's metadata. Both callers degrade safely -- `InferenceCredentials.platform_credential/4` maps it to `:none` and falls through to `PLATFORM_OPENAI_API_KEY`, and `Egress.refresh_platform_chatgpt/2` keeps the token it holds -- so nothing crashes, but that conversation runs on the platform key rather than the subscription. That is ADR 0047 decision 6's "at the next conversation, not within a turn", and it is deliberate: the `Refresher` moduledoc's old promise that "a node that loses the race serves what the winner wrote" no longer holds, and the moduledoc now says so.

## Review follow-ups (316bdebc)

- Removed `Account.revoke_changeset/2`, `expire_changeset/1` and `refresh_changeset/2`. The first two lost their callers when the terminal writes moved onto fenced `update_all`; the third had none before. Each writes on the primary key alone, so leaving them left the most discoverable writers in the module unfenced.
- `mark_expired/1` now reports `:ok | :stale` like `mark_revoked/2`, and `expire_or_serve/1` routes a lost fence through `current_result/1` rather than reporting `:expired` for a grant a reconnect just made active.
- Dropped the `rescue Ecto.StaleEntryError` in `store/3`: `locked_platform_row/0` holds `FOR UPDATE` for the whole transaction, so the optimistic lock cannot lose and the rescue could not fire.
- `current_result/1` no longer raises on a status outside the three literals; the column has no CHECK constraint.
- New test: expiry advances `lock_version` without replacing the generation.

**Correction to the line above, from the second review.** I originally wrote that this test "fails when `mark_expired/1` writes on the primary key alone". That is wrong, and I have re-run it to confirm: my revert-check dropped the fence *and* `inc: [lock_version: 1]` together, so the test failed on the missing version bump. A primary-key-only `update_all` that still increments leaves 9 tests, 0 failures. The test proves the expiry write is versioned, not that it is fenced.

What is actually covered: reverting `current_query/1` itself fails 4 of the 9 lifecycle tests, so the fence predicate is well guarded. What is not covered is `mark_expired/1`'s *use* of it, and the new `:stale -> current_result(row)` branch, which no test reaches -- there is no I/O boundary in that window to hook, unlike the refresh path. Recorded rather than papered over.

### DIFF
--- a/apps/fountain/lib/fountain/platform_chatgpt.ex
+++ b/apps/fountain/lib/fountain/platform_chatgpt.ex
@@ -117,8 +117,10 @@ defmodule Fountain.PlatformChatGPT do
   # the row goes `expired`.
   defp expire_or_serve(row) do
     if lapsed?(row) do
-      _ = mark_expired(row)
-      {:error, :expired}
+      case mark_expired(row) do
+        :ok -> {:error, :expired}
+        :stale -> current_result(row)
+      end
     else
       decrypt(row.access_token_ciphertext)
     end
@@ -288,6 +290,11 @@ defmodule Fountain.PlatformChatGPT do
     end
   end
 
+  # `locked_platform_row/0` holds `FOR UPDATE` for the whole transaction, so
+  # the optimistic lock in `connect_changeset/2` cannot lose a race; it is
+  # there to advance `lock_version` on a reconnect, not to detect one. Two
+  # concurrent connects onto an empty table meet
+  # `platform_chatgpt_account_platform_row` instead and one is refused.
   defp store(attrs, method, actor_user_id) do
     result =
       Repo.transaction(fn ->
@@ -318,8 +325,6 @@ defmodule Fountain.PlatformChatGPT do
       {:error, _changeset} ->
         {:error, :invalid_grant}
     end
-  rescue
-    Ecto.StaleEntryError -> {:error, :stale_grant}
   end
 
   # ── refresh ──────────────────────────────────────────────────────────────
@@ -440,6 +445,7 @@ defmodule Fountain.PlatformChatGPT do
           "active" -> decrypt(current.access_token_ciphertext)
           "revoked" -> {:error, :revoked}
           "expired" -> {:error, :expired}
+          other -> {:error, {:unknown_status, other}}
         end
 
       %Account{} ->
@@ -506,6 +512,12 @@ defmodule Fountain.PlatformChatGPT do
     )
   end
 
+  # `:stale` is narrow here in a way it is not on the refresh path: nothing
+  # blocks between the read in `access_token/0` and this write, so only a
+  # reconnect landing inside those microseconds loses the fence. It is still
+  # routed through `current_result/1` rather than assumed away, because
+  # reporting `:expired` for a grant that is now active is the same class of
+  # wrong answer the fence exists to prevent.
   defp mark_expired(row) do
     {count, _} =
       current_query(row)
@@ -517,6 +529,10 @@ defmodule Fountain.PlatformChatGPT do
         event_type: "admin.platform_chatgpt.expired",
         metadata: %{"actor" => @system_actor, "kind" => row.kind, "account_id" => row.account_id}
       })
+
+      :ok
+    else
+      :stale
     end
   end
 
--- a/apps/fountain/lib/fountain/platform_chatgpt.ex
+++ b/apps/fountain/lib/fountain/platform_chatgpt.ex
@@ -265,13 +265,19 @@ defmodule Fountain.PlatformChatGPT do
   """
   @spec disconnect(keyword()) :: :ok
   def disconnect(opts \\ []) do
-    case platform_row() do
+    {:ok, deleted} =
+      Repo.transaction(fn ->
+        case locked_platform_row() do
+          nil -> nil
+          row -> Repo.delete!(row)
+        end
+      end)
+
+    case deleted do
       nil ->
         :ok
 
       %Account{} = row ->
-        Repo.delete!(row)
-
         Audit.record_admin(%{
           actor_user_id: Keyword.get(opts, :actor_user_id),
           event_type: "admin.platform_chatgpt.disconnected",
@@ -284,9 +290,14 @@ defmodule Fountain.PlatformChatGPT do
 
   defp store(attrs, method, actor_user_id) do
     result =
-      (platform_row() || %Account{})
-      |> Account.connect_changeset(attrs)
-      |> Repo.insert_or_update()
+      Repo.transaction(fn ->
+        case (locked_platform_row() || %Account{})
+             |> Account.connect_changeset(attrs)
+             |> Repo.insert_or_update() do
+          {:ok, account} -> account
+          {:error, changeset} -> Repo.rollback(changeset)
+        end
+      end)
 
     case result do
       {:ok, account} ->
@@ -307,6 +318,8 @@ defmodule Fountain.PlatformChatGPT do
       {:error, _changeset} ->
         {:error, :invalid_grant}
     end
+  rescue
+    Ecto.StaleEntryError -> {:error, :stale_grant}
   end
 
   # ── refresh ──────────────────────────────────────────────────────────────
@@ -341,9 +354,9 @@ defmodule Fountain.PlatformChatGPT do
   # by nothing else: one at a time on this node, no database lock and no
   # connection held across the HTTP round-trip. The row is re-read here
   # because a queued caller may find the refresh already done, and the
-  # write is a compare-and-swap on the refresh token this read started
-  # from, so a node that lost the race to another serves the winner's token
-  # rather than overwriting it.
+  # write checks the generation, version and active state this read started
+  # from. A losing refresher may serve a winner from the same generation,
+  # but cannot overwrite or serve a replacement account.
   def refresh_serialized(mode) do
     case platform_row() do
       nil ->
@@ -376,8 +389,10 @@ defmodule Fountain.PlatformChatGPT do
           swap_in(current, refresh_attrs(fresh), access)
 
         {:error, {:terminal, code}} ->
-          _ = mark_revoked(current, code)
-          {:error, :revoked}
+          case mark_revoked(current, code) do
+            :ok -> {:error, :revoked}
+            :stale -> current_result(current)
+          end
 
         {:error, reason} ->
           Logger.warning(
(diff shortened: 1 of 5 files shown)


BinaryBourbon (changes requested):
The fence itself is right, and the tests earn their place: I reverted `apps/fountain/lib/fountain/platform_chatgpt.ex` to `origin/main` while keeping the new columns, and **7 of the 8 new tests fail**. That is the check I most wanted to see on a concurrency PR. The design also matches ADR 0052's "fence success, terminal failure, reconnect, and disconnect writes by grant ID, generation, version, and active state" line for line, and both audit calls stay outside their transactions.

One thing to fix before merge, plus three smaller ones.

---

### 1. The fence has an unguarded back door: `revoke_changeset/2` and `expire_changeset/1` are now orphaned

`mark_revoked/2` and `mark_expired/1` were the only callers, and both now go through `current_query/1 |> Repo.update_all(...)`. Grep finds no remaining caller anywhere in `apps/`, `ee/` or the test tree:

```
$ grep -rn "revoke_changeset\|expire_changeset" apps ee | grep -i chatgpt
apps/fountain/lib/fountain/platform_chatgpt/account.ex:98:  def revoke_changeset(account, reason) when is_binary(reason) do
apps/fountain/lib/fountain/platform_chatgpt/account.ex:105:  def expire_changeset(account) do
```

This is worse than ordinary dead code. They are public, they are documented as the way to do exactly what this PR just fenced ("The auth server refused the refresh token; the reason is its error code"), and they are now the *most discoverable* way in the module to change `status` — while bypassing the generation/version check entirely. The next person adding a status transition reaches for the changeset, gets no fence, and every test in the new file still passes because none of them go through it.

Delete both, or leave a comment saying they must not be used for lifecycle writes and why.

### 2. `mark_expired/1` swallows a lost fence; `mark_revoked/2` does not

`mark_revoked/2` returns `:stale` and `do_refresh/1` routes that to `current_result/1`. `mark_expired/1` returns the bare `if` and `expire_or_serve/1` discards it:

```elixir
if lapsed?(row) do
  _ = mark_expired(row)
  {:error, :expired}
```

So a lapsed workspace token whose `update_all` matches zero rows still reports `{:error, :expired}` — for a grant that a concurrent reconnect just made active. The window is small (no HTTP round-trip inside it, unlike the refresh path) and it self-heals on the next call, but the asymmetry with `mark_revoked/2` is unexplained and the two read as if they had the same contract.

### 3. The `rescue Ecto.StaleEntryError` in `store/3` looks unreachable

`locked_platform_row/0` takes `FOR UPDATE` inside the same transaction that then runs `connect_changeset/2`'s `optimistic_lock(:lock_version)`. Holding the row lock, the optimistic `WHERE lock_version = ?` cannot lose, and with no row there is no optimistic lock to lose either — the concurrent-insert case lands on `platform_chatgpt_account_platform_row` and comes back `{:error, :invalid_grant}`. No test reaches `{:error, :stale_grant}` from a connect.

If it is reachable, it wants a test. If it is not, drop it — an unexercised rescue reads as a guarantee.

### 4. Nit: `current_result/1`'s inner `case` has no fallback

```elixir
case current.status do
  "active" -> decrypt(current.access_token_ciphertext)
  "revoked" -> {:error, :revoked}
  "expired" -> {:error, :expired}
end
```

`platform_chatgpt_account.status` has no CHECK constraint (see `20260908150000_create_platform_chatgpt_account.exs`), so anything outside the three literals raises `CaseClauseError` mid-refresh. Not a regression — the `String.to_existing_atom/1` it replaced had the same shape — but a `_ ->` clause costs a line.

---

### Two notes, no change requested

**Merge order.** The `Account` moduledoc and the `Refresher` moduledoc both cite ADR 0052, which exists only on #2010. This PR's base is `main`, so if it lands first the tree points at a decision record that is not in `decisions/`. #2010 is already approved — merge it first, or rebase onto it.

**The new `{:error, :stale_grant}` return is a deliberate behaviour change.** Both callers degrade safely (`InferenceCredentials.platform_credential/4` maps it to `:none` and falls through to `PLATFORM_OPENAI_API_KEY`; `Egress.refresh_platform_chatgpt/2` keeps the token it holds), so nothing crashes. But on `main` this window served the winner's token, and now a reconnect during an in-flight refresh sends that conversation to the platform API key for its whole life. That is the correct call given the account-identity pinning you are protecting, and it matches ADR 0047 decision 6's "at the next conversation, not within a turn" — worth saying so in the PR body, since the refresher moduledoc's old promise ("a node that loses the race serves what the winner wrote") is what a reader will remember.

BinaryBourbon (approved):
Second pass, independent of the review that requested changes. All three requested items are genuinely addressed. Approving, with four notes — one of them a coverage gap you should know about before you rely on the follow-up commit's verification line.

## The three requested changes

**1. Orphaned changesets — done.** `revoke_changeset/2`, `expire_changeset/1` and `refresh_changeset/2` are gone and nothing calls them. `grep -rn "revoke_changeset\|expire_changeset\|refresh_changeset" apps ee cli` returns only `Fountain.Connections.Connection`'s unrelated trio (`connections.ex:423,458,469,637`). The tree compiles and 145 tests across the lifecycle, platform-account, inference-credentials, admin-LiveView and audit-guardrail files pass.

**2. `mark_expired/1`'s `:ok | :stale` — correct and reachable.** `current_query/1` is the same predicate the other two writes use, and `:stale` is not decoration: on a multi-replica deployment every conversation that touches a lapsed workspace token calls `expire_or_serve/1`, so all but one lose the fence. Two consequences, both right — the losers now answer `{:error, :expired}` by re-reading rather than by assumption, and only the winner emits `admin.platform_chatgpt.expired`, which is CLAUDE.md's "only record what happened" (on `main` every replica wrote one).

**3. Dropping `rescue Ecto.StaleEntryError` — safe; I checked the mechanism rather than the argument.** Two psql sessions against the test database: session A `BEGIN; SELECT … FOR UPDATE;` then, after a sleep, `UPDATE … WHERE lock_version = 1` and `COMMIT`; session B meanwhile issues the fenced autocommit `UPDATE … WHERE gen = … AND lock_version = 1 AND status = 'active'`. B blocked for the duration of A's lock and then matched **0 rows** — Postgres re-evaluated B's predicate against A's committed row under READ COMMITTED. So while `store/3` holds the row lock no other writer can advance `lock_version`, and `optimistic_lock` cannot lose. That argument only holds if every writer takes or blocks on that lock: `grep -rn "platform_chatgpt_account" apps ee` plus a sweep of `PlatformChatGPT.` call sites confirms the writers are exactly `store/3`, `disconnect/1` (both `FOR UPDATE` in a transaction) and the three fenced `update_all`s (plain autocommit statements, which block). `Device` goes through `connect_from_tokens/3`. The concurrent-insert case does land on `platform_chatgpt_account_platform_row` as the comment says — the index name in `20260908150000` matches the `unique_constraint/2` name, so it returns `{:error, changeset}` → `{:error, :invalid_grant}` rather than raising.

## Findings

### 1. The new expiry test does not guard the fence — only the versioning (medium, non-blocking)

Verified by reverting. Replacing `current_query(row)` with `from(a in Account, where: a.id == ^row.id)` inside `mark_expired/1` — literally "writes on the primary key alone", keeping `inc: [lock_version: 1]` — leaves **9 tests, 0 failures**. The new test asserts `generation` unchanged and `lock_version + 1`, and a PK-only versioned `update_all` satisfies both.

The claim in the commit message is true under the other reading: restoring the old `Account.expire_changeset()` + `Repo.update()` write fails it (`left: 1, right: 2`), because that path did not version at all. So the test proves the expiry write is now versioned, not that it is fenced. Worth saying precisely, since it is the evidence offered for the one item the previous review asked you to fix.

What mitigates it: the predicate lives in one function, and reverting `current_query/1` itself to PK-only fails **4 of 9** (`late refresh success…`, `late terminal refusal…`, and both `a stale … cannot undo a refresh…`). The only uncovered thing is `mark_expired/1`'s *use* of it, plus the new `:stale -> current_result(row)` branch in `expire_or_serve/1`, which no test reaches. I don't think that blocks — but don't carry "verified by reverting" forward as if it covered the fence.

### 2. `Account.statuses/0` is now dead (low)

`@statuses` lost its last real consumer when the three changesets went; `statuses/0` is the only thing left referencing it, and `grep -rn "PlatformChatGPT.Account" apps ee` finds no caller outside the module and its tests. Same argument you made for the changesets — a public accessor for a vocabulary nothing validates against.

### 3. The refresher moduledoc dropped the measurement that justifies the behaviour it still has (low)

The removed line carried a fact: *"measured 2026-09-08: a reused refresh token forks rather than revokes"*. `swap_in/3` still discards the loser's freshly rotated refresh token when the fence fails, so that measurement is exactly what makes the loser's abandoned chain harmless. The replacement prose says what the fence does but not why abandoning a rotated chain is safe. Keep the measured line — the next person to read this will ask.

### 4. ADR 0052 citation still dangles (informational)

`refresher.ex:18` cites ADR 0052; `decisions/` on this branch stops at `0051-record-sandbox-provider-identity.md`. Merge #2010 first or rebase onto it, as already noted.

## Independent pass — what I checked and found clean

- **New error shapes reach safe callers.** `{:error, :stale_grant}` and `{:error, {:unknown_status, other}}` only escape through `access_token/0`. Its two callers are `credential/1` (`_ -> :none`, `inference_credentials.ex:297` then falls to `PLATFORM_OPENAI_API_KEY`) and `Egress.refresh_platform_chatgpt/2` (`with … else _ ->`, `egress.ex:130`, keeps the token it holds). `keepalive/0`'s `{:error, reason}` is `inspect`ed by the worker. The admin LiveView's `chatgpt_error/1` has a `defp chatgpt_error(other), do: inspect(other)` catch-all, and `store/3` no longer producing `:stale_grant` leaves no dead clause there. No caller pattern-matches an exhaustive set. Reading-based.
- **Audit rules.** Both records are outside their transactions; `store/3` and `disconnect/1` record after `Repo.transaction` returns; the three `update_all` writes are not in transactions at all. Metadata carries account id, kind and the server's reason code — no token, no claim. `mark_revoked/2` moving its audit inside the `count == 1` branch is a fix, not a regression.
- **Tenant isolation.** `current_query/1` adds `is_nil(a.user_id)`, which the old CAS did not have; `platform_row/1` and `locked_platform_row/0` both scope the same way. No `_unsafe_` path introduced.
- **Migration.** Newest timestamp; `gen_random_uuid()` has precedent (`20260823100000_create_agent_versions.exs`); per-row default backfill gives each existing row a distinct generation; `null: false` + default is a metadata-only add on modern PG; the `DO $$ … RAISE EXCEPTION` guard runs inside the migration transaction so a tenant row aborts it cleanly; `lock_version > 0` can never be violated by `inc: [lock_version: 1]`; `down` is complete. `:bigint` under a schema `:integer` is fine.
- **Gates.** `mix credo --strict` from the umbrella root: no issues. `mix format --check-formatted` clean on all five changed files. `mix test test/fountain/platform_chatgpt_lifecycle_test.exs test/fountain/platform_chatgpt_test.exs test/fountain/audit_guardrail_test.exs test/fountain/inference_credentials_test.exs test/fountain_web/live/admin_inference_live_test.exs` → 145 tests, 0 failures.

Everything above labelled "verified" I ran; items 3 and 4 under Findings and the caller sweep are reading-based.

PR #2012, opened 2026-09-12
refactor: introduce scoped ChatGPT account storage

jhgaylor:
Extract ChatGPT account storage and the existing platform lifecycle into `Fountain.ChatGPTAccounts`, retaining `PlatformChatGPT` as a compatibility wrapper with explicit platform delegations. Add tenant-scoped metadata reads and generation-pinned credential reads; bearer and account identity come from the same row version.

Owned tokens use the owner's DEK with owner- and token-field-specific authenticated data. Platform tokens keep their deployed encryption format. A redacted server credential struct separates bearer material from its source metadata and has no generic JSON encoder. Metadata reads do not decrypt tokens or refresh grants, and unknown provider failure reasons are sanitized.

The partition runner also discards its generated child protocol table before tests. CI restored a stale table that omitted the new Inspect implementation even after the umbrella compiled it; an app-level run then printed the bearer. The existing redaction assertion caught this, and removing the stale artifact makes Mix rebuild the table from the current implementations.

#2011 is merged. This is the next ADR 0052 (#2010) foundation slice, now based directly on main. User linking, user writes, coordinated user refresh, and conversation/broker integration remain unbuilt. Reads that need renewal return `refresh_required`; they never invoke the platform refresher or paid fallback.

Validation:
- 12 tenant context tests passed, covering ownership, absent owners, wrong-DEK rejection, field/owner ciphertext swaps, platform-format isolation, metadata-only reads, redaction, generation mismatch, and unusable/near-expiry grants.
- Existing platform lifecycle, admin UI, and sandbox tests passed in the focused run. One new test fixture initially reused a stale Ecto struct; it was corrected and all 12 context tests rerun successfully.
- Full `mix precommit` passed before the runner-only fix: compile, formatting, Credo, Dialyzer, Sobelow, release assembly, 5,469 core tests plus 6 doctests, 144 Buzz tests, and 33 Support tests; zero failures (2 core tests skipped, 7 excluded).

- Final runner fix: reproduced [the CI failure](https://github.com/managoat/fountain/actions/runs/34666175266/job/103478700227) locally, restored the same stale protocol table, then ran the corrected partition runner: 904 tests, zero failures, coverage exported. The 12 focused context tests also pass with coverage.
- Release-artifact inspection selects `Inspect.Fountain.ChatGPTAccounts.Grant` and redacts a synthetic bearer. Shell syntax and all 23 CI policy tests pass.

## Review follow-ups (9152adcd)

- `_unsafe_platform_*` renamed to `platform_*`. The functions pin `is_nil(a.user_id)` and cannot reach a tenant's row, so the prefix was claiming a hazard that is not there; two of them read `Application.get_env` and touch no row. `Fountain.PlatformInference` is the deployment-scoped context next door and uses plain names.
- `safe_reason/1` allowlists against a new `OAuth.terminal_codes/0` rather than a second copy of the list. The copy it replaced was already missing `invalid_refresh_token_ciphertext_integrity`, so that revocation reached a tenant as `"provider_error"`. A new test walks every code `OAuth` names.
- `status_for_user/1` now selects `:grant_id` and `:generation`, the pin `credential_for_user/3` takes -- without them nothing could produce those arguments, so the two halves of the API could not compose.
- The `Fountain.PlatformChatGPT` moduledoc's function index and its warning that the refresh margin must exceed the longest turn are restored into `Fountain.ChatGPTAccounts`, which also states which prefix scopes to what. `defdelegate` carries no `@doc`.
- Rebased onto #2011's review fixes; the fence changes there are ported into `Fountain.ChatGPTAccounts` as part of the move.

Still deliberately unbuilt, as before: nothing writes a user grant, `status_for_user/1` and `credential_for_user/3` have no production caller, and reads that need renewal return `:refresh_required` rather than invoking the platform refresher.

## Second-review follow-up (bd018066)

- **Restored the decrypt-failure warning the extraction dropped.** `decrypt/1` in `Fountain.PlatformChatGPT` logged "the stored token does not decrypt under MASTER_SECRETS_KEY; reconnect at /admin/inference"; moving the reads into `Cipher.decrypt_token/2` lost it (three warning sites before the move, two after, no replacement in the tree). It is the only signal that a key rotation killed the grant -- `:undecryptable` becomes `:none`, `InferenceCredentials` reads that as "no grant" and falls through to the platform OpenAI key, and `platform_status/0` never decrypts, so the admin page keeps saying "active". Restored on both ownership paths with different remediation, since a tenant cannot act on an admin URL. The platform message is not quite the one it replaces: it now names the field, so it reads "the stored access_token does not decrypt under MASTER_SECRETS_KEY; reconnect at /admin/inference" -- an access-token failure and a refresh-token failure are different problems and the log should say which. Neither message carries ciphertext or plaintext. Verified by reverting: stripping both warnings fails the new test on the tenant assertion, and stripping only the platform one fails it on the platform assertion, so neither branch can go quiet alone.
- **Guarded the `rm -rf` in `scripts/test-partition.sh`.** It was the only line in that file that could not fail: a typo in the path would remove nothing, exit 0, and silently restore the stale-protocol-table bug the line exists to prevent. Now asserts the app build directory is where it expects whenever a build exists; a cold checkout has nothing stale. Checked both ways -- the real path proceeds, `fountian` exits 1.

Left as noted, not changed: the untouched platform ciphertext format still shares one constant AAD with `PlatformInference`'s keys, so platform blobs remain swappable with each other. Pre-existing and outside this PR's scope; worth its own issue.

## Landing validation

Rebased only this PR’s four commits onto main `86f8f2a2ddd8836296e91ec028ef4547ef5efdc4` after #2011 merged. `git range-diff` reports all four patches unchanged, and all eight PR files are byte-identical to approved head `bd018066`. New head: `57c41591`.

Full local `mix precommit` passed on Elixir 1.19.2 / OTP 28.3 with a dedicated test database: compilation, formatting, strict Credo, Dialyzer (zero errors), Sobelow, production release assembly, 6 doctests + 5,839 core tests, 144 Buzz tests, and 33 Support tests; zero failures (2 skipped, 7 excluded core cases). All 57 CI policy tests and shell syntax checks also passed.

Fresh PR and merge-queue CI must pass against main before landing.


BinaryBourbon (changes requested):
The crypto separation is the good part and it is done carefully: ownership decides the cipher, the AAD binds both the owner and the token field, and the test file proves all four of the swaps that matter (cross-owner, cross-field, wrong DEK, platform ciphertext in an owned row) with no silent fallback to `decrypt_platform/1`. `credential_for_user/3` returning `:not_connected` rather than `:forbidden` for another user's grant id is the right choice too — no existence oracle.

Two things to change, one to decide, and some notes.

---

### 1. `_unsafe_platform_*` inverts the repo's most important safety marker

CLAUDE.md is unambiguous: "Functions prefixed `_unsafe_` bypass tenant scoping... **Never call one as the first fetch in a user-facing request** — ownership must already be established on that request." Every one of the ~40 existing `_unsafe_*` functions in `apps/fountain/lib` names a row operation over tenant-owned data where that warning is load-bearing.

These do the opposite. `platform_row/0` hardcodes `where: is_nil(a.user_id)`; `current_query/1` repeats it. They cannot return a tenant's row under any argument — they are the most narrowly scoped queries in the file. Marking them `_unsafe_` tells a reader to go and establish ownership before calling, and there is no ownership to establish.

Two of them make the point on their own:

```elixir
@spec _unsafe_platform_refresh_margin_seconds() :: non_neg_integer()
def _unsafe_platform_refresh_margin_seconds do
  case Application.get_env(:fountain, :platform_chatgpt_refresh_margin_seconds) do
```

That reads `Application.get_env`. It touches no row, no repo and no tenant. Same for `_unsafe_platform_keepalive_days/0`.

The precedent is right next door: `Fountain.PlatformInference` is the other deployment-scoped context over an admin-owned table, and it uses `key_for/1`, `put_key/3`, `clear_key/2`, `status/0` — no prefix. I would use plain `platform_*` here, which disambiguates against the user-scoped functions without claiming a danger that is not there, and keep `_unsafe_` in reserve for the user-scoped writers when they land in a later slice. A prefix that appears on safe functions is a prefix people stop reading.

### 2. `safe_reason/1` restates `OAuth`'s terminal list and is already out of sync

```elixir
defp safe_reason(reason)
     when reason in ~w(invalid_grant refresh_token_reused refresh_token_expired refresh_token_invalidated),
     do: reason
```

`Fountain.PlatformChatGPT.OAuth` names five:

```
@terminal ~w(refresh_token_expired refresh_token_reused refresh_token_invalidated invalid_grant invalid_refresh_token_ciphertext_integrity)
```

`invalid_refresh_token_ciphertext_integrity` is missing, so that revocation — a real code the auth server sends, and the one a user most needs to see named — reaches them as `"provider_error"`. The list drifted on the commit that introduced it, which is the argument for not having two of them: expose `OAuth.terminal_codes/0` and allowlist against that.

### 3. `status_for_user/1` cannot produce the arguments `credential_for_user/3` needs

The `select:` map carries neither `:id` nor `:generation`, and it is the only user-facing read in the module. So there is no path, even in principle, by which a caller obtains the grant id and generation that the pin requires — the two halves of the API cannot be joined. Adding both fields now is a two-line change and makes the shape testable end to end rather than only from the inside.

---

### Notes, no change requested

**No production caller.** `status_for_user/1`, `credential_for_user/3`, `Cipher.encrypt_user_tokens/2` and the whole `Grant` struct are reachable only from `chatgpt_accounts_test.exs`. The PR body says so plainly and it matches how the #1634 / #1637 stacks were cut, so this is fine — recording it so the choice is on the record rather than discovered later.

**The `PlatformChatGPT` moduledoc went with the body.** The "What is here" index and the operational warning — "The refresh margin must exceed the longest turn the deployment expects: a turn that outlives its access token fails at the proxy, because codex cannot refresh in this mode" — were deleted, and `defdelegate` carries no `@doc`, so `h Fountain.PlatformChatGPT.access_token` is now empty. Nothing is lost outright (the margin warning survives in `docs/configuration.md:264` and ADR 0047), but the module a reader lands on first no longer says what it is for. `Fountain.ChatGPTAccounts` is the natural home for that paragraph.

**The CI fix is unrelated to the refactor.** I reproduced the mechanism locally — deleting `_build/test/lib/fountain/consolidated` does make the child `mix test` regenerate it (20 beams back, tests green), so `scripts/test-partition.sh` works as described and costs almost nothing. My only note is that it would land on its own in a minute and unblock any other branch hitting the same stale table, rather than waiting behind a 900-line refactor. Worth keeping in mind that `scripts/test-libraries.sh` runs the Buzz and Support suites from their own directories and has the same exposure.

**Merge order.** Same as on #2011: the moduledocs cite ADR 0052, which exists only on #2010. That one is approved; it should go in first.

BinaryBourbon (changes requested):
Second pass, independent of the earlier review and of the commit that answers it. All three requested changes are genuinely made, and I verified two of them by reverting. The crypto and the redaction hold up under the swaps I could construct. One thing went missing in the move that nothing on this PR mentions, and it is the reason I am not approving yet.

---

### 1. The `MASTER_SECRETS_KEY` warning was dropped in the extraction (blocking, two lines)

`Fountain.PlatformChatGPT.decrypt/1` on the base branch logged when a stored token would not decrypt:

```
origin/codex/adr52-account-foundation:apps/fountain/lib/fountain/platform_chatgpt.ex:598
  "platform chatgpt: the stored token does not decrypt under MASTER_SECRETS_KEY; "
  <> "reconnect at /admin/inference"
```

`Cipher.decrypt_token/2` replaced it and returns `{:error, :undecryptable}` silently. `grep -rn "does not decrypt" apps ee` on this head finds the `PlatformInference` copy and nothing for ChatGPT. The base had three `Logger.warning` call sites in that module; this head has two, and the missing one is that.

What the operator sees after a key rotation now: `platform_credential/1` maps the error to `:none`, `inference_credentials.ex:297-299` takes `:none` as "no grant" and falls through to `PlatformInference.key_for("openai")`, and `platform_status/0` still reports `status: "active"` because it never decrypts. Silent fallback, admin page says connected, no log line anywhere.

The sibling context is the argument: `platform_inference.ex:166-177` keeps the identical warning for exactly this case, and `admin_live/inference.ex:478` has UI copy for `:undecryptable`. No test ever asserted the ChatGPT one, which is why it could leave without anything going red — so restoring it wants a test alongside, or it will go again. If it was dropped deliberately because `Cipher` is shared between the two owners, log it in the platform caller instead and say so in the commit; what should not stand is a behaviour change the PR describes as a compatibility-preserving extraction.

---

### 2. The three requested changes — all addressed

**The rename.** `grep -rn "_unsafe_platform"` over the whole tree (excluding `.git`, `_build`, `deps`) returns nothing, so no caller was missed, and every production caller still goes through `Fountain.PlatformChatGPT`, whose public arity list is unchanged. I checked the claim the rename rests on rather than taking it: `platform_row/0` (`chatgpt_accounts.ex:638`), `locked_platform_row/0` (`:634`) and `current_query/1` (`:531`) each pin `is_nil(a.user_id)`, and every `update_all` on the platform path goes through `current_query/1`, so no `platform_*` function can read or write a tenant row under any argument. Against CLAUDE.md's definition — "bypass tenant scoping... ownership must already be established" — the prefix was claiming a hazard with no ownership to establish, and the rename is the correct direction. 51 tests across the three ChatGPT files pass.

**`safe_reason/1`.** Verified by reverting: I put the old four-entry `~w(...)` guard back and ran the file. `test/fountain/chatgpt_accounts_test.exs:46` fails with `left: "provider_error"`, `right: "invalid_refresh_token_ciphertext_integrity"`. Restored. The test walks `OAuth.terminal_codes/0` itself, so a sixth code added upstream is covered — the only way it goes quiet is if that list ever returns empty.

**`status_for_user/1` pin fields.** Verified by reverting: dropping `grant_id:` and `generation:` from the select fails `:12` with `** (KeyError) key :grant_id not found`. Restored. The two halves compose.

---

### 3. Crypto — I went looking for the four failure modes and did not find them

Key selection is `account.user_id` nil-ness with no fallback in either direction (`cipher.ex:35,37`); the wrong-DEK test stubs `decrypt_platform/1` to `flunk`, so a fallback would be caught rather than inferred. The AAD `"fountain.chatgpt_grant:#{user_id}:#{field}"` is unambiguous — `user_id` is a `:binary_id` UUID with no `:` in it and `field` comes from a two-atom guard, so no two distinct (owner, field) pairs can collide. Ownership cannot move: `user_id` is not cast by `connect_changeset/2`, which is the only changeset on the schema (`account.ex:64-85`, and `:179` proves it), and the migration backs the shape with both partial unique indexes. Error paths return bare atoms, and `credential_for_user/3` gives `:not_connected` for "no grant" and "another tenant's grant" alike, so there is no existence oracle; reaching `:undecryptable` requires already owning the row.

Note, not a change request: the **platform** format is deliberately untouched, so it still has no owner or field binding — `Crypto.encrypt_platform/1` uses one constant AAD that `PlatformInference`'s keys share, which means platform access and refresh ciphertext are interchangeable with each other and with a platform inference key blob. Pre-existing and explicitly out of scope. It is worth an issue now that the asymmetry sits in one file.

---

### 4. Redaction — robust on every path a caller can reach

Verified by running each of these against a `%Grant{access_token: "SUPERSECRET"}`: plain `inspect`, nested in `{:ok, [grant]}`, `limit: :infinity, printable_limit: :infinity`, and the `Exception.message/1` of both a `MatchError` and a `KeyError` carrying it. All print `#Fountain.ChatGPTAccounts.Grant<source: ..., ...>`. `Jason.encode` raises `Protocol.UndefinedError`, and there is no `String.Chars`, so interpolation raises rather than leaks.

Two ways out, both requiring something to bypass the protocol, neither reachable today because nothing holds a `Grant`:

- `inspect(grant, structs: false)` prints `access_token: "SUPERSECRET"` — opt-in, and no call site in the tree passes it.
- Erlang `~p`: `:io_lib.format(~c"~p", [grant])` prints `access_token => <<"SUPERSECRET">>`. Elixir's `Logger.Translator` uses `inspect` for OTP crash reports so ordinary crashes are fine, but `:sys.get_state` dumps and any non-Elixir formatter are not. Worth remembering when a `Grant` first goes into GenServer state.

---

### 5. `scripts/test-partition.sh`

The fix is real and cheap — I did not take it on the description. I deleted `_build/test/lib/fountain/consolidated`, ran the suite from `apps/fountain`, and the 20-beam table came back rebuilt (the regenerated `Elixir.Inspect.beam` differs from the copy I saved first), 13 tests green in 2.8s. It sits before `mix test`, so the summary-line, zero-count and coverdata guards below it are untouched.

One gap against this script's own standard, which is that no step may pass without proving something: `rm -rf` on a path that does not exist exits 0. A typo in that relative path would silently restore the stale-table bug and the partition would still go green — it is the only line in the file that cannot fail. `[ -d ../../_build/test/lib/fountain/consolidated ] && { echo "..."; exit 1; }` immediately after would close it. Separately, `scripts/test-libraries.sh` runs the Buzz and Support suites from their own directories and has the same exposure; untouched here, which is fine, but it is the same bug waiting.

---

### Gates I ran

`mix format --check-formatted` on all six changed Elixir files: clean. `MIX_ENV=dev mix credo --strict` from the umbrella root: 836 files, no issues. `shellcheck scripts/test-partition.sh`: clean. `chatgpt_accounts_test.exs` + `platform_chatgpt_test.exs` + `platform_chatgpt_lifecycle_test.exs`: 51 tests, 0 failures. `toolchain_lockstep` + `release_pin` + `sobelow_gate` + `audit_guardrail`: 83 tests, 0 failures. The branch is cleanly on top of `codex/adr52-account-foundation@316bdebc`, and a normalised diff of the moved module against the base's `platform_chatgpt.ex` shows finding 1 as the only behavioural difference outside the rename and the `Cipher` indirection.

Merge order unchanged: #2010, then #2011, then this.

BinaryBourbon (approved):
Re-reviewed `bd018066`. The blocking finding is fixed, and I checked the fix rather than the description. Approving.

**The warning reaches the real path.** I did not want this proven only through the direct `Cipher.decrypt_token/2` call the new test makes for the platform half, so I put a temporary probe on the production entry point and ran it: `Fountain.PlatformChatGPT.credential(refresh: false)` — the exact call `inference_credentials.ex:297` makes — emits

```
[warning] platform chatgpt: the stored access_token does not decrypt under MASTER_SECRETS_KEY; reconnect at /admin/inference
```

and `access_token/0` the same. So the operator gets the line at the moment the fallback to the platform OpenAI key actually happens, which was the whole gap. Probe removed.

**Both halves are independently held down.** Stripping both warnings fails the new test on the tenant assertion (`left: ""`). Restoring the tenant one and stripping only the platform one fails it on `platform_log =~ "does not decrypt under MASTER_SECRETS_KEY"` at `:157`. So neither branch can go quiet on its own — which matters more than usual here, given this is the second time this code has moved.

One correction to the description, not to the code: the platform message is not verbatim what it replaced. The original said "the stored token"; this says "the stored `#{field}`", so it now reads `the stored access_token` or `the stored refresh_token`. That is an improvement — a refresh-token failure and an access-token failure are different problems — but the PR body should not claim verbatim. Neither message carries ciphertext or plaintext; the tenant one carries the owner's user id and the field name, which is what an operator needs to route it.

**The `test-partition.sh` guard does what it says.** Exercised the branch directly with the real path and with `fountian`: the real one proceeds, the typo exits 1. The cold-checkout exemption is right — a typo there goes unnoticed, but a checkout with no `_build` has no stale table to remove, so the guard covers exactly the case that produces the bug. shellcheck clean.

**Gates re-run on this head:** the three ChatGPT suites, 52 tests, 0 failures; `mix format --check-formatted` on both changed files clean; `MIX_ENV=dev mix credo --strict` from the umbrella root, no issues.

---

Notes that still stand, neither of them blocking and both fine to leave as you have:

- The platform ciphertext format keeps its single constant AAD, shared with `PlatformInference`'s keys, so platform access and refresh blobs stay interchangeable with each other and with an inference key blob. Out of scope here and now called out in the PR body; worth an issue so the asymmetry against the new tenant format does not just sit there.
- `inspect(grant, structs: false)` and Erlang `~p` still print the bearer. I agree with leaving them: nothing holds a `Grant` yet, and no call site passes either. The thing to remember is that this stops being true the moment a `Grant` goes into GenServer state, where `:sys.get_state` dumps and non-Elixir formatters bypass `Inspect`.

Merge order unchanged: #2010, then #2011, then this.

PR #2013, opened 2026-09-12
fix: coordinate ChatGPT refresh across nodes

jhgaylor:
Two servers can currently exchange the same platform ChatGPT refresh token at once. Generation/version checks protect the stored row, but cannot prevent duplicate upstream exchanges. This change serializes those exchanges with a per-grant PostgreSQL try-lock and serves the committed winner to concurrent refreshers, including forced keepalives.

#2011 and #2012 are merged. This PR is based directly on main and implements the next part of ADR #2010: coordinated platform ChatGPT refresh.

- Use a transaction-scoped advisory lock: the holder retains one checkout; contenders release theirs before bounded, jittered retries. Commit, rollback, worker death and connection loss release the lock without leaving a session lock in the pool. The existing local platform queue limits holders to one per node.
- Re-read generation, version, status and expiry after acquisition. Preserve conditional writes; disconnect and reconnect acquire no refresh lock and can commit while the provider request is in flight. Return rotated credentials only after commit and record terminal revocation's best-effort audit afterward.
- Bound each transaction to 20 seconds, the contention retry window to 5 seconds, and the HTTP exchange with connection/pool/complete-response timeouts. Pin this small exchange to HTTP/1 because the installed Finch version's complete-response timeout applies to that protocol.

Validation:

- 12 new regressions use committed rows and a separate two-connection PostgreSQL pool. They cover concurrent normal/forced refresh, pool capacity with six contenders, unrelated grants, reconnect/disconnect during success/terminal responses, worker death, rollback, transaction expiry, audit ordering/failure, and an actual TCP provider that indefinitely drips unfinished response chunks.
- Ran the two concurrent-refresher regressions against the previous context: both failed with a second upstream request using the original refresh token. Both pass with this change. The local slow-response provider times out at approximately 12 seconds and leaves the grant active and unchanged.
- Full `mix precommit` passed before rebasing onto the reviewed base: compile, formatting, Credo, Dialyzer, Sobelow, production release assembly, 5,481 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests; zero failures, 2 skipped and 7 excluded core cases.
- After rebasing: all 74 affected platform, tenant, lifecycle, admin and new concurrency tests passed on `328ab11e`. The precommit compile, format, Credo, Dialyzer, Sobelow and production release gates also passed on this revision, with the same 74 affected tests passed again by the precommit alias.

The initial concurrency run failed during fixture teardown because ExUnit stopped a per-test pool before cleanup. Moving the pool to module supervision fixed teardown; the behavior assertions had passed. No test was weakened or removed.

This is platform adoption of the lock primitive. User refresh, local queues keyed by user grant, batched user keepalive, and broker request authorization remain follow-up work; user linking remains unavailable. All refresh owners must run the new locking code for deployment-wide exclusion. An upstream rotation followed by a crash before database commit still requires reconnect; this does not provide an atomic exchange with the provider.

## Review follow-ups (`6f8715d4`), rebased onto #2012's `bd018066`

**The HTTP ceiling was 27s against a 20s transaction budget.** `:request_timeout` does not bound a call on its own -- finch 0.23.0 checks it *between* receives (`http1/conn.ex:298`) and the receive that follows may block a full `:receive_timeout` (`:313`), so the two add rather than the larger winning: 12 + 12 + 2 + 1. Reproduced on this branch's own drip harness, changed only to go silent after 11s: 22.8s elapsed with the pool force-disconnecting the connection mid-request. That failure is safe by itself; the reachable bad case is a provider that *completes* the rotation at ~21s, finds the connection gone, cannot commit `swap_in`, and leaves the refresh token rotated upstream with the grant dead until an admin reconnects. Before this branch no connection was held across the exchange at all. `receive_timeout: 3_000` makes it 18 < 20; the existing drip test sends a chunk every 250ms and is unaffected.

**The arithmetic is now pinned, not the wall clock.** The numbers live in two modules and the ceiling is a sum, not a maximum -- neither fact is visible from either file alone, which is how it was wrong to begin with. `OAuth.refresh_timeout_ceiling_ms/0` and `RefreshLock.transaction_timeout_ms/0` make the relationship assertable and a test asserts it. Verified by reverting: at the old `12_000` it fails with `left: 27000, right: 20000`.

**Nothing distinguished a transaction-scoped lock from a session-scoped one.** Swapping `pg_try_advisory_xact_lock` for `pg_try_advisory_lock` left all twelve tests green. "worker death" passes either way because DBConnection drops the whole connection when the client dies brutally; "rollback" passes because advisory locks are re-entrant within a session. Both prove release, neither proves scope. The new test asks the connection itself once the transaction has ended and it is back in the pool, across a clean return, a rollback and a raise. Verified by reverting: the session variant fails it with `left: 3, right: 0`, and it is the only failure.

**Minors, all three taken.** The contention counter is now in `FountainWeb.Telemetry` (without an entry it reached nothing but the test's own handler); the pool-capacity probe's timeout goes 1s to 5s, since it only has to tell "a connection is free" from "both are pinned"; and the moduledoc now says `run/3` must not be called inside a transaction.

Local gates on this head: 97 tests across the coordination, tenant, platform, lifecycle, inference-credential, admin-LiveView and telemetry suites, 0 failures; `mix format --check-formatted` clean; `mix compile --warnings-as-errors` clean; `MIX_ENV=dev mix credo --strict` no issues.

## Landing validation

Rebased only this PR’s three commits onto main `6e632dde8e4f93b1ae62e6ac5f642cf51d19e936` after #2012 merged. `git range-diff` reports all three patches unchanged. The lock, OAuth, refresher, context, and concurrency-test files are byte-identical to approved head `5f993b64`; the telemetry module and metrics test differ only by main’s existing CA-install metric additions. New head: `9c90ff9b`.

`mix precommit` passed on Elixir 1.19.2 / OTP 28.3 with the ten affected ChatGPT, inference, admin, and telemetry test files and a dedicated test database: **134 tests, zero failures**, plus compilation, formatting, strict Credo, Dialyzer (zero errors), Sobelow, and production release assembly. `git diff --check` passed. This local run selected the affected tests; fresh PR and merge-queue CI provide the complete suite against main.

### DIFF
--- a/apps/fountain/lib/fountain/chatgpt_accounts.ex
+++ b/apps/fountain/lib/fountain/chatgpt_accounts.ex
@@ -56,6 +56,10 @@ defmodule Fountain.ChatGPTAccounts do
   a turn that outlives its access token fails at the proxy, because codex
   cannot refresh in this mode. The default is fifteen minutes.
 
+  Platform refresh is coordinated across nodes with a per-grant PostgreSQL
+  try-lock. Only the holder retains a database checkout across the provider
+  request; contenders release theirs between bounded retries.
+
   ## User grants
 
   `status_for_user/1` and `credential_for_user/3` read a tenant's own grant.
@@ -71,7 +75,7 @@ defmodule Fountain.ChatGPTAccounts do
   require Logger
 
   alias Fountain.Audit
-  alias Fountain.ChatGPTAccounts.{Cipher, Grant}
+  alias Fountain.ChatGPTAccounts.{Cipher, Grant, RefreshLock}
   alias Fountain.PlatformChatGPT.{Account, OAuth, Refresher, Tokens}
   alias Fountain.Repo
 
@@ -461,14 +465,48 @@ defmodule Fountain.ChatGPTAccounts do
   end
 
   @doc false
-  # The body of a refresh, run by `Fountain.PlatformChatGPT.Refresher` and
-  # by nothing else: one at a time on this node, no database lock and no
-  # connection held across the HTTP round-trip. The row is re-read here
-  # because a queued caller may find the refresh already done, and the
-  # write checks the generation, version and active state this read started
-  # from. A losing refresher may serve a winner from the same generation,
-  # but cannot overwrite or serve a replacement account.
+  # The local queue bounds platform holders to one per node. The database
+  # try-lock excludes other nodes without parking waiters on pool connections.
+  # Observe identity before waiting, then re-read under the lock: even forced
+  # refresh contenders serve a winner instead of exchanging its rotated token.
   def platform_refresh_serialized(mode) do
+    case platform_row() do
+      %Account{status: "active", refresh_token_ciphertext: cipher} = observed
+      when is_binary(cipher) ->
+        observed.id
+        |> RefreshLock.run(fn -> refresh_locked(observed, mode) end)
+        |> finish_refresh()
+
+      _ ->
+        refresh_without_exchange()
+    end
+  end
+
+  defp refresh_locked(observed, mode) do
+    case platform_row() do
+      %Account{id: id, generation: generation} = current
+      when id == observed.id and generation == observed.generation ->
+        cond do
+          current.status != "active" or current.lock_version != observed.lock_version ->
+            current_result(observed)
+
+          mode == :if_stale and fresh?(current) ->
+            Cipher.decrypt_token(current, :access_token)
+
+          true ->
+            do_refresh(current)
+        end
+
+      nil ->
+        {:error, :not_connected}
+
+      %Account{} ->
+        {:error, :stale_grant}
+    end
+  end
+
+  # Static workspace tokens do not need a refresh lock.
+  defp refresh_without_exchange do
     case platform_row() do
       nil ->
         {:error, :not_connected}
@@ -482,13 +520,17 @@ defmodule Fountain.ChatGPTAccounts do
       %Account{refresh_token_ciphertext: nil} = row ->
         Cipher.decrypt_token(row, :access_token)
 
-      row when mode == :if_stale ->
-        if fresh?(row), do: Cipher.decrypt_token(row, :access_token), else: do_refresh(row)
-
-      row ->
-        do_refresh(row)
+      %Account{} ->
+        {:error, :stale_grant}
     end
   end
+
+  defp finish_refresh({:revoked, account_id, code}) do
+    record_revocation(account_id, code)
+    {:error, :revoked}
+  end
+
+  defp finish_refresh(result), do: result
 
   defp do_refresh(current) do
     with {:ok, refresh} <- Cipher.decrypt_token(current, :refresh_token) do
@@ -501,7 +543,7 @@ defmodule Fountain.ChatGPTAccounts do
 
         {:error, {:terminal, code}} ->
           case mark_revoked(current, code) do
-            :ok -> {:error, :revoked}
+            :ok -> {:revoked, current.account_id, code}
             :stale -> current_result(current)
           end
 
@@ -598,18 +640,17 @@ defmodule Fountain.ChatGPTAccounts do
       )
 
     if count == 1 do
-      record_revocation(row, code)
       :ok
     else
       :stale
     end
   end
 
-  defp record_revocation(row, code) do
+  defp record_revocation(account_id, code) do
     Audit.record_admin(%{
       actor_user_id: nil,
       event_type: "admin.platform_chatgpt.revoked",
-      metadata: %{"actor" => @system_actor, "reason" => code, "account_id" => row.account_id}
+      metadata: %{"actor" => @system_actor, "reason" => code, "account_id" => account_id}
     })
 
     Logger.warning(
--- a/apps/fountain/lib/fountain/chatgpt_accounts/refresh_lock.ex
+++ b/apps/fountain/lib/fountain/chatgpt_accounts/refresh_lock.ex
@@ -10,6 +10,15 @@ defmodule Fountain.ChatGPTAccounts.RefreshLock do
   The callback must re-read the grant and fence its writes. It must not take
   a grant row lock before contacting the provider or emit best-effort audit
   events inside the transaction. Its result is returned only after commit.
+  `run/3` must not itself be called inside a transaction: the lock would
+  attach to the enclosing one and outlive the callback by however long that
+  transaction runs. The one caller is `Fountain.PlatformChatGPT.Refresher`,
+  which holds none.
+
+  Whatever the callback does upstream has to finish inside
+  `@transaction_timeout`, which is a wall clock over the whole checkout
+  rather than a per-query budget. `Fountain.PlatformChatGPT.OAuth.refresh/1`
+  sizes its HTTP timeouts against this number and explains the arithmetic.
   A provider rotation followed by a crash before commit still needs reconnect:
   database exclusion cannot make an external token exchange atomic.
   """
@@ -21,6 +30,10 @@ defmodule Fountain.ChatGPTAccounts.RefreshLock do
   @namespace 52_001
   @wait_timeout 5_000
   @transaction_timeout 20_000
+
+  @doc "The wall clock the holder's whole checkout runs under, in milliseconds."
+  @spec transaction_timeout_ms() :: pos_integer()
+  def transaction_timeout_ms, do: @transaction_timeout
 
   @doc false
   def run(grant_id, fun, opts \\ []) when is_binary(grant_id) and is_function(fun, 0) do
--- a/apps/fountain/lib/fountain/chatgpt_accounts/refresh_lock.ex
+++ b/apps/fountain/lib/fountain/chatgpt_accounts/refresh_lock.ex
@@ -1,0 +1,72 @@
+defmodule Fountain.ChatGPTAccounts.RefreshLock do
+  @moduledoc """
+  Cross-node exclusion for a grant's upstream refresh.
+
+  Only the holder keeps a database checkout across the provider request.
+  Contenders try a transaction-scoped advisory lock and release the checkout
+  before backing off. The transaction releases its lock on commit, rollback,
+  connection loss or worker death; no session lock can leak into the pool.
+
+  The callback must re-read the grant and fence its writes. It must not take
+  a grant row lock before contacting the provider or emit best-effort audit
+  events inside the transaction. Its result is returned only after commit.
+  A provider rotation followed by a crash before commit still needs reconnect:
+  database exclusion cannot make an external token exchange atomic.
+  """
+
+  alias Fountain.Repo
+
+  # Separate from connection, admission and quota locks. Hash collisions only
+  # serialize unrelated grants; they cannot weaken mutual exclusion.
+  @namespace 52_001
+  @wait_timeout 5_000
+  @transaction_timeout 20_000
+
+  @doc false
+  def run(grant_id, fun, opts \\ []) when is_binary(grant_id) and is_function(fun, 0) do
+    deadline = monotonic_ms() + Keyword.get(opts, :wait_timeout, @wait_timeout)
+    timeout = Keyword.get(opts, :transaction_timeout, @transaction_timeout)
+    attempt(:erlang.phash2(grant_id), fun, deadline, timeout)
+  rescue
+    DBConnection.ConnectionError -> {:error, :refresh_unavailable}
+  end
+
+  defp attempt(key, fun, deadline, timeout) do
+    result =
+      Repo.transaction(
+        fn ->
+          case Repo.query!("SELECT pg_try_advisory_xact_lock($1, $2)", [@namespace, key]) do
+            %{rows: [[true]]} -> {:acquired, fun.()}
+            %{rows: [[false]]} -> :busy
+          end
+        end,
+        timeout: timeout
+      )
+
+    case result do
+      {:ok, {:acquired, result}} ->
+        result
+
+      {:ok, :busy} ->
+        # This event is outside the transaction, as is the bounded backoff.
+        :telemetry.execute([:fountain, :chatgpt, :refresh_lock, :contention], %{count: 1}, %{})
+        retry(key, fun, deadline, timeout)
+
+      {:error, _reason} ->
+        {:error, :refresh_unavailable}
+    end
+  end
+
+  defp retry(key, fun, deadline, timeout) do
+    remaining = deadline - monotonic_ms()
+
+    if remaining > 0 do
+      Process.sleep(min(remaining, 25 + :rand.uniform(50)))
+      attempt(key, fun, deadline, timeout)
+    else
+      {:error, :refresh_busy}
+    end
+  end
+
+  defp monotonic_ms, do: System.monotonic_time(:millisecond)
+end
--- a/apps/fountain/lib/fountain/platform_chatgpt/oauth.ex
+++ b/apps/fountain/lib/fountain/platform_chatgpt/oauth.ex
@@ -35,6 +35,40 @@ defmodule Fountain.PlatformChatGPT.OAuth do
   @doc "Codex's OAuth client id."
   def client_id, do: @client_id
 
+  @refresh_connect_timeout 2_000
+  @refresh_pool_timeout 1_000
+  @refresh_receive_timeout 3_000
+  @refresh_request_timeout 12_000
+
+  defp refresh_finch_options do
+    [
+      conn_opts: [transport_opts: [timeout: @refresh_connect_timeout]],
+      protocols: [:http1],
+      pool_timeout: @refresh_pool_timeout,
+      receive_timeout: @refresh_receive_timeout,
+      request_timeout: @refresh_request_timeout
+    ]
+  end
+
+  @doc """
+  The longest `refresh/1` can take before it gives up, in milliseconds.
+
+  A **sum**, not a maximum. `:request_timeout` is checked between receives
+  and each receive may then block for a full `:receive_timeout`, so a call
+  that exhausts the first can still spend the second on one last receive,
+  on top of connecting and waiting for a pool slot.
+
+  `Fountain.ChatGPTAccounts.RefreshLock` holds a database checkout for the
+  whole exchange, so this number has to stay under its transaction budget.
+  `Fountain.ChatGPTRefreshCoordinationTest` asserts exactly that, because
+  the two are set in different modules and the arithmetic is not obvious.
+  """
+  @spec refresh_timeout_ceiling_ms() :: pos_integer()
+  def refresh_timeout_ceiling_ms do
+    @refresh_request_timeout + @refresh_receive_timeout + @refresh_connect_timeout +
+      @refresh_pool_timeout
+  end
+
   @doc """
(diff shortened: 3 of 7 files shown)


BinaryBourbon (dismissed):
Reviewed the lock slice only (base `9152adcd`). The design is sound: one pinned checkout per node, contenders hold nothing, the re-read fence is real, and the revocation audit is correctly outside the transaction. I verified those four by breaking each one and watching the right test fail. Two things below I'd want changed; the rest are notes.

## Verified by reverting

| Break | Result |
|---|---|
| `pg_try_advisory_xact_lock` -> `SELECT true` | both "independent refreshers exchange once" tests fail, mailbox showing the second `{:upstream, ..., "rt_original"}` |
| drop `current.lock_version != observed.lock_version` (chatgpt_accounts.ex:490) | `:force` variant fails: waiter re-exchanges instead of serving the winner |
| `record_revocation` moved inside the transaction | `refute Repo.in_transaction?()` fails at test:202 |
| `Process.sleep(2_000)` before returning `:busy` (backoff inside the transaction) | "many contenders release pool capacity" fails |

Also checked: 12/12 green, four consecutive runs at 13.1s each; `mix credo --strict` clean; CI green on every job; namespace 52001 collides with nothing (existing are 4315 quotas incl. fleet key `0`, 4316 sandbox, 4317 sandbox queue, 4331 connections refresh, plus the one-arg `@first_admin_lock`, a separate objsubid space); `refresh_serialized` has exactly one non-test caller (`Refresher.handle_call`), so at most one checkout per node is pinned and nothing holding the fleet lock can reach this lock — no ordering cycle.

The Finch claim holds for the pinned version. finch 0.23.0, `deps/finch/lib/finch.ex:864`: "This timeout only applies to HTTP/1"; the HTTP/2 pool arms its timer from `opts[:receive_timeout]` instead (`http2/pool.ex:735`). And the pin does not leak into the general case: req 0.7.4 routes `finch: [...]` pool options into a **separate** dynamically started Finch instance named by a hash of those options (`deps/req/lib/req/finch.ex:586-598`), leaving the default `Req.Finch` pool alone.

`transaction_timeout` is a real whole-hold wall clock, not a per-query one: DBConnection arms the deadline once at checkout (`db_connection/lib/db_connection/holder.ex:318`) and does not re-arm it. Probe: 400ms of work under `transaction_timeout: 300` returned `{:error, :refresh_unavailable}`.

## 1. The HTTP exchange can outlive the 20s transaction budget (reproduced)

`:request_timeout` is only checked *between* recvs, and each recv may block a full `:receive_timeout`:

- `deps/finch/lib/finch/http1/conn.ex:298` — the `request_timeout < 0` guard runs before `recv`
- `deps/finch/lib/finch/http1/conn.ex:313-322` — `Mint.HTTP.recv(mint, 0, timeouts.receive_timeout)`, then decrement

So the ceiling is `request_timeout + receive_timeout + connect + pool` = 12 + 12 + 2 + 1 = about 27s, not about 15s.

I reproduced it with your own drip harness, changed only so the provider drips for 11s and then goes silent instead of dripping forever:

```
>>> elapsed=22836ms result={:error, :refresh_unavailable}
[error] Postgrex.Protocol (...) disconnected: ** (DBConnection.ConnectionError)
client #PID<0.506.0> timed out because it queued and checked out the connection
for longer than 20000ms
```

11s of drip leaves about 1s of `request_timeout`, then one recv blocks for the full 12s. 11 + 12 = 23s, matching the measurement.

The failure itself is safe (the backend dies, the lock goes with it, no write lands, the caller gets an error). What I care about is the next case over: if the provider *completes* the rotation at ~21s instead of going silent, the DB connection is already force-killed, so `swap_in` cannot commit. Upstream has rotated the refresh token and we still hold the old one — the grant is dead until someone reconnects at /admin/inference. Your notes call that out for a crash; this makes it reachable from a merely slow provider, and before this PR no DB connection was held across the exchange at all, so a slow-but-successful 25s exchange used to store fine.

One line fixes it: make the HTTP ceiling provably smaller than the transaction budget, e.g. `receive_timeout: 3_000` (12 + 3 + 2 + 1 = 18s < 20s). The existing drip test drips every 250ms, so it still passes.

## 2. Nothing in the suite distinguishes transaction-scoped from session-scoped

Swapping `pg_try_advisory_xact_lock` -> `pg_try_advisory_lock` — the exact failure the moduledoc rules out ("no session lock can leak into the pool") — leaves all 12 tests green:

```
Finished in 13.2 seconds (0.00s async, 13.2s sync)
12 tests, 0 failures
```

"worker death releases the lock" passes because DBConnection disconnects the connection when the client dies brutally, which drops a session lock too. "rollback releases the lock" passes because advisory locks are re-entrant within a session, and the pool has a second connection anyway. Neither observes scope.

Four lines close it. I confirmed this fails on the session variant ("1 advisory lock(s) survived into the pooled connection") and passes as written:

```elixir
%{rows: [[held]]} =
  Repo.query!(
    "SELECT count(*) FROM pg_locks WHERE locktype = 'advisory' AND pid = pg_backend_pid()"
  )

assert held == 0
```

The code is right; the strongest claim in the moduledoc just is not pinned by anything.

## Minor

**The contention counter is emitted but never collected.** `[:fountain, :chatgpt, :refresh_lock, :contention]` (refresh_lock.ex:52) is not in `FountainWeb.Telemetry`'s metrics list, which carries every other `fountain.*` counter. The one signal that would tell an operator cross-node contention is actually happening reaches only the test's `:telemetry.attach`.

**test:183 is the thinnest timing assumption in the file.** `Repo.query!("SELECT 1", [], timeout: 1_000)` with six contenders churning a two-connection pool every 25-75ms. It held over four local runs and CI, but it is the line I would expect to flake first on a loaded runner, and 5_000 asserts the same thing.

**Committed rows and the SQL sandbox.** Cleanup looks complete — I checked that `platform_disconnect`'s audit metadata carries `account_id`, so the disconnect event is caught by the `on_exit` filter too. Worth saying out loud that if that cleanup ever misses a row, the failure surfaces in `platform_chatgpt_test.exs` (which asserts exact `admin_audit_events` lists), not here. I reproduced that blast radius by accident with my own probe and it looked like four unrelated failures in a file this PR does not touch.

**Nesting is undefended, though currently unreachable.** `RefreshLock.run` inside an enclosing `Repo.transaction` would attach the xact lock to the outer transaction and hold it well past the callback. The only prod caller is the `Refresher` GenServer, so it cannot happen today, but the moduledoc's contract list ("must re-read", "must not take a row lock", "must not audit inside") would be the place to add "must not be called inside a transaction".

BinaryBourbon (changes requested):
Reviewed the lock slice only (base `9152adcd`). The design is sound: one pinned checkout per node, contenders hold nothing, the re-read fence is real, and the revocation audit is correctly outside the transaction. I verified those four by breaking each one and watching the right test fail. Two things below I'd want changed; the rest are notes.

## Verified by reverting

| Break | Result |
|---|---|
| `pg_try_advisory_xact_lock` -> `SELECT true` | both "independent refreshers exchange once" tests fail, mailbox showing the second `{:upstream, ..., "rt_original"}` |
| drop `current.lock_version != observed.lock_version` (chatgpt_accounts.ex:490) | `:force` variant fails: waiter re-exchanges instead of serving the winner |
| `record_revocation` moved inside the transaction | `refute Repo.in_transaction?()` fails at test:202 |
| `Process.sleep(2_000)` before returning `:busy` (backoff inside the transaction) | "many contenders release pool capacity" fails |

Also checked: 12/12 green, four consecutive runs at 13.1s each; `mix credo --strict` clean; CI green on every job; namespace 52001 collides with nothing (existing are 4315 quotas incl. fleet key `0`, 4316 sandbox, 4317 sandbox queue, 4331 connections refresh, plus the one-arg `@first_admin_lock`, a separate objsubid space); `refresh_serialized` has exactly one non-test caller (`Refresher.handle_call`), so at most one checkout per node is pinned and nothing holding the fleet lock can reach this lock — no ordering cycle.

The Finch claim holds for the pinned version. finch 0.23.0, `deps/finch/lib/finch.ex:864`: "This timeout only applies to HTTP/1"; the HTTP/2 pool arms its timer from `opts[:receive_timeout]` instead (`http2/pool.ex:735`). And the pin does not leak into the general case: req 0.7.4 routes `finch: [...]` pool options into a **separate** dynamically started Finch instance named by a hash of those options (`deps/req/lib/req/finch.ex:586-598`), leaving the default `Req.Finch` pool alone.

`transaction_timeout` is a real whole-hold wall clock, not a per-query one: DBConnection arms the deadline once at checkout (`db_connection/lib/db_connection/holder.ex:318`) and does not re-arm it. Probe: 400ms of work under `transaction_timeout: 300` returned `{:error, :refresh_unavailable}`.

## 1. The HTTP exchange can outlive the 20s transaction budget (reproduced)

`:request_timeout` is only checked *between* recvs, and each recv may block a full `:receive_timeout`:

- `deps/finch/lib/finch/http1/conn.ex:298` — the `request_timeout < 0` guard runs before `recv`
- `deps/finch/lib/finch/http1/conn.ex:313-322` — `Mint.HTTP.recv(mint, 0, timeouts.receive_timeout)`, then decrement

So the ceiling is `request_timeout + receive_timeout + connect + pool` = 12 + 12 + 2 + 1 = about 27s, not about 15s.

I reproduced it with your own drip harness, changed only so the provider drips for 11s and then goes silent instead of dripping forever:

```
>>> elapsed=22836ms result={:error, :refresh_unavailable}
[error] Postgrex.Protocol (...) disconnected: ** (DBConnection.ConnectionError)
client #PID<0.506.0> timed out because it queued and checked out the connection
for longer than 20000ms
```

11s of drip leaves about 1s of `request_timeout`, then one recv blocks for the full 12s. 11 + 12 = 23s, matching the measurement.

The failure itself is safe (the backend dies, the lock goes with it, no write lands, the caller gets an error). What I care about is the next case over: if the provider *completes* the rotation at ~21s instead of going silent, the DB connection is already force-killed, so `swap_in` cannot commit. Upstream has rotated the refresh token and we still hold the old one — the grant is dead until someone reconnects at /admin/inference. Your notes call that out for a crash; this makes it reachable from a merely slow provider, and before this PR no DB connection was held across the exchange at all, so a slow-but-successful 25s exchange used to store fine.

One line fixes it: make the HTTP ceiling provably smaller than the transaction budget, e.g. `receive_timeout: 3_000` (12 + 3 + 2 + 1 = 18s < 20s). The existing drip test drips every 250ms, so it still passes.

## 2. Nothing in the suite distinguishes transaction-scoped from session-scoped

Swapping `pg_try_advisory_xact_lock` -> `pg_try_advisory_lock` — the exact failure the moduledoc rules out ("no session lock can leak into the pool") — leaves all 12 tests green:

```
Finished in 13.2 seconds (0.00s async, 13.2s sync)
12 tests, 0 failures
```

"worker death releases the lock" passes because DBConnection disconnects the connection when the client dies brutally, which drops a session lock too. "rollback releases the lock" passes because advisory locks are re-entrant within a session, and the pool has a second connection anyway. Neither observes scope.

Four lines close it. I confirmed this fails on the session variant ("1 advisory lock(s) survived into the pooled connection") and passes as written:

```elixir
%{rows: [[held]]} =
  Repo.query!(
    "SELECT count(*) FROM pg_locks WHERE locktype = 'advisory' AND pid = pg_backend_pid()"
  )

assert held == 0
```

The code is right; the strongest claim in the moduledoc just is not pinned by anything.

## Minor

**The contention counter is emitted but never collected.** `[:fountain, :chatgpt, :refresh_lock, :contention]` (refresh_lock.ex:52) is not in `FountainWeb.Telemetry`'s metrics list, which carries every other `fountain.*` counter. The one signal that would tell an operator cross-node contention is actually happening reaches only the test's `:telemetry.attach`.

**test:183 is the thinnest timing assumption in the file.** `Repo.query!("SELECT 1", [], timeout: 1_000)` with six contenders churning a two-connection pool every 25-75ms. It held over four local runs and CI, but it is the line I would expect to flake first on a loaded runner, and 5_000 asserts the same thing.

**Committed rows and the SQL sandbox.** Cleanup looks complete — I checked that `platform_disconnect`'s audit metadata carries `account_id`, so the disconnect event is caught by the `on_exit` filter too. Worth saying out loud that if that cleanup ever misses a row, the failure surfaces in `platform_chatgpt_test.exs` (which asserts exact `admin_audit_events` lists), not here. I reproduced that blast radius by accident with my own probe and it looked like four unrelated failures in a file this PR does not touch.

**Nesting is undefended, though currently unreachable.** `RefreshLock.run` inside an enclosing `Repo.transaction` would attach the xact lock to the outer transaction and hold it well past the callback. The only prod caller is the `Refresher` GenServer, so it cannot happen today, but the moduledoc's contract list ("must re-read", "must not take a row lock", "must not audit inside") would be the place to add "must not be called inside a transaction".

BinaryBourbon (changes requested):
Re-reviewed at `6f8715d4`. The arithmetic fix is right and I confirmed the ceiling test fails at the old value (`left: 27000, right: 20000`, same as you). Agreed on skipping a wall-clock mirror of my repro: the bug was arithmetic, the constants now sit behind named functions, and the file already pays ~12s for one drip test that exercises the real path. Two things below, then this is good.

## 1. CI is red, and it is this change

`FountainWeb.MetricsTest` "every subscribed fountain event has a live producer" (partition 4, run 34669931493). Reproduced locally:

```
[:fountain, :chatgpt, :refresh_lock, :contention, :count] subscribes to
[:fountain, :chatgpt, :refresh_lock, :contention], which no known code path
emits -- this is how #310 happened
```

The guard keeps an explicit producer allowlist, so the new counter needs a line in `test/fountain_web/metrics_test.exs` naming what fires it. With this added it goes to 21 tests, 0 failures:

```elixir
# Fountain.ChatGPTAccounts.RefreshLock.attempt/4, when another node
# holds the grant's lock -- exercised by
# Fountain.ChatGPTRefreshCoordinationTest, which attaches to it.
[:fountain, :chatgpt, :refresh_lock, :contention],
```

## 2. The scope test does not reliably catch the regression it was written for

It reads `pg_backend_pid()` on whichever of the pool's two connections answers, which need not be the one that ran `run/3`. Against the session-lock variant:

- run alone (`mix test ...:255`): **0 failures in 8 runs** — it never catches it
- full file: 3 failures in 4 runs, with `left:` varying 1, 2, 3 (you saw 3, I saw 1 and 2)

So it works by accident of connection reuse, and not at all in the form someone debugging would reach for. Pinning the connection makes it deterministic — `Repo.checkout/1` around the whole body, which is not a transaction so it does not violate the rule you just documented:

```elixir
Repo.checkout(fn ->
  assert {:ok, :done} = RefreshLock.run(grant_id, fn -> {:ok, :done} end)
  assert advisory_locks_held() == 0
  ...
end)
```

Measured: **6/6 detect the session variant, 6/6 pass as written**, run alone both times. The comment above the helper ("Every checkout in this pool ran `RefreshLock.run/3`") is what does not hold — the test process checks out for its own queries too.

## On `receive_timeout: 3_000` — I would rebalance it, non-blocking

3s is tighter than it looks, because `receive_timeout` is the **server think-time** budget, not just an inter-chunk gap. After the request is sent, Finch goes straight into `receive_response([], ...)` (`http1/conn.ex:130-141`) and the `[]` clause calls `Mint.HTTP.recv(mint, 0, timeouts.receive_timeout)` (`:303-314`). So the first receive — waiting for the provider to start answering at all — is bounded by `receive_timeout`, not by `request_timeout`. This drops the time `auth.openai.com` has to begin responding from 12s to 3s.

For a few-hundred-byte JSON body `request_timeout` is nearly idle: the body lands in one segment, so the total-response budget only binds in the adversarial drip case, which is not the case you want to be generous to. `request_timeout: 9_000, receive_timeout: 6_000` holds the same 18s ceiling and doubles the budget for legitimate provider slowness. A first byte at 5s succeeds under 9/6 and fails under 12/3.

Not blocking: the failure is soft (`{:error, reason}`, "keeping the current token", grant untouched, retried next turn), so the worst case is a transient fallback to `PLATFORM_OPENAI_API_KEY` rather than a burnt grant. Your call — I just would not want 3s chosen on the belief that it only governs gaps between chunks.

## Verified

14/14 green locally. Ceiling test fails at the old constant and passes at the new one. `refresh_finch_options/0` changes nothing but `receive_timeout` — same four keys, same values otherwise. The `cipher.ex` logging in the range came in with #2012's base, not this slice. The moduledoc's no-nesting note, the telemetry counter and test:183's 5s all read correctly.

BinaryBourbon (approved):
All three fixed at `5f993b64`, and I re-verified each rather than reading the diff.

- **Scope test.** Deterministic in both directions now: against the session-lock variant it fails **6/6 run alone** (`mix test ...:256`) and **3/3 as a full file**; as written it passes. Before the pin it was 0/8 alone. `Repo.checkout/1` is the right tool — a pinned connection, not a transaction, so `run/3`'s contract holds.
- **Metrics guard.** 21 tests, 0 failures locally with the allowlist entry.
- **The 9/6 split.** 14/14 green over three runs, and the file dropped from 13.2s to 10.4s — the ~2.8s is the drip test now ending at `request_timeout` 9s instead of 12s, which is the arithmetic behaving as intended.

## On the drip test's bounds: the trade is right, keep it

It weakened slightly and harmlessly, and I would not tighten the upper bound.

The upper bound was never what that test was for. `assert {:error, {:token, _timeout}}` is the load-bearing line — it proves the exchange *ended in a timeout* rather than hanging or succeeding — and `assert {:ok, :released} = RefreshLock.run(...)` proves the lock came back. Both are untouched. What `19_000` encoded was arithmetic, and arithmetic now has a dedicated test that fails at the constant instead of at a wall clock on a loaded runner. That is a better home for it, not a weaker one.

A tighter derived bound would have to be `request_timeout + connect + pool`, which means exposing a fourth public function for one assertion's upper limit. Not worth it. `< refresh_timeout_ceiling_ms()` is also the conceptually correct bound: the claim being made is "the exchange finishes inside the budget the transaction gives it", and that is exactly what it says now.

The staleness trap was real, not hypothetical — two hardcoded copies of constants owned by another module went red the moment you rebalanced. Same failure shape as the `@nav` mirror CLAUDE.md warns about.

**One gap, and I think it is fine to accept.** Nothing now pins *which* timeout ends the drip case. Swap the two constants back to 6/9 and the ceiling stays 18s, the ceiling test still passes, and the drip ends at ~6s — still inside `>= 1_000` and `< 18_000`. The rebalance would silently come undone. That is the one thing `>= 11_000` did catch.

I would accept it, because the split is an availability judgement rather than a safety property: getting it wrong costs a little provider headroom, it does not break anything, and the safety property (fits in the transaction) *is* tested. The rationale is also now written at the call site where someone changing the constants will read it. If you want it pinned anyway, the cheap version is to expose the request timeout the way you exposed the ceiling and assert `elapsed >= div(OAuth.refresh_request_timeout_ms(), 2)` — but treat that as optional, not something I am asking for.

## Approving

Everything from both rounds is addressed: the lock is transaction-scoped and now provably so, the exchange cannot outlast the transaction that holds its connection, the re-read fence and the post-commit audit ordering hold, contenders release their checkouts, and the contention counter is both registered and allowlisted. Nice work on the 9/6 reasoning in the call-site comment — it explains the non-obvious half (that `receive_timeout` bounds the first receive) rather than just the numbers.

PR #2014, opened 2026-09-12
feat: renew user ChatGPT grants through bounded workers

jhgaylor:
Near-expiry user ChatGPT grants currently return `:refresh_required`. This adds owner-scoped renewal through bounded workers, then re-reads the caller's pinned generation to return a matching bearer and source snapshot. It never falls through to platform or paid inference.

#2011–#2013 are merged. This PR is based directly on main. User linking, conversation selection, and account API/UI integration remain unbuilt. Keepalive scheduling follows separately in #2015.

- Coalesce callers by owner, grant ID and generation. Admit at most four user workers per node (fewer with a small DB pool), cap each grant at 128 waiters, and apply a 27-second worker deadline. The task supervisor enforces the cap too; coordinator restart tears down its workers. An exchange may finish after its last caller leaves so a rotated token can still be persisted.
- Workers and queue replies carry IDs/status only. After renewal each caller reads its own generation again; no bearer, key or provider response is cached by the coordinator.
- Share platform/user encryption and conditional-write mechanics without mixing ownership. Keep omitted refresh-token rotations intact; reject a changed provider account or user identity. First reads and writes require a verified, claimed, non-suspended owner.
- Re-read under the PostgreSQL refresh lock and fence writes by owner, ID, generation, version and active state. Transient provider errors become a stable retryable error without logging or returning the body. Terminal refusal emits one tenant `chatgpt.reconnect_required` event after commit, with no admin fallback event.

Validation:

- 31 new tests cover coalescing, capacity/waiter limits, worker crash/deadline, caller departure, token redaction, owner/DEK isolation, omitted rotation, source identity, transient and terminal failures, and eligibility. Real independent PostgreSQL connections prove two users and the platform renew separately, same-grant callers issue one exchange, disconnect/reconnect/suspension commit during provider I/O, and tenant audit runs after commit.
- The stale-user-credential test fails on the previous context with `:refresh_required`, and passes with this implementation. The initial synthetic coordinator fixture used Elixir's inspected PID form where Erlang's PID form was required; fixing that fixture made its tests pass. Precommit also identified one call-layout formatting correction, which was applied.
- Full `mix precommit` before rebasing: compile, formatting, Credo, Dialyzer, Sobelow, production release assembly, 5,514 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests, zero failures (2 skipped / 7 excluded core cases).
- After rebasing onto the reviewed timeout/diagnostic fixes: all precommit static/release gates and 108 affected tests passed on `65aa52d8`, with zero failures.

- CI follow-up: the reviewed base added the refresh-lock contention metric but omitted its producer from the metrics test inventory. Registered the existing `RefreshLock.attempt/4` producer; its emission is already exercised by the independent-connection concurrency tests. The previously failing partition now passes all 772 tests and exports coverage on `757bfe02`.

The platform's existing local queue stays separate. PostgreSQL remains the cross-node authority; upstream rotation followed by a crash before commit still requires reconnect. `refresh: false` supports internal reads that must not initiate renewal.

## Landing validation

Rebased this PR’s two commits onto main `2fd1ee2785488ecaf26d3b3827bd01bfec5b22c2` after #2013 merged. Main had extracted the application child list into `children/0`; the conflict was resolved by retaining that structure and adding `Fountain.ChatGPTAccounts.RefreshSupervisor` immediately after the platform refresher. The application file is exactly main plus that one child. All other reviewed implementation is preserved; the metrics test also retains main’s existing CA-install metric entry. New head: `3f84c8ee`.

`mix precommit` passed on Elixir 1.19.2 / OTP 28.3 with fourteen affected ChatGPT, inference, admin, telemetry, and application-ordering test files on a dedicated database: **175 tests, zero failures**, plus compilation, formatting, strict Credo, Dialyzer (zero errors), Sobelow, and production release assembly. `git diff --check` passed. This local run selected the affected tests; fresh PR and merge-queue CI provide the complete suite against main.

### DIFF
--- a/apps/fountain/lib/fountain/application.ex
+++ b/apps/fountain/lib/fountain/application.ex
@@ -96,6 +96,7 @@ defmodule Fountain.Application do
         # (#1040). Supervised and unlinked, a crash here is a log line.
         {Task.Supervisor, name: Fountain.TaskSupervisor},
         Fountain.PlatformChatGPT.Refresher,
+        Fountain.ChatGPTAccounts.RefreshSupervisor,
         {DynamicSupervisor, name: Fountain.ExecutionTransportSupervisor, strategy: :one_for_one},
         FountainWeb.Plugs.RateLimit.Sweeper,
         Fountain.Conversations.Redaction,
--- a/apps/fountain/lib/fountain/chatgpt_accounts.ex
+++ b/apps/fountain/lib/fountain/chatgpt_accounts.ex
@@ -64,10 +64,11 @@ defmodule Fountain.ChatGPTAccounts do
 
   `status_for_user/1` and `credential_for_user/3` read a tenant's own grant.
   Both are scoped by the owner and neither falls through to the platform
-  row. Nothing writes a user grant yet: this foundation does not connect
-  them, coordinate refresh across nodes, or select them for conversations,
-  and a read that needs renewal returns `:refresh_required` rather than
-  invoking the platform-only refresher or a paid fallback.
+  row. Near-expiry reads renew through bounded per-grant workers and the
+  PostgreSQL refresh lock, using only the owner's encryption key. Callers
+  re-read their pinned grant after renewal; the coordinator holds no tokens.
+  User linking, conversation selection and user keepalive scheduling remain
+  unbuilt. No account API exposes this internal credential read.
   """
 
   import Ecto.Query, only: [from: 2]
@@ -75,7 +76,8 @@ defmodule Fountain.ChatGPTAccounts do
   require Logger
 
   alias Fountain.Audit
-  alias Fountain.ChatGPTAccounts.{Cipher, Grant, RefreshLock}
+  alias Fountain.Accounts.User
+  alias Fountain.ChatGPTAccounts.{Cipher, Grant, RefreshCoordinator, RefreshLock}
   alias Fountain.PlatformChatGPT.{Account, OAuth, Refresher, Tokens}
   alias Fountain.Repo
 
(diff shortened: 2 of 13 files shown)


BinaryBourbon (approved):
**Approve.** Reviewed at `f42baca7`. CI is green on that exact SHA (run 34671121299, `head_sha f42baca7…`, conclusion `success`), so credo/dialyzer/sobelow/format/the release assemble and all six partitions are covered; I did not re-run them.

The rebase is clean. `git log --oneline -4` puts this slice at `244718aa` + `f42baca7` on `5f993b64`, and the only `metrics_test.exs` change in the slice is the comment rewording on the contention entry (`[:fountain, :chatgpt, :refresh_lock, :contention]` is present, once). Nothing mangled.

## Verified by running

I broke twelve mechanisms one at a time and re-ran; every one has a test that fails. Full worktree restored after each (`git status --porcelain` empty).

| Mechanism broken | Tests that failed |
|---|---|
| Coalescing key drops `user_id` (worker still gets the real one) | `another owner or generation cannot join an existing grant's worker` |
| `safe_result/1` passes worker returns through | `unexpected worker values cannot be returned as credentials or provider bodies` |
| `max_waiters` cap removed | `waiting callers are capped, and departed callers release their slots` |
| Last caller leaving kills the exchange | `an exchange survives its last caller leaving and a later caller joins it` |
| Owner-eligibility join dropped from `user_grant_query/2` | 5 tests, incl. both `suspend commits while an independent user … response is pending` |
| `encrypt_refresh_tokens/2` always platform-encrypts | 4 tests, incl. `two users and the platform renew independently…` |
| AAD drops the field component | `tenant tokens use the tenant DEK and cannot be swapped between fields or owners` |
| `current_query/1` drops the generation/lock_version/status fence | 4 tests, incl. both `reconnect fences a late user … response` |
| `pg_try_advisory_xact_lock` replaced with `SELECT true` | 4 tests, **3 runs out of 3**, deterministic |
| Contender holds its checkout across the backoff | `two users and the platform renew independently…` (`{:contending, pid, true}`), **3/3** |
| `finish_refresh/1` moved inside the lock transaction | `tenant revocation commits before its best-effort audit` (`refute Repo.in_transaction?()`) |
| `validate_refresh_identity/2` always `:ok` | `refresh cannot change the pinned provider account or provider user` |

The two failure shapes the earlier slices hit are both absent here. The exclusion test is deterministic, not seed-luck. And the pool-leak assertion reads `Repo.checked_out?()` **inside the contending process** (`contending/4` fires from `RefreshLock.attempt/4`, which runs in the waiter task) rather than from a connection it never pinned — that is why breaking it fails 3/3 instead of ~half.

Flake probe: 31 tests × 6 runs, random seeds, all green; then 4 more runs with 20 busy-loops on 10 cores, all green. The barriers are message-based and `await_waiters` polls to a deadline rather than sleeping a fixed amount, so I found no timing assumption to complain about.

Supervision, verified live rather than read: a worker parked inside `RefreshLock.run/3` held exactly 1 advisory lock in `pg_locks` (classid 52001); `Process.exit(RefreshCoordinator, :kill)` killed it with `:shutdown` via `:one_for_all`, advisory locks went to 0, and both children came back. A crash cannot strand the lock.

Budgets line up: `@transaction_timeout` 20s (`refresh_lock.ex:32`) > `refresh_timeout_ceiling_ms/0` 18s (`oauth.ex:66-69`), and the 27s worker deadline covers 5s of lock wait + 18s of HTTP.

## Findings — none blocking

**1. `[:fountain, :chatgpt, :refresh, :failure]` is emitted into nothing.** (`chatgpt_accounts.ex:720`) Verified at runtime: `FountainWeb.Telemetry.prometheus_metrics/0` subscribes only to `…refresh_lock.contention`, and `:telemetry.list_handlers/1` for the event returns `[]`; it is also absent from `Fountain.Telemetry`'s `attach_many` list. `refresh_error/2` deliberately does not log on the user path, so a tenant whose grant keeps failing transiently produces **no log line, no metric and no audit row** — the caller just sees `:refresh_failed`. `metrics_test.exs` only checks the forward direction (every subscribed event has a producer), so it cannot catch this. A `counter("fountain.chatgpt.refresh.failure.count", tags: [:scope])` next to the contention counter closes it.

**2. Four workers can hold 40% of the pool for 20 seconds.** `concurrency = min(4, max(1, pool_size - 2))` (`refresh_supervisor.ex:14`); prod `POOL_SIZE` defaults to 10 (`config/runtime.exs:1375`), so 4 workers, each pinning a checkout across the provider call for up to 20s, plus the platform refresher = 5 of 10 during a provider hang, sustained rather than one-off. `pool_size - 2` reads as "leave room" but does not scale with the pool. Something proportional (`min(4, max(1, div(pool_size, 4)))`) would. Non-blocking because the good half is already right — callers do **not** hold a connection while queued (`refresh_for_user/3` checks its `pinned_user_grant/3` back in before `RefreshCoordinator.run/4`) — and because nothing outside tests calls `credential_for_user/4` or `refresh_for_user/3` yet (grepped `apps/` + `ee/`), so there is no live path today.

**3. Nit: the deadline can discard a successful refresh.** `Task.shutdown(job.task, :brutal_kill)` (`refresh_coordinator.ex:75`) returns `{:ok, result}` and consumes the `{ref, result}` message when the task finished just before the timer fired. The result is dropped and every waiter gets `{:error, :refresh_timeout}` for a grant that was in fact renewed; `credential_for_user/4`'s `with :ok <- refresh_for_user(…)` then returns the error instead of re-reading the now-valid row. A few milliseconds at the 27s mark, self-healing next call. Matching on `Task.shutdown`'s return would handle it.

**4. Nit: the audit exclusion is not on the record.** `swap_in/3` and `mark_revoked/2` mutate tenant-owned state in a context; the terminal path audits `chatgpt.reconnect_required`, the successful-rotation path records nothing. That is very likely the right call (high-volume machine state, same class as `Accounts.touch_api_key/1`), but `audit_guardrail_test.exs` is a hand-maintained list and ChatGPT grants appear in neither `@must_audit` nor `@deliberately_silent`, so CLAUDE.md's "the exclusion is a decision on the record rather than an omission" is unmet. One line in `@deliberately_silent`.

## Reading-based, nothing found

- **Tenant isolation.** Both entry points scope on the owner first (`status_for_user/1`, `pinned_user_grant/3` → `user_grant_query/2`), and `current_query/1` reuses the same scoped query for the write. No `_unsafe_` function is introduced or called. The coordinator is keyed by `{user_id, grant_id, generation}` and carries **only** those three IDs plus a status atom — no bearer, no DEK, no provider body — so a shared worker cannot hand one tenant's token to another even if two owners somehow pinned the same grant id; the re-read is per-caller. `safe_result/1` is the belt on top of that, and both are tested.
- **Crypto.** `encrypt_refresh_tokens/2` and `decrypt_token/2` both dispatch on `account.user_id` read off the persisted row, never on caller input, so ownership stays immutable and a user grant cannot be written platform-encrypted or vice versa. AAD is `fountain.chatgpt_grant:<owner>:<field>` on both sides.
- **`Task.async` prohibition.** Not used for fire-and-forget. Workers go through `Task.Supervisor.async_nolink/5` into a dedicated `Task.Supervisor` with `max_children`, and the coordinator owns the ref and handles the reply — the `Analytics.Sink` shape CLAUDE.md allows. `Task.async` appears only in tests.
- **Bounds.** Monitors are capped at `max_concurrency × max_waiters` = 512; `find_worker/2` is linear over ≤4 jobs. The `rescue RuntimeError` around `async_nolink` (for the `max_children` race while a worker is terminating) is exercised by `the task supervisor cap also refuses admission without crashing the coordinator`.

PR #2015, opened 2026-09-12
feat: renew idle user ChatGPT grants with bounded keepalive jobs

jhgaylor:
Idle user ChatGPT grants currently have no scheduled renewal. This adds a daily, paginated keepalive sweep and a separate provider queue, using the owner-scoped refresh path and generation checks from the preceding PR.

#2011–#2014 are merged. This PR is based directly on main. User linking, account surfaces, and conversation selection remain unbuilt; this prepares renewal for grants once linking is available. Platform keepalive remains on its existing path.

- Select at most 100 eligible, active, refreshable user grants per page. The database projects only owner/grant/generation IDs and applies the existing provisional keepalive interval. The sweep does not load encryption keys or contact the provider.
- Insert unique per-grant jobs and the next page together. Deduplicate incomplete work by owner, grant and generation, and spread execution over a five-minute jitter window. Oban can return an unsaved conflict object when its uniqueness lock is busy; a missing job ID now rolls back the page instead of silently advancing the cursor.
- Run provider jobs in their own queue with concurrency four, behind the existing bounded coordinator. Recheck eligibility, generation and renewal timing on execution. Cancel terminal/unusable grants, retry transient failures with Oban's backoff, and stop after five attempts. Jobs and errors contain no token or provider response body.

Validation:

- 14 new tests cover query eligibility and metadata projection, a real 101-grant pagination boundary, replay/deduplication, jitter bounds, atomic rollback on an unsaved conflict, and malformed arguments. Persisted Oban jobs exercise actual renewal, disconnect/reconnect/suspension cancellation, already-renewed no-ops, tenant terminal audit, and retry exhaustion without persisting an echoed provider secret.
- The initial test run had one fixture assertion failure: `Oban.insert` returns the caller's atom-keyed argument map, while the stored/reloaded job has JSON string keys. The assertion now reads the persisted job, matching the property being tested.
- Full `mix precommit` before rebasing passed compile, formatting, Credo, Dialyzer, Sobelow, production release assembly, 5,528 core tests plus 6 doctests, 144 Buzz tests and 33 Support tests, with zero failures (2 skipped / 7 excluded core cases).
- After rebasing onto the reviewed refresh deadline and decrypt diagnostics: all precommit static/release gates and 239 affected tests passed, with zero failures.

- Final head `fa1eb5ae`, after carrying the telemetry inventory fix: all precommit static/release gates and 49 telemetry, concurrency, and keepalive tests passed, with zero failures.

The six-day setting remains provisional pending the provider idle-lifetime measurement in ADR 0047. A new grant inserted behind a running cursor is picked up by the next daily sweep; on-demand reads continue to renew near-expiry credentials.

## Landing validation

Rebased onto main `5edf3af6f23a5a1b2aa6f7d4871b01001c1223a0` after #2014 merged. The configuration conflict retains main’s `notifications: 5` queue and adds the reviewed `chatgpt_refresh: 4` queue and daily sweep. All keepalive implementation and tests retain their reviewed changes; the test helper also retains main’s additions. Corrected the non-blocking review finding in the platform refresher docstring: user renewal and fleet-wide keepalive scheduling are now described as implemented, with references to their actual modules. New head: `a0f67c94`.

`mix precommit` passed on Elixir 1.19.2 / OTP 28.3 with seventeen affected keepalive, ChatGPT, inference, admin, telemetry, application-ordering, and deadline-notification test files on a dedicated database: **203 tests, zero failures**, plus compilation, formatting, strict Credo, Dialyzer (zero errors), Sobelow, and production release assembly. `git diff --check` passed. This local run selected affected tests; fresh PR and merge-queue CI provide the complete suite against main.


BinaryBourbon (approved):
**Approve.** I ran the suite, six revert-checks and a seed stress loop against `1c43d049`. The Oban specifics this repo normally gets wrong are all right here, and the boundedness claim in the title holds under two independent bounds, not one.

## Non-blocking findings

**1. `refresher.ex:17` now describes built behavior as unbuilt (low).**
`apps/fountain/lib/fountain/platform_chatgpt/refresher.ex:17` still reads "Per-user queues and fleet-wide user keepalive scheduling remain unbuilt under ADR 0052." Per-user queues landed in #2014 and fleet-wide keepalive scheduling lands here. `chatgpt_accounts.ex:70-73` was updated in this diff; this docstring was missed. CLAUDE.md: "remove the caveat in the PR that builds it." One-line fix.

**2. Six moduledocs cite an ADR that does not exist (low, stack-wide).**
`ls decisions/` tops out at `0051-record-sandbox-provider-identity.md`. `grep -rn "ADR 0052"` hits `chatgpt_accounts.ex:3,4`, `chatgpt_accounts/grant.ex:7`, `platform_chatgpt/account.ex:3`, `platform_chatgpt/refresher.ex:17` and both new workers. `git log -S"ADR 0052"` puts the first reference in #2011, so this is inherited rather than introduced here — but by this slice the design (generation fencing, owner-scoped encryption, the advisory lock, the bounded coordinator, the paginated sweep) is fully built and nothing writes it down. Worth landing 0052 before the stack merges.

**3. The keepalive queue and the coordinator have the same width (observation, no action today).**
`chatgpt_refresh: 4` (config.exs:20) equals `RefreshCoordinator`'s `min(4, pool_size - 2)` = 4 (`refresh_supervisor.ex:14`). Four in-flight keepalive jobs can therefore hold every coordinator slot, and a fifth caller gets `{:error, :refresh_busy}` with no retry of its own. That is harmless right now: `grep -rn "credential_for_user" apps ee --include="*.ex"` finds no caller outside this context and its tests. When conversation selection lands, an interactive near-expiry read will be able to lose to the nightly sweep — a narrower queue limit or a priority lane is the cheap fix at that point.

## What I verified by running something

- `mix test test/fountain/workers/chatgpt_grant_keepalive_test.exs test/fountain/workers/chatgpt_keepalive_sweep_test.exs` — 14 tests, 0 failures. Twelve consecutive runs on random seeds: 0 failures. `mix test test/fountain/workers test/fountain/chatgpt_accounts_test.exs test/fountain/chatgpt_accounts` — 145 tests, 0 failures. `mix credo --strict` clean, `mix format --check-formatted` exit 0.
- `gh pr checks 2015` is fully green, and run `34671187537` has `head_sha 1c43d049`, so the green covers the post-rebase tree. `git log --oneline` shows `1c43d049` on `f42baca7`, which `gh pr view 2014` confirms is #2014's head — the rebase is intact and the dropped duplicate commit left nothing behind.
- **Revert-checks (six, all caught):**
  - Relax `insert!/1` to `{:ok, job} -> job` → `chatgpt_keepalive_sweep_test.exs:88` fails.
  - Drop `states: :incomplete` from `ChatGPTGrantKeepalive` → `:112` fails (a completed job would swallow the next sweep's insert).
  - Drop `|> with_eligible_owner()` from `_unsafe_due_user_grants` → `:37` fails (principal / unverified / suspended owners leak into the scan).
  - Drop `schedule_in: :rand.uniform(300)` → `:56` fails, so the jitter window is genuinely asserted, not just bounded from above.
  - Remove the `{:cancel, reason}` clause from the grant worker → 4 failures.
  - Make `_unsafe_due_user_grants` call `Crypto.load_tenant_key/1` → `:37` fails, so "the sweep loads no key" is a real guard and not a stub that never runs.

## What I checked by reading

- **`states: :incomplete` is the correct group.** `deps/oban/lib/oban/job.ex:440` resolves it to `suspended available scheduled executing retryable`, so a grant renewed today does not have its next sweep six days later silently deduplicated against a `completed` row — the exact trap that bites event-driven Oban jobs. Oban's 7-day Pruner would have made this a slow-burn bug against a 6-day interval.
- **The unsaved-conflict guard is defending a real path.** `deps/oban/lib/oban/engines/basic.ex:436-439`: on `{:error, :locked}`, `insert_unique/3` returns `{:ok, %Job{}}` built by `Changeset.apply_action/2` — no `id`, never persisted. `acquire_lock/2` is a no-op unless `testing: :disabled` (basic.ex:543), so this branch is unreachable under `Oban.Testing`, which is why mocking `Oban.insert` is the right call rather than an artificial test. The `is_integer(id)` match plus `Repo.rollback/1` means a page never advances its cursor on an insert that is not durable.
- **Queue and cron reach production.** `config/test.exs:108` adds only `testing: :manual` and `config/runtime.exs` names Oban nowhere, so `chatgpt_refresh: 4` and the `"37 4 * * *"` entry survive into prod config. `chatgpt_keepalive_sweep_test.exs:128-134` asserts both, which is the shape that catches a queue added to a worker but not to the supervisor.
- **Boundedness.** Per page: one query, at most 101 `Oban.insert` calls, no N+1. Total in-flight upstream exchanges are bounded twice over — by the queue (4/node) and independently by the coordinator (4/node) — and #2013's `RefreshLock` pins one checkout per exchange, so at most 4 pinned connections per node for up to 20s. The timeout ladder is consistent end to end: RefreshLock 5s wait + 20s transaction ≤ coordinator worker 27s ≤ `GenServer.call` 30s ≤ the worker's `timeout/1` 35s. The cursor advances strictly (`List.last` under `order_by: asc: a.id`), a page only continues when it is full, and the filters share the query with the `LIMIT`, so a short page is genuinely exhausted rather than filtered-down — the loop terminates.
- **Tenant isolation.** `_unsafe_due_user_grants/2` is a system-level sweep, one of the documented legitimate shapes, correctly prefixed and doc'd, and it projects IDs only. A job cannot act on the wrong owner: `refresh_for_user/3` re-enters through `pinned_user_grant/3` → `user_grant_query/2`, which scopes on `user_id` *and* re-applies `with_eligible_owner/1`, then fences on `generation`.
- **`@terminal_errors` is exactly the reachable terminal set.** I traced every `{:error, atom}` reachable from `refresh_for_user/3`. Terminal and listed: `not_connected`, `stale_grant`, `revoked`, `expired`, `invalid_grant`, `account_mismatch`, `undecryptable`, `no_token`, and `not_found` / `unwrap_failed` from `Crypto.load_tenant_key/1`. Transient and correctly retried: `refresh_failed`, `refresh_busy`, `refresh_unavailable`, `refresh_timeout`. The atoms deliberately absent are unreachable here — `no_refresh_token`, `no_account_id` and `invalid_token` belong to the connect path, and `invalid_id_token` is swallowed by `validate_refresh_identity/2` and `refresh_metadata/1`.
- **Idempotency and races.** A grant renewed between sweep and execution hits `fresh?/1 and not stale_for_keepalive?/1` and no-ops; the two thresholds are the same `platform_keepalive_days/0`, and `chatgpt_accounts.ex:881` handles a nil `last_refreshed_at`, so a swept grant can never fall into the no-op branch by accident. A reconnect changes `generation` → `:stale_grant` → cancel. A disconnect → `:not_connected` → cancel. A replayed page re-inserts against still-incomplete jobs and gets their real IDs back.
- **Audit.** `finish_refresh({:user_revoked, ...})` runs on the return value of `RefreshLock.run/3`, i.e. after the transaction commits, not inside it. It records generation and the provider's reason code, no token or response body. `refresh_error/2` on the user branch emits telemetry and returns a flat `:refresh_failed` rather than the provider body, which `chatgpt_grant_keepalive_test.exs:73-100` pins by asserting a synthetic echo never reaches `job.errors` through all five attempts.
- Not exercised end to end: the production behavior of nested `Oban.insert` uniqueness under `testing: :disabled`, where the `pg_try_advisory_xact_lock` attaches to the sweep's own page transaction and is held until it commits. I read Oban 2.23.0's engine to establish that; I did not run it against a live Oban instance. It is bounded (≤101 locks, one page at a time on a `maintenance: 1` queue) and the code comments show the author knew.

PR #2016, opened 2026-09-12
docs(conversations): record how the server size pin moves through a stack

jhgaylor:
Clarifies when the ConversationServer size limit should be lowered: in the PR that lands the reduction on main, with separate guidance for a stack still in flight. It also records that #1565’s earlier mid-stack lowering was backed out before merge.

The diff changes only the test module’s documentation. It preserves main’s `@pin 2549`, the existing test assertion, and the newer campaign/headroom explanation. This addresses the earlier review’s concern about accidentally raising the limit.

Rebased onto main at `859f60c75989f46c33482f7e11a6a0f9ed0583cc`, resolving the overlapping documentation edits. The server at that base is 2,518 lines.

Validation on the rebased change:
- Existing size-limit test run directly through ExUnit on Elixir 1.19.2 / OTP 28.3: 1 test, 0 failures.
- Elixir formatting check and `git diff --check`: passed.
- Local `mix precommit` was attempted but could not proceed because the isolated worktree has no dependencies installed. Fresh PR CI on `cedef0a5` passed all expected jobs, including all six test partitions, coverage, static analysis, release/API contract checks, `CI required`, and secret scanning.

Refs #1565, #1891, #2032.

### DIFF
--- a/apps/fountain/test/fountain/conversations/conversation_server_size_test.exs
+++ b/apps/fountain/test/fountain/conversations/conversation_server_size_test.exs
@@ -6,9 +6,9 @@ defmodule Fountain.Conversations.ConversationServerSizeTest do
 
   Tracker #1369 refactors the server by subtraction: each sub-issue moves a
   function family into a module under `Fountain.Conversations.*` and lowers
-  `@pin` to the file's new length in the same PR. The pin is the file's line
-  count on `main` at the last move, so a change that makes the file longer
-  fails here and has to say why.
+  `@pin` to the file's new length in the PR that lands on `main` (see below
+  for a stack). The pin is the file's line count on `main` at the last move,
+  so a change that makes the file longer fails here and has to say why.
 
   The shape is the docs-style allowlist that only shrinks (#911): the number
   is not a target, it is a record of where the file is, and the only edit it
@@ -18,8 +18,9 @@ defmodule Fountain.Conversations.ConversationServerSizeTest do
   a number measured against a branch is stale the moment any link below it
   grows the file, which is what review rounds do. #1565 lowered it mid-stack to
   the tip's exact length, left zero headroom, and went red two rounds later
-  when a fix four PRs down added lines. Land the shrink first, then lower the
-  pin in a follow-up, when the number has stopped moving.
+  when a fix four PRs down added lines. That lowering was backed out before
+  merge. Land the shrink first, then lower the pin in a follow-up, when the
+  number has stopped moving.
 
   That is what this pin is mid-way through. #1749 and #1751 each lowered it
   to their own tip's exact length while the #1766/#1767 campaign was in


BinaryBourbon (changes requested):
**REQUEST_CHANGES** — the pin was right when it was measured, but `main` has moved past it. Landing this now **raises** the ratchet by 43 lines.

## The arithmetic

| Ref | `@pin` | actual `conversation_server.ex` | slack |
|---|---|---|---|
| merge-base `d8a3e07` (what this PR measured against) | 2762 | **2689** | 73 |
| **`origin/main` today (`97bcf1a6`)** | **2646** | **2646** | **0** |
| this PR head `6ff03dc` | 2689 | 2689 | 0 |

The measurement was correct when taken: at the merge-base the file really was 2,689 lines and the pin really was 2,762, so `2762 -> 2689` was a genuine lowering.

Since this PR was opened, two links of the #1767 stack landed and each lowered the pin in the PR that landed, which is exactly what the docstring prescribes:

- `94f879c4` #1749 — pin `2762 -> 2652`, file 2652
- `4f0c040c` #1751 — pin `2652 -> 2646`, file 2646

So `main` is already at 2646 with zero slack. This PR's stated goal — "lower the pin to the measured size on `main` so the guard preserves the reduction" — has already been achieved on `main`, and to a tighter number than the one proposed here.

## Why this is blocking rather than merely redundant

Against current `main`, `@pin 2689` is a **raise** of 2646 -> 2689. That is the one edit the test's own moduledoc forbids:

> the only edit it accepts is downward. Lower it when you move something out; never raise it.

And the raise is **silent**. I verified both halves:

1. The merge conflicts on that line (base 2762 / main 2646 / PR 2689), so a human resolves it by hand:
   ```
   CONFLICT (content): Merge conflict in .../conversation_server_size_test.exs
       @pin 2646
       @pin 2689
   ```
   Resolving toward the PR — the natural reading of "this PR is the one that changes the pin" — loosens the ratchet by 43 lines.

2. Nothing catches it afterward. The assertion is `count <= @pin`, so with `main`'s 2,646-line file and a pin of 2689 it stays green. I applied `@pin 2689` to `origin/main`'s tree and ran the test: **1 test, 0 failures.** 43 lines of slack would sit there un-flagged until someone next measured by hand.

That is precisely the slack this ratchet exists to remove, so it is worth not merging by accident.

## What I checked

- **Pin test on the PR head** — passes exactly (`2689 == 2689`): `mix test test/fountain/conversations/conversation_server_size_test.exs` -> `1 test, 0 failures`.
- **Nothing smuggled in** — the diff is one file, +7/-6, confined to the moduledoc and the `@pin` literal. No production code, no other pin, nothing raised elsewhere. Clean.
- **`gh pr checks 2016`** — all green (6 partitions, static, coverage, release/contract, SDKs, Swift). Green here proves only that the branch is self-consistent; it cannot see that `main` has since gone lower.
- **The new docstring sentence is factually correct.** I confirmed #1891 (the #1565 merge) landed with the pin still at 2762 against a 2,689-line file — slack 73. The mid-stack lowering was indeed backed out before merge, as the added line says.

## Suggested resolution

The prose changes are worth keeping; only the number is stale.

1. Rebase onto `main`.
2. **Drop the `@pin` hunk entirely** — leave it at 2646.
3. Keep the two moduledoc edits: "in the PR that lands on `main`" matches what #1749 and #1751 actually did, and "That lowering was backed out before merge" is accurate.

That leaves a docs-only change to the test, which is the part that still has value.

## One non-blocking note for whoever next touches the pin

This file does not shrink monotonically between refactor moves — it went 2774 (#1764) -> 2555 (#1827) -> 2689 (#1891) -> 2652 -> 2646. `main` now sits at zero headroom while the rest of the #1767 stack is still landing, which is structurally the same position the moduledoc blames for #1565 going red two rounds later. Not this PR's problem, and not a reason to add slack pre-emptively, but worth knowing if a later link in that stack turns this test red: the fix is to move code out, not to raise the number.

jhgaylor:
Cross-reference: the same stale-pin arithmetic bites the #1766/#1767 campaign from the other direction — merging its chain tip onto current `main` lands the server at 2673 against the pin of 2646, with every PR in it green. Filed as #2032. Whatever correction that needs may change the number this PR should settle on, so it is worth landing them in the same pass.

jhgaylor:
Pushed the change the review asked for: rebased onto `main` and **dropped the `@pin` hunk entirely**, keeping both moduledoc edits.

Re-derived the arithmetic against `origin/main` as it stands now (`a4822662`, two commits past the `97bcf1a6` the review measured) — the conclusion is unchanged:

| Ref | `@pin` | `conversation_server.ex` | slack |
|---|---|---|---|
| merge-base `d8a3e07` (what this PR measured) | 2762 | 2689 | 73 |
| `origin/main` today (`a4822662`) | **2646** | **2646** | **0** |
| old PR head `6ff03dc` | 2689 | 2689 | 0 |

So `@pin 2689` against today's `main` would have been a 43-line **raise**, not a lowering. The rebase conflicted on exactly that line (base 2762 / main 2646 / PR 2689) as predicted; resolved by taking `main`'s 2646.

The branch is now `ef6d30eb` and its diff against `main` is one file, +6/-5, moduledoc only — `git diff origin/main HEAD` touches no `@pin` line.

Verification:

- `mix test test/fountain/conversations/conversation_server_size_test.exs` from `apps/fountain` -> `1 test, 0 failures`.
- Checked the guard is not passing vacuously, since `count <= @pin` passes just as happily with slack: at `@pin 2645` it fails with `conversation_server.ex is 2646 lines, over the pin of 2645`. So the file is exactly at the pin and the ratchet is tight.
- Re-read both moduledoc edits after the rebase — neither quotes a line count, so they read correctly against 2646. "in the PR that lands on `main`" describes what #1749 and #1751 did, and "That lowering was backed out before merge" matches #1891.

Title and body updated, since the PR no longer lowers anything.

Not dismissing the changes-requested review or self-approving — I made the change, so the re-review is not mine. Leaving it for a human or a fresh reviewer.

The pin's zero headroom against the rest of the #1766/#1767 campaign is #2032 and is untouched here.

BinaryBourbon (approved):
Approved after independently reviewing cedef0a57bacc746f51acf343d44c4479f91103f against current main (859f60c75989f46c33482f7e11a6a0f9ed0583cc).

The previous blocking requests are resolved: this is rebased onto main, the stale pin change is gone, and both requested moduledoc edits remain. The complete diff is documentation-only; main's @pin 2549, the assertion, and the newer campaign/headroom explanation are preserved. The server remains 2,518 lines, so this PR introduces no additional slack or behavioral change.

Validation: independently ran the existing size test through ExUnit (1 test, 0 failures), checked Elixir formatting and git diff --check, and verified all applicable current CI checks pass, including all six test partitions and CI required. No remaining blocking findings.

Issue #2018, opened 2026-09-12
tracker: multiple inference credential sets per account


jhgaylor:
Two customers want the same sentence and need different things.

**The reseller.** One account is really N end customers. He needs each end
customer's turns on that customer's inference credential, never another's.

**The quota-juggler.** One tenant, several Claude subscriptions, exhausting
quota on one and wanting the next agent to run on another.

Today `inference_credentials` is one row per user with four fixed ciphertext
columns and a `unique_constraint(:user_id)`. `InferenceCredentials.select/4`
is the whole selection rule, and it reads that one row.

## Direction

The reseller's answer is **principals**, not credential sets. ADR 0044 already
wrote the objection down -- "one shared service account is not isolation" --
and `Principals.billing_subject_id/1` already gives him one invoice with
per-end-user usage attribution. One gap blocks it (item 10 below).

The juggler's answer is **credential sets**: `inference_credentials` becomes
1:N, an agent names its default and a launch may override it.

## Relationship to ADR 0052

ADR 0052 (user-owned ChatGPT grants, #2010) is merged as Proposed, and
#2011-#2015 + #2017 build its sections 1, 3 and 6. Sections **4 (selection)
and 5 (sandbox identity)** are unbuilt, and that is exactly the ground this
work stands on. This tracker builds section 4's selection plumbing in the
shape 0052 asks for, so the ChatGPT work lands on top of it rather than
colliding, and amends 0052's "no per-agent preference in the first version".

## Stack

1. ADR 0053 + index refresh
2. `select/4` returns a source struct, not a bare atom (pure refactor)
3. Fix: an env/vault secret shadowing a static inference credential resolves
   the source to `:own` (P1, mis-bills on prod today)
4. Inference credential pairs move off the shared `.env` into process env
5. `inference_credentials` gains `name` + `is_default`; 1:N
6. `agents.inference_credential_id` + `allowed_inference_credential_ids`
7. `conversations.inference_credential_id` per-launch override
8. API: CRUD, OpenAPI, contract, TS SDK
9. Console
10. An owner may set an inference credential on a principal it owns
11. Docs, CHANGELOG, SDK bump

## The bug in item 3

`PlatformInference.gate/3` and `select/4` both ask `has_own?`, which reads only
the `inference_credentials` table. A vault secret named `ANTHROPIC_API_KEY`
is invisible to it, but wins in the sandbox env (`egress.ex:223-226`, and
[...]

PR #2019, opened 2026-09-12
docs: propose inference credential sets (ADR 0053)

jhgaylor:
Tenants need to assign different inference credentials to agents while preserving a shared workspace. ADR 0053 proposes named credential sets, agent defaults and launch overrides. Businesses isolating end customers use principals, with an owner-authenticated credential-management route.

This PR is the shared selection and runtime-auth contract for ADRs 0052 and 0053. It keeps the existing grant foundation stack (#2011–#2015 and #2017) and defines the remaining work once:

- One runtime-aware resolver supplies source identity, emitted auth inputs, admission and billing. It accounts for environment/vault precedence, OAuth/API conflicts and runtime aliases, including OpenCode's Google key name.
- Managed custody covers user and platform ChatGPT credentials, including static workspace tokens. Ordinary tenant overrides remain supported; selected subscriptions never silently fall back to API/platform inference.
- Peers persist a non-secret resolved source binding across resume and reattachment. Set IDs alone cannot distinguish key edits or grant replacements. Defaults and source changes cannot silently switch an existing peer.
- Process-only inputs remove shared `.env` persistence, including cleanup on existing machines. Mutable runtime auth state requires isolation. A temporary Codex compatibility guard must atomically include preparing peers before auth-file writes.
- Admission and provisioning validate source revisions and eligibility. Changed inputs require renewed checks before a new binding, including platform ceiling checks when applicable.

The ADR's **Implementation order and acceptance** section specifies dependencies and required proof: shared selection and billing fix; shared auth binding; credential-set consumers; independent broker authorization/draining; user-grant consumers; principal management. User linking does not wait for the entire credential-set console or principal-management stack. Feature activation remains subject to a separate release plan.

Validation: `okf validate decisions` and `okf lint decisions` pass with zero errors and six pre-existing actor-convention warnings, none on ADR 0053. Regenerating the decisions index produces no diff. `git diff --check` passes.

Docs only; the proposed behavior remains unbuilt. Tracker #2018. This explicitly amends ADR 0052's initial restriction on per-agent preference for credential sets while preserving its user-level ChatGPT preference and managed-grant requirements.


BinaryBourbon (approved):
**Approved.** I read the ADR in full and checked its factual claims against `main` rather than taking them on trust.

## What I verified

Every load-bearing assertion about today's code holds:

- `egress.ex:223-226` is the gate-3 comment, and it does say a tenant's own secret of the same name wins over the inference credential.
- The override is real, not just documented: `SpriteEnv.build/4` appends `env.env_vars` and then the merged `secrets` **after** `runtime_module.default_env(agent, env_credentials)`, and the module doc's own #1674 note confirms later entries win.
- `PlatformInference.gate/3` (arity 3, user/model/runtime) calls `InferenceCredentials.has_own?/2`, which is `is_nil(missing_for_model/2)` — the credential row and nothing else. The described mis-bill follows from that.
- `Credential` is one row per user with `unique_constraint(:user_id)` and four ciphertext columns.
- `Conversations.Identity.@process_only` exists with the reasoning decision 4 quotes.
- `Principals.billing_subject_id/1` and `FountainWeb.Plugs.RequireFullScope` exist, so decision 7's gap is stated accurately.

## The ADR rules

- Frontmatter is complete and shaped like 0052: `adr_status: "Proposed"`, `stale_after: 2026-10-12`, `generated.by: human:jhgaylor`. `verified` is correctly **absent** — it is the "checked against the code" marker, and nothing here is built.
- `okf validate decisions` is `"valid": true` with 6 warnings, all pre-existing actor-convention warnings on 0049/0051 and none on 0053.
- The index row is present, matches the frontmatter description, and carries `verified: no`.
- Unbuilt behavior is never described as existing. The status block opens with "Proposed; none of the behavior below is built", the implementation table is written as deliverables and required proof, and the closing paragraph explicitly withholds activation authorization. This is the rule the 2026-07 audit exists for and the ADR respects it.

The decision content is sound. Decision 6's refusal to put the set in the sandbox identity tuple is the right call for the stated use case (one home, two subscriptions), and the alternatives section correctly rejects the vault-per-set shortcut on `vault_id` being *in* that tuple. Decision 7 routing the reseller to principals rather than sets, with ADR 0044 as the reason, keeps tenancy from being built twice.

## One non-blocking ask

**0052 gets no back-reference.** Decision 3 amends ADR 0052 decision 4's "no per-agent preference in the first version", but `0052-user-owned-chatgpt-grants.md` is not touched. The repo has an established convention for this — 0005, 0008 and 0025 each carry an "Amended by NNNN" line in the status block *and* in the frontmatter `description` (which is what lands in the index row). 0052 is Proposed and actively being built right now (#2011–#2015, #2017), so a reader who opens it will follow decision 4 as written without knowing it moved.

One line in 0052's status block and a clause on its description and index row would close it. Worth doing in a follow-up rather than respinning this stack's base.

## A note for later in the campaign

#2022, #2023 and #2025 build decision 2 and decision 5. Once they land, 0053's "**Status:** Proposed; none of the behavior below is built" is stale. The error runs in the safe direction — it understates what exists rather than claiming something unbuilt — so it is not the failure the ADR rule targets, but the campaign's docs PR should correct the status block with an explicit built/not-built accounting rather than leaving it as written.

PR #2020, opened 2026-09-12
chore: correct CI docs and the ee/test classifier gap after #1413

jhgaylor:
Follow-ups #1413 left behind. Docs and one classifier gap; no workflow changes.

**`CLAUDE.md` CI table** still documented an `sdk-clients` job that no longer
exists. Replaced with a row per job that does. Two of those rows carry facts a
reader cannot get from the workflow itself: `swift-sdk` runs no
`sdk/conformance/lint.py`, and a green `sdk-checks` on a docs-only or
reused-tree plan means *correctly skipped*, not *SDKs tested*.

**ADR 0050** sized the merge queue against a full CI run of 19 jobs. That was
true on 2026-09-10, so it stays as a dated measurement rather than being
rewritten. A parenthetical records the current 26 (mixed PR) / 27 (merge group)
and points at `scripts/ci/require-checks.py`, which already carries the live
count. The bigger number strengthens `max_entries_to_build: 1` rather than
changing it.

**`ee/test/` was missing from the classifier's `UNRELATED_PREFIXES`** while
`apps/fountain/test/`, `apps/fountain_buzz/test/` and
`apps/fountain_support/test/` were all present, so an ee-test-only PR selected
all four SDKs. Fail-open, so it ran too much and never too little.

The fix must stay `ee/test/` and must never widen to `ee/`:
`ee/lib/fountain_web/` can move the OpenAPI document that every SDK is
generated against, so it has to keep fanning out. The new test asserts both
halves and a comment in the source says why.

## Verification

- 57 `scripts/ci` tests pass (56 on main).
- Mutation-tested the new guard both ways: dropping `ee/test/` fails 2 tests,
  and widening it to `ee/` also fails 2 — so the tempting over-broad "fix" is
  caught rather than silently breaking OpenAPI fan-out.
- `okf validate decisions` valid, same 6 pre-existing warnings as main; 0050
  is not among them.
- `scripts/conflict-markers.py` clean.
- `docs-style.py` findings are all pre-existing backlog under `docs/`, which
  this PR does not touch.

Deliberately not changed: the `OWNED_FILES` doc→SDK mappings. A
`docs/swift-sdk.md` edit selecting the macos-15 leg is by design and consistent
with the other three SDKs.

https://claude.ai/code/session_01HD7jxhmpwc8fAbWAHby4PE


BinaryBourbon (approved):
**APPROVE** — the CI table rows, the ADR job counts and the `ee/test/` classifier gap all check out against `ci.yml` and the SDK metadata at this HEAD; I re-ran the suite and both mutations myself rather than taking the PR body's word for them. Four documentation nits below, none blocking.

## What I verified independently

**Every new CLAUDE.md row, against `.github/workflows/ci.yml` at 709fc39.**

| Claim | Result |
|---|---|
| `elixir-sdk` ×2, `1.15.8 / OTP 26.2.5.21` + pinned current | true — `matrix.include` is exactly those two pairs; the second is `1.19.2 / 28.3`, which is `.tool-versions` (`erlang 28.3`, `elixir 1.19.2-otp-28`). `sdk/elixir/mix.exs` declares `elixir: "~> 1.15"`, so 1.15.8 really is the declared minimum series |
| `python-sdk` ×2, `3.9` and `3.13` as declared boundaries | true — `matrix.python: ["3.9", "3.13"]`; `sdk/python/pyproject.toml` has `requires-python = ">=3.9"` and classifiers through 3.13 |
| "a built wheel installed into a fresh venv outside the source tree" | true — `verify_wheel.py` builds a venv under `tempfile.TemporaryDirectory`, installs `--no-index --no-deps`, and refuses to proceed unless `fountain.__file__` resolves under `sys.prefix` |
| `typescript-sdk` ×2, minimum Node from `engines.node` (20.19.0) | true — `matrix.node: ["20.19.0", "24"]`; `sdk/typescript/package.json` `engines.node` is `">=20.19"` |
| "the packed tarball installed into a throwaway consumer project" | true — `verify-package.mjs` does `npm pack` into a `mkdtempSync` dir and installs into a scratch `consumer/` |
| `swift-sdk` ×2 on `ubuntu-24.04` and `macos-15` | true |
| "It runs no `sdk/conformance/lint.py`" | true — the other three jobs each have that step; `swift-sdk` has no such step. `scripts/ci/README.md` agrees ("These three extracted jobs lint fixtures…") |
| `changes` "does not run on `push`" | true — `if:` is `pull_request \|\| merge_group \|\| workflow_dispatch` |
| `changes` fail-open on unreadable diff / unregistered path / shared file | true — `from_git` returns `LANGUAGES` on a non-40-hex base, a non-NUL-terminated result and `OSError/CalledProcessError/UnicodeDecodeError`; `classify` returns `LANGUAGES` for any path not owned and not in `UNRELATED_*` |
| `sdk-checks` is an aggregate over the four SDK jobs, green can mean "skipped" | true — `needs` is the four plus the two probes; `gate.py --sdk` expects `skipped` for every SDK job under `reuse` |

Nothing the table says about a job is wrong. No job listed in the table has ceased to exist, and `sdk-clients` no longer appears anywhere in the tree.

**`ee/test/` change (`scripts/ci/sdk_changes.py:29`).** `ee/test/` exists and contains only `.exs` test files plus `test_helper.exs` — no fixture consumed by `scripts/sdk-contract/build.sh`, which boots the app from the umbrella root and projects `dist/openapi.json`. The stated invariant is real, not hand-waving: `ee/lib/fountain_web/controllers/credits_api_controller.ex:18` is `use OpenApiSpex.ControllerSpecs`, so that tree genuinely contributes operations to the document every SDK is generated against. Widening the prefix to `ee/` would have swallowed it. Nothing else under `ee/` is newly swallowed — the prefix matches only `ee/test/`.

**The suite and both mutations, run in my own scratch copy of each tree:**

- HEAD: `Ran 57 tests … OK`. BASE (f249e31): `Ran 56 tests … OK`. Matches the PR body.
- Mutation A, delete `"ee/test/",`: `FAILED (failures=2)` — `test_ee_lib_fans_out_but_ee_tests_do_not` and `test_unrelated_paths_and_mixed_sdk_changes`.
- Mutation B, widen to `"ee/"`: `FAILED (failures=2)` — `test_ee_lib_fans_out_but_ee_tests_do_not` and `test_public_contract_shared_fixtures_and_uncertainty_fan_out`.

Both "fails 2 tests" claims are exactly right. One honest qualification the body does not make: in the widening direction the new test is not the only thing standing there — the pre-existing `test_public_contract_shared_fixtures_and_uncertainty_fan_out` already asserted an `ee/lib/fountain_web/…` path fans out. The new test adds the *drop* direction (which only the amended `test_unrelated_paths_and_mixed_sdk_changes` otherwise covers) and makes the reason explicit.

**ADR 0050 numbers.** I counted the legs myself from `ci.yml` at HEAD, by event:

- mixed PR (code + docs, all SDKs selected): `changes` 1 + `test` 6 + `coverage` 1 + `elixir-static` 1 + `release-and-contract` 1 + `elixir-sdk` 2 + `python-sdk` 2 + `typescript-sdk` 2 + `swift-sdk` 2 + `cli-plugins` 1 + `docs-prose` 1 + `core-distribution` 1 + `compose-fresh-clone` 1 + `compose-pinned-image-boot` 1 + `workflow-checks` 1 + `sdk-checks` 1 + `gate` 1 = **26** (`already-tested` is skipped on `pull_request`).
- merge group: the same plus `already-tested` = **27**.

Both figures are right, and `scripts/ci/require-checks.py` does carry them (the `MERGE_QUEUE` comment: "A full mixed PR runs 26 jobs; a merge group runs 27 because both probes run"), as does `scripts/ci/README.md:75`. Keeping "was 19 … when this was decided" as a dated measurement is the right call.

**OKF.** `okf validate decisions` at HEAD: `"valid": true`, 6 warnings, all `okf/trust/actor-convention` on 0021/0027/0030/0033/0051 — 0050 is not among them. Re-running `scripts/decisions-index.sh` over the HEAD tree produces a byte-identical `decisions/index.md`, so this PR correctly skips an index refresh: the row is built from title/status/verified/description and none of those moved. No `stale_after` is owed either — `decisions/0001-template.md:35` requires it only "while anything described is unbuilt", and 0050 is Accepted and built.

## Nits (non-blocking)

1. `CLAUDE.md:424` — the `changes` row describes only the SDK half of the job. Its most consequential output is `docs_only`, which gates `test` ×6, `coverage`, `elixir-static`, `release-and-contract`, `cli-plugins`, `core-distribution` and both `compose-*` jobs, and it comes from the inline shell step, not `sdk_changes.py`. "Classifies the diff (`scripts/ci/sdk_changes.py`)" attributes the whole classification to the SDK script. The job also emits `tree`, which is the `tested-tree` evidence.
2. `CLAUDE.md:429` — `cli-plugins` also runs `node --test deployed/test/*.test.mjs` ("Deployed-instance runner tests"). `scripts/ci/README.md:83` names it; the new row does not, so a red `cli-plugins` from that step points a reader at Go and Python only.
3. `CLAUDE.md:430` — "Green on a docs-only … plan means *correctly skipped*" is not universally true. `docs/sdk.md` and `docs/{elixir,python,swift}-sdk.md` are `OWNED_FILES`, so a docs-only PR touching one runs that SDK for real; `gate.py:70` says so in as many words ("SDK docs can select a language even when the server plan is docs-only"). The practical warning still holds, but "may mean correctly skipped" would be accurate.
4. The table still omits six jobs that exist: `already-tested`, `core-distribution`, `compose-fresh-clone`, `compose-pinned-image-boot`, `docs` and `workflow-checks`. All six were missing before this PR, so nothing regressed — but given the table's stated purpose ("the job name in a red build tells you which toolchain to go and look at") and that this PR is the one converting it to a row-per-job, they were the natural thing to finish.

Also worth saying plainly: the PR body calls `require-checks.py` "the live count". It is a hand-written comment and a test docstring, not a figure derived from `ci.yml`, so it can drift exactly the way the ADR just did. The ADR's own wording ("carries the current count") is fine; the body's is generous. And `verified:` on 0050 still reads `2026-09-10`, now vouching for a body that carries a 2026-09-12 measurement — `okf` does not care, but a re-stamp would have been tidier.

No CHANGELOG entry is owed (CI/docs only, no user-visible behavior), no docs page is affected (`CLAUDE.md` and `decisions/` are both unpublished at `/docs`), and `scripts/ci/README.md` is already consistent with every number this PR writes.

## Checked / not checked

Checked: `ci.yml` at HEAD parsed and every job's `if:`/matrix enumerated; `sdk/{elixir/mix.exs, python/pyproject.toml, typescript/package.json}`; `verify_wheel.py` and `verify-package.mjs`; `gate.py` plan expectations; `ee/` tree contents and the OpenApiSpex usage in `ee/lib/fountain_web`; the `scripts/ci` suite at BASE and HEAD and both mutations, in isolated copies under my scratchpad; `okf validate decisions`; `scripts/decisions-index.sh` re-run for drift; a repo-wide grep for surviving `sdk-clients` references.

Not checked: I did not run CI, so the job counts are a static leg count from the workflow, not an observed run; I did not run the Swift, Node, Python or Elixir SDK jobs themselves; and I did not verify `.tool-versions` against what `mise install` actually resolves on a runner.

PR #2022, opened 2026-09-12
refactor: select/4 returns a credential source, not an atom

jhgaylor:
**Base: #2019.** ADR 0053 decision 2. Tracker #2018.

A pure refactor with one deliberate vocabulary split. No turn changes how it
is billed, gated or stamped.

## Why now, and why not inside the ChatGPT stack

`InferenceCredentials.select/4` answered `:own | :platform` — one bit for two
questions: *who pays*, and *what actually served the turn*. ADR 0052 decision
4 requires the second ("extend selection to return source identity") and is
unbuilt; #2011-#2015 and #2017 build decisions 1, 3 and 6 and do not touch
`inference_credentials.ex`, `platform_inference.ex`, `sprite_env.ex` or
`turn_machine.ex`. ADR 0053 needs the same thing for a different reason — a
tenant secret that serves a turn the deployment is billed for. Cutting the
struct once, here, means the grant work adds `grant_id` and `generation` to
it instead of landing a second selection refactor beside this stack.

## What it is

```elixir
%Fountain.InferenceCredentials.Source{origin: :own | :platform, scope: ...}
```

`origin` is the billing question in the vocabulary the ledger already uses,
so `TurnMachine.with_inference/2` stamps and `TurnMachine.gate/2` checks the
daily ceiling on exactly the turns they did before. `scope` says where the
value came from: `:credential`, `:platform`, `:none`, `:missing`.

## The one thing that is not a rename

`{:error, :no_credential}` and "this provider needs no credential" both
collapsed into `:own` at `SpriteEnv.select_inference/3`. They are now
`:missing` and `:none`.

Both keep `origin: :own`, so neither is platform-paid and neither turn's
usage map moves — but they are opposite answers to "is anything wrong here".
A local model needs no key; a conversation with no key for a provider that
requires one will fail inside the sandbox. Only the source can tell a surface
which, and decision 5 needs the distinction to report where a turn ran.

## No `kind` field

A field naming the credential atom that served the turn would be wrong for an
account holding both an Anthropic key and an OAuth token: `Runtimes.Claude`
prefers `CLAUDE_CODE_OAUTH_TOKEN` and deliberately never exports the API key
beside it, while `Runtimes.OpenCode` reads only the API key for the same
provider. No derivation in this context is right for both, so the field waits
for a reader that needs it and a derivation that matches the runtime. The
module doc says so, and the ADR does too.

## Verification

- `test/fountain/platform_inference_test.exs`,
  `test/fountain/platform_chatgpt_test.exs`,
  `test/fountain/conversations/conversation_server_platform_inference_test.exs`,
  `ee/test/fountain/credits_chatgpt_inference_test.exs`,
  `ee/test/fountain/credits_inference_test.exs` — 94 tests, 0 failures.
- New `sprite_env_inference_test.exs` (6 tests) over `select_inference/3`, one
  per scope. `async: false` and in its own module rather than beside
  `SpriteEnvTest`: the platform keys are global application env, so a module
  that writes them races every module that reads them (#1214).
- **Revert-checked**: collapsing `Source.missing/0` back into `Source.none/0`
  fails exactly the `:missing` test and nothing else, so the distinction is
  actually asserted.
- `mix credo --strict` clean over 849 files. `mix format` (under mise, from
  `apps/fountain`) clean.
- Swept `apps`, `ee`, `sdk` and `cli` for consumers of the old tuple shape —
  none remain.

Full suite is queued behind another session's stack on this box; I will post
the result on this PR.

## Stack

| # | PR | What |
|---|---|---|
| 2 | **this** | `select/4` returns a source struct |
| 3 | | env/vault secret resolves the source to `:own` (P1, mis-bills on prod) |
| 4 | | inference pairs move off the shared `.env` |
| 5 | | `inference_credentials` gains `name` + `is_default`; 1:N |
| 6 | | `agents.inference_credential_id` + allowlist |
| 7 | | `conversations.inference_credential_id` per-launch override |
| 8 | | API, OpenAPI, contract, TS SDK |
| 9 | | console |
| 10 | | an owner may write a credential on a principal it owns |
| 11 | | docs, CHANGELOG, SDK bump |

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -504,7 +504,7 @@ defmodule Fountain.Conversations.ConversationServer do
       # What the runtime's default_env/2 is handed (gate 3): the real
       # credentials, or placeholders when the conversation is brokered.
       env_credentials: %{},
-      # #1388; see `SpriteEnv.select_inference/2`, which decides both.
+      # #1388, ADR 0053: the `InferenceCredentials.Source` that ran this one.
       inference_source: nil,
       inference_model: nil,
       broker: nil,
--- a/apps/fountain/lib/fountain/conversations/sprite_env.ex
+++ b/apps/fountain/lib/fountain/conversations/sprite_env.ex
@@ -43,15 +43,23 @@ defmodule Fountain.Conversations.SpriteEnv do
   runtime's `default_env/2`, the redaction register — is untouched by the
   feature existing at all.
 
+  The first element is a `Fountain.InferenceCredentials.Source` (ADR 0053
+  decision 2), which the server holds for the conversation's lifetime and the
+  turn's usage stamp is derived from.
+
   `:no_credential` keeps the behaviour that predates platform keys: the
   conversation provisions anyway, and the provider's own auth failure lands
-  on the transcript rather than a refusal invented here.
+  on the transcript rather than a refusal invented here. Nothing served the
+  turn and the deployment is not paying for it, so the source is
+  `Source.missing/0` — `origin: :own`, exactly what this case resolved to
+  before the struct existed.
 
   `runtime` is the conversation's own (`conv.runtime`), which is what the
   sandbox is dispatched on and can differ from the agent's after an edit;
   nil falls back to the agent's.
   """
-  @spec select_inference(map() | nil, map(), String.t() | nil) :: {:own | :platform, map()}
+  @spec select_inference(map() | nil, map(), String.t() | nil) ::
+          {InferenceCredentials.Source.t(), map()}
   def select_inference(agent, own_creds, runtime \\ nil) do
     brokered? = (agent && Fountain.Broker.enabled_for?(agent.user_id)) || false
     runtime = runtime || (agent && agent.runtime)
@@ -60,7 +68,7 @@ defmodule Fountain.Conversations.SpriteEnv do
            brokered: brokered?
          ) do
       {:ok, source, creds} -> {source, creds}
-      {:error, :no_credential} -> {:own, own_creds}
+      {:error, :no_credential} -> {InferenceCredentials.Source.missing(), own_creds}
     end
   end
 
--- a/apps/fountain/lib/fountain/conversations/turn_machine.ex
+++ b/apps/fountain/lib/fountain/conversations/turn_machine.ex
@@ -51,6 +51,7 @@ defmodule Fountain.Conversations.TurnMachine do
 
   alias Fountain.{Agents, Conversations}
   alias Fountain.Conversations.{Conversation, Labels}
+  alias Fountain.InferenceCredentials.Source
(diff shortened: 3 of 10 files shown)


BinaryBourbon (approved):
**Approved.** Reviewed as its own patch against `stack/2018-1-adr`.

## The billing invariant holds

This is the only thing that matters in a refactor that touches the usage stamp, so I checked it directly rather than trusting the description:

- `TurnMachine.with_inference/2` matched `:platform` / `:own`; it now matches `%Source{origin: :platform}` / `%Source{origin: :own}`. Every constructor sets `origin` to exactly what the old atom was at that call site — `credential/0`, `none/0` and `missing/0` are `:own`, `platform/0` is `:platform` — so no turn's stamp moves.
- `TurnMachine.gate/2` went from `inference == :platform` to `Source.platform?(inference)`, which is `true` only for `%Source{origin: :platform}` and `false` for everything including `nil`. Identical to the old comparison, including the `nil` case.
- `CreditPricer` selects on `usage ->> 'inference' = 'platform'`, a string written by `with_inference/2`, so the ledger is downstream of a stamp that did not change.

## The consumer sweep is complete

I re-ran it rather than taking the claim. `inference_source` has exactly three readers in `lib` — `TurnMachine.ctx/2`, and the two `TurnMachine.gate/2` call sites in `ConversationServer` — all of which now carry a struct into functions that were updated in the same patch. The only other `InferenceCredentials.select/4` caller is `start_live.ex:115`, which matches `{:error, :no_credential}` and nothing else; that clause is untouched, so the banner behaves identically.

## The `:none` / `:missing` split

This is the one real behavior distinction in the patch and it is the right one. `select/4`'s `accepted == []` arm and `SpriteEnv.select_inference/3`'s `{:error, :no_credential}` arm both collapsed to `:own` on `main`, and they are opposite answers: a provider that needs no key versus a provider that needs one nobody has. Both keep `origin: :own`, so nothing about billing or the ceiling moves, and the new `sprite_env_inference_test.exs` asserts each scope individually. The revert check described (collapsing `missing/0` into `none/0` fails exactly the `:missing` test) is the right shape of proof for a distinction that would otherwise be untested vocabulary.

`async: false` in its own module, with the #1214 reasoning in the moduledoc, is correct — the platform keys are global application env and a module writing them races every module reading them.

## The `kind` field

Leaving it out is the right call and the moduledoc gives the actual reason: `Managoat.Runtimes.Claude` prefers `CLAUDE_CODE_OAUTH_TOKEN` and never exports the API key beside it, while `Managoat.Runtimes.OpenCode` reads only the API key for the same provider, so any derivation in this context states the wrong credential for an account holding both. "A field arrives with its reader" is the correct discipline for a struct that ADR 0052 will extend with `grant_id` and `generation`.

## Minor

The struct has `@enforce_keys [:origin, :scope]` with four constructors and no public struct literal in `lib`, so the vocabulary cannot drift by accident. Nothing to change.

CI is green across all six partitions plus static analysis and the release/contract job.

PR #1941, opened 2026-09-11
fix: stamp platform inference from the selected credential source

jhgaylor:
A deployment using only its ChatGPT grant selected a platform credential but omitted the inference stamp because no platform API key was configured. The pricer consequently wrote no inference debit, and daily platform spend missed the turn. Stamp platform usage from the credential source selected for the turn; keep tenant-owned usage behavior unchanged.

Closes #1765. Three files, +137/-18. This includes the turn-start stamp and completed-turn usage through the same function.

Validation: a real local active grant with no platform API keys reproduces both missing turn metadata and a zero-debit pricer pass on the parent. With the fix, all 64 focused turn tests and 24 billing tests pass. The billing regression verifies a debit, daily spend, idempotency, and no inference charge for the tenant's own credential. This is local grant-only verification; ADR 0047's separate recorded production measurement is unchanged.

Full local precommit passes with four VM schedulers: 5,366 core tests plus 6 doctests, 144 Buzz tests, and 33 Support tests, with no failures.

PR #2023, opened 2026-09-12
fix: a tenant secret naming a credential is not platform inference

jhgaylor:
**Base: #2022.** ADR 0053 decision 5. Tracker #2018.

**This one is live on production.**

## The bug

A vault or environment secret named `ANTHROPIC_API_KEY` wins over the
account's inference credential in the sandbox. `Conversations.Egress` says so
at the gate-3 split (`egress.ex:223-226`) and `docs/concepts/secrets.md:212`
publishes the rule.

Selection could not see it. `InferenceCredentials.select/4` reads the
decrypted `inference_credentials` row and nothing else, so on a deployment
holding platform keys — the hosted one, since 2026-09-03 — a tenant using the
documented override got:

- the platform key selected,
- the turn stamped `"platform"` by `TurnMachine.with_inference/2`,
- the tokens priced against **their own credits** by `CreditPricer`,
- the spend counted against `PLATFORM_INFERENCE_DAILY_CENTS`,

while their own secret served every one of those turns.

#1941 (merged earlier today) fixed an adjacent facet — stamping a platform
turn when the deployment holds no platform API key. It does not reach this
one, because the stamp is still derived from a selection that cannot see the
secret.

ADR 0052 decision 4 already mandates the fix: *"Environment/vault API-key
overrides must resolve into the same explicit source decision."*

## The fix

`select/4` takes `:secret_keys` and resolves `Source.tenant_secret/0`, which
is `origin: :own` — so nothing is priced and nothing is counted.
`ConversationServer` merges the environment and vault *before* selecting
instead of after; those two calls were already independent, so it is a
reorder, and it takes the file from 2646 lines to 2645.

Three deliberate choices:

- **Presence only, not the value.** The value already reaches the sandbox
  through the secrets path. Merging it into the credentials map would change
  which credential the runtime exports, which is a different change with a
  different blast radius.
- **A credential row is reported ahead of a secret** when an account has
  both. Both are `origin: :own` so no billing turns on the order; the scope
  answers "why was this not platform-paid" and the row is the more specific
  answer, because it is what Fountain itself exports.
- **The managed ChatGPT grant is excluded.** ADR 0052 decision 6 reserves
  `CODEX_CHATGPT_ACCESS_TOKEN` so no configuration may name it, and a
  rotating grant cannot be overridden without breaking custody. *Reserve what
  rotates, resolve what is static.*

## The second table

`@env_names` duplicates a credential → environment-variable mapping
`Fountain.Broker` already holds (minus the grant, and minus the hosts). They
are held together by `inference_credentials_env_names_test.exs` rather than
merged, because a disagreement is silent in both directions: a name only this
table knows is a placeholder nobody substitutes, and a name only the broker
knows is an override selection cannot see — which is this bug again.

## Verification

- 114 tests across the eight files touching selection, pricing or the sandbox
  environment. Zero failures.
- **End to end through a real `ConversationServer`**: a vault secret, a
  platform key configured, and the server resolves `:tenant_secret` and the
  turn stamps `"own"`. (Note for reviewers: the secret has to be written under
  the all-zero DEK `ConversationServerCase` stubs `load_tenant_key/1` to, or
  the server cannot decrypt it and the case silently proves nothing — the
  first draft of this test did exactly that.)
- **Revert-checked**: dropping the `shadowed?/2` clause fails exactly the
  three tenant-secret tests and leaves the two negative cases (a secret for
  another provider; a credential row present too) green.
- `mix credo --strict` clean over 851 files. `mix format` under mise clean.
- `conversation_server.ex` is 2645 lines against a pin of 2646. The pin is
  left alone: lowering it inside a stack that is still in review is how the
  #1565 stack went red on a number it set itself.

## Stack

| # | PR | What |
|---|---|---|
| 2 | #2022 | `select/4` returns a source struct |
| 3 | **this** | a tenant secret resolves the source to `:own` |
| 4 | | the door gate stops refusing those launches on the platform ceiling |
| 5 | | inference pairs move off the shared `.env` |
| 6 | | `inference_credentials` gains `name` + `is_default`; 1:N |
| 7 | | `agents.inference_credential_id` + allowlist |
| 8 | | `conversations.inference_credential_id` per-launch override |
| 9 | | API, OpenAPI, contract, TS SDK |
| 10 | | console |
| 11 | | an owner may write a credential on a principal it owns |
| 12 | | docs, CHANGELOG, SDK bump |

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -680,18 +680,17 @@ defmodule Fountain.Conversations.ConversationServer do
 
     case SpriteEnv.load_tenant_state(conv.user_id) do
       {:ok, dek, own_creds} ->
+        # Before the selection: a tenant secret named after a credential wins
+        # in the sandbox, so it decides the source too (ADR 0053 decision 5).
+        tenant_secrets = SpriteEnv.merge_secrets(env, vault, dek)
+
         {inference_source, inference_creds} =
-          SpriteEnv.select_inference(agent, own_creds, conv.runtime)
+          SpriteEnv.select_inference(agent, own_creds, conv.runtime, tenant_secrets)
 
         bindings = Egress.bindings(conv.user_id)
 
         {merged, bindings, connection_keys} =
-          Egress.add_connection_secrets(
-            conv.user_id,
-            SpriteEnv.merge_secrets(env, vault, dek),
-            bindings,
-            agent
-          )
+          Egress.add_connection_secrets(conv.user_id, tenant_secrets, bindings, agent)
 
         {secrets, brokered} = Egress.split_brokered(conv.user_id, merged, bindings)
 
--- a/apps/fountain/lib/fountain/conversations/sprite_env.ex
+++ b/apps/fountain/lib/fountain/conversations/sprite_env.ex
@@ -57,15 +57,26 @@ defmodule Fountain.Conversations.SpriteEnv do
   `runtime` is the conversation's own (`conv.runtime`), which is what the
   sandbox is dispatched on and can differ from the agent's after an edit;
   nil falls back to the agent's.
+
+  `secrets` is this conversation's merged environment and vault secrets, the
+  same map `merge_secrets/3` returns. Only its **keys** are read: a secret
+  named after a static credential overrides that credential in the sandbox,
+  so the conversation is running on the tenant's own key and must not be
+  billed as platform inference (ADR 0053 decision 5). The values stay where
+  they are — they already reach the sandbox through the secrets path, and
+  merging them into the credentials map would change which credential the
+  runtime picks. Default `%{}` keeps the old answer for a caller with no
+  secrets to hand over.
   """
-  @spec select_inference(map() | nil, map(), String.t() | nil) ::
+  @spec select_inference(map() | nil, map(), String.t() | nil, map()) ::
           {InferenceCredentials.Source.t(), map()}
-  def select_inference(agent, own_creds, runtime \\ nil) do
+  def select_inference(agent, own_creds, runtime \\ nil, secrets \\ %{}) do
     brokered? = (agent && Fountain.Broker.enabled_for?(agent.user_id)) || false
     runtime = runtime || (agent && agent.runtime)
 
     case InferenceCredentials.select(agent && agent.model, own_creds, runtime,
-           brokered: brokered?
+           brokered: brokered?,
+           secret_keys: Map.keys(secrets)
          ) do
       {:ok, source, creds} -> {source, creds}
       {:error, :no_credential} -> {InferenceCredentials.Source.missing(), own_creds}
--- a/apps/fountain/lib/fountain/inference_credentials.ex
+++ b/apps/fountain/lib/fountain/inference_credentials.ex
@@ -175,6 +175,33 @@ defmodule Fountain.InferenceCredentials do
     |> MapSet.new()
   end
 
+  # The environment variable each static credential is exported as. A tenant
+  # secret of the same name overrides it in the sandbox (`Egress`'s gate-3
+  # split, published in `docs/concepts/secrets.md`), which is what
+  # `select/4`'s `:secret_keys` reads.
+  #
+  # The deployment's ChatGPT grant is deliberately absent: ADR 0052 decision 6
+  # reserves `CODEX_CHATGPT_ACCESS_TOKEN` so no configuration can name it.
+  # Reserve what rotates, resolve what is static (ADR 0053 decision 5).
+  # `inference_credentials_env_names_test.exs` holds this table against
+  # `Fountain.Broker.inference_keys/0`, which is where the same mapping is
+  # used to build the placeholders.
+  @env_names %{
+    anthropic_api_key: "ANTHROPIC_API_KEY",
+    claude_code_oauth_token: "CLAUDE_CODE_OAUTH_TOKEN",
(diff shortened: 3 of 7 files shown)


BinaryBourbon (approved):
**Approved.** Reviewed as its own patch against `stack/2018-2-source`. This is a real production mis-bill and the fix is the right shape.

## The premise checks out

I confirmed the override is real rather than only documented. `SpriteEnv.build/4` assembles the sprite env as `runtime_module.default_env(agent, env_credentials) ++ ... ++ env.env_vars ++ secrets ++ proxy`, and its own moduledoc (the #1674 note) states that later entries win. So an `ANTHROPIC_API_KEY` in the merged environment/vault map does overwrite what `default_env/2` exported from the inference credential, exactly as `egress.ex:223-226` claims.

And the mis-bill follows all the way to the ledger: `:platform` → `with_inference/2` stamps `"platform"` → `CreditPricer` selects `usage ->> 'inference' = 'platform'` and debits, and `Billing.platform_inference_spend_today/1` counts it. With `Source.tenant_secret/0` at `origin: :own` the stamp is `"own"`, which neither of those two reads. The fix reaches the thing it claims to.

## `shadowed?/2` is total, and stays total

`Map.fetch!(@env_names, &1)` over `accepted` would raise on a credential missing from the table. It cannot today: `accepted` is `credentials_for_provider/1`, which returns only `[:anthropic_api_key, :claude_code_oauth_token]`, `[:openai_api_key]`, `[:gemini_api_key]` or `[]` — never `:codex_chatgpt_access_token`. More importantly it cannot tomorrow either: `inference_credentials_env_names_test.exs` asserts `Map.keys(env_names()) == Credential.providers()`, so adding a fifth static credential fails that test before it can reach the `fetch!`. `fetch!` over `get` is the correct choice given that guard.

## The second table is held, not merged, and the reason is right

Holding `@env_names` against `Broker.inference_keys/0` in a test rather than merging them is the correct call, and the moduledoc states the failure mode accurately: a disagreement is silent in *both* directions — a name only this table knows is a placeholder nobody substitutes, and a name only the broker knows is an override selection cannot see, which is this bug again. The second test pinning `CODEX_CHATGPT_ACCESS_TOKEN` as present in the broker and absent here is what encodes ADR 0052 decision 6's reservation as an assertion rather than a comment.

## Precedence and the reorder

- The `shadowed?` clause sits between the credential-row clause and the platform clause, which matches the documented precedence and means an account holding both reports `:credential`. Both are `origin: :own`, so nothing about billing turns on the order — the reasoning in the docstring is correct.
- The `ConversationServer` change is a reorder, not a behavior change: `merge_secrets/3` was already computed independently of `select_inference/3` and is now hoisted one call earlier and reused for `Egress.add_connection_secrets/4`. Same map, same arguments.
- Presence-only is the right restraint. Merging the value into `creds` would change which credential `default_env/2` exports — a different change with a different blast radius — and the test asserting `creds == %{}` pins that.

## The end-to-end test

The `ConversationServerCase` test is the one that makes this stick, and the note about writing the vault secret under the all-zero DEK is the detail that would otherwise have made it prove nothing silently. Asserting both `Source.tenant_secret()` *and* `with_inference/2` returning `"inference" => "own"` covers selection and the stamp in one test rather than trusting the link.

## The codex case is a behavior change, and it is the right one

Worth naming explicitly because the patch does not: for a codex agent on an `openai/` model, `accepted` is `[:openai_api_key]`, so a tenant vault secret named `OPENAI_API_KEY` now resolves `:tenant_secret` and the deployment's ChatGPT grant is **not** merged into `creds` — where before it was, and the turn was billed against the platform.

That is correct rather than a regression, and `Managoat.Runtimes.Codex` is why. `prepare_sandbox/3` runs `codex login --with-api-key` whenever `OPENAI_API_KEY` is present anywhere in the sprite env — which a tenant vault secret always is, since `SpriteEnv.build/4` appends secrets after `default_env/2`. So that conversation was *already* running on the tenant's key while carrying a `"platform"` stamp. This patch makes the selection agree with what the sandbox was doing.

## One residual hole, for a follow-up not for this PR

An environment's plain `env_vars` map can also name `ANTHROPIC_API_KEY`, and `SpriteEnv.build/4` appends it after `default_env/2`, so it overrides the inference credential in the sandbox exactly like a secret does. `merge_secrets/3` only reads `decrypted_env/2`, which is `_unsafe_list_secrets/1` — `env_vars` never enters it, and `Environment.changeset/2` puts no restriction on the keys. So a tenant who puts their key in `env_vars` instead of a secret is still selected `:platform` and still mis-billed.

This is not a regression and not in scope for a PR whose title is about secrets — the gate and the provision stay in agreement, so it is the same bug class with a narrower blast radius. But it belongs on the tracker, because "a tenant secret naming a credential is not platform inference" will read as settled once this lands.

Same for the decrypt-failure edge: a secret row whose ciphertext will not decrypt is dropped by `decrypted_env/2` and so is invisible here, while #2025's gate reads the key straight from the table. Vanishingly rare, and the asymmetry lands in the permissive direction at the door.

The `conversation_server.ex` size pin is correctly left alone at 2646 with the file at 2645; lowering a pin mid-stack is how #1565 went red. CI green across all six partitions.

## Verified locally

86 tests across the six files touching selection, the stamp, the size pin and the env-name table, 0 failures, on my own database at the stack tip.

I also re-ran the revert check rather than trusting it. Dropping the `shadowed?/2` clause from `select/4` fails exactly three tests — the two `:tenant_secret` scope assertions and the OAuth-token-by-name one — and leaves both negative controls (a secret for another provider; a credential row present too) green. The fence guards what it says it guards.

Issue #2024, opened 2026-09-12
Flake: ExecutionTransportTest asserts async progress against a 200 ms wall-clock deadline


jhgaylor:
`ExecutionTransportTest` registers a wall-clock deadline and then asserts on
progress through several async hops. On a loaded machine the deadline expires
before the assertions get there.

## The failing assertion

```
1) test identity received before a delayed spawn result still enables cleanup after expiry
   (Fountain.Conversations.ExecutionTransportTest)
   test/fountain/conversations/execution_transport_test.exs:119
   ** (MatchError) no match of right hand side value:

       {:error, :deadline_expired}

   code: launch(execution, fn owner, ref ->
   stacktrace:
     test/fountain/conversations/execution_transport_test.exs:52
     test/fountain/conversations/execution_transport_test.exs:123
```

Two others in the same file fail the same way, and which of the three fail
varies run to run:

- `an unconfirmed spawn is never replayed` (`await_state(execution, "awaiting_identity")`)
- `a blocked writer times out and retires the session without replay`

## How often, out of what

Three consecutive local runs of the file alone, 16 tests each: 3 failures,
then 2, then 2. Reproduced on clean `origin/main` at `97bcf1a6` with
`--seed 0`, so it is not seed-dependent and not a branch's doing (found while
verifying an unrelated branch; the branch showed a subset of the same set).

Local only so far — I have not gone looking for it in CI runs.

**The machine was deliberately CPU-saturated at the time**, by another agent
session running `nproc * 2` busy loops to stress-test refresh contention for
the ADR 0052 stack. That is the trigger, not a confound: it is the same load a
six-partition CI runner can see, and it makes the mechanism unambiguous.

## Mechanism

`register/2` (line 20) sets `deadline_at` to `now + milliseconds` before any
work starts, and the failing tests pass 200 ms (line 119) or rely on the
`io_timeout_ms: 100`/`200` at lines 141, 153 and 190. Between registration and
the assertion the test does a `set_mimic_global` spawn, a
`DynamicSupervisor` start, a message round trip to a task, an `await_state`
poll and a `Repo.get!`. None of that is bounded by 200 ms of wall clock on a
busy box.

So the test **asserts on a state transition but waits on a wall clock**. Same
family as #1592, #1629 and #1721, all closed.

The fix is presumably to give these tests a deadline generous enough that only
a genuinely stuck transport trips it, and to drive expiry explicitly where
[...]

PR #2025, opened 2026-09-12
fix: the door gate asks what the provision will answer

jhgaylor:
**Base: #2023.** ADR 0053 decision 5, second half. Tracker #2018.

## The bug

`PlatformInference.gate/3` refuses a conversation when three things are true:
the deployment holds a key for the model's provider, the tenant holds none of
their own, and the day is spent.

Question two read the `inference_credentials` row alone. So a tenant whose
vault or environment supplies the credential — which is what actually serves
the conversation, and what #2023 just taught selection to see — was refused at
the door on a ceiling they were not spending against. Same root cause as
#2023, opposite symptom: that one over-bills them, this one locks them out.

## The fix

`gate/4` takes the launch's `:environment_id` and `:vault_id`;
`InferenceCredentials.has_own?/3` consults them. All four doors pass what they
have:

| door | source |
|---|---|
| `start_conversation` | the `env_id` / `vault_id` already resolved in the same `with` |
| the attach path | same |
| wake from suspended | `conv.environment_id \|\| agent.environment_id`, `conv.vault_id` |
| re-provision after a dead sandbox | same |

The `\|\| agent.environment_id` fallback is not invented here — it is exactly
how `ConversationServer` resolves the environment at provision
(`conversation_server.ex:733`), so the door asks the question the provision
answers.

`InferenceCredentials.secret_keys/2` is the read: names, never values, only
the four `env_names/0` knows, and joined to `user_id` so an id belonging to
another tenant contributes nothing rather than leaking that it exists.

**It stays cheap.** The read happens only inside the branch that was about to
refuse — the deployment holds a key for this provider *and* the account holds
no row. A deployment with no platform key still costs one primary-key read and
nothing else, which is what the gate's docstring promises and what makes the
feature free for a self-hoster.

## Verification

- 27 tests in `platform_inference_test.exs`, zero failures. Three are new: a
  vault naming the credential, an environment naming it, and a secret for
  another provider (still refused). The environment case asserts the refusal
  *first* with only an unrelated key in that environment, then adds
  `ANTHROPIC_API_KEY` and asserts `:ok` — so it is pinned on the name and not
  on the row existing.
- Cross-tenant: the same vault id presented by a second account is still
  refused.
- **Revert-checked**: dropping `opts` from the `has_own?` call fails the vault
  and environment tests and leaves the other-provider one green.
- `mix credo --strict` clean over 851 files; `mix format` under mise clean.

## One thing reviewers should know

`test/fountain/conversations/execution_transport_test.exs` fails on this
branch. **It also fails on clean `origin/main` at `97bcf1a6` with `--seed 0`**,
with a superset of the same failures, and which of its three tests fail varies
run to run. It is a wall-clock flake in code that landed today (#1748-#1752),
now filed as #2024 with the mechanism. Nothing in this stack touches that file
or anything it exercises.

## Stack

| # | PR | What |
|---|---|---|
| 2 | #2022 | `select/4` returns a source struct |
| 3 | #2023 | a tenant secret resolves the source to `:own` |
| 4 | **this** | the door gate stops refusing those launches |
| 5 | | inference pairs move off the shared `.env` |
| 6 | | `inference_credentials` gains `name` + `is_default`; 1:N |
| 7 | | `agents.inference_credential_id` + allowlist |
| 8 | | `conversations.inference_credential_id` per-launch override |
| 9 | | API, OpenAPI, contract, TS SDK |
| 10 | | console |
| 11 | | an owner may write a credential on a principal it owns |
| 12 | | docs, CHANGELOG, SDK bump |

### DIFF
--- a/apps/fountain/lib/fountain/conversations.ex
+++ b/apps/fountain/lib/fountain/conversations.ex
@@ -3325,7 +3325,11 @@ defmodule Fountain.Conversations do
          # Whose inference key would run this (#1388): refused only when it
          # would be Fountain's and the deployment has spent its day. A door
          # with no platform key configured runs no query here.
-         :ok <- Fountain.PlatformInference.gate(user_id, agent.model, agent.runtime),
+         :ok <-
+           Fountain.PlatformInference.gate(user_id, agent.model, agent.runtime,
+             environment_id: env_id || agent.environment_id,
+             vault_id: vault_id
+           ),
          # A persistent launch lands on the identity's home when there is one
          # (ADR 0023 gate 6): `{:home, sandbox}` leaves the `with` and attaches
          # below. Only when there is none does a machine get provisioned, and
@@ -4312,7 +4316,11 @@ defmodule Fountain.Conversations do
          # Whose inference key would run this (#1388): refused only when it
          # would be Fountain's and the deployment has spent its day. A door
          # with no platform key configured runs no query here.
-         :ok <- Fountain.PlatformInference.gate(user_id, agent.model, agent.runtime),
+         :ok <-
+           Fountain.PlatformInference.gate(user_id, agent.model, agent.runtime,
+             environment_id: env_id || agent.environment_id,
+             vault_id: vault_id
+           ),
          %Sandbox{} = sandbox <- get_sandbox(sandbox_id, user_id) || {:error, :sandbox_not_found},
          :ok <- check_sandbox_api_attach(sandbox, attrs["sandbox_api_access"]),
          :ok <- check_attachable(sandbox, agent, vault_id, env_id),
@@ -5224,7 +5232,9 @@ defmodule Fountain.Conversations do
                  Fountain.PlatformInference.gate(
                    conv.user_id,
                    agent.model,
-                   conv.runtime
+                   conv.runtime,
+                   environment_id: conv.environment_id || agent.environment_id,
+                   vault_id: conv.vault_id
                  ),
                {:ok, _} <- wake_suspended_sandbox(conv.user_id, sandbox_id) do
             case start_conversation_server(conv, sandbox_id, runtime_module, initial_prompt) do
@@ -5533,7 +5543,9 @@ defmodule Fountain.Conversations do
            Fountain.PlatformInference.gate(
              conv.user_id,
              agent.model,
-             conv.runtime
+             conv.runtime,
+             environment_id: conv.environment_id || agent.environment_id,
+             vault_id: conv.vault_id
            ),
          # A fresh sandbox is a fresh placement decision — re-resolve from
          # the agent, so a conversation whose old sandbox died can migrate
--- a/apps/fountain/lib/fountain/inference_credentials.ex
+++ b/apps/fountain/lib/fountain/inference_credentials.ex
@@ -244,10 +244,66 @@ defmodule Fountain.InferenceCredentials do
   True also when the model's provider needs none at all (a local model, a
   gateway): there is nothing missing, which is the same answer
   `missing_for_model/2` gives.
+
+  `opts` may name a launch's `:environment_id` and `:vault_id`. With either,
+  a secret of theirs named after a credential the provider accepts counts as
+  the tenant's own, because it is what will serve the conversation (ADR 0053
+  decision 5). With neither — every caller that has no launch in hand — this
+  is exactly the row question it has always been, and runs no extra query.
   """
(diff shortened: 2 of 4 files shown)


BinaryBourbon (approved):
**Approved.** Reviewed as its own patch against `stack/2018-3-shadow`. I checked the title's claim against the code rather than taking its word, and it holds at all four doors.

## The gate does ask what the provision answers

`ConversationServer.provision` resolves its inputs at `conversation_server.ex:734` and `:751`:

```elixir
env_id = conv.environment_id || (agent && agent.environment_id)
vault = if conv.vault_id, do: Vaults._unsafe_get_vault(conv.vault_id)
```

Against the four gate call sites:

- **`start_conversation`** (`conversations.ex:3192`) passes `env_id || agent.environment_id` and `vault_id` — the same expression the very next line already hands `home_or_new/5`, and `env_id`/`vault_id` are what land on `conv.environment_id`/`conv.vault_id`. Matches.
- **the attach path** (`:3974`) — same expression, and `check_attachable/4` runs after on the same pair. Matches.
- **wake from suspended** (`:4843`) and **re-provision after a dead sandbox** (`:5154`) pass `conv.environment_id || agent.environment_id` and `conv.vault_id`, character-for-character what `provision` computes. Matches.

So the `|| agent.environment_id` fallback is not invented at the door, as the description says. Confirmed.

## Tenant scoping

`secret_keys/2` joins the owning row and filters on `user_id`, for both the environment and the vault query, so a launch naming another tenant's vault contributes nothing and the gate does not distinguish "not yours" from "does not exist". No `_unsafe_` call, no ownership assumption. The test that presents the same vault id under a second account and still gets `{:error, :platform_inference_unavailable}` is the right assertion for this, and it is the one I would have asked for if it were missing.

Names, never values, and `s.key in ^names` bounds the read to the four `env_names/0` knows — so this cannot become an accidental secret-key enumeration surface.

## It really does stay cheap

The claim is that a deployment with no platform key pays nothing. Verified by the shape rather than by the comment: `serves?(...) and not has_own?(...)` short-circuits on `and`, so `has_own?/3` is not called at all when the deployment holds no key for the provider; and inside `has_own?/3`, `secret_keys/2` runs only on the `{_provider, accepted}` branch — i.e. only when `missing_for_model/2` already said the account holds no row. Both queries are further guarded by `if env_id` / `if vault_id`, so a launch with neither runs none. The docstring's promise matches the control flow.

## The tests

The environment test is constructed the right way round: it asserts the **refusal** first with only an `UNRELATED` key in the environment, then adds `ANTHROPIC_API_KEY` and asserts `:ok`. That pins the behavior on the name rather than on the environment row existing, which a one-shot positive test would not have done. The other-provider case (`OPENAI_API_KEY` against an `anthropic/` model, still refused) closes the obvious over-permissive failure.

## Consistency with the rest of the stack

`has_own?/3` has exactly one caller (`PlatformInference.gate/4`), and all four `gate/3` call sites in `lib` are updated; the default `opts \\ []` keeps the arity-2/arity-3 behavior byte-identical for anything that reaches them without a launch. #2023's provision-time `shadowed?/2` and this door-time `secret_keys/2` read the same `@env_names` table, so the two halves cannot drift on which names count.

## Carried over from #2023, not blocking here

The gate inherits the same two gaps I noted there and stays consistent with the provision in both, so neither is a regression: an environment's plain `env_vars` map can name a credential and override it in the sandbox without being a `Secret` row (invisible to both halves), and a secret whose ciphertext will not decrypt is visible here but dropped by `merge_secrets/3` at provision. Tracker material.

On `execution_transport_test.exs`: I did not reproduce it, but the reasoning given — it fails on clean `origin/main` at `97bcf1a6` with a superset of the same failures, the failing subset varies run to run, and nothing in this stack touches that file — is the right standard, and filing it as #2024 with the mechanism rather than re-running until green is what this repo asks for. CI is green across all six partitions here.

## Verified locally

Ran the stack tip against my own database: 86 tests across `platform_inference_test.exs`, `platform_chatgpt_test.exs`, `sprite_env_inference_test.exs`, `inference_credentials_env_names_test.exs`, `conversation_server_platform_inference_test.exs` and the size pin, 0 failures.

I reproduced the revert check rather than trusting it. Dropping `opts` from the `has_own?/3` call fails exactly the vault test and the environment test, and leaves "a secret for another provider does not excuse this one" green — so the gate is pinned on the launch's ids reaching the read, not on the tests happening to pass.

PR #2027, opened 2026-09-12
fix(inference): keep provider credentials out of shared sandbox env

jhgaylor:
Provider credentials written to a shared sandbox `.env` can be read by other conversations. Keep supported credential names and aliases, including the Google runtime alias and managed Codex token, in process environments only. Empty environment writes also replace stale shared-file contents.

The upgrade note is included in this first behavioral slice: scripts that re-source `.env` must use their inherited process environment for provider credentials.

Validation: compilation with warnings as errors; 72 focused tests passed, one existing Linux-only check skipped on macOS.

Base: main. Part of #2018; next #2028.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/identity.ex
+++ b/apps/fountain/lib/fountain/conversations/identity.ex
@@ -52,16 +52,16 @@ defmodule Fountain.Conversations.Identity do
   # cross-conversation read of a credential that brokers another tenant's
   # vault. `Fountain.Broker.process_only_keys/0` names the variables.
   #
-  # The inference credential is per-conversation for the same reason (ADR 0053
-  # decision 4). Which credential runs a conversation is decided per
-  # conversation by `InferenceCredentials.select/4`, and a sandbox carries
-  # several of them (ADR 0023), so a value on the shared disk is a
-  # cross-conversation read of whichever conversation last provisioned.
-  # `InferenceCredentials.env_names/0` names the four; the managed ChatGPT
-  # grant is not among them and never reaches the env file at all.
-  @process_only [@tag_key, "FOUNTAIN_TOKEN", "TRACEPARENT"] ++
+  # Inference auth inputs are per-conversation (ADR 0053 decision 4),
+  # including tenant env/vault secrets with a supported name or alias.
+  # Unbrokered values are bearer material; broker placeholders are harmless
+  # fixed strings, but neither needs to persist in the shared env file.
+  # The managed ChatGPT path also emits a placeholder into the process env.
+  # Strip its reserved name defensively along with the static credentials.
+  # Runtime-owned auth files require separate isolation (decision 6).
+  @process_only [@tag_key, "FOUNTAIN_TOKEN", "TRACEPARENT", "CODEX_CHATGPT_ACCESS_TOKEN"] ++
                   Fountain.Broker.process_only_keys() ++
-                  Map.values(Fountain.InferenceCredentials.env_names())
+                  (Fountain.InferenceCredentials.env_aliases() |> Map.values() |> List.flatten())
 
   @tag_re ~r/(?:^|\s)FOUNTAIN_CONVERSATION_ID=([0-9a-fA-F-]{36})(?:\s|$)/
 
@@ -77,7 +77,7 @@ defmodule Fountain.Conversations.Identity do
 
   @doc """
   The subset of a sprite env that belongs on the machine's disk: everything
-  except the per-conversation identity.
+  except the per-conversation identity and inference auth inputs.
   """
   @spec disk_env([{String.t(), String.t()}]) :: [{String.t(), String.t()}]
   def disk_env(sprite_env) when is_list(sprite_env) do
--- a/apps/fountain/lib/fountain/conversations/identity.ex
+++ b/apps/fountain/lib/fountain/conversations/identity.ex
@@ -20,9 +20,11 @@ defmodule Fountain.Conversations.Identity do
       pairs before the env file is written; the same pairs still reach every
       spawn through `env:`, so the agent's tools inherit them exactly as
       before. What a `source .env` in a setup script loses is the callback
-      token, which it had no business holding — the file is for environment
-      and vault values, which are the same for every conversation on the
-      machine.
+      token and the inference credential, neither of which the file had any
+      business holding — it is for environment and vault values, which are
+      the same for every conversation on the machine. Both still reach the
+      setup script itself: `Provisioning.run_setup_script/4` execs with
+      `env: sprite_env`, the whole list.
 
     * **A session inherits its conversation.** `tag_command/3` wraps the spawn as
       `env FOUNTAIN_CONVERSATION_ID=<id> <cmd> <args…>`. The process gets the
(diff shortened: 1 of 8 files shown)


jhgaylor:
Local `test/fountain/conversations/` run landed: **1282 tests, 4 failures, 2 skipped**.

All four are #2024's family and none touch anything this PR changes:

- `ExecutionTransportTest` — "a blocked writer times out and retires the session without replay", "identity received before a delayed spawn result still enables cleanup after expiry"
- `ExecutionDeadlineWorkerTest` — "a local timeout retains its attempt and restart recovery never replays it", "expiry progresses while every termination slot is blocked"

`ExecutionDeadlineWorkerTest` is a second file in the same family; I have added it to #2024. Everything else in the tree is green, and CI is green on all six partitions for every PR below this one.

apps/fountain/test/fountain/conversations/identity_test.exs
```diff
@@ -20,13 +20,33 @@ defmodule Fountain.Conversations.IdentityTest do
...
                {"SANDBOX_URL", "https://sb.example"},
                {"GITHUB_TOKEN", "ghp_x"}
              ]
     end
 
+    # ADR 0053 decision 4. Which credential runs a conversation is a
+    # per-conversation decision and a sandbox carries several conversations
+    # (ADR 0023), so a value in the shared file is a cross-conversation read
+    # of whichever one provisioned last. A tenant secret of the same name is
+    # not the credential and is not stripped: it belongs to the environment
```
BinaryBourbon: This is not what `disk_env/1` does. It rejects by key name, and `SpriteEnv.build/4` appends the merged environment and vault secrets to the same list, so a tenant secret named `ANTHROPIC_API_KEY` (ADR 0053 decision 5, `docs/concepts/secrets.md`) is stripped from the file exactly like the credential is.

The behaviour looks right to me; the comment has to say it, and the break note has to cover it.

apps/fountain/lib/fountain/conversations/identity.ex
```diff
@@ -49,7 +51,17 @@ defmodule Fountain.Conversations.Identity do
...
   # vault. `Fountain.Broker.process_only_keys/0` names the variables.
-  @process_only [@tag_key, "FOUNTAIN_TOKEN", "TRACEPARENT"] ++ Fountain.Broker.process_only_keys()
+  #
+  # The inference credential is per-conversation for the same reason (ADR 0053
+  # decision 4). Which credential runs a conversation is decided per
+  # conversation by `InferenceCredentials.select/4`, and a sandbox carries
+  # several of them (ADR 0023), so a value on the shared disk is a
+  # cross-conversation read of whichever conversation last provisioned.
+  # `InferenceCredentials.env_names/0` names the four; the managed ChatGPT
+  # grant is not among them and never reaches the env file at all.
```
BinaryBourbon: It does reach the env file: `CodexChatGPT.env/2` is in `SpriteEnv.build/4`, and `disk_env/1` does not strip it. What is true is what your PR body says — the grant is offered only under `brokered: true`, so what lands there is `Broker.placeholder/1`'s fixed string and never a token.

BinaryBourbon (changes requested):
**REQUEST_CHANGES** — the code is right and the mechanism is the right one. Two comments landing with it state the opposite of what the code does, and one of them is the justification a test encodes. Both are cheap to fix.

## What I checked

- Every spawn still gets the four names. `SpriteEnv.build/4` is unchanged, `disk_env/1` is applied only at the two `write_env_file` call sites (`conversation_server.ex:971` provision, `:1174` reattach), and `run_setup_script/4` execs with `env: sprite_env` — the whole list, not the disk subset. The ACP peer, `prepare_runtime_sprite/5`, `clone_repositories/4` likewise.
- Derivation from `InferenceCredentials.env_names/0` rather than a second literal list: right call, and it matches ADR 0053 decision 4 exactly ("the four static credential variable names").
- The behaviour is pinned by a test that fails if reverted: `disk_env/1` over the four `env_names/0` pairs asserting `[]` goes red the moment they come off `@process_only`.
- Swept for other on-disk writers. The remaining one is codex's `$CODEX_HOME/auth.json` (`Managoat.Runtimes.Codex.prepare_sandbox/3` consumes `OPENAI_API_KEY` into `codex login --with-api-key`), which is ADR 0053 decision 6 / ADR 0052 decision 5 and explicitly inherited, not this stage's.

## Must change

**1. `identity_test.exs:31-35` — a tenant secret of the same name *is* stripped.**

The comment says:

> A tenant secret of the same name is not the credential and is not stripped: it belongs to the environment or vault, which is what the file is for

`disk_env/1` rejects by key name, with no provenance. `SpriteEnv.build/4` appends the merged environment and vault secrets to the same list, so a tenant who defines `ANTHROPIC_API_KEY` as an environment or vault secret — the documented override, ADR 0053 decision 5 and `docs/concepts/secrets.md`, the exact path stage 3 taught `select/4` to resolve — now loses that value from `/home/sprite/.env` too.

I think the code behaviour is correct (different conversations on one machine may attach different environments, so that value was never safely machine-scoped either). What has to change is the comment, and the scope of the break note: "The break" section describes a `setup_script` re-sourcing the file "to recover a provider key", which reads as being about the credential Fountain exports. A tenant whose environment carries the key does not think of it that way, and they are affected identically. Either state it here, or add a test pinning it so the next reader learns it from the suite rather than from a support ticket.

To your question about the CHANGELOG: with the break wider than currently described, I would pull the note forward into this PR. #2046 is fine for a break that is announced accurately; it is not fine as the only place a tenant on the documented secret-override path finds out.

**2. `identity.ex:61` — the grant does reach the env file.**

> the managed ChatGPT grant is not among them and never reaches the env file at all

`Fountain.Conversations.CodexChatGPT.env/2` is the second element of `SpriteEnv.build/4`'s concatenation, so `CODEX_CHATGPT_ACCESS_TOKEN` is in `sprite_env` and `disk_env/1` does not strip it — it reaches the file. Your PR body has this right ("what lands in the file is always the broker's placeholder, never a token", because `select/4` offers the grant only under `brokered: true`); the code comment says something stronger and false. The test at `:43` repeats it in its name — "it never reaches the env list" — while asserting something else entirely (`refute ... in process_only_keys()`).

Say what is true: it reaches the file as `Broker.placeholder/1`'s fixed string, never as a token, which is why it is left alone.

## Non-blocking

- The test at `:43` pins the *absence* of a protection. If someone later adds the grant to `@process_only` — strictly safer, costs nothing — this goes red for no reason a reader can act on. Rename it to what it asserts and put the "we deliberately do not strip it, here is why" in the body.
- On a brokered conversation what was sitting in the shared file was `__anthropic_api_key__` — `Broker.placeholder/1` is per key, not per conversation. So the "cross-conversation read of whichever conversation last provisioned" the comment describes is a live exposure for an unbrokered conversation and defence in depth for a brokered one, today and after stage 8. Worth one clause, since the comment is the argument the next reader inherits.
- A pointer from the `@process_only` comment to ADR 0053 decision 6 would stop a reader concluding the machine's disk is now clean of inference credentials — codex's `auth.json` is still there.

BinaryBourbon (approved):
AI-assisted adversarial review — APPROVED

Reviewed exact head `32c7336c6b1851596cd4396854197a55a172ff8b` against base `fba020a072ddcabd22aa734526f6bac9131168f9` in an isolated checkout. No blocking defect found in this slice.

Inspected all 8 changed files: CHANGELOG.md; conversations/identity.ex and provisioning.ex; inference_credentials.ex; identity_test.exs, inference_process_env_test.exs, provisioning_test.exs, and inference_credentials_env_names_test.exs. Traced both provisioning/reattach env-file callers, SpriteEnv assembly/redaction, setup-script exec, and runtime preparation. Canonical provider names, Google runtime alias, tenant env/vault overrides, and managed Codex inputs are filtered by name; process inputs remain intact. Empty writes replace stale file contents. The older human objections about contradictory comments and missing upgrade scope are addressed by current code, tests, and CHANGELOG.

Evidence: source and regression assertions reviewed; current-head GitHub `CI required` and `Detect secrets` passed. I did not run local tests. Runtime-owned auth files remain explicitly outside this slice (ADR 0053 decision 6); this approval does not assert full sandbox credential isolation or waive merge gates.

BinaryBourbon (approved):

Reviewed the entire PR at exact head `34df8f644886e03c2df669a70ddb2d355e5e2eb5` against approved base `0c8443b6be359f28ca988bd0831398d10b307499` in an independent isolated checkout; no blocking defect found in this slice. This records the current head/base explicitly (the earlier review's prose referenced older commits).

Coverage: all 8 changed files — CHANGELOG.md; apps/fountain/lib/fountain/conversations/identity.ex and provisioning.ex; apps/fountain/lib/fountain/inference_credentials.ex; apps/fountain/test/fountain/conversations/identity_test.exs, inference_process_env_test.exs and provisioning_test.exs; apps/fountain/test/fountain/inference_credentials_env_names_test.exs. Read approved-base contributor/reviewer policy and ADR 0053, prior reviews and comments. Traced both provisioning/reattach file-write callers, SpriteEnv assembly/redaction, setup-script exec and pinned runtime adapters. Filtering includes canonical names, the OpenCode Google alias, environment/vault inputs and managed ChatGPT placeholders, while each process retains its env. Empty filtered env now overwrites stale file contents. Prior comment and compatibility-note objections are addressed.

Validation: exact-head targeted suites completed locally on OTP 28.3 / Elixir 1.19.2 with disposable PostgreSQL 18: **52 tests, 0 failures, 1 existing skipped test**. Current-head actual CI run [34752555135](https://github.com/managoat/fountain/actions/runs/34752555135) and Detect secrets succeeded. I did not rerun the entire verification recipe locally. This is a source-review approval, not an assertion that runtime-owned auth files or cross-process access are isolated; ADR 0053 decision 6 and independent CI/maintainer gates remain applicable.

PR #2028, opened 2026-09-12
feat(inference): add named credential sets with verified rename semantics

jhgaylor:
Allow multiple tenant-owned credential sets with a unique name and one default. Existing credentials retain their Default set. Rename persists the new name, rejects duplicates within the tenant, and records accurate before/after audit metadata after commit.

Validation: formatting, compilation with warnings as errors, fresh migrations, and named-set regressions pass, including concurrent default creation and stale promotion/deletion.

Base: #2027. Part of #2018; next #2031.


apps/fountain/lib/fountain/inference_credentials.ex
```diff
@@ -23,14 +23,137 @@ defmodule Fountain.InferenceCredentials do
...
+
+  @doc """
+  Rename a credential set. Audited as `inference_credential_set.renamed`,
+  recording both names: a set's name is a label the tenant chose, not secret
+  material, and a trail that cannot say what a thing used to be called cannot
+  explain a later event that names it.
+  """
+  @spec rename_set(Credential.t(), String.t(), keyword()) ::
+          {:ok, Credential.t()} | {:error, Ecto.Changeset.t()}
+  def rename_set(%Credential{} = set, name, opts \\ []) do
```
BinaryBourbon, line 92: No test covers this function beyond the audit guardrail, and the guardrail would pass on a broken rename: `audited_set/3` reads `set.name` off the struct `Repo.update/1` returned, so a rename that changed nothing still returns `{:ok, set}` and still records `inference_credential_set.renamed`. Needs a test that the name persists, that the event carries `was`/`now`, and that a duplicate name comes back as the `:name` changeset error.

apps/fountain/lib/fountain/inference_credentials.ex
```diff
@@ -23,14 +23,137 @@ defmodule Fountain.InferenceCredentials do
...
+  def get_set(id, user_id) when is_binary(id) and is_binary(user_id) do
+    Repo.get_by(Credential, id: id, user_id: user_id)
+  end
+
+  @doc """
+  Create a named credential set holding nothing yet.
+
+  The first set an account gets is its default, whoever asks for it: an
+  account with no default is an account nothing can read a credential for.
+  Every later one is not, until `set_default/3` says so.
```
BinaryBourbon: `set_default/2`, not `/3` — same on line 107, in the `is_default` field comment in `credential.ex`, and in the migration comment.

BinaryBourbon (changes requested):
**REQUEST_CHANGES** — the schema and the migration are right; `rename_set/3` ships with no test that would fail if it stopped working.

## What I checked

- **Backfill is exhaustive.** `20260510210000_create_inference_credentials.exs` put `unique_index(:inference_credentials, [:user_id])` on the table, so every account has had at most one row. `UPDATE inference_credentials SET name = 'Default', is_default = TRUE` therefore reaches every pre-existing row and cannot produce two defaults for one account. The three shapes all land where the PR says: a row with keys becomes a Default set that `get_for_user/1` reads; a row whose ciphertexts are all `nil` becomes a Default set that `has_any_credential?/1` still reports `false` for; an account with no row still gets its Default built by `put_credential/5` on first write.
- **Ordering in `up`.** Column added, backfilled, then `NOT NULL`, then the old index dropped and the two new ones created — the partial index is created after the backfill, so it never sees a half-populated table.
- **`down` reverses.** Non-default sets are deleted before `unique_index(:user_id)` comes back, which it has to be, and the migration says so and says which row survives. Deliberate data loss, documented in the file rather than only in the PR body.
- **Constraint names match.** `:inference_credentials_user_id_name_index` is what Ecto generates for `unique_index(:inference_credentials, [:user_id, :name])`, and `:inference_credentials_one_default_index` matches the `name:` in the migration. Neither will raise `Ecto.ConstraintError` where the changeset expects an error.
- **Tenant scoping.** `list_sets/1`, `get_set/2` and `create_set/3` are scoped; the mutators take a loaded struct, which is the shape the rest of the repo uses, and `set_default/2` scopes its `update_all` by `set.user_id` rather than by anything the caller passed. No new fetch-by-id without a `user_id`, so nothing here owes an `_unsafe_` prefix.
- **Secrets and audit.** No plaintext is added anywhere; the four new events record the set id and name and no ciphertext, and `Audit.record_resource/4` already establishes that a resource's name belongs in metadata. The `default_changed` recording is outside the transaction, and the already-default no-op records nothing and is tested.
- **`has_any_credential?/1` widening to any set.** Its only production caller is `dashboard_live/index.ex:32`, so the divergence from `status_for_user/1` cannot reach a launch. Fine as built.

## Blocking

**`rename_set/3` has no behavior test, and the guardrail entry it does have would pass on a broken rename.** `audited_set/3` reads `set.name` off the struct `Repo.update/1` returned, so a rename that silently failed to change anything still returns `{:ok, set}` and still records `inference_credential_set.renamed` with the expected action — which is all `audit_guardrail_test.exs` asserts. `rename_set/3` is the one new public mutator with nothing in `inference_credential_sets_test.exs`, and the PR body makes its behavior an explicit decision ("The rename records both names"). Please add a `describe "rename_set/3"` asserting:

- the new name is what `list_sets/1` reads back afterwards;
- the event carries `"was"` and `"now"` — the decision the PR body defends is currently unasserted anywhere;
- renaming onto a name the account already uses comes back as the `:name` changeset error, not an `Ecto.ConstraintError`. That is the one path where the `(user_id, name)` index reaches a user through a rename rather than a create, and only the create half is covered today.

## Fix in the same pass

- **The docs say `set_default/3`; it is `set_default/2`.** Four places: `inference_credentials.ex:62`, `inference_credentials.ex:107`, the `is_default` field comment in `credential.ex`, and the comment in the migration. The commit message and the `@spec` both have it right.

## Non-blocking

- `audited_set/3` is `Audit.record_resource/4` rewritten inline — same `user_id`/`id`/`name` denormalisation, same actor and request_ip defaults. The only difference is which side of the metadata merge wins. Worth collapsing onto the shared helper unless the merge direction is load-bearing.
- `has_any_credential?/1`'s docstring still says it is read "by the conversation-start flow to give a clearer error than 'auth failed in the sprite'". There is no such caller left — the dashboard is the only one. The PR rewrites the paragraph under this line, so it is the natural place to drop the stale half.
- `name` is not trimmed and uniqueness is case-sensitive, so `"  "` is a valid set name and `Work` and `work` can coexist on one account. `validate_length(min: 1)` never fires either, because Ecto casts `""` to `nil` and `validate_required` catches it first.
- `create_set/3` decides `is_default` from `is_nil(get_for_user(user_id))`, which is check-then-act. Two concurrent first-set creations both read `nil` and the loser surfaces "an account has exactly one default credential set" against `:is_default`, a field no form can show. The index does keep the data correct, which is the part that matters.
- ADR 0053 still reads "**Status:** Proposed; none of the behavior below is built" while this PR builds decision 1. Stack-wide rather than this stage's alone — stages 2 to 5 left it too — so presumably stage 12's job, but CLAUDE.md asks for the caveat to come out in the PR that builds the thing.

BinaryBourbon (changes requested):

Exact head `4b96938d47c9f4e3caa20914b0794b8519f876b1`; base `32c7336c6b1851596cd4396854197a55a172ff8b`. Reviewed all 5 changed files: inference_credentials.ex, inference_credentials/credential.ex, the named-set migration, audit_guardrail_test.exs, and inference_credential_sets_test.exs, plus credential callers, tenant scoping, encryption, audit behavior, and migration up/down.

**P2: Reload the set before deciding a rename and its audit metadata.** [rename_set/3, lines 92–102](https://github.com/managoat/fountain/blob/4b96938d47c9f4e3caa20914b0794b8519f876b1/apps/fountain/lib/fountain/inference_credentials.ex#L92) builds the changeset and `was` from a caller-held struct. Reproduction: fetch `Work`; rename it to `Personal`; call `rename_set(original_struct, "Work")`. Ecto detects no name change against that stale struct, returns success, and leaves the database at `Personal`, while recording a fictitious Work→Work rename. A subsequent stale rename to `Client` persists but records `was: Work`, although the actual previous name was `Personal`. Overlapping API requests or a retained LiveView row can create exactly this interleaving. Fetch the current tenant-owned row inside the source lock, derive both mutation and metadata there, return a not-found refusal if it was deleted, and audit only a real committed change.

Evidence: ran the 18 shipped named-set tests and two independent stale-row regressions against this exact head in an isolated checkout and disposable PostgreSQL 18 database: **20 tests, 2 failures**, both independent regressions above. The earlier missing ordinary-rename/duplicate-name coverage objection is fixed, but does not cover this stale-row behavior. Database uniqueness/default promotion/deletion use the new lock and current-row checks; rename needs the same discipline.

Current-head newest full `CI required` and `Detect secrets` passed. An older cancelled duplicate run is not a current test failure. Migration rollback deliberately deletes non-default sets, as documented in its down path and previously reviewed; source verdict and any migration/merge approval gate remain separate.

BinaryBourbon (approved):

Completed an independent whole-PR review of exact head `1b1e8fe8ba68cca2bb2093559ffe912e03548977` against approved base `34df8f644886e03c2df669a70ddb2d355e5e2eb5`. No blocking defect found in this slice.

Coverage: all 5 changed files — apps/fountain/lib/fountain/inference_credentials.ex; apps/fountain/lib/fountain/inference_credentials/credential.ex; apps/fountain/priv/repo/migrations/20260912050000_name_inference_credential_sets.exs; apps/fountain/test/fountain/audit_guardrail_test.exs; apps/fountain/test/fountain/inference_credential_sets_test.exs. Read the approved-base contributor/reviewer instructions, ADR 0053 and prior reviews/comments. Traced default-set reads and legacy provider writes through API/LiveView, provisioning, platform eligibility, avatar generation and dashboard/admin connectivity reporting. Checked tenant scoping, ciphertext handling, source locks, uniqueness/default invariants, audit placement and migration ordering/up/down.

The previous stale-rename blocker is fixed: rename reloads id + tenant under the source lock, derives both the changeset and predecessor name from that current row, suppresses no-op audit, and refuses a deleted/foreign row. The shipped regressions exercise Work→Personal followed by stale Work/Client requests, persisted names, real predecessor metadata, duplicate-name rejection and unchanged audit counts. Default promotion/deletion similarly consult current rows under the lock. The migration backfills every pre-existing singleton before adding the new indexes; legacy readers continue to select its default.

Evidence: migration applied locally and exact-head named-set, existing credential/encryption and audit suites passed: **128 tests, 0 failures**, using OTP 28.3 / Elixir 1.19.2 and disposable PostgreSQL 18. Actual current-head [CI run 34752555617](https://github.com/managoat/fountain/actions/runs/34752555617) and Detect secrets succeeded; the older cancelled duplicate aggregate is separate. I did not rerun the full verification recipe locally. Migration rollback intentionally deletes non-default sets as documented; this AI-assisted source approval neither represents a human migration decision nor waives the approved-base maintainer/migration or merge gates.

PR #2031, opened 2026-09-12
feat(inference): bind agent selection to durable credential sources

jhgaylor:
An agent can select a named credential set without allowing billing, provisioning, and later turns to disagree about which source serves it. Persist nonsecret source identity and revision on conversations and turns; validate and reserve the source atomically at admission. Credential, environment, vault, and platform mutations follow the same lock ordering before changing source rows.

Runtime-aware selection handles provider aliases and conflicting inputs. Source replacement or deletion refuses existing bindings instead of switching them. Codex reserves its source on the sandbox for that machine's lifetime, including after peer failure or deletion; a different source requires a new sandbox. Stale actors and changed live models cannot overwrite admitted turn snapshots.

This stage includes the conversation FK and source migration as internal binding prerequisites. Per-launch selection remains disabled until #2034. All serving nodes must adopt the new lock ordering before enabling this behavior; rolling forward old writers concurrently is unsupported.

Validation: 466 core tests passed, one existing Linux-only check skipped; formatting, compilation, and pre-API contract freshness pass.

Base: #2028. Part of #2018; next #2034. Human review objections remain visible for fresh review.


jhgaylor:
CI caught two real failures in partition 2 and I have pushed the fix.

`ConversationServerCase` stubs `InferenceCredentials.decrypted_for_user/2`. This PR makes `load_tenant_state/2` read through `decrypted_for/3`, and **Mimic does not intercept a module calling itself** — so the real query ran, every credential map came back empty, and the two claude OAuth-fallback tests had no API key to fall back to.

Fixed in the shared harness rather than in the four test files that set credentials, and the new stub **delegates** to `decrypted_for_user/2` rather than answering. Answering directly would have made the case's empty map win over a test's own credentials as soon as a set was involved — a silent wrong answer instead of a failure.

Local: `conversation_server_acp_test.exs`, `_broker_test.exs` and `_platform_inference_test.exs` together, 142 tests, zero failures. #2034 rebased onto this.

BinaryBourbon (changes requested):
**Request changes** — two things: the door gate still asks the *default* set, and the one line that makes the headline claim true has no test that would fail if it were reverted.

Checked: tenant scoping on both write paths and on the read path, the FK and index in the migration and its reversibility, `@snapshot_fields` / rollback, every caller of `load_tenant_state/1` and `decrypted_for_user/2`, the `select/4` fallthrough, and the credits gate at the four `start_conversation` doors (`check_spend/1` runs ahead of the inference gate at each; untouched).

The ownership work is right. `validate_credential_set_owner/1` mirrors `validate_environment_owner/1` including the `valid?: false` short-circuit, both create and update run it, `rollback_agent/3` goes through `update_agent/3` so a snapshot naming a deleted set is refused rather than restored blind, and `decrypted_for/3` resolves through the tenant-scoped `get_set/2` so a cross-tenant id reads as the account's own default. Both directions have a test, and the revert-check you describe is the right one.

## 1. `PlatformInference.gate/4` reads the default set; the provision reads the agent's

`gate/4` -> `has_own?/3` -> `missing_for_model/2` -> `status_for_user/1` -> `get_for_user/1`, which is the row `where is_default`. After this PR, provisioning reads the agent's set instead. So an agent whose credential lives in a non-default set is invisible to the door: on a deployment that holds a platform key for that provider, once the day's ceiling is spent, `start_conversation` refuses with `:platform_inference_unavailable` for a launch that would have run on the tenant's own key.

That is the same failure #2025 removed for tenant secrets, one PR after removing it, and it makes `gate/4`'s own docstring false as written today ("the gate never refuses for a credential the selection would not hand out"). The hosted deployment has held platform keys since 2026-09-03, so the trigger is reachable.

The ingredients are already in the module from #2028: `status_for_set/1` takes a loaded set. `has_own?/3` and `missing_for_model/2` can take an `:inference_credential_id` opt and read `status_for_set(get_set(id, user_id) || get_for_user(user_id))`, and the four `Fountain.PlatformInference.gate(...)` call sites in `conversations.ex` (3192, 3974, 4843, 5154) pass `agent.inference_credential_id` beside the `environment_id` and `vault_id` they already pass.

The opposite direction I checked and am not asking you to change: default set holds the key, agent names an empty set, the gate says `:ok`, and `select/4` falls through to `platform_credential/4` and serves the platform key. The turn genuinely runs on it, `TurnMachine.gate/2` re-checks the ceiling per turn off the resolved `Source`, so nothing escapes the ceiling and the stamp is honest. Worth knowing it happens, though: it is why the PR body's "an agent pointed at an empty set will not reach a model, whatever the default set holds" is not true on a deployment with platform keys. The banner is right, because it asks `select/4`; the sentence is not.

## 2. `conversation_server.ex:753` is untested, and the new stub makes it untestable

"`ConversationServer` passes the agent's set, so a conversation provisions with it" is the point of the PR, and it is one line. No test covers it. The new test file calls `SpriteEnv.load_tenant_state/2` directly with an id it fetched off the agent itself, which passes whatever `ConversationServer` does. And `ConversationServerCase`'s new `decrypted_for/3` stub answers on `user_id` and discards `set_id`, so no existing or future `ConversationServer` test can see the difference either.

Change that line back to `SpriteEnv.load_tenant_state(conv.user_id)` and the suite stays green. Please add one `ConversationServer` test that provisions an agent naming a set holding a distinguishable key and asserts `state.inference_credentials`, with the stub honouring `set_id` — resolving through the real `get_set/2` and `decrypted_for_set/2` (neither is stubbed) keeps the three tests that override `decrypted_for_user/2` working while making the wiring observable.

## Non-blocking

- `decrypted_for_set/2` is public with only two callers, both in this module. It takes a struct nothing re-scopes; the DEK makes it fail closed on another tenant's ciphertext, so this is a naming point rather than a hole, but the repo's rule puts `_unsafe_` on an unscoped function (`Vaults._unsafe_list_secrets/1` is the same shape). Private is the cheaper fix if no later stage needs it.
- The unresolvable-id fallback is silent. Ten lines above, the environment arm logs a warning and provisions without it; this one quietly moves the agent onto the default set's key, which for the reseller in the ADR's Context is the one outcome that must not happen unnoticed. The changeset and `on_delete: :nilify_all` make it hard to reach, so a `Logger.warning` matching the environment arm is enough.
- `Source` gains no `set_id`. ADR decision 2 says it arrives "where there is a set to name", which is here — but it also says a field arrives with its reader, and nothing reads it yet. Fine to leave; worth a line in the stack notes so #9/#10 do not each assume the other added it.

Migration is reversible (`alter/add` plus `create index` under `change`), `nilify_all` is the right choice and is tested, and the index on the FK is there for it.

apps/fountain/lib/fountain/conversations/conversation_server.ex
```diff
@@ -750,7 +750,7 @@ defmodule Fountain.Conversations.ConversationServer do
 
     vault = if conv.vault_id, do: Vaults._unsafe_get_vault(conv.vault_id)
 
-    case SpriteEnv.load_tenant_state(conv.user_id) do
+    case SpriteEnv.load_tenant_state(conv.user_id, agent && agent.inference_credential_id) do
```
BinaryBourbon: This is the line the PR is about, and no test would fail if it went back to `load_tenant_state(conv.user_id)`. It is also where the provision now disagrees with the door: `PlatformInference.gate/4`, a few frames earlier, asked `has_own?/3`, which reads the account's default set, while this reads the agent's.

apps/fountain/test/support/conversation_server_case.ex
```diff
@@ -112,6 +112,16 @@ defmodule Fountain.ConversationServerCase do
     Mimic.stub(Fountain.Crypto, :load_tenant_key, fn _user_id -> {:ok, <<0::256>>} end)
     Mimic.stub(Fountain.InferenceCredentials, :decrypted_for_user, fn _u, _k -> {:ok, %{}} end)
 
+    # `load_tenant_state/2` reads through `decrypted_for/3` since ADR 0053
+    # decision 3, and Mimic does not intercept a module's call to itself — so
+    # stubbing only `decrypted_for_user/2` would leave the real query running
+    # and hand every test an empty credential map. Delegating rather than
+    # answering means a test that overrides `decrypted_for_user/2` (three do)
+    # is still honoured, whether or not the agent names a set.
+    Mimic.stub(Fountain.InferenceCredentials, :decrypted_for, fn u, _set_id, k ->
```
BinaryBourbon: The stub discards `set_id`, so every `ConversationServer` test is blind to which set was asked for, including any future one. Resolving through the real `get_set/2` and `decrypted_for_set/2` (neither is stubbed) would keep the three `decrypted_for_user/2` overrides working and still let a test observe the set.

apps/fountain/lib/fountain/inference_credentials.ex
```diff
@@ -166,8 +166,35 @@ defmodule Fountain.InferenceCredentials do
...
+    case get_set(set_id, user_id) do
+      nil -> decrypted_for_user(user_id, dek)
+      %Credential{} = set -> decrypted_for_set(set, dek)
+    end
+  end
+
+  @doc "The decrypted map of one loaded set, or `%{}` for `nil`."
+  @spec decrypted_for_set(Credential.t() | nil, binary()) ::
+          {:ok, %{atom() => String.t()}} | {:error, :decrypt_failed}
+  def decrypted_for_set(cred, dek) when is_binary(dek) do
```
BinaryBourbon, line 224: Non-blocking: both callers are in this module. It takes a struct nothing re-scopes, so by the repo's rule it is either private or `_unsafe_`-prefixed (`Vaults._unsafe_list_secrets/1` is the same shape). The DEK makes it fail closed, so this is naming, not a hole.

BinaryBourbon (changes requested):
AI-assisted adversarial review — CHANGES_REQUESTED.

Reviewed all 47 changed paths against base `4b96938d47c9f4e3caa20914b0794b8519f876b1`, at exact head `ad615f5e7d2f0276b0018b656cd4e956036a2301`, including real admission/provision/wake/reapply callers, source writers, migrations, tests, trusted base policies/ADRs, and previous human objections. The named-set gate and actor wiring objections now have implementation and regression coverage. Two introduced correctness regressions remain:

1. **P1: an accepted environment reapply strands the conversation.** `Resolver.resolve/4` records environment/vault IDs in the source (resolver.ex:38–45), and `matches/2` compares the entire map (line 63). But `do_reapply_conversation/2` updates the conversation/environment identity without updating or validating that source. Reproduce with a Claude conversation bound to an API-key set, then explicitly reapply to an environment with identical build inputs and only an unrelated `WHO` variable changed: reapply returns `{:ok, updated}`, and the selected credential identity/revision are unchanged, yet the next `SpriteEnv.resolve_inference/4` returns `:inference_source_changed`. The actor subsequently marks the conversation failed; wake cannot repair it. This breaks ADR 0023's supported in-place reapply. Validate the proposed selection and commit compatible source context with the configuration revision under the required lock ordering, or refuse an incompatible change before committing; do not require a new transcript for an unchanged credential. Add an admission-to-reapply-to-runtime regression. Independently reproduced on this head: one diagnostic test, one failure.

2. **P2: rejected channel resumes permanently bind legacy conversations.** At conversations.ex:3221–3223, `check_saved_inference/2` runs before the saved execution allowance check. It calls `InferenceBinding.reserve/2`, which commits `inference_source` even when the request then returns 422 for an unsupported/malformed saved allowance. An unbound legacy conversation therefore acquires today's credential selection on a failed request; a later default change no longer applies even though no resume was admitted. Finish refusal checks before committing a binding, and make the eventual binding reservation part of successful admission. The existing six `SavedAllowanceChannelTest` refusal cases reproduce the mutation both locally and in CI.

Validation: focused source/actor/gate/saved-allowance run: **29 tests, 6 failures** (all six above), plus the independent reapply counterexample. I inspected [exact-head CI failure logs](https://github.com/managoat/fountain/actions/runs/34750540653): formatting fails in conversations.ex, five platform/credit tests retain outdated whole-Source equality expectations, and the deleted-sandbox wake fixture fails at the new binding write, in addition to the six mutation failures. Repair these checks rather than treating earlier test counts as current-head evidence. Secret Scanning passed. No production resources were used. Required CI and maintainer review gates remain in force; existing human threads were left untouched.

BinaryBourbon (changes requested):
AI-assisted adversarial review — REQUEST_CHANGES.

Reviewed the complete current diff (all 53 paths), exact head `02a61afc519075a92cac0ad0e9a369097833c1c0` against base `1b1e8fe8ba68cca2bb2093559ffe912e03548977`, including the original changes, repairs, trusted base policies/ADRs, previous review objections, and affected admission, provisioning, wake/reattach, reapply, channel-resume, turn, mutation, tenant and billing callers. The prior named-set gate, actor wiring, ordinary compatible reapply and refused-channel mutation defects now have working regression coverage. Two issues remain:

1. **P2: a legacy wake can undo a successfully committed reapply.** `apps/fountain/lib/fountain/conversations/inference_binding.ex:21–23` updates the supplied conversation struct without fencing against its persisted configuration revision/current binding. When an existing Claude conversation has `inference_source: nil`, `SpriteEnv.resolve_inference/4` first persists its binding, but `ConversationServer` retains the original nil-bearing `conv`. If a compatible environment reapply commits while that actor is performing sandbox I/O, `Reattachment.prepare_source/5` (line 20) later calls `reserve(conv, old_source)` with that original snapshot. The update restores the old source's environment context while preserving the newly committed conversation environment and configuration revision. The next resolution returns `:inference_source_changed`, so a successful supported reapply strands the conversation even though its credential identity and revision never changed. I reproduced this with the real ConversationServer and real DB resolver/reapply, interleaving reapply at the fake `write_env_file/2` boundary: **1 diagnostic test, 1 failure**, at the assertion that the resulting conversation can resolve its inference source. Fence/reload reservation against current conversation configuration and source under consistent lock ordering; reject or reload a stale wake rather than overwrite newer binding context. Add a legacy-unbound wake/reapply interleaving regression. Simply validating the old credential again does not detect this configuration race.

2. **P2: this PR cannot pass its required static-analysis gate.** `apps/fountain/lib/fountain/conversations.ex:3549` adds the non-nil `where_credential_set/2` clause, but both callers in this exact head always pass nil. Dialyzer reports `pattern_match_cov` (the clause can never match). The newest actual exact-head [CI run](https://github.com/managoat/fountain/actions/runs/34752555575) fails Static analysis with that annotation; this is distinct from the cancelled duplicate run. Keep this stage independently verifiable by moving the dormant clause into the stage introducing non-nil selections, or otherwise correcting the unreachable implementation, without suppressing the gate.

Validation: disposable PostgreSQL 18 database, OTP 28.3/Elixir 1.19.2, exact-head existing focused suites **357 tests, 0 failures** across two runs, plus the failing real-actor counterexample above. I also checked a bound-snapshot control, which passes; the demonstrated race is specifically the legacy nil snapshot retained after initial binding. No provider resources or production credentials were used. I did not rerun the entire policy verification matrix. Required CI, migration/maintainer gates and existing review threads remain in force.

<details><summary>Reviewed changed paths</summary>

- `apps/fountain/lib/fountain/accounts/deletion.ex`
- `apps/fountain/lib/fountain/agents.ex`
- `apps/fountain/lib/fountain/agents/agent.ex`
- `apps/fountain/lib/fountain/conversations/conversation.ex`
- `apps/fountain/lib/fountain/conversations/egress.ex`
- `apps/fountain/lib/fountain/conversations/inference_binding.ex`
- `apps/fountain/lib/fountain/conversations/reattachment.ex`
- `apps/fountain/lib/fountain/conversations/sandbox.ex`
- `apps/fountain/lib/fountain/conversations/sprite_env.ex`
- `apps/fountain/lib/fountain/conversations/turn.ex`
- `apps/fountain/lib/fountain/environments.ex`
- `apps/fountain/lib/fountain/environments/environment.ex`
- `apps/fountain/lib/fountain/environments/secret.ex`
- `apps/fountain/lib/fountain/inference_credentials/credential.ex`
- `apps/fountain/lib/fountain/inference_credentials/resolver.ex`
- `apps/fountain/lib/fountain/inference_credentials/source.ex`
- `apps/fountain/lib/fountain/platform_inference/key.ex`
- `apps/fountain/lib/fountain/vaults.ex`
- `apps/fountain/lib/fountain/vaults/vault_secret.ex`
- `apps/fountain/lib/fountain_web/controllers/fallback_controller.ex`
- `apps/fountain/lib/fountain_web/live/start_live.ex`
- `apps/fountain/priv/repo/migrations/20260912051000_agents_name_a_credential_set.exs`
- `apps/fountain/priv/repo/migrations/20260912052000_conversations_name_a_credential_set.exs`
- `apps/fountain/priv/repo/migrations/20260913120000_bind_inference_sources.exs`
- `apps/fountain/test/fountain/agents_credential_set_test.exs`
- `apps/fountain/test/fountain/conversations/admission_lock_isolation_test.exs`
- `apps/fountain/test/fountain/conversations/bounded_lifecycle_test.exs`
- `apps/fountain/test/fountain/conversations/channel_allowance_lock_order_test.exs`
- `apps/fountain/test/fountain/conversations/conversation_server_broker_test.exs`
- `apps/fountain/test/fountain/conversations/conversation_server_inference_binding_test.exs`
- `apps/fountain/test/fountain/conversations/conversation_server_platform_inference_test.exs`
- `apps/fountain/test/fountain/conversations/execution_allowance_test.exs`
- `apps/fountain/test/fountain/conversations/inference_admission_isolation_test.exs`
- `apps/fountain/test/fountain/conversations/inference_named_set_gate_test.exs`
- `apps/fountain/test/fountain/conversations/inference_reapply_test.exs`
- `apps/fountain/test/fountain/conversations/sprite_env_inference_test.exs`
- `apps/fountain/test/fountain/conversations/termination_attach_order_test.exs`
- `apps/fountain/test/fountain/inference_credential_mutation_test.exs`
- `apps/fountain/test/fountain/inference_source_binding_test.exs`
- `apps/fountain/test/fountain/inference_source_write_order_test.exs`
- `apps/fountain/test/fountain/platform_chatgpt_test.exs`
- `apps/fountain/test/fountain/platform_inference_test.exs`
- `apps/fountain/test/fountain_web/controllers/fallback_controller_test.exs`
- `apps/fountain/test/fountain_web/controllers/saved_allowance_channel_test.exs`
- `apps/fountain/test/support/conversation_server_case.ex`
- `ee/test/fountain/credits_chatgpt_inference_test.exs`

</details>

BinaryBourbon (changes requested):
AI-assisted adversarial review — REQUEST_CHANGES.

Reviewed the complete current diff (all 53 paths), exact head `70bb9504629e956d770574cfbd8cc704dd28b86a` against base `1b1e8fe8ba68cca2bb2093559ffe912e03548977`, including the original changes, repairs, trusted base policies/ADRs, previous review objections, and affected admission, provisioning, wake/reattach, reapply, channel-resume, turn, mutation, tenant and billing callers. The prior named-set gate, actor wiring, ordinary compatible reapply and refused-channel mutation defects now have working regression coverage. One runtime correctness issue remains:

1. **P2: a legacy wake can undo a successfully committed reapply.** `apps/fountain/lib/fountain/conversations/inference_binding.ex:21–23` updates the supplied conversation struct without fencing against its persisted configuration revision/current binding. When an existing Claude conversation has `inference_source: nil`, `SpriteEnv.resolve_inference/4` first persists its binding, but `ConversationServer` retains the original nil-bearing `conv`. If a compatible environment reapply commits while that actor is performing sandbox I/O, `Reattachment.prepare_source/5` (line 20) later calls `reserve(conv, old_source)` with that original snapshot. The update restores the old source's environment context while preserving the newly committed conversation environment and configuration revision. The next resolution returns `:inference_source_changed`, so a successful supported reapply strands the conversation even though its credential identity and revision never changed. I reproduced this with the real ConversationServer and real DB resolver/reapply, interleaving reapply at the fake `write_env_file/2` boundary: **1 diagnostic test, 1 failure**, at the assertion that the resulting conversation can resolve its inference source. Fence/reload reservation against current conversation configuration and source under consistent lock ordering; reject or reload a stale wake rather than overwrite newer binding context. Add a legacy-unbound wake/reapply interleaving regression. Simply validating the old credential again does not detect this configuration race.

The subsequent CI-only repair removes the unreachable channel-set filter from this stage and introduces it with its non-nil callers in #2034. I independently inspected that three-line deletion and the unchanged surrounding callers; the earlier Dialyzer finding is addressed. The repair worker's local Dialyzer log reports **0 errors, passed**; I inspected that log, rather than claiming to have rerun Dialyzer myself. The new exact-head GitHub CI is still running and Detect secrets has passed.

Validation: disposable PostgreSQL 18 database, OTP 28.3/Elixir 1.19.2. On this current head I reran the real-actor counterexample with affected channel/allowance regressions: **16 tests, 1 failure**, solely the reproduced legacy wake/reapply defect above. Before the three-line deletion, the full changed focused suites returned **357 tests, 0 failures** across two runs; a bound-snapshot control also passed. The demonstrated race specifically requires the legacy nil snapshot retained after initial binding. No provider resources or production credentials were used. I did not rerun the entire policy verification matrix. Required CI, migration/maintainer gates and existing review threads remain in force.

<details><summary>Reviewed changed paths</summary>

- `apps/fountain/lib/fountain/accounts/deletion.ex`
- `apps/fountain/lib/fountain/agents.ex`
- `apps/fountain/lib/fountain/agents/agent.ex`
- `apps/fountain/lib/fountain/conversations/conversation.ex`
- `apps/fountain/lib/fountain/conversations/egress.ex`
- `apps/fountain/lib/fountain/conversations/inference_binding.ex`
- `apps/fountain/lib/fountain/conversations/reattachment.ex`
- `apps/fountain/lib/fountain/conversations/sandbox.ex`
- `apps/fountain/lib/fountain/conversations/sprite_env.ex`
- `apps/fountain/lib/fountain/conversations/turn.ex`
- `apps/fountain/lib/fountain/environments.ex`
- `apps/fountain/lib/fountain/environments/environment.ex`
- `apps/fountain/lib/fountain/environments/secret.ex`
- `apps/fountain/lib/fountain/inference_credentials/credential.ex`
- `apps/fountain/lib/fountain/inference_credentials/resolver.ex`
- `apps/fountain/lib/fountain/inference_credentials/source.ex`
- `apps/fountain/lib/fountain/platform_inference/key.ex`
- `apps/fountain/lib/fountain/vaults.ex`
- `apps/fountain/lib/fountain/vaults/vault_secret.ex`
- `apps/fountain/lib/fountain_web/controllers/fallback_controller.ex`
- `apps/fountain/lib/fountain_web/live/start_live.ex`
- `apps/fountain/priv/repo/migrations/20260912051000_agents_name_a_credential_set.exs`
- `apps/fountain/priv/repo/migrations/20260912052000_conversations_name_a_credential_set.exs`
- `apps/fountain/priv/repo/migrations/20260913120000_bind_inference_sources.exs`
- `apps/fountain/test/fountain/agents_credential_set_test.exs`
- `apps/fountain/test/fountain/conversations/admission_lock_isolation_test.exs`
- `apps/fountain/test/fountain/conversations/bounded_lifecycle_test.exs`
- `apps/fountain/test/fountain/conversations/channel_allowance_lock_order_test.exs`
- `apps/fountain/test/fountain/conversations/conversation_server_broker_test.exs`
- `apps/fountain/test/fountain/conversations/conversation_server_inference_binding_test.exs`
- `apps/fountain/test/fountain/conversations/conversation_server_platform_inference_test.exs`
- `apps/fountain/test/fountain/conversations/execution_allowance_test.exs`
- `apps/fountain/test/fountain/conversations/inference_admission_isolation_test.exs`
- `apps/fountain/test/fountain/conversations/inference_named_set_gate_test.exs`
- `apps/fountain/test/fountain/conversations/inference_reapply_test.exs`
- `apps/fountain/test/fountain/conversations/sprite_env_inference_test.exs`
- `apps/fountain/test/fountain/conversations/termination_attach_order_test.exs`
- `apps/fountain/test/fountain/inference_credential_mutation_test.exs`
- `apps/fountain/test/fountain/inference_source_binding_test.exs`
- `apps/fountain/test/fountain/inference_source_write_order_test.exs`
- `apps/fountain/test/fountain/platform_chatgpt_test.exs`
- `apps/fountain/test/fountain/platform_inference_test.exs`
- `apps/fountain/test/fountain_web/controllers/fallback_controller_test.exs`
- `apps/fountain/test/fountain_web/controllers/saved_allowance_channel_test.exs`
- `apps/fountain/test/support/conversation_server_case.ex`
- `ee/test/fountain/credits_chatgpt_inference_test.exs`

</details>

apps/fountain/lib/fountain/accounts/deletion.ex
```diff
@@ -119,7 +119,7 @@ defmodule Fountain.Accounts.Deletion do
       }
     })
 
-    case Repo.delete(user) do
+    case Fountain.InferenceCredentials.with_source_lock(user.id, fn -> Repo.delete(user) end) do
```
BinaryBourbon: [P2] Avoid upgrading the platform lock inside account deletion

`with_source_lock/2` holds shared `inference:platform` then exclusive `inference:<user>`. The delete can cascade to a personal `platform_chatgpt_account`, or nilify `platform_inference_keys.updated_by_user_id`; the new trigger then requests the global lock exclusively. A concurrent same-user source operation holds global shared while waiting for this tenant lock, forming a cycle. I reproduced PostgreSQL 40P01 through the actual `Deletion.delete_user/2` for both FK paths using two independent connections; a user with neither reference passes. Take the needed exclusive global lock before tenant/row locks, or eliminate the upgrade, and add both concurrency regressions.

BinaryBourbon (changes requested):
AI-assisted adversarial panel review — REQUEST_CHANGES.

Reviewed current head `f638c24f0bbacfd2caabfcfcb01ab74a1f7cbdbc` against base `1b1e8fe8ba68cca2bb2093559ffe912e03548977`, including all 54 changed paths, prior objections, trusted base guidance and affected callers. The legacy wake/reapply repair now preserves the newer configuration and supports a subsequent fresh wake. The named-set gate, real-actor wiring, refused-channel mutation and earlier CI findings are addressed.

**P2: account deletion upgrades the global source lock after taking the tenant lock, creating a deadlock with normal source operations.** At `apps/fountain/lib/fountain/accounts/deletion.ex:122`, `Repo.delete(user)` runs inside `with_source_lock/2`, which holds shared `inference:platform` followed by exclusive `inference:<user>`. Deleting a user with a personal ChatGPT grant cascades into `platform_chatgpt_account`; deleting an operator referenced by `platform_inference_keys.updated_by_user_id` updates that row through FK nilification. Both execute the new trigger at migration `20260913120000_bind_inference_sources.exs:38–39`, which requests the global lock exclusively.

A concurrent same-tenant source operation can already hold the shared global lock while waiting for deletion's tenant lock. Deletion then waits for that operation's global lock: a real lock cycle. The diagnostic calls the actual `Deletion.delete_user/2`, pauses immediately after its tenant lock, starts a second connection through `InferenceCredentials.with_source_lock/2`, waits until it is blocked, then releases deletion. Both the personal-grant and operator-attribution cases produced PostgreSQL `40P01 deadlock_detected`; the concurrent source operation was the victim in both observed runs. The no-grant/no-attribution control completed normally. A fourth diagnostic starts actual account deletions for two unrelated users with personal ChatGPT grants; their competing shared-to-exclusive global-lock upgrades also produce 40P01, aborting one deletion. This occurs with current writers; it does not depend on the explicitly unsupported old-writer rolling-upgrade case.

Acquire the appropriate exclusive platform-source lock before tenant or row locks for deletion that can touch these tables, or otherwise redesign the cascade locking to avoid a shared-to-exclusive upgrade. Add independent-connection regression coverage for both cascade paths and the ordinary deletion control. Do not rely on retrying the failed transaction to fix lock ordering.

Validation: the final four-case diagnostic returned **4 tests, 3 failures**, all reproduced deadlocks, with the ordinary deletion control passing. A preceding combined run of 19 existing regression tests plus the first three diagnostic cases returned **22 tests, 2 failures**, solely those diagnostic cases. The independent security/compatibility continuation completed all-path review and **63 targeted tests, 0 failures**, with no additional blocking finding. Earlier workers left passing 61-test security and 42-test compatibility runs. [Exact-head CI passed](https://github.com/managoat/fountain/actions/runs/34754496717), including static analysis, all six test partitions, coverage, release/API contract and SDK checks; this concurrency counterexample is outside the existing suite.

This is a requested human-account review assisted by independent agents, not a claim that the Review Loop service ran its complete verification matrix. Required migration/maintainer gates and existing review threads remain in force. No production resources were used, no code was changed during review, and no merge or thread resolution was performed.

apps/fountain/lib/fountain/conversations.ex
```diff
@@ -1800,10 +1857,35 @@ defmodule Fountain.Conversations do
...
+                inference_source: c.inference_source
+              },
               lock: "FOR UPDATE"
           ) || Repo.rollback(:sandbox_unavailable)
 
+        if conv.user_id != user_id, do: Repo.rollback(:sandbox_unavailable)
+
+        # A serving actor must agree with the persisted binding. Hold the source
+        # lock through insertion so replacement cannot race turn admission.
+        if Map.has_key?(attrs, :inference_source) and
```
BinaryBourbon, line 1873: [P2] Fence autonomous turns against the serving actor's source

This source check is bypassed when `:inference_source` is absent. `ConversationServer.open_autonomous_turn/1` → `Connection.open_autonomous_turn/3` supplies neither that field nor the actor's configuration revision. After reapply commits but before actor refresh, output from the old peer therefore creates a turn stamped with the new persisted source at line 1887. I reproduced a real encrypted-key ACP actor still using Sonnet opening an autonomous turn attributed to Opus in that interval. Pass the actor source/revision through this caller and reject stale callbacks before inserting the turn; add the real-actor interleaving and current-configuration control.

BinaryBourbon (changes requested):
AI-assisted adversarial panel review — REQUEST_CHANGES.

Reviewed exact head `19f4d63b3dfe48f5b353a5dab8cfe1af55b07789` against base `1b1e8fe8ba68cca2bb2093559ffe912e03548977`. Independent correctness, security and compatibility reviewers inspected the full 55-path diff, prior objections, trusted base guidance and relevant callers. The account-deletion lock upgrade is fixed: exclusive platform locking precedes tenant locking, and the four independent-connection regression cases pass. The earlier named-set gate, actor wiring, channel refusal, legacy wake/reapply and CI findings are also addressed.

**P2: background turns can adopt a newly reapplied source while their runtime still uses the old configuration.** The new guard in `apps/fountain/lib/fountain/conversations.ex:1873–1887` checks actor agreement only when `attrs` includes `:inference_source`, then unconditionally stamps the persisted source on the new turn. Normal prompt admission supplies the actor's source and configuration revision. `ConversationServer.open_autonomous_turn/1` calls `Connection.open_autonomous_turn/3`, which supplies neither. That path rereads the newly committed conversation instead of checking the existing peer's configuration.

There is a real interval between the reapply transaction committing and `announce_reapply/2` refreshing the actor. A background ACP notification from the old peer during that interval opens an autonomous turn with the new source/model. The source-attribution promise introduced by this PR is therefore false for autonomous turns even though the normal prompt path refuses a stale actor.

Reproduction on this exact head: start a real ConversationServer/ACP peer with a real encrypted test Anthropic credential, complete a Sonnet turn, update the agent to Opus and explicitly reapply. At the existing post-commit audit boundary, deliver an ACP `agent_message_chunk` before the refresh message. The actor still retains the Sonnet source, but the newly opened autonomous turn is stamped with the Opus source; credential identity and revision are otherwise unchanged. The isolated real-key diagnostic returned **1 test, 1 failure**, specifically the assertion that this stale callback must not open a turn. Provider transport and title generation are faked; resolution, persistence, reapply and the actor/ACP path are real.

Pass the actor's `configuration_revision` and `inference_source` through the autonomous-turn caller into the existing locked admission checks, and refuse/drop stale background output before creating a turn. Add a real-actor post-commit/pre-refresh regression, plus a current-configuration autonomous-turn control. Do not infer the old peer's source from a freshly loaded conversation.

Validation: fresh independent focused runs passed **35 correctness tests**, **64 security tests** and **71 compatibility tests**; these overlap and are not a count of unique tests. The additional real-actor diagnostic above fails. [Exact-head CI passed](https://github.com/managoat/fountain/actions/runs/34755816950), including all six partitions, coverage, static analysis, release/API contract and SDK checks. Green CI does not cover this interleaving.

This is the explicitly requested human-account review assisted by independent agents, not a claim that Review Loop ran its complete service verification matrix. Required CI, migration/maintainer gates and existing review threads remain in force. No source repairs, production provider resources, merges or thread resolutions were performed during this review.

BinaryBourbon (approved):
AI-assisted adversarial panel review — APPROVED.

Reviewed exact head `0fd482c09fa74db7920a591062838b08de32cc8c` against base `1b1e8fe8ba68cca2bb2093559ffe912e03548977`. Three independent reviewers completed fresh reviews of all 61 changed paths, covering correctness/concurrency, security/tenant isolation, and API compatibility/simplicity. All three returned APPROVE with no remaining medium-or-higher findings. The panel inspected the complete PR, prior objections, approved-base guidance and affected callers.

The previous autonomous-turn blocker is addressed: the serving actor now passes its sandbox, configuration revision and source through mandatory `Connection.open_autonomous_turn/5`. Locked admission compares both fences before insertion, including an explicitly nil source. Refusal closes the old connection and halts subsequent output/permission effects. Model-changing and revision-only post-commit/pre-refresh regressions pass, and matching-current-peer background turns still work. The original round-4 real-key diagnostic was rerun unchanged on this head: **1 test, 0 failures**.

The earlier account-deletion lock cycle, legacy wake/reapply race, refused-channel mutation, selected-set gate/actor wiring and static-analysis requests also remain addressed. Exclusive platform locking precedes tenant locking for account deletion; current-row reservations preserve newer configuration; rejected channel resumes leave labels and source binding unchanged. The earlier helper-naming observation remains nonblocking.

Independent local verification used separate exact-head clones and fresh PostgreSQL 18 databases with Elixir 1.19.2 / OTP 28.3:

- Correctness: **240 tests, 0 failures**, plus the unchanged diagnostic above.
- Security: **181 tests, 0 failures**, including second-tenant denials, credential replacement, source write ordering and four account-deletion concurrency cases.
- Compatibility: **162 tests, 0 failures**, plus SDK wire-contract freshness: **174 operations, 202 schemas**.

These suites overlap and are not a count of unique tests. [Exact-head CI passed](https://github.com/managoat/fountain/actions/runs/34757044609), including all six test partitions, coverage, static analysis, release/API contract and SDK checks. [Secret scanning passed](https://github.com/managoat/fountain/actions/runs/34757044432).

This is the explicitly requested human-account review assisted by independent agents. It does not claim that Review Loop ran its complete service verification matrix; its dispatch was skipped. Required migration/maintainer gates and existing review threads remain in force. Provider boundaries used fakes; no production provider resources were used. No code changes, merge, review dismissal or thread resolution were performed during this review.

PR #2034, opened 2026-09-12
feat(conversations): add source-bound per-launch credential selection

jhgaylor:
A launch may override its agent's credential set within the agent's allowlist. The resolved source stays pinned through channel reuse, wake, reprovisioning, and turn admission. Explicit channel selections choose the matching conversation; omitted selections retain the existing binding. Deleted or replaced sources refuse rather than falling back to today's default.

Enables the request inputs on the binding foundations in #2031, adds the launch/channel regression coverage, and preserves the selection for queued requests.

Validation: 21 launch/gate tests pass; formatting, compilation, and pre-API contract freshness pass.

Base: #2031. Part of #2018; next #2042.


BinaryBourbon (changes requested):
**Request changes.** The resolver, the allowlist and the tenant scoping are right and well tested. What is missing is the other half of ADR 0053 decision 6: this is the PR that first lets two conversations on one machine hold different sets, and the refusal decision 6 says must stand until ADR 0052 decision 5 lands is not here.

## What I checked

- `resolve_inference_credential_id/3` against `resolve_environment_id/3`: same three clauses, same allowlist-then-scoped-fetch order, same "the agent's own set is not an override" clause. No divergence.
- Tenant scoping: `InferenceCredentials.get_set/2` is `Repo.get_by(Credential, id:, user_id:)`. A guessed foreign id returns `nil` -> `:inference_credential_not_found`, and behind an empty allowlist it is refused before the fetch. Both asserted. `status_for/2` and `decrypted_for/3` make the same nil-to-default fallback, so an unresolvable id can never report on or run on another account's credential.
- All four `PlatformInference.gate` call sites take `:credential_set_id`; `gate/4` -> `has_own?/3` -> `missing_for_model/3` -> `status_for/2` threads it, and `secret_keys/2` ignores the extra key.
- Migration: nullable, `:nilify_all`, one index, reversible `change`, no backfill needed. The delete test asserts the transcript survives.
- Home identity: `_unsafe_find_home/4` and `check_attachable/4` are unchanged, so the tuple stays `(user, agent, environment, vault)`. Decision 6 holds structurally.

## 1. Decision 6's codex carve-out is not built, and this PR is what makes it reachable (blocking)

ADR 0053 decision 6, at the head of this branch:

> One runtime does not fit. Codex writes an account file into `$CODEX_HOME`, so two peers with different sources on one machine overwrite each other. ADR 0052 decision 5 already requires separate auth locations for exactly this... **until it lands, a codex agent whose set differs from another live codex peer's on the same machine is refused at admission rather than served the wrong account.**

Before this PR that divergence was unreachable: co-tenants on a home share an agent by construction, so with #2031 alone they share the agent's set. This PR adds `resolve_inference_credential_id/3` to `attach_conversation/3` (conversations.ex:3972) and stores the result on the attached row, so the second conversation on a home may now name a different set. `check_attachable/4` (conversations.ex:4130), which is the admission check for that shared machine, still compares only agent, vault, environment and runtime.

The mechanism is live on the wake path, not only at provision: `ConversationServer` re-runs `Provisioning.prepare_runtime_sprite/5` on every reattach (conversation_server.ex:1192, with a comment saying it does so precisely because codex's `auth.json` is written at provisioning), and `Managoat.Runtimes.Codex.prepare_sandbox/3` rewrites `~/.codex/auth.json` from that spawn's `OPENAI_API_KEY`. Two codex conversations on one home with different sets take turns rewriting one machine-global file.

The PR body says the sharing "is only safe because #2027 took the credential off the shared disk". That is true of `/home/sprite/.env` and the four `@process_only` names; it is not true of the file codex derives from them, which is the case decision 6 carves out by name.

What I would like to see: either the admission refusal, next to `check_attachable/4` where the other identity mismatches live (refuse when the sandbox already has a live codex co-tenant whose effective `SpriteEnv.credential_set_id/2` differs), with a test; or, if it belongs to a later stage, name that stage in the PR body and file it. Nothing in stages 9-12 covers admission, so otherwise this ships as decision 6 "implemented".

## 2. The channel-resume key drops the set (blocking, or say why)

`find_channel_conversation/5` (conversations.ex:3122) keys on `(user, agent, vault, environment, channel)`, and its comment gives the reason the environment is in there:

> So is the environment override (#783): an identity that switches environments must not resume a conversation provisioned from the old one.

The credential set is now exactly that kind of override, and `start_or_resume_conversation/2` (conversations.ex:2893-2894) and `channel_conversation/1` (conversations.ex:3113-3114) neither key on it nor resolve it. A channel caller that names a new set silently resumes the conversation running on the old one, and the allowlist check never runs on that path at all. The PR body lists these two call sites but marks only the third, the home-race retry, as deliberate. The home-race one genuinely is: it only feeds `_unsafe_find_home/4`, and `attach_conversation/3` resolves the set properly afterwards. These two need either the set in the key or a comment saying why #783's argument stops here.

## Non-blocking

- `PlatformInference.gate/4`'s docstring still reads "`opts` names this launch's `:environment_id` and `:vault_id`". It is three now.
- The gate's new `:credential_set_id` has no test. It is the change the PR body says prevents "PR 8's own refusal bug", and reverting `status_for/2` to `status_for_user/1` would leave the whole suite green. A test that a named set holding the key stops the gate refusing, where the default set holds none, would pin it.
- Nothing asserts decision 6 itself, two conversations differing only in set landing on one home. Given finding 1, that test is worth having whichever way the refusal goes.
- The precedence rule has "one home" in `SpriteEnv.credential_set_id/2`, but the two door gates re-implement it inline as `cred_set_id || agent.inference_credential_id` (conversations.ex:3198, 3984). `SpriteEnv.credential_set_id(%{inference_credential_id: cred_set_id}, agent)` would make the claim literally true.
- `do_reapply_conversation/2` does not carry the set, so a reapply cannot move a conversation between sets. Reasonable seam, currently unstated.

BinaryBourbon (changes requested):
AI-assisted adversarial review — CHANGES_REQUESTED.

Reviewed all three changed paths against base `ad615f5e7d2f0276b0018b656cd4e956036a2301`, at exact head `1288aba8161d7f81d49ee2e9725c69c54fcaf53e`, including authenticated controller entry, queued attributes, channel lookup/resume, allowlists, shared Codex reservation, and previous human objections. Explicit channel selections now participate in lookup, and the base's identity/revision reservation addresses the prior incompatible-Codex admission objection.

**P2: malformed JSON credential IDs crash the public launch endpoint.** `resolve_inference_credential_id/3` at conversations.ex:4972–4982 handles only nil, an empty string, or a binary. This PR starts reading the new field, while this exact head's request schema does not declare or validate its type. An authenticated JSON `POST /api/conversations` with an owned `agent_id` and `"inference_credential_id": 123` reaches this function and raises `FunctionClauseError` instead of returning a validation refusal. I reproduced this through the real router/controller, without stubbing the resolver. Add a total invalid-input clause returning the stable refusal, or introduce request validation in the same enabling change; cover numeric, boolean, array, and object inputs, including channel resumes. The later API stage does not make this intermediate head safe.

Validation: existing launch suite plus the independent authenticated malformed-ID regression: **20 tests, 1 failure** (the new counterexample; the 19 existing launch tests pass). I also inspected [exact-head CI failure logs](https://github.com/managoat/fountain/actions/runs/34750541064): the saved-allowance mutation, Source equality assertions, and deleted-sandbox wake failure originate in #2031 and are inherited here, not attributed to this diff. The separately reported #2031 reapply regression is inherited as well. Secret Scanning passed. No production resources were used. Required CI and maintainer review gates remain in force; existing human threads were left untouched.

BinaryBourbon (approved):
AI-assisted adversarial review — APPROVE for this PR's source changes.

Reviewed the entire current diff at exact head `c9de53080d7e142d5f5fa43fe347de265daa48c2` against base `70bb9504629e956d770574cfbd8cc704dd28b86a`, including all original changes, repairs, previous review objections, and the final restack delta. All four changed paths were reviewed:

- `apps/fountain/lib/fountain/conversations.ex`: fresh/attached admission, channel lookup/resume, explicit selection, nil/empty selection, allowlists, tenant-scoped lookup, malformed IDs, and the non-nil channel filter now introduced with its callers.
- `apps/fountain/lib/fountain_web/controllers/conversation_controller.ex`: authenticated launch entry and retaining the explicit set in queued attributes through the actual queue drainer.
- `apps/fountain/test/fountain/conversations/launch_credential_set_test.exs`: selection precedence, missing/foreign sets, allowlist narrowing, default changes/deletion, channel identity, source snapshots, stale actor refusal, and Codex shared-machine reservation.
- `apps/fountain/test/fountain_web/controllers/conversation_credential_input_test.exs`: real-router refusals of malformed UUID, numeric, boolean, array and object IDs, with and without an existing channel, before mutation or worker startup.

The earlier malformed-input crash is fixed by the total invalid-input clause and verified through the real endpoint. Explicit channel selections are checked before lookup/resume and select the requested bound conversation; omission preserves the saved source. The Codex incompatibility check is provided by the base's durable identity/revision reservation, reached by this PR's admission paths. No new blocking defect found in this diff.

Validation on this exact head: **33 tests, 0 failures** — the 31 changed regression cases plus two independent real-router/queue/drainer diagnostics. Those additional diagnostics verify that a queued explicit set survives replay and that a subsequently narrowed allowlist refuses replay before creating a conversation. The final head's tree is identical to the earlier fully reviewed `da9bbb0ba3d032fa92a0b92d7f5bd54c847184e2`; the extra filter clause in this diff now belongs to the stage that makes it reachable.

The [newest actual exact-head CI run](https://github.com/managoat/fountain/actions/runs/34753434058) is still pending; Secret Scanning passed. The failed aggregate from cancelled duplicate run `34753433942` is not evidence of a failed test. I did not run the entire policy verification matrix locally.

The inherited legacy wake/reapply race remains a blocking predecessor finding in [the current-head #2031 review](https://github.com/managoat/fountain/pull/2031#pullrequestreview-5190474535); this approval does not claim that defect is fixed or that the stack is ready to merge. Required CI and maintainer gates remain in force. Existing review threads were left untouched; no production/provider resources were used.

PR #2036, opened 2026-09-12
fix: untrack the deps symlink and ignore it by name (#2035)

### LINKED ISSUE #2035: deps is a tracked symlink on main pointing into a local scratchpad path
## Symptom

`deps` is a **tracked symlink on `main`** pointing at an absolute path inside one machine's session scratchpad:

```
$ git ls-tree origin/main deps
120000 blob c220e189dd8fd10449d71c1c1f101c6c783e1a82	deps

$ git cat-file -p c220e189
/private/tmp/claude-501/-Users-jake-dev-BinaryBourbon-fountain/9b86f823-5f5f-46ba-b9d9-dd55978fc91f/scratchpad/wt1744/deps
```

Every fresh clone, every new worktree and every CI checkout materialises `deps` as a symlink to that path. On the machine that created it, it resolves into another agent session's worktree — a directory that can be deleted at any moment. Anywhere else it is dangling.

It is the only tracked symlink in the repository.

## How it got past `.gitignore`

`.gitignore:14` is `/deps/`. The trailing slash makes the pattern **directory-only**, and `deps` here is a symlink, not a directory — so git never ignored it and a `git add -A` in a worktree that had the symlink planted staged it silently.

This is the [never `git add -A` in a review worktree] lesson landing on `main`: the same shape as the `node_modules` symlink incident on #1048, except that one was caught before merge.

## Introduced by

`7b7fdf5b` — "Prepare bounded command transport and guarded ACP writes (#1748)". The commit is otherwise legitimate; `deps | 1 +` sits among ten real files and reads as noise in the stat.

## Impact

Not yet observed breaking a build, which is part of why it survived — but it is wrong for every consumer:

- a contributor cloning the repo gets a dangling symlink where `mix deps.get` expects to write;
- a new git worktree on this machine gets a symlink into *another session's* scratchpad, so two sessions can share or destroy one another's `deps`;
- the path leaks a local directory layout into the public repository.

## Fix

```
git rm --cached deps
```

and tighten the ignore so this cannot recur: `/deps` (no trailing slash) matches the symlink as well as the directory. Worth checking `_build` has the same shape.

---
Found while rebasing #2016: a fresh worktree checked out `deps` as mode `120000` into another session's path.

jhgaylor:
Closes #2035.

`deps` was a **tracked symlink on `main`** (mode `120000`) whose content is an absolute path inside one machine's session scratchpad:

```
/private/tmp/claude-501/-Users-jake-dev-BinaryBourbon-fountain/9b86f823-.../scratchpad/wt1744/deps
```

So every fresh clone, new worktree and CI checkout materialises a dangling symlink where `mix deps.get` expects to write. On the machine that created it, it resolves into **another agent session's worktree**, which that session can delete at any moment. It was the only tracked symlink in the repository.

## Why `.gitignore` did not stop it

`/deps/` is a **directory-only** pattern — the trailing slash is what makes it so. A symlink named `deps` is not a directory, so git never ignored it and a `git add -A` in a worktree staged it silently. It arrived with #1748, where `deps | 1 +` sits among ten real files and reads as noise in the stat.

## This change

- `git rm --cached deps` — untrack the blob. The working-tree symlink is left alone; `deps` is regenerated by `mix deps.get` as always.
- Drop the trailing slash from `/_build/`, `/deps/`, `/dist/`, `/doc/` and `/tmp/` so each matches a symlink as well as a directory, with a comment recording why.

Verified after the change: `git check-ignore -v deps` reports `.gitignore:17:/deps`, so the same symlink can no longer be staged.

No product change and no user-visible behaviour, so no CHANGELOG entry.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # The directory Mix will write compiled artifacts to.
-/_build/
+/_build
 
 # If you run "mix test --cover", coverage assets end up here. Unanchored: in an
 # umbrella the reports land in apps/*/cover/ as well as at the root, and while
@@ -8,13 +8,16 @@
 cover/
 
 # Local build outputs (CLI binaries in CI, `mix openapi.export` locally).
-/dist/
+/dist
 
-# The directory Mix downloads your dependencies sources to.
-/deps/
+# The directory Mix downloads your dependencies sources to. No trailing slash
+# on these: a `dir/` pattern matches a directory only, so a SYMLINK named
+# `deps` was not ignored and a `git add -A` in a worktree staged one onto
+# main, pointing at a local scratchpad path (#2035).
+/deps
 
 # Where 3rd-party dependencies like ExDoc output generated docs.
-/doc/
+/doc
 
 # Ignore .fetch files in case you like to edit your project deps locally.
 /.fetch
@@ -26,7 +29,7 @@ erl_crash.dump
 *.ez
 
 # Temporary files, for example, from tests.
-/tmp/
+/tmp
 
 # ExUnit's `@tag :tmp_dir` writes under the app that runs the test, not the
 # umbrella root, and the behavioural CA install test uses one.
--- a/deps
+++ b/deps
@@ -1,1 +1,0 @@
-/private/tmp/claude-501/-Users-jake-dev-BinaryBourbon-fountain/9b86f823-5f5f-46ba-b9d9-dd55978fc91f/scratchpad/wt1744/deps


BinaryBourbon (approved):
**APPROVE** — premise verified independently, the ignore widening is safe, and the severity is in fact higher than the PR claims.

## The premise holds

`deps` is a `120000` blob on `main`, and it is the **only** tracked symlink in the repo:

```
$ git ls-tree origin/main deps
120000 blob c220e189dd8fd10449d71c1c1f101c6c783e1a82	deps
$ git cat-file -p c220e189
/private/tmp/claude-501/-Users-jake-dev-BinaryBourbon-fountain/9b86f823-.../scratchpad/wt1744/deps
$ git ls-tree -r origin/main | awk '$1=="120000"'   # deps, and nothing else
```

Provenance checks out: `git log origin/main --diff-filter=A -- deps` names `7b7fdf5b` (#1748), where the diffstat line is `deps | 1 +` among 11 files and 1,026 insertions. The "read as noise" explanation is accurate.

## The mechanism is the right one

I looked for the alternative explanations rather than taking the directory-only story on trust:

- No `git add -f` is needed to explain it. `.git/info/exclude` **also** carries `deps/` — the same directory-only pattern, with the same blind spot — so both exclusion sources failed for one reason.
- No nested `.gitignore` covers it (`apps/fountain/.gitignore` is just `*.coverdata`); `core.excludesfile` is unset and there is no `~/.gitignore`.

So a directory-only pattern really is the whole story.

## The widening is safe — I checked it rather than reasoned about it

Nothing is tracked at any of the five paths except the symlink itself:

```
$ git ls-tree -r --name-only origin/main | grep -E '^(_build|deps|dist|doc|tmp)(/|$)'
deps
```

The one I was most suspicious of was `/doc`, because of `docs/`. Confirmed in a throwaway repo with the new patterns that gitignore matches whole path components, not prefixes:

```
docs/page.md     -> NOT ignored   # the published manual is safe
doc/x.md         -> ignored       # unchanged from /doc/
deps, dist       -> ignored       # newly covers file/symlink, which is the point
```

`/dist` is only ever a build output (CLI binaries in CI, `mix openapi.export`), and nothing writes a committed *file* at a bare root `_build`/`dist`/`doc`/`tmp`/`deps` name. Worth remembering that gitignore never affects already-tracked files anyway, so the only exposure would be a *new* file at one of those exact names — which is not a thing this repo produces.

## Untracking breaks no consumer

- **Dockerfile**: never does `COPY . .`; it copies an explicit list (`mix.exs`, `config`, per-app `mix.exs`, `apps`, `ee`, `docs`, `CHANGELOG.md`, `sdk/typescript/examples`, `cli/`), and `deps` is not on it. `deps/` inside the image comes from `mix deps.get --only prod` at line 69.
- **ci.yml**: every `deps` reference is either `mix deps.get` or `actions/cache` with `path: deps`. Removing a blob that checkout used to materialise there can only help the restore.
- No release or asset step globs the path.

## The severity is understated, not overstated

The PR body says a fresh clone "materialised a dangling link where `mix deps.get` writes". That undersells it — a dangling symlink does not merely sit there, it makes the write **fail**:

```
$ ln -s /nonexistent deps && elixir -e 'IO.inspect File.mkdir_p("deps")'
{:error, :enotdir}
```

So `git clone && mix deps.get` was broken for anyone whose machine is not the one in the blob. Main's CI stayed green throughout (8/8 recent `push` runs), which I read as the `actions/cache` restore overwriting the symlink with a real directory before `deps.get` runs — worth knowing, because it means CI would not have caught this and a cold cache is the case that bites. Good reason to land it.

One nuance on the prose: "every fresh clone, worktree and CI checkout materialised a **dangling** link" is not quite true on the machine that created it, where the target is a live directory. The commit message's next sentence states that correctly ("On the machine that created it, it resolved into another agent session's worktree"), so the precise version is on the record. Not worth a revision.

The `.gitignore` comment and the commit message are otherwise accurate about the mechanism.

## CHANGELOG: agree, no entry

Nothing operator- or user-visible changes — no API, runtime, config or image change (the Dockerfile does not copy the path). This is a contributor-workflow fix, and `[Unreleased]` is for things a self-hoster taking an upgrade would need to know. Correct call to leave it out.

## Two non-blocking notes

1. **Same latent class remains elsewhere in the file.** `/sdk/python/dist/`, `/sdk/python/build/`, `/burrito_out/`, `/.build/`, `/.swiftpm/` and `priv/plts/` are still directory-only and would miss a symlink the same way. Genuinely lower risk (nobody runs `git add -A` from `sdk/python/`), and widening them is scope creep on a fix PR. Flagging only so the choice is deliberate.
2. **After merge, five worktrees lose their `deps`.** Of the 171 worktrees on this checkout, 5 currently have `deps` as the symlink; when they pull the deletion, git unlinks the symlink (it does not touch the target) and those worktrees will need their own `mix deps.get`. Benign, just not silent.

Unrelated and out of scope: `apps/fountain/test/fountain/runners_test.exs:18` carries `"/Users/jake/.fountain/runners/mini"`, the only other machine-specific absolute path in a tracked file. It is an inert fixture string in a JSON payload with no filesystem access, so it is harmless — noting it only because I swept for exactly this.

CI is green: all 30 checks pass, including `CI required`, all six partitions, the coverage gate, the release boot check and the compose fresh-clone check.

Issue #2032, opened 2026-09-12
The #1766/#1767 campaign lands ~72 lines over the conversation_server pin, and no PR in it is red


jhgaylor:
## Symptom

Every PR in the #1766/#1767 campaign is green, and the campaign cannot land: merging it onto current `main` puts `conversation_server.ex` over the size pin. No individual PR shows this, because each is measured against a base where the pin is 2762.

## Measured

`main` is at `@pin 2646` with the file at exactly 2646 lines — **zero headroom** — after #1749 (2762 → 2652) and #1751 (2652 → 2646).

Merging the #1979 chain tip `c18710cf` into `origin/main` auto-merges cleanly, with no conflict on the pin line (the branch never edited it, so main's lowering wins silently), and then fails:

```
apps/fountain/lib/fountain/conversations/conversation_server.ex is 2673 lines,
over the pin of 2646.
```

The growth is spread across the campaign, not concentrated in one PR. From the merge-base at 2685:

| Branch | server lines | delta |
|---|---|---|
| merge-base with `main` | 2685 | — |
| `p2-1766-provision-retirement` (#1943) | 2691 | +6 |
| `p2-1766-wake-retirement` (#1944) | 2698 | +7 |
| `p2-1766-provision-start` (#1945) | 2707 | +9 |
| `p2-1767-wake-reset-fence` (#1970) | 2712 | +5 |
| `p2-1767-lifecycle-fence` (#1975) | 2722 | +10 |
| `p2-1767-teardown-intent` (#1983) | 2737 | +15 |
| `p2-1767-termination-actor-fence` (#1980) | 2740 | +3 |
| `p2-1767-orphan-binding` (#1996) | 2741 | +1 |
| `p2-1767-machine-gone-receiver` (#2006) | 2743 | +2 |
| `p2-1767-interruption-binding` (#2000) | 2750 | +7 |
| `p2-1767-termination-client` (#1981) | 2753 | +3 |
| `p2-1767-command-exit-binding` (#2003) | 2755 | +2 |
| `p2-1767-startup-failure-binding` (#2002) | 2757 | +12 |

**+72 lines across roughly a dozen PRs**, all already approved. Landed on a `main` whose pin sits at the file's exact length, the campaign ends ~72 lines over.

## Cause

Exactly what this test's own moduledoc warns about, one campaign later:

[...]

PR #2037, opened 2026-09-12
refactor(conversations): move the provision watchdog out of the server

jhgaylor:
`start_provision_watchdog/2` and its `@provision_deadline_ms` ceiling become `Fountain.Conversations.ProvisionWatchdog`, the #1369 subtraction pattern. The function body moves verbatim (a plain call from the server process, so `self()` still names the server it monitors), taking 81 lines out and putting `conversation_server.ex` at 2565.

## Why now

This is the shrink #2032 needs. `main` sits at `@pin 2646` with the file at exactly **2646** — no room for the next line anyone adds — while the #1766/#1767 retirement and binding campaign is measured at **+68 lines across its 26 open PRs**:

| PR | net lines to `conversation_server.ex` |
|---|---|
| #1943 +6 · #1944 +7 · #1945 +9 | +22 (stack:1766) |
| #1970 +5 · #1975 +10 · #1980 +28 · #1981 +13 · #1982 −16 · #1996 +4 · #1998 +9 · #2002 +7 · #2003 −2 · #2006 −12 | +46 (stack:1767) |
| **total** | **+68** |

Nothing in that campaign is red. Each PR is green against a base whose pin is still 2762, the merges conflict on nothing (the branches never touched the pin line, so main's lowering wins silently), and the assertion is `count <= @pin`, so it fails only at the moment the last one lands. Verified by cherry-picking the stack:1766 chain onto `origin/main` in a scratch worktree: clean merge, compiles, every behavioural test passes, and the size test fails at 2668.

## The pin goes to 2641, not 2565

The number has not stopped moving, and the size test's own guidance is to lower it in a follow-up once it has. 2633 is where the campaign lands; the remaining 8 lines cover #1977 and #1981, still in review rounds. It only ever moves down: 2646 → 2641, then → 2633 in the follow-up.

## Verification

- The moved body is byte-identical to the original modulo the signature line and indentation (diffed whitespace-normalized).
- `mix credo --strict` clean. The two `_unsafe_` reads keep their ownership statements inline rather than growing `.credo.exs`'s exempt list — the watchdog is not a GenServer or a system sweep, it reads exactly the two ids the server handed it.
- `mix format --check-formatted`, `MIX_ENV=test mix compile --warnings-as-errors` clean.
- Size and broker tests pass. `conversation_server_provision_deadline_test.exs` is flaky in this environment independent of this change — 0, 1 and 2 failures across three identical runs, and it fails on unmodified `main` at the same seed. Filed separately.

Refs #1369, #2032, #1766, #1767.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -22,17 +22,9 @@ defmodule Fountain.Conversations.ConversationServer do
 
   alias Fountain.Conversations.{BoundedTurn, CallbackKey, Checkpoints, Connection}
   alias Fountain.Conversations.{Conversation, DetachedRequest, Egress}
-  alias Fountain.Conversations.{Lifecycle, McpServers, Output, Pending, Provisioning, Reapply}
+  alias Fountain.Conversations.{Lifecycle, McpServers, Output, Pending, Provisioning}
+  alias Fountain.Conversations.{ProvisionWatchdog, Reapply}
   alias Fountain.Conversations.{Reattachment, Redaction, SpriteEnv, TurnLaunch, TurnMachine}
-
-  # Absolute ceiling on provisioning (#329). Generous against the summed
-  # default step timeouts (packages 300s + clone 600s + setup 120s). Setup
-  # may opt into up to 900s, but this overall ceiling still applies. It also
-  # catches a step that stalls without raising — the case where
-  # the row sat in `starting` holding a quota slot until the next deploy:
-  # the reaper exempts rows whose server is alive, and the server's own
-  # timers queue behind the stuck handle_continue. Overridable in tests.
-  @provision_deadline_ms :timer.minutes(30)
 
   # ── public api ────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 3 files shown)


BinaryBourbon (approved):
Approving. Verified independently rather than from the description.

**The move is behaviour-preserving.** Diffed the relocated body against `origin/main`'s lines 615-676 whitespace-normalized: identical apart from the signature. `start/2` is still a plain call from the server process, so `self()` names the same server the old `defp` monitored, and `@provision_deadline_ms` keeps the same `Application.get_env` override path the deadline test drives.

**The premise checks out.** Cherry-picked stack:1766's four commits onto `origin/main` in a scratch worktree: clean merge, compiles, all 61 behavioural tests in the three touched files pass, and the only failure is the size test at 2668 > 2646. So the campaign really is blocked, and it really is invisible — the branches never edit the pin line, so main's lowering wins the merge silently instead of conflicting.

**The +68 is right.** Re-derived by summing the per-PR net delta to `conversation_server.ex` across all 26 open PRs in the two stacks, not from #2032's body (which says +72 and is stale).

**On the pin at 2641 rather than 2565.** This is the part worth stating plainly, because it reads as slack: with `count <= @pin` and the pin at the file's own length, *no* PR can ever add a line, so any campaign that grows the server is dead on arrival. That is not a hypothetical — it is #1565, then #1749/#1751, three times now. Sizing the pin to where the file is going, and tightening in a follow-up, is what the moduledoc's own "land the shrink first, then lower the pin in a follow-up" already asks for. It only moves down.

**Gates:** `credo --strict` clean (the two `_unsafe_` reads carry inline ownership statements rather than a new `.credo.exs` exemption, which is the more precise answer for a process that is neither a GenServer nor a sweep), `format --check-formatted` clean, `compile --warnings-as-errors` clean, size and broker tests pass. CI is green end to end including all six partitions and the coverage gate.

One note for the follow-up, not blocking: `conversation_server_provision_deadline_test.exs` is flaky in a local worktree independent of this change (0, 1 and 2 failures across three identical runs, and different tests of the pair failing at the same seed on unmodified `main`). Filed as #2040 so the next person to see it red does not attribute it here.

PR #2038, opened 2026-09-12
fix: remove the committed deps symlink

jhgaylor:
#1748 committed a root `deps` symlink pointing at a scratchpad worktree on one machine:

```
deps -> /private/tmp/claude-501/-Users-jake-dev-BinaryBourbon-fountain/9b86f823-.../scratchpad/wt1744/deps
```

`/deps/` is in `.gitignore`, so the tree was never meant to be tracked; the symlink slipped past because git records it as a file entry, not a directory.

Every checkout but the one it was made on gets a dangling link where mix expects the dependency tree, and `mix deps.get` then writes through it into `/private/tmp` rather than the repo. Nothing goes red — the path is created on demand and the build succeeds against it — which is why it survived a full CI run.

```
$ git ls-tree origin/main deps
120000 blob c220e189dd8fd10449d71c1c1f101c6c783e1a82	deps
```

### DIFF
--- a/deps
+++ b/deps
@@ -1,1 +1,0 @@
-/private/tmp/claude-501/-Users-jake-dev-BinaryBourbon-fountain/9b86f823-5f5f-46ba-b9d9-dd55978fc91f/scratchpad/wt1744/deps


BinaryBourbon (approved):
Approving. Confirmed the entry is real and is what the description says:

```
$ git ls-tree origin/main deps
120000 blob c220e189dd8fd10449d71c1c1f101c6c783e1a82	deps
$ git show origin/main:deps
/private/tmp/claude-501/-Users-jake-dev-BinaryBourbon-fountain/9b86f823-.../scratchpad/wt1744/deps
```

Mode `120000` is a symlink, which is why `/deps/` in `.gitignore` did not stop it — the ignore rule matches a directory path, and git recorded this as a file entry.

Traced to #1748 (`git log -1 -- deps`), merged 2026-09-12 04:17Z. The target is a scratchpad worktree belonging to a different session id, so it resolves on exactly one machine.

Worth saying why no gate caught it, since that is the interesting part: a dangling symlink is not an error to `File.mkdir_p`, so `mix deps.get` creates the target path and populates it, and the build succeeds — against a tree outside the repo. CI stayed green while every dependency lived in `/private/tmp`. The Compose fresh-clone check passes on this PR, which is the check that would have been closest to noticing.

Deleting the tracked entry restores the ignored-directory behaviour. Full CI green, 26 pass.

PR #2042, opened 2026-09-12
feat(api): expose credential sets and correct selection schemas

jhgaylor:
Expose tenant-scoped credential-set CRUD and agent/launch selection fields. Put credential selection fields on AgentRequest rather than Environment, return stable machine-readable refusal codes, and handle malformed or foreign set IDs consistently. Regenerate the server contract and TypeScript wire types from these schemas.

Validation: 120 CRUD/conversation tests and TypeScript/Python contract checks pass.

Base: #2034. Part of #2018; next #2043.


apps/fountain/lib/fountain_web/schemas.ex
```diff
@@ -1615,6 +1667,26 @@ defmodule FountainWeb.Schemas do
               "allows any environment the tenant owns; an empty list forbids overriding; " <>
               "a non-empty list is an allowlist. The agent's own environment always passes."
         },
+        inference_credential_id: %Schema{
```
BinaryBourbon: Wrong resource. This is `Schemas.Environment`, not an agent schema: `Fountain.Environments.Environment` has no `inference_credential_id` or `allowed_inference_credential_ids` column and `EnvironmentJSON.data/1` renders neither, so both fields are now published in `contract.json` and `openapi.ts` describing something the environments API never returns or accepts. The regex matched here because `Environment` also carries a stray `allowed_environment_ids` block with an agent-worded description (pre-existing; worth deleting while you are here). The pair belongs in `AgentRequest` instead, which is still missing it.

apps/fountain/lib/fountain_web/controllers/inference_credential_set_controller.ex
```diff
@@ -0,0 +1,154 @@
...
+
+    with %{} = set <- InferenceCredentials.get_set(id, user.id) || {:error, :not_found},
+         {:ok, _} <- InferenceCredentials.delete_set(set, Audited.attribution(conn)) do
+      send_resp(conn, :no_content, "")
+    else
+      {:error, :is_default} ->
+        conn
+        |> put_status(:unprocessable_entity)
+        |> json(%{
+          error: "the default credential set cannot be deleted; promote another one first",
```
BinaryBourbon: `error` should be the machine code and `message` the sentence. `sandbox_not_resettable` in `FallbackController` is the precedent for a three-field refusal (`error` + `message` + `reason`), and the three clauses you added to that file in this PR already follow it. As written a client branching on `error` gets prose, and `Schemas.Error` declares only `error` anyway. The test asserts `reason` and never `error`, so nothing catches it. Either fix the shape here or move the clause into `FallbackController` beside `cannot_undefault`.

BinaryBourbon (changes requested):
**Requesting changes.** The four routes, the tenant scoping, the audit path and the credits gate are all right. But the schema half of this PR — the part it exists to fix — landed one pair of fields on the wrong resource and missed the agent create body, and both are already baked into `contract.json` and `openapi.ts`.

## Checked

- Router: `/api/account` scope, `[:accepts_json, :api, :require_full_scope]`. Sprite-key refusal is tested on a read and a write.
- Tenant scoping: `list_sets(user.id)`, `get_set(id, user.id)` before every mutation; `rename_set`/`set_default`/`delete_set` take the already-fetched struct. No `_unsafe_` anywhere. Foreign id is 404 on PATCH and DELETE, and the row is asserted untouched.
- No value leakage: `providers/1` derives from `status_for_set/1`, which reads ciphertext byte sizes only and does no query (so no N+1 in `index` either). `refute inspect(body) =~ "sk-secret"` covers it.
- Audit is in the context with `Audited.attribution(conn)` from the controller; `set_default/2` records outside its transaction.
- `Billing.check_spend/1` is untouched on the launch path; the conversation change is the `@queued_attr_keys` allowlist only, which is a real fix.
- Both unique constraints are on the changeset, so a concurrent create racing on `is_default` is a 422, not a 500.
- `sdk/contract/omissions.json` already covers `* /api/account/*` with a reason naming inference credentials — no manifest change needed, as claimed.
- ADR edit is honest: `Source.missing/0` exists at this head.
- `sdk-no-release` label present, so the SDK-bump gate is satisfied.

## Blocking

**1. `Schemas.Environment` got the agent pair.** `schemas.ex:1670` and `:1679` add `inference_credential_id` and `allowed_inference_credential_ids`, with descriptions written about agents, to the `Environment` schema — "A reusable sandbox environment: packages, env vars, repos, networking". `Fountain.Environments.Environment` has neither column and `EnvironmentJSON.data/1` renders neither. This is the same regex over-match you caught in `AgentUpdate`; the verification you describe (each insertion sits after an `allowed_environment_ids` block) could not catch it, because `Environment` also has an `allowed_environment_ids` block — itself a stray with an agent-worded description that predates this PR and is presumably worth deleting in the same pass. The phantom fields are already in `contract.json` and in `openapi.ts` at `components["schemas"]["Environment"]`, so they ship to SDK consumers as a contract that lies.

**2. `Schemas.AgentRequest` did not get them.** `POST /api/agents` is the create body and it still declares `allowed_vault_ids` and `allowed_environment_ids` but neither inference-credential field. Confirmed in the generated output: `AgentRequest` (openapi.ts 2852-2903) has neither, while `Agent` and `AgentUpdate` do. It happens to work at runtime because `create/2` reads raw `params`, so this is a spec and SDK-types gap rather than a broken endpoint — but it is exactly the gap this PR is here to close ("without them an API client could not point an agent at a set"), and a typed client still cannot set the field on create. Counting `Environment` as one of the "four schema sites" is what hid it.

**3. The DELETE 422 inverts the repo's error shape.** `inference_credential_set_controller.ex:127`:

```elixir
error: "the default credential set cannot be deleted; promote another one first",
reason: "is_default"
```

Everywhere else `error` is the machine code and `message` is the sentence, with `reason` as an optional sub-code — `sandbox_not_resettable` in `FallbackController` is the exact precedent (`error` + `message` + `reason`), and the three clauses you added in this same PR follow it. Here `error` carries prose and there is no code at all, so a client branching on `error` the way every other endpoint asks it to gets a sentence. It is also declared against `Schemas.Error`, which has only `error`. The test asserts `body["reason"]` and never `body["error"]`, so it locks the shape in without noticing. Suggest `error: "credential_set_is_default"`, `message: "<the sentence>"`, `reason: "is_default"` — or move the whole clause into `FallbackController` beside `cannot_undefault`, which is where the sibling refusals live.

## Non-blocking

- `update/2` runs rename and promote as two independent audited writes. If the promote fails after the rename lands, the rename persists and the caller gets an error describing neither. Rare, but the response does not say what happened.
- No test asserts an audit row for any of the four operations. The context tests upstream cover the recording, so this is thin rather than absent.
- Deferring `docs/api.md` to the last PR is fine and is stated, but it is the one gate of the seven with no automated guard, so it is the one that can be missed silently.

BinaryBourbon (changes requested):

Reviewed all 11 changed files against base 1288aba8161d, including controller responses, tenant/scope checks, Agent request/response fields, generated contract and TypeScript types, tests, and prior objections. The earlier wrong-resource schema and deletion-error shape defects are repaired.

**P2 — classify the new deletion-error enums in the schema guardrail.** `apps/fountain/lib/fountain_web/schemas.ex:3669–3671` introduces `CredentialSetDeletionError.error` and `.reason`, but the changed `schema_enum_guardrail_test.exs` only classifies the provider enum. The exact-head [partition 4 run](https://github.com/managoat/fountain/actions/runs/34750540471/job/103706263330) fails “no enum is unclassified” for both new fields. This is introduced by this slice and prevents its required test gate from passing. Add these fixed wire enums to the appropriate explicit classification, preserving the guardrail and its assertions; rerun that test.

Evidence: the focused API/principal/console group passes 50 tests locally on the final stacked tree (these API files are unchanged there). Current-head SDK matrices and API-contract/release-boot checks pass. Other failing saved-allowance, platform-inference and wake tests are inherited execution-stage blockers, not introduced by this API diff; they still need resolution before the stack can merge.

BinaryBourbon (approved):

Reviewed the entire current diff at head `79890a69fd6e2edbcf6accc936f9447e3047fc35` against exact base `da9bbb0ba3d032fa92a0b92d7f5bd54c847184e2`, approved-base contributor/review instructions and ADR 0053, plus prior human comments and the current controller/context/client callers. No remaining blocking defect found in this slice.

The earlier wrong-resource schema, missing agent-create fields and deletion-error shape objections are repaired. The new fixed deletion error/reason enums are explicitly classified without weakening the guardrail. Agent create/update/read and conversation-create selection fields agree with actual rendering and validation. Named-set CRUD remains tenant-scoped and full-scope; malformed/foreign IDs refuse, responses expose presence rather than values, and context mutations reload current rows under the source lock. Existing account operations have an explicit SDK omission; generated TypeScript types add this wire surface without breaking existing handwritten client calls.

Independent exact-head diagnostics: 124 API/conversation/enum tests passed, including response-schema validation and denial cases. `scripts/sdk-contract/build.sh --check` passed (178 operations, 208 schemas). Current-head completed CI [34752555497](https://github.com/managoat/fountain/actions/runs/34752555497) passed; cancelled duplicate aggregate failures are separate from that run. The isolated reviewer diagnostics are not a claim to have rerun the full repository recipe.

Coverage (all 11 changed paths):
- `apps/fountain/lib/fountain_web/controllers/agent_json.ex`
- `apps/fountain/lib/fountain_web/controllers/inference_credential_set_controller.ex`
- `apps/fountain/lib/fountain_web/controllers/inference_credential_set_json.ex`
- `apps/fountain/lib/fountain_web/router.ex`
- `apps/fountain/test/fountain_web/controllers/conversation_controller_test.exs`
- `apps/fountain/test/fountain_web/controllers/inference_credential_set_controller_test.exs`
- `apps/fountain/test/fountain_web/schema_enum_guardrail_test.exs`

This is a source approval for this exact slice. It does not waive upstream findings, required CI, independent Review Loop verification, maintainer acceptance or rollout/release gates.

BinaryBourbon (approved):

Reviewed the entire current diff at head `b29133b5f9df613b66747cd2c59b3bd3858d4bbc` against exact base `c9de53080d7e142d5f5fa43fe347de265daa48c2`, approved-base contributor/review instructions and ADR 0053, plus prior human comments and the current controller/context/client callers. No remaining blocking defect found in this slice.

The earlier wrong-resource schema, missing agent-create fields and deletion-error shape objections are repaired. The new fixed deletion error/reason enums are explicitly classified without weakening the guardrail. Agent create/update/read and conversation-create selection fields agree with actual rendering and validation. Named-set CRUD remains tenant-scoped and full-scope; malformed/foreign IDs refuse, responses expose presence rather than values, and context mutations reload current rows under the source lock. Existing account operations have an explicit SDK omission; generated TypeScript types add this wire surface without breaking existing handwritten client calls.

Independent diagnostics on the tree-identical pre-restack revision: 124 API/conversation/enum tests passed, including response-schema validation and denial cases. `scripts/sdk-contract/build.sh --check` passed (178 operations, 208 schemas). Pre-restack completed CI [34752555497](https://github.com/managoat/fountain/actions/runs/34752555497) passed; cancelled duplicate aggregate failures are separate from that run. The isolated reviewer diagnostics are not a claim to have rerun the full repository recipe.

Final-head delta verification: fetched `b29133b5f9df613b66747cd2c59b3bd3858d4bbc` and its live base, independently proved both Git trees and the complete PR diff identical to the revisions reviewed/tested above. The upstream CI repair only relocated a dormant helper between prerequisite slices. Fresh GitHub CI [34753434550](https://github.com/managoat/fountain/actions/runs/34753434550) is queued; earlier successful-tree evidence does not replace this final-commit gate. Required checks remain pending.

Coverage (all 11 changed paths):
- `apps/fountain/lib/fountain_web/controllers/agent_json.ex`
- `apps/fountain/lib/fountain_web/controllers/inference_credential_set_controller.ex`
- `apps/fountain/lib/fountain_web/controllers/inference_credential_set_json.ex`
- `apps/fountain/lib/fountain_web/router.ex`
- `apps/fountain/test/fountain_web/controllers/conversation_controller_test.exs`
- `apps/fountain/test/fountain_web/controllers/inference_credential_set_controller_test.exs`
- `apps/fountain/test/fountain_web/schema_enum_guardrail_test.exs`

This is a source approval for this exact slice. It does not waive upstream findings, required CI, independent Review Loop verification, maintainer acceptance or rollout/release gates.

BinaryBourbon (approved):

Reviewed head `322d3208107534db0178f389093d4cdedd3c7d34` against main `95cdc9285bbf2723a14250c35b4b87335e24d6cb`. This queue candidate includes predecessor #2034 plus #2042's API slice: all 14 unique changed paths were reviewed, including launch admission, channel selection, queued request replay, credential-set CRUD, tenant boundaries and generated schemas.

The independent reviewer found no remaining blocker and ran 155 tests on this exact cumulative head: all passed. Wire-contract freshness passed (178 operations, 208 schemas). Both component patches preserve the previously approved changes. Local in-order squash simulation verified that #2034 then #2042 produces the expected source trees and that #2042's net merge adds only its intended API slice.

The complete integrated stack separately passed 6,370 tests plus 6 doctests, and the reviewed candidate tree is unchanged from its integration snapshot. These test groups overlap. Current PR CI is still running; both PRs will enter the queue in dependency order only after their checks pass. This approval does not bypass queue validation or claim completion of the Review Loop service matrix. No production resources or provider credentials were used.

PR #2043, opened 2026-09-12
feat(api): write and clear credentials in a named set

jhgaylor:
Add PUT and DELETE credential operations beneath a tenant-owned named set. Both routes reuse provider parsing, validation, auditing, and source revision updates while preserving clear not-found and unsupported-provider refusals.

Validation: 11 per-set endpoint tests and TypeScript/Python contract checks pass.

Base: #2042. Part of #2018; next #2044. Previous approval predates these repaired ancestors and is not treated as approval of this revision.


apps/fountain/lib/fountain_web/controllers/inference_credential_controller.ex
```diff
@@ -144,25 +150,120 @@ defmodule FountainWeb.InferenceCredentialController do
...
+      )
+
+  defp write(set, _user, dek, provider, value, conn),
+    do:
+      InferenceCredentials.put_credential_in(set, dek, provider, value, Audited.attribution(conn))
+
+  defp default_set(_user), do: nil
+
+  defp fetch_set(user, id) do
+    InferenceCredentials.get_set(id, user.id) || {:error, :not_found}
```
BinaryBourbon, line 262: `get_set/2` is `Repo.get_by` on a `:binary_id` primary key with no `Ecto.UUID.dump/1` guard, and the path parameter is `type: :string` — so a non-uuid `:id` here raises `Ecto.Query.CastError` out of `Repo` and answers 500 with a dropped connection instead of 404. `Conversations.get_conversation/2` (conversations.ex:641-660) documents the convention and the reason (#1679). The fix is one line inside `get_set/2`, which belongs to #2042 since that PR already exposes two `:id` routes on it.

apps/fountain/lib/fountain_web/controllers/inference_credential_controller.ex
```diff
@@ -144,25 +150,120 @@ defmodule FountainWeb.InferenceCredentialController do
...
+        provider,
+        value,
+        Audited.attribution(conn)
+      )
+
+  defp write(set, _user, dek, provider, value, conn),
+    do:
+      InferenceCredentials.put_credential_in(set, dek, provider, value, Audited.attribution(conn))
+
+  defp default_set(_user), do: nil
```
BinaryBourbon: A named function whose whole body is `nil`, at exactly one call site. `persist(conn, user, nil, provider, value)` reads the same and does not send the reader off to find out what it resolves to.

apps/fountain/test/fountain_web/controllers/inference_credential_in_set_test.exs
```diff
@@ -0,0 +1,210 @@
...
+
+      body =
+        conn
+        |> authed_with_key(key)
+        |> put_json(credential_path(second, "anthropic_api_key"), %{"value" => "sk-ant-second"})
+        |> json_response(200)
+
+      # The default set holds an OpenAI key; this response is about the set
+      # that was written, so it says nothing about that.
+      assert body["data"]["provider"] == "anthropic_api_key"
```
BinaryBourbon: This assertion holds regardless of which set was read: `show/1` renders only the requested provider, so the seeded OpenAI key on the default set can never show up here either way. The test would still pass with `persist/5` reverted to `status_for_user/1` — the behaviour is pinned by the `"set" => true` assertion in the first test, not by this one. Either assert `body["data"]["set"]` under a seeding where the two sets disagree on *that* provider, or drop the test.

BinaryBourbon (approved):
**Approve.** The secret handling is right, and the tenant scoping is right. Four findings, none of them blocking this stage.

## What I checked

- **Encryption.** Both entry points funnel into the new `write_credential/5`, which encrypts with `Crypto.encrypt(plain, dek)` where `dek` comes from `Crypto.load_tenant_key/1` via the controller's `load_dek/1`. The ciphertext-at-rest assertion at `inference_credentials_test.exs:74-85` therefore covers `put_credential_in/5` as well as `put_credential/5` — it is the same private function under both. The controller tests read back through `decrypted_for_set/2`, which only succeeds on a real ciphertext, so they are not merely asserting a 200.
- **Tenant scoping.** `get_set/2` is `Repo.get_by(Credential, id: id, user_id: user_id)`. `fetch_set/2` is the first clause of the `with` in both actions, so a foreign set 404s *before* the provider ping — no provider quota burned on a probe, and nothing stored. `put_credential_in/5` does no scoping of its own and its docstring says so rather than leaving the reader to check; the one caller obeys it.
- **No value anywhere.** Not in the response (`InferenceCredentialJSON.show/1` renders `provider` + a boolean), not in the 422 body (`provider_status` only), not in the audit row (`provider` / `set_id` / `set`), not in a changeset render (`ChangesetJSON` traverses errors, never `changes`), and no `Logger` or `inspect` on the value path. `Audit.record/1` is called outside any transaction.
- **Provider is a closed list twice** — the OpenAPI path enum and `parse_provider/1` as the fail-closed backstop.
- **Scope gate.** Both routes are inside the `/api/account` scope's `require_full_scope` pipeline, so a sprite token is 403.
- **Contract.** `contract.json` and `openapi.ts` regenerated for both operations, `sdk-no-release` applied.

## Findings

**1. A malformed set id is a 500, not a 404 — inherited, but this stage doubles the surface.** `get_set/2` has no `Ecto.UUID.dump/1` guard and the path parameter is `type: :string`, so `PUT /api/account/inference-credential-sets/nope/credentials/anthropic_api_key` raises `Ecto.Query.CastError` out of `Repo` and drops the connection. The repo has this convention written down — `Conversations.get_conversation/2` (`conversations.ex:641-660`, #1679) guards with `dump/1` precisely because the id arrives from a path segment, and `Agents.validate_environment_owner/1` carries the same note. The root is `get_set/2` from #2042, which already exposes two `:id` routes, so the one-line fix belongs there rather than here — I have raised it on that PR's review too. Nothing in the stack tests a malformed id.

**2. One of the ten new tests asserts nothing the others do not.** `inference_credential_in_set_test.exs:57-77`, "the response reports the set it wrote, not the account default", seeds an OpenAI key on the default and then asserts only `body["data"]["provider"] == "anthropic_api_key"` — which is true regardless of which set was read. `show/1` renders only the requested provider, so this test would still pass with `persist/5` reverted to `status_for_user/1`. The behaviour is actually pinned by the first test (`"set" => true` after a write into a non-default set); either strengthen this one to assert `body["data"]["set"]` under a seeding that makes the two sets disagree on *that* provider, or drop it.

**3. `defp default_set(_user), do: nil`.** A named function whose whole body is `nil`, at one call site. `persist(conn, user, nil, provider, value)` reads the same and does not send a reader off to find out what it resolves to.

**4. `provider_error/2` now fields the tenant-scoping miss as well as the ping outcomes.** It works today because `{:error, :not_found}` has a clause, but its terminal clause is `{:error, _network}` rendering "could not reach the provider to validate this credential" as a 502. If `fetch_set/2` ever grows a second error shape, an authorization outcome gets reported as a provider outage. Matching `{:error, :not_found}` in the `with` else before delegating, or a separate `set_error/2`, keeps the two vocabularies apart.

Minor: the description claims "a sprite-scoped key is 403 on both routes" — the test covers the PUT only. The pipeline covers both, so the claim is true; the test is not the thing establishing it.

BinaryBourbon (changes requested):

Exact head `bbaec80addf7103e92045140948089ee9e7a5f2b`; base `73d5d6f0dfac6c48a1f28ca15c3ccb462095f3c4`. Reviewed all 5 changed files: inference_credential_controller.ex, router.ex, inference_credential_in_set_test.exs, sdk/contract/contract.json, and generated openapi.ts. Traced tenant/API-key scope checks, provider validation, tenant DEK encryption, context writes/source revision triggers, status serialization, audit metadata, and client omission policy.

**P2: Re-read the selected set under the mutation lock after provider validation.** [update_in_set/2, lines 185–189](https://github.com/managoat/fountain/blob/bbaec80addf7103e92045140948089ee9e7a5f2b/apps/fountain/lib/fountain_web/controllers/inference_credential_controller.ex#L185) fetches the row before a provider request, then passes that potentially stale struct to [write_credential/5](https://github.com/managoat/fountain/blob/bbaec80addf7103e92045140948089ee9e7a5f2b/apps/fountain/lib/fountain/inference_credentials.ex#L292). If another request deletes this non-default set during validation, the eventual `Repo.insert_or_update` raises `Ecto.StaleEntryError`: the new endpoint crashes with 500 instead of its documented 404. I reproduced this through the real HTTP route by deleting the set inside the mocked provider response. The advisory lock is acquired only after the stale read and does not repair it. Reload by set id + user id inside that transaction, refuse a missing row with `:not_found`, and build the changeset/audit from the current row; keep provider I/O outside the lock. Add this interleaving as an endpoint regression.

The same stale-row path affects [delete_in_set/2](https://github.com/managoat/fountain/blob/bbaec80addf7103e92045140948089ee9e7a5f2b/apps/fountain/lib/fountain_web/controllers/inference_credential_controller.ex#L217): an independent HTTP interleaving that reads an empty set, commits a credential write, then resumes the clear returned 204 while the credential remained. That also recorded a delete without changing the persisted credential. Reloading under the lock gives these mutations and their audit events one consistent order.

Evidence: **13 tests, 2 failures** on this exact head in an isolated checkout/disposable PostgreSQL 18 database: all 11 shipped endpoint tests passed; both independent interleaving regressions above failed. Python contract verifier passed. The earlier malformed-id objection is repaired and its new test passes; foreign-owner and sprite-scope refusals remain enforced, and responses/audit contain no plaintext.

Separate merge evidence: current-head [full CI run 34750540974](https://github.com/managoat/fountain/actions/runs/34750540974) actually fails partitions 1/2/3/5 (saved-allowance state assertions, the CredentialSetDeletionError enum guardrail, source-kind assertions, and deleted-sandbox wake FK handling). These are distinct from the older cancelled duplicate run and must be cleared before merge; `Detect secrets` passed. This source finding is independently blocking regardless of those inherited CI failures.

BinaryBourbon (approved):

Completed an independent whole-PR review at exact head `14f90c919f252c1b07b0903e5b9048f987e7c477` against approved base `79890a69fd6e2edbcf6accc936f9447e3047fc35`. No blocking defect found in this slice.

Coverage: all 5 changed files — apps/fountain/lib/fountain_web/controllers/inference_credential_controller.ex; apps/fountain/lib/fountain_web/router.ex; apps/fountain/test/fountain_web/controllers/inference_credential_in_set_test.exs; sdk/contract/contract.json; sdk/typescript/src/generated/openapi.ts. Read approved-base contributor/reviewer policy and prior GitHub reviews/comments. Traced bearer authentication/full-scope checks, tenant-scoped lookups, provider validation, tenant DEK encryption, current-row mutation/source revision locks, audit metadata, status JSON, fallback errors, schemas and existing SDK omission policy. Both operations fit the existing account-self-service omission; no client manifest or version was weakened.

The prior mutation interleavings are repaired in the current base/context and exercised through these routes: deletion during the provider ping now returns 404 without resurrecting a set or writing a false credential audit; a clear after an interleaved write removes the current ciphertext; deletion after a clear lookup also returns 404. Provider I/O stays outside the source lock, and the mutation re-fetches id + tenant under it. Malformed/foreign IDs and sprite-scoped writes/clears remain refused. Rendering uses the written set's status and contains no credential material.

Evidence on this exact head: **73 shipped tests, 0 failures** across named/legacy credential endpoints and context suites; **4 independent diagnostic tests, 0 failures** covering named-route timeout/network preservation, foreign-set refusal without a provider request, and rename-during-validation with current audit metadata. Isolated OTP 28.3 / Elixir 1.19.2 and disposable PostgreSQL 18; no real provider calls. Server-to-committed wire-contract check passed (180 operations, 208 schemas); Python SDK contract verifier passed (68 operations, 41 schemas, 4 enums). Actual current-head [CI run 34752556267](https://github.com/managoat/fountain/actions/runs/34752556267) and Detect secrets succeeded, separately from the older cancelled duplicate aggregate. I did not rerun the full independent verification recipe locally. This source approval does not waive CI, maintainer or merge gates.

BinaryBourbon (approved):

Completed an independent whole-PR review at exact head `b4e662aab343cf0fbdc29ae8cc75bb0789f4a5a7` against approved base `b29133b5f9df613b66747cd2c59b3bd3858d4bbc`. No blocking defect found in this slice. Revalidated this restack independently: both the head tree and approved-base tree, and the full PR diff bytes, are identical to the complete review immediately preceding it. The CI-only change moves an unreachable helper from #2031 to its enabling stage #2034; it leaves this slice unchanged, including its trusted reviewer instructions.

Coverage: all 5 changed files — apps/fountain/lib/fountain_web/controllers/inference_credential_controller.ex; apps/fountain/lib/fountain_web/router.ex; apps/fountain/test/fountain_web/controllers/inference_credential_in_set_test.exs; sdk/contract/contract.json; sdk/typescript/src/generated/openapi.ts. Read approved-base contributor/reviewer policy and prior GitHub reviews/comments. Traced bearer authentication/full-scope checks, tenant-scoped lookups, provider validation, tenant DEK encryption, current-row mutation/source revision locks, audit metadata, status JSON, fallback errors, schemas and existing SDK omission policy. Both operations fit the existing account-self-service omission; no client manifest or version was weakened.

The prior mutation interleavings are repaired in the current base/context and exercised through these routes: deletion during the provider ping now returns 404 without resurrecting a set or writing a false credential audit; a clear after an interleaved write removes the current ciphertext; deletion after a clear lookup also returns 404. Provider I/O stays outside the source lock, and the mutation re-fetches id + tenant under it. Malformed/foreign IDs and sprite-scoped writes/clears remain refused. Rendering uses the written set's status and contains no credential material.

Evidence from the identical source tree at preceding head `14f90c919f252c1b07b0903e5b9048f987e7c477` (tests were not rerun for this metadata-only restack): **73 shipped tests, 0 failures** across named/legacy credential endpoints and context suites; **4 independent diagnostic tests, 0 failures** covering named-route timeout/network preservation, foreign-set refusal without a provider request, and rename-during-validation with current audit metadata. Isolated OTP 28.3 / Elixir 1.19.2 and disposable PostgreSQL 18; no real provider calls. Server-to-committed wire-contract check passed (180 operations, 208 schemas); Python SDK contract verifier passed (68 operations, 41 schemas, 4 enums). The preceding identical-tree [CI run 34752556267](https://github.com/managoat/fountain/actions/runs/34752556267) succeeded. **Fresh CI for the new commit is still queued/running; it has not passed yet.** Current-head Detect secrets succeeded. The cancelled duplicate run's aggregate failures do not establish an actual test failure in the active run. I did not rerun the full independent verification recipe locally. This source approval does not waive the pending CI, maintainer or merge gates.

BinaryBourbon (approved):

Final queue candidate `f256dba0bbcbe952817683c07205c7f7b1232d1d` against main `c77c381f81953362a5a080dcd823270bcb3754ec`. Its source tree is identical to reviewed integration snapshot `b3115eb91cf0bc901f14d53ef36e5c8f36394a5b`. The locally simulated in-order squash merges preserve each intended incremental diff; a second candidate includes its reviewed predecessor until that predecessor lands.

Independent API and ownership reviewers approved the five-path named-set credential write/clear slice. Scoped row lookup, stale target reload, default mutation ordering, provider validation boundaries and post-commit audit were inspected. The relevant API integration group passed 199 tests and wire-contract regeneration/verifiers; principal/console integration passed 54 tests.

PR #2044, opened 2026-09-12
fix(principals): authorize credential writes against current active ownership

jhgaylor:
Allow a principal's current owner to write or clear its inference credential only while the grant is active. Before claim the issuing application may write; after claim the claimer may write and the original application is refused. Expired, released, or elapsed grants refuse both PUT and DELETE.

Authorization and credential mutation share a transaction and row lock, so a concurrent claim cannot leave the former owner authorized. Audit events are recorded only after commit. Unknown providers return the declared 422 response; malformed principal IDs return 404.

Validation includes real PostgreSQL claim-versus-PUT and claim-versus-DELETE races and audit-after-commit coverage.

Base: #2043. Part of #2018; next #2045.


apps/fountain/lib/fountain/principals.ex
```diff
@@ -134,6 +134,59 @@ defmodule Fountain.Principals do
...
+  principal's default credential set, because it is the principal's
+  credential; the owner is only who asked. The trail is the principal's too,
+  with the acting account named in the metadata.
+  """
+  @spec put_inference_credential(binary(), binary(), atom(), String.t() | nil, keyword()) ::
+          {:ok, Fountain.InferenceCredentials.Credential.t()}
+          | {:error, :not_found | :tenant_key_unavailable | Ecto.Changeset.t()}
+  def put_inference_credential(id, viewer_id, provider, value, opts \\ [])
+      when is_binary(id) and is_binary(viewer_id) do
+    with %ClaimableUser{user_id: principal_id} <- get_claimable_for(id, viewer_id) || :none,
```
BinaryBourbon: `get_claimable_for/2` is the *read* predicate: application OR claimer. After a claim the application is not the owner, yet it still passes here and can swap the inference credential every turn of that tenant runs on. `write_claim/3` revokes the application's keys precisely so it has no standing access to a claimed principal. Gate on ownership instead: the application while `unclaimed`, `claimed_by_user_id` once claimed. A closed (`expired`/`released`) grant should not accept the write either.

apps/fountain/lib/fountain_web/controllers/claimable_user_controller.ex
```diff
@@ -201,6 +203,105 @@ defmodule FountainWeb.ClaimableUserController do
...
+    Principals.put_inference_credential(
+      id,
+      conn.assigns.current_user.id,
+      provider,
+      value,
+      Audited.attribution(conn)
+    )
+  end
+
+  # The path parameter is an enum in the spec, so an unknown provider is a 422
```
BinaryBourbon: This comment is inherited from `InferenceCredentialController`, which installs `plug OpenApiSpex.Plug.CastAndValidate`. This controller does not, and neither does the `:api` pipeline, so the enum is never enforced and this clause is the primary path, not a backstop. An unknown provider therefore renders 404, while both operations declare 422 for it.

BinaryBourbon (changes requested):
**Request changes** — two things: the door gate still asks the *default* set, and the one line that makes the headline claim true has no test that would fail if it were reverted.

Checked: tenant scoping on both write paths and on the read path, the FK and index in the migration and its reversibility, `@snapshot_fields` / rollback, every caller of `load_tenant_state/1` and `decrypted_for_user/2`, the `select/4` fallthrough, and the credits gate at the four `start_conversation` doors (`check_spend/1` runs ahead of the inference gate at each; untouched).

The ownership work is right. `validate_credential_set_owner/1` mirrors `validate_environment_owner/1` including the `valid?: false` short-circuit, both create and update run it, `rollback_agent/3` goes through `update_agent/3` so a snapshot naming a deleted set is refused rather than restored blind, and `decrypted_for/3` resolves through the tenant-scoped `get_set/2` so a cross-tenant id reads as the account's own default. Both directions have a test, and the revert-check you describe is the right one.

## 1. `PlatformInference.gate/4` reads the default set; the provision reads the agent's

`gate/4` -> `has_own?/3` -> `missing_for_model/2` -> `status_for_user/1` -> `get_for_user/1`, which is the row `where is_default`. After this PR, provisioning reads the agent's set instead. So an agent whose credential lives in a non-default set is invisible to the door: on a deployment that holds a platform key for that provider, once the day's ceiling is spent, `start_conversation` refuses with `:platform_inference_unavailable` for a launch that would have run on the tenant's own key.

That is the same failure #2025 removed for tenant secrets, one PR after removing it, and it makes `gate/4`'s own docstring false as written today ("the gate never refuses for a credential the selection would not hand out"). The hosted deployment has held platform keys since 2026-09-03, so the trigger is reachable.

The ingredients are already in the module from #2028: `status_for_set/1` takes a loaded set. `has_own?/3` and `missing_for_model/2` can take an `:inference_credential_id` opt and read `status_for_set(get_set(id, user_id) || get_for_user(user_id))`, and the four `Fountain.PlatformInference.gate(...)` call sites in `conversations.ex` (3192, 3974, 4843, 5154) pass `agent.inference_credential_id` beside the `environment_id` and `vault_id` they already pass.

The opposite direction I checked and am not asking you to change: default set holds the key, agent names an empty set, the gate says `:ok`, and `select/4` falls through to `platform_credential/4` and serves the platform key. The turn genuinely runs on it, `TurnMachine.gate/2` re-checks the ceiling per turn off the resolved `Source`, so nothing escapes the ceiling and the stamp is honest. Worth knowing it happens, though: it is why the PR body's "an agent pointed at an empty set will not reach a model, whatever the default set holds" is not true on a deployment with platform keys. The banner is right, because it asks `select/4`; the sentence is not.

## 2. `conversation_server.ex:753` is untested, and the new stub makes it untestable

"`ConversationServer` passes the agent's set, so a conversation provisions with it" is the point of the PR, and it is one line. No test covers it. The new test file calls `SpriteEnv.load_tenant_state/2` directly with an id it fetched off the agent itself, which passes whatever `ConversationServer` does. And `ConversationServerCase`'s new `decrypted_for/3` stub answers on `user_id` and discards `set_id`, so no existing or future `ConversationServer` test can see the difference either.

Change that line back to `SpriteEnv.load_tenant_state(conv.user_id)` and the suite stays green. Please add one `ConversationServer` test that provisions an agent naming a set holding a distinguishable key and asserts `state.inference_credentials`, with the stub honouring `set_id` — resolving through the real `get_set/2` and `decrypted_for_set/2` (neither is stubbed) keeps the three tests that override `decrypted_for_user/2` working while making the wiring observable.

## Non-blocking

- `decrypted_for_set/2` is public with only two callers, both in this module. It takes a struct nothing re-scopes; the DEK makes it fail closed on another tenant's ciphertext, so this is a naming point rather than a hole, but the repo's rule puts `_unsafe_` on an unscoped function (`Vaults._unsafe_list_secrets/1` is the same shape). Private is the cheaper fix if no later stage needs it.
- The unresolvable-id fallback is silent. Ten lines above, the environment arm logs a warning and provisions without it; this one quietly moves the agent onto the default set's key, which for the reseller in the ADR's Context is the one outcome that must not happen unnoticed. The changeset and `on_delete: :nilify_all` make it hard to reach, so a `Logger.warning` matching the environment arm is enough.
- `Source` gains no `set_id`. ADR decision 2 says it arrives "where there is a set to name", which is here — but it also says a field arrives with its reader, and nothing reads it yet. Fine to leave; worth a line in the stack notes so #9/#10 do not each assume the other added it.

Migration is reversible (`alter/add` plus `create index` under `change`), `nilify_all` is the right choice and is tested, and the index on the FK is there for it.

apps/fountain/lib/fountain/principals.ex
BinaryBourbon: `get_claimable_for/2` is the *read* predicate: `application_user_id OR claimed_by_user_id`. After a claim the application is no longer the owner, yet it still passes here and can swap the inference credential every turn of that tenant runs on. `write_claim/3` revokes the application's keys at the claim precisely so it holds no standing access to a machine that now belongs to somebody, and ADR 0053 decision 7 says "a principal it owns" — `Principals.owner_id/1` is right here in the module. Gate on ownership: the application while `unclaimed`, `claimed_by_user_id` once claimed. A closed (`expired`/`released`) grant should not accept the write either.

apps/fountain/lib/fountain_web/controllers/claimable_user_controller.ex
BinaryBourbon: This comment came from `InferenceCredentialController`, which installs `plug OpenApiSpex.Plug.CastAndValidate`. This controller installs no cast plug and the `:api` pipeline has none, so the enum is never enforced: this clause is the primary path, not a backstop, and an unknown provider renders 404 while both operations declare 422 for it.

BinaryBourbon (changes requested):
> **This supersedes [review 5189125622](https://github.com/managoat/fountain/pull/2044#pullrequestreview-5189125622), which I posted in error against the wrong diff.** That body reviews #2031 (the agent naming a credential set) and says nothing about this change. The analysis below was written against this PR; the earlier post was a payload mix-up on my side, not a judgement about #2031's code. Please disregard it entirely.

**Request changes.** The authorisation predicate is wrong after a claim, and it re-opens a boundary that `write_claim/3` closes deliberately.

## What I checked

Both routes sit in the `/api/claimable-users` scope behind `:require_full_scope`, and the suite proves it from both sides: a `principal`-scoped key gets 403 and a sprite-scoped key gets 403, each asserting nothing was written.

The DEK is the principal's, and the code path shows it rather than the PR body. `put_inference_credential/5` resolves `%ClaimableUser{user_id: principal_id}`, calls `load_principal_key(principal_id)` -> `Crypto.load_tenant_key(principal_id)`, and passes that pair to `InferenceCredentials.put_credential(principal_id, dek, ...)`, which resolves the set with `get_for_user(principal_id)`. Caller ids reach neither the key nor the row. The test that loads the *application's* DEK and asserts `decrypted_for_user(application.id, app_dek) == {:ok, %{}}` pins it from the outside too.

Authorisation reuses `Principals.get_claimable_for/2` rather than reimplementing it, which is the right instinct — but see below, because the helper it reuses answers a different question than this route asks.

Audit is the principal's: `Audit.record` on `set.user_id` with `provider`, `set_id` and `set`, and `by_account` merged in from the caller. `Map.merge(caller_metadata, forced)` puts the forced keys second, so a caller cannot spoof `provider` or `set_id`. `refute inspect(event) =~ "sk-trailed"` covers the value. No transaction wraps it.

Gates: `docs/api.md` states outright that it keeps no second endpoint catalog, so no entry is owed and the generated reference picks these up from the `operation/2` declarations. `sdk/contract/omissions.json` already carries `* /api/claimable-users*`, so `--check` is satisfied. `sdk-no-release` is on the PR and stage 13 carries the 1.32.0 bump. The 422 body is `%{error: "value is required", reason: "empty_value"}`, byte-identical to `InferenceCredentialController`'s.

Refusal coverage is good but has one hole, in item 1.

## 1. Blocking: the application that opened a principal can still write to it after somebody claims it

`get_claimable_for/2` matches `application_user_id == ^viewer_id or c.claimed_by_user_id == ^viewer_id`. That is the correct predicate for a *read* — a closed grant still answering is what makes reconciliation work. It is not the correct predicate for this write, because `write_claim/3` revokes every key the application holds at the moment of the claim, with the reason written beside it:

> Every credential the application held stops working here, atomically with the claim: an anonymous key that outlived the claim would be standing access to a machine that now belongs to somebody.

This route hands that access back in a narrower but worse shape. After the claim the application holds no key to the principal and can no longer claim it or release it — `still_open/1` and `decide_claim/1` see to that — but it can still call `PUT /api/claimable-users/:id/inference-credentials/anthropic_api_key` with a key it controls. Every turn that tenant subsequently runs then goes through the application's provider account: the claimed owner's prompts and completions, on a credential the owner did not set and gets no signal was replaced. That is standing write access to a machine that now belongs to somebody.

ADR 0053 decision 7 says "a principal it **owns**". After a claim the owner is `claimed_by_user_id`, and `Principals.owner_id/1` is in the same module and returns exactly that. The docstring is candid that authorisation mirrors the read, but mirroring the read is the defect, not the justification for it.

Please gate on ownership: the application while `status` is `unclaimed`, the claiming account once `claimed`. And add the refusal test — the suite has "the account that claimed it may write too" but nothing asserting the application loses the ability, so the post-claim behaviour is currently neither chosen nor pinned, and a later refactor could flip it either way silently.

## 2. Blocking, same helper: a closed grant accepts the write and answers 204

There is no status check. An `expired` or `released` principal has had its keys revoked, its sandboxes destroyed and its grant refunded, and `ClaimablePrincipalSweep` deletes it a retention window later — but this route writes a customer's secret onto one and returns 204, so the caller reasonably believes a key is in place that nothing can ever use. Every other write in the module checks status (`still_open/1`, `decide_claim/1`, `release/2`); only the reads deliberately do not. Refuse on a closed grant, or say in the docstring why writing to a dead principal is intended.

## 3. The unknown-provider comment is false in this controller, and the declared 422 is unreachable

```elixir
# The path parameter is an enum in the spec, so an unknown provider is a 422
# before the action runs. This is the fail-closed backstop.
```

True in `InferenceCredentialController`, which installs `plug OpenApiSpex.Plug.CastAndValidate`. `ClaimableUserController` installs no cast plug, and the `:api` pipeline has none either — it is `PutApiSpec`, `RateLimit`, `TenantAPIAuth`, `RateLimit`, `Audit`. So nothing validates the enum, `parse_provider/1` is the primary path rather than a backstop, and an unknown provider returns `{:error, :not_found}` -> **404**. Both operations declare `unprocessable_entity: {"Unknown provider", ...}`, so the OpenAPI document and the SDK contract now state a status this route never renders. Either render 422 for an unrecognised provider or correct the declaration and the comment. No test either way; worth one case.

## Non-blocking

- The controller moduledoc still lists four routes; these two are not in it.
- `docs/build/anonymous-visitors.md` is the page an application integrator reads for this surface, and it never gains the route anywhere in the stack — only the CHANGELOG in stage 13 mentions it.
- `omissions.json`'s reason still reads "reaches these four over plain HTTP"; it is six now.
- No 401 case in the new suite. The pipeline covers that elsewhere, so this is a completeness note rather than a gap.
- `:tenant_key_unavailable` has no `FallbackController` clause and lands on the terminal safety net as a 422 carrying the atom. Same as the account's own route, so pre-existing rather than introduced here.

BinaryBourbon (changes requested):

Reviewed all seven changed files and the actual authorization/mutation callers, including full-scope enforcement, principal DEK selection, current owner before/after claim, release/expiry, audit attribution, generated schemas and both regression files. The prior former-owner write and unknown-provider response defects are repaired. The grant row lock and post-wait owner/status predicate are sound.

**P2 — build the clear from the current credential row after acquiring the source lock.** The new `delete_inference_credential/2` delegates through `Principals.put_inference_credential/5` (`principals.ex:160–179`) to `InferenceCredentials.put_credential/5`. That helper reads the default row before entering `write_credential`'s source-lock transaction. I independently reproduced this sequence using separate PostgreSQL connections and a barrier: the principal's default credential is empty; another transaction holds its source lock; a clear reads the empty row and waits; the lock holder writes a key and commits; the clear acquires the lock, passes current-owner authorization, and returns `{:ok, _}` while the stored key remains. The controller converts that result to 204. The stale nil field is treated as an unchanged Ecto value, so no clear reaches SQL.

This is a **shared context defect inherited from the prerequisite**, also reported on [#2043](https://github.com/managoat/fountain/pull/2043#pullrequestreview-5190395456), exposed by this new principal-management route. Fix the primitive once: acquire the source lock before resolving the default/current row and constructing the changeset, while preserving the grant authorization lock through mutation. Add the principal interleaving regression. No new ownership permission is needed.

Evidence: the shipped API/principal/race/console group passes 50 local tests; the additional principal-clear diagnostic fails 1/1. Current-head SDK and contract checks pass. The completed CI run remains red on inherited execution/schema tests; the separate cancelled duplicate's aggregate failures are not another code finding. Approval is withheld for the reachable mutation defect, independently of those CI blockers.

BinaryBourbon (approved):

Reviewed the whole current diff at head `902663f40aebe0eb7b3d74af0b1d144b3da957b7` against exact base `14f90c919f252c1b07b0903e5b9048f987e7c477`, the identical approved-base contributor/review policies, ADRs 0044/0053 and prior human objections. No remaining blocking defect found in this principal-management slice.

The old application-or-claimer read predicate is only a preliminary scoped lookup. The actual mutation now acquires the source lock, then locks the grant using the current active-owner predicate, rechecks elapsed unclaimed expiry, and retains that authority lock through the write. Claimed principals accept only the claimer; released/expired/elapsed grants and former owners refuse. The default/current credential is resolved after the wait, which repairs the prior successful-clear-that-left-the-key race. Claim and closure use the same grant lock. Values use the principal's DEK and audit trail, with forced acting-account attribution and no secret value in the event. Audit happens after mutation commit. Both routes enforce full scope, the unknown-provider result now matches 422, and generated contract/types match actual responses. No provider ping is promised on this owner-management surface.

Independent exact-head evidence: 21 shipped principal API/concurrency/race tests passed. I additionally adapted a separate barrier-controlled diagnostic to race release with both write and clear; all 3 diagnostic tests passed, including post-commit audit failure. The outer diagnostic transaction intentionally prevented its optional compute cleanup (logged `provider_transaction_open`); authorization, refusal and stored credential assertions passed. The wire-contract check passed (182 operations, 208 schemas). Current-head full CI [34752555026](https://github.com/managoat/fountain/actions/runs/34752555026) passed. These focused local runs do not claim the whole independent verification recipe.

Coverage (all 8 changed paths):
- `apps/fountain/lib/fountain/principals.ex`
- `apps/fountain/lib/fountain_web/controllers/claimable_user_controller.ex`
- `apps/fountain/lib/fountain_web/router.ex`
- `apps/fountain/test/fountain/principal_inference_credential_race_test.exs`
- `apps/fountain/test/fountain_web/controllers/principal_credential_concurrency_test.exs`
- `apps/fountain/test/fountain_web/controllers/principal_inference_credential_test.exs`

This approval covers this exact source slice, not inherited execution behavior or merge readiness. Upstream findings, required CI, independent Review Loop verification and maintainer/rollout gates remain in force.

BinaryBourbon (approved):

Reviewed the whole current diff at head `646e835ef42b215b176ae9c6f70005451fafd973` against exact base `b4e662aab343cf0fbdc29ae8cc75bb0789f4a5a7`, the identical approved-base contributor/review policies, ADRs 0044/0053 and prior human objections. No remaining blocking defect found in this principal-management slice.

The old application-or-claimer read predicate is only a preliminary scoped lookup. The actual mutation now acquires the source lock, then locks the grant using the current active-owner predicate, rechecks elapsed unclaimed expiry, and retains that authority lock through the write. Claimed principals accept only the claimer; released/expired/elapsed grants and former owners refuse. The default/current credential is resolved after the wait, which repairs the prior successful-clear-that-left-the-key race. Claim and closure use the same grant lock. Values use the principal's DEK and audit trail, with forced acting-account attribution and no secret value in the event. Audit happens after mutation commit. Both routes enforce full scope, the unknown-provider result now matches 422, and generated contract/types match actual responses. No provider ping is promised on this owner-management surface.

Independent diagnostics on the tree-identical pre-restack revision: 21 shipped principal API/concurrency/race tests passed. I additionally adapted a separate barrier-controlled diagnostic to race release with both write and clear; all 3 diagnostic tests passed, including post-commit audit failure. The outer diagnostic transaction intentionally prevented its optional compute cleanup (logged `provider_transaction_open`); authorization, refusal and stored credential assertions passed. The wire-contract check passed (182 operations, 208 schemas). Pre-restack full CI [34752555026](https://github.com/managoat/fountain/actions/runs/34752555026) passed. These focused local runs do not claim the whole independent verification recipe.

Final-head delta verification: fetched `646e835ef42b215b176ae9c6f70005451fafd973` and its live base, independently proved both Git trees and the complete PR diff identical to the revisions reviewed/tested above. The upstream CI repair only relocated a dormant helper between prerequisite slices. Fresh GitHub CI [34753434333](https://github.com/managoat/fountain/actions/runs/34753434333) is queued; earlier successful-tree evidence does not replace this final-commit gate. Required checks remain pending.

Coverage (all 8 changed paths):
- `apps/fountain/lib/fountain/principals.ex`
- `apps/fountain/lib/fountain_web/controllers/claimable_user_controller.ex`
- `apps/fountain/lib/fountain_web/router.ex`
- `apps/fountain/test/fountain/principal_inference_credential_race_test.exs`
- `apps/fountain/test/fountain_web/controllers/principal_credential_concurrency_test.exs`
- `apps/fountain/test/fountain_web/controllers/principal_inference_credential_test.exs`

This approval covers this exact source slice, not inherited execution behavior or merge readiness. Upstream findings, required CI, independent Review Loop verification and maintainer/rollout gates remain in force.

BinaryBourbon (approved):

Final queue candidate `c2fd663b97a96c932a8eca3257fc16c712f5bb9e` against main `c77c381f81953362a5a080dcd823270bcb3754ec`. Its source tree is identical to reviewed integration snapshot `f8c9ebbf685fa7497f188a7e6199bd63adab2647`. The locally simulated in-order squash merges preserve each intended incremental diff; a second candidate includes its reviewed predecessor until that predecessor lands.

Independent ownership review approved this complete 10-path cumulative queue diff, including predecessor #2043 and principal authorization. Named-set writes, active current grant ownership, source/grant lock ordering, stale mutation handling, audit behavior and generated contracts were inspected together. Main’s new permanent principal-key expiry is preserved. The API integration group passed 199 tests, and principal/console integration passed 54 tests.

PR #2045, opened 2026-09-12
fix(console): validate and save credentials in the selected set

jhgaylor:
Keep credential status, validation, inline writes, and settings saves on the selected set. Switching selection or submitting stale form state cannot silently write into the account default. Delete confirmation describes refusal of existing bound conversations instead of promising automatic fallback.

Validation: 15 console tests pass.

Base: #2044. Part of #2018; next #2046.


apps/fountain/lib/fountain_web/live/agents_live/form.ex
```diff
@@ -21,9 +21,12 @@ defmodule FountainWeb.AgentsLive.Form do
      |> assign(:user_id, user_id)
      |> assign(
        :missing_credential,
-       InferenceCredentials.missing_for_model(user_id, credential_model(agent))
+       InferenceCredentials.missing_for_model(user_id, credential_model(agent),
+         credential_set_id: agent.inference_credential_id
```
BinaryBourbon, line 25: This is the only one of three `missing_for_model` call sites in this file that names the set. `handle_event("validate", ...)` at line 259 and `handle_event("save_credential", ...)` at line 277 still ask about the account default, so the card this computes disappears on the first form change, and a key pasted into it lands in the default set rather than the one the agent names.

apps/fountain/lib/fountain_web/live/agents_live/form.ex
```diff
@@ -875,6 +880,28 @@ defmodule FountainWeb.AgentsLive.Form do
           />
         </div>
 
+        <div :if={@credential_sets != []} class="space-y-1">
```
BinaryBourbon: Non-blocking: `!= []` shows this for an account with a single set, while the credentials page in this same PR hides the concept below two and says why in its moduledoc. `length(@credential_sets) > 1` would match, and would stop the only set being offered as a choice against the account default when the two differ only if the default later moves.

BinaryBourbon (changes requested):
**Request changes** — the sets half of `/account/inference-credentials` is sound; the agent-form half wires the new set id into `mount/3` only, and the two other call sites in the same file now disagree with it. One of them sends a pasted key to the wrong set.

## What I checked and found clean

- **Auth.** `/account/inference-credentials` and `/agents/:id/edit` are both inside the `:authenticated` `live_session` with `{FountainWeb.Live.Hooks, :require_authenticated_user}`. Unchanged by this PR, correct.
- **Tenant scoping.** Every read is `list_sets(user_id)`. `select_set` takes an id from the client but only `Enum.find`s it inside the user's own list, so a foreign id selects nothing and falls through to `List.first(sets)`. `rename_set` / `make_default` / `delete_set` act on `socket.assigns.set`, which came out of that scoped list. No `_unsafe_*` call, and none needed.
- **No plaintext reaches the client.** `status_for_set/1` yields booleans off ciphertext presence; nothing decrypts on this page, and no credential value lands in an assign.
- **Audit.** All four set mutations pass `FountainWeb.Audited.attribution(socket)`, so the row gets `ui`, and the recording lives in the context. `set_default/2` records outside its transaction, per ADR 0013.
- **`write_into_selected/4`** correctly falls back to `put_credential/5` when the account holds no set yet, which is the path that mints `Default` on a first write. Deleting the selected set reloads onto the default rather than leaving a dangling `@set`.
- **Ownership of a submitted `inference_credential_id`** is enforced in `Agents.validate_credential_set_owner/1` from an earlier stage, so the picker being scoped is defence in depth rather than the only guard. Good.
- Tests cover the mount (both the hidden-controls case and the first-write case) and four mutations.

## Blocking

**1. `handle_event("validate", ...)` drops the set id** (`form.ex:259`, not in this diff). `mount/3` now asks `missing_for_model/3` about `agent.inference_credential_id`; `validate` still asks about the account default. An agent pointed at a set that holds no key for its provider shows the card on mount and loses it on the first keystroke in the form. Pass `params["inference_credential_id"]` through.

**2. `handle_event("save_credential", ...)` writes into the wrong set** (`form.ex:277`). The just-in-time card goes through `InferenceCredentialSave.save/3`, which persists with `put_credential/5` — always the account's **default** set — and then recomputes `missing_credential` against the default too. So with an agent pointed at a non-default set: the card correctly says the key is missing, the user pastes it, the card clears, and the set the agent actually runs on still holds nothing. The conversation then fails inside the sandbox, which is the exact failure #841's card exists to prevent. Either persist through `put_credential_in/5` with the named set, or suppress the inline card when the agent names a non-default set and link to the credentials page instead.

Both are introduced here: before this PR all three call sites agreed on the default. ADR 0053 decision 1 says "everything else that reads without being told which set reads the default", which is fine as a rule — but this file is told, in one of three places, and the incoherence is user-visible.

## Non-blocking

3. The picker renders on `@credential_sets != []`, so it appears for every account with a single set. The credentials page in this same PR deliberately hides the concept below two (`length(@sets) > 1`), and its moduledoc gives the reason. `length(@credential_sets) > 1` would match, and would avoid presenting "— account default —" against "Default" as a choice — they differ only if the default later moves, which is not something the label conveys.

4. No test covers the agent-form half at all. Nothing pins that the picker offers only owned sets, or that a chosen set round-trips into `inference_credential_id` on save. `git grep inference_credential_id` over `test/fountain_web/` finds only `conversation_controller_test.exs`.

5. `"the tabs appear, the default is marked, and selecting one switches the rows"` does not assert the switch. `"Second subscription"` is in the HTML before the click (it is a tab label), `"The four rows below are the"` renders whenever `@set` is truthy, and `render(view) =~ "Set"` matches the per-provider status badge. The test would pass with `select_set` reduced to `{:noreply, socket}`. The test after it does the real work — this one could assert on the provider status the two sets differ by.

BinaryBourbon (changes requested):

Reviewed all five changed files, their context callers and tests. Agent mount, validation and inline credential save now consistently use the selected set, including deletion during provider validation; the earlier default-set misdirection objection is repaired.

**P2 — handle deletion of the selected set before Rename without crashing the page.** At `apps/fountain/lib/fountain_web/live/inference_credentials_live/index.ex:104–117`, Rename sends the socket's cached credential row straight to `InferenceCredentials.rename_set/3`. Select a non-default set, delete it through another tab/API request, then submit Rename in the original tab: `Repo.update` raises `Ecto.StaleEntryError`, which escapes this handler and terminates the LiveView. I reproduced this with an isolated LiveView test (1 test, 1 failure); the stack points to this handler at line 105. Credential saves already handle this stale-row case, but the newly added Rename path does not. Resolve the current tenant-owned row under the mutation lock or translate the stale write into a missing-set result, then show an actionable selection error and refresh the controls. Add the two-session deletion/rename regression.

The existing targeted API/principal/console group passes 50 tests locally. Current-head CI is still red on inherited execution tests and the #2042 enum-classification failure; those are separate from this new console defect. No code was changed during this review.

BinaryBourbon (approved):

Reviewed complete head `7a5007492218b6cddda7d54ff38ad6bda647821e` against exact base `646e835ef42b215b176ae9c6f70005451fafd973`, approved-base contributor/review policies and relevant ADRs.

Reviewed all changed console handlers, shared save helper, current-row mutations, agent selection persistence, authenticated LiveView routes and prior human comments. No remaining blocking defect found in this slice. Agent mount, validation and inline save consistently use the selected set; foreign/deleted choices do not write the account default. The previous two-session Rename crash is repaired: the context reloads under the source lock, the UI handles missing rows, refreshes choices and keeps writes disabled until a valid explicit selection or creation. The invalid-selection guard covers rename, default promotion, deletion, save and clear. First-write/default behavior and presence-only rendering remain intact.

Independent diagnostics on this reviewed source: 18 agent/credential LiveView tests passed, including actual provider-row switching, selected-set persistence, deletion during provider validation and stale rename/default/delete recovery. An initial command included a nonexistent optional test path; the corrected command ran the two actual files and passed. The completed pre-restack CI run was 34752554685 (success); the cancelled duplicate is separate. These focused tests are not the full independent verification recipe.

Final-head delta verification: fetched `7a5007492218b6cddda7d54ff38ad6bda647821e` and its live base, independently proved both Git trees and the complete PR diff identical to the revisions reviewed/tested above. The upstream CI repair only relocated a dormant helper between prerequisite slices. Fresh GitHub CI [34753434253](https://github.com/managoat/fountain/actions/runs/34753434253) is queued; earlier successful-tree evidence does not replace this final-commit gate. Required checks remain pending.

Coverage (all changed paths):
- `apps/fountain/lib/fountain_web/inference_credential_save.ex`
- `apps/fountain/lib/fountain_web/live/agents_live/form.ex`
- `apps/fountain/lib/fountain_web/live/inference_credentials_live/index.ex`
- `apps/fountain/test/fountain_web/live/agent_credential_sets_live_test.exs`
- `apps/fountain/test/fountain_web/live/inference_credential_sets_live_test.exs`

Source approval does not waive inherited findings, required CI, independent Review Loop verification, maintainer acceptance or rollout/release gates.


BinaryBourbon (approved):

Final queue candidate `385c936ec943a1465f216ddac75537a7b870152c` against main `b17fd2402fb3d99cf6618c78accf9d95d7d2c727`. Its source tree is identical to reviewed integration snapshot `5437349c05c06c9d9269d2a539a126289f5197de`. The locally simulated in-order squash merges preserve each intended incremental diff; a second candidate includes its reviewed predecessor until that predecessor lands.

Independent ownership review approved all five console paths. Provider validation and writes retain the selected set; deleting a selection cannot silently retarget a write to the default. Original focused tests passed 39 cases; integrated tests passed 54 cases, including main’s new principal expiry and vault policy.

PR #2046, opened 2026-09-12
docs: document source-bound credential sets and SDK wire types

jhgaylor:
Document named credential sets, agent/launch allowlists, current principal ownership, and durable source-binding behavior. Explain replacement/deletion refusals, compatible configuration reapply, conservative environment revision checks, Codex’s machine-lifetime reservation, and the required coordinated serving-node upgrade. Refresh the ADR implementation inventory and generated index while retaining its Proposed status.

Release TypeScript SDK 2.1.0 for the added generated credential wire types. Preserve the 2.0.0 breaking error behavior and release history already on main. Package version, lockfile, User-Agent, and changelog agree; the release guard confirms the new version is available.

Validation: independent review of all 13 changed paths approved. SDK typecheck, 115 tests, build, contract verification, and 28 conformance cases passed. The complete stack combined with current main passed 6,370 tests plus 6 doctests; fresh API integration tests and generated contract checks passed. Final PR and merge-group CI remain required.

Final slice of #2018, following #2045. This PR targets main after its predecessors land through the merge queue. ADR acceptance, managed-path activation, and coordinated production rollout are separate from this implementation inventory.


BinaryBourbon (changes requested):
**Request changes** — two accuracy items, one of them the rule CLAUDE.md names explicitly. The SDK release half is correct and the manual section is good; both findings are cheap edits.

## What I checked

- `docs/nav.yml`: no page added or moved, so nothing to list. The new `## Credential sets` heading is a section on an already-listed page. No core→extension link. No em dash, no `simply`/`obviously`/`coming soon`, and no colon-ending line followed by a list item, so `docs-style.py` has nothing on the added prose. No new compile-time read, so the Dockerfile `COPY` list is unaffected.
- CHANGELOG `[Unreleased]` placement: `### Fixed` between the first `Added` and the first `Changed`. That block already repeats section headings, so this matches the file.
- Every `Added`/`Fixed` claim against the code at head: `name` + `is_default` columns and the `UPDATE ... SET name = 'Default', is_default = TRUE` backfill (`20260912050000`); `agents.inference_credential_id` and `allowed_inference_credential_ids` (`agent.ex`); `conversations.inference_credential_id`; the four set routes, the per-set write/clear and `PUT /api/claimable-users/:id/inference-credentials/:provider` in `router.ex`; "not part of sandbox identity" against ADR 0053 decision 6 and the identity-tuple comment; the `Fixed` entry against `InferenceCredentials.Source` (`:tenant_secret` is `origin: :own`) and #2025. All of it is built and stated accurately.
- The SDK bump: `package.json`, `package-lock.json`, `USER_AGENT` and `## [1.32.0]` all say 1.32.0, which is exactly what `sdk-release.mjs` checks (bump, changelog heading, user-agent string). `openapi.ts` changed earlier in the stack; #2042/#2043/#2044 carry `sdk-no-release` and this is the only bump open.

## 1. ADR 0053 still says none of this is built (blocking)

`decisions/0053-inference-credential-sets.md` on `main` reads:

> **Status:** Proposed; none of the behavior below is built.

and its frontmatter `description` and the `decisions/index.md` row both open with `Proposed, none built:`. This stack builds all seven decisions, and no stage between #2019 and here touches the ADR. CLAUDE.md is direct about this: *"If a document describes behavior that is not yet in code, say so explicitly … and remove the caveat in the PR that builds it."* This is the last stage, so the caveat lands on `main` contradicting the code that shipped with it — the same ADR-vs-code drift the 2026-07 audit was about, pointing the other way.

Update the status block and the description here (0047's row is the shape: it names what got built and in which PR), then re-run `scripts/decisions-index.sh` so the index row matches. `adr_status` may reasonably stay `Proposed` if you want the verification pass separate; the "none built" sentence may not.

## 2. The upgrade note understates who the `.env` change hits (blocking)

The note is scoped to "the inference credential". The mechanism is not:

```elixir
@process_only [@tag_key, "FOUNTAIN_TOKEN", "TRACEPARENT"] ++
                Fountain.Broker.process_only_keys() ++
                Map.values(Fountain.InferenceCredentials.env_names())

def disk_env(sprite_env), do: Enum.reject(sprite_env, fn {k, _v} -> to_string(k) in @process_only end)
```

`disk_env/1` rejects by **name**. An environment or vault secret called `ANTHROPIC_API_KEY` reaches the sprite env under exactly that key (that is the documented way to override the account credential, and `select_inference/4` reads `Map.keys(secrets)` for precisely that case), so it is now stripped from `/home/sprite/.env` too. That is a wider break than the note's first two sentences describe, and those are the sentences a reader with an Environment secret uses to decide the note is not about them. The operative sentence further down ("a script that re-reads the file for `ANTHROPIC_API_KEY` … finds nothing there now") is correct, so this is one clause: say that a secret of one of those four names is kept out of the file as well, whether it came from the account credential, an environment or a vault.

Related, and not this PR's to fix: `identity_test.exs` in #2027 carries a comment asserting the opposite — "A tenant secret of the same name is not the credential and is not stripped" — and no test covers that case either way. Worth a line on #2027 before the stack lands.

## Non-blocking

- **`docs/api.md` is untouched.** The stack adds three API surfaces and a request field. The page defers route detail to `/api/docs`, so nothing on it becomes false, but its **Inference credentials** section still reads as one credential bag per account, and **Claimable principals** does not mention that an owner can now write a credential on a principal — the capability the PR body calls the reseller's blocker, and the section an integrator reads first. This is the docs stage; two sentences would close it.
- **`docs/concepts/agent.md:139`** names `allowed_environment_ids` and `allowed_vault_ids` as the agent's scoping lists, and the field table at `docs/sdk.md:187` has a row per allowlist. `allowed_inference_credential_ids` is a third one and is in neither.
- **The unset case is missing.** "That list works like `allowed_vault_ids`: empty forbids every override, and a list of ids is an allowlist" leaves out what `nil` means. `agent.ex` says nil = any set the tenant owns, and `docs/sdk.md` states it for `allowed_vault_ids` ("A `null` permits each one"). As written a reader cannot tell what an agent that never set the field permits, which is every agent that exists today.
- **"An account starts with one, called Default."** The row is created lazily — `put_credential/5` builds it on the first write, and `create_set/3` flags the first set default. A brand-new account has no rows and `GET /api/account/inference-credential-sets` returns `[]`. "Your first key lands in a set called Default" would be exact.
- The `Added` entry names the `PUT` on the principal route; the router adds the matching `DELETE` in the same block.

BinaryBourbon (commented):
Supplement to the review above, on finding 1. I went through decisions 1-7 against the code at head rather than against the ADR's prose, because the status line this PR writes has to name what is built and what is not. **Two decisions are not fully built, so the status must not be flipped wholesale.**

## Decision by decision

| # | Built? | Evidence at head |
|---|---|---|
| 1 — named sets | yes | `20260912050000` adds `name` + `is_default`, backfills `'Default'`/`TRUE`, swaps the `user_id` unique index for `(user_id, name)` plus the `is_default` partial index. All three rules are enforced: `create_set/3` sets `is_default: is_nil(get_for_user(user_id))`, `delete_set(%Credential{is_default: true}, _)` returns `{:error, :is_default}`, `set_default/3` on an already-default set is a no-op. "Has this account connected a provider at all" asks every set — `has_any_credential?/1` folds `list_sets/1`. |
| 2 — source, not atom | **partly** | see below |
| 3 — agent names a set, launch overrides | yes | both columns, `SpriteEnv.credential_set_id/2` is `conv.inference_credential_id \|\| agent.inference_credential_id`, allowlist refused at the door as `:inference_credential_not_allowed` rather than clamped. |
| 4 — credentials in process env, not on disk | yes | `Identity.@process_only` gains `Map.values(InferenceCredentials.env_names())`. The upgrade path is covered too, which was worth checking: `write_env_file/2` writes the whole file via `Sandbox.write_file`, and `conversation_server.ex:1174` calls it on every wake/reattach, so a machine provisioned before this change loses the stale entries the next time it wakes rather than keeping them forever. |
| 5 — a tenant secret resolves the source | yes | `secret_keys/2` (names only, tenant-scoped), `select/4`'s `:secret_keys`, `Source.tenant_secret/0`, and `PlatformInference.gate/4` gaining `:environment_id`/`:vault_id` so the door asks what the provision answers. |
| 6 — not part of sandbox identity | **partly** | see below |
| 7 — principals, not sets | yes | `PUT`/`DELETE /api/claimable-users/:id/inference-credentials/:provider`, authorised as the grant's read is, value encrypted under the principal's own tenant key. |

## Decision 2: the struct has no `set_id`

The threading is all there — `ConversationServer` holds `inference_source`, `TurnMachine.ctx.inference` is a `Source.t()`, `with_inference/2` matches on `%Source{origin: ...}` and `gate/2` calls `Source.platform?/1`. But the ADR says:

> The struct carries only what something reads. `set_id` arrives with decision 1, where there is a set to name.

Decision 1 landed in #2028 and `Source` is still `@enforce_keys [:origin, :scope]`. The implementation's own reasoning is sound and I would not change the code — `source.ex` says "a field arrives with its reader" and nothing reads a `set_id` yet — but the ADR sentence promises it as part of this work and it is not here. Fix the sentence, not the struct.

## Decision 6: the codex safeguard is not built (this is the substantive one)

The identity-tuple half is done as decided: `check_attachable/4` still compares only agent, vault, environment and runtime, and the override carries the comment "Not part of the home identity tuple (ADR 0053 decision 6)". The other half of the same paragraph is not:

> until it lands, a codex agent whose set differs from another live codex peer's on the same machine is refused at admission rather than served the wrong account.

Nothing in the stack refuses it. Grepping the whole stack diff for codex turns up only test updates and the ADR text. The path is reachable with what #2034 shipped:

1. Two launches of one codex agent, each naming a different set (decision 3). Same `(user_id, agent_id, environment_id, vault_id)`, so ADR 0023 gives them one machine.
2. `Managoat.Runtimes.Codex.prepare_sandbox/3` runs `codex login --with-api-key`, which writes the machine-wide `~/.codex/auth.json`; codex 0.118+ reads only that file, never `OPENAI_API_KEY` from the process env — so decision 4 does not protect this one, which is why the ADR carves it out.
3. `Provisioning.prepare_runtime_sprite/5` is called from `conversation_server.ex:986` (fresh) **and `:1192` (wake/reattach)**, each time with that conversation's `sprite_env`.

So whichever peer provisions or wakes last owns `auth.json`, and both conversations' codex turns then run on that set's key. The other conversation silently switches accounts mid-life, and its turns bill to a subscription its launch did not name. Same tenant, so this is not a cross-tenant leak, but it is the exact outcome decision 6 says to refuse until ADR 0052 decision 5 lands.

That is a #2034 defect, not a #2046 one, and I have flagged it for that review. What it changes **here** is two things:

- The status line cannot say decision 6 is built. It should say the tuple is unchanged as decided and the codex admission refusal it depends on is still owed.
- `docs/concepts/secrets.md` and the CHANGELOG `Added` entry both state the sharing flat: "The set is not part of sandbox identity. Two conversations that differ only in their set share one machine." For a codex agent that is currently the bug, not the feature. Either the refusal lands before the stack does, or the manual says the caveat.

## So what #2046 should carry

A `**Status:**` line naming decisions 1, 3, 4, 5 and 7 as built in #2018, decision 2 as built without `set_id`, and decision 6's admission refusal as deferred to ADR 0052 decision 5 with a tracker; the frontmatter `description` reworded off "Proposed, none built"; and `scripts/decisions-index.sh` re-run in the same PR so the index row matches, which CI expects. Leaving `adr_status: "Proposed"` is defensible while decision 6 is open — 0046 and 0047 are both "Proposed" with descriptions that say exactly what shipped. What is not defensible is landing "none of the behavior below is built" on top of 5,300 lines that build it.

BinaryBourbon (commented):
**This supersedes the decision-by-decision mapping in review 5189135908 + 5189159101.** I read ADR 0053 at this PR's head. That is not the ADR on `main`: #2019's branch was substantially revised during its own review, and the stack forked before that. `origin/main` has 352 lines, this head has 279. My earlier claim that two requirements "are not in the ADR" was wrong — they are on main, in text the stack never saw. Corrected mapping below, measured against `git show origin/main:decisions/0053-inference-credential-sets.md`.

## First: the stack will revert the reviewed ADR

Only one branch in the stack contains the merged ADR commit `f07925f6`:

```
$ git branch -r --contains f07925f6 --list 'origin/stack/2018-*'
  origin/stack/2018-2-source
```

#2022 was rebased after #2019 merged; stages 3 through 13 were not. They still carry their own copy of the pre-review draft, added from a merge base (`653af872`) that predates the ADR. So:

```
$ git merge-tree --write-tree --name-only origin/main origin/stack/2018-13-docs
decisions/0053-inference-credential-sets.md
CONFLICT (add/add): Merge conflict in decisions/0053-inference-credential-sets.md
```

An add/add conflict is the good outcome — it forces a human to look. The bad outcome is resolving it by taking the branch side, which silently rolls the ADR back to the pre-review draft and drops the acceptance table, the source-binding rules and the alias inventory. **The status edit this PR owes must be written on top of main's 352-line text, not this branch's 279-line one.**

## Corrected BUILT / NOT BUILT

| # | Verdict | Detail |
|---|---|---|
| 1 — a set is a row, not a vault | **built** | Migration, index swap, backfill; `Validator` still pings on save and the console still reports presence without decrypting. |
| 2 — source, not atom | **partly** | see below |
| 3 — agent names a set, launch overrides | **built as schema/API**, but main adds "Replacement and deletion still follow decision 2's source-binding rules", which is not satisfied (see 2). |
| 4 — auth inputs leave the shared env file | **partly** | see below |
| 5 — one runtime-aware resolver | **partly** | see below |
| 6 — not part of sandbox identity | **guard not built** | unchanged from my last review, and main's text is stricter |
| 7 — principals, not sets | **built** | routes, authz, principal's own DEK |

### Decision 2 — three of four parts missing

Main allows the carrier refactor to be plumbing only, and that part is done. The rest is not:

- **`set_id`.** "Add `set_id` with named sets." Named sets landed in #2028; `Source` is still `@enforce_keys [:origin, :scope]`.
- **The persisted binding.** "Before allowing different sources in a shared sandbox, persist a non-secret resolved source reference with the peer and snapshot it for turn accounting … Resume and reattach recover that binding rather than resolve the account's current default again." `inference_source` is `ConversationServer` in-memory state only. The wake path (`conversation_server.ex:1174`–`1192`) re-runs `load_tenant_state/2` and `select_inference/4`, which is re-resolution, precisely what the sentence forbids. No revision or version column exists anywhere, so "distinguish credential replacements within one set" has nothing to distinguish with.
- **Silent rerouting.** "A deleted, ineligible or replaced source cannot silently reroute an existing peer: require an explicit new selection or reauthentication before another turn." Both new foreign keys are `on_delete: :nilify_all`, deliberately — the migrations say so ("deleting a credential set must return the conversations that named it to the agent's"). Delete a set and every agent and live conversation pinned to it silently falls back to the account default on its next wake. That is the ADR's named prohibition implemented as its opposite.

### Decision 4 — the alias inventory is missing, and it is a live hole

The canonical four are handled, and the upgrade path you asked about is genuinely fine: `write_env_file/2` rewrites the whole file via `Sandbox.write_file` and runs on every wake, so an old machine loses its stale entries. But main's text is explicit:

> Four credential columns do not mean four env names: OpenCode emits `GOOGLE_GENERATIVE_AI_API_KEY` for the same Gemini credential that the Gemini runtime receives as `GEMINI_API_KEY`.

`Identity.@process_only` is `Map.values(InferenceCredentials.env_names/0)` — the four canonical names. `Managoat.Runtimes.OpenCode.default_env/2` emits the alias, with a comment saying why: "opencode's google provider is `@ai-sdk/google`, which reads `GOOGLE_GENERATIVE_AI_API_KEY` only." Neither `InferenceCredentials.@env_names` nor `Broker.@inference` knows that name. **So for an opencode agent on a google model, the Gemini credential still lands in `/home/sprite/.env` and is still shared across every conversation on the machine.** The property #2027 claims to establish does not hold for that runtime/provider pair.

`inference_credentials_env_names_test.exs` does not catch it: it asserts that every entry in `env_names/0` agrees with the broker's table, one direction only. Its own moduledoc names the other direction as the bug — "a name only the broker knows is an override selection cannot see, which is the bug decision 5 exists to fix" — and then does not test it, and neither table carries the alias anyway.

### Decision 5 — same root, plus two unbuilt clauses

The env/vault resolution and the door gate agree for the four canonical names, which is the real fix in #2023/#2025 and it is good. Not built:

- **Alias normalization.** `secret_keys/2` matches secret keys against the same four names, so a vault secret called `GOOGLE_GENERATIVE_AI_API_KEY` — the name opencode actually reads — resolves no `:tenant_secret` and the turn still stamps as platform inference. That is decision 5's own bug, surviving under the alias.
- **"Remove competing inference auth inputs from the runtime launch so a second variable cannot change authentication after selection."** Nothing removes anything.
- **"An unusable explicit tenant source returns an actionable error instead of silently using platform inference."** `select/4`'s `{:error, :no_credential}` becomes `Source.missing/0` and the conversation provisions anyway.

### Decision 6 — unchanged, and corroborated

Main's version is stricter than the draft I quoted: an admission guard comparing **resolved source identities and revisions/generations, not set IDs**, an atomic compatibility reservation before any auth-file write, and "if this cannot be proved, keep multi-source Codex admission disabled". None of it exists. `check_attachable/4` compares agent, vault, environment and runtime. I reached this from `~/.codex/auth.json` being machine-global and rewritten by `prepare_runtime_sprite/5` on every wake; review-2034 reached the same defect from `check_attachable/4`.

### The ordering constraint

Main's acceptance table says "Shared auth binding is required before multi-source execution; schema/API slices may be reviewed earlier **without enabling that behavior**." This stack enables multi-source execution — #2034's per-launch override is live — with the shared auth binding stage unbuilt.

## What this changes for #2046

The status edit is still owed and is still the blocker, but it is a bigger edit than I said: written against main's text, it should record decisions 1 and 7 built, 3 built as schema/API, 2/4/5 partly built with the specific clauses above named, and 6's admission guard not built. Plus `scripts/decisions-index.sh`. The manual and CHANGELOG caveat I asked for on the codex case stands, and the alias gap deserves its own tracker against #2027 rather than a docs note.

BinaryBourbon (commented):
**Correction to one finding in review 5189206166.** I said "the stack will revert the reviewed ADR". That is wrong as framed, and I am withdrawing it as a hazard of this burndown. It is a hazard of one path only.

**Why I was wrong.** My `merge-tree` result was real — merging `origin/main` into an un-rebased stage still conflicts add/add on the ADR. But nobody merges these branches; the burndown rebases each stage `--onto main <recorded tip of the branch below>`, which excludes the branch's own copy of the ADR-add, so only the stage's own commits replay. The full ten-stage simulation runs clean and ends at 371 lines = main's 352 plus #2042's 19, with main's hardening intact ("persist a non-secret" ×1, "Upgrade and" ×1, "acceptance" ×3, "alias" ×4).

I can watch it happening rather than take it on trust. Stages 2 and 3 already carry main's 352-line ADR; stage 3 picked it up *between two of my own checks* (`git branch -r --contains f07925f6` returned one branch, then two). Stages 4 onward still read 260 because they are simply not rebased yet, not because they carry a competing edit. My earlier per-stage `gh pr diff --name-only | grep decisions/` — which appeared to show #2025 rewriting the ADR — was a moved-base artifact: with base rebased and head not, that diff reports the merge base's absent file as a 260-line addition. Measuring the branches directly is what settles it.

**What survives unchanged:** the status edit must still be written against main's 352-line text, because the branch a contributor has checked out currently shows the 279-line draft. That is the whole of what is left of this finding.

## Routing the other findings

Two of these are not #2046's to fix and should be tracked where the code is:

- **`on_delete: :nilify_all` on both new foreign keys → #2031 and #2034.** `20260912051000` (agents) and `20260912052000` (conversations) both choose it deliberately and say so. Deleting a credential set therefore returns every agent and every live conversation pinned to it to the account default, with no signal. Main's decision 2: "A deleted, ineligible or replaced source cannot silently reroute an existing peer: require an explicit new selection or reauthentication before another turn." This is that prohibition implemented as its opposite, and it is reachable today because nothing persists a binding to notice the change.

- **The `GOOGLE_GENERATIVE_AI_API_KEY` alias → a tracker against #2027, and it also holes #2023.** One root, two fixes defeated. `Identity.@process_only` is the four canonical names, so the credential still reaches the shared `/home/sprite/.env` for an opencode agent on a google model (`open_code.ex:67-69` says why the alias exists). And `secret_keys/2` matches the same four, so a vault secret under the alias — the name opencode actually reads — resolves no `:tenant_secret` and the turn still stamps as platform inference. #2027's disk fix and #2023's billing fix have the same hole from the same table. `inference_credentials_env_names_test.exs` cannot catch it: it asserts only that every `env_names/0` entry agrees with the broker, while its own moduledoc names the reverse direction as the bug, and neither table carries the alias.

- **The ordering constraint stays on this PR's mapping.** Main's acceptance table: "Shared auth binding is required before multi-source execution; schema/API slices may be reviewed earlier **without enabling that behavior**." #2034 enables it, with the shared auth binding stage unbuilt. That is the clearest statement of why the stack should not stop at #2034, and it is why decision 6's missing admission guard is a defect rather than a deferral.

The blocking finding on #2046 is unchanged and is now just the one: the ADR status block, written against main's text, plus `scripts/decisions-index.sh`.

BinaryBourbon (approved):
AI-assisted adversarial review — APPROVED

Reviewed all 13 changed files against base 1298fb384ad5, the approved-base contributor/review policies and ADRs, actual selection/principal/API callers, SDK exports and release configuration, and the prior reviews including their corrections. No new blocking defect found in this documentation/release slice.

The ADR retains Proposed status, replaces the obsolete “none built” inventory, preserves its acceptance requirements, and refreshes the index. The manual correctly limits Codex sharing to a machine-lifetime binding requiring a new sandbox after incompatible credential replacement; it names whole-environment revision invalidation, coordinated serving-node upgrade, process-env/shared-file behavior, principal current-owner authority, and unbuilt per-peer auth directories/managed user execution. These are implementation descriptions, not deployment or activation evidence. SDK 1.33.0, lockfile, user agent and changelog agree, with no handwritten client-method claim.

Exact-head evidence: [Docs prose gates](https://github.com/managoat/fountain/actions/runs/34750539994/job/103706253840), [Decisions validation/index freshness](https://github.com/managoat/fountain/actions/runs/34750539979), [SDK release gate](https://github.com/managoat/fountain/actions/runs/34750539922), SDK matrices, and API-contract/release-boot checks pass. I read the actual prose logs: style/STE/de-stink reports remain advisory under base policy, including sentence-length advice in secrets.md; successful job status does not mean zero wording advice. The conditional Docs-only job is skipped on this code-inclusive stack. The cancelled duplicate CI's classification failures are separate from the completed run.

This approval covers this slice only. Completed full-stack CI remains red on inherited execution/schema tests, and upstream mutation/console findings remain open. It does not waive those fixes, required CI, ADR/maintainer acceptance, coordinated rollout, cleanup evidence or release/deployment gates.

BinaryBourbon (approved):

Reviewed complete head `4d7e712ca076f188b72f0d9e5fa59a2ce6fc4e64` against exact base `7a5007492218b6cddda7d54ff38ad6bda647821e`, approved-base contributor/review policies and relevant ADRs.

Reviewed all documentation, ADR/index, omission record and TypeScript release changes, against the actual resolver/binding, principal-owner authorization, API rendering, process-env filtering, SDK exports and release configuration. No new blocking defect found in this documentation/release slice. ADR 0053 remains Proposed and distinguishes implementation inventory from acceptance or deployment. The manual explains current-owner authority, no provider ping for principal writes, whole-map env revision invalidation, machine-lifetime Codex binding, new-sandbox requirements, process-only aliases, coordinated serving-node upgrade, and unbuilt per-peer auth/managed execution. Version 1.33.0 agrees across package, lockfile, user agent and changelog. The omission reason accurately covers plain-HTTP owner management and adds no new wildcard scope.

Independent diagnostics on the tree-identical pre-restack revision: 43 manual structure/link/controller tests, TypeScript typecheck, TypeScript contract verification and 111 SDK tests passed. I read the actual successful CI prose job logs: style/STE advice remains, including two 26-word sentences in secrets.md, and is advisory under the approved-base policy. The completed CI run34752556507, Decisions and SDK release gate passed; the conditional docs-only job was skipped, and the cancelled duplicate is not extra test evidence.

The new reapply prose describes the source/configuration transaction that exists in the implementation. A separately reproduced inherited #2031 legacy-wake/reapply race can later overwrite that binding from stale actor state. This documentation approval does not certify that execution guarantee across the race or waive that upstream source blocker; fix it in its owning prerequisite before landing the stack. No acceptance, fleet cleanup, activation, release or deployment evidence is implied.

Final-head delta verification: fetched `4d7e712ca076f188b72f0d9e5fa59a2ce6fc4e64` and its live base, independently proved both Git trees and the complete PR diff identical to the revisions reviewed/tested above. The upstream CI repair only relocated a dormant helper between prerequisite slices. Fresh GitHub CI [34753435110](https://github.com/managoat/fountain/actions/runs/34753435110) is queued; earlier successful-tree evidence does not replace this final-commit gate. Required checks remain pending.

Coverage (all changed paths):
- `CHANGELOG.md`
- `decisions/0053-inference-credential-sets.md`
- `decisions/index.md`
- `docs/api.md`
- `docs/build/anonymous-visitors.md`
- `docs/concepts/agent.md`
- `docs/concepts/secrets.md`
- `docs/sdk.md`
- `sdk/contract/omissions.json`
- `sdk/typescript/CHANGELOG.md`
- `sdk/typescript/package-lock.json`
- `sdk/typescript/package.json`
- `sdk/typescript/src/http.ts`

Source approval does not waive inherited findings, required CI, independent Review Loop verification, maintainer acceptance or rollout/release gates.

BinaryBourbon (approved):

Final queue candidate `2921bcf834f5d13ff58beaa00119758a8e53424f` against main `b17fd2402fb3d99cf6618c78accf9d95d7d2c727`. Its source tree is identical to reviewed integration snapshot `7f9f3f5a71c025c07a6e9960482c0db0f7b3a8a9`. The locally simulated in-order squash merges preserve each intended incremental diff; a second candidate includes its reviewed predecessor until that predecessor lands.

The complete 18-path cumulative queue diff combines the independently approved five-path console slice (#2045) and 13-path documentation/SDK slice. Both source reviews found no blocker. SDK 2.1.0 preserves main’s 2.0.0 breaking behavior and release history; package, lockfile, User-Agent and changelog agree. The registry/release guard passed. SDK typecheck, 115 tests, build, contract verification and 28 conformance cases passed. Documentation matches current source/reapply constraints and keeps ADR0053 Proposed.

PR #2047, opened 2026-09-12
fix(conversations): recheck absolute transport deadlines

jhgaylor:
An execution transport’s relative timer can wake just before its absolute deadline because the remaining time was rounded down to milliseconds. The handler immediately retired the journal, allowing a still-valid turn to be persisted as `interrupted` rather than eventually failing with `wall_time_limit`.

The handler now rechecks the absolute deadline and reschedules early wakes. Scheduling rounds up to the next millisecond. Actual expiry still uses the existing guarded retirement path; explicit close and failure handling retain their interruption semantics.

A deterministic regression delivers an early wake, verifies the journal and turn remain active, and exercises the real writer afterward. It fails against the old handler and passes with the fix. The complete transport and bounded-lifecycle modules pass **37 tests, 0 failures**, including actual independent deadline expiry.

Full `mix precommit` passed on OTP 28.3 / Elixir 1.19.2: compile, formatting, strict Credo, Dialyzer (zero errors), Sobelow, production release assembly, and all suites: **6 doctests + 5,615 core tests, 144 Buzz tests, 33 Support tests; zero failures** (2 skipped, 7 excluded). The full run used the original failing CI seed 495260 with max_cases 4 and unchanged assertions/timeouts.

Discovered while landing #1914: [required CI failure](https://github.com/managoat/fountain/actions/runs/34688178958/job/103538898924). The transport, guard, actor helper and failing lifecycle test were identical on that PR’s merge tree and main. This prerequisite fixes that inherited behavior rather than changing the expected terminal outcome in the test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -271,6 +271,10 @@ upgrade, is in
 
 ### Fixed
 
+- Bounded execution transports recheck their absolute deadline before retiring
+  a turn. A timer that wakes early no longer interrupts a valid turn or hides
+  its eventual wall-time-limit outcome.
+
 - Sandbox `/diff` returns 422 when Git fails after repository discovery,
   instead of reporting an empty diff. Intentional byte-cap truncation still
   succeeds (#1904).
--- a/apps/fountain/lib/fountain/conversations/execution_transport.ex
+++ b/apps/fountain/lib/fountain/conversations/execution_transport.ex
@@ -81,8 +81,7 @@ defmodule Fountain.Conversations.ExecutionTransport do
     Process.flag(:trap_exit, true)
     execution = Keyword.fetch!(opts, :execution)
     owner = Keyword.fetch!(opts, :owner)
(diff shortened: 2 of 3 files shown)


BinaryBourbon (approved):
Reviewed exact head e627aaf91c8a7abfea57376b6b592eee937c9a65 independently. No actionable findings.

The early wake now preserves the active transport and reschedules against the authoritative absolute deadline; rounding up avoids the original sub-millisecond truncation. Due expiry still follows the guarded retirement path, while explicit close, actor death and failed writes keep their existing interruption behavior. The regression exercises the actual writer after a deterministic early wake, and the existing lifecycle coverage verifies eventual wall-time-limit termination.

Confirmed the original failed CI merge and main shared identical transport/guard/actor/test blobs. Inspected the deterministic red result (old handler fences the transport) and the final signed diff. Validation reported by the coder: focused transport/lifecycle 37 tests, zero failures; full precommit including static checks, production release, 6 doctests + 5,615 core tests + 144 Buzz + 33 Support tests, zero failures. Approved.

PR #2050, opened 2026-09-12
refactor(conversations): move the reattach family out of the server

jhgaylor:
The reattachment family — `reattach_running_turn/1`, `reap_orphan_sessions/1`, `attempt_session_attach/4`, `mark_orphan/3` and `find_running_turn/1` — becomes `Fountain.Conversations.Reattachment`, the #1369 subtraction pattern, following #2037.

`Reattachment` is already where the other half of this path lives: `acp_peer/3` is what `attempt_session_attach/4` calls on its ACP arm, and `wait_for_runner/2`, `fail_transport/2` and `disconnect_runner/1` are the same recovery concern. The move puts the whole path in one module and takes **163 lines** out of `conversation_server.ex`, from 2629 to **2466**.

## The move is verbatim

The bodies are byte-identical to the originals apart from three edits, diffed whitespace-normalized:

| line | before | after |
|---|---|---|
| `reattach_running_turn/1` | `defp` | `def` — the server still calls it |
| `find_running_turn/1` | `defp` | `def` — the server calls it on the transient-reattach-failure arm too |
| ACP arm of `attempt_session_attach/4` | `Reattachment.acp_peer(...)` | `acp_peer(...)` — now a local call |

Nothing else, and no behaviour change. `reap_orphan_sessions/1`, `attempt_session_attach/4` and `mark_orphan/3` stay private.

The state coupling needed no adapter. `Reattachment`'s local convention is to take and return the server's `state` map whole (`acp_peer/3`, `wait_for_runner/2`, `disconnect_runner/1` all do), and it reaches for `Pending.from_state`/`into_state` only where it touches that sub-struct. The moved functions do the same, so they moved as they were rather than being reshaped.

`self()` still names the server: every one of these is a plain call from the server process, so `Managoat.Sandbox.attach(..., owner: self(), ...)`, `Process.send_after/3` and `Process.monitor/1` all bind exactly as before.

## The pin goes to 2549, not 2466

`main` was at `@pin 2641`. The `#1766`/`#1767` campaign is still in flight below it, so the pin is not lowered to the file's new length — the moduledoc's own guidance (#1565).

I measured the campaign's remaining net myself rather than reusing #2037's figure, by diffing each **chain tip** against the base it forks from. Summing per-PR `additions - deletions` overcounts a stack, and #1973 is now merged:

| chain | net to `conversation_server.ex` |
|---|---|
| #1975 (`lifecycle-fence`, off `main`) | +10 |
| #1978 → #1980 → #1981 → #1982 → #1983 → #1996 → #1998 → #1999 → #2000 → #2001 → #2002 → #2003 → #2006 → #2007 | +58 |
| #1977 (via #1976) | +5 |
| #1979 (`termination-attach-order`) | 0 |
| **remaining total** | **+73** |

That lands the file at **2539**. `@pin 2549` is that plus 10 lines for the review rounds still to come.

**#2037 measured +68 a day ago; it is +73 today** — two more rounds on #1981 and #1982 moved it, which is exactly the drift the moduledoc warns about. A pin of 2513 (2466 + the older +47 estimate) would have gone red the moment the campaign landed. 2549 is still 92 below the current pin, so the ratchet moves down hard; tighten it to the real length in a follow-up once the campaign has landed.

## Conflict surface

The move is at lines 1249-1411 of the old file. The campaign's remaining PRs edit roughly 1543-1600 and 2038-2130, far outside it, and none of them touches the pin line — so `main`'s lowering wins silently, as it did for #2037.

## Verification

- `mix compile --warnings-as-errors` clean.
- `mix format --check-formatted` clean.
- `mix credo --strict` from the umbrella root: 862 files, **no issues**. The three `_unsafe_` reads (`_unsafe_get_conversation!`, `_unsafe_output_bytes_by_stream`, `_unsafe_orphan_turn`) carry inline `# ownership:` comments rather than a new entry in `.credo.exs`'s `exempt_paths`, as in #2037 — `Reattachment` is not a GenServer or a system sweep, and it reads exactly the conversation the server is bound to and the running turn it looked up for that conversation. `reattachment.ex` was already off the exempt list and already justified its existing `_unsafe_` calls this way.
- **The full `apps/fountain` suite: 6 doctests, 5731 tests, 0 failures, 2 skipped**, against a dedicated database (the shared `fountain_test` is contaminated — it carries an `inference_credentials.name NOT NULL` column that no migration on `main` creates).
- The size test passes at 2466 against the new pin.

Refs #1369, #2032, #1767.

### DIFF
--- a/apps/fountain/lib/fountain/conversations/conversation_server.ex
+++ b/apps/fountain/lib/fountain/conversations/conversation_server.ex
@@ -1164,7 +1164,7 @@ defmodule Fountain.Conversations.ConversationServer do
               sandbox_started_at: Lifecycle.clock_start(sandbox)
           }
 
-          new_state = reattach_running_turn(%{new_state | current_turn: nil})
+          new_state = Reattachment.reattach_running_turn(%{new_state | current_turn: nil})
 
           new_state =
             Reattachment.finish_runner_reconnect(
@@ -1227,7 +1227,7 @@ defmodule Fountain.Conversations.ConversationServer do
             "transient; sandbox row left untouched"
         )
 
-        running_turn = find_running_turn(state.conversation_id)
+        running_turn = Reattachment.find_running_turn(state.conversation_id)
 
         if sandbox.provider == "runner" and
              reason in [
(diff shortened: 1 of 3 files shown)


BinaryBourbon (approved):
Independent review. Verified in a clean worktree against a dedicated database. **Approving** — the move is genuinely verbatim, `self()` is safe, the pin is sound, and the conflict surface is provably empty. Details below, including two corrections to the PR body and an answer on the +47 / +73 question.

## 1. The move is verbatim — confirmed

I extracted the removed block from `origin/main`'s ancestor and the added block from the branch, normalized whitespace, and diffed. After setting aside the two call-site lines (which appear correctly as `+ Reattachment.…` in the server), the entire difference is:

- `defp reattach_running_turn` → `def`
- `defp find_running_turn` → `def`
- `Reattachment.acp_peer(...)` → `acp_peer(...)`
- three `# ownership:` comment blocks (5 lines)

Nothing else. `reattachment.ex` is a single append-only hunk at line 186 of a 189-line file — no pre-existing code reordered or touched.

I also checked the trap that a whitespace diff would not catch: **alias resolution**. `Connection`, `Output`, `Conversations` and `Pending` all resolve to the same `Fountain.Conversations.*` modules in both files; everything else in the moved block is fully qualified (`Managoat.Sandbox`, `Fountain.Conversations.Identity`, `Fountain.RuntimeDispatch`, `Fountain.Repo`); `Logger` is required in both; `import Ecto.Query` is function-local and travelled with `find_running_turn/1`. No same-name-different-module substitution.

## 2. `self()` still names the server — confirmed

Traced the full chain for both call sites, which are both inside `do_reattach/6`:

`handle_continue(:provision, state)` → `provision_with_rows/3` → `dispatch_provision/7` → `reattach/6` → `do_reattach/6`

`reattach/6` wraps the call in `Fountain.Telemetry.span/3`, which invokes the fun inline (`{result, extra} = fun.()` inside `:telemetry.span`, both synchronous). There is no `Task` or `spawn` anywhere in `conversation_server.ex`. So `Managoat.Sandbox.attach(..., owner: self(), ...)` at `reattachment.ex:280`, and the `Process.send_after/3` and `Process.monitor/1` calls in the module, all bind to the server pid exactly as before.

## 3. Conflict surface — empty, and proved a stronger way

The PR body argues this from line ranges, and that argument is **slightly wrong**: `#1982` and `#2007` do have hunks at base lines 1173–1227, much closer to the block (base 1270–1430) than "1543-1600 and 2038-2130" suggests. The conclusion survives, but on better evidence:

- For every campaign tip (`#1977 #1979 #1980 #1981 #1982 #1996 #1998 #2002 #2007`), the reattachment block is **byte-identical** to its own base. No campaign PR touches the moved code.
- Every campaign tip still contains both call-site lines verbatim.
- Real 3-way merges: `merge-tree(main-without-2050, tip)` and `merge-tree(2050, tip)` produce **identical conflict sets**. `#1982` and `#2007` already conflict with main without this PR (they fork from a pre-#2037 base); `#1977 #1979 #1980 #1981` auto-merge cleanly either way.

**This PR adds zero new conflicts.**

## 4. The pin — sound, but the comment is already stale

`main` moved while this was open: **#1975 landed** (a18d1608), taking the file 2629 → **2639**. So:

- The PR still merges cleanly; post-merge the file is **2476**, not 2466.
- The projection is unchanged, because #1975's +10 simply moved from "remaining" to "landed": 2476 + 63 = **2539**, exactly the author's number.
- 2549 is 92 below the current pin of 2641. The ratchet moves down. ✅

The size-test comment still lists "#1975 +10" among the remaining open PRs. Harmless to the arithmetic, but worth correcting in the follow-up so the next person measuring does not double-count it.

One PR family neither of us counted: **stack/2018** (#2022–#2046) also touches this file. It contributes **−1**, so it does not threaten the budget — noting it only so the follow-up's measurement is complete. Final projection **2538**, 11 lines of headroom.

Lowering to 2549 rather than 2466 is the right call and follows the moduledoc's own guidance (#1565's zero-headroom mistake).

## 5. Gates

Run from a clean worktree on a dedicated database:

- `mix compile --warnings-as-errors` — clean
- `mix format --check-formatted` — clean
- `mix credo --strict` (umbrella root) — 862 files, **no issues**
- `mix test test/fountain/conversations/` — 1340 tests, 0 failures
- Full `apps/fountain` suite — **6 doctests, 5731 tests, 0 failures, 2 skipped**

The suite logs `ArgumentError: comparing e.id with nil` from `ExecutionGuard._unsafe_actor_gate/3` and some `MatchError`s from terminating GenServers. **Not from this PR** — it touches three files, none of them `execution_guard.ex`, and the server changes are confined to the moved block plus two call sites. No test fails on them. Flagging separately: an unguarded nil `id` reaching `where: e.id == ^id` on `:lifecycle_check` looks like a latent bug in the #1973 fence worth its own look, independent of this refactor.

## 6. Ordinary review

`Reattachment` remains coherent: it now owns the whole interrupted-turn recovery path, and the moved family sits naturally beside `acp_peer/3`, `wait_for_runner/2`, `fail_transport/2` and `disconnect_runner/1`.

The two new `def`s are justified — `reattach_running_turn/1` and `find_running_turn/1` are each called from the server (lines 1167 and 1230) and nowhere else. The other three correctly stayed `defp`. Minimal public surface.

The three `# ownership:` comments do name what establishes ownership, and match the convention the module already used at lines 54, 130 and 153. Worth saying explicitly that `reattachment.ex` is deliberately **not** in `.credo.exs`'s `exempt_paths` while five sibling helper modules are — inline justification per call site is the stricter and more auditable choice, and I would keep it that way.

Nothing is left dead in `conversation_server.ex` (warnings-as-errors and credo would both catch an orphaned alias or private function).

### Nits, none blocking

1. **`acp_peer/3` is now public with no external caller.** Its only caller became local in this PR. It could be `defp`. Leaving it preserves "pure move", so fold it into the follow-up rather than here.
2. **The module does not say these must run in the server process.** Two of them are now public and depend on `self()` being the server. One line in the moduledoc would protect an invariant that is currently only true by convention.

## On +47 vs +73 — the author's method is right

I measured both independently.

- **Tip vs fork base**: `#2007` head is 2743; its merge-base with main (3f59a76a) is 2685. Net **+58**. Reproduced the author's number exactly.
- **Summing per-PR own-commit deltas** over the same chain: **+141**.

Neither is +47, so that figure is not reproducible today by either method — most likely measured before the review rounds on #1981 and #1982.

The reason summing fails here is concrete and checkable: **the stack is not linearly rebased.** `#1980 → #1981` and `#1981 → #1982` are *not* ancestor links (`git merge-base --is-ancestor` says no). Those two branches were rebased independently onto newer bases, so each one's diff against its own merge-base re-applies work already counted in its parent — #1981 shows +50 and #1982 +52 against bases that are not their parents' heads. Summing double-counts that.

The tip method is sound here because I verified `#2007` is a content superset of the chain: 92–100% of each earlier PR's added lines to this file are present at the tip (the small gaps are lines later rewritten downstream, as expected in a stack).

**Use tip-vs-fork-base. The chain is +58, total remaining is +62 including stack/2018's −1, and the file lands at 2538 against a pin of 2549.**

The general rule worth writing down: sum per-PR deltas only when each PR's merge-base is its parent's current head. Check that before trusting the sum — here it was false for two links out of fourteen, and that was the whole discrepancy.

Issue #1429, opened 2026-09-03
await_registered can give up after one lookup, reinstating the #800 duplicate-sandbox race


jhgaylor:
`ConversationServer.await_registered/2` exists to stop the #800 duplicate-sandbox race: a prompt arriving on a second pod sees a `pending` row, misses the registry, and takes the `:create_new` arm while the first server is still ~20 s from finishing. Two servers, two sprites, a name conflict that kills the loser, and an orphan `ready` row.

**Under load it can give up after a single lookup, which reinstates exactly that race.**

## The defect

```elixir
deadline = System.monotonic_time(:millisecond) + timeout_ms
do_await_registered(conv_id, deadline)

defp do_await_registered(conv_id, deadline) do
  case whereis(conv_id) do
    pid when is_pid(pid) -> {:ok, pid}
    nil ->
      if System.monotonic_time(:millisecond) >= deadline do
        :timeout
      else
        Process.sleep(@registry_poll_ms)
        do_await_registered(conv_id, deadline)
      end
  end
end
```

The deadline is checked **before** the first sleep. If the process is descheduled between computing `deadline` and reaching that check for longer than `timeout_ms`, the function returns `:timeout` having polled the registry exactly once. "Wait for the registry to catch up" degrades to "look once", which is the pre-#800 behaviour.

In production `timeout_ms` defaults to `@registry_settle_ms` = 3000 ms, so this needs a 3-second stall. Unlikely, not impossible under heavy GC or scheduler pressure, and the cost when it happens is the full #800 failure: a duplicate sandbox, ~21 s of wasted provisioning, and an orphan row.

## Why CI sees it

`config/test.exs:67` sets `conversation_registry_settle_ms` to **150**, against `@registry_poll_ms` of 50. That is three polls of headroom, so a 150 ms scheduling stall is enough — and on a runner executing one of six partitions at `max_cases: 20`, that happens.

Observed on PR #1428 (a test-only change touching nothing near this code), partition 1:

```
1) test wake_conversation/2 a pending sandbox whose server appears during the
   settle window is handed the prompt, not raced (#800)
   ** (Mimic.UnexpectedCallError) expected Horde.DynamicSupervisor.start_child/2
      to be called 0 time(s) but it has been called 1 time(s)
```

The wake gave up on the registry and created a second server, which is the assertion's whole point.

## This was addressed once already, on the wrong side

[...]

PR #2052, opened 2026-09-12
test(conversations): wait for the probe to appear in the registry

jhgaylor:
`probe/2` in `termination_client_test.exs` asserted `ConversationServer.whereis(conv_id) == pid` on the **first** read after `start_supervised!/1`. `Probe.init/1` registers in `Fountain.ConversationRegistry` — Horde's CRDT-backed registry — and that write is not guaranteed visible to `lookup/2` the instant init returns. Every test in the file was a coin flip on registry propagation.

It surfaced on **#1982's** partition 6, a PR that touches none of this code:

```
1) test rolling deploy forwards only attribution keys to the actor
   Assertion with == failed
   left:  nil
   right: #PID<0.3654.0>
   test/fountain/conversations/termination_client_test.exs:112
```

The fix waits for the condition being asserted, bounded, with a `flunk` naming what never happened — rather than reading once and hoping.

Three consecutive green local runs (8 tests each); the same file failed in CI on the first read, which is the asymmetry you expect from a propagation race rather than a logic error.

These tests arrived with #1981 earlier today, so this is same-day cleanup of my own change, not an inherited flake.

Refs #1767, #1981.

### DIFF
--- a/apps/fountain/test/fountain/conversations/termination_client_test.exs
+++ b/apps/fountain/test/fountain/conversations/termination_client_test.exs
@@ -125,8 +125,27 @@ defmodule Fountain.Conversations.TerminationClientTest do
 
   defp probe(conv_id, reply) do
     pid = start_supervised!({Probe, {self(), reply, conv_id}})
-    assert ConversationServer.whereis(conv_id) == pid
+    await_registered(conv_id, pid)
     pid
+  end
+
+  # `Probe.init/1` registers in the cluster-wide Horde registry, and a CRDT
+  # registry does not guarantee the write is visible to `lookup/2` the instant
+  # `start_supervised!/1` returns. Asserting on the first read makes every test
+  # in this file a coin flip on registry propagation, which is how partition 6
+  # went red on an unrelated PR. Wait for the thing being asserted.
+  defp await_registered(conv_id, pid, remaining \\ 200) do
+    cond do
+      ConversationServer.whereis(conv_id) == pid ->
+        :ok
+
+      remaining == 0 ->
+        flunk("probe #{inspect(pid)} never became visible in the registry for #{conv_id}")
+
+      true ->
+        Process.sleep(10)
+        await_registered(conv_id, pid, remaining - 1)
+    end
   end
 
   defp events(ctx), do: Audit.list_for_user(ctx.user.id, action_prefix: "conversation.terminated")


BinaryBourbon (approved):
Reviewed independently (own worktree, own `fountain_review_2052` DB, 8 runs with varying seeds — all green).

**Diagnosis checks out.** `Fountain.ConversationRegistry` is Horde's CRDT registry (`application.ex:75`), and `ConversationServer.whereis/1` is a single `Horde.Registry.lookup/2`. `ConversationServer.await_registered/2` already exists in production for exactly this reason (its docstring: "never decides on one lookup (#1429)"), which independently corroborates that a registration write here is not synchronously visible to a lookup — this isn't a novel theory, it's a previously-fixed production bug recurring in a test helper. I couldn't force-reproduce the CI timing locally (quiet single-node box), but the mechanism is sound and matches the observed asymmetry (`nil`, not a wrong pid).

Ruled out the "Probe crashed and restarted" alternative: `start_supervised!/1` gives it a `:temporary` restart, so a real crash would leave `whereis` permanently `nil` and the loop would correctly `flunk` rather than silently pass — it doesn't paper over that failure mode.

**Loop mechanics are fine.** Handles both `nil` and a mismatched pid (same `==` comparison each poll), terminates on every branch, `flunk` names the pid and conv_id. 200×10ms = 2s is reasonable — shorter than the production `@registry_settle_ms` (3s), which is fine since this is intra-node, not cross-node cluster sync.

**One non-blocking suggestion:** this reimplements polling logic that already exists as `ConversationServer.await_registered/2` (returns `{:ok, pid} | :timeout`). Calling that and asserting `{:ok, ^pid} = ConversationServer.await_registered(conv_id)` would drop ~15 lines and exercise the same helper other tests already rely on (see `sandbox_reset_test.exs:172`). Separately, naming the local helper `await_registered/2` when `ConversationServer.await_registered/2` exists with different semantics (no pid arg, returns a tuple) is a bit of an attractive nuisance for a future reader/refactor — a name like `await_probe_registered` would avoid the collision. Neither blocks this PR.

**Point 4 (siblings with the same shape):** grepped every test file that touches `ConversationServer.whereis` / `Horde.Registry.{register,lookup}`. Most are fine (`sandbox_reset_test.exs` already uses the production `await_registered/2`; `execution_deadline_worker_test.exs` is protected by a polling `await_state`). One real hit:

`apps/fountain/test/fountain/conversations/lifecycle_actions_test.exs:322` ("casts `:machine_gone` to every other live server on the machine") registers a stand-in process, does `assert_receive :registered`, then immediately calls `Lifecycle.stop_cotenants/5` — which internally does the exact single-read `ConversationServer.whereis/1` this PR just fixed (`lifecycle.ex:529`). If the registry lookup lags, `stop_cotenants` treats it as "no live server" (a legitimate case elsewhere in the same file) and the cast never fires, so `assert_receive {:cotenant, ...}` times out. Same shape, not covered by this PR.

Weaker/lower-confidence sibling shape (register → signal → immediate use, no retry) in `sandbox_reset_reconciler_test.exs`, `sandbox_reset_concurrency_test.exs`, `sse_stream_test.exs` (`fake_server/1`), and `conversation_call_timeout_test.exs` (setup) — didn't chase these down to a concrete failure mode, but they're built the same way and worth a look in a follow-up rather than blocking this PR.

**Point 5:** agree, no `flake` issue needed — this is same-day cleanup of your own change from #1981, not an inherited flake, per CLAUDE.md's flaky-tests guidance.

Approving. The `lifecycle_actions_test.exs` finding is worth a fast-follow, separate from this PR.

Issue #2057, opened 2026-09-13
Decide: what BROKER_ALLOW_UNENFORCED means, and whether a brokered deployment can run self-hosted runners


jhgaylor:
The third switch on the non-broker path, after the ADR 0019 §9 ratchet (retired, see the 2026-09-12 amendment) and the `BROKER_LISTEN_PORT` off switch (#2056). This one decides whether a sandbox may hold brokered placeholders on a provider that cannot enforce `allow: [broker]`.

## The state today

`Provisioning.check_broker_support/4` refuses a brokered conversation on a provider whose `capabilities/0` omits `:network_policy`, unless `BROKER_ALLOW_UNENFORCED=true`:

| Provider | `:network_policy` |
|---|---|
| Sprites, E2B, Daytona | yes |
| `Managoat.Runner.Adapter` | **no**, deliberately: "the machine is the user's and so is its network" |

So since `BROKER_TENANTS=*` (2026-09-04) no conversation can launch on a self-hosted runner on the hosted deployment. That was an accepted cost at the time because only two accounts ran runners. It is now the permanent state of ADR 0022 on any deployment that brokers, and `BROKER_ALLOW_UNENFORCED` is documented as development-only.

`apply_broker_floor/2` has a matching arm: on a provider without `:network_policy` it publishes `network`/`done` with `enforced: false` and applies nothing. So an unenforced brokered sandbox holds placeholders, a proxy address and a trusted CA, with nothing that stops a process from a direct connection that avoids the proxy. Credential safety holds — a placeholder is worthless anywhere — but containment does not.

## Why it needs deciding rather than drifting

Three of these can each be defended, and they are mutually exclusive:

- **Runners are a first-class provider (ADR 0022).** Then `BROKER_ALLOW_UNENFORCED` is not a development escape hatch, it is the supported runner configuration, and the docs and the ADR both say the wrong thing today. It also needs an answer for what containment means on a machine the tenant owns.
- **Containment is not optional (ADR 0019 §2, "the network floor is not optional").** Then a brokered deployment cannot run runners, ADR 0022 is superseded for brokered deployments, and `SANDBOX_RUNNERS_ENABLED` should say so at boot rather than letting a launch fail per conversation.
- **The runner adapter learns a policy operation.** `Managoat.Runner` would need a way to hold egress on a machine Fountain does not own. That is a library decision in `managoat/managoat_runner`, not a Fountain one, and it is the only option that keeps both ADRs true.

## Blockers

[...]

Issue #2056, opened 2026-09-13
Decide: remove the non-broker path, so every sandbox reaches the internet through the broker


jhgaylor:
Follow-on from the ADR 0019 §9 ratchet retirement (see the 2026-09-12 amendment). That change made brokerage deployment-wide: `BROKER_LISTEN_PORT` is the whole switch, and every tenant of a deployment that sets it is brokered. This issue is about removing the switch itself, so a sandbox cannot reach the internet except through the broker.

## What "remove the off switch" means

`Fountain.Broker.configured?/0` is false when `BROKER_LISTEN_PORT` is blank, and then:

- `Egress.split_brokered/2`, `split_inference/3`, `bindings/1` and `add_connection_secrets/4` are identity functions, so every credential enters the sandbox in the clear;
- `Egress.apply_policy/4` takes its `false` arm, which is `Provisioning.apply_network_policy/3` (the environment's own `networking_type`) rather than `apply_broker_floor/2`;
- Connections and secret bindings have no surface at all.

Removing it deletes the `false` arm everywhere, plus `Provisioning.apply_network_policy/3` and `check_network_policy_support/3`, which have exactly one production caller each and both sit on that arm.

## What it buys

Closes the key-handling half of #1709 by construction: on an unbrokered self-host the platform inference key goes into the sandbox env in plaintext, and there is no way to configure that away today.

## Blockers, each of which is real work or a decision

1. **#1671 — the broker CA install fails on ~18% of setups.** Today that degrades: the sandbox holds placeholders and a proxy address, and the failure surfaces later as a TLS error. With no unbrokered path there is nothing to fall back to, so an 18% silent-failure rate on trust setup becomes an outright provisioning failure class. Fix #1671 first. (Part of it has landed — `Egress.install_ca/3` now emits `[:broker, :ca_install]` and only publishes `broker`/`done` on success — but the ratio alert and the root cause are open.)
2. **The runner decision, tracked separately.** `managoat_runner` advertises no `:network_policy`, so `check_broker_support/4` refuses every brokered conversation on a self-hosted runner. Mandatory brokering makes ADR 0022 unreachable rather than merely awkward. See #2057.
3. **#1638 pulls the other way.** Switchyard needs a persistent, explicit direct-egress opt-in, and it names `Egress.apply_policy/4` and `apply_broker_floor/2` as the code to change. Decide which of the two wins before building either; a sanctioned direct path and a mandatory broker are not both true.
[...]

PR #2058, opened 2026-09-13
refactor: retire the per-tenant broker ratchet (ADR 0019 §9)

jhgaylor:
Brokerage becomes a property of the deployment alone. `BROKER_LISTEN_PORT` is the whole switch, and every tenant of a deployment that sets it is brokered. `Fountain.Broker.enabled_for?/1` and the `:broker_tenants` configuration are gone; `Fountain.Broker.configured?/0` is the gate.

## Why now

The ADR 0019 §9 ratchet existed to widen the hosted deployment one id at a time while classification was proven, and it reached `*` on 2026-09-04 (home-cloud#163). From that day the per-tenant arm has been unreachable in production and exercised only by the tests written for the ratchet itself.

What it left behind is sixteen call sites of the shape `if brokered?(user_id) do X else identity end`, across `Egress`, `Connections`, `SecretBindings`, `ConversationServer`, `SpriteEnv`, `PlatformInference` and five web modules. Every one of those `else` arms is a plaintext-credential path that nothing exercised and nobody could reach. Deleting them is the point; the tidier `Egress` is a side effect.

## What changes

- `Broker.enabled_for?/1` and `config :fountain, :broker_tenants` are removed.
- Six functions lose a `user_id` they only forwarded, rather than keeping an argument the answer no longer depends on: `Egress.brokered?/0`, `Connections.manageable_for?/0`, `Egress.split_brokered/2`, `split_inference/3`, `release/1`, `reattach_policy/3`, and `Lifecycle.destroy/4`.
- `/admin/broker` drops its "Brokered tenants" tile, and `Native.Insights` drops the `tenants` field it fed.
- `BrokerTestHelpers` gains `enable_broker/0` and `disable_broker/0` in place of the per-tenant forms.
- Docs, `.env.example`, `.env.compose.example`, the marketing self-host table and the primitives diagram (both copies, plus the README alt text) stop describing brokerage as per-account.

## `BROKER_TENANTS` keeps exactly one job, and it is not rollout

A **list of ids is now a boot error.** Booting on one would broker the tenants the list deliberately excluded, which is a security-relevant widening no operator asked for, so boot refuses and names the removal — the same shape as the retired `BROKER_URL`.

**`*` stays accepted**, because it carries the assertion of #1686. That guard fails the boot in the dangerous direction: a deployment that means to broker and loses its listener (the `envFrom` drift of #1495) does not half-broker anyone — brokerage turns itself off for everyone, every sandbox gets plaintext GitHub, inference and connection credentials, and the console and the connections keep working, so nothing downstream can tell it from a deployment that meant to broker nobody. Removing the ratchet removes the tenant list that used to carry that intent, and `*` is the last place an operator can state it. It goes when brokerage becomes unconditional.

**No manifest change is needed.** The hosted deployment already sets `BROKER_TENANTS=*`, which stays valid and keeps its guard. That is deliberate: unlike the #1553 flip, this one has no ordering requirement against home-cloud.

## What this does *not* do

Two non-broker paths remain. The ADR amendment names both rather than describing them as decided:

- **The deployment off switch** — `BROKER_LISTEN_PORT` blank still means every credential enters the sandbox in the clear. #2056 carries that decision and names its blockers: #1671, #1638, the 49-of-55 conversation test files that run unbrokered today, and the self-host boot requirements.
- **The unenforced arm** — a provider without `:network_policy` still cannot host a brokered conversation unless `BROKER_ALLOW_UNENFORCED` says so, which is why no conversation runs on a self-hosted runner (ADR 0022) on a brokered deployment. #2057 carries that decision and the ADR 0022 / ADR 0036 amendments it implies.

## Size

75 files, and roughly as many deletions as insertions. It is wide but shallow, and it is one change: a gate cannot be half-removed, and the arity changes have to land with the branch they follow from.

## Verification

- Full suite from the umbrella root: **5812 tests, 0 failures** (2 skipped), plus 144 in `fountain_buzz` and 33 in `fountain_support`. Run against a per-worktree database so it did not contend with other checkouts.
- Both new boot guards **verified by reverting**: stubbing the `case` arms to `:ok` turns exactly the two new `runtime_config_test` cases red.
- `mix credo --strict` clean; `mix format --check-formatted` clean at the root and in `apps/fountain`.
- `docs_test.exs` green, including the four primitives-diagram guards after the SVG file and README alt text were kept in step with the inline copy.
- `okf validate decisions` → `valid: true`. The index needed no regeneration (no status, title or `stale_after` change).
- OpenAPI spec builds and `scripts/sdk-contract/build.sh --check` → ok, 174 operations / 202 schemas. The changed OpenAPI descriptions are not part of the wire contract.
- `docs-style.py` reports the same 3 pre-existing findings as `origin/main` (verified by stashing). `vale` errors and warnings unchanged at 11 / 135; suggestions down by 2.

https://claude.ai/code/session_01LqjRFWWEcwF8Dat6RjWZD8


jhgaylor:
Rebased onto `6e632dde` (the PR was `CONFLICTING` against the three commits main gained, so zero CI ran on the first push). Three conflicts, all the same shape — main added a parameter or prose while this commit changed the broker call — resolved by keeping main's side and swapping `enabled_for?(user_id)` for `configured?()`: `SpriteEnv.select_inference/4` keeps its new `secrets` argument, `PlatformInference.gate/3` keeps `has_own?/3`, and `InferenceCredentials`' ADR 0053 docstring is kept with only the `Fountain.Broker.configured?/0` reference updated.

Full suite re-run on the rebased tree: **5837 tests, 0 failures** (2 skipped), plus 144 in `fountain_buzz` and 33 in `fountain_support`. `mix precommit` also passed end to end on the pre-rebase tree (dialyzer, sobelow, prod release assemble, suite).

Also regenerated `sdk/typescript/src/generated/openapi.ts`, which the "SDK types match the spec" job caught. The diff is three `@description` lines and nothing else — the same "only for accounts the broker is on for" → "only on a deployment that runs the broker" rewording as the docs. `sdk/contract/contract.json` is unchanged, because descriptions are not part of the wire contract, which is why `scripts/sdk-contract/build.sh --check` passed locally while the generated types were stale.

Labelled `sdk-no-release`: the only published change is JSDoc text in generated types, with no type or behaviour change, so it should ride the next real SDK release rather than move `latest` on its own.

BinaryBourbon (approved):
Adversarial review — runtime/deployment scope. Independent Codex agent review submitted through BinaryBourbon. APPROVE at 958b2b7ef629358813fe937f3c01b441b72862e2; the PR remained open at this exact head immediately before submission. No blocking finding in this scope, and the independent review scopes were reconciled before posting.

Inspected runtime parsing and legacy guards, Broker.backend/configured?, listener supervision and telemetry, health/preflight behavior, deployment examples, changelog and ADR 0019. Explicit tenant lists refuse boot instead of silently widening; a wildcard still refuses a missing listener. Invalid ports and missing proxy URLs fail before startup. The application, readiness and provisioning preflight use the same configured native-backend gate; losing a running listener does not change configured? to false and select plaintext fallback.

Independent validation from an exact-head source archive on Elixir 1.19.2 / OTP 28.3: all 22 RuntimeConfigTest cases passed through Config.Reader. Disabling only the wildcard/list guards in the isolated runtime produced exactly two failures (both required exceptions absent); the exact source was restored. A 16-case parent/head matrix covered unset, blank, separators, wildcard, padded wildcard, explicit IDs and mixed/duplicate wildcard values, with and without a listener. A standalone probe compiled the exact Broker, Native and Health source and exercised a real ephemeral listener: brokerage off was ready; configured-but-down failed readiness/preflight; listener startup passed; stopping it failed both again.

The matrix explicitly confirms the intended migration: a configured listener with blank/unset BROKER_TENANTS changes from brokering nobody to brokering everyone. That behavior is declared by the new deployment-wide switch and ADR/changelog, not overlooked tenant-specific authorization. The self-hosted-runner enforcement limitation and deployment off switch remain the named decisions #2057 / #2056.

Limits: my runtime tests did not start the full Fountain application or database, and the listener probe used warm dependency beams. I did not independently execute a detached release, hosted ingress/rollout, the full umbrella suite or all precommit stages. This approval does not claim those checks or infer correctness solely from CI.

BinaryBourbon (approved):
Adversarial review (independent Codex agent), lifecycle and credential scope: no blocking findings.

Reviewed Broker/Egress gate and arity changes, ConversationServer provisioning and reattachment, Lifecycle destruction, SpriteEnv selection, PlatformInference, InferenceCredentials, and SecretBindings. Removed user_id parameters only selected the retired gate; tenant IDs remain on credential queries, DEK loads, native session creation/refresh, and per-turn secret rereads. Fresh and existing conversations split credentials before building sandbox env, reattachment applies broker policy before credential writes, and failed-attempt cleanup still revokes the specific tenant/conversation/token session.

Independent validation: compiled the exact parent and head Broker/Egress/SpriteEnv sources against read-only warm dependencies. All 48 combinations of secrets, inference credentials, and explicit bindings across broker-off and prior :all modes produced identical sandbox placeholders, broker-held values, and tenant-binding precedence. Nil/deleted-agent selection also matched. A stale empty tenant setting could not disable the new deployment gate or put the catalog credential back into sandbox env.

Validation limits: the comparison exercised pure credential paths, not a live proxy or sandbox provider. I inspected the retained integration assertions for plaintext exclusion, wake refusal, refresh, and cleanup, but did not independently rerun the database-backed lifecycle suites. Other review scopes were reconciled before posting. No merge or queue action taken.

BinaryBourbon (approved):
Adversarial review — independent Codex agent review of API/UI, access checks, documentation and public contracts at this exact commit. No blocking findings.

Checked that deployment-wide manageability changes feature availability while record reads and mutations remain tenant-scoped. The separate Connections rollout flag still gates creation, retargeting and re-enabling; revocation and removal stay available. Gmail retains conversation ownership and agent/connection checks. Admin insights retain authorization and session rechecks. No executable callers of the removed broker gate or old arities remain in the source scan; generated TypeScript changes are comments only.

Independent validation in a private checkout/database: 130 tests passed across 13 API, LiveView, insights and documentation files at seed 2058. Two additional reviewer-only tests passed direct foreign-ID API requests and forged LiveView revoke/delete/toggle events, confirming the other tenant's records remained unchanged. OpenAPI export and SDK contract check passed: 174 operations, 202 schemas.

Limits: targeted validation, not a new full-suite run or live OAuth-provider/network-egress exercise. Other independent review scopes were reconciled before this approval. Verified the PR remained open at the reviewed head before submission.

PR #2061, opened 2026-09-13
fix(oauth): explain loopback origins accept any port (#1897)

### LINKED ISSUE #1897: fountain oauth-client under-reports a loopback origin: CORS admits any port, the CLI prints one
Follow-up from the review of `stack:1125` (#1882, #1883, #1884). Not a blocker on any of those; filing so it does not get lost once the stack merges.

## The mismatch

An OAuth client's `origins` field is `Fountain.OAuth.Client.origins_of/1`, which returns the concrete `scheme://host[:port]` of each redirect URI. For a **loopback** redirect that under-reports what CORS actually admits: `origin_keys` stores a loopback origin **port-less** on purpose (RFC 8252 §7.3), so a client registered against `http://localhost:5173` has `http://localhost` on *any* port admitted by `FountainWeb.Plugs.Cors`.

So a developer reading `origins: http://localhost:5173` concludes that moving Vite off 5173 breaks their API calls. It does not. That is the exact half-working failure the `origin_key` column exists to prevent, reported backwards.

## Where it is already fixed, and where it is not

| Surface | Says "any port"? |
|---|---|
| `OAuthClientJSON` schema description (`apps/fountain/lib/fountain_web/schemas.ex`) | yes, since #1882 — "A loopback origin is admitted on any port, not only the port shown here" |
| Generated TypeScript types (`sdk/typescript/src/generated/openapi.ts`) | yes, carried from the schema |
| Console card (`apps/fountain/lib/fountain_web/live/oauth_clients_live/index.ex`) | yes, since #1883 — `origins_line/1` appends "(and a loopback origin on any port)" |
| **`fountain oauth-client` output** | **no** |

`printOAuthClient` (`cli/internal/cmd/oauthclient.go:169`) prints:

```go
fmt.Printf("origins:       %s\n", strings.Join(a.Origins, ", "))
```

A schema description does not reach a terminal, and the CLI is the surface a developer setting up a local app is most likely to be looking at when they need this fact.

## Fix

Mirror `origins_line/1` in Go: when any of the client's `redirect_uris` parses to a loopback host, append the caveat.

```
origins:       http://localhost:5173 (loopback: any port)
```

The host set is `localhost`, `127.0.0.1`, `::1` — `Fountain.OAuth.Client.loopback?/1` is the source of truth and the Go side should carry a comment pointing at it, the way `oauthClient` already points at `oauth_client_json.ex` as the struct it mirrors.

Affects `fountain oauth-client list`, `create`, `update` and `get`, since all four go through `printOAuthClient`. The `--json` path needs no change — a JSON consumer has the schema description.

## Wording note

#1883 phrases it "(and a loopback origin on any port)", which reads slightly oddly for a client whose *only* redirect is loopback: "and" implies a second origin rather than the same one relaxed. Whatever wording this lands on, the two surfaces should match, so it is worth fixing both in the same PR.

jhgaylor:
OAuth apps registered with a loopback redirect previously showed only the registered port in CLI output, even though CORS accepts that origin on any port. The shared create/update printer now appends `(loopback: any port)` for `localhost`, `127.0.0.1`, and `::1`; console cards use the same wording.

The JSON response and origin authorization are unchanged. The current CLI list command only shows redirect URIs and has no origin field; no `get` command exists.

Fixes #1897.

Validation:

- `go test -count=1 ./...` and `go vet ./...` in `cli/` passed with Go 1.26.5.
- The printer regression covers all three hosts, uppercase localhost, IPv6 with and without a port, mixed origins, non-loopback hosts, malformed and relative URIs, and an empty list.
- `mix precommit apps/fountain/test/fountain_web/live/oauth_clients_live_test.exs` passed under Elixir 1.19.2 / OTP 28.3: compile, unused-dependency checks, formatting, strict Credo, Sobelow, dev Dialyzer, production release assembly, and 5 focused tests with 0 failures. The full Elixir test suite was not rerun on this branch.
- Independent source review found no issues with host detection, JSON compatibility, or serial stdout capture in the tests.

### DIFF
--- a/apps/fountain/lib/fountain_web/live/oauth_clients_live/index.ex
+++ b/apps/fountain/lib/fountain_web/live/oauth_clients_live/index.ex
@@ -122,7 +122,7 @@ defmodule FountainWeb.OAuthClientsLive.Index do
     origins = Enum.join(Client.origins_of(redirect_uris), ", ")
 
     if Enum.any?(redirect_uris, &loopback_uri?/1),
-      do: origins <> " (and a loopback origin on any port)",
+      do: origins <> " (loopback: any port)",
       else: origins
   end
 
--- a/apps/fountain/test/fountain_web/live/oauth_clients_live_test.exs
+++ b/apps/fountain/test/fountain_web/live/oauth_clients_live_test.exs
@@ -36,7 +36,7 @@ defmodule FountainWeb.OAuthClientsLiveTest do
     assert html =~ "In development"
     # A loopback redirect was registered, so the card must not imply that
     # :5173 is the only origin CORS will admit.
-    assert html =~ "any port"
+    assert html =~ "(loopback: any port)"
   end
 
   test "shows the validation error rather than saving", %{conn: conn} do
--- a/cli/internal/cmd/oauthclient.go
+++ b/cli/internal/cmd/oauthclient.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"net/url"
 	"strings"
 
 	"github.com/managoat/fountain/cli/api"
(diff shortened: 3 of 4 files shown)


BinaryBourbon (approved):
Adversarial agent review

Reviewed the complete diff at `60bf6c8747e3a74c8d03dad271c7e006dde1b026` against `6e632dde8e4f93b1ae62e6ac5f642cf51d19e936` and issue #1897, using the approved-base contributor and review guidance.

Covered all four changed paths: `cli/internal/cmd/oauthclient.go`, its new printer test, `apps/fountain/lib/fountain_web/live/oauth_clients_live/index.ex`, and its LiveView test. Traced create/update and JSON branches, actual list command behavior, server URI validation/origin-key generation, CORS lookup, and tenant-scoped console loading. The three loopback hosts match the server contract; the annotation is emitted once, non-loopback origins stay unchanged, and JSON/authorization behavior is preserved. The issue's list/get premise does not match the actual command tree.

Independently ran `go test -count=1 ./internal/cmd -run '^TestPrintOAuthClientOrigins$'` on an exact-head CLI archive: passed. Inspected the exact-commit precommit record and underlying LiveView log: 5 tests, 0 failures. I did not rerun the Elixir/full CLI suites. No blocking correctness, isolation, compatibility or simplicity findings.

PR #2062, opened 2026-09-13
test(conversations): pass nil for absent permission params (#1901)

### LINKED ISSUE #1901: pending_test.exs passes an integer where Pending.ask/7 expects request params
Noted on #1855 during the `stack:1635` review, non-blocking, and not fixed there.

## What

`Pending.ask/7` gained a seventh parameter in #1855. It is the request's own `session/request_permission` params, from which the per-request `_meta.fountain.timeout` is read:

```elixir
@spec ask(t(), String.t(), Conversations.Turn.t() | nil, term(), String.t(), list(), map() | nil)
def ask(%__MODULE__{} = pending, conversation_id, turn, request_id, tool, options, params) do
  detached_timeout_ms = DetachedRequest.timeout_ms(params, effective_ask_timeout_seconds(conversation_id))
```

Six call sites in `apps/fountain/test/fountain/conversations/pending_test.exs` pass `60_000` there — lines 100, 120, 136, 159, 185 and 215 — which looks like a leftover from an earlier signature where the argument was a timeout in milliseconds.

## Effect

Harmless at runtime and invisible to every gate:

- `DetachedRequest.timeout_seconds/1` matches on `%{"_meta" => %{"fountain" => %{"timeout" => value}}}`, so an integer falls through its catch-all to `nil`.
- Those six tests therefore exercise the **policy/global fallback** branch, not a per-request timeout, and none of them asserts on `detached_timeout_ms` — they match `pending_permission` partially and check the timer, so the assertions are all still meaningful.
- Nothing catches the type mismatch: dialyzer runs `MIX_ENV=dev` and never sees test files.

## Fix

Pass `nil` at those six sites, which is what the tests mean. If a test ever wants to cover the per-request path at this level, it should pass a real params map (`%{"_meta" => %{"fountain" => %{"timeout" => 60}}}`) — the end-to-end coverage for that currently lives in `conversation_server_acp_test.exs`, which drives a real peer.

jhgaylor:
Fixes #1901.

Six `Pending.ask/7` tests pass `60_000` as the request-params argument, although the function expects a params map or `nil`. Replace those values with `nil` to state that the tests use the policy/global timeout fallback. Existing permission and timer assertions remain intact.

Validation on this exact commit (Elixir 1.19.2 / OTP 28.3, disposable test database): `mix precommit apps/fountain/test/fountain/conversations/pending_test.exs` passed. Compilation, unused-dependency checks, formatting, Credo, Sobelow, Dialyzer and production-release assembly passed; the scoped Pending suite completed with 15 tests, zero failures. `git diff --check` also passed.

### DIFF
--- a/apps/fountain/test/fountain/conversations/pending_test.exs
+++ b/apps/fountain/test/fountain/conversations/pending_test.exs
@@ -97,7 +97,7 @@ defmodule Fountain.Conversations.PendingTest do
       turn: turn,
       pending: pending
     } do
-      {turn, pending} = Pending.ask(pending, conv.id, turn, 7, "bash", ["yes", "no"], 60_000)
+      {turn, pending} = Pending.ask(pending, conv.id, turn, 7, "bash", ["yes", "no"], nil)
 
       assert %{"request_id" => 7, "tool" => "bash", "options" => ["yes", "no"], "asked_at" => _} =
                turn.pending_permission
@@ -117,7 +117,7 @@ defmodule Fountain.Conversations.PendingTest do
       conv: conv,
       pending: pending
     } do
-      {nil, pending} = Pending.ask(pending, conv.id, nil, 7, "bash", [], 60_000)
+      {nil, pending} = Pending.ask(pending, conv.id, nil, 7, "bash", [], nil)
       assert [{"started", _}] = stages(conv.id, "request")
       Process.cancel_timer(pending.permission_timer)
     end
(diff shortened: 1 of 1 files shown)


BinaryBourbon (approved):
Adversarial agent review

APPROVE at 3aff13eed14c651cbf9a62ecd102f52b94063bc0 (reviewed base 6e632dde8e4f93b1ae62e6ac5f642cf51d19e936). No blocking findings.

Covered the entire `apps/fountain/test/fountain/conversations/pending_test.exs` diff against #1901, including all six replacements and surrounding assertions. Traced `Pending.ask/7`, its server adapter, `DetachedRequest.timeout_seconds/1` and `timeout_ms/2`, plus the ACP per-request/policy timeout tests. Both the former integer and `nil` select the same fallback; `nil` correctly expresses absent params under the declared map-or-nil contract. Persistence, peer denial/answer, audit, and timer assertions remain intact. Test ownership and SQL Sandbox isolation are unchanged; no production behavior or API surface changes.

Evidence: independently inspected source and approved-base review guidance; inspected the exact-head validation record and log reporting 15 Pending tests, 0 failures, and current successful CI required/Detect secrets checks. I did not rerun the BEAM suite. The minimal test-only correction meets the linked issue without weakening coverage.

This is the explicitly requested independent agent review, not a Review Loop service verdict.

PR #2065, opened 2026-09-13
test(secrets): isolate global expiry notice configuration (#1932)

### LINKED ISSUE #1932: Flake: secret_forms_live expiry label races the sweeper test's global :secret_expiry_notice_days
`apps/fountain/test/fountain_web/live/secret_forms_live_test.exs:188` fails intermittently, most recently in a **merge queue** build where it ejected an unrelated PR (#1872, the sandbox capacity queue, which touches nothing in vaults or secrets).

## The failing assertion

```
test "add_secret records an optional expiry date and renders its status"
test/fountain_web/live/secret_forms_live_test.exs:188

    assert html =~ "expires in 3d"
    right: "expires in 3d"
    left:  "<div class=..."   # the rendered page, without that label
```

## How often, out of what

One merge-queue build of six for #1872 so far. Run: https://github.com/managoat/fountain/actions/runs/34595416749 (partition 2, 2026-09-11 11:47 UTC). The two earlier #1872 queue failures were a different, real problem (an interaction with #1840) and are fixed; this one is not that.

## The mechanism

`FountainWeb.VaultsLive.Form.expiry_status/1` renders the label only inside its notice window:

```elixir
days = Date.diff(DateTime.to_date(expires_at), Date.utc_today())
cond do
  ...
  days <= SecretExpirySweeper.notice_days() -> {:expiring, "expires in #{days}d"}
  true -> {:ok, "expires #{DateTime.to_date(expires_at)}"}
end
```

`SecretExpirySweeper.notice_days/0` is `Application.get_env(:fountain, :secret_expiry_notice_days, 7)` — global.

`apps/fountain/test/fountain/workers/secret_expiry_sweeper_test.exs` is `use Fountain.DataCase, async: true` and at line 120 does:

```elixir
Application.put_env(:fountain, :secret_expiry_notice_days, 0)
```

While that write is held, `3 <= 0` is false for every concurrent test, so the label falls to the `true ->` branch and the assertion fails. Nothing is wrong with either test in isolation; they only have to overlap. Note the date arithmetic is `Date.diff` on dates, so this is **not** a time-of-day boundary — it is the global write.

## Why the guardrail did not catch it

`apps/fountain/test/fountain/async_global_config_guardrail_test.exs` exists for exactly this class and **allowlists this file**:

```elixir
# :secret_expiry_notice_days — read only by the sweeper under test.
"apps/fountain/test/fountain/workers/secret_expiry_sweeper_test.exs",
```

That justification is no longer true: the key is also read by `expiry_status/1` on every vault form render, and `secret_forms_live_test.exs` asserts on the result. The guardrail's own docstring gives the fix: *"The allowlist is not a blessing... Moving one into an `async: false` module is always correct, and is what to do the moment one of them is suspected in a flake."*

## Suggested fix

Move the `put_env` portion of `secret_expiry_sweeper_test.exs` into a sibling `async: false` module and delete its allowlist entry, the way `quotas_test.exs` and `user_emails_test.exs` were handled. Worth re-reading the other allowlist entries with the same question — each says "read only by X", and this one shows how that goes stale when a reader is added later.

Filed rather than fixed in place because it surfaced mid-burndown of the approved PR stacks, per CLAUDE.md.

https://claude.ai/code/session_016nLzDCqKwgQfuW1V5gVik4

jhgaylor:
Fixes #1932.

The sweeper test temporarily sets the global secret-expiry notice window to zero. When it overlaps the vault-form tests, their three-day expiry label disappears. Move that configuration test into a synchronous sibling module and remove the stale guardrail allowlist entry. Restore the previous configuration after the test, including whether the key was absent.

The other sweeper tests remain asynchronous, and the sweep-disabled and vault-label assertions remain intact.

Validation on this exact commit (Elixir 1.19.2 / OTP 28.3, disposable test database):

```sh
mix precommit \
  apps/fountain/test/fountain/workers/secret_expiry_sweeper_test.exs \
  apps/fountain/test/fountain/workers/secret_expiry_sweeper_config_test.exs \
  apps/fountain/test/fountain_web/live/secret_forms_live_test.exs \
  apps/fountain/test/fountain/async_global_config_guardrail_test.exs
```

Passed compilation, unused-dependency checks, formatting, Credo, Sobelow, Dialyzer and production-release assembly. The combined scoped suite completed with 21 tests, zero failures. `git diff --check` also passed.

### DIFF
--- a/apps/fountain/test/fountain/async_global_config_guardrail_test.exs
+++ b/apps/fountain/test/fountain/async_global_config_guardrail_test.exs
@@ -54,8 +54,6 @@ defmodule Fountain.AsyncGlobalConfigGuardrailTest do
     "apps/fountain/test/fountain/webhooks_test.exs",
     # :retention_days — read only by the pruner under test.
     "apps/fountain/test/fountain/workers/retention_pruner_test.exs",
-    # :secret_expiry_notice_days — read only by the sweeper under test.
-    "apps/fountain/test/fountain/workers/secret_expiry_sweeper_test.exs",
     # :callback_key_ttl_seconds — read when a callback key is minted.
     "apps/fountain/test/fountain_web/api_key_scope_test.exs",
     # :buzz_cli_bin — read only by the buzz paths under test.
--- a/apps/fountain/test/fountain/workers/secret_expiry_sweeper_config_test.exs
+++ b/apps/fountain/test/fountain/workers/secret_expiry_sweeper_config_test.exs
@@ -1,0 +1,35 @@
+defmodule Fountain.Workers.SecretExpirySweeperConfigTest do
+  # The vault form also reads this global notice window. Run after async
+  # tests so disabling the sweep cannot change their expiry labels (#1932).
+  use Fountain.DataCase, async: false
+
+  import Swoosh.TestAssertions
+
+  alias Fountain.Workers.SecretExpirySweeper
+
+  test "notice_days 0 disables the sweep" do
+    previous = Application.fetch_env(:fountain, :secret_expiry_notice_days)
+
+    on_exit(fn ->
+      case previous do
+        {:ok, value} -> Application.put_env(:fountain, :secret_expiry_notice_days, value)
+        :error -> Application.delete_env(:fountain, :secret_expiry_notice_days)
+      end
+    end)
+
+    Application.put_env(:fountain, :secret_expiry_notice_days, 0)
+
+    user = insert_verified_user()
+    vault = insert_vault(user_id: user.id)
+
+    expires_at =
+      DateTime.utc_now()
+      |> DateTime.add(1, :day)
+      |> DateTime.truncate(:second)
+
+    insert_vault_secret(vault, key: "GH_TOKEN", expires_at: expires_at)
+
+    assert :ok = perform_job(SecretExpirySweeper, %{})
+    assert_no_email_sent()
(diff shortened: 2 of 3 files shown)


BinaryBourbon (approved):
Adversarial agent review

Reviewed the complete diff at `007c717c18f082e06c3af9964114aba91bfedc9b` against `6e632dde8e4f93b1ae62e6ac5f642cf51d19e936` and issue #1932, using the approved-base contributor and review guidance.

Covered all three changed paths: `apps/fountain/test/fountain/workers/secret_expiry_sweeper_test.exs`, new `secret_expiry_sweeper_config_test.exs`, and `apps/fountain/test/fountain/async_global_config_guardrail_test.exs`. Traced the sweeper, vault expiry-label reader/tests, and DataCase sandbox/cleanup. The global write now runs in a synchronous module, restoration preserves both absent and existing configuration, and deleting the obsolete exception restores guardrail coverage. The no-email assertion and async sweeper coverage remain intact; production code, tenant ownership and API behavior do not change.

Independently inspected the exact-commit validation record and underlying scoped precommit log: 21 tests, 0 failures across both sweeper modules, vault forms and the config guardrail. Existing form-ID warnings are visible in that log. I did not rerun BEAM tests or claim this single run proves every interleaving. The source change removes the identified overlap; no blocking findings.

PR #2066, opened 2026-09-13
docs: complete the CI job reference table (#2026)

### LINKED ISSUE #2026: docs: CLAUDE.md's CI table omits six jobs, including the one that gates CI's own decision logic
`CLAUDE.md`'s CI pipeline table documents 12 of the 19 jobs in
`.github/workflows/ci.yml`. #2020 rebuilt the table after #1413 split
`sdk-clients`, but deliberately kept scope to the rows that had gone *wrong*
rather than adding the ones that were never there. These omissions predate it.

Measured on `97bcf1a6`:

| Job id | Check name a red build shows |
|---|---|
| `workflow-checks` | `CI policy and alert tests` |
| `already-tested` | `Skip the re-run when a PR already tested this exact tree` |
| `core-distribution` | `Core distribution boots without the extensions` |
| `compose-fresh-clone` | `Compose fresh-clone check` |
| `compose-pinned-image-boot` | `Compose boots the pinned image` |
| `docs` | `Docs` |

(`coverage` also has no row of its own, but the `test` row describes it, so it
is not really undocumented.)

## Why this matters, and which one to do first

The table's stated purpose is in the sentence above it: "the job name in a red
build tells you which toolchain to go and look at." That promise fails for any
job not in the table.

**`workflow-checks` is the one worth adding first.** It runs
`python3 -m unittest discover -s scripts/ci`, which is the suite that gates the
CI decision logic itself — `gate.py`, `sdk_changes.py`, `require-checks.py` and
their tests. Someone who breaks one of those sees a failing check called
`CI policy and alert tests`, a name that appears nowhere in the table, and the
job also runs `scripts/conflict-markers.py` and the alert-fixture evaluation
(#1927, #1939), so the name does not obviously point at `scripts/ci/` either.

`already-tested` is the second most useful, because a reader who sees it
skipped needs to know that is normal on a `pull_request` and that the probe
only runs on `push` and `merge_group`.

## Suggested shape

One row each, in the same style as the existing rows: what it runs, plus the
one fact a reader cannot get from the workflow. For `workflow-checks` that is
"the suite that gates CI's own decision logic"; for `already-tested`, which
events it runs on and what a skip means.

Related: #1413 (the split), #2020 (the table rebuild that set this scope aside).

jhgaylor:
Fixes #2026.

The CI table omitted six jobs, so a failing check could not be mapped back to its tools. Add rows for `workflow-checks`, `already-tested`, `core-distribution`, both Compose checks and `docs`, with their exact GitHub check names. Explain the CI-policy tests, normal probe skips by event, core/bundled release checks, the unpublished release-pin exception, and docs-only selection including reused merge groups.

Validation:
- Compared each row against the current workflow steps, conditions and `scripts/ci/already-tested.sh`.
- A local completeness check confirms all 19 workflow jobs appear in the table: 18 rows, with `coverage` described in the existing `test` row. All six added check names match the workflow exactly.
- `git diff --check` passes.
- `/tmp/fountain-qw-env mix precommit apps/fountain/test/fountain/docs_test.exs` passes on `9b150d13041f570a9e588bb35b3c347fa810c485`: all compile, unused-dependency, format, Credo, Sobelow, Dialyzer and release-assembly stages completed; the scoped suite reports 29 tests, 0 failures. The wrapper selects Elixir 1.19.2 / OTP 28.3 and the isolated test database. This is scoped verification, not a full-suite run on this head.

Documentation only: no workflow, gate script, test or CI policy changes.


BinaryBourbon (approved):
Adversarial agent review

Reviewed the complete `CLAUDE.md` diff at `9b150d13041f570a9e588bb35b3c347fa810c485` against `6e632dde8e4f93b1ae62e6ac5f642cf51d19e936` and issue #2026. Read instructions from the approved base; treated the proposed instructions as review evidence.

Compared all six added rows with `.github/workflows/ci.yml`, `scripts/ci/already-tested.sh`, `scripts/ci/gate.py`, the Compose check script and release workflow. Independently checked that all 19 jobs are represented (coverage is in the test row) and all six added check names match exactly. Event selection, artifact failure fallback, core/bundled release checks, Compose pin exception and docs job selection are accurately described. This changes documentation only and introduces no runtime, tenant-isolation or API changes.

Independently inspected the exact-commit precommit record and underlying docs log: 29 tests, 0 failures. I did not rerun BEAM tests, Docker boots or full CI. No introduced blocking correctness or maintainability findings.

PR #2068, opened 2026-09-13
fix(console): identify and confirm principal key replacement (#2030)

### LINKED ISSUE #2030: Principal key renewal form: no confirmation, and the picker shows bare UUIDs
## Symptom

The "Replace a principal key" form added to `/api-keys` by #1952 has two rough edges next to the controls it sits beside.

## 1. No confirmation on an irreversible action

The revoke button four elements down carries one:

```elixir
phx-click="revoke"
data-confirm="Revoke this key? Any integrations using it will stop working."
```

The renewal form has no `data-confirm`, and it is equally irreversible: it revokes the principal's current credential and mints a replacement, so anything holding the old key stops working the moment the form is submitted. The copy above the select says so in prose ("Its previous key will stop working"), which is the right warning in the wrong place — a select plus a submit button is one misclick.

Suggested: `data-confirm="Replace this principal's key? Its current key will stop working."`

## 2. The select shows bare UUIDs

```elixir
<option :for={id <- @principals} value={id}>{id}</option>
```

`Principals.list_owned/1` selects `o.principal_user_id` and nothing else, so an owner with more than one principal picks between raw UUIDs with no way to tell which machine is which. Widening `list_owned/1` to return the `application_id` (or the claimable's name) alongside the id would make the option label meaningful. That is the part that makes this more than cosmetic: item 1 asks the user to confirm, and right now they have nothing to confirm *against*.

## Scope

Both are console-only and neither is a security issue — ownership is scoped correctly and the forged-id path is covered by a test. Raised as follow-ups during review of #1952, which was approved.

---
Found reviewing #1952 (stack:1688, tracker #1688). The code described here is not on `main` yet; it lands with that stack.

jhgaylor:
The principal-key replacement form now asks for confirmation before replacing the current credential. Its picker shows each principal's application ID alongside its UUID, so owners can identify the key they intend to replace; principals without application metadata retain their UUID label.

A new owner-scoped display query preserves `Principals.list_owned/1`'s UUID return type for account deletion. Credential replacement eligibility is unchanged.

The console has no existing `data-confirm` handler, so a submit-only LiveView hook supplies the prompt. Cancelling or failing to open the dialog blocks submission; interacting with the picker does not prompt.

Fixes #2030.

Validation:

- Extended the existing LiveView renewal test to assert the application label and confirmation, retaining its credential replacement and revocation checks.
- Added a LiveView regression for the exact UUID fallback and exclusion of another owner's principal and application name.
- `node --test deployed/test/principal-confirmation.test.mjs` passed (4 tests). The shipped submit hook blocks cancellation and dialog errors, permits acceptance once, ignores picker events, and removes its listener on destruction.
- `node --test deployed/test/*.test.mjs` passed: 240 tests, 0 failures.
- A local headless Chrome page using the shipped hook passed cancel, accept, dialog-error, picker-click/change and listener-cleanup checks with a bubbling parent submit observer. This checks browser event behavior without a live server.
- `mix precommit apps/fountain/test/fountain_web/live/api_keys_live_test.exs apps/fountain/test/fountain/principals_test.exs` passed under Elixir 1.19.2 / OTP 28.3: compile, unused-dependency checks, formatting, strict Credo, Sobelow, dev Dialyzer, production release assembly, and 64 focused tests with 0 failures. The full Elixir suite was not rerun on this branch; the existing unrelated unused `conn` warning remains in the revoke test.
- `git diff --check` passed. Independent review verified owner scoping, preserved query compatibility, and the final confirmation hook.

### DIFF
--- a/apps/fountain/lib/fountain/principals.ex
+++ b/apps/fountain/lib/fountain/principals.ex
@@ -145,6 +145,19 @@ defmodule Fountain.Principals do
     |> Repo.all()
   end
 
+  @doc "Owned principal IDs and their application names, oldest first."
+  @spec list_owned_details(binary()) :: [%{id: binary(), application_id: String.t() | nil}]
+  def list_owned_details(owner_user_id) when is_binary(owner_user_id) do
+    from(o in Owner,
+      left_join: c in ClaimableUser,
+      on: c.user_id == o.principal_user_id,
+      where: o.owner_user_id == ^owner_user_id,
+      order_by: [asc: o.inserted_at],
+      select: %{id: o.principal_user_id, application_id: c.application_id}
+    )
+    |> Repo.all()
+  end
+
   @doc """
   Whose ledger funds work done as `subject` (ADR 0044 decision 4).
 
--- a/apps/fountain/lib/fountain_web/components/layouts/root.html.heex
+++ b/apps/fountain/lib/fountain_web/components/layouts/root.html.heex
@@ -242,6 +242,22 @@
 
         // ── LiveView hooks ───────────────────────────────────────────────────
         const Hooks = {
+          ConfirmSubmit: {
+            mounted() {
+              this.confirmSubmit = (event) => {
+                let confirmed = false;
+                try { confirmed = window.confirm(this.el.dataset.confirm); } catch (_) {}
+                if (!confirmed) {
+                  event.preventDefault();
+                  event.stopImmediatePropagation();
+                }
+              };
+              // Confirm a submission, not a click on a field inside the form.
+              // Cancellation stops the event before LiveView receives it.
+              this.el.addEventListener("submit", this.confirmSubmit, { capture: true });
+            },
+            destroyed() { this.el.removeEventListener("submit", this.confirmSubmit, { capture: true }); }
+          },
           // Prevents a collapsed <details> section from staying closed when
           // LiveView adds new content to it. Because `open?` is always `true`
           // for turn/recursive sections, LiveView's diff never re-sends the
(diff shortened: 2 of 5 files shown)


BinaryBourbon (approved):
Adversarial agent review

Reviewed head `38f85eb053a720ffbb90bc1e0021eed55ac955a4` and the entire five-file diff against the recorded base `6e632dde8e4f93b1ae62e6ac5f642cf51d19e936`, with issue #2030 and the approved-base correctness, tenant-isolation, API-compatibility and simplicity guidance. No blocking findings.

Coverage: `principals.ex` owner-scoped display join and unchanged UUID-only `list_owned/1` deletion caller; `api_keys_live/index.ex` authenticated mount, escaped application/UUID labels, missing-metadata fallback and existing transactional ownership checks during renewal; `layouts/root.html.heex` hook registration, cancellation/error handling and listener cleanup; both changed test files and their named-principal, foreign-owner, fallback, forged-ID and credential replacement assertions. The pinned LiveView 1.1.30 client listens for submit on window in the bubbling phase, so the form's capture listener stops a cancelled submission before LiveView pushes it.

Validation: independently ran `node --test deployed/test/principal-confirmation.test.mjs` at this head: 4 passed, 0 failed. Inspected the exact-head local validation record and its precommit log, which reaches 64 focused Elixir tests, 0 failures; I did not rerun Elixir or a live-server browser test. Fresh GitHub checks show CI required and Detect secrets successful. This is an independent agent review, not human maintainer approval or a claim that Review Loop service gates ran.

PR #2069, opened 2026-09-13
test(conversations): await cotenant registry visibility (#2053)

### LINKED ISSUE #2053: Flake: lifecycle_actions_test's cotenant test single-reads the Horde registry
Found while reviewing #2052, which fixes the same shape in a sibling file. Not fixed by that PR.

## The test

`apps/fountain/test/fountain/conversations/lifecycle_actions_test.exs:322` — "casts :machine_gone to every other live server" — registers a stand-in server, waits for its own `assert_receive :registered`, then immediately calls `Lifecycle.stop_cotenants/5`.

`stop_cotenants/5` does a **single** `ConversationServer.whereis/1` (`apps/fountain/lib/fountain/conversations/lifecycle.ex:530`). `Fountain.ConversationRegistry` is Horde's CRDT registry (`application.ex:75`), so a registration is not guaranteed visible to `lookup/2` immediately — which is exactly the race #2052 just fixed in `termination_client_test.exs`.

When the lookup lags, `stop_cotenants/5` takes its `nil ->` branch, no cast is sent, and the test's `assert_receive {:cotenant, ...}` times out.

The `:registered` signal the test waits on is not the signal it depends on. That is the same defect as #2052: waiting on one thing and asserting on another.

## Production note, deliberately separate

The `nil` branch is **intentional** — the docstring says "a co-tenant whose server is already gone is not an error here", and the next prompt takes the wake path. So a lagged lookup in production is not a crash; it means one co-tenant misses its `:machine_gone` cast and keeps a stale handle until its next prompt wakes it.

That is a real but bounded degradation, and it is not obviously worth changing: distinguishing "gone" from "not yet visible" would require `ConversationServer.await_registered/2` (which exists, at `conversation_server.ex:67`, built for this exact lag per #1429) on a path whose whole point is to be best-effort. Recording it so the trade-off is a choice rather than an accident.

## Lower confidence, same shape, not chased to a failure

Register → signal → immediate use, with no retry:

- `sandbox_reset_reconciler_test.exs`
- `sandbox_reset_concurrency_test.exs`
- `sse_stream_test.exs` (`fake_server/1`)
- `conversation_call_timeout_test.exs` setup

Worth a sweep rather than four separate issues if someone is in there.

## Frequency

Not observed failing yet — found by pattern-matching against the #2052 failure rather than from a red run. Filing per CLAUDE.md's "record an unexplained failure with what you know" rather than waiting for it to bite someone.

Refs #2052, #1981, #1429.

jhgaylor:
Fixes #2053.

The cotenant test waited for a registration message before calling `stop_cotenants/5`, although Horde may not make that registration visible to lookup immediately. Use the existing bounded `ConversationServer.await_registered/2` helper and assert the exact stand-in PID is visible before exercising the cast. Run the stand-in under the test supervisor so a failed assertion also cleans it up.

The original `:machine_gone` cast assertion remains intact. Production cotenant behavior is unchanged.

Validation on this exact commit (Elixir 1.19.2 / OTP 28.3, disposable test database):

```sh
mix precommit \
  apps/fountain/test/fountain/conversations/lifecycle_actions_test.exs \
  apps/fountain/test/fountain/conversations/termination_client_test.exs
```

Passed compilation, unused-dependency checks, formatting, Credo, Sobelow, Dialyzer and production-release assembly. The scoped lifecycle and termination suites completed with 35 tests, zero failures. `git diff --check` also passed.

### DIFF
--- a/apps/fountain/test/fountain/conversations/lifecycle_actions_test.exs
+++ b/apps/fountain/test/fountain/conversations/lifecycle_actions_test.exs
@@ -14,6 +14,7 @@ defmodule Fountain.Conversations.LifecycleActionsTest do
   use Mimic
 
   alias Fountain.Conversations
+  alias Fountain.Conversations.ConversationServer
   alias Fountain.Conversations.Lifecycle
   alias Managoat.Sandbox.Handle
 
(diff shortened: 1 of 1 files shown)


BinaryBourbon (approved):
Adversarial agent review

Reviewed the complete diff at `da74c0b57978d371164b4a90c452a8ecc983935a` against `6e632dde8e4f93b1ae62e6ac5f642cf51d19e936` and issue #2053, using the approved-base contributor and review guidance.

Covered the sole changed path, `apps/fountain/test/fountain/conversations/lifecycle_actions_test.exs`, and traced `ConversationServer.await_registered/2`, `whereis/1`, `Lifecycle.stop_cotenants/5`, DataCase and the sibling termination probe tests. The test now waits for the actual lookup dependency and pins the exact supervised stand-in PID before invoking the cast. The bounded wait cannot silently accept timeout, supervision cleans up on assertion failure, and the original full `:machine_gone` message assertion remains. The production best-effort missing-server branch, tenant boundaries and API behavior are unchanged.

Independently inspected the exact-commit validation record and underlying scoped precommit log: 35 lifecycle/termination tests, 0 failures. I did not rerun BEAM tests or a full suite; a passing run alone is not the basis for concluding the registration race is addressed. No blocking findings.

PR #2060, opened 2026-09-13
chore: clean stale library configuration and repository guidance (#1717)

jhgaylor:
Refs #1717 (partial cleanup).

Local worktrees appear as untracked files, the library dependency comment assumes every package is still at 0.1, and ADR 0037 still describes the retired Agent Vault coexistence period. Ignore `/.worktrees/`, describe the existing per-library minor pins accurately, and align the ADR status paragraph with the native-only broker already recorded in ADR 0019. Remove the unused `:fountain, :connections_allow_private_hosts` test setting; preserve the live `:managoat_mcp_auth, :allow_private_hosts` setting.

Validation:
- `git check-ignore -v .worktrees/example/README.md` and `git diff --check` pass.
- Evaluated old/new test configuration with `Config.Reader`: the only value removed is the unused Fountain key; the live library key remains `true`.
- `okf validate decisions` passes (0 errors; 6 existing actor-convention warnings in other ADRs). Ran `okf backlinks decisions 0037-component-libraries` before editing and regenerated the decisions index; it is unchanged.
- `/tmp/fountain-qw-env mix precommit apps/fountain/test/fountain/docs_test.exs apps/fountain/test/fountain/connections_test.exs apps/fountain/test/fountain/umbrella_layout_test.exs` passes on `43d23465bb7b7786353d7acd54d97cbffaf36fb8`: all compile, unused-dependency, format, Credo, Sobelow, Dialyzer and release-assembly stages completed; the scoped suite reports 46 tests, 0 failures. The wrapper selects Elixir 1.19.2 / OTP 28.3 and the isolated test database. This is scoped verification, not a full-suite run on this head.

This does not close #1717. The Hermes validation-error bullet is addressed separately in #2070. Its remaining bullets need separate assessment: broker private-upstream configuration documentation, Swift `onboarding_state`, checkpoint event registration, recovery audit evidence, the #1626/#1624 issue-reference mismatch, and optional library patch-release consumption. No dependency versions or runtime broker policy change here.

The ADR edit corrects an implementation-status fact; it does not change the recorded architecture. `decisions/**` is covered by the existing Review Loop human-review policy.

Issue #1717, opened 2026-09-07
Papercuts from the 2026-09-07 week review


jhgaylor:
Small things, one PR:

- `.worktrees/` is untracked and not in `.gitignore`.
- `apps/fountain/mix.exs:53` says "The pins are patch-level (~> 0.1.0)"; acp is `~> 0.2.2`, broker `~> 0.11.0`, runner/sandbox `~> 0.2.0`, runtimes `~> 0.3.1`.
- `decisions/0037-component-libraries.md` status block still says the broker "runs beside the Agent Vault client, selected by BROKER_LISTEN_PORT, until the deployment flips"; #1488/#1494 flipped and deleted it.
- `config/test.exs:148` sets `config :fountain, :connections_allow_private_hosts`, which nothing reads (the library key two lines below is live). `:broker_allow_private_upstreams` is read at `broker/native.ex:65` and set by nothing, so document that it is always false or drop the read.
- `sdk/swift/Sources/FountainKit/Models/Account.swift:9,21` still decodes `onboarding_state` after #1447.
- `integrations/hermes/fountain/client.py:245` prefers `error` over `errors`, so after #1448 every 422 shows `validation_failed` and hides the field map.
- `home_checkpoint.ex:109` publishes a `checkpoint` stage that is in neither `webhooks/events.ex`'s registry nor `events_test`'s `@sources`, so the guard cannot see it.
- #1655's recovery harness records no audit row for a production restart; the journal file is the only trace.
- #1626's body says "keep #1624 open for a deployed rerun", but #1624 is closed.
- Unconsumed test-only patch releases on hex: managoat_docs/mcp_auth/oauth/substitution 0.1.1, managoat_runner 0.2.1. No source changes; bump when convenient.

---
Found in the 2026-09-07 review of the week on `main` (cb93a7ea).

PR #2070, opened 2026-09-13
fix(hermes): show field validation details (#1717)

jhgaylor:
Refs #1717 (Hermes validation-error slice).

When Fountain returns `422` with both `error: validation_failed` and a field-level `errors` map, Hermes currently shows only the generic code. Prefer the field details so the caller sees which input was rejected. Responses without field errors retain the existing error/message fallback; status and the original parsed response body remain available on `FountainError`.

Validation:
- The new request-path regression fails before the fix because the prompt field message is missing.
- All 20 Hermes plugin tests pass after the fix.
- `git diff --check` passes.
- `/tmp/fountain-qw-env mix precommit` passes on `dab792361a894c704942e6bb9a2546376794695c`: compile, unused-dependency, format, Credo, Dialyzer, Sobelow and release-assembly checks completed; the full suite reports core 5,828 tests plus 6 doctests, Buzz 144 tests, and Support 33 tests, all with zero failures (core: 2 skipped, 7 excluded). The wrapper selects Elixir 1.19.2 / OTP 28.3 and an isolated test database.

This closes only the Hermes bullet. The repository cleanup is in #2060; its remaining SDK, runtime, audit and dependency bullets remain open.

### DIFF
--- a/integrations/hermes/fountain/client.py
+++ b/integrations/hermes/fountain/client.py
@@ -242,7 +242,7 @@ def _maybe_json(raw: bytes) -> Any:
 def _describe_http_error(status: int, parsed: Any, url: str) -> str:
     detail = ""
     if isinstance(parsed, dict):
-        err = parsed.get("error") or parsed.get("errors") or parsed.get("message")
+        err = parsed.get("errors") or parsed.get("error") or parsed.get("message")
         if err:
             detail = err if isinstance(err, str) else json.dumps(err)
     elif isinstance(parsed, str) and parsed.strip():
--- a/integrations/hermes/tests/test_plugin.py
+++ b/integrations/hermes/tests/test_plugin.py
@@ -5,11 +5,13 @@
 
 from __future__ import annotations
 
+import io
 import json
 import os
 import sys
 import tempfile
 import unittest
+import urllib.error
 from pathlib import Path
 from unittest import mock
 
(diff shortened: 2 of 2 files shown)


BinaryBourbon (approved):
Adversarial agent review

APPROVE at dab792361a894c704942e6bb9a2546376794695c (reviewed base 2fd1ee2785488ecaf26d3b3827bd01bfec5b22c2). No blocking findings.

Covered the entire `integrations/hermes/fountain/client.py` and `integrations/hermes/tests/test_plugin.py` diff against the Hermes bullet in #1717. Traced HTTP error parsing through `FountainError` and the tool JSON wrapper, and checked the server's `ChangesetJSON` response shape. Preferring nonempty `errors` surfaces actual field details while retaining status/body, coded-error/message fallback, and the existing exception type. No request auth, tenant lookup, or server contract changes; field details are serialized as text, not executed.

Independent validation: all 20 Hermes tests pass. The initial sandboxed run could not bind the loopback fake server; the same suite passed with loopback binding permitted. Ran six additional formatter fallback probes, and verified the new request-path regression fails against the base formatter specifically because the prompt field detail is missing. Current CI required and Detect secrets checks also succeed. No BEAM suite rerun was needed for this Python-only change. The correction is minimal and the regression exercises the public request path and preserved error metadata.

This is the explicitly requested independent agent review, not a Review Loop service verdict. It addresses only the Hermes slice of #1717.

PR #2071, opened 2026-09-13
test(conversations): await timeout fixture registry visibility (#2005)

### LINKED ISSUE #2005: Flake: ConversationCallTimeoutTest's public-function test raced whereis/1 in a full-suite run
Recording an unexplained local failure rather than dismissing it because a rerun passed (CLAUDE.md, *Flaky tests*).

**Test:** `apps/fountain/test/fountain_web/controllers/conversation_call_timeout_test.exs:72` — "the public functions return error tuples rather than exiting", in `FountainWeb.ConversationCallTimeoutTest`.

**How often, out of what:** one failure in one of two local full-suite runs (`mix test` from `apps/fountain`, macOS workstation, isolated DB, Elixir 1.19.2/OTP 28). The same file run alone passed 5/5, and a second full suite on a fixed seed (`--seed 424242`) passed 5,417 tests / 0 failures. No CI run URL: this was local, while merging main into `feat/durable-turn-deadlines` (#1744). The tree under test does not touch `conversation_server.ex`.

**What I do not have:** the failing `left:`/`right:`. The run was piped through a `grep -E` filter that kept only the counts line and the numbered-failure header, so the assertion body was discarded, and the failure has not reproduced since. That is the main reason this is a record rather than a diagnosis.

**Mechanism worth checking first.** The test asserts on what the call *returns*, but waits only on the registration message:

```elixir
stuck = spawn(fn ->
  Horde.Registry.register(Fountain.ConversationRegistry, conv.id, nil)
  send(test_pid, :registered)
  Process.sleep(:infinity)
end)
assert_receive :registered, 2_000
```

`assert_receive :registered` proves `register/3` was *called*, not that `ConversationServer.whereis/1` can see it. Every one of the three assertions goes through `whereis(conv_id)`, and on `nil` `send_prompt/4` takes the `Conversations.wake_conversation/2` branch instead of `call_server/2` — so a lookup that has not settled returns `{:error, :not_running}` / `{:error, :gone}` / `{:error, :not_found}` where the test expects `{:error, :provisioning}`. Same shape as the settled-lookup races in #1544 and #801: wait on the signal you assert on, not on a proxy for it.

The other suspect is the modules global `Application.put_env(:fountain, :conversation_call_timeout_ms, 150)`. 150 ms is a thin margin for three sequential `GenServer.call`s on a loaded workstation, and the env var is process-global — see #1214 for the same hazard.

**Next step for whoever picks this up:** re-run the full suite unfiltered until it reproduces, capture the seed and the assertion, then replay that seed on `main` to separate seed luck from a real regression.

https://claude.ai/code/session_01R9reKpmXUUJf4eUMULogie

jhgaylor:
The timeout fixture sent `:registered` without verifying the registry lookup used by the public calls. If that lookup missed the stand-in, the tests could exercise the missing-server path instead of the timeout path.

Wait for `ConversationServer.await_registered/2` to return the exact supervised stand-in PID before running the existing assertions. The fixture now restores the timeout setting's previous presence and value. The 150 ms call timeout and all five regression tests remain unchanged.

Validation at `4b8bdddf`: scoped `mix precommit apps/fountain/test/fountain_web/controllers/conversation_call_timeout_test.exs` passed every stage with 5 tests and 0 failures. Ten additional runs of the same five tests all passed (50 test executions). `git diff --check` passed. The original full-suite failure discarded its assertion output and has not been reproduced here; this repairs the fixture's readiness and cleanup contract without claiming a confirmed diagnosis of that lost failure.

Fixes #2005

### DIFF
--- a/apps/fountain/test/fountain_web/controllers/conversation_call_timeout_test.exs
+++ b/apps/fountain/test/fountain_web/controllers/conversation_call_timeout_test.exs
@@ -14,8 +14,15 @@ defmodule FountainWeb.ConversationCallTimeoutTest do
   # return value, not an exit.
 
   setup %{conn: conn} do
+    previous_timeout = Application.fetch_env(:fountain, :conversation_call_timeout_ms)
     Application.put_env(:fountain, :conversation_call_timeout_ms, 150)
-    on_exit(fn -> Application.delete_env(:fountain, :conversation_call_timeout_ms) end)
+
+    on_exit(fn ->
+      case previous_timeout do
+        {:ok, value} -> Application.put_env(:fountain, :conversation_call_timeout_ms, value)
+        :error -> Application.delete_env(:fountain, :conversation_call_timeout_ms)
+      end
+    end)
 
     user = insert_verified_user()
     {_key_record, raw_key} = insert_api_key(user)
@@ -23,17 +30,17 @@ defmodule FountainWeb.ConversationCallTimeoutTest do
 
     # The shape of a server whose mailbox is blocked by provisioning:
     # registered under the conversation id, never answering calls.
-    test_pid = self()
-
     stuck =
-      spawn(fn ->
-        Horde.Registry.register(Fountain.ConversationRegistry, conv.id, nil)
-        send(test_pid, :registered)
-        Process.sleep(:infinity)
-      end)
+      start_supervised!(
+        {Task,
+         fn ->
+           {:ok, _} = Horde.Registry.register(Fountain.ConversationRegistry, conv.id, nil)
+           Process.sleep(:infinity)
+         end}
+      )
 
-    assert_receive :registered, 2_000
-    on_exit(fn -> Process.exit(stuck, :kill) end)
+    # Wait for the lookup used by the public calls, not a registration message.
+    assert {:ok, ^stuck} = ConversationServer.await_registered(conv.id, 2_000)
 
     {:ok, conn: authed_with_key(conn, raw_key), user: user, conv: conv}
   end


BinaryBourbon (approved):
Agent-assisted independent adversarial review of 4b8bdddf1aff217667ca200ee43ec269df83447a against base 67e22cb4e0963d039ac0a4268ab95afe21743b39. I did not implement this PR. No blocking findings.

Reviewed the complete conversation_call_timeout_test.exs diff and traced await_registered/2 through the Horde lookup used by the public functions, the call timeout handler, controller responses and deletion path. Exact-PID readiness ensures the supervised sleeping stand-in is the resolved server before assertions run. Task supervision also owns teardown if setup fails. The previous timeout setting's presence and value are restored; async: false, the 150 ms timeout, all five regression tests, 503/error-tuple checks and persisted deletion assertion remain intact. Production code is unchanged.

Evidence inspected: exact-head scoped precommit log reaches 5 tests, 0 failures with exit 0; repeat-run log contains the stated ten additional successful runs. git diff --check passes. Current GitHub CI required, all six test partitions, coverage and secret scanning are successful.

Limitations: I independently inspected source and recorded validation rather than rerunning BEAM tests or the full Review Loop verification recipe. The original full-suite failure lost its assertion output and has not been reproduced; this review supports the concrete fixture readiness/cleanup repair, not a proven diagnosis of that historical failure. This is the user-requested formal code review, not a claim that the four-reviewer Review Loop service ran, and it does not merge or queue the PR.

PR #2072, opened 2026-09-13
test(team): wait for stream subscription readiness (#2067)

### LINKED ISSUE #2067: Flake: TeamStreamTest publishes after a 300 ms sleep instead of waiting for the SSE subscription
`FountainWeb.TeamStreamTest`, `"events from every teammate arrive on one connection, labelled"`, `apps/fountain/test/fountain_web/controllers/team_stream_test.exs:90`.

Found while re-reviewing #2000. Not caused by that branch: it touches neither the team stream nor anything the test reads.

## How often, out of what

Four full-suite runs of `apps/fountain` on tree `159ceb19` (two commits on `859f60c7`), own PostgreSQL database, nothing else running:

| run | seed | result |
|---|---|---|
| 1 | 31337 | **this test failed**, 5,824 tests, 1 failure |
| 2 | 424242 | 0 failures |
| 3 | 31337 (replay) | 0 failures |
| — | 31337 on `859f60c7` | 0 failures |

Plus 3/3 passes of the file alone at seed `31337` on both trees. So it is not seed-dependent: replaying the failing seed on the same tree passed. Local macOS arm64, pinned Erlang 28.3 / Elixir 1.19.2, `max_cases: 20`.

**I did not retain the assertion output** — that run was filtered to test names and totals, and the replay that would have captured it passed. Recording it anyway rather than dropping it, per CLAUDE.md.

## Mechanism

The test waits on a sleep and asserts on something else:

```elixir
task = stream_async(key)
Process.sleep(300)

publish(ada_conv, %{kind: "output", stream: "acp", data: "from-ada"})
publish(linus_conv, %{kind: "output", stream: "acp", data: "from-linus"})
publish(other_conv, %{kind: "output", stream: "acp", data: "not-on-team"})

conn = Task.await(task, 5_000)
...
assert body =~ "from-ada"
```

The 300 ms sleep stands in for "the SSE handler has subscribed to every teammate's topic". Under full-suite load with 20 concurrent cases that is not guaranteed, and a `publish` that lands before the subscription is simply lost — `assert body =~ "from-ada"` then fails on a body that is otherwise well-formed. The `[ada_payload] = Regex.run(...)` on line 118 would fail as a `MatchError` on the same cause.

Same family as #1934 (`FountainWeb.SseStreamTest`, heartbeat under load), different test and different assertion, so filing separately.

## What would fix it

Synchronise on the subscription rather than sleeping — have the stream announce readiness (it already writes `": connected\n\n"` first, which the test on line 125 asserts) and wait for that byte before publishing.

jhgaylor:
Fixes #2067.

The labelled-event test published after a fixed 300 ms sleep, which did not prove that the team stream had subscribed. Observe the real test adapter's `: connected` chunk and hold the stream there until the test has published. Apply the same handshake to the roster-change test's `team` frame, emitted after subscribing to the new conversation. This also prevents fixture work from exhausting the stream's 800 ms idle window.

The adapter still writes the real chunks, and both tests open without `Last-Event-ID`, so their assertions exercise live PubSub delivery. Payload labels and exclusion of the non-team conversation remain covered. Handshakes and task cleanup are bounded; production code is unchanged.

The issue's original assertion output was not retained, so this fixes the identified synchronization gap without claiming to reproduce that historical failure.

Validation on `04e230e8f4d47daefb36395613b0436882a337e4`:

- `mix precommit apps/fountain/test/fountain_web/controllers/team_stream_test.exs apps/fountain/test/fountain_web/controllers/sse_stream_test.exs` passed all stages; the combined suites ran 30 tests with 0 failures. Test scope was these two files, not the full suite.
- Repeated both changed tests together 10 times at seed `31337`: 20 test executions, 0 failures.
- `git diff --check` passed.

### DIFF
--- a/apps/fountain/test/fountain_web/controllers/team_stream_test.exs
+++ b/apps/fountain/test/fountain_web/controllers/team_stream_test.exs
@@ -7,6 +7,7 @@ defmodule FountainWeb.TeamStreamTest do
   """
 
   use FountainWeb.ConnCase, async: false
+  use Mimic
 
   import Phoenix.ConnTest, only: [build_conn: 0]
 
@@ -87,6 +88,44 @@ defmodule FountainWeb.TeamStreamTest do
     })
   end
 
+  defp stream_ready(raw_key) do
+    parent = self()
+
+    Mimic.stub(Plug.Adapters.Test.Conn, :chunk, fn state, body ->
+      result = Mimic.call_original(Plug.Adapters.Test.Conn, :chunk, [state, body])
+
+      if body in [": connected\n\n", "event: team\ndata: {\"reason\":\"changed\"}\n\n"] do
+        # Both frames are written after follow_team has subscribed. Hold the
+        # stream here so fixture work cannot consume its 800 ms idle window.
+        ref = make_ref()
+        send(parent, {:stream_ready, self(), body, ref})
+        assert_receive {:continue_stream, ^ref}, 5_000
+      end
+
+      result
+    end)
+
+    # No Last-Event-ID: the queued broadcasts must reach the live loop, since
+    # replay is skipped for this connection.
+    task = stream_async(raw_key)
+
+    on_exit(fn ->
+      ref = Process.monitor(task.pid)
+      Process.exit(task.pid, :kill)
+      assert_receive {:DOWN, ^ref, :process, _, _}, 5_000
+    end)
+
+    {task, await_ready(task, ": connected\n\n")}
+  end
+
+  defp await_ready(task, body) do
+    pid = task.pid
+    assert_receive {:stream_ready, ^pid, ^body, ref}, 5_000
+    ref
+  end
+
+  defp continue_stream(task, ref), do: send(task.pid, {:continue_stream, ref})
+
   test "events from every teammate arrive on one connection, labelled", %{
     user: user,
     raw_key: key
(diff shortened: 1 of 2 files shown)


BinaryBourbon (approved):
Agent-assisted independent adversarial review, submitted as BinaryBourbon at the user's explicit request. I did not implement this PR. Reviewed the complete two-file diff at 04e230e8f4d47daefb36395613b0436882a337e4 against approved base 7756da61847b9c68da5f466528224ad73fd7061c, the base review-loop policy/instructions, issue #2067, and live checks/review history. No blocking findings.

Scope: apps/fountain/test/fountain_web/controllers/team_stream_test.exs and apps/fountain/test/test_helper.exs, traced through TeamController, ConnCase/DataCase, and the locked Mimic/Plug adapter implementations. Both intercepted frames occur after the relevant real PubSub subscriptions. The stub calls the original chunk implementation and runs in the request Task through Mimic's caller inheritance; it does not replace event delivery. Both changed tests omit Last-Event-ID, so replay cannot satisfy their live-event assertions. Label checks and exclusion of the unrelated conversation remain intact. Ref-scoped handshakes and Task awaits are bounded; cleanup kills and monitors the Task before configuration restoration, Mimic cleanup, and SQL-owner teardown. The synchronous module retains the existing application-environment isolation.

Evidence independently inspected: exact-head GitHub CI run 34738595086 passed, including all six test partitions, CI required, and the separate secret scan. The recorded exact-head precommit command reached its analysis/release stages and the combined team/SSE suites' 30 tests, 0 failures. The corrected repeat-fixed log selects the actual changed tests at lines 129 and 190 and contains 11 successful two-test summaries (22 executions, 0 failures); I did not use the earlier incorrectly selected repeat log. git diff --check passed in my review checkout.

Limitations: I inspected the coordinator's runtime logs rather than launching another BEAM instance. This is an individual source review, not a claim that I executed the Review Loop service's complete independent verification recipe. The original issue did not retain its failing assertion, so the evidence establishes removal of the demonstrated synchronization gap, not reproduction of that historical failure.

PR #2073, opened 2026-09-13
test(execution): synchronize expiry and timeout consequences (#2024)

### LINKED ISSUE #2024: Flake: ExecutionTransportTest asserts async progress against a 200 ms wall-clock deadline

jhgaylor:
Fixes #2024.

Execution transport and deadline worker tests could expire a 200 ms deadline or kill setup/scan work with 100–300 ms task timers before reaching the provider operation they intended to test. Give setup a safety deadline, explicitly expire the journal after identity is buffered, and deliver the timer's kill signal only after blocked spawn/write/termination calls announce themselves. Preserve expiry while all eight termination slots remain occupied, the original deadline and attempt, write fencing, and no replay after restart.

Separate orphan tests retain real production timers and prove tasks expire after abrupt transport/coordinator death, with enough headroom to observe the blocked call before killing its owner. Production behavior is unchanged.

Validation at `0e40be0bf374956ee377523815350bbf4aa3d6ea`:

- Scoped `mix precommit apps/fountain/test/fountain/conversations/execution_transport_test.exs apps/fountain/test/fountain/conversations/execution_deadline_worker_test.exs`: all stages passed, 30 tests, 0 failures.
- Controlled probe: temporarily insert a 300 ms delay immediately after registration in the delayed-identity test. Approved base `89103b9e` fails with `{:error, :deadline_expired}`; this candidate passes the same test. Temporary edits were restored. This demonstrates the setup-delay mechanism; it does not reproduce the original full-suite load failures.
- Independent agent source review found no concrete defects in synchronization, real timer coverage, expiry semantics, or no-replay assertions.

### DIFF
--- a/apps/fountain/test/fountain/conversations/execution_deadline_worker_test.exs
+++ b/apps/fountain/test/fountain/conversations/execution_deadline_worker_test.exs
@@ -37,7 +37,7 @@ defmodule Fountain.Conversations.ExecutionDeadlineWorkerTest do
     })
   end
 
-  defp await_state(id, state, attempts \\ 100)
+  defp await_state(id, state, attempts \\ 500)
   defp await_state(_id, state, 0), do: flunk("journal never reached #{state}")
 
   defp await_state(id, state, attempts) do
@@ -55,14 +55,18 @@ defmodule Fountain.Conversations.ExecutionDeadlineWorkerTest do
     owner = self()
     rows = for _ <- 1..8, do: execution(c.user, true)
 
-    worker(:blocked_pool, fn attempt ->
-      send(owner, {:termination_started, attempt.id, self()})
-      receive do: (:release -> :ok)
-    end)
+    worker(
+      :blocked_pool,
+      fn attempt ->
+        send(owner, {:termination_started, attempt.id, self()})
+        receive do: (:release -> :ok)
+      end,
+      job_timeout_ms: 60_000
+    )
 
     processes =
       for _ <- rows do
-        assert_receive {:termination_started, _id, pid}, 2_000
+        assert_receive {:termination_started, _id, pid}, 5_000
         pid
       end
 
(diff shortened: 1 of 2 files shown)


BinaryBourbon (approved):
Agent-assisted independent adversarial review of 0e40be0bf374956ee377523815350bbf4aa3d6ea against base 7756da61847b9c68da5f466528224ad73fd7061c. I did not implement this PR. No blocking findings.

Reviewed the entire two-file diff: execution_transport_test.exs and execution_deadline_worker_test.exs, plus their production transport/worker and journal interactions. Checked buffered identity before expiry, durable submission before injected task failure, fencing and no replay, original deadline/attempt preservation, eight occupied termination slots, and teardown. The manually delivered untrappable exits exercise failure consequences after provider entry; separate orphan tests still exercise production task timers after abrupt coordinator/transport death. Production code is unchanged.

Evidence inspected: exact-head scoped precommit log reaches 30 tests, 0 failures and exit 0; controlled 300 ms post-registration delay fails the old fixture with deadline_expired and passes this candidate. The probe demonstrates the setup-delay mechanism, not reproduction of historical full-suite load failures. Current GitHub CI required, all six test partitions, coverage and secret scanning are successful.

Limitations: source review and existing validation logs were independently inspected; I did not rerun BEAM tests or the full independent Review Loop verification recipe. This is the user-requested formal code review, not a claim that the four-reviewer Review Loop service ran, and it does not merge or queue the PR.

PR #2075, opened 2026-09-13
test(swift): synchronize timeout conformance with observed output

### LINKED ISSUE #1723: Flake: Swift conformance run-timeout-raises-and-keeps-partial-text loses the partial text on a slow runner
## What failed

`Swift SDK (ubuntu-24.04)` / `swift test --filter ConformanceTests`, scenario `run-timeout-raises-and-keeps-partial-text`, on PR #1722 (no Swift or conformance changes in the PR; the only SDK-facing change is one optional query parameter in the contract).

```
↳ conformance FAILED for swift-kit / run-timeout-raises-and-keeps-partial-text
    A run that outlives its timeout raises, and the error carries what the agent said so far
    error
        expected {
          "kind" : "timeout",
          "partial_text" : "Half an answer"
        }
        got {
          "code" : null,
          "field_errors" : { },
          "kind" : "timeout",
          "partial_text" : "",
          "retry_after" : null,
          "retryable" : false,
          "status" : null
        }
```

First attempt: https://github.com/BinaryBourbon/fountain/actions/runs/34085084699/job/101627491478 (2026-09-07 05:00Z). Rerun of the same commit with no code change: passed. The ubuntu job passed on the eight preceding PR runs the same morning, and on `main` at 2a0f16c9 (02:23Z).

## How often

One failure in nine ubuntu runs on 2026-09-07 (first attempt of PR #1722), passing on rerun.

## Mechanism (probable)

The scenario scripts a stream that sends "Half an answer" and then goes quiet, with a 300 ms client timeout. The kind was `timeout` and the partial text was empty, so the timeout fired before the first chunk had been read: the assertion is on the text having arrived, but the only thing it waits on is the 300 ms clock. On a slow runner the fake server's first frame can lose that race. The other three SDKs did not fail on the same run. #1424 (closed 09-03) was a different divergence in the same scenario (an extra GET after the timeout), not this.

A fix would either have the scenario wait for the first chunk to be observed before starting the timeout clock, or give the timeout more headroom relative to the scripted delay.

jhgaylor:
The timeout conformance scenario could lose partial text when its 300 ms deadline expired before the executor consumed the first output frame. Start this scenario's deadline after the consumer observes text and hold the quiet fixture tail open until cancellation. A per-client internal sleeper seam preserves the public client's normal deadline behavior.

The harness now gives initial output a separate 10-second guard. Missing output throws an explicit harness error, and cancellation of the test cancels and settles its Run before any wait for the result can hang. The event loop and result wait share the cancellation handler; deferred cleanup also handles early errors. This addresses the adversarial review's demonstrated suite stall.

Fixes #1723.

Validation:

- All 86 Swift tests passed with warnings as errors, including nine conformance tests and the new missing-output and cancellation regressions.
- Delaying the first frame by 600 ms still retains partial text. The public-client silent-stream test still times out without waiting for output.
- The independent reviewer repeated the original missing-output reproduction: it now exits with the expected harness failure after 10.678 seconds, instead of surviving the 60-second test limit and requiring an external kill at 80 seconds.
- A delayed-consumer control also completed correctly when the guard expired before the stream request started.
- Strict recursive Swift format lint and the release build passed.
- Scoped `mix precommit apps/fountain/test/fountain/extension_composition_test.exs` passed every gate and all 24 tests in an isolated build/database.

The additional correction changes only the conformance harness. No SDK release or public API change is required.

### DIFF
--- a/sdk/swift/Tests/FountainKitTests/ConformanceTests.swift
+++ b/sdk/swift/Tests/FountainKitTests/ConformanceTests.swift
@@ -70,6 +70,44 @@ struct ConformanceTests {
   }
 
   @Test(.timeLimit(.minutes(1)))
+  func missingOutputFailsWithoutHanging() async throws {
+    let scenario = try timeoutScenarioWithoutOutput()
+    let transport = ScriptedTransport(exchanges: scenario.http, holdQuietTail: true)
+    let observations = Observations()
+    let deadline = ObservedOutputDeadline(outputWaitNanoseconds: 100_000_000)
+
+    do {
+      try await drive(scenario, transport: transport, into: observations, deadline: deadline)
+      Issue.record("missing output must fail the harness")
+    } catch let error as Harness {
+      #expect(error.description == "timed out waiting for the scenario's initial text output")
+    }
+    #expect(transport.unmatched.isEmpty)
+  }
+
+  @Test(.timeLimit(.minutes(1)))
+  func cancellingWhileWaitingForOutputSettlesTheRun() async throws {
+    let scenario = try timeoutScenarioWithoutOutput()
+    let transport = ScriptedTransport(exchanges: scenario.http, holdQuietTail: true)
+    let observations = Observations()
+    // The guard cannot rescue this test before its own one-minute limit.
+    let deadline = ObservedOutputDeadline(outputWaitNanoseconds: 120_000_000_000)
+    let task = Task {
+      try await drive(scenario, transport: transport, into: observations, deadline: deadline)
+    }
+    defer { task.cancel() }
+    await deadline.waitUntilWaitingForOutput()
+    task.cancel()
+    do {
+      try await task.value
+      Issue.record("cancelling the harness must cancel its Run")
+    } catch is CancellationError {
+      // Reaching here proves the unstructured Run no longer strands drive.
+    }
+    #expect(transport.unmatched.isEmpty)
+  }
+
+  @Test(.timeLimit(.minutes(1)))
   func publicClientDeadlineDoesNotWaitForOutput() async throws {
     var scenario = try ConformanceSuite.scenario(
       named: "run-timeout-raises-and-keeps-partial-text")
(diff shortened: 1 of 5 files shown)


sdk/swift/Tests/FountainKitTests/ConformanceTests.swift
```diff
@@ -62,14 +111,45 @@ struct ConformanceTests {
...
+    (output, continuation) = AsyncStream.makeStream()
+  }
+
+  func outputObserved() {
+    continuation.yield(())
+    continuation.finish()
+  }
+
+  func sleep(nanoseconds: UInt64) async throws {
+    for await _ in output { break }
```
BinaryBourbon: [P2] Bound the no-output path before waiting on the deadline gate

If no .text event is emitted, this wait never finishes and holdQuietTail prevents the stream from closing. The one-minute test trait does not provide the claimed completion bound: its cancellation ends run.events, after which drive waits at run.value() on the still-running unstructured follower. In an isolated negative control replacing only the output SSE frame with a comment, the test reported its 60-second limit but was still alive at 80 seconds and needed an external kill. Removing only the sleeper override made the same control fail with empty partial text in 0.334 seconds. Give the gate a bounded failure/cancellation path and explicitly clean up the Run so a text-decoding regression fails the test rather than stalling the suite.

BinaryBourbon (changes requested):
Adversarial review (independent Codex agent): changes requested for a demonstrated harness regression.

The exact-head conformance baseline passed on macOS / Swift 6.1.2. In an isolated archive I replaced only the timeout fixture's output SSE frame with an SSE comment, preserving the turn-start frame, 300 ms timeout, and quiet tail. The delayed-first-frame test reported its 60-second time limit, but the test process still had not exited at 80 seconds and required an external kill. Removing only the new sleeper override made that identical missing-output control fail normally with empty partial text in 0.334 seconds.

The new output gate and indefinitely held quiet tail make a missing/undecodable text event an unbounded suite stall. The Run cancellation limitation predates this PR, but the new harness exposes it: cancellation stops the run.events iteration, then drive awaits run.value(), whose unstructured follower is still waiting for output and whose continuation does not observe cancellation. The one-minute trait therefore reports a failure without bounding completion.

Please bound the wait for initial output and ensure harness cancellation settles/cancels the Run before awaiting its value; add a missing-output negative control that actually terminates. Production default deadline behavior and the successful observed-output synchronization otherwise look sound. No production files or shared worktrees were modified during this review.

sdk/swift/Tests/FountainKitTests/ConformanceTests.swift
jhgaylor: Addressed in 80d132ad19400f2127f265bc735559819dc560d2. The initial-output gate now has a separate 10-second guard, and the harness wraps both event consumption and run.value() in a cancellation handler that cancels the Run, with deferred cleanup for early exits. Added regressions for missing output through the held-open fixture and cancellation while waiting for output. All 86 Swift tests passed. The independent reproduction now exits with an explicit Harness error after 10.678 seconds instead of hanging beyond the test limit.

BinaryBourbon (approved):
Adversarial review follow-up (independent Codex agent): the demonstrated blocker is resolved; no remaining blocking findings at 80d132ad19400f2127f265bc735559819dc560d2.

Verified the pushed correction matches the independently examined harness. Initial output now has a separate 10-second failure guard, cancellation ends the competing waiter, and the cancellation handler plus deferred Run cleanup covers both event consumption and run.value(). The observed-output scenario still uses the real 300 ms deadline afterward.

Independent validation in an isolated copy: all 9 conformance tests passed on the final source, including missing-output and active-drive cancellation regressions. Repeating the original dropped-output-frame reproduction now produces the expected harness failure and exits after 10.678 seconds, inside a 25-second external bound; the previous revision remained alive at 80 seconds. A separate 200 ms delayed-consumer control passed in 0.110 seconds after removing the unnecessary request-count assertion, confirming the guard can safely fire before GET begins.

The earlier changes-requested review remains as the evidence trail. No merge or queue action taken.

PR #2076, opened 2026-09-13
Fix provision deadline tests with explicit watchdog synchronization

### LINKED ISSUE #2040: Flake: provision-deadline tests race their 300ms deadline
## What

`apps/fountain/test/fountain/conversations/conversation_server_provision_deadline_test.exs` fails non-deterministically, on an unmodified `origin/main` checkout.

Three identical consecutive runs of the same tree:

```
4 tests, 0 failures
4 tests, 1 failure
4 tests, 2 failures
```

## The assertions

Both tests in the file can fail, and which one does moves between runs at a fixed seed:

```
1) test a hung provision is killed at the deadline and its rows are failed
   Assertion failed, no matching message after 5000ms
2) test a provision that completes in time is left alone
```

Two runs at `--seed 0`, one on unmodified `origin/main` (`a4822662`) and one with an unrelated refactor applied, failed *different* tests of the pair — so the seed does not determine it.

## Mechanism (partial)

`setup` writes a 300 ms deadline into global application env:

```elixir
Application.put_env(:fountain, :provision_deadline_ms, 300)
on_exit(fn -> Application.delete_env(:fountain, :provision_deadline_ms) end)
```

The watchdog is a bare `spawn/1` that sleeps to that deadline and then reads the sandbox row, so the test is asserting on a 300 ms wall-clock race against provisioning work on a loaded machine. Two candidate causes, not yet separated:

- plain timing: 300 ms is short enough that a busy runner makes "hung" and "completed in time" swap outcomes;
- global app env written by tests that may run concurrently with others reading the same key — the shape of #1214.

The second is testable by checking whether the failures only occur in a full-suite or parallel-worktree run. Both of my repeat runs were the single file alone, which argues for the first.

## What I have not established

Whether this reaches CI. I found it locally while verifying an unrelated change and have not gone through first-attempt CI failures for it. Recording it now rather than dismissing it because a rerun passed.

Environment: darwin 24.5.0, Elixir 1.19.2/OTP 28, worktree sharing `fountain_test` with other checkouts.

jhgaylor:
The provision deadline tests started a 300 ms clock before the server reached the state being asserted. Deliver the watchdog's real deadline message only after provisioning has acknowledged its stalled or completed state, and wait for the watchdog to finish before inspecting the rows. The watchdog now uses a cancellable timer message for its deadline.

A separate zero-delay timer test exercises actual scheduling against already-pending rows. Temporary ExUnit-supervised fixtures ensure a failed assertion cannot leave a hung server or watchdog running after database teardown. The tests retain the terminal-row-before-kill and no-second-sprite contracts.

Fixes #2040.

Validation: `mix test test/fountain/conversations/conversation_server_provision_deadline_test.exs --seed 0 --repeat-until-failure 20` completed 21 consecutive runs (105 tests, 0 failures). `mix precommit apps/fountain/test/fountain/conversations/conversation_server_provision_deadline_test.exs` passed compile, formatting, Credo, Dialyzer, Sobelow, prod release assembly, and the 5-test suite. Validation used an isolated PostgreSQL database and OTP 28.3 / Elixir 1.19.2.


BinaryBourbon (approved):
Adversarial review — independent Codex agent review of this exact commit; no blocking findings.

Checked watchdog monitor/timer ordering, cancellation, pending-versus-ready decisions, terminal row writes before server termination, restart suppression, and ExUnit fixture shutdown before sandbox teardown. The unchanged production ceiling remains real: only the scenario tests inject its message, while the zero-delay case independently exercises actual scheduling.

Independent validation in a private checkout/database: all 5 targeted tests passed at seed 12345. A negative control changing only the scheduled timer message caused the real-timer test to fail (5 tests, 1 failure), while the explicitly synchronized scenarios still passed. Restored the original source afterward.

Limits: targeted tests and source review, not a new full-suite or 30-minute elapsed-time run. Verified the PR head remained unchanged before submission.

PR #2077, opened 2026-09-13
Fix runner LiveView heartbeat ownership and teardown

### LINKED ISSUE #1909: Flake: RunnersLiveTest exits on a FakeDaemon heartbeat outliving its SQL Sandbox owner
`FountainWeb.RunnersLiveTest` "lists runners with live status and forgets one" (`apps/fountain/test/fountain_web/live/runners_live_test.exs:9`) exits with a `DBConnection.OwnershipError`.

## The failure

The test does not fail an assertion — it **exits**, raised out of `:proc_lib`, not out of the test process:

```
** (EXIT from #PID<...>) an exception was raised:
    ** (DBConnection.OwnershipError) ... owner ... is not alive
        (fountain) Fountain.Runners.touch/1
        (managoat_runner) Managoat.Runner.Connection.handle_info/2
        (stdlib) :proc_lib...
```

raised in `Fountain.Runners.touch/1`, called from `Managoat.Runner.Connection.handle_info/2` handling a `:heartbeat`, under `Managoat.Runner.FakeDaemon.Socket`.

## How often, out of what

- **One full-suite run out of one**, observed locally while reviewing #1865 on 2026-09-11 (a verification run of `mix test` from `apps/fountain`, otherwise 4,814 tests green).
- **Passes 6/6 when run in isolation**, so it needs suite load to reproduce — which is why a rerun will look like it cleared.

No CI run URL: this was a local full-suite run, not a CI failure. Recording it per the repo's own rule that an unexplained failure gets written down rather than dismissed because a rerun passed.

## Mechanism

The gap is between what the test waits on and what it asserts on. `on_exit` stops the `FakeDaemon` **from a separate process**, so the daemon's connection process can still be alive and processing after the test process — the SQL Sandbox connection owner — has exited. A `:heartbeat` landing in that window calls `Runners.touch/1`, which tries to check out a connection owned by a process that is already gone.

This is the same family as #1862 (a process outliving its SQL Sandbox owner) but a **different mechanism**: #1862 was a crash-looping `ConversationServer` restarted under the shared Horde supervisor; this is an `on_exit` teardown that does not wait for the process it stops. Fixing #1862 does not fix this.

## Not already fixed

Dated against `git log`, per the repo's guidance for this kind of report:

- `apps/fountain/test/fountain_web/live/runners_live_test.exs` last changed `b55d24c7` (2026-09-01)
- `Fountain.Runners.touch/1` last changed `d0b95971` (2026-08-19)

Both predate the observed failure, so no later fix landed for it.

## Likely direction

Have the teardown wait for the daemon's connection process to actually be down before the test process exits, rather than issuing a stop and returning. The repo has hit the "stop is not the same as stopped" shape before (see the `await_stopped` / Horde deregistration flakes, #1590 and #1544).

jhgaylor:
The runner socket in `RunnersLiveTest` had no SQL Sandbox allowance: explicitly delivering its heartbeat reproduced `DBConnection.OwnershipError` in `Runners.touch/1` while the test was still alive. Grant the socket access to the test's connection and exercise that heartbeat directly. Stop the fake daemon in `try/after` from its linked starter, then await both socket and daemon exits before the test ends.

Fixes #1909.

Validation: the forced heartbeat failed before the allowance fix (3 tests, 1 failure). After the fix, `mix test test/fountain_web/live/runners_live_test.exs --seed 0 --repeat-until-failure 20` completed 21 consecutive runs (63 tests, 0 failures), using an isolated PostgreSQL database and OTP 28.3 / Elixir 1.19.2.

`mix precommit apps/fountain/test/fountain_web/live/runners_live_test.exs` passed compile, formatting, Credo, Dialyzer, Sobelow, prod release assembly, and the 3-test suite.

### DIFF
--- a/apps/fountain/test/fountain_web/live/runners_live_test.exs
+++ b/apps/fountain/test/fountain_web/live/runners_live_test.exs
@@ -14,21 +14,39 @@ defmodule FountainWeb.RunnersLiveTest do
       Runners.register(user.id, %{"name" => "mini", "os" => "darwin", "arch" => "arm64"})
 
     {:ok, daemon} = FakeDaemon.start(online.id, meta: %{user_id: user.id}, name: "mini")
-    on_exit(fn -> FakeDaemon.stop(daemon) end)
+    # The socket is a GenServer, so unlike a Task it does not inherit this
+    # async test's SQL Sandbox allowance through $callers.
+    Ecto.Adapters.SQL.Sandbox.allow(Fountain.Repo, self(), daemon.socket)
+    socket_ref = Process.monitor(daemon.socket)
+    daemon_ref = Process.monitor(daemon.daemon)
 
-    conn = login_user(conn, user)
-    {:ok, lv, html} = live(conn, ~p"/account/runners")
+    try do
+      conn = login_user(conn, user)
+      {:ok, lv, html} = live(conn, ~p"/account/runners")
 
-    assert html =~ "laptop"
-    assert html =~ "lap.local"
-    assert html =~ "mini"
-    assert html =~ "darwin · arm64"
-    assert html =~ "online"
-    assert html =~ "fountain runner"
+      assert html =~ "laptop"
+      assert html =~ "lap.local"
+      assert html =~ "mini"
+      assert html =~ "darwin · arm64"
+      assert html =~ "online"
+      assert html =~ "fountain runner"
 
-    lv |> element("#runner-#{offline.id} button", "Forget") |> render_click()
-    refute Runners.get_runner(offline.id, user.id)
-    refute render(lv) =~ "lap.local"
+      lv |> element("#runner-#{offline.id} button", "Forget") |> render_click()
+      refute Runners.get_runner(offline.id, user.id)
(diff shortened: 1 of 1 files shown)


BinaryBourbon (approved):
Adversarial review — independent Codex agent review of this exact commit; no blocking findings.

Traced the SQL Sandbox owner/allowance chain, heartbeat callback, FakeDaemon's links and stop implementation, and normal/assertion/crash teardown paths. Allowing the socket addresses its independent GenServer database access. Running stop from the linked starter avoids killing that starter, and monitoring both processes bounds their lifetime before the test finishes. The synchronous state call drains the explicitly sent heartbeat before cleanup.

Independent validation in a private checkout/database: all 3 targeted tests passed at seed 12345. Removing only the allowance reproduced DBConnection.OwnershipError in Runners.touch/1 during the forced heartbeat (3 tests, 1 failure). Restored the original source afterward.

Limits: targeted tests and source review, not a new full-suite run or prolonged periodic-heartbeat soak. Verified the PR head remained unchanged before submission.

PR #2078, opened 2026-09-13
test(sse): synchronize heartbeat and idle coverage

### LINKED ISSUE #1934: Flake: SSE heartbeat stream exits before the idle-window assertion

jhgaylor:
Fixes #1934.

Replace the heartbeat test's 700 ms sleep and liveness check against 50/200 ms timers with observable stream behavior. One test runs the real heartbeat timers and verifies two successful heartbeat writes, a third attempted write, and a clean return when the client adapter closes. A second queues each heartbeat before releasing the adapter with a zero idle window, proving delivered heartbeats keep the receive loop open even when idle expiry is immediately eligible. The existing quiet-stream expiry test remains.

The per-connection adapter delegates actual writes to Plug's test adapter. Its barriers, task awaits and cleanup are bounded. Production code and the global test helper are unchanged, so this stays independent of the TeamStream fix in #2072.

The historical full-suite failure was not reproduced; this removes its identified unsynchronized timing assertion and tests heartbeat delivery, rescheduling, idle behavior and disconnect directly.

Validation (isolated database and build directories, pinned Elixir 1.19.2 / OTP 28.3):

- `mix precommit apps/fountain/test/fountain_web/controllers/sse_stream_test.exs` passed every stage, including Dialyzer with zero errors, prod release assembly, and 19 tests / 0 failures. Test scope was this file, not the full suite.
- Full SSE file passed at the issue's seed `246102` (19 tests / 0 failures).
- Removing the production heartbeat reschedule in a temporary negative control made the real-timer test fail at its bounded await; production code was restored immediately afterward.
- Both changed tests passed together ten times at seed `246102` with `ERL_FLAGS='+S 1:1'` (20 test executions / 0 failures).
- `git diff --check` passed; the only changed file is the SSE test.

### DIFF
--- a/apps/fountain/test/fountain_web/controllers/sse_stream_test.exs
+++ b/apps/fountain/test/fountain_web/controllers/sse_stream_test.exs
@@ -308,30 +308,121 @@ defmodule FountainWeb.SseStreamTest do
       assert elapsed < 3_000, "the idle timeout did not fire (took #{elapsed}ms)"
     end
 
-    test "heartbeats hold the connection open past the idle timeout", %{
+    defmodule HeartbeatAdapter do
+      @moduledoc false
+
+      defdelegate send_chunked(state, status, headers), to: Plug.Adapters.Test.Conn
+      defdelegate get_peer_data(state), to: Plug.Adapters.Test.Conn
+      defdelegate get_sock_data(state), to: Plug.Adapters.Test.Conn
+      defdelegate get_ssl_data(state), to: Plug.Adapters.Test.Conn
+      defdelegate get_http_protocol(state), to: Plug.Adapters.Test.Conn
+      defdelegate read_req_body(state, opts), to: Plug.Adapters.Test.Conn
+      defdelegate send_resp(state, status, headers, body), to: Plug.Adapters.Test.Conn
+
+      def chunk(state, body) do
+        send(state.observer, {:heartbeat_chunk, self(), IO.iodata_to_binary(body)})
+
+        if Map.get(state, :pause, false) do
+          receive do
+            :continue_heartbeat -> :ok
+          after
+            5_000 -> raise "test did not acknowledge the heartbeat"
+          end
+        end
+
+        if state.remaining == 0 do
+          {:error, :closed}
+        else
+          Plug.Adapters.Test.Conn.chunk(%{state | remaining: state.remaining - 1}, body)
+        end
+      end
+    end
+
+    defmodule HeartbeatEndpoint do
+      @moduledoc false
+
+      defdelegate init(opts), to: FountainWeb.Endpoint
+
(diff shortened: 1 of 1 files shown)


BinaryBourbon (approved):
Adversarial review (independent Codex agent): no blocking findings.

Reviewed the exact diff against #1934 and traced the real controller receive loop, heartbeat scheduling, disconnect branch, and Plug test-adapter state propagation. The real-timer test cannot finish successfully without rescheduling and two successful writes; the queued-heartbeat test preserves sender order before releasing each chunk and exercises message handling with an immediately eligible idle timeout. The existing quiet-stream expiry check remains. This removes the unsynchronized liveness assertion without changing production behavior or bypassing actual chunk writes.

Validation limit: this review used source and fixture analysis; I did not independently rerun the database-backed SSE suite or reproduce the historical loaded-run failure. The author's reported CI/repetition results were not treated as proof of correctness.

PR #2079, opened 2026-09-13
fix(ci): prevent Mix build locks from waiting on their own reused ports

### LINKED ISSUE #1997: Flake: Mix build lock stalls CI database setup before tests start
While checking the P2 draft stack, the first CI attempt for #1979 stalled during database setup and exhausted the 20-minute job timeout. No tests or test assertions ran in the affected partition.

Evidence: [attempt 1, partition 6](https://github.com/managoat/fountain/actions/runs/34634731091/job/103379807004), head `c18710cfa5d2f996267972b621c0643737c48c43`.

The `Set up database` step runs `mix ecto.create --quiet && mix ecto.migrate --quiet`. It started at 18:42:21 UTC, compiled Fountain, then stopped here at 18:42:40:

```text
==> fountain
Compiling 387 files (.ex)
Generated fountain app
==> fountain_buzz
Waiting for lock on the build directory (held by process 2770)
```

At 19:02:16 the runner canceled the operation and cleanup terminated the orphan `beam.smp` process with PID 2770. The job used the repository-pinned Elixir 1.19.2 / OTP 28.3, restored the test build cache, and successfully installed dependencies and checked PostgreSQL readiness before entering this step.

Frequency: one stalled partition among the six in this PR's first CI attempt. Other P2 drafts have completed the same setup, but a broader failure rate has not been measured. A single retry of the affected jobs was requested on the unchanged head; [attempt 2](https://github.com/managoat/fountain/actions/runs/34634731091/attempts/2) passed database setup, partition 6, coverage and CI required without any code change.

The cause is unresolved. The log does not establish whether the lock holder was a nested Mix invocation or another surviving process. The root migration alias loads app configuration and then shells into the Fountain app, which is a path to investigate; this is not a confirmed diagnosis. Preserve process trees and Mix lock diagnostics if it recurs, and reproduce against the restored cache before changing locking or compilation behavior. Do not disable the build lock or extend the timeout to hide the stall.

Searched existing open and closed `flake` issues before filing; none described this build-lock/database-setup failure. The lifecycle change does not edit Mix tasks or CI setup. No code change was made merely to obtain a green retry.

jhgaylor:
Mix 1.19.2 can wait forever on its own build lock when the OS reuses a listener port. A stale `port_*` file is still hardlinked to a lock file; rewriting it makes the next acquisition probe its own listener. That self-wait can stall CI database setup before tests start.

Backport the exact upstream fix from [elixir-lang/elixir#15765](https://github.com/elixir-lang/elixir/pull/15765), commit `b703302cd1fc91dd1b2df51659daf9b8d6053776`. The wrapper checks Elixir 1.19.2 and source fingerprints, applies the vendored patch to a private temporary copy, and compiles an isolated module. An Erlang `-eval` loads that module before the Elixir CLI adjusts its code path, and also supports embedded release boot. All six pinned core CI jobs export the bootstrap through `ERL_AFLAGS`, so nested Mix VMs use the same fixed lock implementation and namespace.

The installed toolchain is untouched. Locking stays enabled. The wrapper refuses an unreviewed toolchain upgrade; the maintenance notes explain how to retire it when a stable Elixir release includes the fix.

Database setup also emits periodic Mix stack and OS process diagnostics if another stall occurs. The reporter excludes environment variables, messages, dictionary values and command arguments. BEAM snapshots cover the outer VM; nested VMs appear in the OS process listing.

Fixes #1997.

Validation:

- The pinned implementation deterministically waits on its own PID after port reuse; the patched implementation reacquires using real sockets and hardlinks.
- Four regressions pass: normal port reuse, crashed-owner port reuse, inherited cross-VM mutual exclusion, and embedded release boot.
- An independent reviewer verified startup ordering and the patched module origin in ordinary Elixir and embedded Erlang VMs.
- All 57 CI Python tests, ShellCheck and actionlint passed.
- CI installation mode preserved existing `ERL_AFLAGS`; the regression suite passed again with the exported environment.
- Both diagnostic regression tests passed with the patched module, including real contending locks and exclusion of process-dictionary secrets.
- The final branch passed scoped `mix precommit apps/fountain/test/fountain/extension_composition_test.exs`: warnings-as-errors, unused dependencies, formatting, Credo, Dialyzer, Sobelow, production release assembly and 24 tests with zero failures.
- The exact diagnostic `ecto.create` and `ecto.migrate` entry points passed against an isolated database with the backport loaded, including the nested migration Mix VM.
- An actual detached release smoke test passed configuration, HTTP readiness (200), RPC and shutdown with the final loader.

The historical failed runner did not retain its port files or live process tree, so the precise historical cause cannot be proved from that log alone. The regression demonstrates and fixes the upstream self-lock defect on our pinned toolchain.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Unified diffs use a single space to mark an empty context line.
+scripts/ci/mix-lock-15765.patch whitespace=-blank-at-eol
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -286,6 +286,13 @@ jobs:
           elixir-version: "1.19.2"
           otp-version: "28.3"
 
+      - name: Backport the pinned Mix lock fix
+        run: scripts/ci/mix-lock-backport.sh --install
+
+      - name: Mix lock backport preserves port reuse and cross-VM exclusion
+        if: ${{ matrix.partition == 1 }}
+        run: elixir scripts/ci/mix-lock-backport-test.exs
+
       - name: Timing collection preserves ExUnit execution
         if: ${{ matrix.partition == 1 }}
         run: elixir scripts/ci/timing-formatter-test.exs
@@ -448,6 +455,9 @@ jobs:
           elixir-version: "1.19.2"
           otp-version: "28.3"
 
+      - name: Backport the pinned Mix lock fix
+        run: scripts/ci/mix-lock-backport.sh --install
+
       - name: Download coverage exports
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
@@ -531,6 +541,9 @@ jobs:
         with:
           elixir-version: "1.19.2"
           otp-version: "28.3"
+
+      - name: Backport the pinned Mix lock fix
+        run: scripts/ci/mix-lock-backport.sh --install
 
       # Job-scoped keys — see the note in `test`.
       - name: Restore deps cache
@@ -651,6 +664,9 @@ jobs:
         with:
           elixir-version: "1.19.2"
           otp-version: "28.3"
+
+      - name: Backport the pinned Mix lock fix
+        run: scripts/ci/mix-lock-backport.sh --install
 
       # Job-scoped keys — see the note in `test`.
       - name: Restore deps cache
@@ -1375,6 +1391,9 @@ jobs:
           elixir-version: "1.19.2"
           otp-version: "28.3"
 
+      - name: Backport the pinned Mix lock fix
+        run: scripts/ci/mix-lock-backport.sh --install
+
       - name: Restore deps cache
         uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v5
         with:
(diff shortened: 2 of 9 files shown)


BinaryBourbon (approved):
Adversarial review (independent Codex agent acting through BinaryBourbon): APPROVE at 23bb2bc43a1d9453a4974217efec0baae322f4c7. No blocking finding after inspecting the exact nine-file PR diff, #1997, pinned Mix implementation, upstream commit, workflow coverage and diagnostic output paths.

I independently verified that every vendored added/removed line matches upstream b703302cd1fc91dd1b2df51659daf9b8d6053776; only patch hunk grouping differs. The source/version guards, private compilation, early module preload and inherited ERL_AFLAGS preserve the fix across nested VMs. All six pinned core jobs install before Mix; the documented v2 namespace restriction correctly requires isolated builds when unpatched VMs are present. The reporter emits arities instead of frame arguments and excludes arbitrary dictionary values, environment and argv.

Validation from an isolated archive of this exact head on Elixir 1.19.2/OTP 28.3: four lock regressions passed (including the deterministic old-implementation self-wait negative control, crashed-owner port reuse, child-VM mutual exclusion and embedded boot); both diagnostic contention/privacy regressions passed; CI install mode preserved existing +S 2:2 flags and all four lock regressions passed again. All 57 CI Python tests, ShellCheck on the wrapper and actionlint passed. An initial Python invocation used an incomplete source export and failed on missing fixture files; rerunning from the complete exact archive resolved those setup errors.

Limits: this demonstrates the pinned upstream self-lock defect, not the precise historical #1997 interleaving. I independently exercised embedded Erlang boot, not a complete detached Fountain release/HTTP smoke test or all hosted CI jobs.

PR #2080, opened 2026-09-13
test(conversations): await startup and own server teardown

### LINKED ISSUE #1702: Flake: ConversationServer startup test failed on an unchanged file while landing #1645
#1645's body reports "a failure in an unchanged ConversationServer startup test" during its CI runs, with evidence retained only locally. No issue was filed.

Needed: the test name, the assertion with `left:`/`right:`, the run URL and the seed. Replay the seed on `main` before calling it a flake (the #1469 lesson).

---

jhgaylor:
`ConversationServerCase.start_server/2` treated the five-second timeout from `:sys.get_state/1` as a stopped server even when provisioning was still running. The resulting assertion failure could release SQL ownership underneath that server. Wait for provisioning to settle under ExUnit's existing test timeout, and give ExUnit ownership of temporary server children so teardown finishes before the database owner is released. Keep DataCase's separate shared owner instead of replacing it with the test process.

A regression holds sprite creation past five seconds and reproduces `left: :stopped, right: :alive` on the unchanged helper. Another leaves a server running at test completion and checks that teardown stops it and revokes its callback key while the SQL owner still exists. Existing stopped-server behavior remains covered by the provisioning failure tests.

Fixes #1702. Recovered the original local evidence: seed `836199`, `ConversationServerTest`'s “a live server stops without destroying the sprite; the rows say so” (then line 1224), actual `start_server` result `:stopped`, followed by SQL ownership errors in the still-running provision. The original captured exit reason was missing; the new controlled regression establishes this harness race directly.

Validation:
- The delayed-provision regression fails on the unchanged helper (`:stopped` instead of `:alive`) and passes with the fix at seeds 0, 121815 and 836199.
- `ConversationServerTest`: 60 tests, 0 failures at recovered seed 836199, including teardown ownership.
- `mix precommit` with all 25 files using `ConversationServerCase`, seed 836199: 353 tests, 0 failures; compile, format, Credo, Dialyzer, Sobelow and production release assembly passed. Dedicated database and independent builds. Full-suite CI will run on this PR.

### DIFF
--- a/apps/fountain/test/fountain/conversations/conversation_server_test.exs
+++ b/apps/fountain/test/fountain/conversations/conversation_server_test.exs
@@ -51,6 +51,28 @@ defmodule Fountain.Conversations.ConversationServerTest do
   end
 
   describe "provisioning — happy path" do
+    test "a provision outlasting the system-call timeout is still alive", %{conv: conv} do
+      handle = stub_happy_sprite()
+
+      Mimic.stub(Managoat.Sandbox.Sprites, :create, fn _name, _opts ->
+        # Deliberately cross :sys.get_state/1's five-second timeout. This
+        # reproduces #1702 without relying on scheduler load to delay startup.
+        Process.send_after(self(), :finish_create, 5_100)
+        receive do: (:finish_create -> {:ok, handle})
+      end)
+
+      {pid, _ref, settled} = start_server(conv)
+
+      try do
+        assert settled == :alive
+        assert Conversations._unsafe_get_sandbox!(conv.sandbox_id).status == "ready"
+      after
+        # Also drain the old helper's timed-out, still-provisioning server when
+        # running this regression against the parent commit.
+        GenServer.stop(pid, :normal, :infinity)
+      end
+    end
+
     test "drives the sandbox to ready", %{conv: conv, sandbox: sandbox} do
       stub_happy_sprite()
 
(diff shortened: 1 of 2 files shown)


BinaryBourbon (approved):
Adversarial review (independent Codex agent acting through BinaryBourbon): APPROVE at 2dc7daecb467da9fc19ee30abcd25d651743879f. No blocking finding after inspecting the exact two-file diff, #1702, DataCase ownership, ExUnit shutdown/callback ordering, Mimic cleanup and ConversationServer initialization/termination.

The startup barrier now distinguishes slow provisioning from a stopped process: :sys.get_state/2 with :infinity waits behind handle_continue, while ExUnit still bounds an actually hung test. Temporary ExUnit-supervised children cannot restart on abnormal exits and do not link server failures to the test. Removing the redundant shared-mode override preserves DataCase's separate owner; ExUnit waits for child shutdown before on_exit callbacks, so callback-key revocation in terminate/2 can complete before owner release. Existing ordinary provisioning-failure behavior remains covered.

Independent validation used a full archive of this exact head, an independent APFS-cloned build/dependency tree and a dedicated PostgreSQL database, on Elixir 1.19.2/OTP 28.3. The complete changed ConversationServerTest module passed at recovered seed 836199: 60 tests, 0 failures, including the new startup and teardown regressions. As a negative control I replaced only the helper with its unchanged parent version in the isolated tree: the new delayed-provision test failed exactly with left: :stopped, right: :alive. Restoring the exact head helper made both new regressions pass again (2 tests, 0 failures).

Limits: I did not independently rerun all 25 harness consumers, the full umbrella suite or all precommit stages. The historical issue lacks the original captured exit reason; the controlled negative test establishes the concrete helper defect without claiming to reconstruct every historical failure.

PR #2083, opened 2026-09-13
Remove deprecated BILLING_ENABLED fallback

jhgaylor:
Remove the deprecated `BILLING_ENABLED` environment-variable fallback and its boot warning. `CREDITS_ENABLED` is now the sole control, with the existing default of `false`. A deployment setting only `BILLING_ENABLED=true` will leave credits disabled; home-cloud, the known consumer, already sets `CREDITS_ENABLED=true` in its deployment patch and generated manifests.

Update the configuration reference and environment example, add an upgrade note, and replace the alias test with regressions for the retired variable alone and conflicting values for both variables.

Validation:
- Configuration and documentation tests: 52 tests, 0 failures.
- Compilation with warnings as errors, formatting, unused-dependency check, and strict Credo passed.
- `git diff --check` and Destink passed. Advisory docs-style/Vale reports have existing findings outside this change.
- Full `mix precommit` is still running locally (Dialyzer cache rebuild); CI will run the full gates.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -284,7 +284,6 @@ SLACK_OAUTH_CLIENT_SECRET=
 # packs sold through Stripe Checkout, a 402 at zero. That is meaningful for
 # the hosted instance and a lock with no key on a self-hosted one.
 # CREDITS_ENABLED=true
-# (BILLING_ENABLED is the old name and still read, with a warning, for one release.)
 
 # Stripe billing (only with CREDITS_ENABLED=true)
 STRIPE_SECRET_KEY=
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ upgrade, is in
 
 ### Upgrade notes
 
+- `BILLING_ENABLED` is no longer supported. Set `CREDITS_ENABLED=true` to
+  enable credits. With only the old variable set, credits remain off by default.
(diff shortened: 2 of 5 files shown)


BinaryBourbon (approved):
Independent adversarial review by a Codex sub-agent, posted as BinaryBourbon at the maintainer's explicit request. Reviewed all five changed files at fb124c2ff3d5e3d3bed8f4f9b444752e7330dc24 and the surrounding credits gate and production boot checks.

Approved: no actionable correctness or regression findings. The runtime no longer reads BILLING_ENABLED; CREDITS_ENABLED retains its existing parsing and default-off behavior. Production Stripe validation uses the same computed flag. Regression tests cover the retired variable alone and both conflicting combinations, and the upgrade note clearly describes the behavior change. Independently checked that home-cloud's deployment patch and generated manifests both set CREDITS_ENABLED=true.

Validation: 66 focused configuration and documentation tests passed on the PR's migrated isolated database. An initial run against the stale default local database failed on the pre-existing sandboxes.teardown_requested_at migration, followed by a gate-state failure; the same seed passed after selecting the correctly migrated database. All six CI test partitions and release-boot checks have passed. Elixir static analysis remains in progress; this code-review approval does not claim the full CI gate has completed.

PR #2086, opened 2026-09-13
fix: redact secret values from sandbox directory listings

### LINKED ISSUE #1923: list/2 does not redact, so /files returns entry names raw
`Fountain.SandboxFiles.list/2` is the one read in the module that does not redact. `GET /api/sandboxes/:id/files` returns entry names exactly as the filesystem holds them, while every sibling read passes its strings through `redact_with/2`.

## The contradiction

The module's own docstring (`apps/fountain/lib/fountain/sandbox_files.ex:29`, on `main`) states the guarantee:

> Every byte that leaves goes through the same redaction the transcript gets: the values of the identity's environment and vault, plus whatever a live `ConversationServer` registered, replaced with `[REDACTED]`. The `.env` file is on that disk in plaintext, so this is what keeps a third-party app holding the user's key from reading the user's secrets

`list/2` (`:205`) does not do this. It runs `list_script()`, calls `parse_entries/1`, and returns the entries directly — no `secret_values/1`, no `redact_with/2`, no `to_text/1`.

Every sibling does redact:

- `read/3` — `bytes = redact(sandbox, bytes)` (`:244`)
- `diff/3` — `redact_with(values, ...)` on the body (`:303`) and on `repo_root` (`:311`)
- `status/3` — `repo_root`, `branch`, and every change through `redact_change/2` (`:366-369`)
- the failing-command path — `to_text(redact(sandbox, output))` (`:450`)

`redact_change/2` even carries the comment that makes the intent explicit: *"A path is data the agent chose, so it goes through redaction like file content."* A directory entry name is the same kind of data, chosen by the same agent.

## Why it matters

A filename is a plausible place for a secret to land: an agent writing `~/.config/<token>.json`, a checkout directory named after a key, a scratch file named from an environment variable. `/git-status` and `/diff` redact exactly these paths; `/files` does not, so the same name is redacted through one endpoint and raw through another.

The threat model is the one the docstring names: a third-party OAuth app holding a full-scope token is a different principal from the tenant whose vault values redaction protects.

## Fix

Route entry names through `redact_with/2` and `to_text/1` the way `redact_change/2` does, resolving `secret_values(sandbox)` once per call rather than per entry.

Note the interaction with #1907: `list/2`'s cap is `@max_entries`, a **count** of whole entries, not a byte cap, so it cannot bisect a name the way `head -c` bisected file content. No overlap logic is needed here — this is the simple case.

Found during review of #1915. This is the last unredacted exit in the module.

jhgaylor:
Directory listings returned secret-bearing filenames verbatim. Redact environment, vault, and live conversation values from entry names and the response path, resolving secrets once per successful listing. Redaction runs before invalid UTF-8 is converted for JSON; the provider receives the original path.

Preserve directory ordering, entry metadata, duplicate display names, and the 2,000-entry cap. Regression tests cover each secret source, mixed encodings, and exact versus truncated listings.

Fixes #1923.

Validation:
- Sandbox file context, shell script, and API controller suites: 78 tests passed.
- Scoped `mix precommit` passed on `c2bca428967d41233cbe625cb18f6ae8564c7df0`: compilation, unused-dependency check, formatting, strict Credo, Dialyzer (zero errors), Sobelow, production release assembly, and the same 78 tests. This is scoped local validation; the full suite remains for CI.

### DIFF
--- a/apps/fountain/lib/fountain/sandbox_files.ex
+++ b/apps/fountain/lib/fountain/sandbox_files.ex
@@ -207,11 +207,15 @@ defmodule Fountain.SandboxFiles do
          {:ok, absolute} <- resolve_path(sandbox, path),
          {:ok, output} <- run(sandbox, list_script(), [absolute]) do
       entries = parse_entries(output)
+      values = secret_values(sandbox)
 
       {:ok,
        %{
-         path: absolute,
-         entries: Enum.take(entries, @max_entries),
+         path: to_text(redact_with(values, absolute)),
+         entries:
+           entries
+           |> Enum.take(@max_entries)
+           |> Enum.map(fn entry -> %{entry | name: to_text(redact_with(values, entry.name))} end),
          truncated: length(entries) > @max_entries
        }}
     end
--- a/apps/fountain/test/fountain/sandbox_files_test.exs
+++ b/apps/fountain/test/fountain/sandbox_files_test.exs
@@ -163,6 +163,112 @@ defmodule Fountain.SandboxFilesTest do
                SandboxFiles.list(ctx.sandbox, nil)
     end
 
+    test "redacts environment, vault and live conversation values from names and the path", ctx do
+      {:ok, dek} = Crypto.load_tenant_key(ctx.user.id)
+      env = insert_env(user_id: ctx.user.id)
+      vault = insert_vault(user_id: ctx.user.id)
+
+      {:ok, _} =
+        Environments.upsert_secret(env, %{"key" => "TOKEN", "value" => "sk-env-secret"}, dek)
+
+      {:ok, _} =
+        Fountain.Vaults.upsert_secret(
+          vault,
+          %{"key" => "TOKEN", "value" => "sk-vault-secret"},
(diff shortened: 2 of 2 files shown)


BinaryBourbon (approved):
Independent adversarial review: no blocking correctness, security, compatibility, or regression-coverage findings. Reviewed both changed files in full against base 6a6cd94560bb54389d8157d4e86b42e20888ba6a, issue #1923, and ADR 0039; traced the full-scope route and tenant-scoped controller fetch through secret resolution, listing parsing, response encoding, schema, and SDK callers. The fix redacts original name bytes and the response path using one secret lookup, preserves provider arguments and entry metadata, and retains existing ordering, duplicate entries, and the 2,000-entry boundary.

Fresh diagnostic on this exact clean head: sandbox file context, shell-script, and API controller suites passed, 78 tests, 0 failures (seed 113123; prepared local toolchain, --no-compile). Current CI required and Detect secrets are successful. I did not independently rerun the entire cold precommit or all broader verification commands.

Issue #2054, opened 2026-09-13
Completion no longer consults the execution journal, and an already-ended turn leaves the parent running


jhgaylor:
Two related gaps left deliberately out of scope by #1999, recorded with their measurements so the next person does not have to rediscover them.

## 1. The journal bypass

On `main` before #1999, the completion write went through `_unsafe_update_turn/2` → `ExecutionGuard._unsafe_write_turn/3`. `_unsafe_complete_turn/3` takes the parent lock itself and calls `Repo.update!` directly, so the journal's conditions no longer apply at this ending:

- deadline expiry, `uncertain_spawn` and the `:execution_not_started` rollback are not evaluated;
- a normal bounded completion leaves the journal `active` where main left it `ready` (the real actor path still reaches `ready` via `close_bounded_connection`, so the end state matches and only the window is new);
- **a journal-fenced turn no longer idles its parent** — measured at `bounded_lifecycle_test.exs:362`, where `main` leaves `conv=idle` and the branch leaves `conv=running`, with the same `["started", "failed"]` stages and the suite green either way.

**Inert today**, which is why it did not block: `ExecutionLimits.enforced_controls/1` returns `[]` (`execution_limits.ex:58`), so `_unsafe_register_bounded/3` returns `:unbounded` and no `turn_executions` row is written in production at all. This is a latent revert of #1732/#1773-#1793's machinery, not a live incident.

Re-plumbing completion through the guard is the fix, and it is a change of its own — #1999 was already at its fourth review round.

## 2. An already-ended turn leaves the parent running — and this one is NOT journal-dependent

Raised by the reviewer on the final round, and the sharper half. `_unsafe_complete_turn/3` matches `%Turn{status: "running"}`, so any turn already ended by another writer leaves the parent `running` where `main` idled it. Probed with no journal: `main` → `idle`, branch → `running`.

There is no live producer of that shape today. **But #2000's `_unsafe_idle_interrupted_turn/2` lands on exactly it.** If it inherits "already terminal, therefore write nothing" without also inheriting an idle decision, that becomes a live stuck-`running` conversation — and a stuck-`running` parent feeds `_unsafe_sandbox_busy_elsewhere?/4`, so the co-tenant machine bills to `SANDBOX_MAX_LIFETIME_HOURS` and `SandboxReaper` will not reclaim a sandbox with a live server.

**Check this before or while landing #2000**, not after.

## Why they are together

[...]

PR #2087, opened 2026-09-13
test: guard activation instrumentation at every reply writer

### LINKED ISSUE #2055: Nothing pins which turn writers must call Activation.turn_replied/1
`Fountain.Activation.turn_replied/1` decides whether an account counts as activated (#1392). Its moduledoc claimed until yesterday that it is called from a single choke point — `Conversations._unsafe_update_turn/2` — "so a new way for a turn to end is instrumented by construction rather than by remembering".

**That stopped being true at #1473** (`93cbee9a`, 2026-09-03, "recover orphaned running turns"), which added a second call site at `conversations.ex:2300` and did not touch `activation.ex`. The moduledoc was false for nine days and nothing failed. #1999 added a third site and corrected the text.

Nothing enforces it. `activation_test.exs` unit-tests `turn_replied/1` but nothing walks the endings, so the next writer that materializes a reply and forgets the call is silent — and the failure it produces is a real user who never counts as activated.

## Suggested shape

`audit_guardrail_test.exs` generalizes here. Enumerate every function that can set `reply_text` from `nil` and fail unless each calls `turn_replied/1` or appears on a documented exclusion list with a reason:

- `_unsafe_update_turn/2`
- `_unsafe_complete_turn/3`
- `_unsafe_orphan_turn/3`
- `_unsafe_backfill_reply_texts/0`

A bare call-site *count* would have caught #1473 but not the real hole, which is that the backfill writes `reply_text` and never activates. So the check needs to be per-writer, not a count.

The backfill's omission is arguably legitimate — it is a repair path, not a first reply. Under this shape it would be excluded **explicitly, with its reason**, rather than by nobody having noticed. That is the whole value: #2000, #2001 and #2003 each add a turn writer, and the guardrail forces the decision each time instead of leaving it to whoever reviews.

## Why now

Three PRs in the current campaign add writers on this path, and one consequence of the gap is already recorded in #2054: a turn ended by another writer can leave the account permanently unactivated, because the backfill will not activate it later.

Refs #1999, #1473, #1392, #2054, ADR 0013 (the audit guardrail this pattern comes from).

jhgaylor:
Reply-writing paths can omit activation instrumentation without failing a test. Discover reply writers in production source and check each for `Activation.turn_replied/1` or an explicit, documented exclusion. Pin the four current writers so a discovery regression cannot silently reduce coverage. Historical reply backfill remains explicitly excluded; this does not resolve the separate repair behavior in #2054.

The guard checks visible source calls, not runtime control flow. It scans every executable function-body section, including `rescue`, `catch`, `after`, and `else`. Its regression fixtures cover new uninstrumented writers, local helpers, private writers, handler-only writes, uninstrumented clauses sharing an instrumented function's name and arity, and stale or undocumented exclusions.

Closes #2055.

Validation:
- Activation and guardrail suites: 22 tests passed. The three exception-handler regressions failed against the original discovery logic and passed after the repair.
- CI policy suite: 57 tests passed; alert fixtures: 58 cases passed.
- Scoped `mix precommit` passed on `5c6530192ec5b047b295a32157031add0ff7b99a`: compilation, unused-dependency check, formatting, strict Credo, Dialyzer (zero errors), Sobelow, production release assembly, and the same 22 tests. This is scoped local validation; the full suite remains for CI.

### DIFF
--- a/apps/fountain/test/fountain/activation_guardrail_test.exs
+++ b/apps/fountain/test/fountain/activation_guardrail_test.exs
@@ -1,0 +1,261 @@
+defmodule Fountain.ActivationGuardrailTest do
+  @moduledoc """
+  Every reply materializer must decide activation at the write (#2055).
+
+  Like AuditGuardrailTest, intentional silence needs a named function and a
+  reason. Unlike a fixed inventory, discovery scans production source for
+  reply_text maps/keywords and field setters, follows local builder helpers,
+  and checks each function that also calls a Repo insert/update operation.
+  Private writers count; a call elsewhere in the module does not cover them.
+
+  This checks visible calls, not control flow or data flow: dynamic modules or
+  field names, raw SQL and builders in another module need discovery extended
+  when introduced.
+  ActivationTest covers the behavior of the live seams.
+  """
+
+  use ExUnit.Case, async: true
+
+  @root Path.expand("../../../..", __DIR__)
+  @deliberately_silent %{
+    {Fountain.Conversations, :_unsafe_backfill_reply_texts, 0} =>
+      "Historical repair for turns predating reply_text avoids replaying activation " <>
+        "analytics per repaired turn; the analytics/stamping repair gap remains " <>
+        "tracked in #2054."
+  }
+
+  test "every discovered reply writer calls activation or has a documented exclusion" do
+    functions =
+      ["apps/*/lib/**/*.ex", "ee/lib/**/*.ex"]
+      |> Enum.flat_map(&Path.wildcard(Path.join(@root, &1)))
+      |> Enum.flat_map(&(File.read!(&1) |> functions()))
+
+    writers = writers(functions)
+
+    assert Enum.map(writers, &elem(&1, 0)) == [
+             {Fountain.Conversations, :_unsafe_backfill_reply_texts, 0},
+             {Fountain.Conversations, :_unsafe_orphan_turn, 3},
+             {Fountain.Conversations, :_unsafe_update_turn, 2},
+             {Fountain.Conversations, :end_running_turn, 5}
+           ],
+           "Reply-writer inventory changed: review discovery and update the inventory explicitly"
+
+    assert violations(writers, @deliberately_silent) == []
+  end
+
+  test "a new writer is detected even when a neighboring writer calls activation" do
+    source = """
+    defmodule Example do
+      def covered(turn, text) do
+        updated = turn |> Turn.changeset(%{reply_text: text}) |> Repo.update!()
+        Fountain.Activation.turn_replied(updated)
+      end
+
+      def newly_added(turn, text) do
+        turn |> Turn.changeset(%{"reply_text" => text}) |> Repo.update!()
+      end
+    end
+    """
+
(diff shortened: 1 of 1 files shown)


apps/fountain/test/fountain/activation_guardrail_test.exs
```diff
@@ -0,0 +1,261 @@
...
+  end
+
+  defp module_functions(module, body) do
+    expressions =
+      case body do
+        {:__block__, _, expressions} -> expressions
+        expression -> [expression]
+      end
+
+    for {kind, _, [head, [do: body]]} <- expressions, kind in [:def, :defp] do
```
BinaryBourbon: [P2] Discover functions that have implicit try sections

The exact `[do: body]` match excludes every `def`/`defp` with function-level `rescue`, `catch`, or `after`, because its AST body keyword list contains additional keys. For example, `def finish(turn, text) do; turn |> Turn.changeset(%{reply_text: text}) |> Repo.update!(); rescue e -> {:error, e}; end` is a visible reply writer with a supported Repo call, yet discovery returns no writer and violations returns `[]`. Adding this as a new writer leaves the pinned inventory unchanged; adding it as an uninstrumented clause beside an instrumented clause of an existing MFA also passes. This bypasses the per-writer decision #2055 is intended to enforce. Accept the full function-body keyword list and scan its executable sections, then add regression fixtures for an uninstrumented rescued writer and an uninstrumented rescued clause sharing an instrumented MFA.

BinaryBourbon (changes requested):
Independent adversarial review of the entire one-file PR at d7949c08ce21d44b9240599fe51f2764b73a3b74, against base 6a6cd94560bb54389d8157d4e86b42e20888ba6a and issue #2055. The current four production writers and documented historical-backfill exclusion match the intended scope; there is no production/API change. One blocking discovery gap remains: ordinary function-level exception handlers silently remove writers from analysis (inline finding).

Diagnostics: the four original guardrail tests passed via direct ExUnit. An isolated probe using the same private discovery helpers reported missing_activation for a plain writer, but no violation for the identical writer with a function-level rescue, or for an uninstrumented rescued clause added beside an instrumented clause of the same MFA. No repository files were modified. Database-backed activation tests and full CI were not rerun in this review.

BinaryBourbon (approved):
Independent adversarial re-review of the entire one-file PR at 5c6530192ec5b047b295a32157031add0ff7b99a against base 6a6cd94560bb54389d8157d4e86b42e20888ba6a: no blocking findings remain.

The original P2 is resolved: discovery now retains all function-body sections, so function-level rescue/catch/else/after no longer discard visible reply writers. The added fixtures cover writes in both the main body and handlers, plus an uninstrumented rescued clause sharing an instrumented MFA. I independently reran my original probes: the plain writer, rescued writer, and silent rescued same-MFA clause each now report missing_activation. All seven guardrail tests passed via direct ExUnit.

The current four production writers and explicit historical-backfill exclusion remain consistent with #2055; this does not resolve the separate repair behavior in #2054. This review did not rerun the database-backed activation suite or full CI. The guard remains a visible-source-call check with the documented control/data-flow limitations.

PR #2093, opened 2026-09-13
fix(swift): stop cancelled runs before final status fetch

jhgaylor:
Cancelling a `Run` while its event stream is suspended can end the stream normally, allowing the cancelled follower to issue a final status GET. Check task cancellation before that fetch so cancellation stops further requests.

Adds a deterministic regression that waits for follower teardown and asserts that the cancelled run sends no final status request. The regression fails before the fix; 10 conformance tests and 4 Run tests pass with warnings treated as errors.

Found while validating the #2018 stack; this existing main defect is kept in a separate change.

### DIFF
--- a/sdk/swift/Sources/FountainKit/Run/Run.swift
+++ b/sdk/swift/Sources/FountainKit/Run/Run.swift
@@ -155,6 +155,9 @@ public final class Run: @unchecked Sendable {
     emit(.conversation(conversation, url: url))
     do {
       try await withDeadline()
+      // AsyncThrowingStream may finish normally when its task is cancelled.
+      // Do not turn that shutdown into a successful final-status request.
+      try Task.checkCancellation()
       let (ended, died) = read { ($0, failureReason) }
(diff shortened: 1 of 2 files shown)


BinaryBourbon (approved):
Approved after adversarial review of d24aa65db4f4fee1f282f73a4637da57bad1c87c. No blocking correctness or regression findings.

The cancellation check covers the normal-return path from a cancelled AsyncThrowingStream before the final conversation-status GET. The regression waits for the follower to release the Run before inspecting requests, so the immediate settlement of value() cannot hide that trailing request.

Validation: Swift 6.1.2 on macOS, ConformanceTests and RunTests passed with warnings treated as errors (14 tests; documented matrix skips remain). In an isolated source snapshot, removing only the new cancellation guard made the new regression fail on the unwanted final GET; restoring the reviewed code made the suites pass. Existing macOS and Linux Swift CI checks are also green. Local validation used scripted transports; no live service test was run.

Issue #2107, opened 2026-09-13
Agents: replace the legacy null allowlists (vaults, environments, credential sets) with explicit policy


jhgaylor:
Part of #2094. Evidence: main at `0c8443b6`, inspected 2026-09-13.

## Current maintenance burden

allowed_vault_ids: nil retains the old any-tenant-vault behavior, while [] denies all vaults. Conversations.check_vault_allowed treats nil as an unconditional allow after tenant scoping. Removing that branch without migration would change existing agents silently.

## Source

- [apps/fountain/lib/fountain/agents/agent.ex](https://github.com/managoat/fountain/blob/0c8443b6/apps/fountain/lib/fountain/agents/agent.ex)
- [apps/fountain/lib/fountain/conversations.ex](https://github.com/managoat/fountain/blob/0c8443b6/apps/fountain/lib/fountain/conversations.ex)
- [apps/fountain/priv/repo/migrations/20260729000001_add_allowed_vault_ids_to_agents.exs](https://github.com/managoat/fountain/blob/0c8443b6/apps/fountain/priv/repo/migrations/20260729000001_add_allowed_vault_ids_to_agents.exs)

## Status (2026-09-15)

The vault half is done. #2119 (merged 2026-09-13) added the generated `vault_access` column (`all_tenant_vaults` | `allowlist`) to agents and agent versions, migration `20260913180000`, an API round-trip test, and the upgrade note in `docs/guides/operate/upgrade.md`; `Conversations.check_vault_allowed` delegates to `Agent.vault_allowed?/2`, which fails closed, so the legacy nil branch is gone. The decision comment below records why this shape and not the alternatives.

Decision 6 of #2176 (2026-09-14) widened this issue to the two sibling allowlists that still carry the nil-means-any shape on `main`. That is the open work.

## Removal criteria: vaults (done, #2119)

- [x] Choose and document explicit policy representation for unrestricted tenant vault access versus a finite allowlist; preserve tenant isolation.
- [x] Migrate existing nil policies deliberately, including agent version snapshots and apply/API/SDK serialization where relevant.
- [x] Remove the implicit legacy-nil behavior after every supported record and writer uses the explicit representation.
- [x] Publish migration guidance and cover future vault creation semantics so migration does not silently broaden or narrow access.

## Removal criteria: the two sibling allowlists (open)

Same two-value policy, same behaviour-preserving migration (`nil` → all, `[]` and a populated list → allowlist), one PR per allowlist because each changes a table.

[...]

Issue #2094, opened 2026-09-13
Tracker: retire legacy forks and compatibility layers


jhgaylor:
The 2026-09-13 scan of main at `0c8443b6` found an explicit Buzz fork, a Mix backport, an upstream Git pin, and old-server, old-message, old-disk and old-public-contract compatibility paths. Track their removal as bounded changes with concrete retirement gates.

The objective is to eliminate duplicate implementations and obsolete compatibility branches. Every child records source evidence, migration requirements and validation. No priority, owner or delivery date is inferred by this tracker.

## Execution rules

- Recheck current main and linked work before implementation; #2018 has an active credential-set stack.
- Upstream-dependent issues must verify a released fix before repinning or deleting workarounds.
- Internal message retirement requires a supported deployment/peer version floor and upgrade ordering.
- Persistent data and public contracts require explicit migration/release behavior. Preserve security invariants and retained raw history.
- Do not rewrite applied migrations; retire persistent database accommodations with forward migrations.

## Scope boundary

The OpenAI-compatible API, Codex/Sprites capability adaptation, provider journaling/replay shims, forward-compatible decoding and both supported Swift products serve current integrations. They are not legacy removal targets merely because they adapt an interface. Existing #1656/#1703 cover the Codex/Sprites behavior. The unfinished protected ChatGPT broker adoption is separate from deleting the internal PlatformChatGPT facade.

## Visibility

Open repository issues and PRs were inspected before filing. The GitHub token lacks read:project, so Project membership, fields and automation could not be verified; these issues do not imply a Now/Next commitment.

## Removal checklist

- [ ] #776 — Buzz source fork: retire only after block/buzz#6088, #6101 and #6097 are included in the selected upstream release. The existing issue owns this work.
- [ ] #2095 — CI: retire the pinned Mix.Sync.Lock backport
- [x] #2096 — CI: replace the Dialyxir Git pin with a released Hex dependency
- [ ] #2097 — ACP: retire Gemini session-store consolidation after the upstream fix
- [x] #2098 — CLI: remove the pre-bulk legacy apply implementation
- [x] #2099 — Conversations: remove rolling-upgrade message compatibility handlers
- [x] #2100 — Inference: remove the PlatformChatGPT compatibility facade
- [x] #2101 — Clients: remove legacy vendor transcript parsing and stale parser references
- [ ] #2102 — Sandboxes: migrate old metadata and remove reconciliation fallbacks
[...]

PR #2119, opened 2026-09-13
refactor(agents): enforce explicit vault access policy

jhgaylor:
The legacy nullable vault allowlist is now represented by an explicit persisted policy mode and used by both conversation authorization and team vault selection.

- `nil` remains `all_tenant_vaults`, including future vaults in the same tenant.
- `[]` and finite arrays remain `allowlist`, preserving deny-all and finite access.
- Existing API/export shapes remain `allowed_vault_ids`; `vault_access` is internal and generated from the existing values.
- Historical snapshots distinguish omitted keys (`unchanged`), explicit JSON null (`all_tenant_vaults`), arrays (`allowlist`) and malformed values (`invalid`). Invalid and unsaved/inconsistent in-memory policies fail closed.
- A forward migration adds generated stored columns to agents and agent_versions. The original arrays and snapshot JSON remain authoritative for mixed-version writers.

Validation:
- Isolated `fountain_2107_test` database created and migrated.
- 10 focused vault-policy tests and 48 affected team tests passed, including old writers, future vaults, deny-all/finite policies, snapshots, API/apply round trips, malformed policies and tenant boundaries.
- `git diff --check` and formatter checks pass.
- Full `mix precommit` and GitHub CI are running on the repaired head.

Refs #2107 and #2094.


BinaryBourbon (approved):
Adversarial whole-PR review: APPROVED at 7c77c2e44554dec11bcc5a9a538f366f87bfe323 against base 9237a3a5b4c428057b51ae72eb657bd07e58bc61. No unresolved actionable findings on this revision.

Reviewed all ten changed files and traced generated-column writes, agent/version persistence and rollback, conversation start/resume/reapply authorization, team binding selection, API serialization and tenant exports. The authoritative nullable array remains compatible with old writers; the generated mode is returned after writes. Authorization still combines policy checks with tenant-scoped resource lookup. Missing snapshot keys preserve the current value, explicit null restores unrestricted tenant access, and malformed values are rejected by restoration validation. Neither generated field is accepted as a client-controlled policy or added to the public payload.

The migration rewrites both tables under exclusive locks in one bounded transaction; its documented maintenance-window requirement matches the implementation. Down removes only derived columns and requires readers compatible with the old schema. Production table-size/lock behavior has not been measured in this review.

Found and reproduced the original head's failure: the new team regression called Agents.update_agent/2 without aliasing Fountain.Agents (team_test.exs:568). Original-head focused run: 147 tests, 1 failure, matching CI partition 3. The new head adds precisely that alias. Repeating the same five suites on this reviewed head passed 147 tests, 0 failures: vault policy, conversation start, team, API/apply round trips, and docs guardrails. Exact dependencies and migrations were installed in an isolated local checkout/database; git diff --check passed.

Current-head CI run 34755652465 has passed all six test partitions (including the previously failing partition 3), static analysis, docs, release/API and most SDK checks; macOS Swift and coverage were still pending at the final check. Secret scan 34755652439 passed. This is a standalone adversarial review, not completion of the Review Loop service's full verification recipe; no production migration or deployment was performed.

PR #2121, opened 2026-09-13
fix(agents): tolerate homes removed before deletion fence

### LINKED ISSUE #2048: Lifecycle fence makes a missing home row fail agent deletion hard
Found during the re-review of #1972 (`codex/p2-1767-forced-home-fence`), which has since landed. Non-blocking for that PR; worth its own ticket.

## What

`Agents.delete_agent` now propagates `{:error, :not_found}` from the teardown fence when the agent's home sandbox row has already gone. Two call sites surface it to a user:

- `apps/fountain/lib/fountain_web/controllers/agent_controller.ex:181`
- `apps/fountain/lib/fountain_web/live/agents_live/index.ex:83`

Before the fence, a home row that had disappeared was simply nothing to tear down and the agent deleted cleanly. Now the deletion fails.

## Why it matters

A row missing at that moment is not a reason to refuse the delete — it is the state the delete wants to reach. The user is left with an agent they cannot remove, and retrying does not help because the row stays missing.

## Suggested shape

Treat `{:error, :not_found}` from the fence as success on the deletion path specifically, the same way `Lifecycle.destroy/5` already tolerates an already-terminal row. A fence refusal that means "there is nothing here" and one that means "someone else is using this" should not be the same answer to a caller trying to delete.

## Not established

I have not checked how reachable this is in production — whether a home row can actually vanish while its agent still exists. If it cannot, this is latent and the fix is cheap insurance rather than a bug anyone will hit.

Refs #1767, #1972.

jhgaylor:
Agent deletion now succeeds when a persistent home disappears after the home query and before its teardown fence. The agent-specific loop skips only `:not_found`, without counting the missing home as destroyed; other fence errors still stop deletion, and direct home teardown retains its missing-row error.

The regression deletes the actual row immediately after the listing query, then runs the real fence. It fails with `{:error, :not_found}` before the fix and passes afterward without provider I/O or a teardown audit event.

Validation:
- 134 tests passed across teardown fence, forced home fence, agents, sandbox mode, sandbox reset, and admin broker LiveView suites.
- `mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict`, and `git diff --check` passed.
- Used an isolated worktree/dependency build and dedicated local PostgreSQL database. Full `mix precommit`/CI verification remains pending; local toolchain was Elixir 1.19.5 / OTP 28.

Closes #2048.


BinaryBourbon (approved):
Reviewed commit `044c9296e476f16d710a57fa8c77eff12b903672`.

Reviewed the entire two-file diff at `044c9296e476f16d710a57fa8c77eff12b903672` against `95cdc9285bbf2723a14250c35b4b87335e24d6cb`, using the base contributor and review instructions. No actionable medium-or-higher correctness, security, compatibility, or simplicity findings.

Traced `Agents.delete_agent/2` through the home query, teardown fence, retirement, and agent-row deletion. The new continuation catches only a missing home and preserves the count and all other fence errors. Direct teardown keeps its error contract. Inspected both added regressions and the existing transaction/audit/fence tests: the telemetry hook deletes the real row after listing, isolates by process, and checks absence of provider calls and teardown events.

Validation in this review: exact live-head comparison and `git diff --check` passed; source, callers, and regression assertions inspected. I did not rerun ExUnit or the full precommit suite; prior PR validation is not presented as my own execution evidence.

PR #2122, opened 2026-09-13
fix(conversations): handle retirement with other validation errors

### LINKED ISSUE #2039: Retirement races match the changeset error list exactly, in four places
## What

The #1766 campaign handles a sandbox retiring mid-flight by matching the changeset `update_sandbox/2` rejects with:

```elixir
{:error, %Ecto.Changeset{errors: [status: {"sandbox is retired", []}]}} ->
```

That literal appears four times — three in `conversation_server.ex` (#1943, #1944, #1945) and one in `lifecycle.ex` (#1946) — and the #1767 chain adds more of the same shape.

## Why it is fragile

The pattern only matches when the retirement error is the *sole* error on the changeset. `prevent_sandbox_revival/1` adds it to whatever `Sandbox.changeset(current, attrs)` already produced, so one more validation firing beside it — a new field constraint, a stricter status enum — makes `errors` a two-element list and every one of these sites falls through to the path it was written to avoid:

- `do_fresh_provision/6` and the wake path raise `MatchError` and take the conversation down;
- the provision-completion site fails the conversation and releases a session a replacement may own;
- `park/4` raises inside the reclaim path.

No test would catch it. Each PR's own negative test (`"an unrelated ready-write error still fails provisioning"`) asserts a *different* single error, so it passes either way.

## Suggested shape

A predicate beside the error's producer in `Fountain.Conversations`, plus a wrapper the call sites use:

```elixir
def sandbox_retired?(%Ecto.Changeset{errors: errors}),
  do: errors[:status] == {"sandbox is retired", []}

def sandbox_retired?(_), do: false

@spec claim_sandbox(Sandbox.t(), map()) :: {:ok, Sandbox.t()} | :retired | {:error, term()}
def claim_sandbox(sandbox, attrs) do
  case update_sandbox(sandbox, attrs) do
    {:ok, updated} -> {:ok, updated}
    {:error, error} -> if sandbox_retired?(error), do: :retired, else: {:error, error}
  end
end
```

Call sites then match the atom `:retired`, which is shorter than the pattern it replaces and robust to a second error appearing. `{:error, :sandbox_reset_pending}` passes through unchanged, so `park/4` keeps its separate clause.

## Not doing it now

The #1766 chain carries 22 approved stack:1767 PRs on top of it, so rewriting those four branches to swap the call sites would force a rebase of the whole campaign. This is a follow-up for after it lands — the conversion is mechanical once the file has stopped moving.

Found while reviewing stack:1766 for merge readiness. Refs #1766, #1767, #2032.

jhgaylor:
Retirement races during provisioning, wake and park now recognize the retirement error even when the changeset also contains other validation errors. A shared predicate scans every status error, and `claim_sandbox/2` converts that rejection to `:retired` while preserving unrelated errors and `:sandbox_reset_pending`.

Regression coverage exercises real retirement plus invalid-mode validation at all four lifecycle writes, preserving replacement conversations, sandboxes and broker credentials. It also covers multiple status errors in either order, ordinary success, unrelated validation failures and reset fences.

Validation after merging current main: 108 focused lifecycle/inference tests passed; the 33-test broker suite passed again after moving its retirement pause past inference reservation; compilation with warnings as errors, formatting and strict Credo passed. Full precommit and CI validation remain for the draft PR. Existing unused-variable warnings in the broker test file were unchanged.

Closes #2039.


BinaryBourbon (approved):
Reviewed commit `3f93c76cefe0f6007f14712e872ce0c13167ae26`.

Reviewed the complete diff: `conversations.ex`, `conversation_server.ex`, `lifecycle.ex`, and all four changed lifecycle/sandbox test files. Traced the four claim writes through the locked persisted-row validation, retirement/reset fencing, provider cleanup, prepared broker credentials, and usage-event emission. The retirement predicate recognizes additional validation errors without weakening unrelated failures or changing the underlying update API. The regressions exercise real retired rows and preserve replacement resources.

No actionable medium-or-higher correctness, tenant-isolation, compatibility, or maintainability defects found. CI run [34759872309](https://github.com/managoat/fountain/actions/runs/34759872309) has successful six test partitions, static analysis, coverage, release/API checks and CI required; Detect secrets also succeeded. This was an independent source review and CI inspection; I did not rerun the tests locally.

PR #2125, opened 2026-09-13
fix(conversations): announce reused turns only after acceptance

### LINKED ISSUE #1924: A refused connection reuse publishes turn|started twice, so the turn renders as never finishing
Found while reviewing #1913 (the #1667 fix). Same defect, different path, and #1913's gate does not cover it.

## The shape

`Fountain.Conversations.Connection.resume/7` publishes `turn`/`started` before it prompts:

```elixir
Output.publish_stage(conversation_id, "turn", "started", %{
  turn_id: turn.id, turn_number: turn.turn_number,
  mode: "continue", connection: "reused"
})
```

When the idle peer then refuses the reuse, `ConversationServer.resume_acp_connection/5` drops the connection and starts over on the **same turn row**:

```elixir
{:error, _reason} ->
  state = drop_connection(state, "peer_refused_reuse")
  run_fresh_turn(state, conv, turn, prompt, TurnMachine.agent_for(conv), images, true)
```

`run_fresh_turn/7` publishes `turn`/`started` too. #1913 gated that publish on `is_nil(state.turn_session_retry)`, but this path never sets the flag — it is not a session restart — so a second `started` goes out for one turn.

## Why it matters

It is the failure #1913 exists to fix, on the same surface. `managoat/demos`, `apps/fountain-conversations`, `src/lib/blocks.ts`: the timeline pairs a stage `started` with its terminal partner, and collapses a repeated `started` only when `top.children.length === 0`. Events land in between, so the dedup does not fire — the second `started` opens a **nested** `turn` section whose outer half is closed by the end-of-stream sweep with `ended: null`. It renders as a turn that never finishes.

Present since #817, and untested: `peer_refused_reuse` appears twice in `lib/` and **zero times** in `test/`.

## The fix is probably a different question

#1913's gate asks "was this turn restarted after a lost session". The honest question for both paths is **"has this row already announced itself"** — a per-turn-row latch on the `started` publish, cleared where the row is. That covers this path, the #1667 path, and whatever the third one turns out to be.

Worth confirming against the other `publish_stage(_, "turn", "started", _)` call sites before settling on a shape.

## Not blocking #1913

#1913 is correct on the path it addresses and its new assertion pins that sequence. This is the sibling it uncovered.

jhgaylor:
A live ACP peer can refuse the next prompt, causing a fresh connection to retry the same turn row. Both paths previously published `turn/started`, leaving a nested unfinished turn in clients that pair stage events.

Publish the reused-turn start only after the peer accepts the prompt. On refusal, the fresh launch publishes the row's sole start; the conversation actor processes queued peer output after this call returns, preserving start-before-output ordering. Current session-loss recovery stays on its initialized peer and retains its existing single-start behavior.

Validation: the new regression fails on the original code with duplicate starts and passes with the fix. It covers refusal, fresh fallback, a later reused turn, and `started → output → done` ordering. Connection, ACP server, turn-machine, and bounded-lifecycle suites passed (228 tests); pinned Elixir 1.19.2 / OTP 28.3 compilation with warnings as errors, repository formatting, strict Credo (906 files), and `git diff --check` passed. Tests used an isolated local database. Full precommit and CI remain pending.

Closes #1924.


BinaryBourbon (approved):
Reviewed commit `dc7e96b280a3545597a58eef21c09ba349b2786b`.

Reviewed the complete diff in `connection.ex`, `connection_test.exs`, and `conversation_server_acp_test.exs`. Traced Peer.prompt acceptance/refusal through the actor, fresh launch, same-session recovery, output persistence, and subsequent reused turns. Peer notifications queue on the calling actor, so publishing the accepted start before returning still orders it before output; refusal leaves the sole start to the fresh launch of the existing row. The regression checks the refusal path, row reuse, subsequent success and durable event ordering.

No actionable medium-or-higher correctness, tenant-isolation, compatibility, or maintainability defects found. CI run [34759822278](https://github.com/managoat/fountain/actions/runs/34759822278) has successful six test partitions, static analysis, coverage, release/API checks and CI required; Detect secrets also succeeded. This was an independent source review and CI inspection; I did not rerun the tests locally.

PR #2126, opened 2026-09-13
fix(principals): check owner eligibility during credential renewal

### LINKED ISSUE #2029: A suspended or unverified owner can renew a claimed principal's credential
## Symptom

`Principals.renew_owned_credential/3` mints a live `principal`-scoped key for an owner whose account is suspended or whose email was never verified. `claim/4` refuses that same account.

## Mechanism (`apps/fountain/lib/fountain/principals.ex`, as of #1951)

`claim/4` gates on `check_claimer_eligible/1`:

```elixir
defp check_claimer_eligible(%User{} = claimer) do
  cond do
    claimer.principal -> {:error, :ineligible}
    is_nil(claimer.email_verified_at) -> {:error, :ineligible}
    not is_nil(claimer.suspended_at) -> {:error, :ineligible}
    Fountain.Billing.check_spend(claimer.id) != :ok -> {:error, :ineligible}
    true -> :ok
  end
end
```

`renew_owned_credential/3` calls none of it. It goes straight to the ownership subquery and mints.

Two of those four checks should not be there on renewal, and that part is right:

- `check_spend/1` is correctly skipped. `billing_subject_id/1` routes principal spend to the owner's ledger, so the balance gate still bites at spend time. Refusing renewal on a zero balance would strand a machine behind a credential the owner cannot replace, which is the failure #1688 exists to fix.
- `claimer.principal` cannot be true here, since ownership comes from the `principal_owners` row.

The other two are a different kind of check. Suspension is an operator action, not a balance, and the minted key authenticates **as the principal** — which is not itself suspended — so a suspended owner gets working credentials for compute on their own ledger. Unverified is the same shape.

## Not a regression

The API keys page already lets a suspended owner mint an ordinary key for themselves, so this widens an existing hole rather than opening one. Raised as a follow-up during review of #1951, not as a blocker; both that PR and #1952 were approved.

## Fix

Either extract the suspension/verification half of `check_claimer_eligible/1` and call it from `renew_owned_credential/3`, or decide deliberately that renewal is a recovery path that outlives suspension and say so in a comment — the current silence reads as an oversight either way. If the first, the wider question is whether a suspended account should be able to mint any API key.

---
Found reviewing #1951 (stack:1688, tracker #1688). The code described here is not on `main` yet; it lands with that stack.

jhgaylor:
Credential renewal now rejects a claimed principal's suspended or unverified owner with `:ineligible` before revoking or minting keys. It reloads and locks the owner inside the renewal transaction so a pending suspension must commit before eligibility is decided. Ownership failures remain `:not_found`, and verified, unsuspended owners can still renew at zero balance.

The owner lock precedes the claim lock to preserve user-before-dependent-row ordering. Revocation and minting remain atomic, runtime callback keys stay untouched, and existing audit writes still happen after commit. Claim and renewal share the suspension/verification predicate; claim retains its separate spending check.

Validation:
- Both suspended/unverified regression cases failed on the previous implementation and pass with the fix, preserving existing credentials and emitting no audit events.
- 159 tests passed across principals, principal claim replay/concurrency, audit guardrails, and API-key LiveView suites. Includes real concurrent suspension/renewal serialization, zero-balance renewal, ownership isolation, failed-mint rollback, and renewal/claim-replay serialization.
- Compile with warnings as errors, formatting, strict Credo, and diff checks passed on base `95cdc928` after applying its migrations to the dedicated local test database.
- Full precommit/CI verification remains pending. Local toolchain: Elixir 1.19.5 / OTP 28. The existing unused `conn` warning in API-key LiveView tests remains unchanged.

Closes #2029.


BinaryBourbon (approved):
Reviewed commit `070031e839ad95c847316cbd2311ff543b0cb1bd`.

Reviewed the entire three-file diff against base `95cdc928`, plus the renewal LiveView caller, suspension update, claim/replay path, credential revocation/minting, and existing ownership/rollback regressions. The current owner is read under a share lock before the claimed-principal row lock; ownership remains tenant scoped, ineligibility returns before any key mutation, and audit writes remain after the transaction. The shared eligibility predicate preserves claim’s separate spending requirement and allows zero-balance renewal.

Validation: independently traced the concurrent suspension test through the actual row lock and checked both suspended/unverified rejection assertions and caller error handling. Confirmed the remote head matches `070031e839ad95c847316cbd2311ff543b0cb1bd`. This was a source review; I did not rerun the Elixir suite or full precommit, and the PR’s reported test results are not independent execution evidence from this review.

Issue #1716, opened 2026-09-07
fountain-fixture runtime: disabling it leaves fixture agents un-updatable, and it ships in the public OpenAPI enums


jhgaylor:
#1644's deterministic ACP fixture is gated by `DEPLOYED_ACP_FIXTURE_ENABLED` + `DEPLOYED_ACP_FIXTURE_USER_ID` (`config/runtime.exs:45-49`), checked in `agent.ex:62-66,114-129`, `RuntimeDispatch` and `prepare_sandbox`; a tenant gets 422. Fine.

Edges:
- Disabling it later fails every changeset on an existing fixture agent (`validate_inclusion`), so those agents cannot be edited or deleted through the API.
- `fountain-fixture` sits in the public runtime enum (`schemas.ex:114,348,718`, `contract.json:7837`) and therefore in every SDK's type.
- `deployed_acp_fixture_test.exs:97-102` (the SHA test) is tautological, and the file is missing from `scripts/test-timings.tsv`.

---
Found in the 2026-09-07 review of the week on `main` (cb93a7ea).

PR #2128, opened 2026-09-13
fix(agents): allow maintenance of disabled fixture agents

jhgaylor:
Disabling the deployed ACP fixture no longer prevents its owner from editing an existing fixture agent. The changeset permits retaining the persisted fixture and its owner while runtime dispatch and sandbox preparation continue to enforce the live fixture gate. New fixture creation, switching an ordinary agent to the disabled fixture, changing its owner, and unsupported fixture configuration remain denied.

The API regression reproduced the previous 422 and now passes. Separate tests establish that deletion already worked and remains tenant-scoped. The fixture installation test now checks the actual installed bytes against a pinned SHA-256 value, and the fixture/API regression files have measured timing entries.

Validation: 92 focused tests passed across fixture, disabled-fixture API, agent changeset/controller, and schema-enum guardrail suites. Compile with warnings as errors, formatting, strict Credo, and diff checks passed. Full precommit/CI remains pending; local toolchain was Elixir 1.19.5 / OTP 28 with an isolated database.

Refs #1716. This does not close the public-enum portion: the current request enums are also the CastAndValidate admission contract for the documented opt-in fixture API, and response enums describe persisted fixture agents even after disabling the runtime. Removing that value alone would break the test-deployment API and response validation. Separating the internal fixture interface from the public contract remains follow-up design work.


BinaryBourbon (approved):
Reviewed commit `be4310f390343d6470bc2660bb30c631f348ab5b`.

Reviewed all four changed files at `be4310f390343d6470bc2660bb30c631f348ab5b` against `95cdc9285bbf2723a14250c35b4b87335e24d6cb`, using base instructions. No actionable medium-or-higher findings.

Followed create/update changesets, tenant-scoped API update/delete, snapshot writes, RuntimeDispatch, and fixture installation. The maintenance exception requires a loaded fixture with its persisted owner, while new creation, ordinary-runtime conversion, unsupported fixture fields, and actual runtime admission remain gated. The owner and second-tenant API regressions exercise the relevant boundary; the installation test checks actual written bytes against a fixed digest. Both timing additions correspond to real test files. The documented public-enum separation remains a follow-up and is not claimed as fixed here.

Validation in this review: exact live-head comparison and `git diff --check` passed; inspected every changed test and relevant callers. I did not rerun ExUnit or full precommit; the PR author's test results are not my execution evidence.

PR #2129, opened 2026-09-13
fix(conversations): refuse fenced prompts before configuration reload

### LINKED ISSUE #2009: A fenced conversation whose configuration was reapplied answers 200 queued, then does nothing
Found during an adversarial review of the #1732 bounded-execution stack (#1749). **Not introduced by that stack** — it is inherited from the reply-before-continue structure already on main — but it is a hole in the stack's stated promise that a caller of `send_prompt` hears `{:error, :execution_fenced}`.

## The scenario

A conversation has **both** an unresolved bounded execution **and** a configuration reapply committed since its server last read `configuration_revision`.

1. `handle_call({:send_prompt, ...})` preflights pass; `kick_turn/4` calls `TurnMachine.open/5`, which returns `:configuration_changed` — the stale revision is checked before the fence.
2. `kick_turn/4` returns `{:noreply, %{state | handle: nil}, {:continue, {:reapply_prompt, ...}}}`.
3. `replying_ok/1` turns that into `{:reply, :ok, state, continuation}` — **the caller is told `:ok` here**, before the continuation runs.
4. The continuation reaches `handle_continue(:provision, state)`, which calls `ExecutionGuard._unsafe_interrupt(conv.id)`, gets `{:ok, {:bounded, _}}`, and returns `{:stop, :normal, state}`. The server terminates.

## What the caller sees

`200 {"status": "queued"}` from the conversations controller. No turn is ever opened, no `:execution_fenced` is returned, and the server stops. The prompt is silently lost.

If provisioning had succeeded rather than stopping, the same prompt would then reach `kick_turn/4` inside `handle_cast({:initial_prompt, ...})`, where a fence refusal has no caller to answer and becomes a `Logger.info` — same outcome by a different route.

## Why it happens

The reply is sent *with* the continuation, so any refusal discovered after that point has nobody to tell. Anything that can fail in `handle_continue` after a `{:reply, :ok, _, {:continue, _}}` has the same shape of problem; the fence is just the case that makes it visible.

## Repro sketch

Register a bounded execution that owes a remote stop (claim the spawn, bind an identity, then fail the turn — the journal lands on `ready`), bump the conversation's `configuration_revision` so the server's copy is stale, then `POST /api/conversations/:id/prompts`. Expect `409`/`422` naming the fence; observe `200 queued`.

## Notes

- Reachable only where bounded execution is active, which today needs an operator-set ceiling — see #1864 for why that is not currently settable in production. So this is not live for tenants yet, which is why it is filed rather than held.
- The related gap where `:execution_fenced` reached callers as a bare 422 with no message is fixed separately in #2008.

https://claude.ai/code/session_01R9reKpmXUUJf4eUMULogie

jhgaylor:
When a configuration reapply leaves a conversation actor on an older revision while its previous bounded execution still owes a remote stop, prompt admission returned `configuration_changed`. The actor acknowledged the prompt as queued before its reload continuation discovered the execution fence and stopped, silently losing the prompt.

Check the unresolved journal before returning the stale-revision reload decision, under admission's existing parent lock. The caller now receives `execution_fenced` synchronously. Unfenced reloads and normal admission retain their existing behavior.

Validation: the real-actor regression fails on original main (`:ok` instead of the fence error), then passes with this fix. Bounded lifecycle, conversation reapply, and server refresh suites pass (53 tests), including a missed-refresh prompt that successfully reloads the unfenced configuration. Compilation with warnings as errors passed on pinned Elixir 1.19.2 / OTP 28.3 with an isolated database. Repository formatting, strict Credo (919 files), and `git diff --check` passed. Full precommit and CI remain pending.

Closes #2009.


BinaryBourbon (approved):
Reviewed commit `e085d7e9dfd5485176aca913c75b217b2906dbce`.

Reviewed the complete diff in `conversations.ex` and `bounded_lifecycle_test.exs`, plus the admission lock order, ExecutionGuard open-state predicate, TurnMachine error mapping and actor configuration-reload continuation. The earlier journal check uses the already-locked parent, returns the existing execution_fenced refusal before prompt acknowledgment, and leaves the normal revision reload and final admission fence intact. The regression verifies no turn insertion, no actor replacement and no loss of the existing handle/journal.

No actionable medium-or-higher correctness, tenant-isolation, compatibility, or maintainability defects found. CI run [34760335962](https://github.com/managoat/fountain/actions/runs/34760335962) reports six successful test partitions, static analysis, release/API checks and CI required; Detect secrets also succeeded. This was an independent source review and CI inspection; I did not rerun tests locally.

PR #2131, opened 2026-09-13
fix(sdk): forward sandbox API access on conversation creation

### LINKED ISSUE #1711: sandbox_api_access: "none" silently reverts to "owner" on a channel re-create that omits the field, and only the TypeScript SDK can send it
`conversations.ex:1798-1803` (#1628): when a client re-creates a conversation on a channel and omits `sandbox_api_access`, the value defaults to `owner` rather than refusing or inheriting `none`. The CHANGELOG says `none` "refuses channel resumes with a different setting"; an omitted field is not a different setting, so it slips through, and the application that chose `none` for mutually untrusted work gets a sandbox holding the owner's callback credential.

Also: the field is claimed by the TypeScript SDK only; Python, Elixir and Swift cannot send it. Either treat an omitted field on a channel re-create as "inherit", or 422 it; and claim the field in the other three manifests.

---

jhgaylor:
Python, Elixir, and both Swift clients now send an explicitly selected `sandbox_api_access` when creating a conversation. Omitting it leaves the field absent, so resumed channels inherit their existing policy. Swift's typed request/response models expose the policy, and all three SDK manifests claim the field and its supported values.

The server already preserves a channel's `none` policy when the request omits the field. A new API regression confirms that it resumes the same conversation and sandbox, retains `none`, and issues no callback key. No server behavior change is needed. Python and Elixir patch releases 0.3.1 are prepared with matching runtime/user-agent versions and changelogs because their public run helpers change; this draft does not merge or publish them.

Validation:
- Python regression reproduced the missing argument before the fix; all 50 Python tests and contract verification pass.
- All 59 Elixir SDK tests pass, including contract and explicit/omitted option coverage.
- Both Swift SDK suites pass: 96 tests, including wire serialization, typed encoding/decoding, contracts, and existing conformance checks. Strict Swift formatting passes.
- 15 server policy/callback tests pass; root compile with warnings as errors, formatting, strict Credo, and diff checks pass.
- Rebased onto `c77c381f` and reran server/static checks plus all three SDK contract verifiers. Full precommit/CI verification remains pending; local Elixir was 1.19.5 / OTP 28.

Release validation: both 0.3.1 versions were unused on PyPI/Hex when checked; the Elixir release guard, documentation build, local Hex package build, 59 Elixir SDK tests, and 50 Python tests pass. Publishing remains a consequence of a later authorized merge under the repository release workflow.

Closes #1711.


BinaryBourbon (approved):
Reviewed commit `eb3021aafa722223501fd1148818842036bd15e0`.

Reviewed all 23 changed paths against base `c77c381`, including Elixir/Python/Swift request construction, Swift response decoding and wire-value handling, SDK manifests, versions, docs, and server regression coverage. Traced the server’s create/resume validation and callback-key behavior. Explicit `none`/`owner` values are forwarded and an omitted value stays omitted, preserving an existing channel’s policy; optional arguments preserve existing caller shapes. The Swift docs correctly spell `SandboxAPIAccess.none` where Optional.none would be ambiguous.

Independent validation on head `eb3021aafa722223501fd1148818842036bd15e0`: Python `test_sdk.py` passed all 21 tests, including the three-value request-body regression; the Python contract verifier passed (68 operations, 41 schemas, 5 enums). I inspected but did not rerun Elixir, Swift or server tests, nor full precommit. This review does not substitute for those release/CI checks.

PR #2133, opened 2026-09-13
fix(conversations): consult execution journal during actor completion

### LINKED ISSUE #2054: Completion no longer consults the execution journal, and an already-ended turn leaves the parent running

jhgaylor:
Actor completion bypassed the execution journal, allowing a successful turn write after its deadline or before its command identity was confirmed. If another writer had already ended the current turn, completion also left its conversation running.

Arbitrate actor completion under parent, journal, then turn locks. Commit the journal's authoritative outcome, reply materialization, and eligible parent idle update together. An already-ended latest turn can release its parent without rewriting its result or emitting another terminal event. Reassigned actors and superseded turns retain their existing fences; deadline events remain single; successful waiting turns retain their detached request metadata. Explicit journal refusals reach callers without partial completion writes.

Validation on pinned Elixir 1.19.2 / OTP 28.3 with an isolated database:

- Six new regression cases fail on original main, covering journal retirement, expiry, uncertain spawn, completion before spawn, and terminal-parent recovery with and without a journal.
- The full conversations directory passed: 1,518 tests, 0 failures, 2 existing skips. After the final waiting-metadata preservation, 224 focused journal, actor, waiting, and binding-race tests passed.
- Independent PostgreSQL connection races cover completion-first and reassignment-first ordering for bounded and unbounded turns, including atomic visibility of the result, reply, and parent status.
- Compilation with warnings as errors, repository formatting, strict Credo (920 files), and `git diff --check` passed.

Bounded controls remain disabled by the existing capability gate. This restores their completion contract without enabling enforcement. Full precommit and CI remain pending.

Closes #2054.


BinaryBourbon (approved):
Reviewed commit `ccf7fa2472e0539fe7e90cd0274e089697a536d4`.

Reviewed the complete diff in `conversations.ex`, `conversation_server.ex`, `execution_guard.ex`, `turn_machine.ex`, `turn_completion_binding_isolation_test.exs`, `turn_completion_journal_test.exs`, and `turn_parent_test.exs`. Traced parent/journal/turn locking, admission and reassignment, completion versus deadline/uncertain-spawn arbitration, reply materialization, detached permission metadata, bounded retirement, duplicate terminal-event suppression, startup failure and parent generation checks. The database writer and journal now commit together; declined transitions retain their state, and already-ended latest turns can release the parent without another terminal announcement.

No actionable medium-or-higher correctness, tenant-isolation, compatibility, or maintainability defects found. Independently ran the three changed test modules plus `bounded_lifecycle_test.exs` against this exact head in an isolated source/build copy and dedicated local PostgreSQL database: **68 tests, 0 failures** (pinned mise toolchain). CI run [34761003993](https://github.com/managoat/fountain/actions/runs/34761003993) also reports six successful test partitions, static analysis, release/API checks and CI required; Detect secrets succeeded. I did not independently rerun the full suite.

PR #2134, opened 2026-09-13
fix(auth): expose existing connection management capability

### LINKED ISSUE #1922: GET /api/auth/me reports one Connections boolean, so a client that hides connect also hides revoke
Residue of #1693, which split the Connections gate into `enabled_for?/1` (may *add* a credential path) and `manageable_for?/1` (may list, revoke and delete what exists). The server now keeps the revoke and delete doors open for a tenant whose `connections` flag is off but whose credentials are still being brokered.

`GET /api/auth/me` did not get the same split. It still reports a single boolean:

```elixir
connections_enabled: Fountain.Connections.enabled_for?(user.id)
```

A client that hides its whole Connections surface on that field — which is the obvious way to read it — hides **revoke** along with **connect**. That is the exact bug #1693 fixed on the server, relocated to the client: the tenant sees no way to turn off a credential that is still in use.

Nothing is broken today, because the field means what it always meant and the console reads the context directly rather than the API. It becomes wrong as soon as a client trusts it for anything but the connect button.

## What it needs

A second field, so the two questions a client has to ask are both answerable. Shape is the open part — `connections_manageable` alongside the existing boolean is the smallest thing; replacing both with an object is tidier and a bigger break.

Left out of #1693 on purpose: a new response field pulls in the OpenAPI spec, the SDK contract check and four SDK regenerations, which is disproportionate to the fix it rode along with. It wants its own PR and the seven API-route gates that go with one.

jhgaylor:
`GET /api/auth/me` now reports `connections_manageable` separately from `connections_enabled`. Clients can keep list/revoke/delete controls visible after the Connections rollout flag is disabled, while continuing to hide controls that add credential paths. Both fields are false without a configured broker; `connections_enabled` retains its existing behavior.

The additive field is described in OpenAPI and the API manual, projected into the SDK contract and generated TypeScript types, and exposed by Swift's typed `AuthMe`. All four SDK manifests claim both capabilities. The shared auth conformance scenario verifies that disabled creation and enabled management survive every client's response handling.

Validation:
- Three capability assertions failed before the server change; 66 focused API, rollout, schema, and docs tests pass after it.
- TypeScript: 115 tests, typecheck, build, generated-type freshness, contract, and 28 conformance cases pass.
- Python: 49 tests and contract verification pass. Elixir SDK: 56 tests pass. Swift: 93 tests and strict formatting pass.
- Root compile with warnings as errors, formatting, strict Credo, SDK-contract freshness, conformance lint, and diff checks pass.
- Rebased onto `b17fd240` and rechecked server/docs/contract behavior. Full precommit/CI remains pending. The standalone docs-style advisory reports three pre-existing findings outside the added paragraph (`docs/api.md:128`, `docs/catalog/runtimes/codex.md:62`, `docs/concepts/environment.md:73`).

Release intent: `sdk-no-release` applies because the TypeScript change is generated API-contract exposure only, and Python/Elixir already preserve the response map unchanged. No package release or package version change is requested. Swift's optional typed fields remain compatible with responses from older servers.

Closes #1922.


BinaryBourbon (approved):
Reviewed commit `01fdd38f4ede11230bb0482f54363e9b380237e1`.

Reviewed all 14 changed paths against base `b17fd24`, including the auth response, actual connection/provider/binding management gates, schema and committed contract, all four manifests, Swift decoding, generated TypeScript, shared scenario and docs. The additional boolean exposes the existing management predicate without changing authorization. The rollout-off/broker-on case remains distinct from creation permission; Swift fields are optional for older servers, and map-based clients retain the new fields.

Independent validation on head `01fdd38f4ede11230bb0482f54363e9b380237e1`: conformance lint passed all 28 scenarios; Python’s conformance adapter passed 28 tests; Python contract verification passed (68 operations, 42 schemas, 4 enums); `git diff --check` passed. I inspected but did not rerun the Fountain, Elixir SDK, Swift or TypeScript suites or regenerate OpenAPI during this review. Full CI remains separate.

PR #2136, opened 2026-09-13
fix(sandboxes): add confirmed admin recovery for pending resets

### LINKED ISSUE #1894: A failed provider delete strands a sandbox and its quota slot with no way back

jhgaylor:
A failed reset already retries in the background, but the accepted admin recovery path in #1894 remained missing: the existing Reap action releases capacity without confirming provider deletion. Add **Retry reset** for fenced persistent sandboxes in `ready` or `suspended`.

The action rechecks the administrator's current role, verified email, suspension and session version before the unscoped lookup and provider call. It uses the existing retry finalizer with an explicit provider probe: a definitive `not_found` confirms retirement; an existing machine gets another delete attempt. Uncertain probes, failed deletes and disabled providers retain the fence and quota reservation. The existing finalizer preserves single-winner retirement accounting and notifications. The admin audit records the operator, tenant, sandbox and result; no unconfirmed release path is added.

Update the sandbox manual and API conflict message to point to automatic retries and the admin action. The existing 409/error field contract is unchanged. The separate pending/starting reset restriction requested in the issue is already documented in the changelog.

Validation: 231 distinct focused reset, admin UI, retry worker, audit, docs and fallback tests passed (the final docs wording was checked again with 43 passing tests). These include provider-lost machines, successful delete, uncertain probe/delete, disabled providers, retained quota, ineligible and malformed requests, and four forms of access revocation after mount. Compilation with warnings as errors, formatting and strict Credo passed. Independent source review found no actionable issues. The prose checker reports only three existing warnings in untouched docs/api.md:128, docs/catalog/runtimes/codex.md:62 and docs/concepts/environment.md:73. Full precommit and broad CI remain for this draft.

Closes #1894.


BinaryBourbon (approved):
Reviewed commit `ec45253bbaf12e236853738fa70ff936bd96641e`.

Reviewed the complete seven-file PR diff against the merge base of supplied base `0d24bb3` and head `ec45253bbaf12e236853738fa70ff936bd96641e`, including the admin event, reset context, LiveView, fallback response, both test changes and docs. Traced the existing reset worker, provider get/destroy contract, provider enablement, row-lock finalization, quota/usage transition, transcript completion and audit handling. The new action refreshes administrator eligibility before the unscoped lookup, validates IDs, and re-reads the pending owned row. Only definitive absence or successful deletion retires it; provider uncertainty keeps its fence and reservation, and concurrent completion still uses the existing finalizer election.

Validation: source inspection of the successful/missing/uncertain/disabled-provider, stale-authority, invalid-row, repeat-event and suspended-home regressions and their production paths; `git diff --check` passed; remote head matched the reviewed commit. I did not rerun the Fountain Elixir suite or full precommit. The review covers the PR’s merge-base diff, not unrelated two-dot differences caused by branch divergence, and makes no claim about future merged-tree validation.

PR #2137, opened 2026-09-13
fix(conversations): interrupt turns before destroying machines

### LINKED ISSUE #1707: Terminating a conversation mid-turn on the destroy branch orphans the turn instead of interrupting it, so it is never priced
`conversation_server.ex:1613-1650` (`:terminate_conv`): the kept-sandbox branch calls `interrupt_turn` first; the destroy branch calls only `drop_connection` (which closes autonomous turns) and stops `:normal` with `current_turn` still set. `terminate/2` (`:2207-2220`) then orphans it with `orphaned_at`, and both `CreditPricer` and `SandboxUsage` exclude orphaned rows.

Not a regression of #1473 (before it the turn stayed `running` and was also never priced), but it now formalises "terminate mid-turn = free turn". Interrupt in both branches.

Related: an orphaned turn later completed by a server on a partitioned node goes through `TurnMachine.finish` and `_unsafe_update_turn`, which overwrites status to `completed` and leaves `orphaned_at` set, so a turn that in fact completed is unbilled. Documented trade-off; noting the billing consequence.

---

jhgaylor:
Terminating a conversation while destroying its machine previously stopped the actor with a current turn still attached. Normal-stop orphan bookkeeping then excluded that otherwise known interrupted interval from usage and credit pricing. Interrupt the turn before dropping the connection and destroying the machine, matching the existing shared/home-machine path.

Seven real-actor regressions cover destroyed, shared and home machines with bounded/unbounded turns, exact SandboxUsage attribution, one idempotent CreditPricer debit, unresolved journal fences, and a reassignment during EOF that preserves both the successor and replacement machine. The original destroy path fails the new `orphaned_at == nil` assertion.

Validation: 97 focused termination, binding, turn-parent, bounded-lifecycle and billing tests passed; formatting, warnings-as-errors compilation, strict Credo (929 files) and diff checks passed. Provider behavior is stubbed; bounded cases use a synthetic journal identity and do not claim remote termination confirmation. Full precommit and live provider acceptance were not run locally.

The related late-orphan-completion note in the issue describes an older writer: current completion already requires a running turn and preserves terminal orphan outcomes. Conservative exclusion of truly orphaned intervals remains intentional because their actual end time is unknown. This PR fixes explicit termination, without changing that billing policy. It is independent of #2133's journal-aware completion changes.

Closes #1707


BinaryBourbon (approved):
Reviewed commit `83d8e67cb1e1a3bc353e2a50b158e882c592832e`.

Reviewed the complete diff in `conversation_server.ex` and `ee/test/fountain/conversations/termination_billing_test.exs`. Traced admission fencing, both interruption branches, kept/shared-machine behavior, connection closure, provider destruction, final binding validation and normal-stop orphaning. Also traced the interval through SandboxUsage and CreditPricer, including the orphan exclusion and debit idempotency key. Interrupting first preserves an ended, billable turn and clears actor state before normal shutdown; bounded retirement retains the remote-stop obligation. The seven regressions cover bounded/unbounded destroyed/shared/home cases and a reassignment during teardown.

No actionable medium-or-higher correctness, tenant-isolation, compatibility, or maintainability defects found. CI run [34761548979](https://github.com/managoat/fountain/actions/runs/34761548979) reports six successful test partitions, static analysis, release/API checks and CI required; Detect secrets also succeeded. This was an independent source review and CI inspection; I did not rerun this PR's tests locally.

Issue #1715, opened 2026-09-07
Tracker: sandbox files lifecycle coordination and provider no-wake guarantee


jhgaylor:
## Current scope — reassessed 2026-09-17

This is an umbrella for the remaining sandbox-files lifecycle and provider-contract work, not one implementation PR. Assessment: static inspection of main at `ae736fe0eff2aed7ac07675041b09185002c5425`, the pinned provider adapter, current provider documentation and live issue/PR history. No new race reproduction or tests were run during this assessment.

### Completed

- [x] **Symlink confinement:** #2139 merged on 2026-09-14. File/directory targets are checked against physical allowed roots, with real-shell regressions. The original lexical-confinement defect is addressed. This is physical path validation, not descriptor-relative protection against adversarial filesystem mutation between resolution and access; that limitation is documented and is not being folded into the remaining lifecycle fix.

### Remaining work

- [ ] #2394 — **Coordinate file-read admission with Fountain-controlled park/destroy.** One focused implementation PR, estimated medium once its admission contract is explicit. All four operations (list, read, diff, status) must participate in the existing machine-owner/lease protocol, with fresh eligibility checks, bounded cleanup and deterministic concurrency regressions.
- [ ] #2395 — **Decide the no-wake guarantee under autonomous provider sleep.** A Fountain lease can coordinate Fountain's lifecycle operations; it cannot prevent Sprites from sleeping independently and waking when exec is called. Decide whether to require an atomic provider capability or explicitly narrow ADR 0039's promise. No contract change is authorized merely by this split.

## Pickup order

1. Re-read #2386's merged state and #2344's current decisions. **Recommend landing #2394 after #2386** (ADR 0058 stage 9a), because it makes `destroying` durable intent that readers must refuse even after its lease expires. #2386 was open at this assessment; do not treat that snapshot as its future state.
2. Investigation and deterministic reproduction for #2394 can start now. The sequencing recommendation is not a declared hard blocker, and there is no need to wait for all of #2344 or its later stage 9b cleanup. Reuse the existing ownership protocol; another status check or a mailbox alone is insufficient.
3. #2395 can be investigated independently. The lifecycle fix must state its narrower guarantee and must not claim to prove that provider exec never wakes an autonomously sleeping VM.
4. Before closing this umbrella, both children need explicit disposition. If the decision retains strict provider no-wake, add and complete the necessary capability/adapter/validation follow-ups here; deciding the contract alone does not implement it.

[...]

PR #2139, opened 2026-09-13
fix(sandbox-files): confine symlink targets to physical roots

jhgaylor:
Sandbox file listing and reading followed symlinks after checking only the requested path lexically. Both now resolve the target physically and compare it with the physical allowed roots in the provider's execution namespace. An escaped target returns the existing `path_outside_sandbox` error before emitting file bytes or directory entries. The shared root guard also serves git discovery, preserving mapped runner homes and sandbox-facing repository paths.

Real-shell regressions cover outside directory/file symlinks, in-root symlinks, symlinked runner homes, either allowed root, trailing newlines, missing/dangling paths, wrong kinds and unreadable files. The two escape regressions fail on the original implementation by returning outside bytes and filenames. Context/controller checks verify mapped root arguments and error translation.

Validation: all 85 focused shell, context and API controller tests passed; formatting, warnings-as-errors compilation, strict Credo (930 files), and diff checks passed. Shell behavior was exercised locally on macOS; Linux and the wider suite remain CI validation. No live provider acceptance or full precommit was run locally.

This is physical path validation, not descriptor-relative confinement against an adversarial symlink swap between resolution and opening. That limit is stated in the module documentation.

The separate ready-then-exec parking race remains open. `SandboxFiles.ready?/1` checks the supplied row before provider exec, while lifecycle suspension/parking runs independently. The pinned Sprites adapter's suspend call is a no-op and its exec path has no conditional no-wake option. Sprites documents that commands [wake sleeping machines](https://docs.sprites.dev/quickstart/), and its [exec API](https://sprites.dev/api/sprites/exec) exposes no such condition. Another row check cannot make remote execution atomic; a complete remedy needs coordinated read/park admission and a provider capability or an explicit contract change for autonomous sleep. This exceeds a second small local PR and is retained under the issue's `big bug` label.

Refs #1715


BinaryBourbon (approved):
Reviewed commit `32fb6cde8e1f7a6c72bea41c5843903c71249857`.

Reviewed all four changed files at `32fb6cde8e1f7a6c72bea41c5843903c71249857` against `0d24bb38ba333269d493d3f7f4fcf9f70d646c1c`, using base review instructions and ADR 0039. No actionable medium-or-higher findings in this change.

Traced the full-scope, tenant-scoped controller fetch through lexical resolution, provider host-path mapping, all four shell operations, exit translation, and redaction. Physical targets are compared against physical roots before list entries or file bytes are emitted; positional argv remains data, the sentinel preserves trailing newlines, and git returns the sandbox-facing root. Inspected every changed regression and retained parser/redaction behavior. The documented concurrent symlink-swap and ready-versus-park limitations are real remaining scope limits, not claims this patch resolves them.

Independent validation: executed 10 diagnostics against the exact raw shell fragments extracted from this head, using local macOS bash and disposable paths. Passed directory and file escapes (exit 9, empty stdout), allowed symlinks, mapped homes, either root, trailing-newline files/directories, and git status/diff root mapping. Exact live-head comparison and `git diff --check` passed. I did not run the full ExUnit suite, Linux/provider acceptance, or precommit, and do not claim those from the author's evidence.

PR #2140, opened 2026-09-13
fix(agents): snapshot references removed by source deletion

### LINKED ISSUE #1052: Environment deletion changes agent config without writing a version

jhgaylor:
Deleting an environment silently nilified an agent's default environment without creating a version; environment and vault allowlists also retained deleted IDs. The newest snapshot could therefore disagree with the live agent and attribute the cleanup to a later edit.

Delete each source in the same transaction as scoped reference cleanup and a new snapshot for every affected agent. Lock the source before scanning fresh agent rows in ID order, and use `RETURNING` with the existing snapshot writer. Environment-changing edits acquire the existing source advisory lock before the agent row to preserve lock order. Audits happen after the transaction, and existing home teardown stays outside it. A deletion or snapshot failure rolls back all reference and version changes.

Only references to the deleted resource change. Empty allowlists remain empty, unrestricted `nil` remains unrestricted, database-derived vault access remains consistent, historical configurations with disabled runtimes can still be cleaned, and older snapshots are untouched. Update the agent manual and ADR 0029's resolved gap. This cleanup covers references present at deletion; the existing allowlist validators still permit a later stale-ID submission. This PR does not introduce FK-like integrity or change authorization policy for those arrays.

Validation: 323 focused tests passed across new transaction/concurrency regressions, agents, environments, vaults, home teardown, source write ordering, version history, audit and docs. The new tests verify both edit/deletion orders, concurrent new and changed environment FKs, real database failures at deletion and snapshot insertion, untouched cotenants and historical snapshots, and nil/empty access semantics. Compilation with warnings as errors, formatting and strict Credo passed. The ADR index regenerated without changes; independent source review found no actionable issues. The prose checker reports only three existing warnings in untouched docs/api.md:128, docs/catalog/runtimes/codex.md:62 and docs/concepts/environment.md:73. Full precommit and broad CI remain for this draft.

Closes #1052.


BinaryBourbon (approved):
Reviewed commit `d77498133acfb845eabd4609e106e0d278b0e44b`.

Reviewed the entire seven-file diff at `d77498133acfb845eabd4609e106e0d278b0e44b` against `0d24bb38ba333269d493d3f7f4fcf9f70d646c1c`, using base contributor/review instructions and ADRs 0013/0029. No actionable medium-or-higher findings.

Followed environment/vault deletion from tenant-scoped API/LiveView callers and the Buzz cleanup caller, through home retirement, the source-lock wrapper, the fresh ordered agent scan, RETURNING writes, version insertion, source deletion, and post-transaction audit. Checked update/rollback/manifest callers and the source-before-agent lock ordering for environment FK changes. Cleanup preserves nil versus empty allowlists, unrelated configuration, generated vault-access values and historical snapshots. The docs match this scope; later stale allowlist submissions remain explicitly outside it.

Inspected every new regression, including real lock waits in both edit/deletion orders, concurrent new and changed FKs, actual deletion/version constraint failures, cotenant isolation, historical runtimes, snapshot freshness and audit attribution. Exact live-head comparison and `git diff --check` passed. This is a completed source review; I did not rerun the database concurrency suite or full precommit and do not represent the PR author's runs as independent execution evidence.

PR #2141, opened 2026-09-13
fix(sandboxes): enforce unique live provider machine names

jhgaylor:
Two live sandbox rows could claim the same provider and machine name before any provider identity was recorded. Add a concurrent partial unique index on `(provider, sprite_name)` for pending, starting, ready and suspended rows, and return a `machine_name` changeset error on collisions. Failed and terminated rows retain the existing historical-name reuse behavior.

Production preflight ran on 2026-09-13 in an explicitly read-only transaction, returning aggregate counts only: 0 duplicate live name groups, 0 duplicate provider-instance groups, 0 live names outside their tenant namespace, and 381 live rows without a recorded provider instance ID. This is evidence at inspection time; the migration remains fail-closed if a collision appears before deployment. It retains the configured advisory migration lock while building the index concurrently.

This is the database portion of #1919. Provider adoption on 409 still needs tenant ownership validation; the current Sprites SDK neither accepts creation labels nor retains returned labels, so that work spans the SDK, sandbox adapter and Fountain provisioning. The issue remains open with `big bug` for that coordinated work. No production schema or provider state was changed.

Validation:
- Six of nine new database regressions fail without the index; all nine pass after the migration. They cover every live state, cross-account collisions, independent providers, terminal history and direct-write revival.
- Full `mix precommit` passed with the pinned toolchain and a dedicated local database: 6 doctests, 6,202 core tests, 144 Buzz tests and 33 Support tests; zero failures (core: 2 existing skips, 7 excluded). Compile, formatting, Credo, Sobelow and Dialyzer passed.
- Independent review found no issues with the lifecycle predicate, constraint propagation or migration-lock configuration.

Refs #1919.


BinaryBourbon (approved):
Reviewed commit `d8d3c583533907e0da685a034d46bb2ca65cd5e6`.

Reviewed all three changed files at `d8d3c583533907e0da685a034d46bb2ca65cd5e6` against `0d24bb38ba333269d493d3f7f4fcf9f70d646c1c`, using the base contributor and review instructions. No actionable medium-or-higher findings in the database uniqueness change.

Checked the physical `sprite_name` mapping, non-null name/provider/status columns, all six lifecycle states, create/update changesets, initial reservation rollback/error propagation, provider-scoped name generation (including runner identity), and the configured PostgreSQL advisory migration lock. The partial index retains every live state including suspended, permits terminal history, and enforces collisions across accounts while allowing distinct providers. Its named constraint is surfaced on the application field. Concurrent DDL runs outside the migration transaction while retaining the supported advisory lock. Inspected all nine database regressions, including revival onto an occupied name.

Validation in this review: exact live-head comparison and `git diff --check` passed; full migration/schema/caller/test source inspection. I did not execute the migration or ExUnit suite and made no production database/provider calls. The author's read-only preflight and precommit results are not my own validation. The separately documented provider-adoption ownership work remains unresolved and is not represented as fixed by this approval.

PR #2142, opened 2026-09-14
fix(config): delete the dead PHOENIX_REQUEST_LOG switch and widen the reference guard

jhgaylor:
## What

An audit of `.env.example` against every environment variable the app reads. The count in the prompt (16 of 103) was the file's uncommented lines; it names 133 variables, every one has a reader, and `docs/configuration.md` is the full reference by design. One real gap came out of it.

`PHOENIX_REQUEST_LOG` was the only variable the app reads that no reference documents. It was also wired to nothing: `config/prod.exs` is evaluated when the release is built, so a value set on a deployment never arrived (#210 introduced it with a comment saying to set it at the deployment). Phoenix's default is the value the read produced, so the switch is deleted rather than moved.

`config_reference_test.exs` only scanned `runtime.exs`, which is how this slipped past it. The direct `System.get_env` scan now covers every config file and the lib tree of every app, `ee/` included. The quoted-literal pass stays on `runtime.exs`, the only file where every such literal is a variable name. The reverse, compose, render and fly checks ask the same union, so `RELEASE_NAME` no longer needs an exemption.

## Verified

The guard was broken two ways and failed naming the variable each time: a stray read appended to `config/prod.exs`, and a stray read in a lib module. `docs_test.exs` and `mix format --check-formatted` pass.

## Follow-ups

Three further PRs retire dead switches and freeze tunables nobody sets; they are independent of this one.


BinaryBourbon (approved):
Reviewed both changed files at 4fde1206d79373a1fd279c0cf817a7d71ea7047a against a1512cb9e3c298c3a742b46d878c646b500dabff. No blocking correctness, security, or compatibility regression found. Phoenix 1.8.13 defaults `:logger` to `true` and installs its request telemetry handlers, so deleting the override preserves normal production request logging.

Validation: the standalone configuration-reference suite passed (11 tests, 0 failures). Four isolated mutation probes added an undocumented environment read to prod config, core lib, Buzz lib, and ee lib; each correctly failed and named the variable. Current-revision CI is green, including all six test partitions, static analysis, release boot, CI required, and secret scanning. Local diagnostics used Elixir 1.19.5/OTP 28; I did not rerun the full application suite locally. There is no existing Review Loop App review on this revision; its skipped dispatch is not approval evidence.

Nonblocking, pre-existing documentation cleanup: `apps/fountain/priv/help/runbook.md:221` still tells operators to set `PHOENIX_REQUEST_LOG=false` on the Deployment. That instruction already failed with the old build-time read; remove it now that the switch is retired. This is not a newly introduced runtime regression.

PR #2144, opened 2026-09-14
refactor(credits): retire contact rent and message pricing

jhgaylor:
Part 1 of 3 of #2143 (remove Team comms). Team comms had zero users, and rent was the only thing that made the Team context reach into `ee/`. This is the ee half.

## What goes

- `Fountain.Credits.Rent`, `Workers.CreditRentCollector` and its 06:47 cron entry, the rent-due email and its worker clause.
- The credit pricer's message pass (`burn_message`), and the `burn_rent` / `burn_message` ledger reasons. Rows already written keep their reason; nothing writes new ones.
- The finance panel's contact and message lines, `Finance.rate_card/1`'s four contact rates, the AgentMail/AgentPhone rows of `Billing.Reconciliation` and `ProviderInvoice.providers/0`.
- The four customer prices on `Credits.price_card/0` and the marketing copy that quoted them.
- `Team.Comms` no longer gates provisioning on a month of rent or charges one; `Team.Contact` loses its two rent stamps (the columns go with the table in part 3).

## Env vars no longer read

`CREDIT_NUMBER_CENTS`, `CREDIT_INBOX_CENTS`, `CREDIT_EMAIL_MESSAGE_CENTS`, `CREDIT_SMS_MESSAGE_CENTS`, `AGENTMAIL_INBOX_CENTS`, `AGENTPHONE_NUMBER_CENTS`, `AGENTMAIL_MESSAGE_CENTS`, `AGENTPHONE_MESSAGE_CENTS`. Documented under Upgrade notes in the CHANGELOG; ADR 0030 records the retirement of decision 4 and the comms rows of decision 3 (index regenerated, `okf validate` clean).

## Verification

`mix precommit` green on a fresh database: credo no issues, dialyzer 0 errors, sobelow, prod release assembles, 6175 core tests + 144 buzz + 33 support, 0 failures.

## Stack

1. **this PR** — ee: rent and message pricing
2. #2145 — SDKs drop the comms surface (`sdk-no-release`)
3. #2146 — core: delete Team comms, drop the tables, regenerate the contract, bump the SDKs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017e1HGs98jna3gzSj2vG62q

Issue #2143, opened 2026-09-14
Remove Team comms: teammate SMS numbers, inboxes, rent and message pricing


jhgaylor:
Team comms (a teammate's AgentMail inbox and AgentPhone number, the `fountain-comms` MCP tools, inbound texts as prompts, rent from the credit balance and per-message pricing) has zero users, and rent is the only thing that makes the Team context reach into `ee/`. Rip it out.

Landed as a stack of three PRs, each based on the one below:

1. **ee side** — retire `Credits.Rent`, the rent collector, the rent-due email, the pricer's message pass, the finance panel's contact and message lines and the four customer prices. Team stops calling ee.
2. **SDKs** — drop `commsStatus` / `comms_status` and the `TeamCommsStatus` / `TeammateContact` types from all four SDKs and their contract manifests, ahead of the wire change (`sdk-no-release`; the bumps ride in 3).
3. **core** — delete `Fountain.Team.Comms` and its subtree, `Team.Contact`, `Team.CommsMessage`, the two controllers, the routes, the MCP injection, the `team_comms` flag, the AgentMail/AgentPhone config, the docs page, and drop the `team_contacts` and `comms_messages` tables; regenerate the SDK contract and bump the SDKs.

Outside this repo, once 3 lands:

- `managoat/demos` `apps/fountain-team`: the contact dialog calls `GET /api/team/comms` and `POST /api/team/:agent_id/contact`, which will 404. Remove it.
- PostHog flag `team_comms` (project `jakegaylor.com`, id 831574) and the `AGENTMAIL_API_KEY`, `AGENTPHONE_API_KEY`, `AGENTPHONE_WEBHOOK_SECRET` secrets in `fountain/prod` Infisical become inert; delete when convenient.
- AgentPhone's master webhook still points at `fountain.inevitable.fyi/api/webhooks/agentphone`; the endpoint will answer 404. Delete the webhook and release any purchased number/inbox with the providers directly.

jhgaylor:
Follow-ups outside this repo, done 2026-09-14 after #2146 deployed (prod on the merge commit, `team_contacts` and `comms_messages` dropped, migration 20260914090000 applied):

- **AgentPhone**: the master webhook pointing at `/api/webhooks/agentphone` deleted; the two Fountain-provisioned numbers (`+17282331011`, `+14065594534`) released and their personas deleted. Only the account's own number and persona remain.
[...]

PR #2145, opened 2026-09-14
refactor(sdk): drop the team comms surface from the four SDKs

jhgaylor:
Part 2 of 3 of #2143 (remove Team comms), based on #2144. The client half, landed before the wire drops the operations so no SDK names an endpoint the server has stopped serving.

## What goes

- `team.commsStatus()` (TypeScript, both Swift products) and `comms_status` (Python, Elixir), with the `TeamCommsStatus` and `TeammateContact` types and `Teammate.contact` in FountainKit.
- `GET /api/team/comms` leaves the four contract manifests and joins the omissions allowlist beside the contact operations, until part 3 stops serving them.
- The Elixir README and `docs/elixir-sdk.md` no longer list `comms_status`.

Changelogs carry an Unreleased entry each; the version bumps ride in part 3, which regenerates the TypeScript types from the trimmed spec (`sdk-no-release` here).

## Verification

TypeScript: 115 tests, contract check ok, `tsc --noEmit` clean. Python: 49 tests (unittest, as CI runs it), contract check ok. Elixir: 56 tests. Swift: `swift test --filter ContractTests` passes on a clean build.

## Stack

1. #2144 — ee: rent and message pricing
2. **this PR** — SDKs drop the comms surface
3. #2146 — core: delete Team comms, drop the tables, regenerate the contract, bump the SDKs

https://claude.ai/code/session_017e1HGs98jna3gzSj2vG62q


BinaryBourbon (approved):
Reviewed as its own patch (`git diff stack/team-comms-1-rent...stack/team-comms-2-sdks`). All four SDKs drop the call and the types cleanly, and a grep of the tip for `comms_status|commsStatus|TeamCommsStatus|TeammateContact|team/comms` across sdk/ and docs/ finds only the changelogs, the still-published docs page that PR 3 deletes, and the manifest schema claims noted below. The intermediate state is right for the order: the SDKs stop calling `GET /api/team/comms` before the server stops serving it, `omissions.json` takes the operation over so `build.py --check` has a home for it, and Swift's `Teammate` losing `contact` is safe against a server that still sends the key because `Decodable` ignores unknown keys. `docs/elixir-sdk.md` and the Elixir README no longer list `comms_status`. `sdk-no-release` is correct here since no version moves.

Nits, no action required:

- The four manifests still claim `TeamCommsStatusResponse` (and `TeamCommsStatus`/`TeammateContact` in `typescript.json`) as schemas the client reads, which is no longer true after this patch. The check passes because the contract still has them, and PR 3 removes the claims with the schemas, so nothing is wrong on main at any point. Just noting the manifest lags the code by one stage.
- The Elixir changelog's bare `comms_status/1` is not a fully-qualified reference, so ExDoc will not try to autolink it; PR 3's rewording keeps that property.

BinaryBourbon (approved):
Re-reviewed at `fc7f6774` as its own patch (`git diff stack/team-comms-1-rent stack/team-comms-2-sdks`). The patch is the one approved before, rebased: all four SDKs drop the call and the types, `omissions.json` takes `GET /api/team/comms` over so `build.py --check` stays satisfied while the server still serves it, the manifests drop the operation, and the docs and READMEs no longer list `comms_status`. A grep of the tip for `comms_status|commsStatus|TeamCommsStatus|TeammateContact|team/comms` across `sdk/` and `docs/` finds only the changelogs, the still-published docs page #2146 deletes, and the one-stage manifest schema claims already noted. CI is green at this sha. `sdk-no-release` is still right: no version moves here.

One thing to know before queueing, not a defect in the patch: `main` moved under the stack. #2131 released Elixir and Python SDK 0.3.1 today, with a `## [0.3.1]` heading at the top of `sdk/elixir/CHANGELOG.md` and `sdk/python/CHANGELOG.md`, and this PR's `## [Unreleased]` heading is added at the same spot, so `git merge-tree` reports a content conflict in both files against today's `main`. GitHub only checks a PR against its base, so this shows as mergeable now; the moment #2144 lands and GitHub retargets this to `main`, it will flip to conflicting, get no CI, and the queue will refuse it. Rebase onto `main` once #2144 is in, with the Unreleased section above 0.3.1. Nothing else in the patch touches a file `main` changed.

Issue #1487, opened 2026-09-03
Delete the Agent Vault client, and take ADR 0019 off Proposed


jhgaylor:
Part of #1488 (native-broker flip tracker). The last step: once production runs the native broker and has stayed up, the vendor client stops being a supported path and comes out.

**Do not open this before the flip has run and settled.** While `BROKER_URL` is a live rollback, `Fountain.Broker.AgentVault` is the rollback.

## Delete

- [ ] `Fountain.Broker.AgentVault` (438 lines) and its dispatch arms in `Fountain.Broker` (`impl/1`, the `:agent_vault` clauses in `delete_vault/1` and `list_vaults/0`)
- [ ] `BROKER_URL`, `BROKER_TOKEN` and their validation in `config/runtime.exs`, including the "both set" boot check, which has nothing left to conflict with
- [ ] `BROKER_LOG_RETENTION_HOURS` **only if** the native egress log issue did not adopt it — it should, so expect this one to stay
- [ ] the reaper's vault branch in `Fountain.Workers.BrokerVaultReaper` (`Broker.backend()` `:agent_vault` arm, `vaults/0`, `delete/2`), leaving the native session sweep. Rename the worker; "vault reaper" stops meaning anything
- [ ] `test/support/broker_test_helpers.ex` and `broker_test.exs`'s 60 Agent Vault tests
- [ ] the Agent Vault row from the backend table in `Fountain.Broker`'s moduledoc, and `backend/0` itself if only one backend remains
- [ ] `marketing_html.ex:1496` still tells self-hosters to "Set `BROKER_URL` and `BROKER_TOKEN`" — that instruction becomes wrong at the flip, not at the deletion. Fix it in whichever PR lands first.

## ADR 0019 and the docs

- [ ] ADR 0019 stops saying the broker runs "beside Agent Vault until the flip" (the 2026-09-02 amendment, §8 and §11)
- [ ] its status moves off **Proposed** — the amendment says it stays Proposed "until the flip has happened and the vendor client is deleted", which is exactly this PR
- [ ] refresh `verified` / `stale_after` frontmatter and regenerate the index (`scripts/decisions-index.sh`), in the same PR, per CLAUDE.md
- [ ] `CLAUDE.md`'s `managoat_broker` entry says fountain "runs it beside the Agent Vault client" — drop the beside
- [ ] self-hosting docs: the broker section describes a vendor service to deploy; after this it is a listener in the app

## Acceptance

[...]

PR #2150, opened 2026-09-14
chore(config): retire five dead switches

jhgaylor:
Tier 1 of the env-var cleanup: switches for things that no longer exist or that a standard variable already covers. Independent of #2142.

- **`HONEYCOMB_ENDPOINT`, `HONEYCOMB_API_KEY`**: vendor shortcuts for `OTEL_EXPORTER_OTLP_ENDPOINT` and `OTEL_EXPORTER_OTLP_HEADERS` (`x-honeycomb-team=<key>`). The exporter config is now written only when an endpoint is set. `otel_config_test.exs` gains the standard-spelling case and a check that headers alone do not turn export on. Nothing in production sets either shortcut.
- **`DNS_CLUSTER_QUERY`**: a second discovery mechanism beside `CLUSTER_DNS_QUERY`, whose docs row said to leave it unset. The `DNSCluster` child, the `dns_cluster` dependency and the row go.
- **`BROKER_URL`, `BROKER_TOKEN`**: boot guards for the Agent Vault backend removed in #1487. A leftover value is now ignored; the CHANGELOG upgrade note says so, and the two guard tests go with them.

Left alone on purpose: `BROKER_TENANTS=*` (a deliberate assertion from #2058, set in production), the `DEPLOYED_ACP_FIXTURE_*` pair (a whole test-deployment feature, not a switch), and the three `FOUNTAIN_EXECUTION_*` variables (the #1732 stack landed on 2026-09-12; the `.env.example` note saying nothing enforces them is what is stale).

Verified: `otel_config_test`, `runtime_config_test`, `config_reference_test` and `docs_test` pass (65 tests); `mix format --check-formatted` at both levels.


BinaryBourbon (changes requested):
Adversarial review of `f8c99c8cd94d08acb96cdaaf8479b590817c8038`: changes requested for one operability defect in the configuration reference.

**[P2] Remove the remaining instructions to enable telemetry with retired Honeycomb variables.** In [docs/configuration.md lines 397–402](https://github.com/managoat/fountain/blob/f8c99c8cd94d08acb96cdaaf8479b590817c8038/docs/configuration.md#L397), the endpoint's default is still `HONEYCOMB_ENDPOINT`, and the following paragraph still says either `HONEYCOMB_ENDPOINT` or `HONEYCOMB_API_KEY` enables export. This PR stops reading both. An operator following that published setup now boots successfully with `traces_exporter: :none` and silently loses all exported traces; I reproduced that with both Honeycomb variables set and the standard endpoint absent. Update the default to unset and the paragraph to require `OTEL_EXPORTER_OTLP_ENDPOINT`, with authentication in `OTEL_EXPORTER_OTLP_HEADERS`. The obsolete shortcut claim in `config/runtime.exs` lines 1419–1420 should be corrected in the same fix. The changelog's correct migration note does not correct the contradictory live reference.

Reviewed all 10 changed paths, including dependency/lockfile removal, supervision, clustering, broker guards, tests and upgrade guidance. No other blocking defect found. The retained `CLUSTER_DNS_QUERY` topology still uses `Cluster.Strategy.DNSPoll`; an isolated config probe confirmed that topology and intentional retired-broker-variable behavior.

Validation: 25 OTel/runtime-config tests passed on the exact head using standalone ExUnit and the head's source modules. A broader standalone config-reference attempt passed 10 of 11 cases; its remaining case could not run because the partial harness did not load `FountainWeb.MarketingHTML` (environment limitation, not a PR finding). Current GitHub CI required, all six build/test partitions, static analysis, release checks and secret scanning are successful. Full application/production integration was not rerun locally. No Review Loop App summary or existing review was present; its dispatch check was skipped.

jhgaylor:
Addressed in f151f7b0: the `OTEL_EXPORTER_OTLP_ENDPOINT` row's default is now unset, the `OTEL_EXPORTER_OTLP_HEADERS` row names the Honeycomb header form, the paragraph says only the endpoint turns export on (headers alone do not), and the stale runtime.exs comment about `HONEYCOMB_API_KEY` is gone. The only remaining mentions of the shortcuts are the two comments that say they are gone. docs, config-reference and OTel suites pass.

BinaryBourbon (approved):
Fresh adversarial re-review of `c7abc9471dcae4783f47f1bd64d4f126d3e4b03b`: approved. Two independent reviewers examined all 10 changed paths for correctness, security, and operator compatibility; neither found a blocking defect.

The previous P2 finding is resolved. The configuration reference now requires `OTEL_EXPORTER_OTLP_ENDPOINT`, describes authentication through `OTEL_EXPORTER_OTLP_HEADERS`, and no longer advertises the retired Honeycomb shortcuts. The obsolete runtime shortcut instructions are removed.

The retained DNSPoll clustering path remains configured, and `BROKER_TENANTS=*` still refuses startup without a broker listener. The dependency removal, retired broker-variable behavior, tests, and upgrade guidance were reviewed.

Validation: each reviewer independently ran the exact-head OTel/runtime-config suites (25 tests, 0 failures per run), with additional isolated probes for retired-variable behavior, native broker configuration, and retained clustering. GitHub reports 27 successful checks and 3 conditional skips on this head, including CI required, all six test partitions, static analysis, coverage, release boot, and secret scanning. Full application/integration suites were not repeated locally; standalone harnesses loaded only the modules needed for these config tests. This is a manual review, not a completed Review Loop App run; its dispatch is skipped.

Optional cleanup: `config/runtime.exs:230–231` still compares the tenant-list boot guard to the removed `BROKER_URL` guard. That stale internal comment does not affect behavior and does not block approval.

PR #2151, opened 2026-09-14
chore(config): fix the principal bounds and ChatGPT timings as constants

jhgaylor:
Tier 2 of the env-var cleanup, part one: eight tunables nobody sets stop being environment variables. Independent of #2142 and #2150.

- **Six `PRINCIPAL_*` bounds** (ADR 0044): `config.exs` already held the same values and `runtime.exs` re-read them from the environment. The runtime block goes; `config.exs` now says the numbers are fixed. `principals_ceiling_test.exs` still overrides them through the application env.
- **`PLATFORM_CHATGPT_REFRESH_MARGIN_SECONDS`, `PLATFORM_CHATGPT_KEEPALIVE_DAYS`** (ADR 0047): the two accessor functions return their values directly, and the rationale from the runtime comment (the margin must exceed the longest turn; six days sits inside the auth server's eight-day window) moves into their docstrings.
- Docs rows, `.env.example` blocks and the empty table the ChatGPT rows left behind are removed. A CHANGELOG upgrade note says a value set for one is now ignored.

Nothing in production or any deploy surface sets any of the eight.

Verified: `config_reference_test`, `docs_test`, `runtime_config_test`, `principals_ceiling_test`, `chatgpt_refresh_coordination_test` and `codex_chatgpt_test` pass; `mix format --check-formatted` at both levels.


BinaryBourbon (approved):
Approved after an independent adversarial review of all seven changed files at `6cc4ea100b05276c577f4a160213b0659af2e6c5` against base `5e966ca4bd152d1232e8d5c133a9cb55e12047b6`.

No blocking correctness, security, or API compatibility defects found. I traced principal creation ceilings, TTL/grant clamping and purge retention, plus both platform and tenant ChatGPT refresh/keepalive callers. The retained values match the previous defaults, principal application-env test overrides still work, and the removed variables have no remaining runtime/deployment readers in the repository. The upgrade note explicitly documents the intentional override removal.

One nonblocking documentation follow-up: ADR 0047 decision 3 still says the refresh margin “is config” (`decisions/0047-codex-platform-chatgpt-account.md:188`). Update that sentence to describe the fixed fifteen-minute margin.

Validation: `git diff --check` passed. The current commit has 27 successful GitHub Actions checks, including all six test partitions, static analysis, release/API contract, docs, coverage and secret scanning; three conditional jobs are skipped. No Review Loop App review or summary was present, and its dispatch job was skipped.

Local execution limitation: a standalone configuration-reference diagnostic passed 10 tests and failed one because `FountainWeb.MarketingHTML` was not loaded outside Mix. I stopped the separate checkout’s cold dependency rebuild before database setup/test execution; I am not claiming a complete local suite pass. The complete test evidence above is from GitHub CI.

PR #2153, opened 2026-09-14
chore(config): fix five housekeeping tunables as constants

jhgaylor:
Tier 2 of the env-var cleanup, part two. Independent of #2142, #2150 and #2151.

Five tunables with a sane default that no deployment ever changed stop being environment variables: `BROKER_SESSION_TTL_SECONDS` (six hours), `LOG_OUTPUT_BUDGET_MB` (50), `UNVERIFIED_PRUNE_AFTER_DAYS` (30), `SECRET_EXPIRY_NOTICE_DAYS` (7) and `AGENTPHONE_BASE_URL`. Every reader already fell back to the same value in the lib, so only the `runtime.exs` reads, the docs rows, the `.env.example` and compose entries go. The tests that override the application env keys still can. A CHANGELOG entry says a value set for one is now ignored.

Kept, after checking what production and the docs actually rely on:

- `UNVERIFIED_PRUNE_EXEMPT`: production sets it.
- `AGENTMAIL_BASE_URL`: it selects the EU region host.
- `OTEL_SERVICE_NAME`: production sets it to the default, but its lines sit inside hunks #2150 rewrites; a one-line follow-up once that lands.

Verified: `config_reference_test` (including the compose and render guards), `docs_test`, `runtime_config_test`, `compose_passthrough_config_test`, the sweeper, pruner, log-budget and output suites pass (100 tests); `mix format --check-formatted` at both levels.


BinaryBourbon (approved):
Approved after an independent adversarial review of all nine changed files at 9554f2a216424f8cbfd883d071bfb7a9ba10d388. I found no actionable correctness, security, or compatibility defect beyond the explicitly documented retirement of these environment overrides.

Traced every removed runtime key to its remaining consumer: broker sessions retain 21,600 seconds, output retains 50,000,000 bytes, account pruning retains 30 days, expiry notices retain 7 days, and AgentPhone retains https://api.agentphone.ai. The application-env overrides used by tests remain available. UNVERIFIED_PRUNE_EXEMPT and AgentMail's regional host selection remain wired through. Compose, examples, public documentation, and the changelog agree with the intended change.

Validation: 50 config-reference, runtime-config, and Compose-passthrough tests passed using direct ExUnit against the isolated head checkout. A separate release-config probe passed for absent, blank, zero, malformed, and nondefault retired variables, while verifying the retained exemption and EU-host settings. The first local Compose run hit this machine's missing automatically discovered OS CA bundle; explicitly loading /etc/ssl/cert.pem resolved that environment failure without changing repository files. DB-dependent suites and full compilation were not rerun locally; current-head GitHub CI supplies that coverage. CI required, Detect secrets, all six test partitions, coverage, static analysis, release/contract, SDKs, and Compose checks are successful. No existing review or Review Loop App verdict was present; the dispatch job was skipped.

Issue #2152, opened 2026-09-14
Connections: platform connectors become extensions, the OAuth client becomes a library


jhgaylor:
Tracker for the Connections split discussed 2026-09-13/14. Connections stays in core as the way a tenant authenticates to an OAuth-based MCP service; the parts that are a product on top of it move out through the two seams the repo already has.

## Plan

1. **Provider quirks become struct data.** The two hooks `Fountain.Connections.Platform` owns (extra authorize parameters; Slack's nested token body) become fields on `Fountain.Connections.Provider`, so `Fountain.Connections.OAuth` knows no service by name. Unblocks 3 and 4.
2. **The Gmail MCP server becomes an extension** (`apps/fountain_google`, ADR 0043). The JSON-RPC tool layer, the Gmail API client and the controller move; the extension contributes its server through the existing `conversation_mcp_servers/2` callback. No new callback. Independent of 1.
3. **The OAuth client becomes a library.** `Fountain.Connections.OAuth` (authorize URL, code exchange with PKCE and `resource`, refresh, revoke, account label) moves into `managoat_mcp_auth` next to discovery and registration, as the complete client side of MCP authorization. Fountain maps its provider row onto the library's config struct.
4. **Extensions contribute connection providers.** An ADR adds the eleventh `Fountain.Extension` callback, `connection_providers/0`. The Google provider moves into `fountain_google`; Microsoft and Slack each become their own extension (`fountain_microsoft`, `fountain_slack`). `Fountain.Connections.Platform` is deleted and ADR 0033 decision 1 is amended.

## Out of scope here

The `oauth2` tenant provider kind, the `connections` rollout flag, and the SDK contract for `POST /api/connection-providers` stay as they are. The Gmail route was never in an SDK manifest.

Each step is its own PR (or a short stack), labelled `stack:<this issue>`.

jhgaylor:
Progress, 2026-09-14:

- Step 1: #2154 (root) — quirks become fields on the provider. Precommit green.
- Step 4a: #2155 (on #2154) — ADR 0054, the `connection_providers/0` callback, registry composition and boot validation.
- Step 2 (`fountain_google`, closes #1529), step 3 (`managoat_mcp_auth` client + the Fountain draft), and steps 4c/4d (`fountain_microsoft`, `fountain_slack`, stacked on #2155) are in progress on their own branches.
- Step 4b (the Google provider into `fountain_google`, `Platform` loses its builders) follows once step 2 has a branch to stack on.

[...]

PR #2154, opened 2026-09-14
refactor(connections): the platform quirks become fields on the provider

jhgaylor:
Step 1 of #2152.

`Fountain.Connections.OAuth` used to ask `Fountain.Connections.Platform` two questions about every provider, both keyed on a service's slug: which extra authorize parameters to send, and whether the token body nests the grant. A provider that comes from somewhere else (an extension, step 4) cannot rely on a slug the client recognises, and a client that names services cannot leave for a library (step 3).

## What changes

- `Fountain.Connections.Provider` gains two virtual fields: `authorize_params` (a map merged over the standard authorize-URL parameters) and `token_body_nest` (the key a provider nests the user's grant under). The platform builders set them; a tenant row has neither.
- `Fountain.Connections.Platform` loses `authorize_params/1` and `normalize_token_body/2`. Google, Microsoft and Slack carry the same values on their structs.
- `Fountain.Connections.OAuth` reads the fields and names no service. The Slack lift is a private clause keyed on `token_body_nest`.
- ADR 0033 decision 1 says where the two hooks went (body only, frontmatter untouched, so the index is unchanged). CHANGELOG line under Unreleased.

No wire change and no behaviour change: the composed authorize URLs and the lifted Slack token body are now asserted through the client (`OAuth.authorize_url/3`, `OAuth.exchange_code/3` against a `Req.Test` stub) rather than through the hooks.

## Verification

- `mix test test/fountain/connections test/fountain/connections_test.exs test/fountain_web/live/connections_live_test.exs test/fountain/conversations/egress_test.exs`: 80 tests, 0 failures.
- Revert check: with the nested-token lift broken, `the client lifts slack's authed_user grant to the top level` fails (`{:ok, grant}` vs `{:error, ...}`); restored, 12 tests, 0 failures.
- `mix compile --warnings-as-errors`, `mix credo --strict` (no issues), `mix format` from both roots. Full `mix precommit` running; result posted as a comment.

## Stack map (#2152)

1. **this PR** `stack/connections-1-quirks` (root)
2. `stack/connections-2-gmail-ext` (independent root): the Gmail MCP server becomes `apps/fountain_google`
3. the OAuth client moves into `managoat_mcp_auth`; a Fountain PR stacked on this one pins the new version
4. ADR + `connection_providers/0` callback, stacked on this one; Google, Microsoft and Slack providers become extensions


jhgaylor:
Full `mix precommit` on the branch, read from the log rather than the exit code:

- compile with warnings as errors, unused deps, format, `credo --strict` (no issues), sobelow, dialyzer (`Total errors: 0, Skipped: 0`), prod release assembled (`Release created at _build/prod/rel/fountain_server`).
- core suite: `6 doctests, 6274 tests, 0 failures, 2 skipped`.
- `scripts/test-libraries.sh`: `fountain_buzz` 144 tests, 0 failures; `fountain_support` 33 tests, 0 failures. (The first pass of these two failed on `relation "buzz_identities" does not exist`: my worktree database had been migrated from `apps/fountain`, which skips extension migrations. Migrated from the umbrella root and re-run; the failures were environmental, not the branch.)

jhgaylor:
Replaced by #2156: this branch received no workflow events for any push, reopen or synchronize (zero check suites on the head commit), while a dispatch of ci.yml on the same ref ran. Same commit, fresh branch.

PR #2155, opened 2026-09-14
feat(extensions): an extension contributes connection providers (ADR 0054)

jhgaylor:
Step 4a of #2152, stacked on #2156 (step 1).

`Fountain.Extension` gains its eleventh callback, `connection_providers/0`. An extension hands the host config-backed `Fountain.Connections.Provider` structs, the shape a platform provider has always had; the host lists them after its own in `Platform.all/0`, answers them from `get/1`, reserves their slugs against tenant providers, and drives them with the one OAuth client. The extension contributes the provider, never the flow, so no token passes through extension code.

This is the seam that lets Google, Microsoft and Slack leave core with the products behind them (steps 4b to 4d). Until they move, the host's three builders stay in `Platform` and the registry is "the host's three, then each installed extension's".

## What changes

- `Fountain.Extension`: the callback, its doc, the `[]` default and the moduledoc bullet.
- `Fountain.Extensions`: `connection_providers/0,1` isolated like the admin figures (`safe_admin` became `safe_list`), the callback in the required set, and a boot check that refuses a non-list, a non-struct entry, a `user_id`, an id that is not the slug, a slug of the wrong shape or kind, and a slug the host or an earlier extension owns. A disabled extension is not asked.
- `Fountain.Connections.Platform`: `builtin_slugs/0` for the host's three; `all/0`, `get/1` and `slugs/0` are the whole registry.
- Fixtures: `Enabled` contributes one provider (`fixture-svc`); `Disabled` contributes one that must never appear; six NOT-configured fixtures exercise the validation. Every test that enumerated the registry now sees four platform providers, on purpose: the seam is proved on the real path the console and the API take.
- ADR 0054 (Accepted, partially built: the three moves are named as not yet built, `stale_after` set), ADR 0043 decision 3 amended for the eleventh, index regenerated, `okf validate` clean for both files. CLAUDE.md's "adding a callback needs an ADR" line updated. CHANGELOG under Added.

## Verification

- `mix test test/fountain/extensions_test.exs test/fountain/extension_composition_test.exs test/fountain/connections test/fountain/connections_test.exs test/fountain_web/controllers/connection_controller_test.exs test/fountain_web/controllers/connection_provider_controller_test.exs test/fountain_web/controllers/connections_controller_test.exs test/fountain_web/live/connections_live_test.exs test/fountain/extension_guard_test.exs test/fountain_web/api_spec_test.exs`: 180 tests, 0 failures.
- Revert check: with `Platform.all/0` no longer appending extension providers and the boot check removed from `validate/2`, 8 of the new tests fail (the registry composition, the end-to-end provider test, and six validation cases); restored, 0 failures.
- `mix compile --warnings-as-errors`, `mix credo --strict` (no issues), `mix format` from both roots, `okf validate decisions` (no finding on 0043 or 0054). Full `mix precommit` running; result posted as a comment.

## Stack map (#2152)

1. #2156 `stack/connections-1-provider-quirks` (root; replaced #2154, whose branch received no workflow events): the platform quirks become fields on the provider
2. `stack/connections-2-gmail-ext` (independent root): the Gmail MCP server becomes `apps/fountain_google`
3. the OAuth client moves into `managoat_mcp_auth`, stacked on 1
4. **this PR** (4a), stacked on 1; then 4b `fountain_google` takes the Google provider, 4c `fountain_microsoft`, 4d `fountain_slack`, and `Platform` loses its builders


jhgaylor:
Full `mix precommit` on the branch, read from the log: compile with warnings as errors, format, `credo --strict`, sobelow, dialyzer (`Total errors: 0`), prod release assembled; core suite `6 doctests, 6286 tests, 0 failures, 2 skipped`; `fountain_buzz` 144 tests, 0 failures; `fountain_support` 33 tests, 0 failures.

apps/fountain/lib/fountain/connections/platform.ex
```diff
@@ -46,20 +52,38 @@ defmodule Fountain.Connections.Platform do
...
 
-  @doc "The platform provider for a slug, or nil."
+  @doc "The platform provider for a slug, the host's or an extension's, or nil."
   @spec get(String.t()) :: Provider.t() | nil
-  def get("google"), do: google()
-  def get("microsoft"), do: microsoft()
-  def get("slack"), do: slack()
+  def get(slug) when is_binary(slug) do
+    builtin(slug) ||
+      Enum.find(Fountain.Extensions.connection_providers(), &(&1.slug == slug))
```
BinaryBourbon, line 77: [P1] Handle existing connections when an extension provider disappears

This makes a previously valid platform provider disappear when its extension is disabled/removed, or when connection_providers/0 raises and safe_list/3 returns []. Its persisted connections remain active with provider_id: nil. Connections.provider_for/1 then returns nil: revoke_at_provider/1 passes that to unlock_provider/1 (connections.ex:437), so both revoke and delete raise FunctionClauseError before changing local state. The same assumption exists in do_refresh/2 (line 639), and implicit_hosts/2 dereferences nil.token_hosts (line 549), which is reached by Conversations.Egress.connection_bindings/3 during credential assembly. A connector failure can therefore prevent removal of its credentials and break conversation startup, contrary to the intended isolation.

Make missing-provider resolution an explicit supported state across these callers: allow local revoke/delete with remote revocation skipped as best effort, return a controlled unavailable error for refresh, and omit unavailable-provider tokens/bindings from sandbox assembly. Add a regression that connects an account while its extension is enabled, then makes the extension unavailable and exercises revoke/delete plus credential assembly.

apps/fountain/lib/fountain/extensions.ex
```diff
@@ -549,6 +582,90 @@ defmodule Fountain.Extensions do
...
+
+      p.user_id != nil ->
+        {:error,
+         "#{where} contributes #{inspect(p.slug)} with a user_id; an extension's " <>
+           "provider is config-backed and has none"}
+
+      p.id != p.slug ->
+        {:error, "#{where} contributes #{inspect(p.slug)} whose id is not its slug"}
+
+      p.kind not in Fountain.Connections.Provider.kinds() ->
```
BinaryBourbon, line 648: [P2] Keep config-backed MCP providers out of tenant rediscovery

Provider.kinds/0 includes "mcp", so this now explicitly accepts an extension MCP provider with user_id: nil and exposes it through Connections.get_provider/2. POST /api/connection-providers/<extension-slug>/discover accepts any %Provider{kind: "mcp"}; unlike update/delete, ConnectionProviderController.discover/2 has no tenant-row guard. It calls Connections.rediscover_provider/2, whose first operation is Crypto.load_tenant_key(p.user_id); load_tenant_key/1 only accepts a binary, so the newly supported provider produces a FunctionClauseError/500. The LiveView rediscover event has the same missing guard (even though its button is hidden for platform rows).

Reject config-backed providers at both rediscovery entry points and guard the context operation as tenant-only, matching the existing edit/delete boundary. Add an installed MCP-provider fixture assertion that discovery by slug returns the normal not-found response rather than attempting tenant-key loading.

BinaryBourbon (changes requested):
Adversarial whole-PR review of 6423cabf0fe45cf49c4879ab85273d6431cbd877 against stacked base c81c5ba085493fd94c655f37ab21374a85e0dea1.

Changes requested for two integration defects in the newly extensible provider registry:

1. Existing connections cannot be revoked/deleted safely after their extension becomes unavailable; refresh and broker binding construction also assume the provider can never disappear.
2. A permitted config-backed MCP provider is admitted to the tenant-only rediscovery endpoint, which crashes on its nil user_id.

The inline comments give the triggers, affected callers and minimal fixes. I inspected all 14 changed paths, the extension lifecycle/validation, provider resolution and persistence, OAuth client, connection management API/LiveView, broker binding path, regression assertions, fragment and ADR changes. The stack base is preserved.

Validation was left to CI as requested; no local tests were run. Current-head CI run 34812403134 has a partition-5 failure in Fountain.Conversations.TerminationClusterTest (a 5-second :peer.call timeout at termination_cluster_test.exs:9), leaving coverage skipped and CI required red. I found no causal link from this diff to that failure and am not presenting it as either finding above. This is a personal source review, not a claim that the Review Loop service verification ran.

BinaryBourbon (approved):
Adversarial whole-PR re-review of 310742dd178f625347f5550fd820298eb97459aa against stacked base f96d133489fa0f3ee9d8ffaf2c6eb544c4816411.

Approved. No remaining concrete medium-or-higher defect found. Both objections from my previous review are repaired: an unavailable extension now produces a controlled provider_unavailable result, allows local revoke/delete, and contributes no tokens or implicit hosts to new sandbox credential assembly; config-backed MCP providers are rejected by the context, API and forged LiveView rediscovery paths. I traced the new lifecycle regression assertions through the implementation, including fresh and stale grants, a retained provider and a second tenant.

Reviewed all 18 changed paths: registry composition and boot validation, callback isolation and extension lifecycle, tenant slug reservation and ownership, OAuth start/callback/refresh/revoke callers, encrypted persistence and response allowlists, broker assembly, API/LiveView callers, fixtures and regression tests, and the contributor-guide/changelog/ADR changes. The existing API shapes and stacked base are preserved.

Validation: no local test suite run. GitHub CI run 34813837596 is a successful pull_request run by jhgaylor for this exact head; all six core partitions, static analysis, coverage, release/API-contract and core-distribution checks passed. The head's secret-scan and Decisions checks also passed. The earlier TerminationClusterTest timeout occurred on the superseded head; this review does not claim a root-cause fix for that earlier failure. This is the requested personal source review, not a claim that Review Loop service verification ran.

CI evidence: https://github.com/managoat/fountain/actions/runs/34813837596

PR #2156, opened 2026-09-14
refactor(connections): the platform quirks become fields on the provider

jhgaylor:
Step 1 of #2152.

`Fountain.Connections.OAuth` used to ask `Fountain.Connections.Platform` two questions about every provider, both keyed on a service's slug: which extra authorize parameters to send, and whether the token body nests the grant. A provider that comes from somewhere else (an extension, step 4) cannot rely on a slug the client recognises, and a client that names services cannot leave for a library (step 3).

## What changes

- `Fountain.Connections.Provider` gains two virtual fields: `authorize_params` (a map merged over the standard authorize-URL parameters) and `token_body_nest` (the key a provider nests the user's grant under). The platform builders set them; a tenant row has neither.
- `Fountain.Connections.Platform` loses `authorize_params/1` and `normalize_token_body/2`. Google, Microsoft and Slack carry the same values on their structs.
- `Fountain.Connections.OAuth` reads the fields and names no service. The Slack lift is a private clause keyed on `token_body_nest`.
- ADR 0033 decision 1 says where the two hooks went (body only, frontmatter untouched, so the index is unchanged). CHANGELOG line under Unreleased.

No wire change and no behaviour change: the composed authorize URLs and the lifted Slack token body are now asserted through the client (`OAuth.authorize_url/3`, `OAuth.exchange_code/3` against a `Req.Test` stub) rather than through the hooks.

## Verification

- `mix test test/fountain/connections test/fountain/connections_test.exs test/fountain_web/live/connections_live_test.exs test/fountain/conversations/egress_test.exs`: 80 tests, 0 failures.
- Revert check: with the nested-token lift broken, `the client lifts slack's authed_user grant to the top level` fails (`{:ok, grant}` vs `{:error, ...}`); restored, 12 tests, 0 failures.
- `mix compile --warnings-as-errors`, `mix credo --strict` (no issues), `mix format` from both roots. Full `mix precommit` running; result posted as a comment.

## Stack map (#2152)

1. **this PR** `stack/connections-1-provider-quirks` (root; replaces #2154, whose branch received no workflow events)
2. `stack/connections-2-gmail-ext` (independent root): the Gmail MCP server becomes `apps/fountain_google`
3. the OAuth client moves into `managoat_mcp_auth`; a Fountain PR stacked on this one pins the new version
4. ADR + `connection_providers/0` callback, stacked on this one; Google, Microsoft and Slack providers become extensions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2157, opened 2026-09-14
refactor(connections): the OAuth client moves to managoat_mcp_auth

jhgaylor:
Step 3 of #2152, stacked on step 1 (#2156). **Draft until managoat_mcp_auth 0.2.0 is on hex** — see below.

## What moved

The OAuth 2.0 authorization-code client behind Connections is now `Managoat.McpAuth.Client` in the `managoat_mcp_auth` library (managoat/managoat_mcp_auth#4), so that library holds the whole client side of MCP authorization: RFC 9728/8414/7591 discovery and registration, the SSRF guard, and the flow itself. Step 1 made the client name no service, which is what let it leave.

`Fountain.Connections.OAuth` is a thin adapter now:

- `to_config/1` maps a `%Provider{}` (with its plaintext `client_secret` already unlocked) onto the library's `Managoat.McpAuth.Client.Config`: the RFC 8707 `resource` from `resource/1`, `trusted_urls?` and `require_refresh_token?` from `user_id: nil` (a platform provider's URLs are ours and its expiring token must come with a refresh token; a tenant provider is guarded on every fetch and takes what it gets), and the Req options and timeout from `:connections_req_options` / `:connections_timeout_ms`.
- `configured?/1`, `code_verifier/0`, `authorize_url/4`, `exchange_code/4`, `refresh/2`, `revoke/2` and `resource/1` keep their signatures and return shapes and delegate.
- The ported implementation (~330 lines) is deleted here; it lives in the library with 56 tests of its own at 100% coverage.

No wire or behaviour change. The `Req.Test` plug in `config/test.exs` keeps its name (`Fountain.Connections.OAuth`, passed through as the config's `req_options`), so no test file changed. ADR 0033, CLAUDE.md's repo layout, the CHANGELOG and the moduledocs that named the client say where it lives now.

## Hex dependency status

`apps/fountain/mix.exs` pins `{:managoat_mcp_auth, "~> 0.2.0"}`. That version is **not on hex yet**: it publishes when managoat/managoat_mcp_auth#4 merges (its `ci` and `release gate` checks are green). Until then CI here fails at `mix deps.get`. Once it is published, one follow-up commit on this branch updates `mix.lock` and the PR leaves draft. No `path:` dep is committed.

## Verification

Run locally with the dependency pointed at the library branch (`path:`), then flipped back to the hex pin for the commit:

```
cd apps/fountain && mix test test/fountain/connections test/fountain/connections_test.exs \
  test/fountain_web/controllers/connections_controller_test.exs \
  test/fountain_web/controllers/connection_controller_test.exs \
  test/fountain_web/live/connections_live_test.exs test/fountain/audit_guardrail_test.exs \
  test/fountain_web/controllers/connection_provider_controller_test.exs \
  test/fountain_web/controllers/connections_rollout_test.exs \
  test/fountain_web/controllers/gmail_mcp_controller_test.exs test/fountain/umbrella_layout_test.exs
193 tests, 0 failures

mix compile --warnings-as-errors      # clean
mix format (both roots)                # clean
mix credo --strict                     # 8631 mods/funs, found no issues.
MIX_ENV=dev mix dialyzer               # Total errors: 0, Skipped: 0, Unnecessary Skips: 0
okf validate decisions                 # "valid": true; index unchanged
```

The three `unused alias Google` warnings in test files are pre-existing on `main` (those files were not touched).

## Stack

| Step | PR |
|---|---|
| 1 — the platform quirks become fields on the provider | #2156 (`stack/connections-1-provider-quirks`; #2154 was its first head) |
| 2 — the Gmail MCP server becomes an extension | not open yet; independent of this |
| 3 — this: the client moves to `managoat_mcp_auth` | #2157 |
| 4 — extensions contribute connection providers (ADR 0054) | #2155, also on step 1 |

Lands after step 1; GitHub retargets it to `main` when #2156 merges.


apps/fountain/mix.exs
```diff
@@ -57,7 +57,7 @@ defmodule Fountain.MixProject do
       {:managoat_acp, "~> 0.4.2"},
       {:managoat_broker, "~> 0.14.0"},
       {:managoat_docs, "~> 0.1.0"},
-      {:managoat_mcp_auth, "~> 0.1.0"},
+      {:managoat_mcp_auth, "~> 0.2.0"},
```
BinaryBourbon, line 111: **[P1] Publish and lock the client dependency before this can merge.** A clean checkout cannot resolve this new requirement: the current CI logs report `managoat_mcp_auth ~> 0.2.0` does not match any version and abort before compilation/tests. The [live Hex package API](https://hex.pm/api/packages/managoat_mcp_auth) still lists 0.1.1 as latest, [upstream PR #4](https://github.com/managoat/managoat_mcp_auth/pull/4) is still open, and `mix.lock:45` still locks 0.1.0, which lacks the delegated Client modules. This blocks every server build, including production/release assembly. After the upstream 0.2.0 release exists, resolve it and commit the resulting root `mix.lock`, then let CI validate that published artifact. Keeping this draft until then is appropriate; the outstanding lockfile follow-up is required.

BinaryBourbon (changes requested):
Adversarial review of 0738c4fd against stacked base c81c5ba0: **changes requested**.

The extraction is still blocked by its dependency release and lockfile. This is acknowledged in the draft description, but remains an unresolved blocker on the current revision.

I reviewed all 10 changed files, compared the adapter with the removed implementation and the proposed upstream client/config at managoat/managoat_mcp_auth@067e6237, and traced authorization callbacks, tenant/provider lookup, secret unlocking, refresh/revocation and the existing connection tests. The config mapping preserves the platform/tenant URL-guard and refresh-token policies, PKCE/resource parameters, provider quirks and Req.Test settings in that proposed upstream source. I found no additional concrete regression in those paths.

CI evidence: [current-head run 34812408550](https://github.com/managoat/fountain/actions/runs/34812408550) fails dependency installation across all six test partitions, static analysis, release/API-contract and core-distribution jobs; these are dependency-resolution failures, not test flakes. No local test suite was run, as requested. This is a personal source review, not a Review Loop service-verification result. Once the dependency is published and the lockfile committed, CI needs to exercise the actual released artifact.

BinaryBourbon (approved):
Adversarial whole-PR review of 033d58f17f98e69a8879b2225025df903197a43a against stacked base f96d133489fa0f3ee9d8ffaf2c6eb544c4816411: approved. The dependency-publication/lockfile blocker from the previous changes-requested review is resolved.

Verified the actual released dependency: Hex reports managoat_mcp_auth 0.2.0 published at 2026-09-14 06:28:52 UTC. Inspected Client and Config extracted from the cached Hex tarball; both its inner checksum and archive SHA-256 match this PR's committed mix.lock. The public release checksum also matches. The review therefore covers the released implementation rather than only the earlier proposed upstream source.

Reviewed all 11 changed paths, compared the released client against the removed implementation, and traced callbacks, provider lookup/unlocking, refresh/revoke, API serialization and the existing connection regression tests. The adapter preserves tenant URL guarding versus trusted platform URLs, refresh-token policy, PKCE/resource parameters, client authentication, provider quirks, error/grant shapes and Fountain's Req.Test/timeout configuration. The removed undocumented dig/req helpers have no repository callers. Documentation, dependency requirements and ADR amendments match the extraction. No actionable medium-or-higher defect found.

Local verification: four isolated diagnostic tests passed using the archived PR sources and the released Client/Config, with existing compiled dependency beams: platform/MCP authorize-URL differential comparison; Slack nested exchange/refresh differential comparison; rejection of an expiring platform grant without a refresh token; and tenant localhost rejection before exchange/refresh/revoke HTTP calls. These diagnostics used HTTP stubs and no database; they are not a full clean-build or application-suite run.

GitHub Actions pull_request run 34813782920, authored and triggered by jhgaylor for this exact head, succeeded: all six test partitions, static analysis, release/API contract, coverage and CI required passed; secret scanning also passed. This supersedes the old dependency-install failure. This is an independent manual review, not a Review Loop service-verification result.

PR #2159, opened 2026-09-14
refactor(connections): the platform quirks become fields on the provider

jhgaylor:
Probe: same commit as #2156, opened without labels to find out why that PR receives no workflow events. Will be closed or completed in a minute.


jhgaylor:
Probe done: the cause is a CHANGELOG.md conflict with main, so GitHub builds no merge commit and creates no workflow runs (same class as the conflicted-PR-means-zero-CI note). Fixed by rebasing #2156.

PR #2161, opened 2026-09-14
refactor(connections): the Microsoft provider becomes the fountain_microsoft extension

jhgaylor:
ADR 0054 decision 6, first of three (#2152 step 4c): the Microsoft platform connection provider leaves core for `apps/fountain_microsoft`.

## What moved

- **`apps/fountain_microsoft`**, an AGPL umbrella app modeled on `fountain_support`: `FountainMicrosoft.Extension` (`use Fountain.Extension, id: :microsoft`) implements exactly two callbacks, `connection_providers/0` and `docs/0`. No route, no migration, no `mod:`, no cron.
- **`FountainMicrosoft.Provider`** builds the one `%Fountain.Connections.Provider{}` core used to build in `Platform.microsoft/0`, verbatim: same slug, endpoints, `userPrincipalName` label, PKCE, `authorize_params: %{"prompt" => "select_account"}`, `MICROSOFT_ACCESS_TOKEN`, `graph.microsoft.com`, and the same default scopes with the `MICROSOFT_OAUTH_SCOPES` override.
- **Config, per ADR 0054 decision 5**: `config/runtime.exs` reads `MICROSOFT_OAUTH_CLIENT_ID` / `_SECRET` / `_SCOPES` into `config :fountain_microsoft` (absent stays absent, `""` counts as absent); the test client pair in `config/test.exs` moves to that namespace. `docs/configuration.md` is unchanged and the config reference guard still passes.
- **The manual page** `docs/catalog/connections/microsoft.md` moves to `apps/fountain_microsoft/docs/` under the same slug, so `/docs/catalog/connections/microsoft` is unchanged; `FountainMicrosoft.Docs` embeds it and the nav merges it back into `Catalog`.
- **The scope-override test** moves with the provider (`scope_override_test.exs`, async: false).

## What core no longer knows

- `Fountain.Connections.Platform` loses `microsoft/0` and `@microsoft_scopes`; `@builtin` is `~w(google slack)`. `client_env_var/1` and `short_name/1` stay (convention, decision 5).
- `platform_oauth_clients` / `platform_oauth_scopes` in `config/runtime.exs` no longer name Microsoft.
- `docs/nav.yml` drops the page; `docs/catalog/connections/index.md` and `slack.md` mention Microsoft by name without a link (a core page must not link an extension's page). The connections console blurb says "Platform providers" rather than naming the three.
- Registered where every extension is: `installed_extensions` (`Code.ensure_loaded?`), the Dockerfile `COPY`, CLAUDE.md's layout block, an `extension_guard_test.exs` row (module regex `FountainMicrosoft\.` plus the old builder name `Fountain\.Connections\.Platform\.microsoft\b`, route `"/(api/)?microsoft`, priv `**/*microsoft*`), and `scripts/test-timings.tsv` loses the deleted file.
- ADR 0033 decision 1 gets an amendment note; ADR 0054's status block and description record Microsoft as built, Google and Slack as not yet; `decisions/index.md` regenerated, `okf validate` clean.

Core tests that enumerated the registry now assert the host's own in order and the contributed tail by membership: a root `mix test` loads the extension (so `Platform.all/0` has four entries there) while a run from `apps/fountain` does not (three). The refresh-token test uses Google; the two-labels test uses Google and the fixture provider `fixture-svc`.

No API change: the OpenAPI descriptions that say "Google, Microsoft, Slack" are prose and were left alone so the export stays byte-identical.

## Verification

- `apps/fountain_microsoft`: `mix test` → `28 tests, 0 failures` (registry lists it, `Platform.get("microsoft")` is the extension's struct field for field, `OAuth.configured?/1` true, authorize URL carries the picker and S256, refresh-token insistence, scope override, docs structural checks, merged-manual links, sidebar placement).
- `apps/fountain`, the connections/extension/docs/api-spec files plus `config_reference_test.exs` → `221 tests, 0 failures`.
- Guard broken on purpose (a `FountainMicrosoft.Extension` mention in `platform.ex`) → `17 tests, 1 failure` naming the line; restored.
- `mix openapi.export` raw output vs the same on `origin/stack/connections-4a-provider-callback` (detached worktree, own `deps.get`): `cmp` identical, MD5 `92ca0e2e4350518bc56eaa7ab90a619c` both sides.
- `mix compile --warnings-as-errors`, `mix format --check-formatted` (root, `apps/fountain`, `apps/fountain_microsoft`), `mix credo --strict` → `8715 mods/funs, found no issues`.
- Full `mix precommit` from the root (`DATABASE_URL` set; dialyzer `done (passed successfully)`, sobelow, `Release created at _build/prod/rel/fountain_server`): `6 doctests, 6289 tests, 0 failures, 2 skipped (7 excluded)`, exit 0.
- `scripts/test-libraries.sh`: `fountain_buzz: 144 tests, 0 failures`, `fountain_microsoft: 28 tests, 0 failures`, `fountain_support: 33 tests, 0 failures`, exit 0.

## Stack map (#2152)

#2154 step 1 (root: the quirks become fields) → #2155 step 4a (the `connection_providers/0` callback) → **this PR, 4c (Microsoft)** → 4d (Slack, stacked on this one).

#2152 step 2 (`fountain_google`) and 4b (the Google provider into it) are on a separate root.


apps/fountain/lib/fountain/connections/platform.ex
```diff
@@ -25,28 +28,21 @@ defmodule Fountain.Connections.Platform do
...
+  products a tenant actually consented to, and `<SLUG>_OAUTH_SCOPES`
+  (space-separated) lets an operator narrow or grow the default request — a
+  deployment whose Google app verification does not cover a scope simply
+  does not ask for it.
   """
 
   alias Fountain.Connections.Provider
 
-  @builtin ~w(google microsoft slack)
+  @builtin ~w(google slack)
```
BinaryBourbon, line 40: [P2] Preserve cleanup of existing Microsoft connections on core upgrades

A deployment already using BUNDLE_EXTENSIONS=false can have active Microsoft connections before this PR, because Microsoft is still built into the base. After upgrading the same distribution, this removal makes Platform.get("microsoft") return nil for those persisted rows. Connections.revoke/2 calls revoke_at_provider/1, which passes that nil to unlock_provider/1 (connections.ex:437); no clause accepts it. The console Revoke/Remove actions therefore crash, and DELETE /api/connections/:id returns an exception instead of deleting the connection. Refresh and implicit_hosts/2 have the same unchecked missing-provider assumption.

Handle an absent provider before shipping the extraction: allow local revoke/delete without a provider request and make token refresh/host lookup return a controlled unavailable result. Add a regression retaining an active Microsoft row while its extension is absent. The shared lifecycle fix requested on #2155 can supply the implementation; this PR introduces the upgrade trigger for existing Microsoft users.

BinaryBourbon (changes requested):
Adversarial whole-PR review of 594f519239df6b9576aea51c7b7121dd151d58b9 against its actual stacked base 6423cabf0fe45cf49c4879ab85273d6431cbd877.

One blocking upgrade regression: existing Microsoft connections on a core distribution become impossible to revoke/delete when this extraction removes their provider from the registry. See the inline finding. This is a concrete Microsoft upgrade trigger for the missing-provider lifecycle issue also identified on parent PR #2155; the shared repair can land there and be inherited here, with a regression covering a persisted Microsoft connection after the provider disappears.

Reviewed all 36 changed paths, including provider field/config parity, registry callers and tenant boundaries, extension discovery and bundled/core packaging, manual merging and links, API compatibility, migrated tests, documentation and changelog fragment. No other blocking defect found.

No local test suites run, as requested. Current-head CI passed partitions 1–4 and 6, static analysis, release/API-contract checks and core-distribution checks. Partition 5 failed in the unchanged TerminationClusterTest at termination_cluster_test.exs:9 (:peer.call timed out after 5000 ms), so CI required is red and coverage was skipped. That failure is recorded separately; it is not the basis for this source-review finding. CI evidence: https://github.com/managoat/fountain/actions/runs/34812488825/job/103876456695

This is the requested personal adversarial review, not a Review Loop service verification result.

BinaryBourbon (approved):
Adversarial whole-PR re-review of 55cf0355cb8452825afc4289251fc35a4a90d2b9 against its actual stacked base 310742dd178f625347f5550fd820298eb97459aa.

Approved. No remaining concrete medium-or-higher defect found. The previously reported core-upgrade cleanup defect is repaired in the inherited #2155 base. The new CoreUpgradeTest retains fresh and stale Microsoft grants while removing the extension, then checks controlled token unavailability, empty synthetic credentials and host bindings, preserved unrelated input, local revoke, and deletion. I traced those assertions through the shared implementation.

Reviewed all 37 changed paths, including the moved provider's fields and config namespace, OAuth and tenant-scoped connection callers, registry membership and slug reservation, extension installation, release inclusion/exclusion and Docker inputs, API compatibility, migrated and new tests, merged manual/navigation/links, unchanged AGPL license text, changelog and ADR changes. The Microsoft slug, endpoints, token host, environment key and OAuth behavior remain compatible in the bundled distribution.

Validation: an isolated local Elixir diagnostic compared the base builder with the new provider and confirmed complete struct equality for default, empty and explicitly overridden scopes, using the reviewed Provider schema and builder source. No local database/full test suite or Docker build was run. CI run 34813901772 is a successful pull_request run by jhgaylor for this exact head. Its log records fountain_microsoft: 29 tests, 0 failures, with coverage exported; all six core partitions, static analysis, coverage, release/API-contract and core/bundled distribution checks passed. Current-head secret-scan and Decisions checks also passed. The prior head's TerminationClusterTest timeout is historical; this review does not claim its root cause was repaired.

CI evidence: https://github.com/managoat/fountain/actions/runs/34813901772

This is the requested personal source review, not a claim that Review Loop service verification ran.

PR #2162, opened 2026-09-14
refactor: remove fourteen public functions nothing calls

jhgaylor:
## What

A `mix_unused` sweep of `apps/fountain` (run in a throwaway worktree, not added to the repo) reported 1,680 public functions with no static caller. After filtering protocol impls, callbacks and router-dispatched actions, and cross-checking every survivor with a repo-wide grep across core, `ee/`, both extension apps and every test tree, fourteen had no caller in shipped code. This removes them and the tests that exercised only them: **639 lines deleted, 12 added.**

| Function | Why it had no caller |
|---|---|
| `Conversations.list_conversations_by_activity/1` | backed the retired `/conversations` LiveView (#869) |
| `Conversations._unsafe_list_active_conversations/0` | backed the retired left-nav (#869) |
| `Team.list_addable_agents/1` | backed the retired `/team` add picker (#869) |
| `Accounts.update_preferences/2` + `User.preferences_changeset/2` | backed the retired conversation view settings (#869) |
| `Apps.new_conversation_url/0`, `Apps.team_url/1` | last callers left with #2115 |
| `InferenceCredentials.has_own?/3` (+ `secret_keys/2`, `shadowed?/2`) | superseded by `resolve/4` in #2031 |
| `PlatformInference.serves?/3` | superseded by `resolve/4` in #2031 |
| `Environments._unsafe_get_environment/1`, `_unsafe_get_environment!/1` | only their own tests read them |
| `Vaults._unsafe_get_vault!/1`, `_unsafe_get_vault_by_name/1` | only their own tests read them |
| `Conversations._unsafe_dispatch_prompt/3` | added in #1749, never wired to the API door |
| `Credits.lot_remaining_for/2` | added in #1551, the refund path never used it |
| `Connections.McpServers.connection_entry?/1` | `remote_entry?/1` is the one the sync uses |

Four fewer `_unsafe_` entry points is the part worth having on its own.

## Kept on purpose

- `Accounts.update_execution_limits/3`: its docstring says the admin control is still to come. Documented intent, not dead code.
- `SpriteEnv.select_inference/4` and `load_tenant_state/2`: superseded by `resolve/4` too, but their tests exercise real credential-set semantics and deserve a rewrite of their own rather than a delete. Follow-up.
- The 27 other functions only tests call are test seams (`_unsafe_fenced?`, `_unsafe_capture`, the timeout accessors, the docs nav source, and so on). Left alone.
[...]

PR #2163, opened 2026-09-14
chore: a monthly dead-code report, advisory and off by default

jhgaylor:
## What

The follow-up to #2162, which ran `mix_unused` by hand in a throwaway worktree. This makes that run one command and one scheduled job, without adding a gate.

- **`apps/fountain/mix.exs`**: `mix_unused` as a dev-only dep, its compiler enabled only when `MIX_UNUSED=1` is set. An ignore list drops the shapes that are never dead (protocol impls, OTP and Phoenix callbacks, router-dispatched controllers, LiveViews and plugs, the `Repo`, `Mailer`, `OAuth` and `Docs` `use` surfaces, the extension seam) and skips the "should be private" analyzer, which is hygiene rather than dead code. Over the current tree the report goes from 1,680 lines to 495. With the variable unset nothing changes: same compilers, same build.
- **`scripts/dead-code.sh`**: runs the Elixir report and `golang.org/x/tools/cmd/deadcode` (pinned) over both Go modules.
- **`.github/workflows/dead-code.yml`**: runs both on the first of the month and on demand, uploads each report as a 90-day artifact, and writes the counts and top namespaces to the job summary so the trend is visible without downloading anything. `permissions: contents: read`, pinned actions, its own cache keys.
- **CONTRIBUTING.md** gets a "Finding dead code" section on how to read the report: the five shapes the tracer gets wrong, and the test-only-caller case where the right answer is sometimes to restore the call site rather than delete. CLAUDE.md's CI section points at it in one paragraph.

## Why not a gate

The tracer sees only static calls inside `apps/fountain`. Anything reached through `apply/3`, an MFA tuple, an extension app or a test reads as unused. The residual 495 would need a hand-tended allowlist on every PR, and #2162 found fourteen real deletions in it. A monthly report someone reads is the right cost for that yield.

## Verified

- `mix format --check-formatted`, `MIX_ENV=test mix compile --warnings-as-errors`, `mix credo --strict`, `mix deps.unlock --unused` (lock unchanged), the `scripts/ci` policy suite (57 tests), `docs_test.exs` (CHANGELOG is a docs page), shellcheck on the script, and the workflow parses.
- `scripts/dead-code.sh elixir` on this branch: 495 hints. `scripts/dead-code.sh go`: two unreachable exported functions in `cli/`, both API another module may import, none in `apps/fountain_buzz/cli`.

https://claude.ai/code/session_01HkopMNnchbyzYhYEbT6UbV


scripts/dead-code.sh
```diff
@@ -0,0 +1,54 @@
...
+go_report() {
+  echo "## Go: unreachable functions (golang.org/x/tools/cmd/deadcode)"
+  echo
+  for mod in cli apps/fountain_buzz/cli; do
+    echo "### $mod"
+    # -test: a function only a test reaches is not unreachable. An exported
+    # function in cli/api or cli/credentials may still be one another module
+    # imports (apps/fountain_buzz/cli does), so read those lines as API, not
+    # dead code.
+    (cd "$root/$mod" && go run "golang.org/x/tools/cmd/deadcode@$DEADCODE_VERSION" -test ./...) || true
```
BinaryBourbon: **[P2] Preserve analyzer execution failures instead of reporting them as zero findings.** This `|| true` discards any `go run` failure, including a package-loading error or an unavailable tool download. The script then exits successfully with only section headings, and the workflow counts zero `unreachable func` lines and publishes `none`, indistinguishable from a successful clean scan. The Elixir branch has the same issue at lines 23–25 and additionally discards diagnostic stderr. Advisory findings do not need to fail a merge gate, but failed analysis must remain visible: retain compiler/tool exit status and diagnostics, and tolerate only the filtering command's expected no-match result after a successful analyzer invocation. If the workflow deliberately stays green on execution failure, mark that report incomplete rather than emitting a zero count.

apps/fountain/mix.exs
```diff
@@ -41,11 +43,61 @@ defmodule Fountain.MixProject do
...
+  # before acting on a line of it.
+  defp unused_compiler do
+    if System.get_env("MIX_UNUSED") in ["1", "true"], do: [:unused], else: []
+  end
+
+  defp unused_config do
+    [
+      # Not MixUnused.Analyzers.Private: "public but only called from its own
+      # module" is hygiene, not dead code, and it is ~350 lines of the report.
+      checks: [MixUnused.Analyzers.Unused, MixUnused.Analyzers.RecursiveOnly],
```
BinaryBourbon: **[P2] Use an analyzer selection mechanism supported by the locked release.** Published `mix_unused` 0.4.1's `MixUnused.Config.extract_config/2` copies only `ignore`, `severity`, and `warnings_as_errors`; it never reads this `checks` key. Its default still contains `MixUnused.Analyzers.Private`. Consequently a public function that is called within its own module still produces a `hint: ... should be private` diagnostic, which `scripts/dead-code.sh` includes and the job summary counts as an unused function. This directly contradicts the intended exclusion above and contaminates the monthly dead-code counts with live functions. Filter those diagnostics explicitly in the report, or use a released analyzer version that actually supports this configuration, and remove the ineffective option. Confirm the report excludes a function with a real same-module caller while retaining genuinely unused functions.

BinaryBourbon (changes requested):
Adversarial review of f9d31f7f against main 5e499e49: **changes requested** for two report-correctness defects below.

Reviewed all seven changed files, the scheduled/manual workflow and its permissions/caches/artifacts, both report commands and Go modules, compiler activation and dependency scope, the locked analyzer's published source, and contributor guidance. The report remains separate from merge gates and the new dependency is dev-only; I found no concrete change to tenant authorization, credential handling or the runtime API.

[Current-head CI run 34812285658](https://github.com/managoat/fountain/actions/runs/34812285658) has passing static analysis, release/API-contract, core distribution, SDK checks and five test partitions. Partition 5 fails in the unchanged `TerminationClusterTest` while calling the peer receiver (`:gen_server.call(..., 5000)` times out at test line 9); coverage is skipped and the aggregate check fails. I found no causal connection between that failure and this diff, so it is not one of the findings and is not being declared a confirmed flake. CI still needs resolution before merge.

No local test suite was run, as requested. This is a personal source review, not a Review Loop service-verification result.

scripts/dead-code.sh
```diff
@@ -0,0 +1,57 @@
...
+
+elixir_report() {
+  echo "## Elixir: public functions no compiled module calls (mix_unused)"
+  echo
+  # --force so the tracer sees every module, not just the ones that changed.
+  # Keep compiler diagnostics visible and propagate analyzer failures. Only
+  # dead-code hints enter the report: mix_unused 0.4.1 cannot disable its
+  # Private analyzer through config, and a same-module caller is still live.
+  (cd "$root/apps/fountain" &&
+    MIX_UNUSED=1 MIX_ENV=dev mix compile --force) |
```
BinaryBourbon: [P2] Reset the analyzer manifest before forcing a fresh report

The monthly workflow restores `_build` across revisions (dead-code.yml:50-56), including `unused.manifest`. In locked mix_unused 0.4.1, `update_manifest/2` does `Map.merge(cache, data)`, so `mix compile --force` replaces traces for existing modules but never removes calls belonging to a deleted module. A removed controller's old calls can therefore keep now-unused context functions absent from the report indefinitely.

Reproduced with the real 0.4.1 analyzer in an isolated two-module fixture: `Caller.work/0` calls `Target.work/0`, and `Caller` is ignored exactly as this PR ignores controllers. (1) Initial compile: no Target hint. (2) Delete Caller and run `mix compile --force`: still no Target hint. (3) Remove `_build/dev/lib/probe/.mix/unused.manifest` and repeat the same compile: `hint: Target.work/0 is unused`. All commands exit zero. The probe used Elixir 1.19.5; the relevant cache-merge behavior is in the locked analyzer source.

Clear Fountain's analyzer manifest before each forced report (or exclude/invalidate it when restoring caches), while retaining dependency build caches. Add a regression that removes a caller between two reports and checks that the newly unused callee appears.

BinaryBourbon (changes requested):
Adversarial whole-PR re-review of 2fe21eb90bd26afbbc620e399a5507f75a3ada33 against main at 5e499e49e7c7511f98561d4169dc995c970881b7: changes requested for one reproduced report-correctness defect.

[P2] Clear the analyzer manifest before a fresh report. scripts/dead-code.sh:25-26 forces recompilation but preserves mix_unused's unused.manifest; .github/workflows/dead-code.yml:50-56 restores that manifest with _build across source revisions. In the locked mix_unused 0.4.1, update_manifest/2 merges cached call data with the current trace using Map.merge(cache, data), retaining entries for deleted modules. When a controller (ignored by unused_config/0) is deleted, its stale calls can keep a now-unused context function out of every subsequent report. This directly affects the retired-controller sweep this report is intended to support. See the inline finding for the reproduction and minimal remedy.

The previous two findings are resolved: same-module “should be private” advice is filtered out, and analyzer failures propagate with visible diagnostics and an incomplete workflow summary. I reviewed all ten changed paths, compiler activation/dependency scope, analyzer source matching the locked 0.4.1 package metadata, both Go modules, workflow permissions/cache/artifact behavior, contributor guidance, and the distributed termination fixture. The latter preserves the actual five-second request timeout and attribution/refusal assertions. No additional concrete runtime API or tenant-boundary defect found.

Verification: the five new report tests pass in an isolated copy of this head; bash syntax validation passes. The first sandboxed test attempt was blocked by the sandbox forbidding tee to open /dev/stderr; the unchanged tests all passed outside that restriction. A real, database-free mix_unused 0.4.1 fixture reproduced the manifest defect with local Elixir 1.19.5; CI pins 1.19.2. [CI run 34813646445](https://github.com/managoat/fountain/actions/runs/34813646445), a jhgaylor pull_request run on this exact head, succeeded including all six partitions, static analysis, release/API contract, coverage, SDK checks, and CI required; current-head secret scanning also succeeded. I did not run the full local Elixir/Go suites or dispatch the new scheduled workflow. Base, head and BinaryBourbon identity were rechecked before submission. This is an independent agent-assisted review, not a Review Loop service verification result.

BinaryBourbon (approved):
Adversarial whole-PR re-review of `a7eeee1dc2c13da8d881d84f55440963de7ee83c` against `5e499e49e7c7511f98561d4169dc995c970881b7`: approved after two independent source reviews; neither found a remaining actionable medium-or-higher defect.

The previous stale-caller finding is resolved. Each Elixir report now cleans Fountain's dev artifacts with the unused compiler enabled before forced compilation, removing the analyzer manifest that otherwise retained deleted modules' calls. I independently ran the real locked-analyzer regression: deleting the ignored caller between reports now exposes `Target.work/0`, while dependency BEAM timestamps and a separate test-environment artifact remain unchanged. The same-module advice filter and analyzer-failure propagation remain intact.

Reviewed all 11 changed paths: the scheduled workflow, both contributor guides, Fountain's Mix configuration and lockfile, both distributed-termination fixture files, the changelog fragment, both Python report tests, and the report script. Traced the locked analyzer's manifest/clean callbacks, both Go modules, workflow permissions/caches/artifacts/failure summaries, and the termination request/fence/refusal path. No runtime API or tenant-boundary change was introduced by the test-fixture adjustment.

Local verification: the real deleted-caller regression passed (1 test, Elixir 1.19.2/OTP 28.3); all report filtering/failure regressions passed (5 tests); `bash -n scripts/dead-code.sh` passed. I did not run the full local application or Go suites, or dispatch the scheduled workflow.

CI remains a separate merge blocker: [run 34816102895](https://github.com/managoat/fountain/actions/runs/34816102895) has a failed `CI policy and alert tests` job and therefore failed `CI required`. Its unchanged changelog guard fails when `git diff --name-only 5e499e49...HEAD` exits 128; the checkout log shows a shallow merge against a newer main revision. This supports a checkout-history issue, not a regression in the changed report files. The six test partitions, static analysis, coverage, release/contract and SDK checks succeeded. This review does not waive the failing gate or claim merge readiness.

This is an independent agent-assisted review, not a Review Loop service verification result. BinaryBourbon identity and the unchanged PR base/head were verified immediately before submission.

PR #2168, opened 2026-09-14
refactor(connections): the Gmail MCP server becomes the fountain_google extension

jhgaylor:
Step 2 of #2152, stacked on #2167 (Slack) so the stack is linear and step 4b (the Google provider into this app) can follow. Closes #1529, which ADR 0043 specified as `fountain_gmail`; the app is `fountain_google` because the Google connection provider joins it next.

## What moves

- `Fountain.Connections.Mcp` (the Gmail JSON-RPC tool layer) → `FountainGoogle.Mcp`; `Fountain.Connections.Gmail` (the API client) → `FountainGoogle.Gmail`; `FountainWeb.GmailMcpController` → `FountainGoogle.McpController` behind `FountainGoogle.Router`, mounted at `/mcp/gmail` so `POST /api/mcp/gmail/:conversation_id/:connection_id` is unchanged. The `fountain-gmail` manual page moves to the extension's `docs/` through `docs/0`.
- The extension contributes its server through the existing `conversation_mcp_servers/2` callback: for each `%{"connection" => id}` entry with no URL whose connection is active, the tenant's, and on the Google provider, it returns the HTTP server in the ACP shape Buzz uses. No new callback.
- Core no longer knows the Fountain-served shape: `Fountain.Connections.McpServers.resolve/4` drops connection-only entries (the remote shape with a URL stays), so the server reaches every runtime through `session/new` and is never written to `.mcp.json`. The `extensions_test` example of a core static prefix moved off the Gmail route.
- Wiring: `installed_extensions`, `Dockerfile`, the guard row (`FountainGoogle\.`, `Fountain\.Connections\.(Mcp|Gmail)\b`, `FountainWeb\.GmailMcp[A-Z]`, route `/mcp/gmail`, priv `*gmail*`), boundary test, CLAUDE.md layout, `scripts/test-timings.tsv`, `:gmail_req_options` → `config :fountain_google, :req_options`. Core pages that linked `fountain-gmail` now name it without a link. ADR 0043 decision 3 notes the build.
- Changelog fragment `changelog.d/2152-google-extension.md` (the last commit carries the fragment and the `CHANGELOG.md` revert alongside a test fix; squash-merge collapses that).

## Verification (on the rebased tree)

- `apps/fountain_google`: `mix test` → 44 tests, 0 failures. One test named a Slack connection as "the other provider"; with Slack an extension of its own it now uses the fixture extension's provider.
[...]

PR #2167, opened 2026-09-14
refactor(connections): the Slack provider becomes the fountain_slack extension

jhgaylor:
Step 4d of #2152, stacked on #2161 (Microsoft). The same move, for Slack.

## What changes

- `apps/fountain_slack`: an AGPL OTP app depending on `:fountain` with two callbacks. `connection_providers/0` returns the Slack provider built from `config :fountain_slack` (`SLACK_OAUTH_CLIENT_ID` / `_SECRET` / `SLACK_OAUTH_USER_SCOPES`, read in `config/runtime.exs` where the app loads); `docs/0` carries the `slack (connection)` page at its old URL. Boundary test, provider test (field for field against what core built), scope-override test, docs and manual tests, and `FountainSlack.CoreUpgradeTest`.
- Core builds Google alone. `Fountain.Connections.Platform` keeps the registry functions and knows Slack only as a row an installed extension contributes. The Slack builder, its scope constant, the `:fountain` config keys and the core docs page are gone; the guard test has a `:fountain_slack` row.
- Slack's two quirks travel as data on the struct (`authorize_params` with `user_scope`, `token_body_nest: "authed_user"`), which is what step 1 was for.
- Changelog fragment `changelog.d/2152-slack-extension.md`.

## Verification

- `apps/fountain_slack`: `mix test` → 30 tests, 0 failures (includes the core-upgrade test: persisted Slack grants on a core-only deployment stay revocable and deletable and contribute no token or implicit host).
- `apps/fountain`, the connections/extension/docs/api-spec/config-reference/lifecycle set → 226 tests, 0 failures.
- Guard broken on purpose (a `FountainSlack.Extension` mention in `platform.ex`) → 22 tests, 1 failure naming the line; restored.
- `mix format` at all three roots (no diff), `mix compile --warnings-as-errors`, `mix credo --strict`, `python3 scripts/changelog.py check`. Full `mix precommit` running; result posted as a comment.

## Stack map (#2152)

#2156 (step 1) → #2155 (4a, ADR 0054) → #2161 (4c Microsoft) → **this PR** (4d Slack) → step 2 (`fountain_google`, next) → 4b (Google provider into it). #2157 (step 3, OAuth client library) sits on #2156.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2170, opened 2026-09-14
refactor(connections): the Google provider joins the fountain_google extension; core builds none

jhgaylor:
Step 4b of #2152, stacked on #2168 (step 2). The last move of ADR 0054 decision 6: after this, core builds no platform provider.

## What changes

- `FountainGoogle.Provider` builds the Google provider from `config :fountain_google` (`GOOGLE_OAUTH_CLIENT_ID` / `_SECRET` / `_SCOPES`, read in `config/runtime.exs` where the app loads), field for field the struct core built; `FountainGoogle.Extension.connection_providers/0` contributes it beside the Gmail MCP server the same app serves. The `google (connection)` manual page moves into the extension and keeps its URL.
- `Fountain.Connections.Platform` keeps the registry functions over what installed extensions contribute; `builtin_slugs/0` is `[]`; the Google builder, its scope constant and `Fountain.Connections.Google` are gone. `config/runtime.exs` reads no provider client into `:fountain` any more (the `platform_oauth_clients` loop is gone; each extension has its own block). Test config, `.env.example`, `docs/configuration.md`, the connections catalog index, CLAUDE.md and the guard row (`Fountain\.Connections\.Google\b`, `Platform\.google\b`) follow. The Microsoft and Slack pages no longer link the Google page (an extension page must not link another extension's).
- The core suite's default connection is now the fixture extension's provider (`fixture-svc`, which gains a revoke endpoint and an app-env-readable client id for the console's "not configured" test). Every core test that asserted Google's env key, hosts, scopes, consent parameters or start/callback paths asserts the fixture's; the browser OAuth flow test runs against the fixture's endpoints. The lifecycle test's "retained" connection is a tenant provider's, the one kind no extension removal touches. The "slug the host already owns" validation case is gone with the host's last slug (`ProvidersTakeGoogle` fixture removed); the cross-extension duplicate case stays.
- The Google specifics live in the extension's suite: `FountainGoogle.ProviderTest`, `ScopeOverrideTest`, `CoreUpgradeTest` (a core-only deployment that stored Google grants keeps them revocable and deletable and contributes no token or implicit host), and the manual test covers the moved page.
- ADR 0054: status **built**, `stale_after` removed, description and decision 6 updated, index regenerated, `okf validate` clean. ADR 0033 decision 1 says Google has moved. Changelog fragment `changelog.d/2152-google-provider-extension.md`.

- Platform identity is explicit in Gmail discovery, endpoint admission, and console setup instructions: a tenant provider named `google` never qualifies as a platform grant. Tenant accounts retain remote/stdio guidance after extension installation.
- A new core migration replaces slug-only connection uniqueness with partial indexes for platform `(user_id, provider, account_email)` and tenant `(user_id, provider_id, account_email)` identities. Same-label accounts can coexist, and reconnecting updates only the corresponding grant. Upgrade preserves existing credentials. Rollback restores the old constraint first and refuses if newly coexisting identities would violate it; it never deletes grants.

## Verification

- Both reported lifecycle regressions reproduced on the previous revision before their fixes.
- Targeted core/provider/LiveView/migration checks: 38 tests, 0 failures. Google extension: 60 tests, 0 failures.
- Full `mix precommit` passed on the isolated test database: compilation, formatting, Credo, Dialyzer (0 errors), Sobelow, production release assembly, core 6 doctests / 6,069 tests / 0 failures / 2 skipped; extension suites Buzz 136, Slack 30, Support 33, Google 60, Microsoft 29, all 0 failures.
- Regression coverage includes rendered setup instructions, tenant/platform same-label coexistence and separate reconnects, database duplicate rejection, migration upgrade/rollback credential preservation, and refusal of rollback once the old schema cannot represent the data.
- `python3 scripts/changelog.py check` and `git diff --check` passed.

## Stack map (#2152)

#2156 (step 1) → #2155 (4a, ADR 0054) → #2161 (4c Microsoft) → #2167 (4d Slack) → #2168 (step 2, `fountain_google`) → **this PR** (4b). #2157 (step 3, OAuth client library) sits on #2156. With this PR the tracker's four steps are all open.


jhgaylor:
Full `mix precommit` on the branch, read from the log: compile with warnings as errors, format, `credo --strict`, sobelow, dialyzer (`Total errors: 0`), prod release assembled; core suite and every sibling suite at 0 failures (literal lines in the run log below, posted from the session that ran it).

jhgaylor:
Literal lines: core `6 doctests, 6067 tests, 0 failures, 2 skipped`; `scripts/test-libraries.sh`: buzz 136, slack 30, support 33, google 56, microsoft 29, all 0 failures; exit 0.

BinaryBourbon (changes requested):
Adversarial review of `2e4b0663efc876bb7f88c687cae2951d5c62eace` against `3038a66e0bd3fd86bb29a53c61ed34e32c762889`: **changes requested**.

**[P2] Distinguish platform Google grants from tenant providers before exposing Gmail tools.** Removing the built-in reservation in `apps/fountain/lib/fountain/connections/platform.ex:37` makes `google` a valid tenant-provider slug whenever the Google extension is absent. A tenant can create that provider and its connections on a core distribution, then the operator can install `fountain_google`. Boot validation checks collisions between extension contributions, without checking tenant rows, so those connections survive.

Both Gmail entry points currently identify the provider by slug alone: `apps/fountain_google/lib/fountain_google.ex:81` selects an active connection with `provider: "google"`, and `apps/fountain_google/lib/fountain_google/mcp_controller.ex:65` accepts the same slug. A surviving tenant connection has a non-null `provider_id`; `Connections.provider_for/1` consequently resolves its tenant OAuth provider, and `Connections.access_token/1` returns that provider's credential. The Gmail tool path then sends it to `gmail.googleapis.com`. This exposes Gmail tools for the wrong service and forwards a credential outside its intended service.

Require `provider_id: nil` in both Google connection guards, and add a regression that creates a tenant provider named `google` while the extension is absent, installs the extension, and verifies that its connections are omitted from Gmail server discovery and rejected by the Gmail endpoint without an API request. The intended removal of core's reserved slug can remain.

Validation: an isolated diagnostic using the head's provider changeset accepted the tenant `google` provider with no validation errors; the actual Gmail controller guard extracted from this revision returned `:ok` for its non-null-`provider_id` connection. Source tracing confirmed the installation and token-delivery paths. The moved provider also matched the base's complete struct across eight client/scope configurations. Current-head CI and secret scanning pass. This review covered the full PR diff and affected callers; I did not rerun the full database suite locally.

BinaryBourbon (changes requested):
**Changes requested** after three independent adversarial reviews of the full PR at `28b87819dc16eb4b4d9c81a78151315fc4494498`, against stacked base `3038a66e0bd3fd86bb29a53c61ed34e32c762889`, covering correctness, security/tenant isolation, and compatibility.

The previous credential-routing finding is fixed: both Gmail discovery and endpoint admission now require `provider_id: nil`. Two P2 defects remain in the same newly supported lifecycle: creating a tenant provider named `google` while the extension is absent, then installing Google.

1. **[P2] Give tenant Google providers usable connection instructions.** In `apps/fountain/lib/fountain_web/live/connections_live/index.ex:866–873`, the account card still tests only the `google` slug and extension presence. It displays `{"gmail":{"connection":"<tenant-connection-id>"}}` and hides the remote URL/stdio instructions. Following that example silently produces no server: core drops connection-only entries, and the corrected Google contributor rejects the tenant provider ID. A diagnostic using a real LiveView mount reproduced the displayed snippet, the missing remote instructions, and both empty resolver results. Require platform provenance (`provider_id == nil`) in the Gmail display predicate and use the remote instructions otherwise. Add a LiveView regression for both tenant and platform grants after installation.

2. **[P2] Preserve distinct provider identities in connection uniqueness.** Removing Google's permanent reservation in `connections/platform.ex:37` makes a tenant Google grant and a platform Google grant with the same account label possible. However, `apps/fountain/lib/fountain/connections/connection.ex:74` and the unique index created in `20260826031178_create_connections.exs:28` still constrain `(user_id, provider, account_email)`. After creating a tenant `google` connection for `me@example.com` and installing the extension, `Connections.connect(user.id, "google", grant)` fails with `connections_user_id_provider_account_email_index`: its lookup correctly distinguishes provider provenance, but its database constraint does not. A diagnostic reproduced this against actual persisted rows. Use separate tenant/provider-ID and platform/slug uniqueness (for example, appropriate partial indexes), update the changeset constraints, and test that same-label tenant/platform grants coexist while reconnects remain idempotent.

Validation: current-head GitHub CI, including `CI required` and secret scanning, is green. The local precommit log shows 6,067 core tests plus all five extension suites with zero failures; the Google suite has 58 tests. Two additional isolated database/LiveView diagnostics confirmed the defects above using synthetic data. No repository changes were made during this review.

BinaryBourbon (approved):
**Approved** after three independent adversarial whole-PR reviews of `8a4f22f3ba589189135413e3d5587b900cae8daf` against stacked base `3038a66e0bd3fd86bb29a53c61ed34e32c762889`. Reviewers covered all 51 changed paths, with separate emphasis on migration/correctness, security/tenant isolation, and product/API compatibility. No remaining actionable findings.

All three previous findings are addressed:

- Gmail discovery and endpoint admission require platform provenance (`provider_id: nil`), so tenant providers named `google` cannot send their credentials to Gmail.
- Console setup instructions use the same identity check and retain remote/stdio guidance for tenant providers.
- Separate partial unique indexes allow same-label tenant/platform accounts while preserving duplicate protection, independent reconnects, connection IDs, and environment keys. Changeset constraints match the new indexes.

The extraction preserves the Google provider's OAuth parameters, scope configuration, credential destinations, public configuration names and manual URLs. Missing-extension behavior fails closed for token access and preserves local cleanup. The new migration builds both replacement indexes before dropping the old one and preserves existing grants. Its rollback deliberately refuses once coexisting identities cannot fit the older uniqueness rule; it does not delete or combine credentials.

Validation: current-head GitHub CI is green, including `CI required` and secret scanning. Reviewed local precommit evidence: 6 doctests and 6,069 core tests with zero failures (2 skipped), plus Buzz 136, Slack 30, Support 33, Google 60 and Microsoft 29 tests, all with zero failures. The targeted checks and regression coverage exercise the prior failures, identity-specific reconnects, duplicate rejection, and migration upgrade/rollback. The reviewers inspected these logs and the source independently; they did not each rerun the full suite or perform live-provider OAuth exchanges.

Issue #333, opened 2026-08-02
ci: precommit drifts from CI, the manifest image pin isn't verified, and deploys aren't gated on CI


jhgaylor:
Three separate CI/CD gaps, all small, filed together.

## 1. `mix precommit` is not "the same checks CI runs"
`mix.exs:68` runs `credo --strict --mute-exit-status` — precommit can never fail on credo, while CI gates on `mix credo --strict` (`ci.yml:84-85`). Combined with sobelow being absent from precommit entirely, CLAUDE.md's "same checks CI runs" is false in two places. Local-green/CI-red drift only, but it's the kind that wastes a push cycle every time.

## 2. The manifest pin step can silently publish a stale image tag
`.github/workflows/publish-manifests.yml:87-95` rewrites the image with
`sed -E "s|ghcr\.io/binarybourbon/fountain:sha-[0-9a-f]+|${IMAGE}|"` and never verifies the substitution happened. If `k8s/deployment.yaml`'s image line stops matching that pattern — a digest pin, a registry rename — sed no-ops, `git diff --stat` merely prints, and the artifact publishes with the **old** tag. That recreates exactly the manifest/image mismatch (#241) this workflow exists to prevent. It matches today (`k8s/deployment.yaml:38`). Fix: `grep -q "$IMAGE" k8s/deployment.yaml` after the sed, and fail if absent.

## 3. The deploy pipeline has no CI dependency
`build.yml` triggers on every push to main independent of `ci.yml`, and `publish-manifests.yml:31-35` triggers on **build** success only. A main commit whose CI run fails still gets built, published, and reconciled by Flux. Mitigated by PR-only merges and the watched-rollout convention, but nothing structural stops it.

Related and already documented, noted here for completeness: `build.yml:16-24` `paths-ignore` includes `k8s/**`, so a manifest-only merge triggers neither build nor publish and reaches the cluster only via a manual `workflow_dispatch`.

PR #2174, opened 2026-09-14
ci: resolve the issue and PR numbers a PR cites, as a comment

### LINKED ISSUE #1014: CI: resolve issue and PR citations, so a made-up number cannot ship
`#1006` was cited in 20 places across 13 files as the tracking number for retiring the GitHub Pages docs site. It is an unrelated open PR ("ci(sdk): make CI the only publisher, and the version bump part of review"). The number was invented in the first commit of that work and copied outward from there into `docs.ex`, `compiler.ex`, `router.ex`, `docs_controller.ex`, the docs tests, `ci.yml`, `CLAUDE.md`, `.vale-ste.yml`, `docs-style.py`, `superpowers/README.md` and two CHANGELOG lines.

It survived `mix precommit`, three green CI runs and two merges. Nothing in the pipeline resolves a citation. It surfaced only because someone happened to list open PRs for an unrelated reason. Fixed in #1012.

This is the third time: `fountain-adr-numbers-in-old-issues` records #470 and #148 citing ADRs that turned out to be different documents.

## The obvious check does not work

A "does this issue exist?" lint returns 200 for #1006 and moves on. Existence is necessary and not sufficient, and shipping only that check is worse than shipping nothing, because it looks like a guard while catching none of the real cases.

## The rule with teeth

A citation in code is a reference to history: why this line looks the way it does. So:

> A newly-added `#N` must resolve **and already be closed or merged** at CI time.

#1006 was `OPEN`, so this catches it. It also catches the general shape of the mistake, which is guessing at a number that does not exist yet, including your own PR's.

## Scope to added lines only

There are 2,059 `#N` citations in the tree. Checking all of them needs an allowlist the size of the corpus. Diff-scoped, it is a ratchet by construction, the same shape `scripts/docs-style.py` already uses for the style backlog.

## The regex is the hard part, not the API call

Real collisions in this repo, all of which a naive `#\d+` eats:

```
#111827  #000000   pure-numeric hex colors
&#106;   #115;     HTML numeric character references (markdown_test.exs, the XSS tests)
```

`106` and `115` are plausible issue numbers. Discriminators that hold today:

- Cap at the repo's current highest issue-or-PR number (1013 at time of writing). Every 6-digit hex colour drops out.
- Reject anything preceded by `&`, which kills the character references.
- Require the shape `(#N)` or `#N` at a word boundary.

Three-digit pure-numeric hex (`#000`) still collides and needs care.

## Suggested staging

1. **Comment, do not gate.** A PR comment listing each newly-added citation with its resolved state and title:

   ```
   #1006 → [OPEN] ci(sdk): make CI the only publisher, and the version bump part of review
   ```

   Next to a docs comment about MkDocs that is wrong at a glance, to a human or a model. No false-positive tax, because it blocks nothing. This is most of the value for the least risk.

2. **Then gate**, once the regex has proven itself over a few weeks, by promoting "resolves and is closed" to a hard failure.

Shape: `scripts/check-issue-refs.py`, wired into the `checks` job beside the other prose gates, `pull_request` only since it needs the diff and the API. Public repo, so `github.token` is enough.

Validate it by pointing it at the diff of #1008 and confirming it flags `#1006` before trusting it anywhere.

## The part that needs no lint

Prefer generated references over authored ones. `f89523f`'s trailing `(#1008)` is right and the inline `(#1006)` was wrong because GitHub generated the first and a human typed the second.

The corollary is a convention worth writing into `CLAUDE.md` regardless of whether this lint gets built: **do not cite the current change's own number inside the current change.** You cannot know it. Let the merge subject carry it, and cite it explicitly only from follow-up work.

## What this will not catch

A citation to a closed, plausible-looking, wrong issue. Only the resolved-title comment helps there, which is another argument for building step 1 first.

Follow-up to #1008 and #1012.

jhgaylor:
Closes #1014.

`#1006` was cited in 20 places across 13 files as the tracker for retiring the MkDocs site. It was an unrelated open PR, guessed in the first commit and copied outward, and nothing in CI resolves a citation. Stage 1 of the fix, as the issue lays it out: comment, do not gate.

## What this adds

- `scripts/ci/check_issue_refs.py` walks a diff's added lines for `#N` and `github.com/managoat/fountain/{issues,pull}/N`, resolves each through the GitHub API and prints one line per number with its state (`OPEN`, `CLOSED`, `MERGED`, `MISSING`) and title, plus `path:line` for each place it was cited.
- `--comment --pr N` upserts one comment per PR, found by a marker, so re-runs update rather than stack. A PR that adds no citation gets no comment. The report also goes to the step summary, which is what survives when a fork's read-only token cannot comment.
- `--strict` is stage 2: exit 1 on an open or missing citation. Not wired yet; the README says what to flip and names the trade-off (a PR that cites the issue it fixes cites an open issue).
- `scripts/ci/test_issue_refs.py`: 23 tests, API mocked at the one seam. `test_the_1008_diff_flags_1006` is the validation the issue asked for, with #1006 pinned at the state it had then.
- The `workflow-checks` job gets the step, `pull_request` only, `continue-on-error: true` so an API outage cannot block a merge. The changelog guard's base resolution moves into its own step so both diff-scoped checks read `PR_BASE_SHA`; the guard itself is unchanged.
- CLAUDE.md gets the convention the issue asked for regardless: do not cite the current change's own number inside the current change.

## The regex

A match needs no `&`, `#` or word character before it, no leading zero, no `;` after it, and a value at or under the repo's newest issue-or-PR number (read from the API), which drops every six-digit colour. Stylesheets are skipped. Three-digit numeric colours in prose can still match, which is the reason stage 1 comments rather than gates.

## Validation

Pointed at #1008's diff it lists `#1006` at 18 locations with its SDK-publisher title. `#1006` has since merged, so today it reads `MERGED`; the unit test pins `OPEN`.

Run over this branch's own diff, the one open citation it reports is #1014. It also flags the README's own `#333` colour example as a real closed issue, which is exactly the false-positive shape the README describes.

No changelog fragment: nothing a user or operator observes changes.

https://claude.ai/code/session_01KZXLxPERK164yNsdpEfUQw


scripts/ci/check_issue_refs.py
```diff
@@ -0,0 +1,267 @@
...
+        "A citation in code names history, so a newly added number should already "
+        "be closed or merged; a title that does not match the sentence around it "
+        "is a made-up number (#1014). This report is advisory.",
+    ]
+    return "\n".join(lines) + "\n"
+
+
+def upsert_comment(repo, pr, body):
+    """One comment per PR, found by its marker, updated on every run."""
+    for comment in api("GET", f"/repos/{repo}/issues/{pr}/comments?per_page=100"):
```
BinaryBourbon: [P2] Paginate both report lookups

This request only reads the first 100 comments. If a PR already has 100 comments when the first citation report is posted, every rerun misses that report and creates another. The no-citations check at lines 251–252 has the same limit, so removing the citations leaves old reports uncorrected. A mocked reproduction with 100 initial discussion comments returned `created` on three consecutive upserts and retained three reports.

Use a shared paginated lookup for upsert and empty-report correction. Add a regression with the existing report on page two that verifies both updating it and clearing it after all citations are removed.

scripts/ci/check_issue_refs.py
```diff
@@ -0,0 +1,267 @@
...
+        if comment["body"].startswith(MARKER):
```
BinaryBourbon: [P2] Check report ownership before updating a marked comment

The public marker alone does not identify this script's report. If someone copies it into an earlier comment, this loop selects that person's comment for PATCH instead of the genuine bot report. The job's write permission can overwrite another person's comment; that person also retains control of the adopted report. A mocked reproduction with an outsider-authored marked comment before the real bot report selected the outsider's comment for PATCH.

Require both the marker and the expected authenticated report author's identity, including the lookup at lines 251–252. Add a regression ensuring another user's marked comment is left untouched and the genuine bot report is updated or created.

GitHub documents [editing others' comments with write access](https://docs.github.com/en/communities/moderating-comments-and-conversations/managing-disruptive-comments#editing-a-comment) and [pull-request write permission for this endpoint](https://docs.github.com/en/rest/issues/comments#update-an-issue-comment).

BinaryBourbon (changes requested):
Adversarial review verdict: **Changes requested** for `277db4b9ccaa5a906034fde6eb109e6c5257d797` (base `74ad641f1f7c2cc06904f67a35a0c3c40ea58f68`).

Four independent Codex reviewers covered all five changed files:
- Correctness/tests: changes requested — comment pagination.
- Security: changes requested — comment ownership validation.
- Product/API behavior: changes requested — comment pagination violates the documented single-report behavior.
- Simplicity/maintainability: changes requested — comment pagination.

The two distinct P2 findings are attached inline. Please paginate both report lookups and restrict updates to the expected report author, with regression coverage for a report on page two and a different user's copied marker.

Validation: all 104 Python CI policy tests pass, including the 23 new citation tests. Focused mocked API reproductions confirm duplicate report creation after 100 comments and selection of another user's marked comment for PATCH. Current-head required GitHub checks (`CI required`, `Detect secrets`) pass.

This is a manually requested review posted as BinaryBourbon; it is not an installed Review Loop service verdict. No source changes were made.

scripts/ci/check_issue_refs.py
jhgaylor: Addressed in 3c1646d2: `find_report/3` follows the pages and matches the marker only on a comment by the expected author (`github-actions[bot]` under Actions, the token's user locally); both the upsert and the no-citation correction go through it. Regressions: a report on page two (update and clear), an outsider's pasted marker left untouched, and a full last page followed by an empty one.

scripts/ci/check_issue_refs.py
jhgaylor: Addressed in 3c1646d2: `find_report/3` follows the pages and matches the marker only on a comment by the expected author (`github-actions[bot]` under Actions, the token's user locally); both the upsert and the no-citation correction go through it. Regressions: a report on page two (update and clear), an outsider's pasted marker left untouched, and a full last page followed by an empty one.

BinaryBourbon (approved):
Adversarial re-review verdict: **Approved** for `3c1646d2bdd5134928cbb724b08f4d243b402ac0` (base `74ad641f1f7c2cc06904f67a35a0c3c40ea58f68`).

Four fresh independent Codex reviewers covered all five changed files. Correctness/tests, security, product/API behavior, and simplicity/maintainability each returned APPROVE with no actionable findings.

Both findings from my previous review are addressed:
- The shared `find_report` lookup paginates for both report updates and correction after all citations are removed.
- Report selection requires the marker and the expected author's login, so another user's copied marker is ignored.

Validation: all 110 Python CI policy tests pass, including 29 citation tests. Independent diagnostics placed the genuine report on page three behind copied markers and verified that only that report was updated across repeated runs. Current-head CI and secret scanning pass.

This approval supersedes my changes-requested verdict on `277db4b9`. This is a manually requested review posted as BinaryBourbon, not an installed Review Loop service verdict. No source changes were made.

PR #2177, opened 2026-09-14
feat: expose agent plan snapshots from managoat_acp 0.4.3

### LINKED ISSUE #1571: acp: the `plan` update is dropped, so no client can render a checklist
`Managoat.ACP.Blocks` documents its own gap, at `lib/managoat/acp/blocks.ex:23`:

> | `plan`, `available_commands_update` | dropped — no equivalent yet |

So every ACP adapter that emits a `plan` session update has it thrown away, and
`Fountain.Conversations.Blocks` has no kind to put a checklist in — the nine are
`text thinking tool_use tool_result init result error raw permission_request`.
A client watching an agent work sees the tool calls but never the agent's own
plan, which is the one block that says *where the turn is going* rather than
what it just did.

This is squarely Blocks' job as its moduledoc states it: "the one seam between
what the runtime wrote and what a transcript shows", and ADR 0014's principle
that a client on another origin never re-parses a vendor dialect. Adding it here
gives Conversations, Team and Workbench checklists at once, with no work in any
of the three.

## The part that isn't free

The two runtimes spell it differently, and one of them is incremental:

- **Codex** sends the whole list on every `update_plan`. Straight map to a block.
- **Claude** sends incremental `TaskCreate` / `TaskUpdate` / `TaskList` / `TaskGet`
  calls, each mutating one entry of a list **the transcript never states in full**.
  Rendering these as they arrive gives you a pile of unrelated tool calls, not a
  checklist. The list has to be rebuilt in the parser and each call replaced by
  the resulting snapshot.

So `:plan` wants a `body` that is the full list at that point in the turn, not
the delta that produced it — which also keeps it correct for a client that
joins mid-turn or replays from `/events`.

## Prior art worth reading

siteboon/claudecodeui solved exactly this, both halves, in
`server/shared/message-unification.ts`. It rewrites Claude's incremental task
calls and Codex's whole-list `update_plan` onto one canonical shape so the chat
"needs exactly one renderer per surface" — same conclusion as ADR 0014, reached
from the client side. Its `TASK_TOOLS` set and the snapshot rebuild are the bits
to study.

**Licence:** claudecodeui is AGPL-3.0 and `managoat_acp` is Apache-2.0. Read it
for the approach; do not copy code across.

## Scope

- `:plan` added to `@kinds` and the `to_json/1` wire form in `Managoat.ACP.Blocks`
- the incremental rebuild for claude's task calls, whole-list passthrough for codex
- `Fountain.Conversations.Blocks` inherits it via the `acp` stream; check
  `Fountain.Runtimes.LegacyBlocks` needs nothing for pre-ACP turns
- the kinds enum is API surface (`?blocks=true` on `/events` and the streams), so
  this is an additive wire change — worth a line in the docs

`available_commands_update` is the other half of that table row and is not part
of this; it is a UI affordance, not a transcript block.

jhgaylor:
Agent plan updates were dropped before they reached conversation clients. This pins the published `managoat_acp` 0.4.3 release and exposes its `plan` blocks through Fountain's events and streams. Each `body` is a complete ordered checklist snapshot, so clients can render live output and individual replay events without reconstructing task deltas.

The library implementation and Hex release are complete in managoat/managoat_acp#16. It handles native ACP plans, Claude's incremental task tools, and Codex whole-list plans. Legacy vendor parsing has already been removed from Fountain, so no LegacyBlocks change is needed.

Fountain's block enum and OpenAPI schema now include `plan` and array bodies. The composed body schema puts nullability on the typed string alternative so strict OpenAPI validators accept null as well as text and plan arrays. Regenerated the SDK contract and TypeScript types; the TypeScript turn follower keeps plans out of reply text, and Swift exposes the checklist as `Block.planEntries`.

**TypeScript SDK 5.0.0 is a breaking release.** Public `Block.body` now includes a checklist array. Because the generated Block is not a discriminated union, checking `block.kind` alone does not narrow `body`. The changelog includes migration guidance: use `typeof block.body === "string"` before string operations, and `Array.isArray(block.body)` to read checklist entries. Package, lockfile and User-Agent versions are synchronized. Updated the chat example and docs accordingly.

Validation:
- Library: full `mix precommit`, 245 tests, 0 failures, 95.92% coverage; required CI and Hex publication passed.
- Fountain: focused block/docs and schema-enum guardrail tests; warnings-as-errors compilation, formatting, strict Credo, unused-lock check and generated-contract freshness passed.
- SDK 5.0.0: TypeScript typecheck, all 117 tests, contract verification, build and release guard passed.
- Swift decoding/contract tests and strict format lint; Python and Elixir contract checks passed.
- Strict schema regression reproduced the old nullability compilation failure and now validates null, text, empty/populated plans and rejection of invalid bodies. The raw exported schema also passed direct strict validation.
- The compiled-test runner now links the SDK's `node_modules` into its temp copy, so the strict schema test also runs on the Node 20 leg (it could not resolve `@redocly/ajv` there on fb2a26bf).
- Full CI, secret scanning and the SDK release gate pass on 294c9e8f. Both prior review findings (major-version release, nullable on the typed string alternative) are addressed and their threads resolved; re-review requested.

Closes #1571.


sdk/typescript/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@managoat/fountain-sdk",
-  "version": "4.0.0",
+  "version": "4.1.0",
```
BinaryBourbon: [P2] Release the incompatible Block.body type at a major-version boundary

The public `Block.body` changes from `string | null | undefined` to a union that also includes an array. Existing consumers can fail compilation even when they only handle text blocks:

```ts
function text(block: Block): string {
  if (block.kind === "text") return block.body?.trim() ?? "";
  return "";
}
```

This compiles against the base's generated Block type and fails against this head with TS2339 (`trim` does not exist on the array alternative). `Block` is not a discriminated union, so checking `kind` does not narrow `body`. The PR's own example edits are another instance of callers needing migration.

Publishing this as 4.1.0 makes the incompatible types eligible for existing 4.x dependency ranges, despite the SDK changelog's explicit SemVer contract. Preserve the existing public typing compatibility, or release this shape as 5.0.0 with synchronized package/lock/User-Agent metadata and a Breaking changes entry. Migration guidance should require a body-type check for the current generated shape; checking the kind alone is insufficient. [SemVer's major-version rule](https://semver.org/#spec-item-8) applies to incompatible public API changes.

BinaryBourbon (changes requested):
Adversarial review verdict: **Changes requested** for `9eb7e5931d5317a053ad419cb8ce7864656ccfd9` (base `74ad641f1f7c2cc06904f67a35a0c3c40ea58f68`).

Two independent reviewers covered all 22 changed paths and relevant callers. Both approved their correctness/security source reviews, including the upstream managoat_acp 0.4.3 implementation. My coordinating API compatibility review found the P2 release-boundary defect attached inline: the public TypeScript body type breaks existing consumers but is being published as a minor release.

Validation: TypeScript typecheck, all 116 tests, and contract verification pass. Swift's focused turn-follower/contract run passes all 10 tests. An independent Python diagnostic verified plan preservation, empty snapshots, text exclusion, and foreign-turn filtering. Current-head CI, secret scanning, and the SDK release gate pass. A separate compatibility fixture compiles against the base's public Block type and fails with TS2339 against this head.

Please preserve the existing public TypeScript compatibility or release the wider type as a major version with breaking-change migration guidance. No runtime/security blocker was found in plan propagation.

This is a manually requested review posted as BinaryBourbon, not an installed Review Loop service verdict. No source changes were made.

apps/fountain/lib/fountain_web/schemas.ex
```diff
@@ -1989,7 +1991,27 @@ defmodule FountainWeb.Schemas do
           type: :string,
           enum: Fountain.Conversations.Blocks.kinds()
         },
-        body: %Schema{type: :string, nullable: true},
+        body: %Schema{
+          nullable: true,
```
BinaryBourbon: [P2] Put nullability on a typed oneOf alternative

The old body schema was a nullable string. The replacement puts `nullable: true` on a wrapper with no `type`, while both oneOf branches are nonnullable. Under OpenAPI 3.0 this does not preserve nullability. Against the committed base/head schemas, @redocly/ajv compiles the base but rejects this head with `"nullable" cannot be used without "type"`; OAS30Validator accepts null in the base and rejects it in the head. This breaks downstream schema compilation/validation even though Fountain's own contract checks and generated TypeScript tolerate the definition. No ordinary server emission of a null body was established; the reproduced failure is in strict schema consumers.

Remove nullable from the wrapper and make exactly the typed string branch nullable, leaving the array branch nonnullable. That shape independently validates null, strings, empty/populated plan lists and rejects numeric bodies. Regenerate the contract/types and add a schema-validation regression covering those alternatives. Adding a wrapper `type: string` would reject plans, and making both branches nullable would make null match twice under oneOf.

See [OpenAPI 3.0 nullable and composition rules](https://spec.openapis.org/oas/v3.0.3.html#schema-object).

BinaryBourbon (changes requested):
Adversarial re-review verdict: **Changes requested** for `58bf15aeb8517a0a964ae3eb1394fc12ba6fe0ba` (base `74ad641f1f7c2cc06904f67a35a0c3c40ea58f68`).

The previous release-version finding is resolved: package, lockfile and User-Agent agree on 5.0.0, and the breaking-change guide correctly explains body-type narrowing. Its migration example compiles.

Two independent reviewers covered all 22 changed paths and identified one separate P2 OpenAPI defect attached inline: Block.body uses nullable on a typeless oneOf wrapper. Strict schema tooling rejects that definition, and the resulting schema no longer validates null despite the declared nullable public type.

Validation: current-head TypeScript typecheck, all 116 tests, contract verification and build pass; conformance lint and Python contract verification pass. Base/head probes with OAS30Validator and @redocly/ajv reproduce the schema regression; the second reviewer independently reproduced it from the exported schemas. The focused Elixir block suite passed 7 tests in isolation with the updated dependency source, and additional plan-state/replay diagnostics passed. Moving nullability into the typed string alternative accepts null, strings, empty/populated plan arrays and rejects numeric bodies. Required CI, secret scanning and the SDK release gate pass.

This is a manually requested review posted as BinaryBourbon, not an installed Review Loop service verdict. No source changes were made.

sdk/typescript/test/block-schema.test.ts
```diff
@@ -0,0 +1,34 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import AjvModule from "@redocly/ajv";
```
BinaryBourbon, line 4: [P2] Make the new dependency available to compiled tests

This bare import works from the checkout, but scripts/test-compiled.mjs emits tests into a fresh temporary tree and copies only source, package metadata and fixtures there. It never copies or links node_modules, so installing @redocly/ajv in the original checkout does not make it resolvable from the emitted test. The required Node 20.19.0 job now fails before this test executes with ERR_MODULE_NOT_FOUND; running the compiled runner locally reproduces it. Stage or otherwise resolve the installed test dependencies from the temporary output, then run node scripts/test-compiled.mjs to verify the fix.

BinaryBourbon (changes requested):
Manual adversarial review of fb2a26bfbfffffcdddef7b4dc0820ee89f109803 against 74ad641f1f7c2cc06904f67a35a0c3c40ea58f68, consolidated from two independent whole-PR reviewers and a separate validation pass. This is a human-account review, not an installed Review Loop App verdict.

The previous nullable-composition issue is fixed, and the SDK 5.0.0 bump/migration guidance covers the widened Block.body type. One blocking test-integration defect remains: the new Ajv import cannot resolve from the compiled runner's isolated output tree.

Validation: TypeScript typecheck and contract verification pass; direct SDK tests pass 117/117. Running node scripts/test-compiled.mjs reproduces ERR_MODULE_NOT_FOUND for @redocly/ajv (116 pass, one test file fails). This matches the current required Node 20.19.0 CI failure: https://github.com/managoat/fountain/actions/runs/34837934054/job/103956009608 . The local reproduction used Node 26.7.0. Full Elixir and Swift suites were not rerun in this pass.

Verdict: CHANGES_REQUESTED. Make the installed test dependencies resolvable from the compiled output and rerun the minimum-runtime job.

sdk/typescript/package.json
jhgaylor: Addressed in 58bf15ae: released as 5.0.0 with package.json, package-lock.json and `USER_AGENT` in sync, and a **Breaking changes** entry in `sdk/typescript/CHANGELOG.md` whose migration example checks `typeof block.body === "string"` rather than `kind` alone. The 11:17 re-review confirmed this finding resolved.

apps/fountain/lib/fountain_web/schemas.ex
jhgaylor: Addressed in fb2a26bf: `nullable: true` moved off the wrapper onto the string alternative only, with the array branch nonnullable; contract and generated TypeScript regenerated. `sdk/typescript/test/block-schema.test.ts` compiles the committed `Block.body` projection with `@redocly/ajv` in strict mode and checks that null, `""`, text, `[]` and a populated checklist validate while numbers, booleans, objects, entries missing `status` and an unknown status are rejected. 294c9e8f makes that test resolvable on the Node 20 compiled-test leg, which was the one red job on fb2a26bf.

BinaryBourbon (approved):
Adversarial re-review verdict: **APPROVED** for `294c9e8fe4a74e484a2bbb9a2854bf8227c91020` against base `74ad641f1f7c2cc06904f67a35a0c3c40ea58f68`.

Three independent whole-PR reviewers and a coordinating maintainability/validation review covered all 24 changed paths and relevant callers:

- Correctness: APPROVE. Plan snapshots survive parsing/replay and stay out of assistant text; targeted Elixir diagnostics passed.
- Security: APPROVE. The exact Hex 0.4.3 artifact matches both locked checksums. Session filtering still precedes normalization, and existing tenant-scoped event/stream access and redaction boundaries remain intact.
- API compatibility: APPROVE. Server schema, generated TypeScript, Swift decoding, and the other SDK/CLI consumers agree with the intended snapshot contract.
- Maintainability and validation: APPROVE. No concrete medium-or-higher defect found.

All three earlier findings are resolved: the incompatible TypeScript body type ships at 5.0.0 with migration guidance and synchronized metadata; nullability is on the typed string alternative; the compiled test runner makes installed dependencies resolvable from its temporary output tree.

Validation on this revision: TypeScript typecheck, 117/117 direct tests, 117/117 compiled tests, contract verification and build pass. Focused Elixir diagnostics pass 10 tests using the head's Blocks/Schemas and pinned ACP source with cached dependencies, without booting the application; three additional plan lifecycle/replay diagnostics also pass. Python's 28 conformance scenarios, contract verifier, and shared conformance lint pass. Current-head GitHub CI required, secret scanning and SDK release gate all pass, including Node 20.19.0 and both Swift platforms. Local TypeScript/Python socket fixtures required running outside the restricted sandbox. Full database-backed Elixir and Swift suites were not rerun locally; their evidence comes from current-head CI.

This is the user's explicitly requested manual review posted as BinaryBourbon, not an installed Review Loop App verdict. No source changes were made and no merge was requested.

Issue #2178, opened 2026-09-14
Flake: AuditTest's window assertion shares a 200-row page with audit rows leaked by unboxed_run tests


jhgaylor:
## What fails

`apps/fountain/test/fountain/audit_test.exs:211`

```
1) test _unsafe_list_events/1 — cross-tenant with filters since and until bound
   the window inclusively (Fountain.AuditTest)
   Assertion with == failed
   code:  assert earlier |> planted.() |> Enum.map(& &1.action) == ["old.event"]
   left:  []
   right: ["old.event"]
   stacktrace:
     test/fountain/audit_test.exs:229: (test)
```

## How often, out of what

Locally, on a **reused** test database: the first full-suite run against a
freshly created database passed (6008 tests, 0 failures); every run after
that failed. Four consecutive full-suite runs, including one on an unmodified
`origin/main` (74ad641f) with the same seed, all failed the same way. No CI
run URLs — this is reproducible locally and CI does not see it, for the reason
below.

Reproduce:

```sh
export DATABASE_URL=postgres://postgres:postgres@localhost:5432/fountain_test_flake
MIX_ENV=test mix ecto.create && MIX_ENV=test mix ecto.migrate
mix test            # passes
mix test            # passes or fails
mix test            # fails
```

## The mechanism

It is not a race. It is committed state that survives the SQL Sandbox and
accumulates in the database run over run.

1. Twenty-three test files use `Ecto.Adapters.SQL.Sandbox.unboxed_run/2` to
   exercise real concurrency (`agent_source_deletion_test.exs`,
   `inference_source_write_order_test.exs`,
   `principal_inference_credential_race_test.exs`, the
   `conversations/*_order_test.exs` and `*_isolation_test.exs` set, and
   others). Inside an unboxed run every write **commits**.
2. Those tests create verified users, which audit `account.registered`,
   `credit.granted`, `auth.email.verified`, `agent.created`, `api_key.created`.
   They then delete the users. `audit_events.user_id` is
   `ON DELETE SET NULL`, so the user rows go and the audit rows stay, with a
   null `user_id`.
3. The rows are never cleaned up. After six full-suite runs against one
   database: `users` = 0 rows, `audit_events` = 774 rows, every one with
   `user_id IS NULL`. Those three files alone add 39 rows per run (measured:
   774 → 813).
4. `Audit._unsafe_list_events/1` is deliberately cross-tenant and pages with
[...]

Issue #939, opened 2026-08-22
acp: a permission policy per launch, replacing the constant auto-allow in permission_outcome/1


jhgaylor:
Part of #643. **This is the whole of the safety story that can ship today**, and it ships with no behavior change.

## Where the bypass actually lives

Not in argv any more. Since the ACP conversions (#671–#675) deleted the legacy
spawn path, claude, codex and opencode carry no permission flag at all — they
ask, over `session/request_permission`, and Fountain answers. The answer is a
constant:

```elixir
# acp/peer.ex:747
defp permission_outcome(params) do
  options = Map.get(params, "options") || []

  chosen =
    Enum.find(options, &(&1["kind"] == "allow_always")) ||
      Enum.find(options, &(&1["kind"] == "allow_once")) ||
      List.first(options)
  ...
```

Plus one pre-approval at session setup, `enableAllProjectMcpServers`
(`claude.ex:93`), so the first turn does not stall on a server the sandbox has
no human to approve.

That is the entire rail-off surface for every shippable runtime, in two places
we own. It is not four vendor flags and it is not unchangeable — it is one
function that already has the request, the tool name and the offered options
in its hands, and it throws them away.

## Decided 2026-08-21

Recorded in ADR 0014 gate 3 (#944). These are settled; do not relitigate them
in the PR.

- **Per-tool, not per-launch-only.** The policy is a map of tool → verdict, not
  a single value. This is what makes jhgaylor/fountain-team#3's "Always allow
  `<tool>`" buildable rather than dropped.
- **A launch may narrow the agent's map; it may never widen it.** That single
  rule is the no-escalation guarantee, and it is why this needs **no
  `allowed_permission_*` list** beside it the way `environment_id` needed
  `allowed_environment_ids`. A launch asking for more than the agent grants is
  rejected at the door, not silently clamped.
- **Global default stays `auto_allow`**, so this ships as a no-op for every
  existing conversation.
- **`auto_deny` never synthesises an option.** A `reject_*` kind when the agent
  offered one, `cancelled` when it did not.
- **Audit denials and policy changes; not allows.** A turn makes dozens of tool
  calls and a row per allow would turn the trail into a transcript (ADR 0013).
- **A sprite may not answer its own prompt** — it holds a `FOUNTAIN_TOKEN`, and
  that loop is closed by name rather than left to chance. (Enforced in #940,
  where the endpoint lives; noted here because it constrains the actor model.)

### One thing to measure, not decide

[...]

PR #2179, opened 2026-09-14
refactor(api): the permission policy is one named OpenAPI schema

### LINKED ISSUE #1899: The permission_policy OpenAPI fragment is copy-pasted five times in schemas.ex
Found while reviewing the `stack:1635` set (#1852 → #1860) and deliberately not fixed there, because the fix is a contract decision rather than a cleanup.

## What

`FountainWeb.Schemas` declares `permission_policy` inline in five places — `AgentResponse`, `AgentRequest`, `AgentUpdateRequest`, `Conversation` and `ConversationCreateRequest`. Each copy is now ~22 lines: the `ask_timeout` property with its `minimum`/`maximum` and an 8-line description, plus the `additionalProperties` `oneOf` union of the verdict enum and an integer.

The duplication predates #1852 — the `additionalProperties` enum was already five-way — but that PR roughly tripled the size of the copied fragment, and #1635's ceiling work then had to edit all five again. Any further change to the policy wire shape is a five-place edit with nothing to catch a missed one.

## Why it was not fixed in the stack

The file's own idiom is a named nested schema module referenced by name (`UsageTotal`, and `PendingPermissionRequest` added in #1859). Applying it here is the obvious fix:

```elixir
defmodule PermissionPolicy do
  OpenApiSpex.schema(%{title: "PermissionPolicy", ...})
end
```

But that turns the inline object into a `$ref`, which renames the generated TypeScript type — the five call sites stop producing an anonymous intersection and start producing `components["schemas"]["PermissionPolicy"]`. That is arguably better for SDK consumers, and it is a contract change that wants deciding on its own rather than riding along in a bug fix.

A module attribute would keep the emitted spec byte-identical, but attributes are lexically scoped to the module being defined, so one declared on `FountainWeb.Schemas` is not visible inside the nested `defmodule`s — which is why there is no module-attribute precedent in the file.

## Decide

- Named schema module and accept the generated-type rename (one SDK minor, no runtime change), or
- leave it and accept the five-place edit.

Either is fine; the point of the issue is that the current state is a silent five-way edit surface and nobody has chosen.

Related: #599 derived the enum lists to stop exactly this class of drift.

jhgaylor:
`FountainWeb.Schemas` declared `permission_policy` inline in five places —
`Agent`, `AgentRequest`, `AgentUpdate`, `Conversation` and
`ConversationCreateRequest`. Each copy was the same ~27 lines, so any change to
the policy wire shape was a five-place edit with nothing to catch a missed one.

It is now the `PermissionPolicy` schema module, following the file's own idiom
(`UsageTotal`, `PendingPermissionRequest`), referenced from all five. Net
−134 lines in `schemas.ex`.

## The decision

Named schema module, per the issue's first option, with the generated-type
rename accepted. `permission_policy` is now
`components["schemas"]["PermissionPolicy"] | null` in the TypeScript SDK
instead of an anonymous intersection repeated five times. The **type body is
byte-identical**, so this is a rename and not a new shape; TypeScript is
structurally typed and existing consumer code keeps compiling.

The emitted document is otherwise unchanged — same properties, same
`minimum`/`maximum`, same enum members, same `oneOf`, same descriptions.
Nothing was reworded.

## The five descriptions, and why this is `allOf` rather than a bare `$ref`

They are **not** identical, and none of them is wrong where it stands:

| Schema | What its description says |
|---|---|
| `Conversation` | the stored per-launch override, "or null if it had none" |
| `ConversationCreateRequest` | "may only narrow", 422 `permission_policy_widens`, 422 `permission_policy_unenforceable` |
| `Agent` | "an unset default is auto_allow — today's behaviour", 422 `permission_policy_unenforceable` |
| `AgentRequest`, `AgentUpdate` | same as `Agent`, plus "a conversation may narrow this at launch, never widen it" (identical to each other) |

A single shared description would have to carry the launch-only
`permission_policy_widens` clause onto the response schemas, where a reader
would take it as describing something the response does — which is the case the
brief said to flag rather than silently flatten.

So the call sites are `%Schema{allOf: [PermissionPolicy], nullable: true,
description: ...}` — the shape `usage_total` on `Teammate` already uses in this
file. Each keeps its own prose verbatim; only the shape is shared. Emitted:

```json
"permission_policy": {
  "allOf": [{"$ref": "#/components/schemas/PermissionPolicy"}],
  "nullable": true,
  "description": "<unchanged, per call site>"
}
```

There is a second, harder reason. `OpenApiSpex.schema/1` sets `x-struct`, and
these five schemas are live in `CastAndValidate` on the agent and conversation
controllers. A bare `$ref` would make the cast build a
`%FountainWeb.Schemas.PermissionPolicy{}` — a struct whose only field is
`ask_timeout` — and **silently drop every per-tool key**. `Cast.AllOf` relaxes
`x-struct` to `nil` (`cast/all_of.ex:10`), so the cast keeps returning a plain
map, exactly as the inline object did. `nullable` is read off the wrapper
before the `allOf` dispatch (`cast.ex:138`), so `null` still casts to `nil`.

`agent_controller_test.exs:382` is the proof: it posts
`%{"default" => "auto_allow", "Bash" => "ask"}` through `CastAndValidate`,
asserts the keys round-trip, and asserts `%{"Bash" => "banana"}` is refused.
Both still pass.

## Guardrail

`schema_enum_guardrail_test.exs` loses its five `permission_policy.{}` rows and
gains one `{PermissionPolicy, "{}"}`. Its walker does not follow a module
reference by design, so the enum is now declared and checked in exactly one
place — which is the drift #599 built that test for.

## SDK versioning

The TypeScript SDK goes **5.0.0 → 5.1.0** (it was 4.0.0 → 4.1.0 until #2177 shipped 5.0.0 under it; merged main and re-bumped), not `release:skip-sdk`.

- `sdk/typescript/src/generated/openapi.ts` is under `PUBLISHED_GLOBS`, so
  `scripts/sdk-release.mjs guard` requires a bump or the label. A real type
  change earns the bump.
- **Minor, not major**: `components` is publicly exported
  (`src/schemas.ts:110` → `src/index.ts:32`), so `PermissionPolicy` is a new
  publicly reachable name. Nothing is removed and no shape changes, so it is
  additive. Not a patch, because the public type surface gained a name.
- `CHANGELOG.md` has the `## [5.1.0]` heading and `USER_AGENT` says `5.1.0`;
  `node scripts/sdk-release.mjs guard origin/main` reports "Release looks
  well-formed."
- **Python, Elixir and Swift need nothing.** Their release gates are
  path-filtered to `sdk/python/**`, `sdk/elixir/**` and the Swift package, none
  of which this touches; they vendor no generated types. All three contract
  verifiers pass against the new `contract.json` unchanged. Only
  `manifests/typescript.json` mentions `permission_policy`, as an optional
  field of `ConversationCreateRequest` — still a property, still optional.

`contract.json` moves as expected: five inline objects become
`allOf: [{ref: PermissionPolicy}]` plus one new component. No requiredness,
enum, format or nullability changed.

## Gates run

| Gate | Result |
|---|---|
| `mix format --check-formatted` (root and `apps/fountain`, under mise 1.19.2) | clean |
| `mix compile --warnings-as-errors` | clean |
| `mix credo --strict` | 945 files, no issues |
| `mix test` (`apps/fountain`) | 6008 tests, 0 failures |
| `mix openapi.spec.json` + `jq empty` | ok |
| `scripts/sdk-contract/build.sh` then `--check` | ok, 177 operations / 203 schemas |
| TypeScript: `npm ci`, `generate`, `typecheck`, `test`, `conformance`, `verify-contract` | 115 + 28 tests, 0 failures |
| Python `scripts/verify_contract.py` | ok (67 ops, 41 schemas, 5 enums) |
| Elixir `mix test test/contract_test.exs` | 1 test, 0 failures |
| Swift `swift test --filter ContractTests` | 2 tests, passed |
| `python3 scripts/changelog.py check` | 16 fragments, 18 entries |
| `python3 scripts/conflict-markers.py` | clean |
| `mix precommit` (fresh test DB) | **exit 0** — dialyzer 0 errors, sobelow, prod release assembled, 6008 + 288 tests, 0 failures |

## Review rounds

**Round 1 (41eba420).** Review caught that `{allOf: [$ref], nullable: true}`
does not accept null under OpenAPI 3.0: `nullable` relaxes only the node it
sits on, and the wrapper has no type. Fixed by putting `nullable: true` on the
`PermissionPolicy` component itself, which is the `type: :object` +
`nullable: true` pair the five inline copies always had. The wrappers keep
their own flag because OpenApiSpex answers null there before dispatching to
the composition. `check_nullable_composition` in `scripts/sdk-contract/build.py`
now fails the build on this shape, `scripts/ci/test_sdk_contract_nullable.py`
pins it against fourteen shapes cross-checked with `OAS30Validator`, and
`project_schema` records a component's own `nullable` so the fix is visible in
`contract.json`. Two pre-existing fields in the same state
(`Conversation.sandbox`, `Turn.usage`) are ratcheted in `KNOWN_NOT_NULLABLE`
and tracked in #2189.

**Round 2 (45d8bbfe).** The null-through-the-cast test the review asked for
found that `POST`/`PUT /api/agents` with `"permission_policy": null` was a
500 from the NOT NULL column, on main as well as here. Same contract this PR
guards, so it is fixed here: a null change becomes `{}`. The test asserts
the response is `{}`, because an agent's policy is never null in storage.
`sdk/contract/README.md` documents the nullable trap beside the default trap.

**Merge (69aca252).** #2177 shipped SDK 5.0.0 while this PR carried 4.1.0, so
the four release files conflicted; resolved on main's state with this release
re-bumped to 5.1.0. Re-export and regenerate on the merged tree are
byte-identical to the merge result.

**Round 3 (0689d6ac).** Review caught that the new guard evaluated `oneOf`
as `anyOf`: two branches that both take null make null match twice, and
`oneOf` refuses that, so the guard passed a shape a validator rejects. Now
exactly one branch must accept null; four shapes added to the ground truth,
all eighteen cross-checked against `OAS30Validator`.

## One thing found on the way: #2178

`mix precommit` first came back red on `Fountain.AuditTest` "since and until
bound the window inclusively". It reproduces on unmodified `origin/main`
(74ad641f) with the same seed, so it is not from this change, and it is not a
race: 23 test files use `Sandbox.unboxed_run/2`, whose writes **commit**. They
create verified users (five audit rows each) and then delete them;
`audit_events.user_id` is `ON DELETE SET NULL`, so the audit rows stay. After
six suite runs against one database: `users` = 0, `audit_events` = 774, all
with a null `user_id`. `Audit._unsafe_list_events/1` is cross-tenant and pages
at `limit: 200`, so once that backlog fills the page the test's own hour-old
row falls off the end. CI never sees it because it builds a fresh database;
a workstation database hits it on about the second run.

Filed as #2178 rather than fixed here, so this diff stays about the schema.
`mix precommit` against a fresh database is green.

Closes #1899

https://claude.ai/code/session_01TQor5fs74TZ74L4cbJ1oLz


apps/fountain/lib/fountain_web/schemas.ex
```diff
@@ -468,32 +508,8 @@ defmodule FountainWeb.Schemas do
             "Immutable sandbox callback credential policy. none never issues a callback token."
         },
         permission_policy: %Schema{
-          type: :object,
+          allOf: [PermissionPolicy],
```
BinaryBourbon, line 521: [P2] Preserve nullability inside the composed policy schema

This `allOf` references a non-nullable object schema. `nullable: true` on the typeless wrapper does not relax the referenced schema's constraint under OpenAPI 3.0. The same change occurs on Agent, AgentRequest, AgentUpdate, Conversation and ConversationCreateRequest. Existing agents/conversations without a policy still serialize `permission_policy: null`, and clients can still submit null, so standards-based validation now rejects legitimate payloads.

Reproduced with `openapi-schema-validator` 0.9.0's `OAS30Validator` against the actual base/head committed contracts: null passes all five base properties and fails all five head properties with `None for not nullable` / `None is not of type 'object'`; a valid dynamic tool-policy map passes both. OpenApiSpex's early null handling accepts the new wrapper, so its casting tests do not expose this external-contract regression.

Restore a valid nullable composition while preserving plain-map casting and dynamic tool keys. One independently tested approach is a nullable union with an `allOf` object branch and an explicit null-only branch. Merely adding `type: object` to the outer wrapper leaves the inner non-nullable constraint intact. Regenerate contract/types and add exported-schema null validation alongside the dynamic-key regression. See the [OpenAPI 3.0 schema/composition rules](https://spec.openapis.org/oas/v3.0.3.html#schema-object).

BinaryBourbon (changes requested):
Adversarial review verdict: **Changes requested** for `7114619be7ea1b68b1448d72d25c06f7ba461a11` (base `74ad641f1f7c2cc06904f67a35a0c3c40ea58f68`).

Two independent reviewers covered all nine changed paths and both requested changes for one P2 wire-contract regression: all five policy fields cease to accept null under OpenAPI 3.0 validation, although the server still returns and accepts null. The inline finding includes the reproduction and repair direction.

Validation: TypeScript typecheck, 115 tests and contract verification pass. OpenApiSpex casting probes preserve dynamic tool keys, null, timeout bounds and invalid-verdict rejection. An independent OAS30Validator probe of the committed base/head contracts accepts null in all five base fields and rejects it in all five head fields; valid policy maps pass both. A second reviewer exported the schemas directly from both revisions and independently reproduced the failure with @redocly/ajv. Current-head CI, secret scanning and SDK release gate pass.

This is a manually requested review posted as BinaryBourbon, not an installed Review Loop service verdict. No source changes were made.

jhgaylor:
Confirmed and fixed in 41eba420. The finding was right, and my verification in the first pass was the wrong kind for this class of bug.

## The fix

`nullable: true` moves onto the `PermissionPolicy` component. That is what the five inline copies it replaced said all along — each wrote `type: :object` and `nullable: true` on one node, and the extraction had separated the pair. One line.

```elixir
OpenApiSpex.schema(%{
  title: "PermissionPolicy",
  type: :object,
  nullable: true,      # <- here, not only on the five wrappers
  ...
})
```

The wrappers keep `allOf: [PermissionPolicy], nullable: true, description: ...`. The wrapper's own `nullable` stays because OpenApiSpex answers null at `cast.ex:138`, before it dispatches to the composition — dropping it would push null into `Cast.AllOf` with a nil accumulator.

I checked your warning about `type: :object` on the wrapper, and it is correct — the branch still refuses. Row D below.

## External-validator evidence

`openapi-schema-validator` 0.9.0, `OAS30Validator`, against the real exported `dist/openapi.json`. Before (commit 7114619b):

```
schema                         null       policy map   bad verdict
------------------------------------------------------------------------------
Agent                          REJECT     ACCEPT       rejected
    null  -> None for not nullable
AgentRequest                   REJECT     ACCEPT       rejected
    null  -> None for not nullable
AgentUpdate                    REJECT     ACCEPT       rejected
    null  -> None for not nullable
Conversation                   REJECT     ACCEPT       rejected
    null  -> None for not nullable
ConversationCreateRequest      REJECT     ACCEPT       rejected
    null  -> None for not nullable

RESULT: FAILURE
```

After (41eba420):

```
schema                         null       policy map   bad verdict
------------------------------------------------------------------------------
Agent                          ACCEPT     ACCEPT       rejected
AgentRequest                   ACCEPT     ACCEPT       rejected
AgentUpdate                    ACCEPT     ACCEPT       rejected
Conversation                   ACCEPT     ACCEPT       rejected
ConversationCreateRequest      ACCEPT     ACCEPT       rejected

RESULT: all five accept null and a policy map, and reject a bad verdict
```

"policy map" is `{"default": "auto_allow", "Bash": "ask", "ask_timeout": 3600}` and "bad verdict" is `{"Bash": "banana"}`, so the same run shows the dynamic per-tool keys still validate and an unknown verdict is still refused. The casting half is unchanged and still covered by `agent_controller_test.exs:382`.

## Why I did not take the union form

I tested it, and it works — but so does the one-line component fix, and it emits less. All six shapes against `OAS30Validator`:

| shape | null |
|---|---|
| A0 `{nullable, allOf:[$ref]}` (what you caught) | REJECT |
| A `nullable` on the component | **ACCEPT** |
| B `{nullable, oneOf:[{allOf:[$ref]}, {nullable, enum:[null]}]}` (yours) | **ACCEPT** |
| C same with `anyOf` | **ACCEPT** |
| D `{type: object, nullable, allOf:[$ref]}` | REJECT |

B and C add a two-branch union at five call sites and turn the generated TypeScript into a union; A restores the base document's exact `type: object` + `nullable: true` pairing. Happy to switch if you prefer the explicit null branch.

## The blind spot, closed

You are right that my casting probes could not see this, so the guard is document-level:

- **`check_nullable_composition`** in `scripts/sdk-contract/build.py` fails the build when a property says `nullable: true` and a standards validator would still reject null. It runs on the exported document in `release-and-contract`.
- **`scripts/ci/test_sdk_contract_nullable.py`** pins the rule against 14 shapes — every one cross-checked against `OAS30Validator` — plus the ratchet behaviour and the five fields as they stand in the committed projection. Runs in `workflow-checks` on bare `python3`; the third-party validator is deliberately not a test dependency, so its agreement is recorded rather than re-derived.
- **`project_schema` now records a schema's own `nullable`.** Worth knowing: without that, this fix diffed `contract.json` to **nothing**. The projection recorded nullability on properties but not on components, so the committed artifact was byte-identical either way. `sdk/contract/README.md` lists nullability as part of the contract; now it is.

I verified the guard by reverting the one-line fix and rebuilding: `build.sh` exits 1 naming all five fields, the external validator goes back to REJECT on all five, and doctoring `contract.json` fails `test_the_component_is_nullable`. All three layers go red.

## Two more of the same class, not fixed here

The guard immediately found:

```
Conversation.sandbox  {"nullable": true, "oneOf": [{"$ref": ".../Sandbox"}]}    null -> REJECT
Turn.usage            {"nullable": true, "oneOf": [{"$ref": ".../TurnUsage"}]}  null -> REJECT
```

Both confirmed with `OAS30Validator`, both real — `Turn.usage`'s own description says "null while the turn runs". I did **not** fix them here: `Sandbox` and `TurnUsage` are referenced from other places, so hoisting `nullable` onto those components is a different judgement than it was for `PermissionPolicy`, and it would put unrelated schemas in this diff. They are recorded in the guard's `KNOWN_NOT_NULLABLE` ratchet and tracked in #2189. The list only shrinks — a new one fails the build, and so does a stale entry.

Say the word if you would rather they landed here.

## Versioning

Still 4.1.0 minor. The five properties resolve to exactly the 4.0.0 type (`components["schemas"]["PermissionPolicy"] | null` where the named type now carries the `| null`), so nothing breaks; the added public name is what earns the minor. CHANGELOG wording updated to say the named type includes null.

## Gates

`mix precommit` exit 0 on a fresh database — dialyzer 0 errors, sobelow, prod release assembled, 6008 + 288 tests, 0 failures. `scripts/ci` 94 tests OK. TypeScript typecheck / 115 tests / 28 conformance / verify-contract. Python, Elixir and Swift contract verifiers. `build.sh --check` ok.

Leaving the verdict to you.

jhgaylor:
Round two, independently re-verified from a clean worktree at 41eba420, then one more commit.

## The finding, re-checked from scratch

Exported `dist/openapi.json` from the branch and ran `openapi-schema-validator` 0.9.0's `OAS30Validator` against each of the five `permission_policy` properties with the document's `components` attached:

```
schema                       null     policy   bad verdict
Agent                        ACCEPT   ACCEPT   REJECT
AgentRequest                 ACCEPT   ACCEPT   REJECT
AgentUpdate                  ACCEPT   ACCEPT   REJECT
Conversation                 ACCEPT   ACCEPT   REJECT
ConversationCreateRequest    ACCEPT   ACCEPT   REJECT

ratchet Conversation.sandbox   null -> REJECT
ratchet Turn.usage             null -> REJECT
```

Then the guard, both ways: with `nullable` stripped from the `PermissionPolicy` component in a copy of the document, the validator goes back to REJECT on all five and `build.py --check --spec <copy>` exits 1 naming all five fields. Restored, `build.sh --check` is `ok (177 operations, 203 schemas)`. The two ratchet entries are still genuinely broken, so `KNOWN_NOT_NULLABLE` is accurate as committed.

## What the null test found: 45d8bbfe

The review asked for null validation beside the dynamic-key regression in `agent_controller_test.exs`. Writing it — `PUT /api/agents/{id}` with `"permission_policy": null` — turned up that the null the document has advertised since #939 was a **500** on the agent doors:

```
** (Postgrex.Error) ERROR 23502 (not_null_violation)
   null value in column "permission_policy" of relation "agents"
```

The cast accepted it, the changeset accepted it, and the NOT NULL / `{}`-default column refused it. Pre-existing (the base document says nullable on `AgentRequest` and `AgentUpdate` too), but it is the same contract this PR hoists onto the component and guards, so it is fixed here rather than left for a third round: a null change becomes `{}`, which is what "no policy" is for an agent. `put_change` drops a value equal to the stored one, so a no-op PUT records no change. Verified by reverting the one pipeline line — the test goes back to the not-null violation.

The response after the PUT is `{}`, not `null`: an agent's policy is never null in storage, and the test says so. `Conversation.permission_policy` is a nullable column and was already right.

Also in that commit: `sdk/contract/README.md` gets a "nullable trap" section beside the "default trap" it already documents, so both checks `build.py` enforces are written down where the contract is explained. Changelog fragment carries the null fix.

## Gates, on the fresh isolated database `fountain_test_2179`

| Gate | Result |
|---|---|
| `mix compile --warnings-as-errors`, `mix format --check-formatted` (root and app), `mix credo --strict` | clean |
| `MIX_ENV=dev mix dialyzer` | Total errors: 0 |
| `mix test` (umbrella root: core, ee, buzz, support) | core 6009 tests 0 failures; buzz 136, support 33, slack 30, google 60, microsoft 29, all 0 failures |
| `python3 -m unittest discover -s scripts/ci` | 94 tests OK |
| `scripts/sdk-contract/build.sh --check` | ok |
| TypeScript `generate` (no diff), `typecheck`, `test` 115, `conformance` 28, `verify-contract` | all pass |
| `node scripts/sdk-release.mjs guard origin/main` | "Release looks well-formed. Merging this publishes 5.1.0." (was 4.1.0; #2177 shipped 5.0.0 under this PR, so 69aca252 merges main and re-bumps) |
| `python3 scripts/changelog.py check`, `scripts/conflict-markers.py` | clean |

One workstation note, not a PR finding: the shared `fountain_test` database on this machine carries committed sandbox rows from other worktrees' `unboxed_run` tests, and the suite run against it fails 347 tests with `503 fleet_full` and friends. Against a fresh database the same tree is green. That is #2178's neighbourhood, not this change.

scripts/sdk-contract/build.py
```diff
@@ -358,6 +365,138 @@ def check_defaults(contract: Dict[str, Any]) -> List[str]:
...
+    """
+    node = resolve_ref(spec, node)
+    if not isinstance(node, dict) or depth > 20:
+        return False
+
+    for keyword, needs_all in (("allOf", True), ("anyOf", False), ("oneOf", False)):
+        branches = node.get(keyword)
+        if isinstance(branches, list) and branches:
+            results = [accepts_null(spec, branch, depth + 1) for branch in branches]
+            if (all(results) if needs_all else any(results)) is False:
```
BinaryBourbon: [P2] Require exactly one matching branch for `oneOf`

`oneOf` takes the `any(results)` path here, so the new guard reports a nullable schema as valid when null actually matches multiple branches. For example:

```json
{
  "nullable": true,
  "oneOf": [
    {"type": "object", "nullable": true},
    {"type": "string", "nullable": true}
  ]
}
```

On this head, `accepts_null(spec, node)` returns `True` and `check_nullable_composition(spec)` produces no finding for a property with this shape (with the unrelated ratchet entries cleared in the isolated probe). `openapi-schema-validator==0.9.0` independently returns `OAS30Validator(node).is_valid(None) == False`: null matches both branches, whereas [OpenAPI `oneOf` requires exactly one](https://spec.openapis.org/oas/v3.0.3.html#discriminator-object). The new build gate therefore permits the nullable-contract mismatch it is intended to catch. This is a defect in the new guard, not a remaining failure of the five repaired permission-policy fields.

Evaluate `oneOf` with exactly one true result (for example `sum(results) == 1`), retain `any(results)` for `anyOf`, correct the matching docstring, and add an overlapping-nullable-branches regression. All 13 current nullability tests pass without exercising this case.

BinaryBourbon (changes requested):
Adversarial review verdict: **Changes requested** for `69aca2523463d60f956e820ebe02220759c5b721` (base `913c52414ccc4bb0d5416ab5ebe6a439053acb3f`).

Three independent reviewers examined the full 14-file diff for correctness, security/tenant isolation, and API compatibility, with an additional coordinating maintainability pass. They agree on one P2 defect in the newly added nullability guard: `oneOf` is evaluated as `anyOf`. The inline finding contains the independently reproduced validator disagreement and the requested repair.

The earlier permission-policy nullability regression is repaired. The named component retains `type: object` with `nullable: true`, dynamic policy keys survive casting, and agent null changes normalize to an empty policy. No additional introduced tenant or permission-enforcement bypass was found. TypeScript release metadata consistently updates 5.0.0 to 5.1.0 after the merge from main.

Validation: the current head's Python CI suite passes all 123 tests, including the 13 new nullability tests; Python and TypeScript contract verifiers pass. A separate `openapi-schema-validator` 0.9.0 probe at this head rejects null for the overlap counterexample while the new guard accepts it. Earlier-head in-memory runtime diagnostics passed 30 casting probes across the five fields and checked null create/update, omission, repeated clear and invalid verdict handling; the merge leaves the agent changes and permission-policy schema fragment unchanged. These probes are not database integration tests. Updated-head CI is still running; passing jobs include the API contract, SDK release gate, secret scan and both TypeScript SDK jobs.

This is a manually requested review posted as BinaryBourbon, not an installed Review Loop service verdict. No source changes were made.

scripts/sdk-contract/build.py
jhgaylor: Confirmed and fixed in 0689d6ac. `oneOf` now requires exactly one null-accepting branch (`sum(results) == 1`); `anyOf` keeps `any()`, `allOf` keeps `all()`, docstring corrected. Four shapes added to the ground truth, each cross-checked against `OAS30Validator` (18 of 18 agree): your overlapping pair under `oneOf` (rejects) and under `anyOf` (accepts), `oneOf` with one nullable branch of two (accepts), and `oneOf` over a nullable `$ref` plus an explicit null branch (rejects). With the `oneOf` line reverted, two of the four fail. `build.sh --check` on the real document is still ok, so `KNOWN_NOT_NULLABLE` is unchanged.

jhgaylor:
Round three: the `oneOf`-as-`anyOf` defect in `accepts_null` is fixed in 0689d6ac, with the overlapping-branches regression and an 18-of-18 cross-check against `openapi-schema-validator` 0.9.0. Reply on the inline thread has the detail. Python CI suite: 123 tests OK. The real document still passes `build.sh --check` with no ratchet change.

changelog.d/1899-permission-policy-schema.md
```diff
@@ -0,0 +1,26 @@
+### Changed
+
+- The OpenAPI document declares the permission policy once, as the
+  `PermissionPolicy` component, instead of repeating it inline on `Agent`,
+  `AgentRequest`, `AgentUpdate`, `Conversation` and `ConversationCreateRequest`
+  (#1899). The wire shape is unchanged: the five properties are now a `$ref` to
+  that component, keep their own descriptions, and still accept the `null` a
+  policy-less agent or conversation carries. A generated client gains a named
+  type where it had an anonymous object with the same body. TypeScript SDK
+  4.1.0 follows.
```
BinaryBourbon, line 10: [P3, nonblocking] Update the SDK release reference to 5.1.0

After the merge from main, package.json, package-lock.json, USER_AGENT and the TypeScript changelog consistently release 5.1.0 over 5.0.0. This fragment still says 4.1.0, so the rolled server changelog would point readers at the wrong SDK release. Update this reference to 5.1.0. This documentation mismatch does not block approval.

BinaryBourbon (approved):
Adversarial review verdict: **Approved** for `0689d6ac8a01442a7799dcfda900c8589c448377` (base `913c52414ccc4bb0d5416ab5ebe6a439053acb3f`).

Three fresh independent reviewers covered all 14 changed paths for correctness, security/tenant isolation, and API compatibility, with a coordinating maintainability pass. No blocking findings remain. Both earlier findings are repaired: the shared PermissionPolicy component preserves nullability, and the document guard now requires exactly one matching branch for oneOf. Agent null normalization preserves the existing policy-clearing behavior and tenant boundaries.

Fresh validation: all 123 Python CI tests pass; all 18 nullability fixture shapes agree with independent OAS30Validator 0.9.0 validation; Python and TypeScript contract verifiers pass. Resolving the five extracted PermissionPolicy references makes the entire committed contract equal to the base contract. All applicable current-head GitHub checks pass, including CI required, the six server test partitions, static analysis, API contract, SDK checks, SDK release gate, and secret scanning. Local Elixir/database integration tests were not rerun in this review environment; that coverage comes from the current-head CI results.

One nonblocking P3 documentation correction is noted inline: the server changelog fragment still says SDK 4.1.0, while the release metadata consistently specifies 5.1.0.

This is a manually requested review posted as BinaryBourbon, not an installed Review Loop service verdict. Approval covers this revision only; no source changes or merge were performed.

Issue #2187, opened 2026-09-14
refactor(conversations): one helper builds the resolver call for admission, wake and provision


jhgaylor:
Stage 6 of #2176. Depends on stage 4 (#2185).

## What

Admission (`conversations.ex:4988-5000`), wake (`:5010-5020`), provision (`conversations/sprite_env.ex:43-57`) and `validate_source/2` (`inference_credentials.ex:617-629`) each build the `resolve/4` call by hand, and two of them pass `credential_set_id` beside an `expected_source` that already carries `set_id` (the resolver ignores the former whenever the latter is present, `resolver.ex:10`). One helper builds the call, in a new small module (`Fountain.Conversations.InferenceResolution` or similar), **not** in `conversations.ex` (the size pin) and not in `SpriteEnv`.

Two shapes:

- **a new selection**: `credential_set_id` (the launch's override, else the agent's set, else nil for the account default), `environment_id`, `vault_id`;
- **a re-validation**: `expected_source` and the environment/vault it names, and nothing else.

`SpriteEnv.credential_set_id/2` (`sprite_env.ex:36-41`) moves in as the "which set" rule for the first shape.

## Proves unchanged

`launch_credential_set_test.exs`, `inference_admission_isolation_test.exs`, `inference_source_binding_test.exs`, the wake/resume tests in `conversations_test.exs` that assert `:inference_source_changed`.

## Size

Under 250 lines. `conversations.ex` shrinks. No table changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SmajkEG6N13cDrrDVUJHsx

Issue #2186, opened 2026-09-14
refactor(inference): derive Source.origin from scope; the stored key stays


jhgaylor:
Stage 5 of #2176. Independent, but lands after stage 4 (#2185) to avoid a conflict in `source.ex`.

## What

`Source.origin` is derivable: `:platform` is set by exactly one constructor (`inference_credentials/source.ex:28`) and `:own` by the other four, so `origin == (scope == :platform)` for every value the code constructs. Drop the struct field; keep the stored shape.

Decision 2 of #2176: **the key stays in the dumped map, unchanged, and no migration runs.**

1. Remove `:origin` from `defstruct` and `@enforce_keys` (`source.ex:11-23`). `platform?/1` reads `scope` (`:30`).
2. `dump/1` still writes `"origin"` (`"platform"` when scope is `:platform`, else `"own"`), so `conversations.inference_source`, `turns.inference_source` and `sandboxes.codex_inference_source` keep the shape every stored row has, and the `matches/2` comparison of a dumped source with an `expected_source` (`resolver.ex:63`) keeps working on rows written before this change.
3. `load/1` ignores the stored `"origin"` key.
4. `TurnMachine.with_inference/2` (`conversations/turn_machine.ex:900`, `:905`) matches on `scope` (`:platform` versus the four own scopes) instead of `origin`.
5. Tests that pin `origin: :platform` / `origin: :own` on the struct (`platform_chatgpt_test.exs`, `ee/test/fountain/credits_chatgpt_inference_test.exs:29`, `conversation_server_platform_inference_test.exs:183-359`) pin `scope` instead.
6. **Add the test the memo found missing:** `Source.load/1` over a literal map in the stored shape (every key, string-valued, as a row written on 2026-09-13 has it), asserting the struct and that `dump(load(map)) == map`. Renaming a `kind` atom would make `load/1` decode `nil` and every wake refuse with `:inference_source_changed`; nothing loads a historical map today.

## Why the key stays

The API and SDK do not expose the inference source's origin (the `origin` in `sdk/contract/contract.json` is the turn's `user`/`autonomous`), so removing the key is possible; it buys nothing but a forward migration over three JSON columns and a loader that must accept both shapes for the life of the rows. Not worth it.

## Proves unchanged

The three test files above, `inference_source_binding_test.exs`.

## Size

Under 150 lines. No table changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

Issue #2183, opened 2026-09-14
refactor(inference): the platform step moves into PlatformInference, and a runner conversation on a brokered deployment gets its test


jhgaylor:
Stage 3 of #2176. Independent of stages 1 and 2.

## What

The resolver's "platform policy" step names `Fountain.ChatGPTAccounts` and queries the `platform_chatgpt_account` table directly. Move that step into `Fountain.PlatformInference`, which already owns the one deployment-credential rule the code has (a stored row beats the env var, `key_for/1`), so the resolver asks one question of one module: *the deployment's credential for this provider and runtime*.

Paths under `apps/fountain/lib/fountain/`; lines against `main` at `05c18224`.

1. `Resolver.platform/4` (`inference_credentials/resolver.ex:155-167`) becomes a call to a new `PlatformInference.credential_for(provider, runtime)` returning `{:ok, kind, value} | :none`. The body moves: `openai` + `codex` on a brokered deployment takes `ChatGPTAccounts.platform_credential(refresh: false)`, else `key_for/1`. `PlatformInference` reads `Fountain.Broker.configured?/0` itself; the resolver stops forwarding `brokered` (`resolver.ex:32`, `:157`).
2. `Resolver.platform_reference/3` (`resolver.ex:247-277`, the identity/revision binding for a platform source, including the `Repo.one(from a in Fountain.PlatformChatGPT.Account, ...)` read) moves next to it as `PlatformInference.reference(kind, creds, dek)`. After this, `resolver.ex` aliases neither `ChatGPTAccounts` nor `PlatformChatGPT.Account`.
3. The `platform_chatgpt_test.exs:426` test "an unbrokered conversation never takes the grant" simulates an unbrokered deployment by passing `brokered: false`. It now configures the broker off for the test the way the egress tests do, so it tests the deployment fact rather than the seam.

## The test decision 4 of #2176 asks for

Add, in `conversation_server_platform_inference_test.exs` or beside it: **a conversation on a self-hosted runner, on a brokered deployment, under `BROKER_ALLOW_UNENFORCED`, selects the ChatGPT grant, and the sandbox environment carries the `__codex_chatgpt_access_token__` placeholder and never the token.** That is the case the memo's risk table found no test for. Decision 4 (deployment fact) is what makes it the expected behaviour: credentials are brokered on both `contained` and `direct` sandboxes (#2057), so the grant is safe wherever the broker carries it.

## Proves unchanged

[...]

Issue #2182, opened 2026-09-14
refactor(inference): delete the two callerless wrappers, PlatformInference.gate/4 and SpriteEnv.select_inference/4


jhgaylor:
Stage 2 of #2176. Independent of stage 1; either may land first.

## What

Delete the two public wrappers that exist only for their own tests, and rewrite those tests against the functions production actually calls.

Paths under `apps/fountain/lib/fountain/`; lines against `main` at `05c18224`.

1. **`PlatformInference.gate/4`** (`platform_inference.ex:334-341`). No caller in `lib/` or `ee/lib`. Production admission calls `InferenceCredentials.resolve/4` and then `PlatformInference.gate_source/1` (`conversations.ex:4988-5000`). Delete `gate/4` and its `@doc`; keep `gate_source/1` and `check_ceiling/0`. Its fifteen test references (`platform_inference_test.exs` ×12, `platform_chatgpt_test.exs` ×2, `conversations/inference_named_set_gate_test.exs` ×1) become `resolve/4` + `gate_source/1`, which is what they were testing through the wrapper.
2. **`SpriteEnv.select_inference/4` and `load_tenant_state/2`** (`conversations/sprite_env.ex:59-96`). No caller outside tests (`sprite_env_inference_test.exs` ×16, `agents_credential_set_test.exs` ×6, `sprite_env_test.exs` ×2; `test/support/conversation_server_case.ex:111` names `load_tenant_state/2` in a comment). Delete both. Tests that exercised selection through `select_inference/4` call `resolve/4` with a persisted environment/vault, or `resolve_inference/4` where they were really testing provisioning. The `overrides:` option of `select/4` loses its only caller here (`sprite_env.ex:90`); leave the option for stage 4 to remove with `select/4`.

## Note for #1709

The per-account daily bound decided on #1709 names `PlatformInference.gate/2`. After this stage the gate is `gate_source/1`; the bound lands there (it will need the user, so `gate_source/2`), not in a revived `gate/4`.

## Proves unchanged

The rewritten tests must cover every case the wrapper tests covered (ceiling, explicit named set, grant counted as platform, tenant key wins, unusable explicit selection). `launch_credential_set_test.exs`, `inference_admission_isolation_test.exs`.

## Size

Under 400 lines, most of it test rewrites. No table changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SmajkEG6N13cDrrDVUJHsx

PR #2195, opened 2026-09-14
refactor(inference): one resolver entry point, four options, a struct inside

### LINKED ISSUE #2185: refactor(inference): one resolver entry point, four options, a struct inside
Stage 4 of #2176. Depends on stages 1, 2 and 3 (#2181, #2182, #2183).

## What

One public entry point with four options, and a struct instead of a keyword list between the resolver's two halves.

Paths under `apps/fountain/lib/fountain/`; lines against `main` at `05c18224`.

1. **`InferenceCredentials.select/4` goes.** After stage 2 its only production caller is the verified-landing banner, `start_live.ex:120` (`needs_credential?/2`), which already holds the agent and can call `resolve/4` with `credential_set_id: agent.inference_credential_id` and match `scope: :missing`. `resolve/4` runs one short transaction under the source lock per dashboard render, which the banner tolerates; measure it once on the rig before merging and say the number in the PR. The tests that call `select/4` directly (`platform_chatgpt_test.exs`, `conversation_server_platform_inference_test.exs:369`, the ee credits test) call `resolve/4` with an inserted set.
2. **The option list is exactly** `credential_set_id`, `environment_id`, `vault_id`, `expected_source`. `brokered` went in stage 3; `refresh`, `allow_missing`, `secret_keys` in stage 1; `overrides` and `override_entries` were never external. Unknown keys raise (a `Keyword.validate!/2`), so the next dead option cannot grow back silently.
3. **`Resolver.select/4` becomes a private function taking a struct** (`%Resolver.Inputs{}` or similar: model, runtime, own, overrides, provider) rather than the merged keyword at `resolver.ex:28-33`. No behaviour change.
4. The `select/4` docstring (`inference_credentials.ex:546-563`) and the `resolve/4` line get one `@doc` describing the order the resolver applies, the one #2176 §1 lists. That paragraph is the reference the ADR corrections (stage 7) point at.

## Proves unchanged

`launch_credential_set_test.exs`, `inference_admission_isolation_test.exs`, `inference_named_set_gate_test.exs`, `inference_source_binding_test.exs`, and the verified-landing LiveView test for the banner.

## Size

Under 400 lines. No table changes.

https://claude.ai/code/session_01SmajkEG6N13cDrrDVUJHsx

jhgaylor:
Stage 4 of #2176. Base: `stack/2176-3-platform-step` (#2183); depends on #2181, #2182 and #2183.

## What changed

- **`InferenceCredentials.select/4` is deleted.** Its one production caller, the verified landing's banner (`StartLive.needs_credential?/2`), calls `resolve/4` with `credential_set_id: agent.inference_credential_id` and shows the banner for a `:missing` source. It also shows it for `{:error, :inference_credential_unusable}`: under an explicit set the resolver refuses `:missing` and `:platform` rather than substituting, so an agent pointed at a set that holds nothing would be refused at launch too, and that is the banner's question. (Today's banner said "fine" for that agent when a platform key was configured, while the launch would have been refused; it now says what the launch will do.)
- **The option list is exactly** `credential_set_id`, `environment_id`, `vault_id`, `expected_source`, enforced by `Keyword.validate!/2`; a test proves an unknown key raises. `refresh`, `brokered`, `overrides` and `override_entries` no longer exist. `PlatformInference.credential_for/2` (was `/3`) answers from the row with `refresh: false` fixed, as the resolver always asked.
- **`Resolver.select/1` is private and takes `%Resolver.Inputs{}`** (provider, runtime, the set's own values, the normalized overrides); `accepted/2` is private; `selected_overrides/2` is gone. No behaviour change in selection.
- **`resolve/4` has the `@doc`** stating the seven-step order (#2176 §1) and a `@spec`; the `select/4` docstring is gone with it. Comments in `PlatformInference`, `ChatGPTAccounts`, `CodexTransport`, `InferenceCredentials` and `StartLive` that named `select/N` name the resolver or `credential_for/2`.
- **Tests** that called `select/4` with a map of values resolve against persisted rows: the seven selection cases in `platform_inference_test.exs` (plus the stored-key case) with `put_credential/4`, the empty-value case by writing an encrypted `""` into a set the way `agents_credential_set_test.exs` does; the five grant cases in `platform_chatgpt_test.exs` (the revoked-grant case revokes through `platform_access_token/0` first, since the resolver never dials out); the two alias cases in `inference_source_binding_test.exs` with vaults; the codex-turn stamp test and the ee credits test with the user they already had.

## Banner timing

The issue asks for the cost of one `resolve/4` per dashboard render. Measured on the lane's test database with a throwaway test (`:timer.tc` around the banner's exact call, twenty iterations after one warm-up, starter agent, no set):

| min | median | max |
|---|---|---|
| 287 µs | 302 µs | 452 µs |

One short transaction under the source lock; the previous code did a DEK load and a set decrypt outside any lock, so the render pays one advisory lock pair more.

## What stayed

`Source.origin` (#2186). The helper that builds the call at admission, wake and provision (#2187).

## Verification

`mise exec -- mix precommit` at the umbrella root on `fountain_test_2176a`:

```
6 doctests, 6010 tests, 5 failures, 2 skipped
```

The five failures are the environmental trust-store ones of the earlier stages; extension suites buzz 136, slack 30, support 33, google 60, microsoft 29 tests, 0 failures.

Reverted to verify: with the `Keyword.validate!/2` line removed, the unknown-option test fails; restored, it passes. Targeted files before the gate: 147 tests, 0 failures (the banner LiveView tests included).

Changelog fragment `changelog.d/2185-banner-agrees-with-launch.md` (`### Fixed`) records the banner change above; everything else is internal.

Part of #2176
Closes #2185


apps/fountain/lib/fountain_web/live/start_live.ex
```diff
@@ -102,29 +102,26 @@ defmodule FountainWeb.StartLive do
...
-        )
-      )
-    else
-      # A tenant key that will not load is a bigger problem than this banner,
-      # and it is not this page's to report.
+    # default set holds. Under an explicit set the resolver refuses that
+    # case as unusable rather than substituting the platform key, so a
+    # launch would be refused too: the same answer for this banner.
+    case InferenceCredentials.resolve(user_id, agent.model, agent.runtime,
+           credential_set_id: agent.inference_credential_id
```
BinaryBourbon: [P2] Include the agent environment in this resolution. With an empty explicit set, an environment ANTHROPIC_API_KEY, and a configured platform key, the base hides this banner but this revision shows it because the set-only resolution returns :inference_credential_unusable. Admission passes agent.environment_id and successfully selects :tenant_secret. Add a regression for this working configuration.

BinaryBourbon (changes requested):
Adversarial whole-PR review of `fb20243605cb00dcadea25f87e0f8f61ccff5ccc` against `1bc69f29f1a6713249698dcc2817412b53af6717`.

**Changes requested: preserve environment credentials when deciding whether to show the banner (P2).** In `StartLive.needs_credential?/2`, the resolver receives the explicit set but not `agent.environment_id`. With an empty named set, a valid `ANTHROPIC_API_KEY` in the agent's environment, and a configured platform key, the previous banner stays hidden while this revision shows “no inference credential yet.” Actual admission passes the environment and successfully selects `:tenant_secret`. Thus the new explicit-set `:inference_credential_unusable` branch introduces a false warning for a working launch. Pass the agent's environment into the banner resolution and add a LiveView regression with those three conditions.

Evidence: an isolated Elixir diagnostic loaded the actual base/head resolver and extracted banner functions, with in-memory database/provider boundaries. The same fixture returned `{banner_needed: false, launch_scope: :tenant_secret}` on the base and `{banner_needed: true, launch_scope: :tenant_secret}` on the head. This is a controlled source diagnostic, not a full Phoenix/database integration test.

Inspected selection precedence, alias normalization, explicit-set refusal, platform grant selection without refresh, all removed-function callers, persisted-row test replacements, and the banner's admission path. GitHub checks are complete with success/skipped conclusions at review time. The local full Mix diagnostic could not run in these isolated worktrees because dependencies are absent; I have not reproduced the PR author's complete precommit run. No Review Loop App summary or prior review was present. This is a code-review verdict, not an assertion that the separate Review Loop verification policy completed.

<details>
<summary>Reviewed changed paths</summary>

- `apps/fountain/lib/fountain/conversations/codex_transport.ex`
- `apps/fountain/lib/fountain/inference_credentials/resolver.ex`
- `apps/fountain/lib/fountain_web/live/start_live.ex`
- `apps/fountain/test/fountain/conversations/conversation_server_platform_inference_test.exs`
- `apps/fountain/test/fountain/inference_source_binding_test.exs`
- `apps/fountain/test/fountain/platform_chatgpt_test.exs`
- `apps/fountain/test/fountain/platform_inference_test.exs`
- `changelog.d/2185-banner-agrees-with-launch.md`
- `ee/test/fountain/credits_chatgpt_inference_test.exs`

</details>

apps/fountain/lib/fountain_web/live/start_live.ex
jhgaylor: Fixed in bfcd097f. `needs_credential?/2` now passes `environment_id: agent.environment_id` beside the set, the way admission does, so an environment key clears the banner for an explicit set that holds nothing.

Two regressions. `start_live_test.exs` ("an agent whose set is empty but whose environment holds the key is not") stays async and needs no platform key: the empty explicit set is refused as unusable either way, and the environment key is what clears it. `start_live_platform_banner_test.exs` is a new `async: false` module (the platform key is global application env, the same reason `platform_inference_test.exs` is `async: false`) and sets the key for the exact three conditions, plus the inverse: with nothing in the environment the banner still shows.

Verified by reverting: without the `environment_id` line both no-banner tests fail and nothing else in either file does. Stages 5 and 6 are rebased on the new tip.

jhgaylor: Rebased onto main after the roots merged; the fix is now 92c8c63d (same diff).

BinaryBourbon (approved):
Fresh adversarial review of `1f4353fdf47f591c4ab2cf8cda803484a54ba37b` against `aef22f8afd8d00aed0a2fc3053363b27c5e03852`: **approved**. No remaining blocking defect found.

The previous P2 is fixed in the implementation: `StartLive` now supplies `agent.environment_id` together with the agent's named set. LiveView regressions exercise an empty explicit set plus a usable environment credential, both with and without a platform key; the inverse still warns. I traced the generated launch request to admission and confirmed the same default environment/set and nil vault context. An independent second pass challenged that context and found no blocking issue.

Reviewed all 14 changed paths, including resolver option validation, runtime precedence, alias conflicts, tenant-scoped overrides, explicit-set refusal, source identity/revision binding, platform key/grant policy without provider I/O, removed-function callers, and converted persisted-row tests.

Validation: **106 targeted tests passed on this exact head**, across all changed test files plus the existing platform-banner suite, with pinned Elixir and a dedicated freshly migrated local PostgreSQL database. `git diff --check` passed. The initial run against the existing shared database had four audit-list failures caused by a pre-existing audit row (id 1007, timestamp 2026-09-14 10:49:50Z); unchanged tests passed on the isolated database. No shared database rows were deleted. Another 29 targeted tests for the preceding stack changes passed before the final rebase.

The head was rebased during review. The old/new PR deltas have identical stable patch IDs; I inspected the new base's ChatGPT-cleanup interactions and reran the 106 tests on the final head. Trusted review policy/instructions are unchanged. GitHub CI was still finishing at the last check; this approval does not waive its required gates. Full local precommit and separate Review Loop service verification were not run. This is a commit-specific code-review verdict.

<details>
<summary>Reviewed changed paths</summary>

- `apps/fountain/lib/fountain/conversations/codex_transport.ex`
- `apps/fountain/lib/fountain/inference_credentials/resolver.ex`
- `apps/fountain/lib/fountain_web/live/start_live.ex`
- `apps/fountain/test/fountain/conversations/conversation_server_platform_inference_test.exs`
- `apps/fountain/test/fountain/inference_source_binding_test.exs`
- `apps/fountain/test/fountain/platform_chatgpt_test.exs`
- `apps/fountain/test/fountain/platform_inference_test.exs`
- `apps/fountain/test/fountain_web/live/start_live_platform_banner_test.exs`
- `apps/fountain/test/fountain_web/live/start_live_test.exs`
- `changelog.d/2185-banner-agrees-with-launch.md`
- `ee/test/fountain/credits_chatgpt_inference_test.exs`

</details>

Issue #2176, opened 2026-09-14
Design memo: simplify how a run chooses its credentials


jhgaylor:
The 2026-09-14 complexity review found the two credential questions a run answers before a sandbox can speak to a model (which inference credential goes in, and who is acting) spread across roughly 9,000 lines and seven ADRs, each a reasonable answer to a real problem, with nobody having decided there would be one credential model. This is the design memo for simplifying it. It was written read-only against `main` at `74ad641f`; nothing is implemented. The full memo is at the bottom, under the fold.

## Headline findings

- **The resolver already applies one order.** The work is reducing what it arbitrates between, not designing a new resolver.
- **Four of the thirteen "resolver options" are not resolver options** (`authorize`, `actor`, `request_ip`, `metadata` are write-path options). Of the rest: `refresh: false` at the three `conversations.ex` sites is a no-op the resolver overwrites (`resolver.ex:30`); `secret_keys` has no caller anywhere; `allow_missing` is forced true and passed by nobody; `brokered` is a test seam; `override_entries` is resolver-internal; `overrides` is reachable only through `SpriteEnv.select_inference/4`, which has no production caller.
- **Two callerless wrappers**: `PlatformInference.gate/4` and `SpriteEnv.select_inference/4` (plus `load_tenant_state/2`) exist for their own tests only.
- **The two ChatGPT families are one mechanism with two owners** (same table, cipher, lock and OAuth module), and the user-owned half is dormant: nothing in production inserts or reads an owned row, so about 700 lines are reachable only from their own tests.
- **Removal-issue status**: #2112 merged; #2106 was closed by #2114 (the code now requires an explicit `auth_mode`); #2107 is still open, `vault_access` landed in #2119, and the nil-means-any shape survives for the environment and credential-set allowlists.
- **ADRs describing what is not there**: 0047 cites `PlatformInference.gate/3` and `serves?/2` (neither exists) and describes compare-and-swap refresh where the code has an advisory try-lock; 0052's status says nothing is built while 0053's says decisions 1, 3 and 6 are; 0053 is still Proposed with its whole stack merged; 0008 names `Fountain.Runtimes.*`; 0054 amends 0033's closed-registry rule without a note.

## Target model

[...]

PR #2197, opened 2026-09-14
refactor(conversations): one helper builds the resolver call

### LINKED ISSUE #2187: refactor(conversations): one helper builds the resolver call for admission, wake and provision

jhgaylor:
Stage 6 of #2176. Base: `stack/2176-5-origin` (#2186); depends on #2185.

## What changed

- **New module `Fountain.Conversations.InferenceResolution`** (98 lines with its docs), not in `conversations.ex` and not in `SpriteEnv`:
  - `select/3`, a new selection for admission: `credential_set_id` (the launch's override, nil for the agent's set), `environment_id` (nil for the agent's) and `vault_id`; `Keyword.validate!/2` on those three.
  - `revalidate/3`, a re-validation for wake, resume, provision and reapply: the conversation's `inference_source` is the expected source and its `runtime` the runtime, the agent's model is the model, `environment_id` / `vault_id` are the rows it runs against now. Reapply overrides `expected_source` (the stored source with the new configuration merged in) and `runtime` (the agent's, which the row takes only after the source resolves).
  - `credential_set_id/2`, moved from `SpriteEnv` unchanged; `launch_credential_set_test.exs` calls it there.
- **Call sites:** `resolve_admission_inference/5` calls `select/3`; `resolve_saved_inference/2` (wake and resume) and `resolve_reapplied_inference/4` call `revalidate/3`; `SpriteEnv.resolve_inference/4` calls `revalidate/3` when the conversation has a stored source and `resolve/4` with `credential_set_id/2` when it has none (a conversation admitted before sources were stored). `conversations.ex` no longer aliases `SpriteEnv`.
- **`InferenceCredentials.validate_source/2`** drops its redundant `credential_set_id` in place: it is the context's own function and calling `Conversations.InferenceResolution` from there would invert the dependency.

`conversations.ex`: 6,059 → 6,057 lines, 249,375 → 249,116 bytes. No size pin exists on it (only `conversation_server.ex` has one).

## Where the issue and the code differ

The issue describes a re-validation as "`expected_source` and the environment/vault it names, and nothing else". Wake passes the agent's *current* environment (`conv.environment_id || agent.environment_id`) and provision the rows it just loaded, and a configuration that moved since admission is exactly what the `expected_source` comparison is there to refuse (`legacy_wake_reapply_test.exs`, `inference_reapply_test.exs`). So the helper takes the current rows from the caller and only the set is taken from the stored source, which is the redundancy the issue identified.

## Verification

`mise exec -- mix precommit` at the umbrella root on `fountain_test_2176a`:

```
6 doctests, 6016 tests, 5 failures, 2 skipped
```

The five failures are the environmental trust-store ones of the earlier stages; extension suites buzz 136, slack 30, support 33, google 60, microsoft 29 tests, 0 failures.

Reverted to verify: with `revalidate/3` passing no `expected_source`, three tests fail across the reapply, wake and admission-isolation suites; restored, all pass. Targeted files before the gate: 224 tests, 0 failures (`conversations_test.exs` included).

No changelog fragment: internal.

Part of #2176
Closes #2187


apps/fountain/lib/fountain/conversations/inference_resolution.ex
```diff
@@ -0,0 +1,89 @@
...
+  """
+  @spec revalidate(map(), map() | nil, keyword()) :: result()
+  def revalidate(conv, agent, opts) do
+    opts = Keyword.validate!(opts, [:expected_source, :runtime, :environment_id, :vault_id])
+
+    InferenceCredentials.resolve(
+      conv.user_id,
+      agent && agent.model,
+      Keyword.get(opts, :runtime, conv.runtime),
+      expected_source: Keyword.get(opts, :expected_source, conv.inference_source),
```
BinaryBourbon, line 92: [P1] Only omit credential_set_id when an expected source actually exists. resolve_saved_inference also calls this for legacy conversations with inference_source: nil. Such a conversation/agent may name a nondefault set; this now resolves the account default. resume_channel then calls InferenceBinding.reserve, whose legacy branch persists that wrong default binding. Retain credential_set_id(conv, agent) for the nil-source case and cover a legacy resume with distinct named/default credentials.

BinaryBourbon (changes requested):
Adversarial whole-PR review of `04aa50d2657dbe1f496dcb11ba0f3a5fa89add4a` against `8436d48dea5b8a2ac91679c271a993d80e5079e9`.

**Changes requested: retain explicit-set selection when a legacy conversation has no stored source (P1).** `InferenceResolution.revalidate/3` unconditionally drops `credential_set_id`, but `resolve_saved_inference/2` also handles historical conversations with `inference_source: nil`. Previously that path passed the conversation/agent set via `SpriteEnv.credential_set_id/2`; now `expected_source: nil` makes the resolver choose the account default. For a legacy Claude channel conversation whose agent names a nondefault set, `resume_channel/4` calls this helper at `conversations.ex:3261` and immediately reserves the resulting source. `InferenceBinding.current_source/2` permits an initial legacy binding, so resume can permanently pin the default credential instead of the configured named set. Wake's gate can likewise evaluate a different credential from the one provision uses. The source migration leaves historical bindings nil, and existing legacy tests explicitly support that state.

Preserve the conversation/agent credential-set fallback whenever there is no expected source, as the new SpriteEnv legacy branch already does. Add a legacy channel-resume regression with distinct default/named credentials and assert the persisted binding uses the configured set; cover a conversation override as well.

Evidence: an isolated Elixir diagnostic executed the production resolver and new helper with in-memory database/provider boundaries. For the identical nil-source conversation and agent naming `named`, the old call selected set `named`; `revalidate/3` selected set `default`. Inspected the real resume/reserve path to establish that the wrong source is persisted, not only returned. This diagnostic does not claim a full database integration run.

Reviewed admission, wake, channel resume, reapply, provision, source validation, helper tests and legacy lifecycle tests. GitHub checks are complete with success/skipped conclusions at review time. The full local Mix diagnostic could not run because the isolated worktree has no installed dependencies. No Review Loop App summary or prior review was present. This is a code-review verdict, not completion of the separate Review Loop verification policy.

<details>
<summary>Reviewed changed paths</summary>

- `apps/fountain/lib/fountain/conversations/inference_resolution.ex`
- `apps/fountain/lib/fountain/conversations/sprite_env.ex`
- `apps/fountain/test/fountain/conversations/launch_credential_set_test.exs`

</details>

apps/fountain/lib/fountain/conversations/inference_resolution.ex
jhgaylor: Fixed in d93bbdd2. `revalidate/3` now passes `credential_set_id: credential_set_id(conv, agent)` on every call. The resolver ignores it whenever an `expected_source` is present (`resolver.ex:22`), so the stored-source path is unchanged, and a legacy row with no stored source selects the set the agent or the launch named. The SpriteEnv provision no longer needs a legacy branch of its own and is back to one `revalidate/3` call; the moduledoc says why the set stays.

Three regressions in `launch_credential_set_test.exs`: the helper with a nil source picks the agent's set and then a conversation override; a legacy channel conversation (source nulled after admission) resumes on the agent's set; and on its own `inference_credential_id` override. Both resume tests assert on the persisted binding after `resume_channel` reserves it, with distinct default and named credentials from the module's setup.

Verified by reverting: without the helper line all three fail and nothing else in the file does.

jhgaylor: Rebased onto main after the roots merged; the fix is now f3fe1ee2 (same diff).

BinaryBourbon (approved):
Fresh adversarial whole-PR review of `f3fe1ee2427d3ffe969955818f94ebc7fc2487e3` against `76de6f4e783181a4d392c1aa0e2c8362943617cc`.

**Approved. The prior P1 legacy credential-selection finding is addressed.** `revalidate/3` now carries `credential_set_id(conv, agent)`. A nil-source conversation retains its configured conversation/agent set; when an expected source exists, the resolver continues to take its set from that source. Traced channel resume through the actual reservation/persistence path and checked both new persisted-binding regressions (agent set and conversation override), plus the direct helper regression. The added fallback does not relax stored-source revision matching.

Reviewed all five changed paths and relevant admission, wake, channel resume, reapply, provision, source validation, binding/locking and credential-scope callers. Reapply still supplies its proposed runtime/context while retaining the pinned source; provision remains under `with_current/2` and reserves before provider work. No blocking correctness, tenant-isolation or compatibility issue found in the current delta.

Evidence: an isolated diagnostic executed this head's production `Source`, `Resolver` and `InferenceResolution` with in-memory database/key/provider boundaries: **7 tests, 0 failures**. It covers legacy agent/override/default selection, stored-source precedence and replacement refusal, reapply runtime override, and admission precedence. Inspected the repository's database regression assertions; did not execute the database integration suite locally. `git diff --check` passed. GitHub's current checks, including all six test partitions, static analysis, coverage, `CI required` and secret scanning, are successful (conditional jobs skipped). Head/base/state and existing reviews were rechecked before posting.

This is a fresh code-review verdict, not completion of the separate Review Loop service verification recipe. No Review Loop App summary is present.

<details>
<summary>Reviewed changed paths</summary>

- `apps/fountain/lib/fountain/conversations/inference_resolution.ex`
- `apps/fountain/lib/fountain/conversations/sprite_env.ex`
- `apps/fountain/test/fountain/conversations/launch_credential_set_test.exs`

</details>

PR #2207, opened 2026-09-15
chore(precommit): the default test stage runs the changed files' tests; --full runs the whole suite

jhgaylor:
## What changes

`mix precommit`'s `test` stage runs `scripts/precommit-tests.py`, the tests for the files the branch changed, instead of the whole suite. Every other stage is untouched. `mix precommit --full` runs the whole suite in that stage, `mix precommit --list` says which command the stage runs in each mode, and the exit-status contract is unchanged: the last line is still `precommit: PASSED` or `precommit: FAILED at <stage>`.

The selector works from the working tree against `git merge-base origin/main HEAD`, plus untracked files:

- a changed `*_test.exs` runs itself;
- a changed lib file runs its mirror test and every test in that app with the same stem (`search.ex` runs `search_test.exs` and `search_controller_test.exs`; `ee/lib` looks in `ee/test` and `apps/fountain/test`);
- a docs change runs the manual's tests, the same files `scripts/test-docs.sh` names (a test pins the two lists together);
- a change to something every test reads escalates to the whole suite on its own: `mix.exs`, `mix.lock`, `config/`, `coverage.exs`, a `test/support` tree, a `test_helper.exs`, a migration;
- a lib file with no matching test is named in the output, never skipped in silence; every selected path is checked to exist before `mix test` sees it, because mix drops an unmatched path without failing when another matches;
- one `mix test` per app directory, with `ee/test` reached from `apps/fountain` as `../../ee/test`.

`python3 scripts/precommit-tests.py --print` shows the selection without running it.

## Why

Main takes 56 to 72 merges a day and the local gate ran all ~6,000 tests before each push. CI runs the whole suite on every plan, so a dependent the selection misses costs one CI round trip, not a merged defect. The static stages, sobelow and the release assemble stay in the default run because they are the checks CI cannot catch cheaply or at all locally.

## Why not `mix test --stale`

It was the first design and it was measured. Two findings:

1. **It crashes on this repo.** `apps/fountain`'s `test_paths` include `"../../ee/test"`; Mix keys the stale manifest by that relative path and looks each compiled file up by `Path.relative_to(file, cwd)`, which cannot express a path above cwd and falls back to the absolute one. The first ee file raises `KeyError` in `Mix.Compilers.Test.each_file/4`. An absolute `test_paths` entry fixes it, and was tried.
2. **With that fixed, it selects almost everything.** The stale closure follows runtime references (`dependent_modules/3` checks `module in r or module in c or module in e`), and in a Phoenix app every context module reaches the router and endpoint. A probe edit to `Fountain.Search` and `Fountain.Workers.WelcomeEmail` selected 5,516 of 5,963 core tests plus 27 in `fountain_buzz`.

The mix.exs change is not in this PR. It is a one-line fix for anyone who wants `--stale` for local iteration, but it changes the ee test paths that appear in failure output and it does not serve the goal here.

## Measured in the umbrella, from the root

| Run | Command | Result | Wall time |
|---|---|---|---|
| Whole suite, baseline (`--stale` with the absolute-path fix, first run) | `mix test --stale` | 6 doctests + 5,963 core tests, 136 buzz, 30 slack, 33 support, 60 google, 29 microsoft, 0 failures | 126 s |
| Second `--stale` run, no edits | `mix test --stale` | "No stale tests" in all six apps | 2 s |
| Third `--stale` run after a probe edit to `Fountain.Search` and `Fountain.Workers.WelcomeEmail` | `mix test --stale --trace` | 5,516 core + 27 buzz tests ran | 162 s (trace mode) |
| **The selector**, same probe edit | `python3 scripts/precommit-tests.py` | `search_test.exs`, `search_controller_test.exs`, `../../ee/test/fountain/workers/welcome_email_test.exs`: 18 tests, 0 failures | 5 s including the recompile |
| The selector on this branch's own diff | `python3 scripts/precommit-tests.py` | 0 test files selected (scripts and Markdown only), stage passes and says so | 1 s |

## Verification

All in a private worktree at the branch tip, against a private test database (`fountain_test_precommit_stale`), through `mise exec`.

- `python3 -m unittest discover -s scripts/ci -p 'test_*.py'`: 121 tests, OK (10 new: 9 selection rules in `test_precommit_tests.py`, 1 stage test in `test_precommit.py`).
- `shellcheck scripts/precommit.sh`: clean. `python3 scripts/conflict-markers.py`, `python3 scripts/changelog.py check`, `git diff --check`: clean.
- `mise exec -- mix precommit` (default mode) on this branch: `precommit: PASSED (10 stages, 229s)`. The test stage reported `0 test files selected for the changed paths` and took 0s; dialyzer 105s and the release assemble 106s were the cost.
- `mise exec -- mix precommit --full test`: 6 doctests + 5,963 core tests with 1 failure, 136/30/33/60/29 in the sibling apps with 0 failures, `precommit: FAILED at test (exit 2) after 132s`. The failure is `Fountain.AuditTest` "since and until bound the window inclusively", the shape #2178 describes: this private database had run the suite four times and held 387 committed audit rows, so the test's 200-row page lost its own row. The same suite passed 0 failures on the first run against the fresh database earlier in the session (the 126s baseline above), and this PR changes no Elixir code.
- `mix precommit --full`, `mix precommit --full credo test`, `mix precommit test --full` argument shapes: covered by the stage test with the stub `mix`; the alias in `mix.exs` already passes arguments through unchanged.

Not verified: a cold checkout's first run (no `_build`), and the selector against a branch with a renamed lib file (a rename shows as the new path only, so the old test file is not selected unless it changed).


BinaryBourbon (changes requested):
Independent adversarial integration review of fbfaa48de06dc805fcb60d61387742b7d0567751 against 4224258d6e773201501e50959469f82defc8b572, covering all seven changed files and their Mix callers.

Changes requested: include `ee/test/test_helper.exs` in the shared-input full-suite fallback (`scripts/precommit-tests.py:45-56`). I independently reproduced that changing this existing helper yields `full_reason=None`, no selected tests, and an ignored-path message. `apps/fountain/mix.exs` declares `../../ee/test` in `test_paths`, so the helper is loaded by the core suite; deleting it breaks that suite while the new default test stage selects nothing. This is the same selector finding raised by the other reviewer; no duplicate inline comment.

Validation: all 18 targeted precommit/selector Python tests passed in an isolated archive of the reviewed head. Additional isolated executable-stub checks passed for invocation outside the checkout, core/EE and extension working directories, `MIX_ENV=test`, preservation of a second-app exit status of 23 with later apps skipped, `--full` argument order/repetition and stage deduplication, config-triggered root-suite fallback, and side-effect-free `--print`. No additional integration blocker found. This is a manual source/integration review; I did not run the full Elixir suite or the hosted Review Loop verification recipe.

scripts/precommit-tests.py
```diff
@@ -0,0 +1,260 @@
...
+
+# Any of these changed means the whole suite: they are read by every test.
+FULL_SUITE_PATTERNS = [
+    "mix.exs",
+    "mix.lock",
+    "coverage.exs",
+    "config/*",
+    "apps/*/mix.exs",
+    "apps/*/test/support/*",
+    "apps/*/test/test_helper.exs",
```
BinaryBourbon, line 53: [P2] Include the EE test helper in the full-suite fallback

`ee/test/test_helper.exs` exists and is loaded through `apps/fountain/mix.exs`'s `test_paths`, but this pattern only covers helpers under `apps/`. Changing that EE helper currently puts it in `ignored` and returns success with zero tests, contrary to the documented fallback for a changed `test_helper.exs`. In an isolated reproduction, even a helper containing `raise "invalid helper"` returned 0 without invoking Mix. Include `ee/test/test_helper.exs` in the shared-input patterns and add a regression that verifies it requests the full suite.

scripts/precommit-tests.py
```diff
@@ -0,0 +1,260 @@
...
+        result = subprocess.run(
+            ["git", "merge-base", ref, "HEAD"], cwd=ROOT, capture_output=True, text=True
+        )
+        if result.returncode == 0:
+            base = result.stdout.strip()
+            break
+    if base is None:
+        return None, "no merge base with origin/main or main"
+    diff = subprocess.run(
+        ["git", "diff", "--name-only", base], cwd=ROOT, capture_output=True, text=True, check=True
```
BinaryBourbon, line 185: [P2] Preserve rename source paths for shared-input escalation

Git rename detection makes `--name-only` emit only the destination. An isolated `git mv apps/fountain/test/support/factory.ex apps/fountain/lib/fountain/factory.ex` produced an R100 rename, but `changed_files()` returned only the lib path; selection consequently had no `full_reason` and zero tests. The old `test/support` path changed and should trigger the promised full-suite fallback. Disable rename detection for this path inventory (`--no-renames`) or collect both sides of renames, and cover a shared-input move with an actual-Git regression.

BinaryBourbon (changes requested):
Independent adversarial review of fbfaa48de06dc805fcb60d61387742b7d0567751 against 4224258d6e773201501e50959469f82defc8b572. Reviewed all seven changed files, the approved-base review guidance, and the Mix/test callers. Requesting changes for two reproducible gaps in the documented shared-input fallback; the deliberate limits of source/stem selection are otherwise respected.

Validation: all 18 precommit Python tests passed in an isolated exact-head archive. Additional isolated Git/stub-Mix diagnostics confirmed the two findings, ordinary source-deletion selection, first-app failure propagation and stop behavior, and full-suite fallback for known shared inputs and a missing merge base. I did not run the full Elixir suite or the hosted Review Loop verification recipe.

jhgaylor:
Closing in favour of a smaller change: the test stage moves behind `--full` with no selector. CI runs the whole suite on every PR and again in the merge queue, so the local suite was buying little for 400 new lines. The finding that `mix test --stale` crashes on the ee test path stands on its own.

Issue #2227, opened 2026-09-15
Tracker: reduce the conversation field tax across API, SDKs and CLI


jhgaylor:
## Outcome

Make an ordinary additive conversation field a server implementation and API-contract change followed by deterministic regeneration, rather than a coordinated rewrite of every SDK and CLI. Keep client behavior (stream following, retries, permission handling, pagination and run completion) independently implemented and tested.

## Current cost

The main duplication is request construction: `sdk/typescript/src/client.ts`, `sdk/python/src/fountain/client.py`, `sdk/elixir/lib/fountain.ex` and the Swift clients enumerate launch options and assemble the same JSON. `cli/internal/cmd/conv.go` repeats a subset. Typed Swift models also repeat wire fields and CodingKeys.

The often-quoted nine-place tax is not universal: TypeScript response types already derive from OpenAPI (`schemas.ts`), Python/Elixir return maps, and Go conversation output uses maps. An optional response field should not require edits to all clients or every conformance fixture.

## Approach

- Pass contract-shaped conversation inputs separately from client execution options.
- Generate wire models where typed clients need them; reuse the existing OpenAPI/contract pipeline.
- Preserve current public helpers as compatibility adapters, with explicit collision/omission semantics.
- Give CLI users a JSON file/stdin path so new fields do not require dedicated flags.
- Keep server serialization explicit and conformance expectations independently authored.

## Deliverables

- [x] #2228 — SDKs: define conversation request passthrough and ship the TypeScript pilot (merged)
- [x] #2229 — Python SDK: forward conversation request fields without expanding run keywords (merged)
- [x] #2230 — Elixir SDK: forward conversation request maps separately from run options (merged)
- [x] #2231 — Swift SDK: generate conversation wire models and adopt request passthrough (merged)
- [x] #2232 — CLI: accept conversation creation JSON from a file or stdin (merged)
- [x] #2233 — SDK contract: make field propagation automatic and keep conformance focused on behavior (merged)

## Completion criteria

- A representative additive optional request field reaches supported SDK run paths and CLI JSON creation without adding another handwritten field translation to each client.
- An additive response field appears in generated typed models or existing map responses without handwritten model edits.
- Request forwarding preserves omission, explicit null, false, zero and empty values according to the API contract; local timeout/collection settings never become request fields.
- Existing run/name-resolution APIs remain compatible; behavioral differences require focused scenarios.
[...]

Issue #1406, opened 2026-09-02
Conversations: return a durable turn identity when a prompt is accepted


jhgaylor:
## Completed correlation scope — September 18, 2026

The accepted September 17 first cut is complete: caller-supplied `client_request_id` is carried through prompt submission and conversation creation to the resulting turn and its start event, with SDK, CLI, and ACP support. It is a correlation value; repeating it is not an idempotent retry.

## Completion evidence

- [x] #2374 — prompts-route validation/response, turn persistence and rendering, start-event correlation, both delivery paths, controller/concurrency coverage, and manual/schema updates.
- [x] #2376 — the same field on conversation create; the genuine-field propagation exercise recorded in #2248.
- [x] #2390 — all four SDKs, both Swift surfaces, follow-up prompts, and all five channel-resume forwarding branches, with per-language regression and omission coverage.
- [x] #2403 — CLI `--client-request-id` and ACP `_meta.clientRequestId`; #2389 is closed.
- [x] #2405 — shared channel-resume conformance across all five adapters, with negative probes proving missing forwarding is detected; #2388 is closed.

All five PRs are merged. The original bug was the missing durable caller correlation, not a promise that a turn already exists when a wake request is acknowledged. Requests omitting the field retain the established behavior.

## Separate accepted follow-up

#2375 owns idempotent retries and authoritative identity during wake: #2407 reserves durable prompt identity, and #2408 resolves repeated client keys without new execution. Those issues remain open with their design and rollout gates. Closing this issue does not claim they are implemented.

The accepted design and its rationale remain in the [September 17 comment](https://github.com/managoat/fountain/issues/1406#issuecomment-5717338773). Parent paydown tracker: #2247. The original broader proposal is retained in this issue's edit history.

jhgaylor:
## Design for the first cut (2026-09-17)

Picked up from #2247. Read against `main` at `02883de91`.

### Why option 1 (return `turn_id`) is not the contract

`POST .../prompts` has two paths, and only one of them has a turn when it answers:

- **A live server.** `ConversationServer.handle_call({:send_prompt, ...})` opens the turn through `TurnMachine.open/6` before it replies, so the id exists. But the same call also answers `:ok` when `kick_turn/4` was refused `:at_capacity` or `:no_command`, and when it deferred the prompt behind `:configuration_changed`. "Accepted" does not imply "a turn exists".
[...]

PR #2234, opened 2026-09-15
feat(sdk): accept API-shaped conversation requests in TypeScript

### LINKED ISSUE #2228: SDKs: define conversation request passthrough and ship the TypeScript pilot
Parent initiative: #2227

## Problem

`RunConfig` and the body builder in `sdk/typescript/src/client.ts` manually mirror launch fields even though `ConversationCreateRequest` already exists in the API contract. A new field requires edits to both the schema and the ergonomic client.

## Scope

Define the shared request-vs-execution-options boundary and implement it in TypeScript as the reference for the other SDKs. Write an ADR recording the selected public shape, compatibility approach and field semantics; describe unimplemented language work as proposed.

## Acceptance criteria

- Expose a conversation request type derived from the generated `ConversationCreateRequest` and forward the supplied wire object without a field-by-field copy list.
- Separate API inputs from local timeout, collection and transport/run settings.
- Preserve the existing `run(prompt, config)` surface through an adapter or compatible overload; retain name resolution as a small convenience. Document how a raw ID and a name-based option interact; reject ambiguous collisions or specify one tested precedence.
- Audit fields that change the response/lifecycle (for example queued creation) before allowing them through a run handle that expects an immediate conversation. Unsupported semantics must fail clearly, not silently disappear. This issue need not implement queued runs.
- Preserve absent vs explicit null, false, zero and empty values according to the contract; do not use truthiness to filter the raw request.
- A recording transport proves an API option not on the old RunConfig list reaches the request, local options stay local, and old helpers still form the expected request.
- Existing TypeScript run, contract and conformance checks pass; document examples and public-surface release implications.

## Starting points

`sdk/typescript/src/client.ts`, `schemas.ts`, `run.ts`; `sdk/contract`; `apps/fountain/lib/fountain_web/schemas.ex`.

jhgaylor:
## Summary

Conversation launch fields can now be sent through `runRequest(request, options)` without adding another ergonomic option and JSON mapping. `ConversationInput` comes from the existing generated OpenAPI types; local execution settings are separate, and existing name-based `run` calls remain compatible.

The new path rejects queued and promptless creation before HTTP because a Run follows an immediately started turn. ADR 0056 records this boundary for the remaining clients. This prepares TypeScript SDK 5.2.0; merging the version change follows the existing package release workflow.

Closes #2228. Part of #2227.

## Validation

- Typecheck, build and SDK contract verification passed.
- All 120 TypeScript tests passed, including shared conformance and new request forwarding/lifecycle tests.
- Packed-package consumer checks passed for declarations, Node/browser entry points and browser bundling.
- ADR validation passed (only existing metadata warnings); conflict-marker, changelog and diff checks passed.

Channel following retains the existing history-based turn selection; durable prompt-acceptance identity remains tracked separately in #1406.


sdk/typescript/src/client.ts
```diff
@@ -164,6 +165,30 @@ export class Fountain {
...
+    if (request.queue != null && request.queue !== false) {
+      throw new TypeError("runRequest does not support queued creation; use api.request");
+    }
+    const body = { ...request };
+    return new Run(this.api, {
+      start: async () => {
+        const conversation = await this.api.data<ConversationRecord>(
+          "POST", "/api/conversations", { body },
+        );
+        const turnNumber = body.channel_id ? await this.nextTurnNumber(conversation.id) : 1;
```
BinaryBourbon: [P2] Dispatch the prompt when channel creation resumes an existing conversation

With `runRequest({agent_id, channel_id: "bound-channel", prompt: "execute this"})` and an existing idle channel conversation, the server returns that conversation with `meta.resumed: true`. `Conversations.start_or_resume_conversation/2` takes `resume_channel/4` (`apps/fountain/lib/fountain/conversations.ex:2894-2934`), which validates the binding and merges labels but never sends the supplied prompt. This new client path discards the response metadata, reads the history and starts following `max + 1` without ever posting `/api/conversations/:id/prompts`. Consequently the requested work never starts and the default Run waits indefinitely (or times out when configured).

I reproduced this with the actual new SDK method and a recording transport matching that server response: history contained completed turn 1, requests were only `POST /api/conversations`, `GET .../turns`, `GET .../stream`, and the Run raised `Timed out after 40ms waiting for turn 2`. The new channel test hides this by using `fake.onTurn` to fabricate turn 2 on creation, which the real resume branch does not do.

Preserve the created/resumed distinction and explicitly submit/follow the prompt on resume, with a regression transport that models the actual resume response and asserts prompt delivery; alternatively reject unsupported channel runs before HTTP. The legacy helper shares this limitation, but the new entry point currently accepts this lifecycle as supported, contrary to #2228's requirement that unsupported semantics fail clearly. This is deterministic missing delivery, separate from #1406's concurrent prompt-identity problem.

BinaryBourbon (changes requested):
Adversarial review: **CHANGES REQUESTED** for `50cb67ddb3abf88657cbdea13bb4c19eab5d946a`.

One blocking P2 finding is attached to the new channel-run path: an existing channel resumes without delivering the requested prompt, then waits for a turn that was never started.

Reviewed all 11 files in the actual PR diff, including generated-input typing/public exports, request serialization and run options, server create/resume/queue behavior, SDK release metadata, documentation and ADR 0056.

Validation: local `npm run typecheck`, `npm run build`, `npm run verify-contract`, all **120 TypeScript tests** (including conformance), and `git diff --check` passed in an isolated snapshot. A separate recording-transport diagnostic reproduced the channel failure described inline. The local checks used Node 26.7.0; current-head CI also passed its supported Node 20/24 SDK jobs, server tests, release gate, ADR validation and secret scan ([CI run 34932821076](https://github.com/managoat/fountain/actions/runs/34932821076)). I did not rerun the server's database suite locally. Review Loop dispatch was skipped and no App review/summary exists on this PR.

BinaryBourbon (approved):
Independent adversarial review: **APPROVE** for `60fe9e4431e4321e2c7cb55edc8748eea7510ce8` against base `ab6bd0bd3195126c5a531f62e79817efc02acd0e`.

Reviewed the entire 12-file PR diff, including the original API-shaped input/type exports, request serialization, local Run options, server create/resume/queue and prompt lifecycle, backwards compatibility, test fixture validity, SDK 5.2.0 release metadata and packaging, README, changelogs, and ADR 0056. No remaining concrete blocking defects found.

The prior P2 missing-prompt finding is addressed: the client preserves `meta.resumed`, reads cursor/history before dispatch, posts the prompt and images exactly once on resume, and follows turn 1 for newly created/fresh channel conversations. The regression now models a resume that does not itself start a turn.

Actual local validation in a new isolated worktree (Node 26.7.0): typecheck, build, contract verification, all **122 TypeScript tests** including conformance, packed npm declarations/Node/browser consumer checks and browser bundling, conflict-marker/changelog/diff checks, and ADR validation (zero errors; six pre-existing metadata warnings) passed. Eight additional independent recording-transport diagnostics passed: plain creation, new channel, fresh channel, resumed empty and multiple-turn histories, null image forwarding, cursor-discovery failure fallback, and prompt-busy error propagation.

Current-head GitHub CI is green, including supported Node 20/24 SDK jobs, server partitions, release/contract checks, SDK release gate, decisions and secret scan ([CI run 34939109663](https://github.com/managoat/fountain/actions/runs/34939109663)). This is an explicitly requested independent human-account review, not a Review Loop service run. Review Loop dispatch was skipped and there is no App review/summary; I did not execute its separate full verification recipe or rerun the server database suite locally. The existing history-based concurrent prompt identity limitation remains tracked in #1406.

PR #2235, opened 2026-09-15
feat(python): forward API-shaped conversation requests

### LINKED ISSUE #2229: Python SDK: forward conversation request fields without expanding run keywords
Parent initiative: #2227

## Problem

`sdk/python/src/fountain/client.py` lists launch fields in the run signature and again in an optional-field dictionary. New server options are unavailable through the run helper until both lists change.

## Scope and dependencies

Follow the request/execution boundary and compatibility semantics established by the TypeScript pilot. Keep dictionary-based responses; do not add a runtime model framework merely to pass through fields.

## Acceptance criteria

- Add a mapping-based conversation request path that forwards wire fields without a per-field allowlist in the SDK.
- Keep existing keyword-based calls working through a compatibility adapter and retain name resolution.
- Preserve explicit null, false, zero and empty values; omission remains different from explicit null. Define conflicts between legacy keywords and the wire request consistently with the pilot.
- Keep timeout, event collection and other local execution options outside the JSON body; handle unsupported lifecycle-changing inputs as specified by the pilot.
- Recording-transport tests exercise an option absent from the old keyword list, omission/null behavior and legacy calls. Existing run/conformance and contract checks pass.
- Document the new path and when callers should use names vs raw IDs; follow the package's release rules for the public change.

## Starting points

`sdk/python/src/fountain/client.py`, `run.py`, `http.py`, `tests/test_conformance.py`.

Shared request-boundary reference: #2228.

jhgaylor:
## Summary

`Fountain.run_request(request, *, timeout=None, collect_events=False)` sends API-shaped launch fields directly, including fields absent from the existing keyword helper. Local settings remain separate, values preserve omission/null/false/empty distinctions, and the existing name-based `run` API stays compatible.

The method refuses queued and promptless starts before HTTP. It follows the request boundary proposed in #2234; it can be reviewed independently. Prepares Python SDK 0.5.0.

Closes #2229. Part of #2227.

## Validation

- All 53 Python SDK tests passed, including shared conformance and new request forwarding tests.
- Contract verification and source compilation passed.
- Source distribution and wheel built successfully; all 53 tests also passed against the installed wheel in an isolated environment.
- Conflict-marker, changelog and diff checks passed.


sdk/python/src/fountain/client.py
```diff
@@ -83,6 +83,35 @@ def plan() -> Any:
...
+        if not isinstance(prompt, str) or not prompt.strip():
+            raise ValueError("run_request requires a non-empty prompt; use request for promptless creation")
+        if body.get("queue") is not None and body["queue"] is not False:
+            raise ValueError("run_request does not support queued creation; use request")
+
+        def plan() -> Any:
+            conversation = self.api.data("POST", "/api/conversations", body=body)
+            turn_number = (
+                self._next_turn_number(str(conversation["id"]))
+                if body.get("channel_id") else 1
```
BinaryBourbon: [P2] Follow the turn that a fresh channel launch actually started

A supported request with `channel_id` and `fresh=True` creates a new conversation and submits its prompt as turn 1. If that turn exists by the following GET /turns, `_next_turn_number` returns 2, so this Run ignores the completed first turn and waits for a turn it never submitted (indefinitely with the default timeout).

This reproduces with the existing unmodified `FakeFountain`: call `client.run_request({"agent_id": AGENT_ID, "prompt": "hi", "channel_id": "raw", "fresh": True}, timeout=0.2).result()`. The fixture emits a completed turn 1, but the result raises `TimeoutError: ... waiting for turn 2`. The new channel test masks the defect by patching `_next_turn_number` to return 1. The real server's `start_or_resume_conversation` fresh branch likewise invokes `start_conversation`.

Use turn 1 for a known new/fresh conversation and distinguish creation from resume when selecting/submitting the followed turn; alternatively refuse unsupported channel launches before HTTP. Add a regression exercising actual turn history rather than mocking the selection helper. Legacy `run` shares this problem, but copying it into the new public helper leaves its accepted launch shape broken even with a single caller; no concurrent-attribution redesign is needed to reproduce this.

BinaryBourbon (changes requested):
Adversarial review of 7ac701bc712a59b77c92a61cda04b18bca55b6bb against ab6bd0bd3195126c5a531f62e79817efc02acd0e: changes requested for one reproducible defect in the new run_request channel path.

Reviewed all eight changed files, request serialization and value preservation, Run/TurnFollower behavior, the server create/channel-resume implementation, package version consistency, and existing regression/conformance coverage. The finding concerns the newly exposed helper's supported inputs; legacy run already shares this limitation, and I am not describing a regression in that method or requiring #1406's concurrent prompt-attribution redesign.

Validation: all 53 Python SDK tests passed in an isolated snapshot; contract verification, conformance lint, source compilation, changelog validation, and git diff --check passed. An additional diagnostic using the existing HTTP fixture without mocking turn selection reproduced the failure below. Current-head CI 34933003881, Python release gate 34933003882, and secret scanning 34933003965 are successful. Review Loop dispatch 34933003995 was skipped; all reported checks belong to github-actions and no Review Loop App summary/review exists. I did not rerun the server suite locally.

BinaryBourbon (approved):
Independent adversarial review of d8fd2096b61a6b58b39b4fa1790bbad69629e57f against base ab6bd0bd3195126c5a531f62e79817efc02acd0e: approved. No remaining blocking findings.

Reviewed the full eight-file PR diff, API-shaped serialization and omission/null/false/empty preservation, local validation, Run/TurnFollower lifecycle, channel create/fresh/resume behavior against the actual server/controller and JSON renderer, HTTP failures, regression fixtures, documentation, and package/version metadata. The earlier fresh-channel finding is fixed: created conversations follow turn 1; resumed channels obtain history before sending prompt/images and follow the next turn. The new tests exercise actual turn selection without mocking it.

Validation on this revision: all 54 Python SDK tests passed from source and again against the built wheel installed in an isolated environment; sdist/wheel builds, contract verification, conformance lint, source compilation, changelog/conflict-marker checks and git diff --check passed. Eight additional loopback HTTP diagnostics passed for zero/four prior turns, explicit null/empty images, and busy/gone/credit/sandbox-unavailable prompt errors.

Exact-head GitHub Actions CI 34939111396, Python release gate 34939111566 and secret scanning 34939111546 succeeded. Review Loop dispatch 34939111401 was skipped; there is no Review Loop App review or execution evidence. This is an independent human-account review, not a claim that the Review Loop service verification recipe ran. I did not rerun the Elixir/server suite locally; its current-head CI checks are green. Concurrent prompt attribution remains the existing separate limitation; these diagnostics use one caller.

PR #2237, opened 2026-09-15
feat(cli): create conversations from API-shaped JSON

### LINKED ISSUE #2232: CLI: accept conversation creation JSON from a file or stdin
Parent initiative: #2227

## Problem

`cli/internal/cmd/conv.go` manually assembles a limited creation body from flags. Every new server option otherwise needs a new flag and mapping before CLI users can use it.

## Scope

Provide a JSON file/stdin input path for the existing conversation creation command. Keep convenient flags for common settings. A dedicated flag for every future API field must no longer be required.

## Acceptance criteria

- Document and implement a JSON object input from a file and from stdin, using the existing authenticated API client and POST /api/conversations path.
- Forward request keys without a duplicated CLI field allowlist; preserve explicit null, false, zero and empty values. The API remains responsible for semantic validation and authorization.
- Define behavior when JSON is combined with explicit flags/name-resolution conveniences. Reject ambiguity or use a documented tested precedence; flag defaults must not overwrite supplied JSON values.
- Invalid JSON, a non-object body and input-read errors fail before sending a request.
- Preserve current output/error behavior and existing flag-only calls. Inputs which change creation's response/lifecycle must be handled explicitly; do not assume every response is an immediate conversation.
- Transport/command tests prove a previously unmapped field reaches the server, stdin works, values survive, conflicts are handled and legacy invocations remain compatible.
- Add CLI help and manual examples.

## Boundaries

Conversation creation only. Do not generate every CLI command or introduce typed Go conversation models where existing maps already preserve fields. Coordinate collision and lifecycle semantics with the SDK pilot; file/stdin plumbing can proceed independently.

Shared request-boundary reference: #2228.

jhgaylor:
## Summary

`fountain conv create --file <path|->` accepts the full API-shaped conversation request from a file or stdin and prints the complete API response. New request fields no longer need a CLI flag or body mapping; null, false, empty values and integer precision survive forwarding.

Creation is separate from `run`'s watching behavior, so promptless starts and 202 queued responses work without inventing a turn to follow. Name-based flags cannot be mixed into the JSON command. Existing `run` invocations remain unchanged.

Closes #2232. Part of #2227.

## Validation

- Complete Go module tests and vet passed.
- Command/HTTP tests cover file and stdin, exact request values, authentication, 201/202 responses without extra requests, malformed input, file errors, conflicting arguments and server validation errors.
- Generated CLI reference and manual tests passed.
- Conflict-marker, changelog and diff checks passed.


BinaryBourbon (approved):
## Verdict: approved

Adversarial review of `9942bd1bea3165c482e66ab689c593b1a75c832b`; no blocking defects found across all six changed files.

Traced the registered command through file/stdin decoding, the existing profile-aware authenticated client and POST /api/conversations. The JSON object boundary rejects null, arrays, malformed/trailing input and read failures before HTTP; raw JSON field values preserve omitted keys, explicit null/false/zero/empty values and large integers without an allowlist. Required input, positional arguments and unknown launch flags cannot silently overwrite request fields. Successful creation responses are printed intact without assuming a turn or opening a stream, including queued 202 responses; server validation errors retain the existing client error path. The existing run command is unchanged.

Validation: in an isolated checkout of this exact revision, `go test -mod=readonly -count=1 ./...` and `go vet -mod=readonly ./...` both passed for the complete CLI module, including command/HTTP and CLI documentation parity tests. `git diff --check` passed. Current-head [CI run 34933570257](https://github.com/managoat/fountain/actions/runs/34933570257) and secret scanning passed. Server/manual ExUnit suites were not rerun locally. Review Loop dispatch was skipped and no App review/summary was present; this is an independent review verdict.

Issue #2231, opened 2026-09-15
Swift SDK: generate conversation wire models and adopt request passthrough


jhgaylor:
Parent initiative: #2227

## Problem

`FountainKit/Models/Conversations.swift` manually repeats response properties, request properties, initializers and CodingKeys. The package also ships the map-based `Fountain` product; both run surfaces need the request/execution split without an unrelated transport rewrite.

## Dependencies

Use the pilot's public boundary and compatibility rules. Resolve the relevant published-schema nullability defects through existing issue #2189 before claiming generated Conversation/Turn models correctly accept server responses. Coordinate release conventions with #1414 without absorbing that initiative.

## Acceptance criteria

- Extend the existing OpenAPI/contract pipeline to deterministically generate the conversation request/response models and required referenced types for FountainKit. Choose and document a bounded generation mechanism; avoid a new general-purpose schema framework.
- Preserve wire names, requiredness, dates, defaults and nullability. Request encoding must distinguish unset from explicit null where the API does; ordinary Optional encoding must not silently erase that distinction.
- Preserve public model usability/protocol conformances and existing helper calls, or document an intentional versioned migration. Do not introduce an accidental source-breaking initializer change.
- Route conversation launch requests through the generated request model in FountainKit and a wire mapping in Fountain, separated from local run options and without a handwritten field projection.
- Generation has a stale-output check and does not require Swift-only CI to boot the Elixir server. Reuse committed contract artifacts where appropriate.
- Model fixtures prove decoding of real nullable responses and encoding of omission/null/false values. Both Swift products' contract and conformance suites pass.
- Demonstrate an additive conversation field updating generated output without a handwritten property/CodingKeys edit.

## Boundaries

Generate wire representation only. Keep run handling, streaming, HTTP behavior and permission logic handwritten; do not combine the products or rewrite every resource in this issue.

Shared request-boundary reference: #2228.

PR #2239, opened 2026-09-15
fix(api): make nullable response compositions accept null

### LINKED ISSUE #2189: Conversation.sandbox and Turn.usage say nullable: true and reject null in the published contract
## What

Two properties in the published OpenAPI document say `nullable: true` and do
not accept `null`:

```json
"Conversation.sandbox": {"nullable": true, "oneOf": [{"$ref": "#/components/schemas/Sandbox"}]}
"Turn.usage":           {"nullable": true, "oneOf": [{"$ref": "#/components/schemas/TurnUsage"}]}
```

Both are the shape #1899 fixed on `permission_policy`. In OpenAPI 3.0
`nullable` relaxes the type of the node it sits on and nothing else. These
wrappers carry no `type`, so their `nullable` is vacuous; the referenced
schema is where `type: object` lives, and it never saw the flag. Adding
`type: object` to the wrapper does not help.

`openapi-schema-validator` 0.9.0, `OAS30Validator`:

```
Conversation.sandbox  null -> REJECT: None is not valid under any of the given schemas
Turn.usage            null -> REJECT: None is not valid under any of the given schemas
```

## Why it matters

The server really does send null for both, and `Turn.usage`'s own description
says so — "null while the turn runs, when the runtime reported none, or on
turns that predate the field". So a consumer doing standards-based validation
against the published document rejects a response the server legitimately
returns. Our own suites cannot see it: `OpenApiSpex` resolves null before it
ever reaches the composition, so every cast test passes.

## The fix

`permission_policy` was fixed by hoisting `nullable: true` onto the referenced
component, because all five of its uses are nullable. That is not obviously
right here — `Sandbox` and `TurnUsage` may be referenced from places that
should stay non-null — so check the other references first. Where the component
must stay non-nullable, the alternative is an explicit null branch:

```elixir
%Schema{nullable: true, anyOf: [%Schema{allOf: [Sandbox]}, %Schema{nullable: true, enum: [nil]}]}
```

Both forms were checked against `OAS30Validator` while building the guard.

## Guard

`check_nullable_composition` in `scripts/sdk-contract/build.py` (added in
#2179) catches this class on every `release-and-contract` run. These two are
recorded in its `KNOWN_NOT_NULLABLE` ratchet with this issue number. Fixing
one means deleting its line — the guard fails on a stale entry too, so the
list cannot rot.

Found by that guard while landing #1899 (#2179); left out of that PR so its
diff stays about the permission policy.

jhgaylor:
## Summary

Conversation sandbox and turn usage values can be null in API responses, but their OpenAPI compositions rejected null under standard OpenAPI validation. Add explicit null branches (including teammate turn usage), regenerate the contract and TypeScript declarations, and remove the two nullable-composition exceptions.

The committed-contract regression check pins both repaired fields. Fix its existing allowlist-test cleanup so test mutations cannot leak between cases.

Closes #2189. Prerequisite for Swift wire-model generation in #2231; part of #2227.

## Validation

- Pinned `mix precommit`
- 21 focused schema tests
- 14 nullable-contract tests
- Contract export, TypeScript generation, typecheck and contract verification
- Changelog, conflict-marker and diff checks

Release decision: `release:skip-sdk` — the TypeScript declaration already included null before this repair; regeneration only changes its redundant union spelling. This PR repairs the server document without cutting an SDK release.


sdk/typescript/src/generated/openapi.ts
```diff
@@ -3713,7 +3713,7 @@ export interface components {
             /** @enum {string} */
             runtime: "claude" | "codex" | "gemini" | "opencode" | "acp" | "fountain-fixture";
             runtime_session_id?: string | null;
-            sandbox?: components["schemas"]["Sandbox"] | null;
+            sandbox?: (components["schemas"]["Sandbox"] | (null)) | null;
```
BinaryBourbon, line 3716: [P2] Resolve the SDK release decision for the generated declaration change

This changes a shipped SDK source file while sdk/typescript/package.json remains at 5.1.0. The required [SDK release gate](https://github.com/managoat/fountain/actions/runs/34936316464/job/104274881002) fails at this exact head: `base: 5.1.0`, `head: 5.1.0 (unchanged)`, `shipped: sdk/typescript/src/generated/openapi.ts`. The PR has no release-exception label or documented no-release decision, so it cannot satisfy the repository's required workflow as submitted.

Resolve this under the existing release policy: include coherent releasable package metadata (version/lockfile, USER_AGENT and SDK changelog), or have the maintainer explicitly document and approve the deliberate no-release exception. The generated unions appear semantically equivalent for TypeScript consumers, which may support that decision, but the required gate still needs to be resolved; do not remove the generated changes merely to hide the failure.

BinaryBourbon (changes requested):
Adversarial review of 770dcf107237f6b7aca9e4d20748cd13cb6ecc46 against b03c63ebbc34395a9429f5f65b5c4318d666e60f: changes requested for the unresolved required SDK release gate below.

Reviewed all seven changed files, the full diff, nullable response rendering, non-null component uses, generated TypeScript declarations, projection logic, and guard cleanup. The schema repair looks correct: the explicit branch admits null while Sandbox/TurnUsage remain non-null components; valid object behavior is preserved. No additional source correctness findings.

Validation: 14 focused nullable-contract tests passed; all 96 CI-script tests passed after preparing the disposable snapshot with Git metadata and allowing /dev/stderr access (the first sandboxed/archive run failed on those environment restrictions). Independently cross-checked all 18 guard shapes with OAS30Validator 0.9.0 and checked that the new union accepts null/valid objects while rejecting malformed objects/scalars/arrays. Python and TypeScript contract verification, changelog validation, conflict-marker scan, and git diff --check passed. I did not rerun Elixir suites or full contract export locally.

Exact-head CI 34936316248, including coverage, and secrets 34936316337 are green; SDK release gate 34936316464 is red. Review Loop dispatch 34936316568 was skipped. All reported checks belong to github-actions, and no Review Loop App summary/review exists.

BinaryBourbon (approved):
Independent adversarial re-review of all seven files in the actual merge-base diff for 770dcf107237f6b7aca9e4d20748cd13cb6ecc46, with current base b03c63ebbc34395a9429f5f65b5c4318d666e60f. No remaining findings; approving this revision.

The earlier release-gate finding is resolved: the PR now records a no-release rationale and carries release:skip-sdk. The generated TypeScript change is an equivalent nullable union, so this rationale matches the existing contribution/release policy. The newer exact-head SDK release-gate run 34937165025 is skipped under that policy; the earlier failure 34936316464 is historical.

Reviewed server response rendering (including teammate last-turn usage), all three repaired compositions, non-null Sandbox/TurnUsage component uses, generated declarations, projection behavior, empty exception ratchet, cleanup isolation, documentation, and changelog. OAS30Validator 0.9.0 independently agrees with all 18 guard shapes. Additional diagnostics reconstructed the inspected repaired schemas and verified null/valid/malformed behavior, preserved non-null component semantics, and projection round trips.

Local checks passed: 14 focused nullable tests; all 96 CI-script tests; TypeScript typecheck, all 117 tests (including conformance), build and contract verification; Python contract verification; 28-scenario conformance lint; changelog, conflict-marker and diff checks. Initial broader tests encountered sandbox /dev/stderr and loopback restrictions; the unrestricted reruns passed.

Exact-head GitHub CI 34936316248 (including coverage, server tests, contract export and SDK jobs) and secret scanning 34936316337 are successful. Checks are authored by github-actions. I did not rerun the Elixir suites, a fresh server export/generation, or the full independent Review Loop service verification recipe locally. This is an independent human-account source review with the checks above, not a claim that the service verification ran.

Issue #2175, opened 2026-09-14
Design memo: one owner per conversation lifecycle verb


jhgaylor:
The 2026-09-14 complexity review rated the Conversations subsystem (about 20,500 lines, the two most-churned files in the repo) the biggest engineering priority in the codebase and named the defect: the lifecycle verbs `start`, `wake`, `reapply`, `interrupt`, `terminate`, `resume`, `retire` and `stop` have no single owner. This is the design memo for giving each verb one owner. It was written read-only against `main` at `74ad641f`, re-run 2026-09-15 against `4224258d` after the #2176 credential stack landed (the re-run added one interacting rule and re-mapped every `conversations.ex` citation; nothing else moved), and filed as eleven stage issues under `stack:2175` on 2026-09-15. Nothing is implemented. The full memo is at the bottom, under the fold.

## Headline findings

- **The verb table the review hand-assembled was wrong in four places.** `DetachedRequest.resume_prompt/3` is a pure JSON builder, `redaction.ex` defines no `terminate`, `start_or_resume`/`attach`/`release` were missing, and `execution_guard.ex` (29 `_unsafe_` functions) was omitted although it defines the durable half of five verbs.
- **Interrupt has four ad-hoc doors** into `ExecutionGuard._unsafe_interrupt`; the plan reduces that to two owned ones.
- **`resume` is four unrelated functions sharing a word**, and `retire` is per-thing. Neither gets a single owner; both get renamed by meaning.
- **`execution_*` is three concerns in five files** (policy plus allowance; journal plus scheduler; transport). No rename while ADR 0046 stays Proposed.
- **No fifth `from_state/into_state` pair.** The new owners act on ids, rows and a handle, not server fields.

## Ownership table

| Verb | Owner | Server keeps |
|---|---|---|
| start | new `Conversations.Launch` | nothing |
| wake | new `Conversations.Wake` | nothing |
| reapply | existing `Conversations.Reapply` | `handle_call({:refresh_configuration, ..})`, the `:reapply_prompt` continue |
| interrupt | new `Conversations.Interruption` | `interrupt_turn/1` |
| terminate + release | new `Conversations.Termination` (absorbs `Lifecycle.retire_terminated_sandbox`) | `terminate_machine`, `terminate_kept_machine`, `finish_termination`, OTP `terminate/2` |
| resume | stays split by meaning | `resume_acp_connection` |
| retire | stays per-thing | `retire_bounded_turn` delegate |
[...]

PR #2260, opened 2026-09-15
refactor(conversations): the doors only one owner uses move into that owner

jhgaylor:
Refs #2175

Eight `@doc false` doors that `Fountain.Conversations` kept only so `Launch` and `Wake` could reach them moved into the one owner that actually calls each, as `defp`. Every one had zero bare callers left in `conversations.ex` and exactly one remote caller module (verified with `grep -rn "Conversations.<name>(" apps ee`, tests included).

Moved into `apps/fountain/lib/fountain/conversations/launch.ex` (placed beside the callers that use them):
- `resolve_sandbox_api_access/2` (both clauses)
- `check_sandbox_api_name/2` (both clauses)
- `resolve_sandbox_mode/2` (all three clauses)
- `resolve_parent_id/2` (all three clauses) — its inner `get_conversation/2` call is now qualified `Conversations.get_conversation/2` since that stayed the public door it already was
- `get_root_conversation_id/1`
- `broadcast_graph_update/1`
- `reserve_inference/1` — placed once, ahead of its two callers (`create_attached_conversation/3` and `reserve_initial_conversation/4`)

Moved into `apps/fountain/lib/fountain/conversations/wake.ex`:
- `check_saved_inference/2` — its inner call is now `Conversations.resolve_saved_inference/2`, the existing public door for `Launch.resume_channel/4` that stayed behind

Every "door for ... (#22xx)" / "not part of the context's public surface" comment line that no longer applies was dropped along with the move; substantive explanatory comments (ADR references, lock-order notes, the `#1632`/`#1679`/`#783` rationale) moved unchanged. No new public door was added: every helper these called either stayed a public function already (`get_conversation/2`, `resolve_saved_inference/2`) or was a module call outside `Conversations` (`InferenceBinding.reserve/2`, `Phoenix.PubSub.broadcast/3`).

Untouched: every door with a local caller or two owners (`resolve_vault_id`, `resolve_environment_id`, `resolve_permission_policy`, `resolve_sandbox_provider`, `with_sandbox_lock`, `broadcast_sidebar_update`, `mint_machine_name`, `require_provider_commit_boundary`, `insert_conversation_row`, `after_conversation_created`, `record_execution_allowance_created`, `resolve_admission_limits`, `record_sandbox_usage`, `stamp_terminated_at`, `end_running_turn`).

[...]

PR #2262, opened 2026-09-15
fix(conversations): interrupting a dead conversation reconciles its turn instead of provisioning

jhgaylor:
Refs #2175 (settles open decision 1, decided by Jake, 2026-09-15).

## Before

`Fountain.Conversations.Interruption.interrupt/2` on a registry miss calls
`interrupt_dead/1`, which calls `wake_for_interrupt/1`
(`Wake.wake_conversation_for(conv_id, nil, :interrupt)`). The `purpose:
:interrupt` flag only skipped the allowance preflight — when the sandbox
probe (`Wake.maybe_reuse_sandbox/1`) answered `:create_new` (no sandbox, or
one whose machine is gone), the wake provisioned a brand-new sandbox and
started a server exactly as a prompt would, passing `Billing.check_spend/1`
along the way. The subsequent `:interrupt` call to the new server then
queued behind `handle_continue(:provision)` and timed out to `{:error,
:provisioning}`. No test covered that arm.

## After

`Wake.wake_conversation_for/3` now branches on `purpose` for a `:create_new`
probe:

- `purpose: :work` is unchanged — it still provisions a fresh sandbox
  (`create_fresh_sandbox_and_start/4`).
- `purpose: :interrupt` no longer provisions. It reconciles instead: finds
  the conversation's running turn with `Reattachment.find_running_turn/1`
  and reconciles it through `Conversations._unsafe_orphan_turn/2` — **the
  same door the reattach path already uses for the same shape of loss** (a
  turn the row still calls `running` with no process left to finish it),
  not a new writer. That call sets, on the turn: `status: "interrupted"`,
  `ended_at`, `orphaned_at` (all only if the turn was still `running`); on
  the conversation: `status: "idle"` (only when this is the conversation's
  latest turn and it is still `running`). The arm then answers `{:error,
  :not_running}`, which `interrupt/2` already maps for a caller (`:ok` |
  `{:error, :not_found | :not_running}`).

`ExecutionGuard._unsafe_interrupt/1` (the bounded-journal half) and the
`:reuse`/`:provisioning`/`{:error, _}` arms of `maybe_reuse_sandbox/1` are
untouched. `check_spend/1` still sits where it did; it is simply never
reached on this arm now.

## Audit

`Termination.audit_lifecycle/5` records `conversation.interrupted` only on
`result == :ok`. This arm returns `{:error, :not_running}`, so **no**
`conversation.interrupted` row is written for it — asserted in the new
test. `_unsafe_orphan_turn/2`'s own `conversation.turn.orphaned` audit still
fires unchanged (it is the reattach path's existing writer, not new).

## Docs

Updated `Wake`'s moduledoc, `Interruption.wake_for_interrupt/1`'s doc, and
added a comment on the new `:create_new when purpose == :interrupt` clause
and its `reconcile_dead_interrupt/1` helper; `interrupt/2`'s doc was already
accurate and untouched.

## Test

New file `apps/fountain/test/fountain/conversations/interrupt_dead_reconciles_test.exs`
(2 tests): asserts `Interruption.interrupt(conv.id)` on a `running`
conversation whose sandbox is `terminated` returns `{:error, :not_running}`,
orphans the running turn (`status: "interrupted"`, `orphaned_at` set),
idles the conversation, leaves the sandbox count unchanged, never calls
`Horde.DynamicSupervisor.start_child/2` (`reject/1`), and writes no
`conversation.interrupted` audit row; and a mirror test that the same
dead-sandbox shape still provisions fresh for a plain `Conversations.wake_conversation/2`
(`purpose: :work`), so the change is provably scoped to `:interrupt`.

"Proves unchanged" set, run by explicit path from `apps/fountain`, each
non-zero:

| File | Tests |
|---|---|
| `test/fountain/conversations_wake_test.exs` | included in the 318-test run below |
| `test/fountain/conversations/wake_race_test.exs` | included |
| `test/fountain/conversations/interruption_admission_isolation_test.exs` | included |
| `test/fountain/conversations/bounded_lifecycle_test.exs` | included |
| `test/fountain/conversations/conversation_server_test.exs` | included |
| `test/fountain_web/controllers/conversation_controller_test.exs` | included |
| `test/fountain_web/controllers/saved_allowance_wake_test.exs` | included |
| `test/fountain/audit_guardrail_test.exs` | included |
| `test/fountain/conversations/interrupt_dead_reconciles_test.exs` (new) | 2 |

Run together: `318 tests, 0 failures`.

`mix precommit --full`: `precommit: PASSED (10 stages, 550s)`. Re-ran
`mix precommit compile format credo test` after rebasing onto a moved
`main`: `precommit: PASSED (4 stages, 179s)`.

## Changelog

User-observable (an interrupt on a dead conversation now answers
`not_running` instead of spinning up a machine and timing out):
`changelog.d/2175-interrupt-dead-reconciles.md` added.

Closes nothing on its own; settles #2175 open decision 1.


BinaryBourbon (changes requested):
## Independent Codex adversarial review — correctness and tests

**Verdict: REQUEST_CHANGES**

Reviewed the complete PR diff at `9d7a85d328f181aac798c35ae2d85c20514fa5b5` against merge base `ddf411362dfc12a51bc623800fd5da4c1e268281`. Coverage: `apps/fountain/lib/fountain/conversations/interruption.ex`, `apps/fountain/lib/fountain/conversations/wake.ex`, `apps/fountain/test/fountain/conversations/interrupt_dead_reconciles_test.exs`, and `changelog.d/2175-interrupt-dead-reconciles.md`. Confirmed the inspected changed files match that head.

### P1 — Fence reconciliation to the sandbox that was probed

At `wake.ex:367-370`, reconciliation selects the latest running turn after the provider probe and calls `_unsafe_orphan_turn` without `expected_sandbox_id`. During a delayed not-found response for the old sandbox, another wake can rebind the conversation and admit a successor. This code then selects that successor and marks it interrupted/orphaned, idling its parent even though the not-found result concerned the old sandbox. The recovery transaction's existing binding check only runs when its caller supplies the expectation; its latest-turn check does not reject this case.

Pass `expected_sandbox_id: conv.sandbox_id` to orphan recovery and handle a changed binding without writing the successor. Add a regression that pauses the old provider probe, commits a replacement binding and running successor, then releases the old not-found response; the successor's status and orphan marker must remain unchanged. `orphan_binding_isolation_test.exs` already exercises the underlying guard.

### P2 — Apply the no-provision decision to the registry-timeout arm

The new interrupt branch at `wake.ex:342-343` covers only `:create_new`. A running conversation left on a `pending` or `starting` sandbox instead produces `{:provisioning, sandbox_id}`; when `await_registered` times out, lines 329-330 still call `create_fresh_sandbox_and_start` for an interrupt. With normal admission gates satisfied, that creates a replacement sandbox and starts a server, contrary to the newly documented no-provision behavior for a dead server.

Branch on `purpose` after that timeout too, reconciling for `:interrupt` and preserving fresh provisioning for `:work`. Test both pending and starting rows with a timed-out registry lookup, rejecting `Horde.DynamicSupervisor.start_child/2` and asserting no additional sandbox row.

Validation: `git diff --check` passed. I traced the affected callers, recovery transaction, reservation path and existing isolation tests. These findings are source-supported failure scenarios; I did not execute local ExUnit tests or compile this PR.

BinaryBourbon (changes requested):
## Independent Codex review — lifecycle ownership and recovery

**Verdict: REQUEST_CHANGES** for commit `9d7a85d328f181aac798c35ae2d85c20514fa5b5`.

### [P1] Fence dead-interrupt recovery to the sandbox that was probed

`apps/fountain/lib/fountain/conversations/wake.ex:368-370`

The new path selects the latest running turn after probing the old sandbox, then calls `_unsafe_orphan_turn/2` without `expected_sandbox_id`. While the probe is waiting, a concurrent work wake can replace that sandbox and begin a new turn. The stale probe can then return `:not_found`, and this lookup selects the live replacement turn. Recovery marks it interrupted and orphaned and idles the conversation without interrupting the replacement runtime. Orphaned intervals are excluded from billing/usage attribution.

`ExecutionGuard._unsafe_recover_turn/3` explicitly supports a locked binding check, but omitting the option disables it; unbounded turns have no execution journal to supply an alternative check. Existing `turn_parent_test.exs` assertions cover this exact missing-fence distinction.

Pass `expected_sandbox_id: conv.sandbox_id` from the observed conversation, including an explicit `nil`, and preserve a replacement's turn when the locked binding changed. Add a regression that blocks the old provider probe, rebinds the conversation and starts a successor, then releases the stale probe.

### [P2] Apply the no-provision rule after a pending sandbox times out

`apps/fountain/lib/fountain/conversations/wake.ex:342-343` (also inspect the remaining timeout arm at lines 329-330)

The new guard handles only the literal `:create_new` result. A `pending` or `starting` sandbox takes `{:provisioning, sandbox_id}` instead; when no server appears during `await_registered/1`, that arm still calls `create_fresh_sandbox_and_start/4` for `purpose == :interrupt`. A dead provision therefore still allocates a replacement and starts compute on an interrupt's behalf, contrary to the stated no-provision behavior. A running co-tenant can retain its status when `follow_cotenants/3` moves it onto a pending replacement, so this is not restricted to a conversation that the interrupt door necessarily rejects.

Make the timeout fallback purpose-aware while retaining the wait/handoff for a server that appears. Cover running conversations on pending and starting sandboxes with a registry timeout and assert that no sandbox or child is created.

I reviewed all four changed files and traced the scoped API interrupt entry, wake dispatch, orphan recovery locks, side effects, and existing/new regression assertions. No cross-tenant lookup regression found. I did not run local tests; the two added tests cover a terminated sandbox and the ordinary work path, not these interleavings or timeout fallback.

BinaryBourbon (changes requested):
## Independent Codex adversarial review — API/product contracts and maintainability

**Verdict: REQUEST_CHANGES** for `9d7a85d328f181aac798c35ae2d85c20514fa5b5` against merge base `ddf411362dfc12a51bc623800fd5da4c1e268281`.

Reviewed all four changed files and traced interruption through the tenant-scoped controller, wake selection, and orphan recovery.

### P1 — Fence reconciliation to the sandbox that was found dead

`apps/fountain/lib/fountain/conversations/wake.ex:367–370`

A provider probe can finish after a concurrent work wake has rebound the conversation to a replacement sandbox and started another turn. The new helper then queries the latest running turn and calls `_unsafe_orphan_turn/2` without checking that the conversation still belongs to the sandbox it probed. It can therefore mark the replacement's live turn interrupted/orphaned and the conversation idle while its runtime continues. Orphan attribution also excludes that live work from normal usage accounting.

Pass `expected_sandbox_id: conv.sandbox_id` into the existing recovery helper, whose parent-lock check already rejects a changed binding. Add a deterministic regression that pauses the dead-machine probe, rebinds and starts a successor, then releases the interrupt and verifies that the successor remains running.

### P2 — Apply the no-provision policy after a registry-settle timeout too

`apps/fountain/lib/fountain/conversations/wake.ex:342–346` (remaining path at lines 306–330)

The new purpose guard covers only `:create_new`. A running conversation with a stranded `pending`/`starting` sandbox takes `{:provisioning, sandbox_id}`; if its server never appears, the timeout still unconditionally calls `create_fresh_sandbox_and_start/4`, even for `:interrupt`. That allocates a replacement for the cancellation request, contradicting the documented and explicitly settled “interrupting a dead server never provisions” policy.

Route an interrupt's registry timeout to dead-turn reconciliation, with the same binding protection, while preserving fresh provisioning for `:work`. Cover pending/starting sandbox rows and assert no new sandbox or server is created.

The new tests cover a terminal sandbox and the ordinary work wake, but neither case above. `git diff --check` passed. I did not execute Elixir tests locally; these findings come from the control flow and existing recovery contracts.

jhgaylor:
Pushed a fix for both findings from the three independent Codex reviews (commit `4f0b120c`, on top of the original `a644762d` after a rebase onto `main`).

## P1 — fenced reconciliation to the sandbox that was probed

`reconcile_dead_interrupt/1` now passes `expected_sandbox_id: conv.sandbox_id` into `Conversations._unsafe_orphan_turn/3`. `conv.sandbox_id` is the sandbox this wake actually probed, read before the probe ran. `_unsafe_orphan_turn/3`'s existing recovery transaction re-locks the conversation row and rejects a changed binding as `{:error, :ownership_changed}`, writing nothing to either the turn or the conversation in that case. Either outcome, the arm still answers `{:error, :not_running}` — the interrupt was for the dead incarnation, and a live successor found instead is not this wake's to touch.

New regression: **"a stale probe response after a concurrent rebind does not touch the live successor"**. Stubs the sprite probe so that, before it answers `{:error, :not_found}` for the old sandbox, it rebinds the conversation to a freshly inserted replacement sandbox and admits a running successor turn — simulating a concurrent `:work` wake winning the race while this interrupt's probe is still in flight. Asserts the successor turn is still `running` with `orphaned_at` nil, the conversation is still `running` and bound to the replacement, and the interrupt still answers `{:error, :not_running}`.

## P2 — the registry-timeout arm is now purpose-aware too

The `{:provisioning, sandbox_id}` arm's `:timeout` branch (reached when a `pending`/`starting` sandbox's server never turns up during the registry-settle wait) now checks `purpose` the same way the flat `:create_new` probe result does: `:interrupt` reconciles through the same fenced `reconcile_dead_interrupt/1` helper, `:work` keeps provisioning fresh via `create_fresh_sandbox_and_start/4` unchanged. The wait-and-handoff when a server does appear during the settle window is untouched.

New regressions (one `test` in a `for status <- ~w(pending starting)` loop, so two runtime tests): a running conversation with a running turn on a `pending` sandbox, and again on a `starting` sandbox, with no server registered so `await_registered/1` times out against the real registry (config/test.exs's 150ms settle window, no stub needed). Both assert `Horde.DynamicSupervisor.start_child/2` is never called (`reject/1`), the sandbox row count is unchanged, the turn is orphaned (`status: "interrupted"`, `orphaned_at` set), and the result is `{:error, :not_running}`.

## Docs and changelog

Reworded the `Wake` moduledoc paragraph, `Interruption.wake_for_interrupt/1`'s doc, and the changelog fragment to say "dead or stranded sandbox," since the no-provision rule now covers both the flat `:create_new` probe and the registry-timeout arm, not just the former.

## Test count

`interrupt_dead_reconciles_test.exs` now has 5 tests total: the original 2 (dead-sandbox reconcile vs. plain-prompt still-provisions), plus the 3 new ones above (1 stale-probe/P1, 2 pending+starting timeout/P2).

## Gate

`mise exec -- mix precommit compile format credo test` (after rebasing onto `main` at `82c57d9c`, resolving one mechanical conflict in `wake.ex` where PR #2260 added an unrelated private function, `check_saved_inference/2`, at the same insertion point as this PR's `reconcile_dead_interrupt/1` — kept both, no logic changed): `precommit: PASSED (4 stages, 132s)`. The full `mix precommit --full` had already passed on the lib change alone before this fix; a rerun on the reused lane DB hit the known #2178 accumulated-`audit_events` shape (unrelated to this PR), cleared with `delete from audit_events` and reconfirmed.

BinaryBourbon (approved):
## Independent Codex adversarial re-review — correctness and tests

**Verdict: APPROVE**

Reviewed the entire four-file diff at `4f0b120c77d8fe195d57dfd784df5aa354570111` against merge base `82c57d9c45ca1c2a8634c683a150bc7c7d85c68e`, including the original behavior change and subsequent fixes. All inspected changed files match that head.

### Previous findings

- **P1, stale probe orphaning a rebound successor: resolved.** `reconcile_dead_interrupt/1` now passes the observed `conv.sandbox_id` as `expected_sandbox_id`. Traced that option through `_unsafe_orphan_turn/3` into `_unsafe_recover_turn/3`: the parent is locked and a changed binding is rejected before journal or turn mutation. The regression changes the binding and inserts a running successor during the provider probe, then verifies the successor remains running and unorphaned.
- **P2, provisioning after registry timeout: resolved.** The `pending`/`starting` timeout arm now reconciles for `:interrupt` and retains fresh provisioning for `:work`. Both stranded statuses have regressions rejecting server startup and checking that no sandbox row is added. The original dead-sandbox test still verifies turn reconciliation, parent idling and audit behavior; its companion retains the ordinary work-wake path.

No new actionable correctness defect found. The existing conditional recovery remains idempotent if another process already ended the selected turn. The target branch at `c671cb9dd07e4b6b88eac2587089d447600fcd00` differs from the merge base only in ADR documentation and still exports the wake delegate used by the new test.

Coverage:

- `apps/fountain/lib/fountain/conversations/interruption.ex`
- `apps/fountain/lib/fountain/conversations/wake.ex`
- `apps/fountain/test/fountain/conversations/interrupt_dead_reconciles_test.exs`
- `changelog.d/2175-interrupt-dead-reconciles.md`

Validation: `git diff --check` passed. The coordinator verified [CI run 34954296184](https://github.com/managoat/fountain/actions/runs/34954296184) succeeded for this exact head, including all six partitions, static checks and coverage. I inspected the regression assertions and related recovery code; I did not run local ExUnit or compilation checks.

BinaryBourbon (approved):
## Independent Codex re-review — lifecycle ownership and recovery

**Verdict: APPROVE** for commit `4f0b120c77d8fe195d57dfd784df5aa354570111`.

I reviewed the complete four-file PR and the affected interrupt, wake, orphan-recovery, and locked execution-guard paths. Both findings from my previous review are resolved:

- **P1, stale probe recovering a replacement's turn:** reconciliation now supplies `expected_sandbox_id: conv.sandbox_id`. The existing recovery guard compares it against the locked parent before changing the turn or execution journal, including when the expected binding is explicitly `nil`. The new regression simulates a rebind before the old probe returns and asserts that the successor remains running and has no orphan timestamp.
- **P2, provisioning after a pending/starting timeout:** the registry-timeout branch now distinguishes interrupt from work. Interrupt uses the same fenced reconciliation helper; ordinary work retains fresh provisioning. New tests cover both pending and starting sandbox states and reject child creation while asserting unchanged sandbox count.

No additional actionable security defect found. Tenant-scoped authorization remains at the API caller, and recovery retains the established conditional writes and post-commit events. The original terminated-sandbox and ordinary-work regressions remain present.

**Validation limits:** I inspected the regression assertions but did not run local tests or a runtime race reproduction. The coordinating reviewer verified passing [GitHub CI for this exact head](https://github.com/managoat/fountain/actions/runs/34954296184). I also checked that the test's `Conversations.wake_conversation` call remains valid on the supplied current target revision.

BinaryBourbon (approved):
## Independent Codex adversarial re-review — API contracts and maintainability

**Verdict: APPROVE** for `4f0b120c77d8fe195d57dfd784df5aa354570111` against merge base `82c57d9c45ca1c2a8634c683a150bc7c7d85c68e`.

Re-reviewed the complete four-file change and the surrounding interruption, wake, recovery and controller paths. Both findings from my previous review are resolved:

- **P1, stale-probe reconciliation:** the helper now supplies `expected_sandbox_id: conv.sandbox_id`. Existing recovery checks that binding under the parent lock before writing. The new deterministic regression rebinds during the provider probe and verifies that the successor remains running and un-orphaned.
- **P2, registry timeout provisioning:** both `:create_new` and the pending/starting registry-timeout path reconcile for `:interrupt`, while ordinary work retains fresh provisioning. New tests cover both pending and starting rows and reject new worker creation.

The result still reaches the existing HTTP 409 `not_running` contract. Orphan reconciliation uses the established conditional writes and audit path; the documentation and changelog now describe both dead and stranded sandboxes. No new blocking defect found.

`git diff --check` passed. I inspected test assertions but did not execute Elixir tests locally. The coordinating reviewer verified successful exact-head [GitHub CI](https://github.com/managoat/fountain/actions/runs/34954296184), including all six test partitions, static checks and coverage.

Issue #2269, opened 2026-09-15
Swift SDK: finish legacy and behavioral wire-model seams


jhgaylor:
Parent: #2251; initiative: #2247. Follows #2267 and the resource-generation batch.

## Remaining finite inventory

Straightforward resource DTOs are moving to the existing contract generator. Four seams still combine wire declarations with public compatibility or client behavior:

- `AuthMe`: `onboardingState` is a published optional property absent from the current wire contract. Decide and document whether to preserve it through a compatibility extension or retire it in a versioned migration; generate the remaining contract-owned account payload.
- `AdminUserPage`: retain the `users`/`hasMore` convenience while consuming generated data/meta envelope types instead of duplicating page wire keys.
- `ConversationBindingUpdate` / `ConversationReapplyRequest`: retain the three-state binding API and omission/null semantics while deriving supported request fields from the contract.
- `LogEvent`, `Block`, `PermissionOption`, `PermissionRequest`: inventory raw-versus-normalized event shapes before separating contract-owned DTOs from custom decoding and stream normalization. Preserve permissions, plan bodies, unknown events and partial/malformed payload behavior.

## Acceptance

- [x] Every family above has generated wire declarations or an explicit, documented compatibility decision with no unowned migration.
- [x] Preserve public names, optional-property source APIs, accepted historical decoding and behavior; version intentional breaks.
- [x] Add compiled consumer and non-null/null/omission regressions for affected seams; retain independent endpoint/stream conformance.
- [x] Show an additive field propagates without another handwritten registry wherever generation applies.
- [x] Delete superseded declarations and update `contributing/swift-wire-models.md` and #2251.

`JSONValue`, `ConversationInputField`, string-backed enum wrappers, resource methods and stream-following behavior remain handwritten by design. This is not a client-behavior generator or a reason to merge the two Swift products. No protocol-retirement work is included.

BinaryBourbon:
## Decisions for the four seams (2026-09-15)

**1. `AuthMe` — done, in #2289 (green, awaiting review).** Generated from `AuthMeResponse`. `onboardingState` is **retired**, not preserved: the server dropped `users.onboarding_state` in v0.16.0 (#1393, ADR 0038, NC-6 of ADR 0007) and the TypeScript SDK dropped it in its 1.17.0, so Swift was the last client carrying it and it has decoded `nil` from every reachable server for two releases. `role` and `email_verified` are pinned — the guard from #2287 demanded both by name on its first real migration.

[...]

Issue #2251, opened 2026-09-15
Swift SDK: eliminate remaining duplicated wire models in bounded batches


jhgaylor:
Parent follow-on tracker: #2247. Foundation: #2227.

## Problem

#2240 generates the conversation boundary, but FountainKit still has handwritten wire models, including reused Sandbox shapes and models in Models/Agents.swift and Models/Sandboxes.swift. The generator also has explicit compatibility type mappings. We should account for the remaining duplication rather than treating the first generated file as complete coverage.

## Acceptance criteria

- [x] Inventory handwritten DTOs and reused/aliased referenced types; distinguish contract-derived data from behavior/domain conveniences. Record which are generated, intentionally handwritten, or require a migration.
- [x] Prioritize the duplicated conversation dependencies (including Sandbox and usage/value-type mappings) and then other repeated wire shapes. Split execution into bounded PRs or linked child issues with a finite inventory.
- [x] Extend the existing contract/generation pipeline for those shapes without creating another source of field lists or a general client-behavior generator.
- [x] Preserve requiredness, dates, server-default omission, nullability, unknown enum handling and request omission-vs-null semantics; test actual non-null and nullable payloads.
- [x] Preserve public names, conformances and initializers where possible. Document and version intentional migrations rather than accidentally changing source APIs.
- [x] Delete superseded handwritten model/property/CodingKey declarations and only genuinely redundant verification claims. Keep endpoint coverage and behavior-dependent manifest claims.
- [x] Demonstrate an additive field in each newly covered family reaches generated output with no handwritten property registration; keep stale-output checks usable without booting Elixir in Swift CI.

Depends on #2231/#2240. Both Swift products remain supported; merging them is not a prerequisite.

## Bounded implementation batches (2026-09-15)

1. #2267 — Sandbox, SandboxDetail and nested types, Runner, conversation tree, and Usage aliases. Review-requested optional-property fixes pushed; current CI green, re-review pending.
2. #2270 — 22 resource roots across agent, environment/vault, connection, team, account/catalog/apply and admin models, plus nested payloads. Deletes 620 handwritten model lines. Stacks on #2267; CI is green. Local validation: 118 Swift tests, strict format/release build, both-product propagation, 105 tooling tests, per-root/nested additive probes, and external-import property/optional compatibility compilation.
[...]

Issue #2248, opened 2026-09-15
API/SDK: verify the field-tax rollout and close the foundation loop


jhgaylor:
## Completion verified — September 18, 2026

Closing as completed after checking the acceptance evidence on September 18, 2026.

- Every acceptance criterion is checked.
- #2270 merged and #2227 is closed.
- The final genuine-field exercise landed in #2376, following the prompts-route work in #2374. The evidence is no longer only the synthetic probe.
- The SDK resume-branch findings recorded here were addressed separately in #2390; the CLI/ACP and shared conformance follow-ups landed in #2403 and #2405 (#2389 and #2388 are closed).
- Independent SDK ownership/release work #1414 is also closed by #2404. Idempotent retries remain separate accepted work in #2375.

This closes the rollout-verification scope, without claiming the separate idempotency program is implemented. Parent tracker: #2247.

---

## Original scope and evidence (preserved)

Parent follow-on tracker: #2247. Foundation: #2227.

## Scope

Close the operational/documentation loop for #2227 once its implementation PRs land. This is separate from the ongoing debt-retirement work.

## Acceptance criteria

- [x] Rebase/retarget the remaining stacked PRs as their bases merge; #2241 runs against the actual integrated main, has green required CI and independent review before landing.
- [x] Verify the TypeScript, Python and Elixir publish/tag jobs and installable artifacts for their intended versions. Verify Swift and CLI availability under their existing server/release conventions; do not claim an unreleased merge is available to consumers.
- [x] Refresh ADR 0056's implementation-status paragraph (it currently calls non-TypeScript paths unbuilt), SDK documentation, and the #2227 checklist with exact merged/released references.
- [x] Run the generation/check and propagation entry points from a clean integrated checkout. Preserve independent conformance expectations.
- [x] Use the next genuine conversation field as an end-to-end confirmation: record server changes, generated diffs and handwritten client edits actually required. Do not invent a public field just for the exercise; until a real addition occurs, explicitly record that the evidence is the synthetic probe.
- [x] Close #2227 only when its own completion criteria and release/documentation follow-through are satisfied; remaining debt stays in this follow-on tracker.

Dependencies: #2227's implementation PRs. Broader release-policy redesign remains #1414.

## Verification evidence (2026-09-15)

- All seven foundation PRs merged. Integrated checkout `ddf411362dfc12a51bc623800fd5da4c1e268281` passed the full generation check and cross-client propagation probes, including a temporary nested referenced field. No fake public API field was added.
[...]

Issue #2247, opened 2026-09-15
Tracker: finish API and SDK simplification after the field-tax foundation


jhgaylor:
## Outcome

The accepted API/SDK paydown checklist is complete as of September 18, 2026. The API-shaped request boundary, first-party caller migration, shared launch implementations, Swift wire generation, protocol retirement, SDK maintenance conventions, and caller correlation all have merged completion evidence.

## Reconciled checklist

- [x] #2248 — rollout and genuine-field verification. #2270 merged, #2227 closed, and #2376 supplied the real `client_request_id` exercise. #2248 is now closed.
- [x] #2249 — first-party callers migrated; Fountain #2264 and managoat/demos#77 are the recorded evidence.
- [x] #2250 — shared launch implementation landed in #2265.
- [x] #2251 — Swift generation completed through the bounded batches and #2269's recorded follow-through; the issue is closed.
- [x] #2252 — OpenAI-compatible and AG-UI retirement completed under ADR 0057; the issue is closed. The external site cleanup is also recorded as complete in the prior tracker history.
- [x] #1414 — ownership, independent releases/tags, and SDK catalog landed in #2404 on September 18. The issue is closed; the old statement that this work had not started is superseded.
- [x] #1406 — accepted correlation scope delivered by #2374, #2376, and #2390. CLI/ACP forwarding and shared resume conformance followed in #2403 and #2405; #2389 and #2388 are closed.

The live issue states and merged PRs were rechecked for this reconciliation. This is completion of the named paydown scope, not a new production smoke or a new verification of registry artifacts.

## Continuing work has its own records

- #2375 remains open for idempotent retries, split into #2407 and #2408. Correlation is implemented; idempotent retries are not. The September 17 scope split on #1406 and the current #2375 plan establish this boundary.
- #2094 continues the separate compatibility-retirement program.
- New API/SDK defects or capabilities should be concrete issues with their own acceptance criteria, rather than keeping this completed tracker open indefinitely.

Earlier inventories, release evidence, and sequencing notes remain in the comments and issue edit history. No priority, owner, iteration, or deadline is changed by closing this tracker.

jhgaylor:
## Checklist re-checked against main (2026-09-15, 13:00Z)

Three children moved after this tracker was last written at 11:22Z. Re-verified against `origin/main` rather than against the issue text.

[...]

PR #2270, opened 2026-09-15
refactor(swift): generate resource wire models from the contract

jhgaylor:
## Summary

FountainKit's resource DTOs still repeated contract fields and CodingKeys. Extend the existing Swift generator to agent, environment/vault, connection, team, account/catalog/apply and admin resources, including nested envelope payloads. Delete 620 lines of handwritten model declarations.

Preserve public names, nested types, initializer ordering, historical optional properties, string-backed unknown enums and dynamic JSONValue APIs. Agent inputs retain their string policy convenience and gain numeric policy values and explicit null. Shared Skill/Secret schemas reject incompatible definitions.

A type that already shipped handwritten must not become harder to decode, so properties exposed here for the first time are optional even where the contract requires them: `Catalog.firstRequest`, every `Catalog.mcpServers` item field and `ConnectionProvider.kind`, `.platform`, `.redirectURI`, `.scopes`, `.tokenHosts`. A response from a server older than those fields still decodes. Wholly new types take contract requiredness directly. `redirect_uri` is named `redirectURI`, matching the six sibling URL properties on the same type, before the name is released.

The finite inventory in `contributing/swift-wire-models.md` records coverage, both compatibility tables' sizes, the handwritten DTOs outside `Models/` and the four remaining legacy/custom-decoding seams in #2269. Resource methods, pagination, stream following and permission behavior remain independently tested. Swift docs also identify the now-published v0.17.1 request APIs and separate released changes from this unreleased generation work.

## Validation

- 119 Swift tests passed, including contract/conformance, resource decoding, optional/null fields, numeric policies and old initializer order; existing matrix/live skips retained.
- `payloadsFromAnOlderServerStillDecode` decodes payloads that omit every `OPTIONAL_COMPAT` key. Verified by deleting pins: it fails with `keyNotFound`, and passes again once restored. `ConfigTests` keeps its pre-PR minimal-catalog fixtures (`{}`, `{"apps": null}`, `{"apps": {}}`), so that file leaves the diff entirely.
- 106 Python tooling tests passed, including independent additive-field probes across all 22 resource roots, nested payloads, shared skills, both secret schemas, and a new guard that every `OPTIONAL_COMPAT` pin still names a live property. Verified by misspelling a pin: the guard fails.
- Strict Swift format and release build with warnings as errors passed. Generated-output check passed for both products.
- The additive-field propagation probe covers the conversation roots only, not the 22 resource roots; the resource roots are covered by the Python probes above. No committed test imports FountainKit non-`@testable`, so public access is a compile-time property of the generated output rather than a gate.
- Conflict-marker, changelog and diff-whitespace checks passed.

Refs #2251, #2248, #2247. Remaining compatibility migration: #2269. Review follow-ups (inline-name collision with a real schema, `additionalProperties` over sibling `properties`, array-of-enum typing, dead `REUSED`): #2277.


BinaryBourbon (changes requested):
**Changes requested.** Adversarial review across four dimensions: generator correctness, source/wire compatibility, test efficacy, and docs/citations. Reviewed `64659816` against `7e1d8c32` (this PR's own commit only, not the stack below it).

The refactor itself is sound and I verified the headline claims rather than taking them:

- `generate-swift.py --check` → exit 0; committed output reproduces from the contract, and the check genuinely fails when I flipped `firstRequest` to optional.
- `swift test` → **118 passed**, exactly as claimed. 5 skips, all pre-existing. `swift format lint --strict` exit 0; `swift build -c release -Xswiftc -warnings-as-errors` clean.
- 105 Python tests pass; `RESOURCE_ROOTS` is genuinely 22 and the probe iterates all of them.
- **No existing property changed optionality**, no property was dropped, no type silently changed, and initializer order matches all four deleted handwritten inits field-for-field. `swift-api-digester` (validated against a control) reports one removed decl in the whole module.
- The conflict guards are real: `Incompatible reused inline shape: AgentSkillsItem` and `Incompatible secret field: key` both fire on broken input.
- `camel()` gaining `ids`→`IDs`/`ip`→`IP` is name-*preserving* (reproduces `allowedVaultIDs`, `requestIP`), not a rename.

## Blocking

**1. Three undisclosed decode regressions, only one of them disclosed.** Six newly-required properties across two types make previously-decodable payloads throw:

- `ConversationWire.generated.swift:884,888,889,892,895` — `ConnectionProvider` went from 7 properties (only `id` required) to 26 with five required: `kind`, `platform`, `redirectUri`, `scopes`, `tokenHosts`. The exact field set the deleted `Connections.swift` documented now fails with `keyNotFound("kind")`.
- `:1138` — `Catalog.firstRequest`. `Catalog` has no custom `init(from:)`, so this fails the **entire** `client.catalog()` call, which also backs `conversationURL()`.
- `:1466-1471` — `CatalogMcpServersItem` has all five required; one item missing `verified_on` fails the enclosing `Catalog` decode.

Against a current server none of these fire — `ConnectionProviderJSON.summary/1` and `catalog_controller.ex:80` emit every key unconditionally, and all are `required:` in `schemas.ex`. The exposure is SDK-newer-than-server skew, real for a self-hosted AGPL product with no documented minimum server version, plus mocks and fakes. `README.md:100` advertises `from: "0.17.1"`, which SPM resolves up to `<1.0.0`, so existing consumers adopt this without opting in.

The issue is disclosure: `sdk/swift/CHANGELOG.md` names only `Catalog.firstRequest`. `changelog.d/2251-swift-resource-models.md:3` — the fragment that becomes the public changelog — says only "preserving existing public names, dynamic JSON APIs and nullable request semantics" and mentions none of the six. Either add the five `ConnectionProvider` fields and `CatalogMcpServersItem` to `OPTIONAL_COMPAT` alongside their siblings, or disclose all six in the fragment.

**2. `ConfigTests.swift:120` — the test that would have caught this was neutered instead of the behavior pinned.** The new `catalogFixture()` helper injects `first_request` into *every* fixture, including the `"{}"`, `{"apps": null}` and `{"apps": {}}` arguments at `:100-104` that exist precisely to exercise a minimal catalog. Those arguments are now dead. Restoring the pre-PR fixture gives `keyNotFound("first_request")`. Note `OPTIONAL_COMPAT` pins every *other* pre-existing `Catalog` property (`apps`, `models`, `package_managers`, `runtimes`, `sandbox_providers`) — `first_request` was the one missed. There is no direct `Catalog` decode test anywhere in the suite.

**3. `:889` — `redirectUri` is the lone `Uri` among seven URL-ish properties** (`authorizeURL`, `connectURL`, `mcpURL`, `revokeURL`, `tokenURL`, `userinfoURL`). Newly exposed, so this mints a permanent public-API inconsistency. One line in `camel()` now; a breaking change after release.

## Non-blocking

**4. `OPTIONAL_COMPAT` is load-bearing and largely unguarded.** It carries the whole backward-compat story, and deleting `('Connection','scopes')`, `('ConnectionProvider','configured')`, `('TeamSchedule','enabled')`, `('ApiKey','created_at')` flips four public properties `Optional` → non-`Optional` with **every gate green**: `--check` exit 0, 9/9 Python, 118/118 Swift. The only signal was a compiler warning in Tests, and `-warnings-as-errors` compiles Sources only. Whether a pin is guarded is incidental — assertions written `x?.y` catch it, `x == false` / `x != nil` do not.

**5. `contributing/swift-wire-models.md:25-27` deleted the numbers that made "finite" auditable, in the PR that grew the table 9 → 44** (20 owner types). The old text enumerated all nine entries by name; the replacement says "finite" and discloses no growth, two lines above asserting the table "is not a registry to extend." `:32-33` reads exhaustive but omits `networking_config` (×2) and `AgentVersion.config` of the 19 `TYPE_OVERRIDES`.

**6. The inventory silently covers `Models/` only.** `PageMeta` (`Client/APIClient.swift:223`), `APIErrorBody` (`Errors/FountainError.swift:6`) and TeamResource's inline `Body` structs (`:22,:61,:124`) are hand-maintained contract-shaped DTOs with hand-written `CodingKeys`, in no inventory row and none of #2269's four seams. #2251 asks to inventory handwritten DTOs without scoping to `Models/`.

**7. `contributing/conversation-callers.md:48,52` states two things that are now false.** "The migration is implemented in demos#77 … It awaits review/merge" — demos#77 merged `10:49:47Z`, ~6 min after this commit. "demos#76 owns generated input adoption" — closed COMPLETED `10:49:48Z`. This is a standing contributor record, not a snapshot, and the edit is arguably out of scope for a Swift-models PR.

**8. `generate-swift.py:133,137` — `TYPE_OVERRIDES` makes the typed `Repository` schema unreachable.** The override is consulted before the `ref` branch, so adding a *required* `Repository` field changes nothing and raises nothing — the one place "new fields follow contract requiredness" provably does not hold. Source-compatible with the deleted `[JSONValue]?` so likely deliberate, but it is the only new table entry with no comment saying so, and nothing pins it.

**9. Two PR-body validation claims do not cover this change.** The "Swift propagation probe" injects only into `ConversationCreateRequest`/`Conversation`/`Turn` — zero hits for any of the 22 resource roots. The "external-import compile check" has nothing committed: all 15 test files use `@testable import`, and `ResourceWireTests` calls the internal `APIClient.decode`, so the suite cannot prove public access. Both claims are true; neither is a gate.

## Follow-ups, not this PR

Two generator holes are **byte-identical at `7e1d8c32`** and inherited from #2267, so they belong in a follow-up:

- An inline object whose synthesized `<Owner><Key>` name collides with a real schema is silently discarded and typed as the unrelated schema — the reuse guard checks `self.nested` but `render()` resolves `self.schemas` first. Confirmed live: making `Connection.provider` an inline `{slug,name}` types it as `ConnectionProvider` with no error. Not reachable on today's contract (the synthesized-name ∩ schema-name intersection is empty), but this PR enlarges the surface considerably.
- `additionalProperties` silently wins over sibling `properties`, contradicting the module docstring. This PR pulls the contract's only both-shaped node (`networking_config`) into scope for the first time, though a `TYPE_OVERRIDE` neutralizes it today.

Also minor: array-of-enum loses typing (`Catalog.sandbox_api_access` → `[String]?` while scalar `Conversation.sandbox_api_access` is `SandboxAPIAccess`), and `REUSED = set()` at `:20` is never populated, making `if ref not in REUSED` at `:224` dead code.

Items 1–3 are small fixes. Everything else can land as follow-ups if you prefer.

jhgaylor:
Thanks — items 1–3 are fixed, plus 4–7 and the comment in 8. Rebased onto `main` (parent #2267 landed as `44c70b61`) and retargeted; the diff is now against `main`. New commit: `4d4617ff`.

**1. Decode regressions.** Took the first option, for all six properties and the sixth type's siblings: `('Catalog','first_request')`, `('ConnectionProvider', kind/platform/redirect_uri/scopes/token_hosts)` and all five `CatalogMcpServersItem` fields are pinned in `OPTIONAL_COMPAT`, under a comment that names the second reason the table exists — a property this SDK exposes for the first time on a type that already shipped handwritten. Wholly new types still take contract requiredness. Both changelogs now disclose the optional exposure rather than a required field.

**2. `ConfigTests`.** Reverted to the pre-PR fixtures; `catalogFixture()` is gone and the `{}` / `{"apps": null}` / `{"apps": {}}` arguments are live again. The file is byte-identical to `main` and has left the diff. The behavior is pinned directly instead: `payloadsFromAnOlderServerStillDecode` decodes a catalog without `first_request` (and a bare `{}`), an `mcp_servers` item with only `slug`, the legacy seven-field `ConnectionProvider`, and payloads missing `Connection.scopes`, `APIKey.created_at` and `TeamSchedule.enabled`/`one_off`. Verified by deleting those five pins: `keyNotFound("first_request")`, failing; restored from a copy, passing.

**3. `redirectUri`.** `camel()` learns `uri`→`URI` and `uris`→`URIs`, alongside the existing `id`/`ids`. The property is `redirectURI`, done before release.

**4. `OPTIONAL_COMPAT` guard.** Two now. The Swift test above fails on a deleted pin whatever the assertion style elsewhere. `test_optional_compat_pins_reach_a_live_property` fails when a pin stops naming a live property — the same flip by a different route, since a contract rename makes a pin a silent no-op. `render()` keeps its models for it. Verified by misspelling `redirect_uri` as `redirect_url`: fails.

**5. The numbers.** Restored and made auditable: 9 entries over 5 owner types → 55 over 21, stated in the same paragraph as "not a registry to extend". `TYPE_OVERRIDES` is now given as 19 entries and the enumeration includes networking config and agent-version config.

**6. Inventory scope.** Added a row for `PageMeta`, `APIErrorBody` and TeamResource's inline request bodies, marked handwritten outside `Models/`, unmigrated, and outside #2269's four seams.

**7. `conversation-callers.md`.** Corrected: demos#76 closed completed with demos#77, merged 2026-09-15. The claim that no deployed app was migrated is kept and now says why merged there is not deployed here.

**8. `Repository`.** Deliberate; the override now says so in a comment. The rest of 8 and the two inherited generator holes are #2277.

**9. PR body.** Rewritten. Both claims now say what they cover and that neither is a gate: the propagation probe covers the conversation roots, the 22 resource roots are covered by the Python probes, and no committed test imports FountainKit non-`@testable`.

Gates: 119 Swift tests, 106 Python, `--check` exit 0, strict format, release build with warnings as errors, propagation probe.

BinaryBourbon (approved):
**Approved.** Re-review of `4d4617ff` against my `CHANGES_REQUESTED` on `64659816`. All three blocking items are resolved, and I verified each by measurement rather than by reading the commit message. Two small things below are worth landing first, but neither is user-facing and neither warrants another round trip.

First, the rebase is faithful: `069a116a`'s own patch is byte-identical to the `64659816` I reviewed, so nothing was quietly revised on the way onto `main`. The whole delta is the fix commit.

## Blocking items — all resolved

**1. The six newly-required properties are gone.** I re-ran my inventory (every non-optional property in the generated file, resolved against the handwritten baseline at `069a116a^`): **0 newly-required properties on previously-handwritten types**, down from 6. `ConnectionProvider.kind/platform/redirectURI/scopes/tokenHosts` and all five `CatalogMcpServersItem` fields are `Optional`; `Catalog.firstRequest` is `Optional`. A parallel enumeration of all 71 non-optional stored properties, matching nested types by enclosing type rather than generated name, also found **0 violations** of the new rule.

**2. `ConfigTests.swift` is byte-identical to the pre-generation original.** `diff` against `069a116a^` is empty — the `catalogFixture` padding helper is gone and the `"{}"`, `{"apps": null}` and `{"apps": {}}` arguments genuinely exercise the minimal-catalog path again. This is the right fix; restoring the fixture and pinning the behavior separately is better than what I asked for.

**3. `redirectURI`** now matches its six `*URL` siblings, via `camel()` learning `uri`→`URI`. Fixed before release rather than after.

Suites: `generate-swift.py --check` exit 0, Python 9 → **10 tests**, Swift 118 → **119**. Both new guards are real, not decorative. I broke the Python one myself — renaming a pinned key to simulate a contract rename fails `test_optional_compat_pins_reach_a_live_property` with a clear message, which closes the "a pin becomes a silent no-op" failure mode properly.

## Worth fixing before merge (not blocking)

**`ResourceWireTests.swift:156` — one of the eleven pins this commit adds is not covered by the guard added to cover it.** The fixture is `{"runtimes":["claude"],"mcp_servers":[{"slug":"m1"}]}`, which *supplies* `slug`, so `('CatalogMcpServersItem', 'slug')` is the one new pin whose absence is never exercised. Deleting that pin and regenerating:

```
1469:  public var slug: String?  ->  public var slug: String
python3 -m unittest ... test_swift_generation  ->  Ran 10 tests ... OK
swift test                                     ->  119 tests passed
generate-swift.py --check                      ->  exit 0
```

A public property flips to non-`Optional` — the exact whole-response decode failure this pin exists to prevent — with every gate green. The other ten new pins each correctly go red (`keyNotFound("dcr"), codingPath: [mcp_servers, Index 0]`). Fix is two characters: `"mcp_servers":[{"slug":"m1"},{}]`.

**The "every" claim is not true and will be relied on.** `ResourceWireTests.swift:151-153` says "*Every* property pinned in the generator's `OPTIONAL_COMPAT` decodes from its absence; dropping a pin fails here"; the commit message and `contributing/swift-wire-models.md:36` repeat it. Measured against the 55-entry table, **23 of 55 pins are actually exercised**. Six have their owner decoded but the key supplied (`CatalogMcpServersItem.slug`, `ConnectionProvider.configured/connect_url/env_key/name/slug`); 26 have an owner the guard never decodes (the nine sandbox/runner pins, `AdminSandbox.status`, `AdminUser.role`, `AgentVersion.config/inserted_at`, `ApiKeyCreatedResponse.name/prefix`, `AuditEvent.inserted_at`, `CatalogSandboxProviders.default/enabled`, `SearchHit.snippet/ts`, `Secret.environment_id`, `TeammateLastTurn.*`, `TeammatePresence.label`). Several are caught incidentally by other tests; `CatalogSandboxProviders.default` and `.enabled` are silently droppable the same way `slug` is. Narrowing the sentence to what it does costs nothing and stops the next contributor trusting it.

## Non-blocking notes

**4. The documented rule does not describe the code.** `swift-wire-models.md:26-34` and `generate-swift.py:192-196` both end "wholly new types take contract requiredness directly." `CatalogMcpServersItem` never shipped handwritten — it is a wholly new type — yet all five fields are pinned; `CatalogFirstRequest`, also wholly new and under an equally newly exposed parent, is not. Two siblings under one parent, opposite treatment, and the rule predicts neither. The real distinction looks like field history (an old server could emit `mcp_servers` items lacking `verified_on`; no server ever emitted a partial `first_request`), which is a better rule than the one written down.

**5. Both changelogs' blanket clause is false, though unreachably so.** "Properties these models expose for the first time decode as optional, so a response from an older server still decodes" — `CatalogFirstRequest.curl/placeholders/prompt/typescript` are newly exposed and non-optional, so a present-but-partial `first_request` fails the whole response. I had this checked against the server: `catalog_controller.ex:85-103` is the only construction site and is an unconditional four-key map, the key was introduced whole in #1443, and no other endpoint emits it — **no code path produces the failing payload**. Wording, not risk.

**6. `TYPE_OVERRIDES` has no guard, while the new prose implies it does.** `generate-swift.py:425-427` says "the compatibility tables' own guard" (plural) and the doc says "Neither table is a registry to extend", but only `OPTIONAL_COMPAT` got one. Making an entry stale (`('Agent','mcp_servers')` → a dead key) changes `mcpServers` from `JSONValue?` to `[String: JSONValue]?` — a source-breaking public type change — with 119 Swift and 10 Python green.

**7. "Two kinds of entry" is a comment, not a distinction the code can make.** `:154` and `:197` are two `.update({...})` calls into one flat set, and `fields()` does a single membership test. Nothing can tell a historically-optional pin from a newly-exposed one, so the halves will merge by accident on the next edit. A second set, or a dict with a reason, would make the rule enforceable.

**8. The DTO inventory is still incomplete.** `swift-wire-models.md:22` adds `PageMeta`, `APIErrorBody` and "TeamResource request bodies", but leaves out `TeamResource.swift:61-66` and `:124-129` (two `Reply` structs, `Decodable`, hand-written `CodingKeys` with `conversation_id` — response DTOs, which "request bodies" does not cover), plus `AdminResource.swift:60` and `ConversationsResource.swift:63`.

**9. Five pins are no-ops** — `('Secret','environment_id')` and the four `TeammateLastTurn` entries — because the contract already leaves those optional; removing `Secret.environment_id` produces a byte-identical file. Pre-existing from `069a116a`, and the new guard passes them since it only checks presence.

Doc counts I re-counted rather than trusted: `OPTIONAL_COMPAT` 55 entries over 21 owners and `TYPE_OVERRIDES` 19 both match what the doc now states, and the seven named override categories cover all 19 exactly. The demos#76/#77 wording is now correct in tense and causality.

Approving because nothing here is a user-facing regression and the fix demonstrably closed what I blocked on. Items 4–9 are fine as follow-ups; I'd land the two-character fixture fix and the "every" wording with this PR.

BinaryBourbon:
Pushed `711198ac` onto this branch — the two items I flagged as worth fixing before merge. Tests and contributor docs only; the generator is untouched, so the generated file is unchanged and `--check` stays in sync.

**The guard now covers the pins it claimed.** `payloadsFromAnOlderServerStillDecode` supplied six pinned keys rather than omitting them, and never decoded `CatalogSandboxProviders` at all. The fixture now decodes a second `mcp_servers` entry (`{}`), an empty `sandbox_providers` object, and a bare `ConnectionProvider` beside the populated one. Verified by deleting each pin, regenerating and running the full suite:

```
RED  (the guard catches it)      ('CatalogMcpServersItem','slug')
RED  (the guard catches it)      ('CatalogSandboxProviders','default')
RED  (the guard catches it)      ('CatalogSandboxProviders','enabled')
RED  (the guard catches it)      ('ConnectionProvider','configured')
RED  (the guard catches it)      ('ConnectionProvider','name')
```

All five were green before. `('CatalogMcpServersItem','slug')` is the one that mattered most — it was added by `4d4617ff` to close a review finding and was itself droppable with every gate green.

**The docs now say what the guards actually do.** Dropped the "every property pinned" claim from both the test docstring and `swift-wire-models.md`; the pin test cannot see a deleted pin, the decode tests catch only the keys their fixtures omit, and a new pin needs the omission that proves it. Also replaced "wholly new types take contract requiredness directly", which the code contradicts: `CatalogMcpServersItem` is wholly new and pinned, `CatalogFirstRequest` is wholly new and not. The doc now records the actual distinction — a nested type reached through a newly exposed property is pinned, and `CatalogFirstRequest` is the exception because `first_request` arrived whole in #1443.

Green here: 119 Swift, 106 Python, `swift format lint --strict` clean, `generate-swift.py --check` in sync.

Items 4–9 from my review are untouched and still fine as follow-ups — in particular `TYPE_OVERRIDES` has no guard at all, and "two kinds of entry" in `OPTIONAL_COMPAT` is a comment rather than something the code can enforce.

PR #2275, opened 2026-09-15
docs: define OpenAI-compatible and AG-UI retirement boundaries

jhgaylor:
## Outcome and current status

Retire Fountain's OpenAI-compatible and AG-UI public APIs and their request-defined tool bridge. Preserve native conversations and the actively used pattern of agent-configured tools calling client applications.

**This PR is documentation only.** It adds proposed ADR 0057, the source/consumer inventory and its index entry. No endpoint, runtime behavior or database column has been removed. The plan is reviewed at `0993695ef42003bc407d94e6955c6e9e1b204427`; implementation remains a follow-up under #2252. Documentation approval/CI is not implementation or deployment evidence.

### Read first

- [ADR 0057: proposal and cutover](https://github.com/managoat/fountain/blob/0993695ef42003bc407d94e6955c6e9e1b204427/decisions/0057-retire-public-compatibility-protocols.md)
- [Inventory: consumers, exact source map, mixed tests, documentation and validation](https://github.com/managoat/fountain/blob/0993695ef42003bc407d94e6955c6e9e1b204427/contributing/protocol-retirement.md)
- Parent implementation issue #2252; initiative #2247.
- Current checkout's `CLAUDE.md`, `CONTRIBUTING.md`, `contributing/server.md`, `contributing/docs.md` and `scripts/ci/README.md`. Recheck current repository instructions before editing.

### Maintainer context already established

- Work through this initiative **one item at a time**. Start with #2252; do not start the remaining Swift work #2269 in parallel. Someone else is handling #2270; recheck its status and avoid overwriting its work.
- The maintainer reports no known consumers needing a transition period and no users explicitly relying on request-defined tools. Do not repeat that discovery question. This is maintainer evidence, not a production traffic/database census.
- People **do** configure tools on agents that call client applications. Preserve those tools, their MCP configuration, credential substitution and callback authentication. A URL pointing back to an application does not make it part of the retiring bridge.
- Native `runRequest` replaces delegation calls, not a stock `ChatOpenAI`/`HttpAgent` integration or a `role: tool` continuation loop. Preserve truthful documentation of that loss.
- Improving the ergonomics of agent-configured application tools is separate future work; this retirement does not claim to solve it.

### Approval boundary and repository state

[...]

Issue #2252, opened 2026-09-15
API: retire OpenAI-compatible and AG-UI surfaces with an explicit migration


jhgaylor:
Parent follow-on tracker: #2247. Foundation: #2227.

## Outcome

Capture the separate simplification discussed before #2227: retire the OpenAI-compatible and AG-UI protocol surfaces and the code that exists solely to support them. This concerns Fountain's public compatibility endpoints, not OpenAI/Codex inference/runtime support.

## Starting evidence

- apps/fountain/lib/fountain_web/controllers/openai_controller.ex and agui_controller.ex
- /v1/chat/completions, /v1/models, /v1/models/{model}, /api/agui/:agent_id and related router pipelines
- apps/fountain/lib/fountain/caller_tools.ex and the caller-tool callback/bridge paths
- openai_compat flags, API schemas, sdk/contract/omissions.json, fixtures and docs/integrations/{openai-compatible,openbot,langchain}.md

#2094 previously preserved these as current integrations. Retirement needs an explicit consumer/cutover record, not an assumption that protocol adaptation is unused.

## Acceptance criteria

- [ ] Inventory current consumers and shared versus exclusive dependencies; record the migration path to native conversation APIs/SDKs and any capability without a replacement.
- [ ] Record the compatibility/release decision and removal behavior for these public endpoints, including consumer notices/docs and any data/active-run handling needed at cutover.
- [ ] Remove controllers, routes/pipelines, feature flags, schemas and protocol-only event/message translations. Regenerate contracts and update omissions/endpoint decisions.
- [ ] Trace and remove caller-tool bridge plumbing, callbacks, persistence fields and dependencies that become unused. Preserve core agent/runtime tools, native permissions, native SSE and any genuinely shared behavior. Use forward migrations for persistent data changes.
- [ ] Remove or rewrite stale integration pages/examples and reconcile #1196's AG-UI documentation plans with the retirement decision.
- [ ] Verify retired routes have the documented behavior and native conversation creation/streaming/permissions still work; run relevant server, schema/SDK, docs and release checks.
- [ ] Record deleted production code, remaining shared dependencies, and any explicitly deferred removal with its own issue.

This is a separate removal initiative, not required to merge the field-tax foundation. Do not remove OpenAI model pricing, credentials, inference adapters or Codex support by name association.

## Inventory and cutover proposal (2026-09-15)

[...]

Issue #2273, opened 2026-09-15
Conversations: drop retired caller-tool storage after the deployment floor advances


jhgaylor:
Parent: #2252. This is the final persistence cleanup after protocol/bridge code is retired, not permission to drop a column while old servers still use it.

`conversations.caller_tools` was added by `20260826041202_add_caller_tools_to_conversations.exs`. It stores request-defined tool schemas. Parked calls/results/timers live in ConversationServer memory. Agent-configured MCP tools, callback credentials, native permission requests and retained log history are different data and must stay.

## Deployment gate and acceptance

- [ ] The preceding retirement release no longer reads/writes `caller_tools`, including create, attach, queued launch and reapply/session paths.
- [ ] Record that every server node/job using the old schema has stopped and document the minimum rollback version. Leaving the physical column for the first release keeps old Ecto readers safe during rolling deployment.
- [ ] Use a new forward migration to drop only `conversations.caller_tools`; never edit the applied 20260826041202 migration.
- [ ] Test upgrading a database with representative legacy rows. Preserve conversations, sandboxes, channel bindings, agent MCP configuration, callback-key fields and historical logs/webhook deliveries.
- [ ] Document that dropping the column destroys its obsolete definitions; re-adding the column on downgrade does not restore their contents. Record backup/rollback handling before release.
- [ ] Regenerate/check relevant contracts if any public surface changes and reconcile #2252's final persistence item.

The maintainer reports no known request-defined-tool users needing a transition period, but that is not a database census or permission to break active agent-configured application tools.

BinaryBourbon:
## The schema field is gone; this issue's precondition is now the deployment floor alone

Recording state so this is actionable when the floor moves. The #2252 stack removes **every read and write** of `conversations.caller_tools`: the Ecto field and its `cast`, `Conversations.set_caller_tools/3`, both `Launch` write sites, the queued-attrs allow-list key, and the MCP injection that read it. That is what makes a server running this code a **non-reader**, which is the condition this issue was waiting on.

The applied migration `20260826041202_add_caller_tools_to_conversations.exs` is untouched, and the physical column and its rows are intact.

**What is left here**, unchanged from the original plan:

- [ ] Confirm no server or job older than the retirement release is still running. An older Ecto reader selects the column, so this is a rolling-deploy boundary, not a code question.
[...]

Issue #1423, opened 2026-09-03
Buzz harness test collides with itself across concurrent suites: unique_integer is per-node, TMPDIR is not


jhgaylor:
`apps/fountain/test/fountain/buzz/harness_test.exs:29` builds its scratch directory as:

```elixir
dir = Path.join(System.tmp_dir!(), "buzz-harness-#{System.unique_integer([:positive])}")
```

`System.unique_integer/1` is unique **per BEAM node**, not per machine. Two `mix test` runs on the same host are two nodes, so both start their counters low and hand out the same small integers. `System.tmp_dir!()` is shared between them. The two runs therefore collide on the same path, and `on_exit` in one deletes the directory the other is still launching from.

The visible symptom is a failure in `Fountain.Buzz.HarnessTest` plus lines like:

```
.../priv/buzz-acp-launch.sh: line 21: /var/folders/.../T/buzz-harness-198/buzz-acp: No such file or directory
```

The failing example varies with timing. The file passes on its own every time.

## Why it is worth fixing rather than tolerating

It is invisible in CI, because the six partitions run on six separate machines and nothing else on a runner is running the suite. It only appears when two suites overlap on one host, which is exactly what happens when several worktrees or agents build in parallel — and then it looks like a real regression in whatever branch happened to be running. It cost time in this session twice, and my first diagnosis of it was wrong in a way that mattered: I told an agent it would stop once the other suites finished, but the collision depends on overlap at any moment, not on how many suites there are.

## Fix

Seed the path with something node-unique rather than node-local. Any of:

- `System.pid()` (the OS pid of the BEAM) joined with the unique integer,
- `:erlang.unique_integer` plus `node()`, or
- simply `Path.join(System.tmp_dir!(), "buzz-harness-#{System.pid()}-#{System.unique_integer([:positive])}")`.

The `on_exit` cleanup then only ever removes a directory this run created.

## Worth checking while there

Grep for other tests building a shared-`$TMPDIR` path from `System.unique_integer/1` alone. The same reasoning applies to every one of them, and this is unlikely to be the only instance.

Found while running four suites concurrently for the ADR 0038 onboarding work.

jhgaylor:
I grepped for the sibling instances I suggested checking for, and there are five more. Every one builds a shared-`$TMPDIR` path from `System.unique_integer/1` alone, so every one has the same collision:

| File | Line | Prefix |
|---|---|---|
[...]

PR #2279, opened 2026-09-15
feat(api)!: retire the OpenAI-compatible and AG-UI protocol surfaces (stage 1: routes, controllers and the public cutover)

BinaryBourbon:
Stage 1 of the ADR 0057 retirement (#2252): the **wire surface only**. The tool bridge those dialects fed stays on `main` after this PR and is removed in stage 2, so this one is reviewable as "four routes and their two controllers, gone".

Implementation-phase authorization for the removal was given by the maintainer after #2275's documentation PR was queued. This PR removes source; it removes no data and deploys nothing.

## What goes

| Retired | Was |
|---|---|
| `POST /v1/chat/completions` | `OpenAIController` (1,175 lines) |
| `GET /v1/models` | ″ |
| `GET /v1/models/:model` | ″ |
| `POST /api/agui/:agent_id` | `AguiController` (686 lines) |

Both controllers are exclusive production code for their protocol and go whole, with their routes, their two suites, and their four operations plus five exclusive schemas (`ChatCompletion`, `ChatCompletionRequest`, `Model`, `OpenAIError`, `RunAgentInput`) out of the contract and the generated TypeScript types.

`POST /api/mcp/caller/:conversation_id` — the fifth retired path — stays until stage 2, because removing it means removing `Fountain.CallerTools` and the parked-call plumbing with it.

## The retirement behaviour is three answers, not one

ADR 0057 proposed "the ordinary unmatched-route 404". That is true of `/v1` and **not** true of `/api/agui`, and I checked rather than asserted it. `apps/fountain/test/fountain_web/protocol_retirement_test.exs` records what a client actually gets:

| Path | JSON client | No key | `Accept: text/event-stream` |
|---|---|---|---|
| `/v1/*` | 404 | 404 | 404 |
| `/api/agui/*` | 404 | **401** | **406** |

`/v1` matches no route at all, so `NoRouteError` renders 404 before anything authenticates. `/api/agui` falls through to `FountainWeb.Plugs.ExtensionDispatch`, whose scope pipes through `:accepts_json` inside `:api` — so `TenantAPIAuth` answers a keyless call 401, and content negotiation refuses the event-stream `Accept` an AG-UI client actually sends with 406, both before dispatch runs.

Neither status is a surviving dialect handler, and the suite asserts that rather than claiming it: each case is compared against an `/api` path that never existed, and the two answer identically. **ADR 0057's cutover text needs this correction**; I will make it in the stage that touches the ADR rather than editing an approved document from a code PR.

## Added after review

Two defects the review caught, both fixed in `ba96b61d`:

**The caller-tool MCP append moved into this PR.** Removing the two dialect controllers removed the only parties that could ever hand a parked caller-tool call back to a client — but `McpServers.fountain_served/2` still advertised `fountain-caller` for any row with persisted `caller_tools`. On a legacy `openai:`/`agui:` row continued natively, which ADR 0057 promises works, the agent could still select one, park a call nobody could answer, and cycle 60-second waits until the permission deadline failed it. A delayed tool failure is worse than no tool, so the append goes here rather than with the rest of the bridge in #2280: **not advertised is not selectable**. Three regressions cover it, including a team-channel case so an empty list cannot pass for a correct result, and an inverted `session/new` assertion in the ACP suite that would have caught it.

**The ADRs are updated in this PR, not a later one.** A change that builds part of an architecture decision has to say so. ADR 0057 goes to Accepted with a status section naming exactly what this change builds and what each later change in the stack adds — written so that section is updated by each of them and always describes the tree it merges into. ADR 0035 is superseded rather than edited (`status: deprecated`, `adr_status: "Superseded by 0057"`), body kept as the record of what was built and why. `decisions/index.md` regenerated; `okf validate decisions` → `valid: true` with the six pre-existing actor-convention warnings.

## Preserved

- OpenAI and Codex inference, credentials, model catalog and runtimes. They share the vendor's name with the retired gateway and nothing else; no provider-side `/v1` path was touched.
- Native create/attach/prompt/`/events`/`/stream`/permissions, and the SDKs over them.
- `SavedAllowanceChannelTest` keeps every native assertion — it loses only its `:openai` and `:agui` arms.
- Existing conversations with `openai:` / `agui:` channel bindings. Those are opaque strings on the row; the `channel_id/1` helpers that built them were controller-local. Such a conversation continues by its tenant-scoped native id.

## TypeScript SDK 6.0.0

Major, deliberately. `schemas.ts` re-exports `components` and `paths`, and `index.ts` re-exports that, so the five removed schema keys were part of the published type surface even though no hand-written method used them — the same reasoning by which 5.1.0 was a *minor* for adding one. No runtime export, method, option or request changed. The release gate agrees: `Release looks well-formed. Merging this publishes 6.0.0.`

## Validation

Run against a private `DATABASE_URL`, because the shared test database had 187 live sandbox rows and 1,644 committed audit rows left in it and produced three unrelated `:fleet_full` / audit-paging failures.

- `mix precommit` — **PASSED** (9 stages, 198s).
- 325 tests, 0 failures across the retirement suite, `api_spec_test`, `api_pipeline_responses_test`, `saved_allowance_channel_test`, `attach_test`, `mcp_servers_test`, `callback_key_test`, `pending_test`, `conversation_server_acp_test`, `conversation_server_redaction_test`, `audit_guardrail_test`, and the bridge suites that stay green until stage 2.
- `sdk-contract/generate.py --check`, `sdk/conformance/lint.py`, `changelog.py check`, `conflict-markers.py`, `git diff --check` — all pass.
- The contract diff is exactly the four operations and five schemas: no operation added, no surviving operation or schema changed. Checked semantically, not by eyeball, and specifically for #2270's Swift work leaking in — it did not; Swift generation produced no diff.
- The new suite was verified by reverting: re-adding the two routes turns 6 of its 10 tests red.

## The documentation cutover is in this PR

Third review round, correctly: `contributing/protocol-retirement.md` — the approved base — says the source-removal PR carries **"the public migration guide"**, and forbids leaving shipped examples calling removed endpoints. Splitting that off was my staging error, not a deferral the plan allowed.

So this PR now also carries:

- **Four migration pages at their existing URLs** — [openai-compatible](https://github.com/managoat/fountain/blob/api/retire-compat-protocols-routes/docs/integrations/openai-compatible.md), [openbot](https://github.com/managoat/fountain/blob/api/retire-compat-protocols-routes/docs/integrations/openbot.md), [langchain](https://github.com/managoat/fountain/blob/api/retire-compat-protocols-routes/docs/integrations/langchain.md), [gateways](https://github.com/managoat/fountain/blob/api/retire-compat-protocols-routes/docs/integrations/gateways.md). Each says what a call gets today, what to use instead, and **what has no replacement** — the no-code integration itself, which nothing native gives back. The AG-UI page documents the 406 specifically, since `Accept: text/event-stream` is what those clients send.
- **Cross-links reconciled**: `clients.md`, `feature-status.md`, `nav.yml`, docs index, build guide, glossary, troubleshooting.
- **The three examples removed.** Every one called a deleted route; none can be ported. `examples_test.go`'s skip list named two litellm files and fails on an entry matching nothing — which is how it caught them — so it empties rather than disappears. `contributing/conversation-callers.md` held the deepagents example pending *"protocol retirement #2252 provides its migration"*; it now records that there is none.
- **Flag guidance**: `.env.example`, `.env.compose.example`, `docker-compose.yml` and `configuration.md` stop naming `openai_compat` as what `FEATURE_FLAGS_ON` turns on. Worded for what is true *here* — the endpoints are gone so the flag gates nothing; deleting the flag itself is still #2282.

Docs gate: **5 suites, 0 failures**. `go test ./internal/manifest/` green.

## What is still a follow-up, and why that is safe

#2280 (bridge internals) and #2282 (the flag) stay separate. Both are invisible to a user at this merge: nothing advertises the caller tools any more, and nothing documented promises the flag does anything. Neither leaves a claim in the release that this PR contradicts — which was the actual defect.

Refs #2252, #2247.


jhgaylor (changes requested):
Adversarial review of `731c97e0dbe8114aed14a2efea17b6aca852f2f4` against recorded base `115d3e5e49b76351cf66843472fe3d60c8017241`.

Requesting changes for two medium-severity defects:

1. **The staged removal leaves legacy caller tools advertised but impossible for any client to answer.** At `apps/fountain/lib/fountain_web/router.ex:653`, this change removes the OpenAI/AG-UI controllers that were the only callers of `pending_caller_calls/1` and `answer_caller_tools/2`, while the retained path at `apps/fountain/lib/fountain_web/router.ex:530` and `McpServers.fountain_served/2` still inject `fountain-caller` whenever a persisted conversation has `caller_tools`. Trigger: continue an existing `openai:`/`agui:` conversation through its native ID, as the PR says is supported, and let the agent select one of its persisted caller-defined tools. The MCP call parks and can only cycle through 60-second waits until the permission deadline resolves it as an error; there is no remaining wire path that exposes the call to the client or accepts its answer. This turns the promised native continuation into a delayed tool failure until the follow-up PR lands. Minimal remedy: make the caller-bridge deactivation atomic with this route removal (for example, include the stage-2 removal here), and add a regression covering a native prompt on a legacy row with persisted `caller_tools`; at minimum, stop injecting those tools in this change so they cannot be selected without an answer path.

2. **The authoritative ADRs become false at merge.** The route removal above implements part of the architecture change, but `decisions/0057-retire-public-compatibility-protocols.md:4,14-18` will still say the decision is Proposed and that no endpoint or source implementation has been removed, while `decisions/0035-openai-compatible-endpoint.md:16-20` will still say the removed gateway is accepted, built behavior. This is directly contrary to the base repository rule to update an ADR in the PR that builds the change, and the approved retirement inventory likewise requires the implementation PR to supersede ADR 0035 and update ADR 0057. Trigger: merge this PR by itself. Consequence: Fountain's source-of-truth architecture record contradicts the running router and misdirects subsequent implementation/release work. Minimal remedy: update ADR 0057 to accurately record this partial implementation and the observed 404/401/406 behavior, supersede the affected parts of ADR 0035, and regenerate/validate `decisions/index.md` in this PR. Later cleanup stages can extend that record, but the merged state must not claim nothing has been built.

I inspected all supplied paths: `apps/fountain/lib/fountain/conversations/prompt_input.ex`, both deleted protocol controllers, `apps/fountain/lib/fountain_web/router.ex`, both deleted controller suites, `saved_allowance_channel_test.exs`, `protocol_retirement_test.exs`, the changelog fragment, `sdk/contract/contract.json`, `sdk/contract/omissions.json`, and the four TypeScript package/generated files. I also traced the retained caller-tool/session launch path and applicable ADR/policy text. A semantic contract comparison found exactly the four intended operations and five exclusive schemas removed, with no shared operation/schema changes.

All reported required CI jobs are green, including six test partitions, static analysis, release/API contract, SDK jobs, coverage, and `CI required`; the docs job is skipped. I ran `git diff --check` and the semantic contract comparison locally. I did not run local server tests because the shared worktree is on another dirty branch; CI is supporting evidence and does not exercise the legacy-row/native-continuation gap above.

jhgaylor (changes requested):
Adversarial whole-PR re-review of head `67302a2eb230040441d310de04e9e3d20df48ed8` against recorded base `978efe75a31eb45b8e735db6efb3d7022fadf9b2`.

Requesting changes for three medium-severity defects. Both blockers from my review of the prior head remain unresolved: the merge from current `main` did not deactivate the caller-tool bridge or update the architecture record.

1. **Legacy caller tools are still advertised after their only client answer path is removed.** `apps/fountain/lib/fountain_web/router.ex:530-533` retains `POST /api/mcp/caller/:conversation_id`, and `apps/fountain/lib/fountain/conversations/mcp_servers.ex:122-149` still appends `Fountain.CallerTools.conversation_mcp_servers/2` whenever a persisted conversation has `caller_tools`. This PR removes both controllers that called `pending_caller_calls/1` and `answer_caller_tools/2`. Trigger: continue an existing `openai:`/`agui:` conversation through its native ID, as the PR promises, and let the agent select one of its persisted caller-defined tools. The MCP call parks and repeatedly returns `pending` until its deadline because no surviving wire surface exposes the call to a client or accepts the answer; the native turn degrades into a delayed tool failure. Minimal remedy: deactivate/remove the request-tool bridge atomically with these routes (or at minimum stop injecting `fountain-caller` for persisted rows), and add a regression that exercises a native prompt on a legacy row containing `caller_tools`.

2. **The authoritative ADRs still contradict the implementation.** `decisions/0057-retire-public-compatibility-protocols.md:14-18` says the decision is Proposed, no endpoint/source is removed, and ADR 0035 remains the shipped behavior; `decisions/0035-openai-compatible-endpoint.md:16-20,59-66` still says the deleted gateway is accepted, built behavior. Trigger: merge this PR. Fountain's source-of-truth architecture record then directs maintainers toward routes absent from the router and falsely claims the implementation has not started. This also violates the approved-base inventory's cutover sequence and the repository rule to update an ADR in the PR that builds the change. Minimal remedy: update ADR 0057 to record the implemented partial state and actual 404/401/406 behavior, supersede the affected parts of ADR 0035, and regenerate/validate `decisions/index.md` in this PR.

3. **The public manual and shipped examples still instruct users to call endpoints this PR deletes.** For example, `docs/integrations/openai-compatible.md:3-19,31-38,80-90` says Fountain serves `/v1/chat/completions`, provides a working curl command, and links the runnable example; `docs/integrations/openbot.md:9-23,45-56` says `/api/agui/:agent_id` is a working AG-UI endpoint and tells operators to configure/test it. `docs/integrations/clients.md`, `gateways.md`, `langchain.md`, `reference/feature-status.md`, the three protocol examples, and the env/Compose flag guidance repeat those claims. Trigger: a user follows the manual or an included example after this release. They receive the newly asserted 404/401/406 responses instead of the advertised integration. The approved-base `contributing/protocol-retirement.md` explicitly requires the source-removal PR to include public migration guidance and says not to leave shipped examples calling removed endpoints. Minimal remedy: make the documented cutover atomic here—retain the inbound URLs as truthful migration pages, port or remove the broken examples and cross-links, and remove the obsolete `openai_compat` configuration guidance; run the focused docs checks.

Coverage: I inspected all 15 changed paths: `apps/fountain/lib/fountain/conversations/prompt_input.ex`; both deleted protocol controllers; `apps/fountain/lib/fountain_web/router.ex`; both deleted controller suites; `apps/fountain/test/fountain_web/controllers/saved_allowance_channel_test.exs`; `apps/fountain/test/fountain_web/protocol_retirement_test.exs`; `changelog.d/2252-retire-openai-agui-routes.md`; `sdk/contract/contract.json`; `sdk/contract/omissions.json`; `sdk/typescript/CHANGELOG.md`; `sdk/typescript/package.json`; `sdk/typescript/src/generated/openapi.ts`; and `sdk/typescript/src/http.ts`. I also traced the retained caller-tool controller/state/session assembly and launch paths, native/API/SDK callers, public docs/examples/config, tenant/auth boundaries, the approved-base policy/instructions, ADRs 0035/0057, and the retirement inventory. A semantic contract comparison found exactly the intended four operations and five exclusive schemas removed, with no surviving operation/schema changes.

Checks: all current reported CI jobs are green, including six test partitions, static analysis, release/API contract, TypeScript/Python/Elixir/Swift SDK jobs, CLI/plugins, core distribution/Compose, secret scanning, coverage, the SDK release gate, and `CI required`; the Docs job is skipped. Locally, `git diff --check` passed and the semantic contract comparison passed. I did not rerun the server suite in the shared dirty worktree. Green CI supports the mechanical deletion and generated contract, but it does not exercise the legacy-row/native-continuation failure or validate documentation omitted from this diff.

BinaryBourbon:
Both findings were real and both are fixed in `ba96b61d`. Thank you — the first one in particular was a defect my staging created, not one the plan had.

**1. Caller tools advertised with no answer path.** Correct, and worse than a cosmetic ordering problem: on a legacy `openai:`/`agui:` row continued natively — which ADR 0057 explicitly promises works — the agent could select a persisted tool, park a call, and cycle 60-second waits until the permission deadline failed it. A delayed tool failure is worse than no tool.

I took your minimal remedy rather than folding stage 2 in, because the whole append is four lines and the rest of the bridge is not: `McpServers.fountain_served/2` no longer appends `fountain-caller` at all, so those tools are no longer selectable. `Fountain.CallerTools`, the adapter, the route and the parked-call plumbing still go in #2280.

Three regressions, at both layers:
- a row with persisted `caller_tools` gets no bridge server;
- the same row **on the team channel** still gets its team tools and not the bridge — so an empty list cannot pass as a correct result (the vacuity guard your finding implies);
- `conversation_server_acp_test.exs` had a test asserting the *opposite* — that such a row's tools ride in `session/new`. It is inverted, which is the same guarantee one layer down and the thing that would have caught this.

**2. ADRs false at merge.** Also correct, and it is the base-repo rule. ADR 0057 is now Accepted with a status section that says exactly what this change builds and what each later change in the stack adds — written so that section is updated by each of them and always describes the tree it merges into. ADR 0035 is superseded rather than edited (`status: deprecated`, `adr_status: "Superseded by 0057"`), with its body kept as the record of what was built and why.

ADR 0057 also now carries the 404/401/406 correction: its cutover text said requests "receive the ordinary unmatched-route HTTP 404", which holds for `/v1` and not for the `/api` paths. `decisions/index.md` regenerated; `okf validate decisions` → `valid: true` with the six pre-existing actor-convention warnings.

**Validation:** `mix precommit` PASSED (9 stages); full core suite **5,915 tests, 0 failures**, 2 skipped; `scripts/test-docs.sh` 5 suites 0 failures; changelog/conflict/whitespace clean.

Two things to note on your review method, since you flagged not running local tests: the shared checkout genuinely did have another agent working in it — that is also what produced a contract-check flap I briefly mis-filed as #2281 (now closed as not-a-defect). I ran the suite against a private `DATABASE_URL`. One failure you would otherwise hit is environmental: `Fountain.AuditTest`'s window assertion fails once a reused test database has accumulated committed audit rows (#2178), and is green after `TRUNCATE audit_events`.

jhgaylor (changes requested):
Adversarial whole-PR re-review of head `ba96b61d70782d4cafe3463c52002c64dbbeba58` against recorded base `978efe75a31eb45b8e735db6efb3d7022fadf9b2`.

Verdict: **CHANGES_REQUESTED** for one medium-severity defect.

Disposition of prior blockers:

1. **Resolved — legacy caller tools no longer have a stranded selectable path.** `apps/fountain/lib/fountain/conversations/mcp_servers.ex:120-149` now omits the caller-tool append while preserving extension and team servers. `mcp_servers_test.exs:175-200` covers a persisted legacy row and a non-vacuous team-channel case, and `conversation_server_acp_test.exs:2578-2599` verifies that `session/new` receives no `fountain-caller` server. I traced `for_session/3`, extension/team composition, connection egress, callback-token handling, and the retained bridge internals; the prior delayed parked-call failure is no longer reachable through the advertised session configuration.

2. **Resolved — the architecture record now matches this partial implementation.** `decisions/0035-openai-compatible-endpoint.md:6-20` is superseded by 0057; `decisions/0057-retire-public-compatibility-protocols.md:14-36` records exactly which route/controller and caller-advertisement work is built and which bridge/flag/manual work remains; its cutover section records the observed 404/401/406 behavior. `decisions/index.md` reflects both states, and the current `okf validate` check passes.

3. **Still blocking — the public manual, runnable examples, and operator flag guidance advertise endpoints this PR removes.** `docs/integrations/openai-compatible.md:3-19,31-38,52-90` says Fountain serves `/v1/chat/completions` and `/v1/models`, gives working SDK/curl setup, and links `examples/openai-chat`; `docs/integrations/openbot.md:9-23,45-56` says `/api/agui/:agent_id` is live and tells operators to test it. `docs/integrations/clients.md:53-74`, `gateways.md:1-25`, `langchain.md:10-31`, `docs/reference/feature-status.md:10-13`, all three protocol examples, `.env.example`, `.env.compose.example`, and `docker-compose.yml` repeat those live-product claims. Trigger: merge this route deletion and have a user follow the shipped manual/example or enable `openai_compat`. Consequence: the documented integrations fail with this PR's 404/401/406 retirement responses, while the release still tells users they work. This is not merely future cleanup: approved-base `contributing/protocol-retirement.md:120-140` requires the source-removal PR to carry the public migration guide and specifically forbids leaving shipped examples calling removed endpoints. Minimal remedy: keep the established manual URLs as truthful retirement/migration pages, reconcile the clients/gateway/LangChain/feature-status cross-links, port or remove the three broken examples, remove obsolete `openai_compat` operator guidance, and run the focused docs checks.

Coverage: I inspected all 21 changed paths: `apps/fountain/lib/fountain/conversations/mcp_servers.ex`, `prompt_input.ex`, both deleted protocol controllers, and `router.ex`; `conversation_server_acp_test.exs`, `mcp_servers_test.exs`, both deleted controller suites, `saved_allowance_channel_test.exs`, and `protocol_retirement_test.exs`; the changelog fragment; ADRs 0035/0057 and `decisions/index.md`; `sdk/contract/contract.json`, `sdk/contract/omissions.json`, and the TypeScript changelog, package version, generated OpenAPI types, and user-agent version. I also traced retained caller-tool state/controller/launch paths, native session MCP assembly, extension/team/connection ordering, tenant/auth routing, public docs/examples/config, SDK exports/versioning, the approved-base policy/instructions, relevant ADRs, and the retirement inventory.

Evidence: every current substantive GitHub check for this exact head is green, including all six server partitions, static analysis, release/API contract, TypeScript/Python/Elixir/Swift SDK jobs, CLI/plugins, core distribution and Compose, secret scanning, coverage, the SDK release gate, Decisions, and `CI required`; the Docs job is skipped. Locally, `git diff --check` passed. A semantic contract comparison found exactly the four intended operations and five exclusive schemas removed, with no added or changed surviving operation/schema. I did not rerun server tests in the shared checkout because it is on a different commit; the author's reported `mix precommit`, full core, and docs runs are supporting evidence, not substituted for the current-head CI above. Green CI establishes the code/contract mechanics but does not repair or validate the omitted published manual and examples.

BinaryBourbon:
Third blocker fixed in `6a1182d0`. You were right and I checked the citation rather than taking it: `contributing/protocol-retirement.md` says the source-removal PR carries **"the public migration guide"**, and the *Published surfaces to reconcile* list above it names exactly the pages, examples and env/Compose comments you flagged. Splitting that into a later stage was my staging error, not something the approved plan allowed — a release whose own manual tells people to call routes it deletes is the defect, regardless of what a follow-up PR intends to do.

**Folded into this PR:**

- **Four migration pages at their existing URLs** — `openai-compatible.md`, `openbot.md`, `langchain.md`, `gateways.md`. The inventory asks for migration pages where a page has inbound links, so the URLs stay and the content changes: what a call gets today, what to use instead, and **what has no replacement**. That last section is the honest one — these dialects existed so a client that already spoke someone else's protocol needed no code at all, and nothing native returns that. The AG-UI page documents the **406** specifically, because `Accept: text/event-stream` is what an AG-UI client sends, so it is the most likely thing a stale integration actually meets.
- **Cross-links**: `clients.md` (the roster rows and the prose), `feature-status.md`, `nav.yml`, docs index, build guide, glossary, troubleshooting.
- **The three examples removed**, with the loss disclosed rather than a migration implied. `examples_test.go`'s `notOurs` skip list named two litellm files, and it fails on an entry matching no file — which is how it caught them — so the list empties rather than disappears, since it is the documented door for third-party YAML. `contributing/conversation-callers.md` held the deepagents example pending *"protocol retirement #2252 provides its migration"*; it now records that there is none, and why.
- **Operator flag guidance**: `.env.example`, `.env.compose.example`, `docker-compose.yml`, `docs/configuration.md`, `feature-status.md`.

One deliberate narrowing there. The flag still **exists** at this merge — deleting it is #2282 — so the wording says only what is true here: the endpoints are gone, so there is nothing for `openai_compat` to switch on. I did not write "the flag is gone", which would have been the same class of false-at-merge claim you flagged in the ADRs.

**Why #2280 and #2282 can still follow.** Both are invisible to a user at this merge: nothing advertises the caller tools any more (your blocker 1), and nothing documented promises the flag does anything (this one). Neither leaves a claim in the released product that this PR contradicts, which was the actual defect in all three findings. If you would rather see the whole thing as one PR, say so and I will collapse it.

**Validation:** `scripts/test-docs.sh` **5 suites, 0 failures** — that gate checks every page is in `nav.yml`, every page on disk is in the nav, and every internal link and anchor resolves, so the migration pages and the removed cross-links are covered. `go test ./internal/manifest/ -count=1` green. `mix precommit` PASSED (9 stages). `changelog.py check`, `conflict-markers.py`, `git diff --check`, `okf validate decisions` (`valid: true`, six pre-existing warnings) all pass.

Also worth flagging since you noted the shared checkout twice: that was real, and one consequence reached my own work — an earlier `git add -A` swept 1,308 untracked scratch files (`work/`, `scripts/destink/`, `node_modules` included) into a commit on #2280. Stripped from all four branches; `git ls-tree -r --name-only` is clean on each head now.

jhgaylor (changes requested):
Adversarial whole-PR re-review of head `6a1182d0f6b46339c8d546f4ab9bdb9df89fa9e8` against recorded base `978efe75a31eb45b8e735db6efb3d7022fadf9b2`.

Verdict: **CHANGES_REQUESTED** for two medium-severity defects.

1. **Accepted ADR 0057 contradicts the tree it claims to describe.** `decisions/0057-retire-public-compatibility-protocols.md:26-30` still lists manual/example reconciliation as “not yet built” and promises that this section always describes the merged tree, but this head adds the migration pages and deletes all three examples. More materially, `:79-93` says `/api/mcp/caller/*` now falls through to extension dispatch with 404/401/406 behavior and that `protocol_retirement_test.exs` verifies both `/api` paths. The route is still live at `apps/fountain/lib/fountain_web/router.ex:530-533`, and `caller_mcp_controller_test.exs:43-52` proves an authenticated `tools/list` still returns 200 for a legacy row. Trigger: consult the accepted ADR to operate or finish the stack, or call that retained endpoint. Consequence: the source-of-truth architecture record misstates both completed work and a live authenticated callback surface. Minimal remedy: move the manual/examples into “built so far”; mark the caller-route fallthrough as future until #2280 removes it (or fold that removal in here); and remove the false present-tense test-coverage claim.

2. **The public migration is still incomplete and can discard existing thread continuity.** `docs/build/pieces.md:48-50` says in present tense that AG-UI binds a coworker channel, while `apps/fountain/lib/fountain_web/controllers/llms_controller.ex:105-110` publishes “editors over ACP, AG-UI” through `/llms.txt`; `docs/index.md:136` also says Feature Status lists two gated features when the edited page lists one. The new migration pages then tell users to create a native conversation (`openai-compatible.md:26-38`, `openbot.md:30-40`) but omit the approved-base inventory’s continuity path (`contributing/protocol-retirement.md:60-66`): find the tenant-scoped `openai:<key>` or `agui:<threadId>` conversation and prompt its existing ID while preserving agent/environment/vault identity. Trigger: an existing integration follows the migration page. Consequence: it can open a fresh sandbox instead of continuing the existing conversation, losing the working thread’s state, while other published surfaces still advertise the removed protocol. Minimal remedy: remove or historicalize the remaining live AG-UI claims, correct the feature count, and add the explicit legacy-thread continuation recipe.

Prior blocker disposition:

- **Resolved:** persisted caller tools are no longer advertised. The legacy-row, non-vacuous team, and ACP `session/new` regressions cover that boundary.
- **Still blocking:** the ADR and public migration are substantially improved but remain inaccurate as above.

Nonblocking cleanup found during the whole-tree sweep: `apps/fountain/lib/fountain/agents.ex:24-27` still explains `get_agent/2` through the removed AG-UI route; `sdk/contract/README.md:145-147` still counts removed `ChatCompletionRequest.stream`; `sdk/typescript/package-lock.json:2-8` remains 5.2.1 while the package is 6.0.0; and `scripts/test-timings.tsv:191,229` retains both deleted controller suites.

Coverage and evidence: all 50 changed paths and the retained caller/session/auth/extension paths were inspected under independent correctness, security, API, and maintainability lenses. No executable tenant-isolation or credential-boundary blocker was found. The semantic contract comparison removes exactly the intended four operations and five schemas, with no surviving operation/schema changes; `git diff --check` passed. All substantive GitHub checks for this exact head are green, including six server partitions, static analysis, release/API contract, all SDKs, CLI/plugins, Compose/core distribution, coverage, decisions, secrets, and `CI required`; the standalone Docs job is skipped. On an isolated exact-head archive, 98 docs/manual tests and 32 focused retirement/MCP tests passed with 0 failures.

jhgaylor:
Both blockers fixed in `2c8f7e55`, and all four cleanups with them.

**1. The ADR described a tree that was not this one.** Correct on both counts, and the second was the worse of the two.

- The manual and examples are now in **Built so far**, since this head carries them.
- The response table claimed `/api/mcp/caller/*` falls through to extension dispatch, and that `protocol_retirement_test.exs` covers both `/api` paths. Neither was true here — the route is declared, and `caller_mcp_controller_test.exs` proves an authenticated `tools/list` still answers 200. The table now carries `/api/agui/*` only, and a paragraph says plainly that the caller route is **a live authenticated callback surface with no remaining caller** until #2280 deletes it, and why nothing can reach it from a retired dialect meanwhile.

I did not fold #2280 in. Instead the ADR's status section now genuinely moves with the stack: #2280 adds the caller route to the table and drops that paragraph, #2282 records the flag. Each stage's own commit does it, so the section is true at every merge rather than only at the end. That is the promise the section makes, and it was not keeping it.

**2. The migration could cost people their sandbox.** This is the one I am glad you caught — it is the only defect in this round that reaches a user. Both pages said "create a native conversation" and said nothing about the one they already have, so a client following the page opens a fresh sandbox and loses the thread it was trying to continue.

Both pages now lead with the continuity recipe from `contributing/protocol-retirement.md`, **before** the what-to-use-instead table: filter `GET /api/conversations?channel_id=openai:<key>` (or `agui:<threadId>`), then `POST /api/conversations/<id>/prompts`. With the two ways to get it wrong — replaying the host transcript, which feeds the agent its own words back, and resuming by channel with an incomplete `agent_id`/`environment_id`/`vault_id`, which resolves to a different machine and loses the sandbox by another route.

The three remaining live claims are gone: `docs/build/pieces.md` (present-tense AG-UI channel binding), `llms_controller.ex` (`/llms.txt` publishing "editors over ACP, AG-UI"), and the docs index's feature count.

**Cleanups**, all four: `agents.ex` explained `get_agent/2`'s malformed-id promise through the retired route — reworded rather than dropped, because the promise belongs to every caller handing it a raw path segment, not to that one route; `sdk/contract/README.md` counted `ChatCompletionRequest.stream` among four default-carrying properties, now three; `test-timings.tsv` lost this change's two rows.

`package-lock.json` deserves a note — it was **not** merely cosmetic. It sat at 5.2.1 against a 6.0.0 `package.json`, which `npm ci` rejects. Nothing in this repo's CI runs `npm ci` against the SDK, so it would have surfaced in a consumer's install rather than here. Fixed and verified with `npm ci --dry-run`.

**Validation:** `mix precommit` PASSED (9 stages); full suite **5,915 core + 136/30/33/60/29 extensions, 0 failures**; `scripts/test-docs.sh` 5 suites 0 failures; `okf validate decisions` valid with the six pre-existing warnings; SDK release gate still well-formed for 6.0.0.

jhgaylor (changes requested):
Adversarial re-review verdict: **CHANGES_REQUESTED**

Reviewed the full 56-path change at head `2c8f7e55dce8b0a2bc759c45cd1ba6b6657663af` against base `978efe75a31eb45b8e735db6efb3d7022fadf9b2` through four independent lenses (correctness, tenant/security boundaries, API/SDK compatibility, and maintainability/verification).

## Blocking findings

### 1. A sprite callback key can act on another conversation in the same tenant (medium)

`POST /api/mcp/caller/:conversation_id` is retained in `router.ex:530-533`, but `CallerMcpController.build_ctx/2` checks only that the path conversation belongs to the authenticated account and has caller tools (`caller_mcp_controller.ex:30-44`). It does not bind a `sprite`-scoped key to that conversation's `callback_api_key_id`.

This matters because `TenantAPIAuth` authenticates the callback token as the account while exposing the presented key as `current_api_key`; `SandboxKey` explicitly documents that same-tenant checks alone are insufficient for cross-conversation writes. A compromised sandbox A can name a live legacy conversation B in the same account, list B's request-defined tool schemas, and call a tool against B. That reaches `ConversationServer.park_caller_tool/4`, mutating B's pending caller-tool state and emitting stages/webhooks until the call expires.

Prefer deleting the caller MCP route/controller in this retirement change. If it must remain staged for #2280, require `B.callback_api_key_id == current_api_key.id` for sprite keys (while retaining intended owner/full-key behavior) and add a two-conversation, same-tenant denial test.

### 2. The documented continuity lookup can select the wrong or dead conversation (medium)

`docs/integrations/openai-compatible.md:32-44` and `docs/integrations/openbot.md:36-47` query only `channel_id` and then tell the reader to prompt "the id" returned. That result is not singular: `list_conversations/2` applies only supplied filters, includes terminal rows, and orders by `updated_at`; the controller tests already demonstrate that a channel-only query returns idle and terminated conversations together. The retired launch binding also included agent, vault/environment identity, and live conversation/sandbox state.

With a reused legacy thread key, the recipe can therefore prompt another agent's workspace or a terminal conversation, defeating the continuity guidance. Raw URL interpolation also mishandles reserved characters in thread keys.

Use encoded query parameters, narrow by the known agent and promptable state, and require the operator to verify agent/environment/vault plus live sandbox state before prompting by conversation ID. Document multiple-match and no-live-match cases.

### 3. The public webhook catalog still advertises the retired request-tool loop (medium)

`docs/reference/webhooks.md:72` still presents `caller_tool.started/done` as a current event produced when an agent calls a tool supplied on the client's request. This PR removes both clients capable of answering that loop and removes caller-tool advertisement; only transitional bridge/event-vocabulary internals remain. The trusted-base retirement inventory explicitly names the webhook catalog as a published surface to reconcile.

An operator can still configure `conversation.caller_tool.*` from the manual and then silently receive no events on supported runs. Mark the entry historical/retired and explain that vocabulary/filter compatibility remains temporarily for legacy rows and #2280 cleanup, or remove it atomically with the remaining bridge.

## Verification and prior-finding disposition

The previous ADR/live-claim, caller-tool advertisement, package-lock, contract-count, timing, feature-count, and stale-comment blockers are resolved. The continuity section was added but remains ambiguous as described above.

All substantive exact-head GitHub checks are green (six server partitions, static analysis, API contract/release, SDKs, CLI/plugins, Compose/core, coverage, decisions, secrets, and `CI required`; standalone Docs is skipped). Independent local checks covered the exact tree: `git diff --check`, contract normalization (exactly four operations and five exclusive schemas removed, no surviving operation/schema changed), 98 docs/manual tests, 32 focused protocol-retirement/MCP tests, and the TypeScript install/contract suite.

jhgaylor:
All three fixed, in `cf16e3cb` and `513edd93`. The first one I verified in the code before acting on it, and you're right.

### 1. The sprite-key cross-conversation gap — route deleted here

Confirmed: `build_ctx/2` does `get_conversation(conv_id, user.id)` and nothing else. `Conversations.sandbox_owns?/2` exists for exactly this rule and the labels path uses it; this route never did. A compromised sandbox could name any live legacy row in the account and park calls on its turn.

One correction to the framing, not to the finding: this **predates the retirement** — the route always scoped that way, and the dialects were never required to reach it. So it isn't a regression this PR introduces. That doesn't make it less worth fixing, and it doesn't change the remedy.

I took your preferred option and deleted the route, controller and suite here rather than adding a binding check to code scheduled for deletion. The route was already dead weight in this PR — nothing advertises those tools any more. **All five retired method/path combinations are now unrouted in one change**, which is also what the ADR wanted to be able to claim. Verified by reverting: re-adding a `/api/mcp/caller` route turns five tests red across the retirement and spec suites.

### 2. The continuity lookup

Also right, and this is the second time this recipe has been wrong — it was worth the second look. `list_conversations/2` applies only the filters given, orders by `updated_at`, and returns `failed` and `terminated` rows that cannot take a prompt. Both pages now:

- filter by `agent_id` and `status=idle,running,pending` as well as `channel_id`;
- pass the thread key through `--data-urlencode`, so `/`, `?`, `&` or a space in a key doesn't break the query;
- say explicitly that the result is a **list, not an answer**, and that the reader must confirm `agent_id` / `environment_id` / `vault_id` before prompting;
- cover both the several-matches case (thread key reused across configurations — pick by identity, not recency) and the no-match case (nothing to continue; a fresh conversation is correct).

### 3. The webhook catalogue

Moved here from #2280, since nothing can publish a `caller_tool` stage once this change lands: the `@retired` vocabulary and the docs page both come forward. The two types stay valid as filters so a stale endpoint remains editable, and the page says why.

That turned the catalogue guard red, because it reads `publish_stage/4` call sites out of source and `Pending` still holds them. Rather than weaken it, it now knows about `@retired` in both directions, plus a new test pinning the transitional state itself — call sites present, stage out of the catalogue, stage in `@retired`.

Worth flagging: the first version of that test's failure message told the next person to drop `caller_tool` from `@retired` once the call sites go. That is exactly backwards — the vocabulary has to **outlive** the call sites, because what it buys is a stale endpoint staying editable, and it leaves at the rollback floor (#2273). Corrected in a follow-up commit, and #2280 deletes the test when its call sites go while keeping `@retired`.

### Stage consistency

Since the ADR status section promises to describe the tree it merges into, I checked it mechanically at all four heads rather than trusting it:

| stage | caller route | `CallerTools` | flag | ADR: 5 unrouted | ADR: "Not yet built" |
|---|---|---|---|---|---|
| #2279 | gone | present | present | yes | yes |
| #2280 | gone | gone | present | yes | yes |
| #2282 | gone | gone | gone | yes | no |
| #2283 | gone | gone | gone | yes | no |

### Validation

`mix precommit` PASSED (9 stages). Full suite **5,918 core + 136/30/33/60/29 extensions, 0 failures**. `scripts/test-docs.sh` 5 suites 0 failures. `sdk-contract --check` ok (173 operations, 198 schemas). `okf validate decisions` valid with the six pre-existing warnings. Changelog, conflict-marker and whitespace checks pass.

jhgaylor (dismissed):
Adversarial re-review verdict: **CHANGES_REQUESTED**

Reviewed all 63 changed paths at head `1f1f0ccc78e80bd422277edfa36a709413a5e2ef` against base `978efe75a31eb45b8e735db6efb3d7022fadf9b2` through independent correctness, tenant/security, API/SDK, and maintainability/verification lenses.

## Blocking findings

### 1. Retained MCP callbacks let one sprite assume another conversation's authority (high)

The deleted caller route is fixed, but every remaining sandbox-facing MCP callback repeats its authorization flaw. A `sprite` key authenticates as the account, while `callback_api_key_id` and `SandboxKey.id/1` provide the exact conversation binding. The controllers do not compare them:

- `TeamMcpController.build_ctx/3` (`team_mcp_controller.ex:29-39`) accepts any same-tenant team conversation and derives `ctx.self` from the conversation named in the path.
- `FountainGoogle.McpController.build_ctx/3` (`mcp_controller.ex:39-53`) accepts any same-tenant conversation whose agent names the connection, then resolves the server-held OAuth token for mailbox read/send/modify.
- `FountainBuzz.McpController.build_ctx/2` (`mcp_controller.ex:35-46`) accepts any same-tenant Buzz conversation, then resolves and decrypts that identity's vault material for signed publishing.

Trigger: sandbox A uses its valid callback token with conversation B's ID. The controller accepts B and grants B's configured tool identity. This enables Gmail reads and external sends, publishing as another Buzz identity, or impersonating teammate B in owner-delegated team messages. Cross-tenant IDs remain hidden, but the per-sandbox tool and credential boundary is bypassed.

Centralize a callback authorization check before resolving agent config, connections, vaults, or MCP context. For a sprite key, require the presented key ID to equal the target conversation's non-null `callback_api_key_id`; return a neutral 404 on mismatch. Preserve full-scope owner-key behavior explicitly if intended. Add A-token/B-conversation denial tests and exact-key positive tests for Team, Google, and Buzz, asserting no side effect or audit on refusal.

### 2. The continuity recipe still cannot verify sandbox liveness (medium)

`docs/integrations/openai-compatible.md:51` and `openbot.md:54` tell readers to verify that a candidate's sandbox still exists before prompting it, but their `GET /api/conversations` query cannot supply that evidence. `list_conversations/2` does not preload `:sandbox`, so `ConversationJSON` renders `sandbox: null` for list rows. An `idle` conversation can retain a sandbox row marked `terminated` or `failed`; prompting it makes `Wake` provision a fresh machine, contradicting the recipe's “Same sandbox, same history” promise and losing workspace continuity.

For each candidate, document `GET /api/conversations/<id>` (whose show path loads the sandbox), require a non-null/non-terminal `data.sandbox`, and verify its agent/environment/vault identity before prompting. Cover duplicate channel candidates where an idle conversation's sandbox is terminal.

### 3. Brand-new webhook endpoints may subscribe only to retired, never-emitted events (medium)

`Events.valid_filter?/1` (`events.ex:114-120`) accepts retired exact and wildcard filters globally, and `Endpoint.validate_event_types/1` uses that predicate for both creation and update. The new regression at `webhooks_test.exs:455-465` therefore locks in successfully creating an endpoint whose sole filter is `conversation.caller_tool.started` even though the event is explicitly never emitted.

That preserves stale rows but also creates new valid-looking endpoints that silently receive nothing, contradicting the catalog's save-time guarantee. Grandfather retired values only when they were already present and unchanged on an existing endpoint; reject them on create and when newly added during update. Test both cases.

## Verification and prior-finding disposition

The earlier caller-route vulnerability, ambiguous channel-only lookup, live webhook-catalog claim, ADR state, package lock, and public-cutover blockers are otherwise repaired. All five retired path/method combinations are now unrouted, and the wire contract still removes exactly four operations and five exclusive schemas without changing surviving operations or schemas.

All substantive exact-head GitHub checks are green: six server partitions, static analysis, API contract/release, SDKs, CLI/plugins, Compose/core, coverage, decisions, secrets, and `CI required`; the standalone Docs job is skipped. Local exact-head evidence: `git diff --check`, Go manifest tests, TypeScript typecheck, 126 SDK tests, and the TypeScript contract verifier passed. The local Mix/docs attempt was blocked before app tests by the Lumis precompiled-NIF environment, so I am not claiming that run passed.

Nonblocking cleanup: `scripts/test-timings.tsv:197` still names the newly deleted caller-controller suite, and retained callback/extension comments still describe `/api/mcp/caller` as live.

jhgaylor (changes requested):
Adversarial re-review verdict: **CHANGES_REQUESTED**

Reviewed all 63 changed paths at head `1f1f0ccc78e80bd422277edfa36a709413a5e2ef` against base `978efe75a31eb45b8e735db6efb3d7022fadf9b2` through independent correctness, tenant/security, API/SDK, and maintainability/verification lenses.

## Blocking findings

### 1. The continuity recipe still cannot verify sandbox liveness (medium)

`docs/integrations/openai-compatible.md:51` and `openbot.md:54` tell readers to verify that a candidate's sandbox still exists before prompting it, but their `GET /api/conversations` query cannot supply that evidence. `list_conversations/2` does not preload `:sandbox`, so `ConversationJSON` renders `sandbox: null` for list rows. An `idle` conversation can retain a sandbox row marked `terminated` or `failed`; prompting it makes `Wake` provision a fresh machine, contradicting the recipe's “Same sandbox, same history” promise and losing workspace continuity.

For each candidate, document `GET /api/conversations/<id>` (whose show path loads the sandbox), require a non-null/non-terminal `data.sandbox`, and verify its agent/environment/vault identity before prompting. Cover duplicate channel candidates where an idle conversation's sandbox is terminal.

### 2. Brand-new webhook endpoints may subscribe only to retired, never-emitted events (medium)

`Events.valid_filter?/1` (`events.ex:114-120`) accepts retired exact and wildcard filters globally, and `Endpoint.validate_event_types/1` uses that predicate for both creation and update. The new regression at `webhooks_test.exs:455-465` therefore locks in successfully creating an endpoint whose sole filter is `conversation.caller_tool.started` even though the event is explicitly never emitted.

That preserves stale rows but also creates new valid-looking endpoints that silently receive nothing, contradicting the catalog's save-time guarantee. Grandfather retired values only when they were already present and unchanged on an existing endpoint; reject them on create and when newly added during update. Test both cases.

## Verification and prior-finding disposition

The earlier caller-route vulnerability, ambiguous channel-only lookup, live webhook-catalog claim, ADR state, package lock, and public-cutover blockers are otherwise repaired. All five retired path/method combinations are now unrouted, and the wire contract still removes exactly four operations and five exclusive schemas without changing surviving operations or schemas.

Security note: the retained Team/Google/Buzz callbacks are tenant-scoped rather than callback-key-bound. Base ADR 0045 explicitly accepts that account-wide authority for `owner`-mode sprite keys and identifies per-conversation callback authorization as an unchosen alternative, so it is not a defect in this PR.

All substantive exact-head GitHub checks are green: six server partitions, static analysis, API contract/release, SDKs, CLI/plugins, Compose/core, coverage, decisions, secrets, and `CI required`; the standalone Docs job is skipped. Local exact-head evidence: `git diff --check`, Go manifest tests, TypeScript typecheck, 126 SDK tests, and the TypeScript contract verifier passed. The local Mix/docs attempt was blocked before app tests by the Lumis precompiled-NIF environment, so I am not claiming that run passed.

Nonblocking cleanup: `scripts/test-timings.tsv:197` still names the newly deleted caller-controller suite, and retained callback/extension comments still describe `/api/mcp/caller` as live.

BinaryBourbon:
Both fixed in `745f8c81`. The first is a fix of mine that over-corrected, which is worth saying plainly.

### 1. A new endpoint could subscribe to a retired event

Correct, and my regression locked it in. Making `valid_filter?/1` accept retired types *globally* kept stale rows editable — the point — but it also let anyone create an endpoint whose only filter is `conversation.caller_tool.started`, which saves cleanly and then receives nothing. That is exactly what save-time validation exists to prevent, so I traded one silent-nothing failure for another.

The grandfathering belongs to **the row**, not the vocabulary:

- `valid_filter?/1` goes back to meaning "in the catalogue".
- A separate `retired_filter?/1` recognises a retired value, and its docstring says it is not a licence to subscribe.
- `validate_event_types/1` accepts a retired value only when it is already in `changeset.data.event_types`. On a new record that is the `[]` default, so nothing is grandfathered; on an update a newly added retired type is not in the stored list and is refused like any other unknown.

Four tests cover both halves — existing row stays editable and unmodified, create refused, *newly added* retired type on an existing row refused, and the kept value never delivered. **Verified by reverting**: widening the check back to `valid_filter?(type) or retired_filter?(type)` turns two of them red.

The setup writes the retired type onto the row with a raw `Ecto.Changeset.change/2`, bypassing today's validation, because that is the only way to reach the state a pre-retirement release left behind — which is the state being grandfathered.

### 2. The continuity recipe asked for evidence the query cannot give

Also right, and I checked it before changing anything: `list_conversations/2` preloads `[:agent, :agent_version, turns: first_turn_query()]` and **not** `:sandbox`, while `get_conversation/2` preloads `[:sandbox, :agent, :vault, :agent_version]`. So every list row renders `sandbox: null` and my "verify the sandbox" instruction was unfollowable.

Both pages now add a per-candidate `GET /api/conversations/<id>` step, with the comment saying why the list cannot answer it, and require before prompting:

- `data.sandbox` non-null and its `status` not `terminated` or `failed` — an `idle` conversation can still point at a dead machine, and prompting that one provisions a fresh sandbox, the outcome the whole section exists to avoid;
- `data.agent_id` / `environment_id` / `vault_id` matching the identity the work was done under.

The duplicate-candidate case now says to expect some of them to have terminal sandboxes, and that if none survives both checks a fresh conversation is the right answer.

### Non-blocking cleanups

Both done: `test-timings.tsv` loses the deleted caller-controller suite, and the Buzz and Google extension moduledocs stop listing `/api/mcp/caller/:id` among the core routes their mounts must not shadow.

Two references deliberately kept: `caller_tools_test.exs` still names the path because the module it tests still exists in this change (#2280 deletes both), and the applied `20260826041202` migration is never edited.

### Validation

`mix precommit` PASSED (9 stages). Full suite **5,920 core + 136/30/33/60/29 extensions, 0 failures**. `scripts/test-docs.sh` 5 suites 0 failures. Changelog, conflict-marker and whitespace checks pass.

Noted on your local run being blocked by the Lumis precompiled-NIF environment — that is why I run against a private `DATABASE_URL`; the shared checkout also accumulates committed audit rows that fail `Fountain.AuditTest` (#2178) and leftover live sandbox rows that produce `:fleet_full`.

jhgaylor (changes requested):
## Verdict: changes requested

Reviewed all 67 changed paths in the exact range `978efe75a31eb45b8e735db6efb3d7022fadf9b2..745f8c8150c10cf59662a34e41795d118a1de1d2` through four independent adversarial lenses: correctness, tenant/security boundaries, API/SDK compatibility, and maintainability/verification.

### 1. Medium — the continuity recipe still cannot establish that the original sandbox exists

`docs/integrations/openai-compatible.md:51-77` and `docs/integrations/openbot.md:54-80` now fetch the show endpoint and treat a non-null, nonterminal `data.sandbox` database row as sufficient to prompt, ending with “Same sandbox, same history.” That endpoint only preloads stored state; it does not contact the provider.

Actual liveness is first tested inside the mutating prompt path: `Wake.maybe_reuse_sandbox/1` and `probe_sandbox/4` (`apps/fountain/lib/fountain/conversations/wake.ex:43-102`) probe a `ready`/`suspended` machine, and provider `:not_found` becomes `:create_new`. A stale row therefore passes the documented checklist, then the prompt provisions a fresh machine and loses the disk/runtime context the guide promises to preserve.

Either expose and use a non-mutating provider-liveness/refuse-replacement path, or describe continuation as best-effort and explicitly warn that prompting may reprovision. The database show response cannot support the current guarantee.

### 2. Medium — the new Connections guidance collapses two independent gates

`docs/configuration.md:419-424` says Connections is turned on by `BROKER_LISTEN_PORT`, and `.env.example:189-193` says it needs no `FEATURE_FLAGS_ON` entry. But `Fountain.Connections.enabled_for?/1` requires both `Broker.configured?/0` and `FeatureFlags.enabled?(:connections, user_id)`. With `POSTHOG_PROJECT_API_KEY` configured, an absent or false PostHog answer disables Connections creation; `feature_flags_test.exs` pins that behavior. `docs/reference/feature-status.md` and ADR 0019 also correctly describe the flag and broker as separate gates.

A self-hoster following the changed configuration guidance can configure the broker and PostHog yet leave every Connections creation route disabled. State the actual split: without PostHog, `connections` defaults on; with PostHog, enable it remotely or force it with `FEATURE_FLAGS_ON=connections`; the broker remains independently required.

### 3. Medium — the retired-event guard cannot detect a newly reintroduced emitter

`apps/fountain/test/fountain/webhooks/events_test.exs:65-75` exempts every source pair whose stage appears in `Events.retired/0`, while lines 90-103 merely assert that at least one `caller_tool` site remains during this transitional stage. A new `publish_stage(..., "caller_tool", ...)` site or status is therefore indistinguishable from the expected unreachable legacy sites and leaves CI green.

That can become observable delivery: `apps/fountain/lib/fountain/webhooks.ex:288-300` dispatches through `Events.matches?/2` without a catalogue check, and grandfathered stored filters remain. Pin the exact transitional retired call sites/statuses and reject additions; when stage 2 deletes them, replace that allowance with a zero-retired-publish-sites invariant.

### Prior-finding disposition

- The webhook creation defect is fixed: new/newly added retired filters are rejected while unchanged stored values remain editable.
- The prior callback-key concern is not a defect under base ADR 0045's account-authority model.
- The route/controller, contract, generated TypeScript, package metadata, example, timing, and caller-tool-advertisement repairs remain coherent.

Nonblocking: `docs/reference/webhooks.md:95-100` still says Fountain “keeps accepting” retired filters without clearly limiting that to unchanged values already stored on an existing endpoint.

All substantive exact-head GitHub checks are green, including six test partitions, static analysis, release/API contract, SDKs, CLI/plugins, Compose/core, coverage, decisions, secrets, and `CI required`; Docs was conditionally skipped. `git diff --check` passed. No local Mix result is claimed.

BinaryBourbon:
All three fixed in `efc520e2`, and the nonblocking note with them.

**1. The continuity recipe.** You were right that the show route never contacts the provider: the probe is `Wake.maybe_reuse_sandbox/1` inside the prompt, and a provider `:not_found` becomes `:create_new`. There is no non-mutating liveness path for a `suspended` sandbox at all — `SandboxFilesController` refuses a parked machine with `409 sandbox_not_ready` rather than waking it for a read — so the recipe cannot promise what it promised. Both pages now say what the row rules out (a null or terminal sandbox, the wrong identity), what it cannot (a `ready`/`suspended` row the provider has since lost), and how much a caller can know beforehand: for `ready`, `GET /api/sandboxes/:id/git-status` is a read that proves the machine answers (`200`) or did not (`503 sandbox_unreachable`); for `suspended`, the wake the prompt performs is the probe. Continuation is stated as best-effort, with the two after-the-fact tells: `data.sandbox_id` changed on the show route, or a `session` stage event with `data.reason: fresh_sandbox` in the event feed. I first wrote that as a webhook check and then read `Webhooks.payload/3`: it carries no stage data, so it is the feed, not the webhook.

**2. The Connections gates.** `Connections.enabled_for?/1` is `Broker.configured?() and FeatureFlags.enabled?(:connections, user_id)`, as you said. The table row and `.env.example` now state the split — without PostHog the flag reads on by itself; with PostHog, enable it there or force `FEATURE_FLAGS_ON=connections`; the broker is required either way — and the hosted-rollout paragraph stops calling the broker "the switch that decides which accounts", since the broker is deployment-wide and the flag is what selects accounts. `docs/reference/feature-status.md` and ADR 0019 were already right and are unchanged.

**3. The retired-event guard.** `events_test.exs` now pins the two `Pending` sites by file, stage, status and count in `@retired_call_sites`, exempts only those from the catalogue assertion, and asserts the found set equals the pin. Verified by breaking it three ways — a `caller_tool` publish in another file, a new status in `pending.ex`, and a second copy of a pinned site — and each turns it red (the first two fail both tests, the duplicate fails the pin). When stage 2 deletes the sites the map goes empty and the same assertion becomes "no retired stage is published from anywhere", which is the invariant you asked for.

**Nonblocking.** `webhooks.md` now limits the allowance to a value already stored on an existing endpoint and says a new endpoint or an added filter fails with `unknown event`.

Local: `events_test`, `webhooks_test`, `config_reference_test`, `docs_test` and `scripts/test-docs.sh` green on a private test database; `mix format --check-formatted` clean; conflict-markers and changelog check clean.

jhgaylor (changes requested):
## Verdict: changes requested

Fresh adversarial review of all 67 changed paths in exact range `978efe75a31eb45b8e735db6efb3d7022fadf9b2..efc520e2f877a55c151ae7e27361eb2f2e91a7ef` through correctness, tenant/security, API/SDK, and maintainability/verification lenses.

### Medium — the “exact” retired-emitter guard still ignores production publisher files

`apps/fountain/test/fountain/webhooks/events_test.exs:21-35` hard-codes 12 source files, and `published_sites/0` at lines 46-51 reads only those. The current tree has production `publish_stage` callers outside that list: `home_checkpoint.ex:114`, `machine_events.ex:45,85`, `provision_watchdog.ex:79`, `turn_launch.ex:57`, and `wake.ex:601,631`.

Adding `Output.publish_stage(..., "caller_tool", "started", ...)` to any of those already-existing publisher modules leaves both the catalogue check (lines 83-108) and the new retired-site frequency assertion (lines 110-137) green because the file is never scanned. That contradicts the changed test's claim at lines 60-69 that an emitter in “another file” is detected and that an empty future map proves no retired publisher exists anywhere.

This is externally observable, not just test style: `Fountain.Webhooks.do_dispatch/2` (`apps/fountain/lib/fountain/webhooks.ex:288-300`) builds the event type and applies `Events.matches?/2` without a catalogue check, while grandfathered exact filters and `*` can still match. A supposedly retired event introduced in an omitted module would therefore be delivered while CI remained green.

Please discover all production `apps/fountain/lib/**/*.ex` publisher files (or independently assert that the maintained inventory equals every file containing a publisher), then exempt only the two intended `Pending` sites by file/stage/status/count. Add a regression showing a publisher in a newly discovered/currently omitted module fails the guard.

### Prior findings

The sandbox-continuity guidance, two independent Connections gates, retired-filter create/update behavior, callback route removal, route/contract/generated-TypeScript deletion, and public migration material are otherwise resolved and coherent. No other medium-or-higher defect was found.

All substantive exact-head checks are green, including `CI required`, six test partitions, coverage, static analysis, release/API contract, SDK matrices, CLI/plugins, Compose/core distribution, decisions, and secret scanning. Exact-range `git diff --check` passed; no local Mix result is claimed.

BinaryBourbon:
Fixed in `47b1ab03`. You were right, and the claim in the pin's own comment was false on the tree it shipped in: the list named twelve files while `home_checkpoint.ex`, `machine_events.ex`, `provision_watchdog.ex`, `turn_launch.ex` and `wake.ex` published outside it.

**What changed.** The list is gone. `publisher_files/1` walks `lib/**/*.ex` for every file that calls `publish_stage(` with a literal stage, and both directions of the catalogue test and the retired-site pin read from that walk. A second, wider pattern reads every call that names its stage whatever it does with the status, and the pin has to agree with that view as well, so `publish_stage(id, "caller_tool", status)` cannot hide behind a computed status. "The walk reaches the publishers" names the six files the list missed and fails if the walk stops finding any of them. Only the two `Pending` sites remain exempt, by file, stage, status and count.

**The regression you asked for.** A test writes one module the inventory never heard of into a scratch tree, publishing a retired stage once with a literal status and once with a computed one, and asserts the walk reports both and that the result fails the pin. I also broke it on the real tree three ways, each red: a literal `caller_tool` publish appended to `wake.ex` (catalogue and pin both fail), a computed-status one in `machine_events.ex` (the wider-view assertion fails), and a brand-new module (both fail). All restored; `git status` clean apart from the test.

**One thing the walk surfaced that this change does not fix.** `home_checkpoint.ex` has published a `checkpoint` stage with a computed status since #1079, and that stage is in neither the catalogue nor the docs page, so `*` subscribers receive `conversation.checkpoint.done`/`failed` while nobody can subscribe by name. The literal-status pattern cannot read that site, and the wider view is only asserted for retired stages here. Filed as #2294 rather than folded into a retirement PR; it predates this change and is on `main` today.

Local: `events_test` (16) and `webhooks_test` green on a private database; `mix format --check-formatted` clean.

BinaryBourbon:
One more commit, `befc219a`: partition 4 on `47b1ab03` was red on `Fountain.TmpDirTest`'s guardrail, not on the walk. The new regression built its own path under `$TMPDIR`, which is the #1423 spelling that guardrail exists to refuse. It now uses `Fountain.TmpDir.mkdir!/1`, which scopes the scratch tree to the OS process and removes it at test end, so the explicit `on_exit` is gone too. `events_test` and `tmp_dir_test` green together locally.

jhgaylor (approved):
## Verdict: approved

Reviewed all 67 changed paths in `978efe75a31eb45b8e735db6efb3d7022fadf9b2..befc219a80c430f28d927529edaf66a9b17e028e` through independent correctness, tenant/security, API/SDK, and maintainability/verification reviews. No unresolved medium-or-higher findings.

The previous retired-emitter blocker is resolved. The guard recursively discovers production files under `lib/`, includes the formerly omitted publishers, and checks both literal and computed statuses against the two permitted `Pending` sites. In an isolated archive, adding a new module publishing `caller_tool.started` failed both the catalogue and retired-site assertions; using a computed status failed the wider retired-stage assertion. Removing the mutation restored all 16 tests.

The temporary-directory fixture failure on `47b1ab0` is also resolved: `befc219a` uses the existing `Fountain.TmpDir.mkdir!/1` helper and its cleanup. I ran the actual webhook catalogue and temporary-directory suites against the current commit with the pinned Elixir toolchain: **23 tests, 0 failures**. This was standalone ExUnit execution, not a full local Mix run. The diff whitespace check passed.

The prior continuity guidance, Connections gates, stored-filter grandfathering, callback-route retirement, native MCP preservation, contract/type removals, SDK 6.0.0 metadata, and public migration pages remain coherent. The existing unrelated computed `checkpoint` catalogue gap is tracked separately in #2294.

All substantive GitHub checks for this revision are successful, including the previously failing partition 4, the remaining server partitions, coverage, static analysis, release/API contract, SDKs, CLI/plugins, Compose/core, decisions, secret scanning, and `CI required`.

Issue #1202, opened 2026-08-26
Tool bridge on /v1/chat/completions: the caller's tools come back as tool_calls, Fountain's own tools run in the sandbox


jhgaylor:
## Goal

A client that sends `tools` on `POST /v1/chat/completions` gets them **back as real `tool_calls`** when the agent uses one, answers with `role: "tool"` messages on the next request, and the turn resumes. Fountain's own tools (the sandbox's shell, files, git, the MCP servers the agent was configured with) keep running inside the sandbox with no round trip and no approval, exactly as today.

This is "the other half of the integration" that `AguiController`'s moduledoc and `docs/integrations/openbot.md` deliberately left out (#643 is the approvals corner of it), and it is what makes a Fountain agent usable *inside* a framework loop rather than only as a leaf: a LangGraph `create_agent`, a Deep Agent, Open WebUI's tool servers, LiteLLM function calling, the `openai` SDK's `tool_runner` all work by round-tripping tool calls, and today ADR 0035 decision 4 refuses every one of them ("never a tool call, `tools` ignored").

## Why this is not what ADR 0035 refused

Decision 4's reasoning was: emitting the *sandbox's* tool use as a `tool_call` asks the client to run something that already ran. That still holds and stays refused. A **caller-defined** tool is the opposite case: it only exists on the client, the sandbox cannot run it, and the client is the only party that can return a result. Emitting it is the protocol working as designed. So the amended rule is: *a tool call is emitted if and only if the tool came from the request's `tools`.*

## Mechanism (sketch)

The sandbox already receives Fountain-served MCP servers per conversation — `team_mcp_servers/1`, `team_comms_mcp_servers/1`, `buzz_mcp_servers/1` in `ConversationServer`, each a `POST /mcp/<kind>/:conversation_id` authenticated by the conversation's `callback_token`. The bridge is one more of those:

1. **Register.** On a request with `tools`, the controller stores the tool list on the conversation (name, description, JSON schema; `tools` may change between requests, last write wins) and the sandbox's MCP config gains `POST /mcp/caller/:conversation_id`, which lists exactly those tools. The agent sees them beside its own; nothing tells it they are remote.
2. **Call.** When the agent invokes one, the MCP handler parks the call as a pending `caller_tool_call` (id, name, arguments) on the `ConversationServer`, and the turn's block stream carries a new block kind, `caller_tool_call` (like `permission_request` in `Blocks`, which is the existing "the turn is blocked on the outside" shape).
[...]

PR #2280, opened 2026-09-15
feat(conversations)!: retire the request-defined tool bridge (stage 2)

BinaryBourbon:
Stage 2 of the ADR 0057 retirement (#2252). **Stacked on #2279** — review that first; this PR's diff against `main` will include it until #2279 lands and this one is retargeted.

#2279 unrouted all five retired paths, the caller MCP adapter and its route included — those moved there after review, because that route scoped only to the tenant and any sandbox key in the account could drive it. **This PR is what is left once nothing can reach the bridge: the internals.**

## Gone

`Fountain.CallerTools` and its MCP layer · `set_caller_tools/3` · the `caller_tools` schema field, the write on both launch paths and its key in the queued-attrs allow list · the bridge append in `McpServers.fountain_served/2` · the four parked-call entry points on `ConversationServer` with their `handle_call` clauses, the `caller_calls` field, the deadline handler and the turn-end drain · `Pending`'s parked-call half.

Nothing publishes a `caller_tool` stage after this, so the transitional catalogue guard #2279 added goes too. `@retired` stays — it keeps a stale endpoint editable and leaves at the rollback floor (#2273), not here.

Server size pin **2364 → 2223**.

## Kept, deliberately

- **`Pending`'s permission behaviour.** It shared the module with the parked calls and none of the mechanism.
- **The `conversations.caller_tools` column**, for #2273 to drop once no old reader is left. The *schema field* goes now — that is precisely what makes this server a non-reader, which is the precondition #2273 is waiting on. The applied `20260826041202` migration is untouched.

## The regression this is actually risky for

People configure tools **on an agent** that call their own application. Those are not the retired bridge — they come off the agent row, resolve `${VAR}` against the environment and vault, and authenticate with the tenant's own credential — and the maintainer named them as the thing not to break. A new `describe` block in `mcp_servers_test.exs` pins it end to end:

- the substituted URL and headers reach `session/new`;
- again on a later turn with a rotated callback token;
- with **no** callback token at all (the Fountain-served lists are gated on that credential; the agent's own servers must not become so by accident);
- still ordered ahead of the team tools it shares a turn with.

The "again on a later turn" case compares two lookups, so two *absent* servers would have compared equal and passed vacuously — exactly the regression the test exists to catch. It asserts the shape first. (Guard-the-guard, per the suite's own convention.)

## The webhook catalogue: retired vocabulary, not a data migration

**This section replaced a migration after review.** The first version of this PR shipped a forward migration that stripped `conversation.caller_tool.*` from stored subscriptions and defaulted an emptied one to the three turn/provision events. That was wrong twice over: it handed endpoints event classes their owners never chose, and its no-op `down/0` contradicted ADR 0057's own statement that this phase stays reversible while the physical tool definitions remain.

The migration is deleted. The actual problem is narrower: `Webhooks.Endpoint.validate_event_types/1` re-validates the **whole** `event_types` array on every update, so a stored retired type would be refused the next time an owner changed an unrelated field like the URL — a row made uneditable by a removal it had no part in. That needs the vocabulary to keep accepting what it used to emit; it never needed anyone's data rewritten.

`Fountain.Webhooks.Events` grows a `@retired` list: **not** in `types/0`, **not** in the catalogue the manual renders, never published, and still accepted by `valid_filter?/1`. A filter matching no type never fires, and `matches?/2` is a string compare that never consults the module. Stored subscriptions are left exactly as their owners wrote them, and a rollback restores the emitter to find its filters intact. An entry leaves `@retired` only once the rollback floor passes the release that retired it — for `caller_tool`, #2273's gate.

`docs/reference/webhooks.md` loses the `caller_tool` row and gains a **Retired events** section explaining all of that, including that tools configured *on an agent* are a different mechanism and are not retired.

**Verified by reverting:** narrowing `valid_filter?/1` back to the catalogue alone turns four of the new tests red. Coverage at both levels — `events_test` for the vocabulary (a typo inside a retired stage is still refused; a guard that `retired/0` is not merely empty) and `webhooks_test` for the real scenario: an endpoint created carrying a retired type can still have its URL changed, and its `event_types` come back untouched.

## Validation

Run against a private `DATABASE_URL` (the shared test database still carries leftover live-sandbox and audit rows).

- **Whole suite**: `fountain` 6 doctests + **5,889 tests, 0 failures**, 2 skipped; `fountain_buzz` 136, `fountain_slack` 30, `fountain_support` 33, `fountain_google` 60, `fountain_microsoft` 29 — all 0 failures.
- `mix precommit` — **PASSED** (9 stages).
- Contract regeneration produced **no diff**: this route was never in the OpenAPI document (it is in `api_spec_test`'s unspecced allowlist, where its row is now removed). 173 operations before and after.
- `changelog.py check`, `sdk-contract --check`, `conformance/lint.py`, `conflict-markers.py`, `git diff --check` — all pass.
- Final `rg` census across `apps ee config`: the only remaining hits are the applied 20260826041202 migration, this PR's own migration, and prose. Extension router moduledocs in `fountain_google` and `fountain_buzz` that listed `/api/mcp/caller/:id` as a core route are corrected.

## Still ahead in #2252

Stage 3 (the `openai_compat` flag and exclusive quiet-timeout settings) and stage 4 (docs, examples, ADR 0035 superseded and ADR 0057's status — including the correction to its cutover text that #2279 documents).

Refs #2252, #2247.


jhgaylor (changes requested):
Reviewed the exact stacked range `731c97e0dbe8114aed14a2efea17b6aca852f2f4..1d442c0da7af29c59d060503c79c914be8103948` under the correctness, tenant-isolation, API-compatibility, and simplicity policies. I found two blocking issues:

1. **The webhook migration irreversibly broadens subscriptions and breaks rollback** (`apps/fountain/priv/repo/migrations/20260915130000_strip_retired_caller_tool_subscriptions.exs:38`, `:62-66`). When an endpoint subscribed only to `conversation.caller_tool.*`, `up/0` replaces that explicit filter with the three default turn/provision events. The endpoint therefore starts receiving event classes its owner never selected. The no-op `down/0` also means rolling this code release back restores the old caller-tool emitter but not the filters that consumed it, even though ADR 0057 says this code-removal phase is reversible while the physical tool definitions remain. Keep the retired filter vocabulary valid but non-emitting until the rollback floor, or preserve an empty/no-delivery filter without defaulting it and add migration/rollback coverage; defer irreversible cleanup until it can be recovered or the rollback floor is explicit.

2. **The public webhook catalogue still advertises the event types this head removes** (`docs/reference/webhooks.md:72`). After this change, a user following that row and saving `conversation.caller_tool.started`, `conversation.caller_tool.done`, or `conversation.caller_tool.*` gets `unknown event`, and no such event can be emitted. Update the published catalogue/migration guidance in this change so the documented accepted filters and behavior match `Fountain.Webhooks.Events` and the data migration.

GitHub reports all executed CI jobs green, including `CI required`, six Build and Test partitions, static analysis, release/API contract, SDK jobs, secret scanning, and the coverage gate; the Docs job was skipped. I did not run the Elixir suite locally because the shared worktree contains unrelated in-progress edits. The exact-range `git diff --check` passed.

jhgaylor (changes requested):
Reviewed the exact stacked range `67302a2eb230040441d310de04e9e3d20df48ed8..6f4841596f666feafa90ddf2ae48fc1854fbbe01` (head `6f4841596f666feafa90ddf2ae48fc1854fbbe01`), not `main..head`, under the correctness, tenant-isolation, API-compatibility, and simplicity policies.

The two blockers from my review of old head `1d442c0da7af29c59d060503c79c914be8103948` remain unresolved on this head:

1. **The webhook migration changes tenant delivery scope and destroys the data needed for rollback** (`apps/fountain/priv/repo/migrations/20260915130000_strip_retired_caller_tool_subscriptions.exs:38`, `:62-66`). If an endpoint explicitly subscribed only to `conversation.caller_tool.*`, `up/0` replaces that no-longer-emitting filter with the three default turn/provision events. It therefore starts delivering event classes the owner never selected. The no-op `down/0` cannot restore either exact or wildcard caller-tool filters; rolling this PR back restores the bridge and its emitter from the recorded base but leaves those receivers unsubscribed, contrary to ADR 0057's explicit guarantee that this code-removal release is reversible while the physical tool column remains. A moduledoc and changelog disclosure do not preserve tenant intent. The smallest repair is to avoid rewriting stored filters: let unchanged legacy filters remain valid-but-non-emitting on unrelated endpoint edits while rejecting them when a user newly selects/changes filters, and add an upgrade/rollback regression over exact-only, wildcard-only, mixed, global-wildcard, and unaffected subscriptions. Otherwise preserve enough state for an exact `down/0` and do not widen an emptied subscription.

2. **The published webhook catalogue still documents filters that this head rejects and can never emit** (`docs/reference/webhooks.md:72`). After `Fountain.Webhooks.Events` removes `caller_tool`, a user following that row and saving `conversation.caller_tool.started`, `conversation.caller_tool.done`, or `conversation.caller_tool.*` receives `unknown event`; existing receivers get no such event. Update the public catalogue and explain the migration behavior in the same change. This is particularly necessary because the Docs job was skipped for this CI plan.

Coverage: I inspected the PR text, discussion, historical review, base/head topology, all current checks, the approved-base Review Loop/reviewer/verification instructions, `CLAUDE.md`, `CONTRIBUTING.md`, the server/docs guides, ADRs 0024/0035/0043/0056/0057, affected callers and repository-wide symbol/doc references. I inspected every changed path: `apps/fountain/lib/fountain/caller_tools.ex`, `apps/fountain/lib/fountain/conversations.ex`, `apps/fountain/lib/fountain/conversations/conversation.ex`, `apps/fountain/lib/fountain/conversations/conversation_server.ex`, `apps/fountain/lib/fountain/conversations/launch.ex`, `apps/fountain/lib/fountain/conversations/mcp_servers.ex`, `apps/fountain/lib/fountain/conversations/pending.ex`, `apps/fountain/lib/fountain/webhooks/events.ex`, `apps/fountain/lib/fountain_web/controllers/caller_mcp_controller.ex`, `apps/fountain/lib/fountain_web/controllers/conversation_controller.ex`, `apps/fountain/lib/fountain_web/router.ex`, `apps/fountain/priv/repo/migrations/20260915130000_strip_retired_caller_tool_subscriptions.exs`, `apps/fountain/test/fountain/audit_guardrail_test.exs`, `apps/fountain/test/fountain/caller_tools_test.exs`, `apps/fountain/test/fountain/conversations/attach_test.exs`, `apps/fountain/test/fountain/conversations/conversation_server_acp_test.exs`, `apps/fountain/test/fountain/conversations/conversation_server_redaction_test.exs`, `apps/fountain/test/fountain/conversations/conversation_server_size_test.exs`, `apps/fountain/test/fountain/conversations/mcp_servers_test.exs`, `apps/fountain/test/fountain/conversations/pending_test.exs`, `apps/fountain/test/fountain/extensions_test.exs`, `apps/fountain/test/fountain_web/api_spec_test.exs`, `apps/fountain/test/fountain_web/controllers/caller_mcp_controller_test.exs`, `apps/fountain/test/fountain_web/controllers/saved_allowance_channel_test.exs`, `apps/fountain/test/fountain_web/protocol_retirement_test.exs`, `apps/fountain_buzz/lib/fountain_buzz/mcp_router.ex`, `apps/fountain_google/lib/fountain_google/router.ex`, and `changelog.d/2252-retire-caller-tool-bridge.md`.

Verification: `git diff --check` passed for the exact range. GitHub reports all substantive checks green for this head: `CI required`, six Build and Test partitions, coverage, Elixir static analysis, release/API contract, both Elixir/Python/TypeScript/Swift SDK matrices, CLI/plugins, core distribution, Compose, policy/alerts, and secret scanning. The Docs job and reuse/dispatch jobs were skipped. I did not run the Elixir suite locally because the shared checkout is on a later stacked branch rather than this exact head. There is no checked-in regression test for the new data migration; the PR description's manual database table is not executable rollback evidence.

BinaryBourbon:
Both findings were right, and the first one changed my mind about the approach entirely rather than getting patched. Fixed in `5e2517e0`.

**1. The migration was the wrong instrument, so it is deleted.** You are correct on both counts: substituting the three defaults for an emptied subscription hands an endpoint event classes its owner never chose, and a no-op `down/0` contradicts ADR 0057's own claim that this phase stays reversible while the physical tool definitions remain.

Taking your first remedy — *keep the retired filter vocabulary valid but non-emitting until the rollback floor*. The problem it was aimed at is narrower than I had it: `validate_event_types/1` re-validates the **whole** `event_types` array on every update, so a stored retired type is refused the next time an owner changes an unrelated field. That needs the vocabulary to keep accepting what it used to emit; it never needed anyone's data rewritten.

`Fountain.Webhooks.Events` grows a `@retired` list — not in `types/0`, not in the catalogue the manual renders, never published, still accepted by `valid_filter?/1`. A filter matching no type never fires, and `matches?/2` is a string compare that never consults the module. **Stored subscriptions are now left exactly as their owners wrote them**: nothing rewritten, widened or dropped. A rollback restores the emitter and finds its filters intact. An entry leaves `@retired` only once the rollback floor passes the release that retired it — for `caller_tool`, #2273's gate, which the moduledoc names.

**2. Published catalogue.** `docs/reference/webhooks.md` loses the `caller_tool` row and gains a **Retired events** section: the two types are still accepted as filters, why that is deliberate, and that dropping them is the owner's call whenever it suits. It also notes that tools configured *on an agent* are a different mechanism and are not retired, since that is the confusion this whole retirement invites.

**Verified by reverting**, not just by passing: narrowing `valid_filter?/1` back to the catalogue alone turns four of the new tests red. Coverage at both levels — `events_test` for the vocabulary (a typo *inside* a retired stage is still refused; a guard that `retired/0` is not merely empty), and `webhooks_test` for your actual scenario: an endpoint created carrying `conversation.caller_tool.started` can still have its URL changed, and its `event_types` come back untouched.

Also folded in from #2279's review, which lands underneath this: the caller-tool MCP append moved into that PR, so a legacy row's tools stop being advertised in the same change that removes the controllers which could answer them. The regression is adapted here, where `caller_tools` is no longer a schema field at all.

**Validation:** `mix precommit` PASSED (9 stages); whole suite **5,897 core tests + 136/30/33/60/29 across the extensions, 0 failures**; `scripts/test-docs.sh` 5 suites 0 failures; changelog/conflict/whitespace clean.

On the shared worktree you hit: that was real — another agent was working in this checkout. I ran against a private `DATABASE_URL`. Worth knowing for next time: `Fountain.AuditTest`'s window assertion fails once a reused test DB has accumulated committed audit rows (#2178), green after `TRUNCATE audit_events`; and leftover live sandbox rows there produce `:fleet_full`.

jhgaylor (dismissed):
Adversarial verdict: **APPROVED** for the exact stacked range `ba96b61d70782d4cafe3463c52002c64dbbeba58..b739c688f2a5f7e48acd9409d6d2c36f8ca21bb8` (head `b739c688f2a5f7e48acd9409d6d2c36f8ca21bb8`), not `main..head`.

The two blockers from the historical reviews are resolved on this head:

1. **Stored webhook intent and rollback are preserved.** The forward migration that stripped `conversation.caller_tool.*`, widened emptied subscriptions to defaults, and had no reconstructive rollback is absent from the exact range. `Fountain.Webhooks.Events` instead keeps the two exact retired types and the retired stage wildcard valid while removing them from `types/0` and `catalogue/0`. No subscription row is rewritten; live dispatch has no caller-tool type to emit; and rolling back restores the old emitter against the owners' unchanged filters. The new unit/context coverage pins exact retired types, the wildcard, rejection of typos, absence from the live catalogue, retention through an unrelated endpoint update, and a nonempty retired-list guard.

2. **The public catalogue now matches behavior.** `docs/reference/webhooks.md` removes the live `caller_tool` row and explicitly documents that the two exact types and `conversation.caller_tool.*` remain accepted-but-non-emitting until the rollback floor, while agent-configured MCP tools are not the retired request-defined bridge. The changelog says the same.

I found no medium-or-higher correctness, tenant-isolation, API-compatibility, security, rollback, or maintainability defect. The bridge removal is internally complete: the module/controller/route, schema/cast/write paths, queue projection, ConversationServer entry points and handlers, parked-call state/timers/drain, webhook emission catalogue, audit guardrail entry, and focused tests are removed; the physical column and its applied migration remain for rolling compatibility; legacy values are not advertised; and agent, connection, extension, team, native SSE, and permission paths remain intact. Repository-wide symbol tracing found only intended history, the retained applied column migration, retirement docs/tests, and two low-impact stale timing rows. Those deleted-test entries in `scripts/test-timings.tsv` and the obsolete bridge comment at `apps/fountain/lib/fountain/conversations/launch.ex:416` are non-blocking cleanup: partitioning enumerates existing files rather than this manifest, and neither changes runtime behavior.

Path coverage (all 30 changed paths): `apps/fountain/lib/fountain/caller_tools.ex`, `apps/fountain/lib/fountain/conversations.ex`, `apps/fountain/lib/fountain/conversations/conversation.ex`, `apps/fountain/lib/fountain/conversations/conversation_server.ex`, `apps/fountain/lib/fountain/conversations/launch.ex`, `apps/fountain/lib/fountain/conversations/mcp_servers.ex`, `apps/fountain/lib/fountain/conversations/pending.ex`, `apps/fountain/lib/fountain/webhooks/events.ex`, `apps/fountain/lib/fountain_web/controllers/caller_mcp_controller.ex`, `apps/fountain/lib/fountain_web/controllers/conversation_controller.ex`, `apps/fountain/lib/fountain_web/router.ex`, `apps/fountain/test/fountain/audit_guardrail_test.exs`, `apps/fountain/test/fountain/caller_tools_test.exs`, `apps/fountain/test/fountain/conversations/attach_test.exs`, `apps/fountain/test/fountain/conversations/conversation_server_acp_test.exs`, `apps/fountain/test/fountain/conversations/conversation_server_redaction_test.exs`, `apps/fountain/test/fountain/conversations/conversation_server_size_test.exs`, `apps/fountain/test/fountain/conversations/mcp_servers_test.exs`, `apps/fountain/test/fountain/conversations/pending_test.exs`, `apps/fountain/test/fountain/extensions_test.exs`, `apps/fountain/test/fountain/webhooks/events_test.exs`, `apps/fountain/test/fountain/webhooks_test.exs`, `apps/fountain/test/fountain_web/api_spec_test.exs`, `apps/fountain/test/fountain_web/controllers/caller_mcp_controller_test.exs`, `apps/fountain/test/fountain_web/controllers/saved_allowance_channel_test.exs`, `apps/fountain/test/fountain_web/protocol_retirement_test.exs`, `apps/fountain_buzz/lib/fountain_buzz/mcp_router.ex`, `apps/fountain_google/lib/fountain_google/router.ex`, `changelog.d/2252-retire-caller-tool-bridge.md`, and `docs/reference/webhooks.md`.

Verification: exact-range `git diff --check` passed. GitHub reports `CI required`, all six Build and Test partitions, coverage, Elixir static analysis, release/API contract, both Elixir SDK versions, both Python and TypeScript SDK versions, both Swift platforms, CLI/plugins, core distribution, both Compose jobs, policy/alerts, documentation classification, and secret scanning successful. The Docs job and reuse/dispatch jobs were skipped. I did not rerun the Elixir suites locally because the shared checkout is on a later stacked branch, not this exact head. BinaryBourbon separately reports `mix precommit`, the full core/extension suites, and `scripts/test-docs.sh` passing; I treated that as supporting evidence, not proof.

BinaryBourbon:
Thanks for the approval. The two non-blocking items are cleared in `4c1f79b8`, plus one they led to:

- **`scripts/test-timings.tsv`** — the four rows for deleted suites are gone. Inert as you say (partitioning enumerates files on disk), but a manifest naming deleted tests is one nobody can trust to be current.
- **`launch.ex:416`** — worse than stale: the comment described the `caller_tools:` line this stack deleted, so it was left sitting above `labels:` describing nothing.
- **Following that comment turned up a third thing.** `Launch.channel_conversation/1`'s doc cited *"a tool answer on the bridge (#1202)"* as the request shape it exists for — and that was its **only production caller**, so the function is now reachable from its test alone.

  I kept it rather than deleting it, and said so in the doc rather than leaving it to be discovered: it is the one read-only door onto the channel resolution `start_or_resume_conversation/2` performs, and duplicating that resolution elsewhere is how the two drift. `saved_allowance_channel_test.exs` still pins that the two agree, including for a foreign tenant. The doc now states plainly that nothing in production calls it and says to remove it with its test if nothing has claimed it by the release. Happy to delete it instead if you would rather not carry it — it is a two-line change either way, and I did not want to make that call silently inside an approved PR.

Whole suite re-run after the change: **5,897 core + 136/30/33/60/29 across the extensions, 0 failures**. `mix precommit` PASSED.

One thing worth flagging because it touched this PR's history: an earlier `git add -A` of mine swept the untracked `work/` and `scripts/destink/` scratch directories into a commit here, and when I cleaned it up, a later `git add scripts` pulled `scripts/destink/node_modules` straight back in — 711 files. Both are stripped now and all four branch heads verify clean (`git ls-tree -r --name-only` → 0). Neither directory is gitignored, which is why this is so easy to do twice; might be worth a `.gitignore` entry independently of this stack.

jhgaylor:
Housekeeping on this PR, and an apology for the noise: I blanked this description a few minutes ago with a stray `--body-file /dev/null` while posting elsewhere. It is restored, and updated for the scope change below. Nothing about the commits changed.

**Scope moved after #2279's fifth review round.** The caller MCP adapter and its route are now removed *there*, not here. The reason is worth knowing: `CallerMcpController.build_ctx/2` scoped only to the account and never bound a sandbox's callback key to the conversation named in its path, the way `Conversations.sandbox_owns?/2` does for labels — so while the route existed, any sandbox key in the tenant could park calls on another conversation's turn. It predates the retirement, but it is only reachable while the route is there, and the route was already dead weight once nothing advertised those tools. Deleting it beat adding a binding check to code scheduled for deletion.

So all five retired paths go unrouted in one change, and **this PR is now purely the internals** nothing can reach: `Fountain.CallerTools`, the registration write, the schema field, and the parked-call plumbing in `ConversationServer` and `Pending`.

The `@retired` webhook vocabulary and the docs page also moved to #2279, since nothing can publish a `caller_tool` stage once that lands. What this PR does with it: the transitional catalogue guard added there is deleted here, because its call sites are gone — while `@retired` itself **stays**, since it is what keeps an endpoint still naming a retired type editable, and it leaves at the rollback floor (#2273) rather than here.

Your approval was on the previous head. Re-review when convenient — the diff is smaller than it was.

jhgaylor (changes requested):
## Verdict: changes requested

Reviewed the complete stacked diff `1f1f0ccc78e80bd422277edfa36a709413a5e2ef..06a30793af3ca57a6a8b506bd47abf78000bc89f` through four independent adversarial lenses: correctness, security/tenant boundaries, API/SDK compatibility, and maintainability/verification.

### Medium — the retirement guard permits a retired webhook emitter to return unnoticed

`apps/fountain/test/fountain/webhooks/events_test.exs:59-85` now removes every retired stage from `published_pairs()` before checking that source publication sites belong to the public catalogue, and this change deletes the former transitional assertion that tracked the remaining `caller_tool` sites. A future `publish_stage(..., "caller_tool", ...)` can therefore be introduced while this guard stays green.

That would be a real delivery regression, not merely a catalogue mismatch: `apps/fountain/lib/fountain/webhooks.ex:288-300` derives the event type and calls `Events.matches?/2` without consulting `Events.known?/1`. Retired exact filters and `*` remain valid, so subscribers could receive an event that the retirement contract says is never emitted. The later test around lines 147-155 only checks catalogue membership and cannot prove absence of source publication sites.

Please make the stable invariant explicit: assert that `published_pairs()` contains no stage (preferably no exact `{stage, status}` pair) listed by `Events.retired/0`, while retaining the separate compatibility assertion that historical retired filters remain valid. Once the last emitter is gone, retirement should mean zero publish sites.

### Prior-scope disposition

- The destructive stored-filter migration concern is resolved: this range adds no migration and leaves historical filters intact.
- The caller bridge implementation, persisted application field, parked calls, and executable bridge references are coherently removed; the physical database column is intentionally retained for rolling compatibility.
- Agent-configured application MCP servers remain separate and covered, and the API contract/SDK artifacts are unchanged.
- The inherited behavior allowing a new endpoint with only retired filters is unchanged from the stacked base and is not charged to this PR.

Nonblocking cleanup: ADR 0057 repeats the stage-2 removal in adjacent bullets, and `extensions.ex:125` still says “team and caller servers” although caller servers are gone.

All substantive exact-head checks are green, and `git diff --check` passed. Static parsing passed in an exact-head archive; focused Mix tests could not start because the environment failed to download the `lumis` precompiled NIF (`enoent`).

jhgaylor:
Fixed in `31547414`. This one was a genuinely better invariant than the one I wrote, not just a stricter version of it.

**The skip made the guard blind in the direction that matters.** I had the catalogue walk skip retired stages so it would stay green while `Pending` still held the call sites. Once those sites went, I left the skip in place with a comment about it costing nothing — but it does cost something: a re-added `publish_stage(_, "caller_tool", _)` would sail past it.

And you're right that this is a delivery regression rather than a bookkeeping one. `Webhooks.dispatch` derives the type and calls `Events.matches?/2` without consulting `known?/1`, and a retired exact filter — or plain `*` — is still a valid subscription. So the event would actually be delivered to an endpoint the retirement promised would never hear from it again. The catalogue-membership test later in the file cannot see that, because it only looks at the catalogue.

So the skip is gone and the invariant is stated directly: **`published_pairs()` contains no stage listed by `Events.retired/0`**. The failure message names the two ways out — delete the call site, or take the stage out of `@retired` and put it back in the catalogue and the docs page. The catalogue walk goes back to covering every pair it finds, with no exclusions.

The compatibility assertions that retired filters stay valid are untouched, since those are about stored subscriptions rather than emission — the two halves are now clearly separate tests with separate reasons.

**Verified by reverting**: adding a `publish_stage(_, "caller_tool", _)` call site to `Pending` turns the new guard red with exactly that message.

**Non-blocking cleanups, both done.** ADR 0057 carried a duplicate of its own stage-2 bullet from an earlier merge — removed. `Fountain.Extensions` still promised that a failing extension leaves "the host's own team and caller servers" untouched; there are no caller servers left, so it now says team.

On your note that the new-endpoint-with-only-retired-filters behaviour is inherited from the base and not charged here: agreed, and it is fixed in #2279 (`745f8c81`). `valid_filter?/1` is catalogue-only again, `retired_filter?/1` is the narrow recogniser, and `Endpoint.validate_event_types/1` grandfathers a retired value only when it is already on the stored row — so create is refused and a newly added one on update is refused, while an existing row stays editable.

**Validation:** `mix precommit` PASSED (9 stages). Full suite **5,900 core + 136/30/33/60/29 extensions, 0 failures**. `scripts/test-docs.sh` 5 suites 0 failures. `okf validate decisions` valid.

Also worth knowing given the `lumis` NIF failure you hit: that download is what blocks a cold `MIX_ENV=test` compile here too. Once it is cached the focused suites run fine, and I use a private `DATABASE_URL` because the shared checkout accumulates committed audit rows that fail `Fountain.AuditTest` (#2178) and leftover live sandbox rows that produce `:fleet_full`.

jhgaylor (approved):
## Verdict: approved

Fresh adversarial re-review of all 22 changed paths in exact stacked range `745f8c8150c10cf59662a34e41795d118a1de1d2..3154741492fc08ae606057cbd66276f4647ee29b` through correctness, tenant/security, API/SDK, and maintainability/verification lenses.

No medium-or-higher defects found.

The prior blocker is resolved: the retired-emitter invariant now directly asserts that every value in `Events.retired/0` has zero publish sites, and the catalogue walk no longer skips retired events. The bridge removal is coherent across schemas, writes, queueing, runtime state, handlers, timers, and tests. The physical database column remains for rolling compatibility, agent-configured MCP plus extension/team paths remain preserved and covered, and contract/SDK artifacts are unchanged. Findings inherited from the base PR were excluded.

Low, nonblocking: the changelog statement that retired caller-tool values “stay valid webhook filters” is broader than the implementation. Existing unchanged stored values are grandfathered, while new subscriptions and additions are rejected; the wording should be tightened in a follow-up.

All substantive exact-head checks are green. Diff whitespace validation passed; no local Mix result is claimed.

BinaryBourbon:
**Rebased onto `main` and retargeted**, now that #2279 has landed. Approval carries over — the diff is the same content you approved — but two things are worth stating rather than leaving to be noticed.

**Rebuilt rather than merged.** #2279 was squash-merged, so replaying this branch's original commits collided against the squashed parent (the classic add/add on files both sides "created"). I rebuilt the branch as `main` + the exact reviewed diff instead, which is why the history is one commit now. `gh pr diff` is 26 files, all stage 2.

**It surfaced a regression this branch was carrying.** An earlier merge here had silently reverted #2279's `.env.example` rewrite back to an older wording — the one that predates the ADR 0057 sentence and the two-gates explanation for Connections. Merging would have landed that revert on `main`. Main's version stands; this branch no longer touches the file at all. Worth knowing that a stacked branch can quietly undo its own parent's work through a resolved merge, and neither CI nor the PR diff against the old base showed it.

Re-verified on the rebuilt branch: `mix precommit` PASSED (9 stages), full suite **5,902 core + 136/30/33/60/29 extensions, 0 failures**, `scripts/test-docs.sh` 5/5, `okf validate decisions` valid, changelog and whitespace clean.

Queuing once checks go green.

PR #2282, opened 2026-09-15
refactor(flags): retire the openai_compat feature flag (stage 3)

jhgaylor:
Stage 3 of the ADR 0057 retirement (#2252). **Stacked on #2280** (which is stacked on #2279) — land those first.

The `openai_compat` flag gated the `/v1` endpoints that stages 1 and 2 removed, so it now gates nothing.

## The flag

`@flags` keeps `connections`, which has its own rollout and stays in `@on_without_posthog`. `FEATURE_FLAGS_ON` itself is untouched and still documented — it is the no-PostHog path for any future flag — but **no shipped feature needs a key in it today**, so `.env.example`, `.env.compose.example`, `docker-compose.yml` and the configuration reference now say that rather than naming a flag that no longer exists. A deployment that listed `openai_compat` there can drop the variable.

## Why a flag removal touched 39 lines of the flag suite

`FeatureFlagsTest` used `:openai_compat` as its stand-in for "some flag", so the suite was pinned to whichever real feature happened to exist. It now uses a neutral `"test_flag"` string — `enabled?/2` takes the PostHog key directly, so those tests need no entry in `@flags` at all and the next flag removal will not reach them. This is the "generic feature-flag tests can use a neutral string test flag" item in #2252's handoff.

The `:connections` assertions are deliberately left alone: those cover `@on_without_posthog`, which is about that flag specifically. `AnalyticsTest`'s flag-stamping test moves to the same neutral key, and `SavedAllowanceChannelTest` drops an override it no longer needs.

## The inventory's "exclusive quiet-timeout settings"

There were none. Those turned out to be in-module constants in the two controllers (`@wait_seconds` and friends) and left with them in stage 1; no configuration key was exclusive to the retired dialects. Shared SSE heartbeat settings, generic feature flags and the connection rollout are untouched. Recording it here because the handoff lists it as a work item and "checked, nothing to remove" is the honest outcome rather than a silent omission.

## Validation

- **Whole suite** against a private `DATABASE_URL`: `fountain` 6 doctests + 5,889 tests, `fountain_buzz` 136, `fountain_slack` 30, `fountain_support` 33, `fountain_google` 60, `fountain_microsoft` 29 — **0 failures**.
- `mix precommit` — **PASSED** (9 stages).
- `scripts/test-docs.sh` — 13 tests, 0 failures (this PR edits `docs/configuration.md`, which is guarded).
- `changelog.py check`, `conflict-markers.py`, `git diff --check` — pass.

[...]

PR #2283, opened 2026-09-15
docs: close the retirement record — ADR 0057 final, inventory marked executed (stage 4)

jhgaylor:
Stage 4 and the last of the ADR 0057 stack (#2252). **Stacked on #2282** → #2280 → #2279; land those first.

**Scope changed after review.** The migration pages, the example removals and the flag guidance that were originally here have moved into #2279, because `contributing/protocol-retirement.md` requires the source-removal PR to carry the public migration guide and forbids leaving shipped examples calling removed endpoints. What is left is the part that genuinely belongs last: closing the record once the whole stack is in.

## What this PR is now

- **ADR 0057 → fully built.** #2279 set it to Accepted with a status section naming what was built there and what each later change adds. This is the final pass: the status now describes the whole stack in order, including why the caller-tool MCP append had to move into #2279 rather than wait for #2280 — those controllers were the only parties that could hand a parked call back, and a tool nobody can answer is worse than no tool. The consequences section records the actual numbers (`conversation_server.ex` 2,364 → 2,213) instead of the pre-implementation estimate.
- **`contributing/protocol-retirement.md` marked executed**, kept as ADR 0057's audit trail rather than deleted: what was surveyed, at which commits, and which of it was shared rather than exclusive. Its line counts are labelled a snapshot of `44c70b61`. It also names the two items still deliberately open (#2273, and the release/site coordination) and points at the place the plan and the implementation disagree — the retirement response, which is a plain 404 only on `/v1`.
- **`feature-status.md`**: #2279 said the endpoints are gone so the flag gates nothing, which was what was true there. With #2282 merged the flag itself is gone, so this says that instead.
- **`decisions/index.md`** regenerated.

`decisions/0035` is untouched here — #2279 supersedes it, and stage 4 originally carried a reflow of the same paragraph with no change of meaning, which is dropped.

## Validation

- `scripts/test-docs.sh` — **5 suites, 0 failures**.
- `scripts/decisions-index.sh` is idempotent on re-run; `okf validate decisions` → `valid: true`, 6 warnings, the same six pre-existing actor-convention ones #2275 recorded.
- `changelog.py check`, `conflict-markers.py`, `git diff --check` — pass.
- Documentation-only, so no Elixir suite, per CLAUDE.md.

`CHANGELOG.md` deliberately untouched — the release roll owns it.

## Still open after this stack

Each with its own gate: the `conversations.caller_tools` column drop (#2273), the site catalog and examples (managoat/site#3), the AG-UI outreach section (#1196), and choosing and verifying the breaking release tag. A code merge is not a deployment.

Refs #2252, #2247.


BinaryBourbon (approved):
Adversarial verdict: APPROVED

Reviewed the exact stacked range `ef1bffc70c876b42683d5268d3006d1b6eb38745..d369b3e6ce611b5ceb707d39ecccf6593b3cf394`, not main-to-head. Three independent reviewers covered correctness, security/tenant isolation, and API/docs/SDK behavior; the coordinating review covered simplicity/maintainability and adjudication. No newly introduced medium-or-higher finding remains in this four-file documentation delta.

Coverage: `contributing/protocol-retirement.md`, `decisions/0057-retire-public-compatibility-protocols.md`, `decisions/index.md`, and `docs/reference/feature-status.md`. The completion record matches source removal of the retired routes/controllers, request-tool bridge, flag, and runnable examples. It explicitly leaves physical column removal, release/deployment, and external-site coordination behind their own gates. Preserved historical data and stored retired webhook filters are consistent with the rollback boundary. This approval covers only this stacked delta; it does not waive the separate changes requested on parent PR #2282.

Accepted nonblocking precision notes under the policy's low-severity allowance:

- ADR 0057:37-39 could say retired event filters “remain accepted when already stored on an endpoint.” New/newly added retired filters are rejected; only unchanged stored values are grandfathered.
- ADR 0057:143 could distinguish the previous 2,364-line size ceiling from the actual 2,334-line starting file. The final file has 2,213 lines. This does not change the retirement outcome.

Verification: current-head docs-plan CI passed Docs, SDK matrices, CI policy and the CI-required gate (run `34997620178`); Decisions and secret scanning also passed. Server/static/release/coverage checks were intentionally skipped by that docs plan, not represented here as newly executed. An isolated archive of this exact head passed `okf validate` with zero errors and six unrelated existing warnings; regenerating the ADR index matched committed bytes; `git diff --check` passed. The attempted focused local Elixir checks were blocked before execution by Mix.PubSub's sandbox-denied TCP socket (`:eperm`). No Review Loop App summary is present and dispatch was skipped: this is the requested manual adversarial review, not a claim that the service's independent full verification recipe ran.

BinaryBourbon:
**Rebased onto `main` and retargeted**, now that #2282 has landed. Approval carries over — 4 files, all stage 4.

Rebuilt as `main` + the reviewed diff, same as the rest of the stack. **Diffing the result against `main` before committing caught two hunks that would have reverted work already there**, which is the third time this stack has produced that failure:

- ADR 0057's `openai_compat` bullet — `main` explains that a deployment drops only that entry and keeps any other, because `connections` still decides the Connections rollout wherever PostHog answers flags. This branch carried the cruder "no shipped feature needs a key in it".
- `changelog.d/2252-retire-openai-compat-flag.md` — same distinction, same revert.

Both come from the #2282 work and are more accurate than what this branch had, so `main`'s wording stands in both. The `feature-status.md` hunk is the opposite case and is kept: it says the flag *is gone* rather than "on its way out", which only became true when #2282 landed.

Also corrected while here: the inventory's "Executed" note still described the planned stage order. It now says what actually happened — routes, controllers, caller MCP adapter and the whole public cutover together in #2279, then the bridge internals in #2280, then the flag in #2282.

Verified: `scripts/test-docs.sh` 5/5, `okf validate decisions` valid with the six pre-existing warnings, `decisions-index.sh` idempotent, changelog/conflict/whitespace clean. Documentation-only, so no Elixir suite per CLAUDE.md.

Queuing once checks go green. After this the code side of #2252 is complete; #2273, managoat/site#3 and the release tag remain.

Issue #2255, opened 2026-09-15
refactor(conversations): Termination, tranche 2 — the forced-teardown family


jhgaylor:
## Completion verified — September 18, 2026

Closing as completed after checking the child records on September 18, 2026.

- Stage issues #2256, #2257, #2258, #2259 and #2268 are closed.
- #2285 and #2309 merged, covering the remaining decisions recorded in this tracker's comments.
- The latest status said the tracker was open only as the parent of #2307. That issue is now closed (September 16); its machine-ownership program was tracked separately in #2344, which is also closed after #2425 merged.

No outstanding deliverable remains in this forced-teardown tranche. The original design and its comments remain available as history.

---

## Original scope and evidence (preserved)

All citations are against `main` at `269f2fe6` (2026-09-15, with every #2175 stage merged).

## Context

#2175 gave each conversation lifecycle verb one owner. Stage 1 (#2209, landed as
#2223) created `Fountain.Conversations.Termination` and moved the terminate and
release *clients*, `retire_terminated_sandbox/2` and the lifecycle audit into it,
but explicitly left three row-writing functions behind in `Fountain.Conversations`:
`_unsafe_release_conversation/2`, `_unsafe_finish_conversation_termination/2` and
`_unsafe_fence_sandbox_for_teardown/2` (#2209 body, "not moved by this stage (a
second tranche may; see the open decisions on #2175)"). #2175's second memo (issue
comment, 2026-09-15) named the rest of that pile: the teardown fence, the admin
reap, `_unsafe_destroy_homes_for_agent`/`_unsafe_destroy_home`, and
`Fountain.Accounts.Deletion`'s teardown — "the whole forced-teardown family," too
large for a #1369-sized stage. #2175's landed decision 6 (posted 2026-09-15) says:
file it as its own tracker with stage issues now, run it as its own stack later.
This is that tracker.

`Fountain.Conversations.Termination` already exists at
`apps/fountain/lib/fountain/conversations/termination.ex` (212 lines). Its
moduledoc states the boundary stage 1 drew: "the row writes stay in
`Fountain.Conversations`" and names the same three functions. Stage 3 below
proposes moving one of those three (the fence) into `Termination` per decision 6;
doing so means updating that sentence, which is a documentation change, not a
behaviour change — flagged under Open Decisions since it revises a boundary
stage 1 stated deliberately.

## Family inventory

| Function | File:line (main@269f2fe6) | Callers | Locks | Audit | Tests |
[...]

PR #2285, opened 2026-09-15
refactor(conversations): the admin reap records its own audit event

jhgaylor:
Refs #2255 (settles decision 4).

## Before

`Fountain.Conversations.Termination.reap_sandbox/1` recorded nothing. Its two
admin-driven callers each built and recorded the identical
`admin.sandbox.reaped` event afterwards:

- `FountainWeb.AdminController.reap_sandbox/2`
- `FountainWeb.AdminLive.Sandboxes.handle_event("reap_sandbox", ...)`

## After

`Termination.reap_sandbox/2` takes an `opts` keyword (`reap_sandbox(sandbox_id,
opts \\ [])`, so the 1-arg call every other caller uses is unchanged). When
`opts[:admin_user_id]` is present and the reap returns `{:ok, outcome}`, the
context itself records `admin.sandbox.reaped` — after the row write, outside
any transaction (the reap opens none) — with the same `event_type`,
`target_user_id: nil`, and metadata keys and values as the two callers used to
write. On `{:error, _}` nothing is recorded, same as before. Both callers now
pass `admin_user_id: admin.id` (the controller's
`conn.assigns.current_user.id`, the LiveView's `socket.assigns.current_user.id`)
and their own `Audit.record_admin` calls are gone.

`reap_all_for_user/1` (called by `Accounts.suspend_user/1`) passes no admin id
and so still records nothing per sandbox, exactly as today — the docstring
says so explicitly now.

## The guardrail

`audit_guardrail_test.exs`'s `@must_audit` table drives `Fountain.Audit.record/1`
via `Audit.list_recent_for_user/2`, which reads `audit_events`. Admin actions
go through the separate `Audit.record_admin/1` into `admin_events`, keyed by
actor/target rather than a tenant user, so that table cannot express this
event. Per the addendum, the pin lives in the reap's own test coverage
instead: two new cases in `conversation_audit_test.exs` assert that
`Termination.reap_sandbox(id, admin_user_id: admin.id)` leaves exactly one
`admin.sandbox.reaped` `AdminEvent` with the expected `actor_user_id`,
`target_user_id` and metadata, and that `reap_sandbox(id)` with no option
leaves none.

## Files touched

- `apps/fountain/lib/fountain/conversations/termination.ex` — `reap_sandbox/1`
  → `reap_sandbox/2` plus a private `audit_reap/3`; 440 → 473 lines.
- `apps/fountain/lib/fountain_web/controllers/admin_controller.ex` — drops its
  `Audit.record_admin` call, passes `admin_user_id:`; 608 → 603 lines.
- `apps/fountain/lib/fountain_web/live/admin_live/sandboxes.ex` — same;
  320 → 315 lines.
- `apps/fountain/test/fountain/conversations/conversation_audit_test.exs` —
  two new cases; 198 → 230 lines.

## Tests run unchanged (from `apps/fountain`)

- `test/fountain/conversations/conversation_audit_test.exs` — 11 tests, 0
  failures (includes the 2 new cases above)
- `test/fountain/conversations/sandbox_reset_test.exs` +
  `test/fountain/conversations_read_state_test.exs` — 45 tests, 0 failures
- `test/fountain_web/live/admin_sandboxes_live_test.exs` +
  `test/fountain_web/controllers/admin_controller_test.exs` — 41 tests, 0
  failures
- `test/fountain/accounts_suspension_test.exs` +
  `test/fountain/audit_guardrail_test.exs` + `test/fountain/audit_test.exs` —
  119 tests, 0 failures

Full gate from the worktree root: `precommit: PASSED (10 stages, 508s)`.

No behaviour change a user can observe; no changelog fragment (the admin
trail keeps the same event with the same fields).

Refs #2255 — settles decision 4. Decisions 2 and 3 (unifying the "is anyone
else on this machine" predicates, and routing the reaper through the
advisory lock) are a separate PR per the tracker's last comment; #2255 stays
open until that one lands too.


BinaryBourbon (approved):
Adversarial review verdict: **APPROVE**

Reviewed base `978efe75a31eb45b8e735db6efb3d7022fadf9b2` through head `d81d6188731bdd47cceb81822865cbdaff7e39ad`.

Two independent adversarial passes traced every changed path and the reap callers through the API controller, admin LiveView, suspension path, audit persistence, and regression coverage. The context-owned audit matches #2255 decision 4: authenticated admin attribution is forwarded, successful admin reaps preserve the prior event shape, errors remain silent, and suspension reaps remain unaudited.

No tenant-isolation or API compatibility regression found. Focused verification completed with 233 tests and 0 failures; formatting, diff checks, required CI, and secret scanning pass on this head.

No blocking findings.

PR #2286, opened 2026-09-15
refactor(conversations): one liveness scan for the reaps, and the reaper writes under the sandbox lock

jhgaylor:
Refs #2255 (settles decisions 2 and 3).

## Decision 2 — one liveness scan, pure refactor

`Termination.reap_sandbox/1` and `Workers.SandboxReaper` each scanned a
sandbox's preloaded conversations for a live `ConversationServer`
themselves — the same "is any server alive on this machine" question asked
twice. Adds two functions to `Fountain.Conversations.Lifecycle`, over a
`%Sandbox{}` with `:conversations` preloaded (both callers already carry
that from the query that fetched the sandbox, so this runs no second
query):

- `live_conversation_ids/1` — the conversation ids with a live server, for
  `Termination.reap_sandbox/1`, which needs the list to act on (it
  terminates each live conversation).
- `any_server_alive?/1` — the boolean, deferring to
  `live_conversation_ids/1`, for `SandboxReaper.release_stuck_sandboxes/0`
  and `sweep_abandoned_sandboxes/0`, which only decide whether to touch the
  row.

`SandboxReaper.server_alive?/1` is deleted; both its call sites now call
`Lifecycle.any_server_alive?/1`. Same registry lookup (`ConversationServer.whereis/1`),
same set, no behaviour change.

## Decision 3 — the reaper's writes under the sandbox lock, and a durable park claim

What shipped here grew across four rounds of adversarial review, each
finding closed a real race:

- **Round 1**: the reaper's terminal/parked writes now run under
  `Conversations.with_sandbox_lock/2` (namespace 4316), the same advisory
  lock turn admission takes, with the row revalidated fresh under the lock
  before each write — closing the original admission-before-lock overwrite.
- **Round 2**: the locked recheck also enforces the 15-minute grace window
  the outer scan uses, and every date comparison uses `DateTime.before?/2`
  instead of struct `<` (which compares fields alphabetically and gets
  month/year boundaries backwards).
- **Round 3**: a durable claim, `sandboxes.park_claimed_at` (migration
  `20260915150000_add_sandbox_park_claim.exs`), replaces writing
  `suspended` before the provider checkpoint/suspend calls run. The row
  stays `ready` with a live claim while those calls are outside any lock;
  only a second locked write, after the provider call is known to have
  succeeded, flips it to `suspended`. `Wake.maybe_reuse_sandbox/1` and
  `Launch.check_attachable/4` both refuse a live-claimed `ready` sandbox
  with a new retryable `{:error, :sandbox_parking}` (503, `Retry-After`).
  `Lifecycle.park_claim_live?/2` / `park_claim_ttl/0` (ten minutes) is the
  one shared definition of "live".
- **Round 4**: closed the remaining wake-vs-claim race
  (`Wake`'s reuse arm now re-checks the claim and registers the server
  inside the same lock the reaper's claim step takes), made the
  failed-suspend fallback revalidate liveness and a reset fence exactly
  like the successful-suspend path does, added recovery for a stale claim
  left behind by a crash (reconciled against the provider's own reported
  state, resuming and restarting the max-lifetime clock if the provider
  says it's genuinely suspended), and fixed `TeamScheduleRun`'s snooze
  list to include `:sandbox_parking` (now read from
  `SandboxQueue.transient_errors/0` at compile time, one list instead of
  two).
- **Round 5**: `maybe_reuse_sandbox/1` is now fully side-effect-free —
  recovering a stale claim is compute, gated by every admission check
  (account suspension included) before it ever runs, and an interrupt
  never recovers at all. Recovery became a serialized takeover with an
  exact-stamp CAS across both locks (`claim_stale_park/2`,
  `finish_stale_park/3`, `finalize_stale_park/3`) instead of a
  read-then-write that a fresh reaper claim (or the same operation simply
  outliving the TTL) could erase mid-flight; the reaper's own
  `finalize_park/2` gained the mirror compensation for a suspend that
  succeeds after its claim was taken over. Added `sandboxes.woken_at`
  (same migration) — a durable, database-visible publication of a wake's
  own registration, since Horde's registry is asynchronous and a reaper on
  another node can still read `nil` moments after a registration commits
  elsewhere; `stuck_eligible?/2` and `ready_abandoned?/3` both require it
  nil or past the grace window. The final locked re-read (reuse and
  recovery both) now validates full admissibility — status, the reset/
  teardown fence, and that the conversation's own `sandbox_id` has not
  been rebound — not just the park claim.

`:sandbox_parking` joins the queue's transient-error vocabulary
(`SandboxQueue.transient_errors/0`, `Team.Schedules.describe_error/1`) and
the fallback controller's 503 map. `docs/sdk.md`'s error table is narrowed
to say the SDKs don't map it to `NotReadyError` yet — filed #2291 for that
client work. ADR 0017's "Park claim" section describes the whole protocol,
including the takeover CAS and the `woken_at` marker.

**Rollout note**: a replica on the previous release neither writes nor
reads `park_claimed_at` at all, so a reaper pass or a wake landing on an
old replica during a rolling deploy reproduces the pre-claim race until
that replica drains. Accepted as a one-rollout window, not solved — the
migration is additive and nullable and old code never touches the column.

Provider I/O never runs under `with_sandbox_lock/2`;
`Managoat.Sandbox.suspend/1`/`resume/1` and the home checkpoint always run
outside it, and `Horde.DynamicSupervisor.start_child/2` (a supervisor call,
not provider I/O — the server's real provisioning runs later in its own
`handle_continue(:provision)`) is the one exception explicitly reasoned
about in round 4. No fence, no `:sandbox_kept` refusal, no
`sandbox.teardown_requested` row — the reaper's premise (no live server,
an orphaned row) and `update_sandbox`'s revival guards are exactly as they
were.

## Tests

`sandbox_reaper_lock_order_test.exs` and `sandbox_park_claim_test.exs`
(new) hold the sandbox's advisory lock on an independent connection or
simulate the equivalent race deterministically via a Mimic side effect,
covering: the original admission-before-lock overwrite (round 1); the
grace-window and chronological-comparison regressions (round 2); the
durable claim visible mid-flight with a concurrent wake refused, a
competing owner clearing the claim, and a provider-suspend failure
(round 3); the wake-registration window, the failed-suspend-with-live-
server case and its mirror, and stale-claim recovery including a real
`Managoat.Sandbox.resume/1` call (round 4); the suspended-account gate
ahead of recovery, the takeover CAS losing to a fresh reaper claim and
winning against a mid-suspend reaper (with the resulting compensation),
three `woken_at` grace regressions across both reaper passes, and the
reset/teardown fence and missing/terminal cases at the final locked
re-read (round 5). `sandbox_queue_drain_test.exs` and
`team_scheduler_test.exs` cover `:sandbox_parking` going back in line
rather than being consumed, the latter end-to-end with no stubbed
`send_prompt`. `fallback_controller_test.exs` covers the 503 shape.

Proved unchanged, run by explicit path from `apps/fountain` (28 in
`sandbox_reaper_test.exs`, the rest combined): `conversations_wake_test.exs`,
`wake_race_test.exs`, `attach_test.exs`, `sandbox_queue_test.exs`,
`conversation_audit_test.exs`, `sandbox_reset_test.exs`,
`conversations_read_state_test.exs`, `lifecycle_actions_test.exs`,
`lifecycle_fence_test.exs`, `audit_guardrail_test.exs`,
`admin_sandboxes_live_test.exs`. A combined run across every touched and
proves-unchanged file: 350 tests, 0 failures.

`mise exec -- mix precommit --full` from the worktree root, on the head
commit: `precommit: PASSED (10 stages, 189s)`.

Refs #2255. A second PR follows for decision 4 (the duplicated
`admin.sandbox.reaped` audit); #2255 stays open until that lands.


BinaryBourbon (changes requested):
Adversarial review verdict: **CHANGES REQUESTED**

Reviewed base `978efe75a31eb45b8e735db6efb3d7022fadf9b2` through head `2671c6c3025b714d1ed9f8bc1f01b6eca5304ff4`.

**High — the reaper can act on a stale pre-lock liveness verdict and retire newly active work.** In `SandboxReaper.release_stuck_sandboxes/0` and `sweep_abandoned_sandboxes/0`, candidate selection, server-liveness checks, and activity bounds are evaluated before `Conversations.with_sandbox_lock/2`. If turn admission wins that lock after the scan and commits a running turn, the reaper later acquires the lock and unconditionally writes `failed`, `suspended`, or `terminated` without reloading or revalidating. The idle path additionally invokes provider suspend and `HomeCheckpoint.on_park/1` before taking the lock, so it can interrupt newly admitted work even while admission holds the lock.

This was reproduced against the exact head: provider suspension occurred while concurrent admission held the shared advisory lock; the final database state had a running turn and conversation on a suspended sandbox. The added lock-order tests hold an inert lock and then assert the stale reaper write proceeds, so they do not cover the admission-wins interleaving.

Please re-read and revalidate status, reset state, liveness, and activity eligibility under the shared lock, and make provider suspend/checkpoint participate in a safe claim/ordering protocol so a stale pre-lock verdict cannot affect newly active work. Add admission-wins regressions covering stuck release, idle park, and expiry.

Required CI is green, but it does not exercise this race.

jhgaylor:
Addressed in a second commit, `fix(conversations): revalidate the reaper's writes under the sandbox lock`.

**The finding, restated:** every write site read its candidate list, liveness and activity bounds before ever reaching `Conversations.with_sandbox_lock/2`. An admission that won the lock and committed a running turn *after* that read but *before* the reaper's own turn at the lock was then overwritten by a stale verdict on the reaper's side. The idle path made it worse by calling provider suspend and `HomeCheckpoint.on_park/1` before the lock too, so it could suspend a machine admission had just claimed.

**The fix, one per write path:**

- `release_stuck_sandboxes/0`: `stuck_eligible?/2` (status in `@active_statuses`, `reset_requested_at` nil, `updated_at < cutoff`, not `Lifecycle.any_server_alive?/1`) is now the *one* function both the scan and a fresh re-read inside the lock call — it cannot drift because there's only one copy. The recheck reloads the sandbox (with `:conversations`) fresh under the lock; if it's no longer eligible, nothing is written and no audit is recorded ("became active under the lock; left alone").
- The abandoned sweep (idle park + max-lifetime expiry): `ready_abandoned?/3` does the same thing, pinned to the specific bound (`{:expired, :idle}` or `{:expired, :max_lifetime}`) the outer scan classified the candidate under, with `check_bounds/2` recomputed from the freshly reloaded row's activity.
- Idle park specifically: the lock now **claims the row first** — `claim_for_suspend/1` revalidates and writes `suspended` inside the lock, with no provider I/O there. `HomeCheckpoint.on_park/1` and `Managoat.Sandbox.suspend/1` run only after that write commits and the lock releases. On a provider suspend failure, the existing fallback (mark it `terminated` instead, so pass 2 still destroys the leaked sprite) is itself revalidated: it only proceeds if the row is *still* exactly the `suspended` state this run just claimed, so a legitimate wake in between (say, the very failure that made the provider call fail also woke the machine) is not clobbered by the fallback.
- Max-lifetime expiry has no provider I/O of its own (pass 2 destroys the sprite separately), so it only needed the revalidated write.

**Why turn-admission freshness alone doesn't defeat the max-lifetime path:** its clock is anchored to `inserted_at`/`last_resumed_at` (a continuous run), not last activity, so a fresh turn doesn't reset it — only a live, registered server does (a reattach in flight). The regression test for that path simulates liveness directly rather than a fresh turn, and says why in a comment.

**Tests** (`sandbox_reaper_lock_order_test.exs`, rewritten): six cases, two per path (stuck release, idle park, max-lifetime expiry). One holds the sandbox's advisory lock inertly and shows the write still happens exactly as before (the behaviour decision 3 introduced). The other — the admission-wins regression the review asked for — holds the same lock and, while holding it, commits (or, for the liveness case, makes true) exactly what a concurrent admission would, then asserts the reaper's pass left the row and any turn untouched, recorded no `sandbox.*` audit, and — for idle park — never reached `Managoat.Sandbox.suspend/1` (`Mimic.reject/1`).

Every existing `sandbox_reaper_test.exs` case (28) passes unchanged. `mix precommit --full`: `precommit: PASSED (10 stages, 154s)`.

BinaryBourbon (changes requested):
Adversarial re-review verdict: **CHANGES REQUESTED**

Reviewed base `b59dbd7b1ef2b7dbcb0d595db08438c2ea869347` through head `f7b05c7f6a086c9b7cb557fe5dc6846b55fc6fcf`.

The new locked rechecks close the originally reported admission-before-lock overwrite, but blocking problems remain:

1. **High — a wake after the new suspension claim can still be suspended by stale provider I/O.** `claim_for_suspend/1` commits `status: "suspended"` under the sandbox advisory lock and releases it; only afterward does `suspend_claimed/2` checkpoint and call the provider's suspend operation. The real wake path coordinates through the separate quota reservation locks, resumes the provider, writes `ready`, and can start work before that delayed suspend completes. On E2B and Daytona, where suspend is a real pause/stop, the reaper can therefore leave a live/`ready` database state over a provider-suspended machine and record a false `sandbox.suspended` audit. An isolated barrier test reproduced this exact ordering against the head. The added regression covers admission before the claim, not wake after the claim. Please serialize wake against provider suspension through completion, using a durable non-wakeable operation claim/generation or an equivalent protocol, and add a wake-after-claim regression.

2. **Medium — the locked abandoned recheck drops the established grace condition.** The outer scan requires `updated_at < grace_cutoff`, but `ready_abandoned?/3` does not. A sandbox refreshed while the reaper waits can still be retired when registry liveness is briefly stale, defeating the 15-minute grace intended to protect wake/registration. Use one complete predicate, including the grace cutoff, for both the scan and the locked recheck, with a refreshed-timestamp regression.

3. **Medium — `stuck_eligible?/2` compares `DateTime` structs with `<`.** Elixir structural ordering is not chronological across calendar boundaries (`~U[2026-09-30 00:00:00Z] < ~U[2026-10-01 00:00:00Z]` is false while `DateTime.compare/2` returns `:lt`). The locked recheck can therefore strand genuinely stale provisioning rows around month/year boundaries. Use `DateTime.before?/2` or `DateTime.compare/2` and cover a boundary.

4. **Required CI fails on this head.** Elixir static analysis stops at formatting: `sandbox_reaper.ex` has a `Logger.info/1` call that `mix format --check-formatted` requires on one line. `CI required` consequently fails, and Credo/Dialyzer never run.

The six test partitions, coverage, release/API, SDK, Compose, policy, and secret-scan jobs otherwise pass, but they do not cover the reproduced post-claim race.

jhgaylor:
Addressed in a third commit, `fix(conversations): a durable park claim so wake serializes against the reaper's suspend`.

## Finding 1 (High) — wake could still be suspended by stale provider I/O

The previous round committed `status: "suspended"` under the sandbox's advisory lock and only afterward ran the checkpoint and the provider's suspend call outside it — so a wake winning that window reattached to a row that said `suspended` while the sprite was still running.

This round adds a durable claim, `sandboxes.park_claimed_at` (nullable `utc_datetime_usec`, migration `20260915150000_add_sandbox_park_claim.exs`), and splits the park into three steps:

1. **Claim**, under the advisory lock: revalidate, and if still eligible and the existing claim (if any) is nil or stale, write `park_claimed_at` — status stays `ready`. No provider I/O here.
2. **Provider I/O**, outside any lock: checkpoint, then `Managoat.Sandbox.suspend/1`, exactly as before.
3. **Finalize**, under the lock again, only once the provider call is known to have succeeded: reload, and only if the row is still `ready` with this run's exact claim stamp and still has no live server does it become `suspended` (claim cleared in the same write). If the claim no longer matches, nothing is written. If the claim matches but a server is now live — a wake that accepted a claim gone stale while this run's provider call was still in flight — the claim is cleared without touching status, and the reaper best-effort resumes the machine to undo a suspend that may already have reached the provider.

`Wake.maybe_reuse_sandbox/1` and `Launch.check_attachable/4` both refuse a `ready` sandbox carrying a live claim with a new retryable `{:error, :sandbox_parking}`, rather than probing or attaching through the park in flight. A stale claim (past `Lifecycle.park_claim_ttl/0`, ten minutes — one definition, `Lifecycle.park_claim_live?/2`, used by every caller) is ignored and falls through to the ordinary path, so a claim from a run that crashed before finalizing cannot strand the row.

I checked whether to mirror `Lifecycle.park/4` (the `ConversationServer`'s own idle/max-lifetime park) instead of inventing a new protocol: it re-reads the sandbox with a plain, unlocked read and writes under a row-level `FOR UPDATE` lock, not the advisory lock, and that's safe only because the server is that sandbox's single serializing owner for as long as it lives. The reaper has no equivalent exclusivity against a concurrent wake or turn admission, so I kept the durable-claim protocol above rather than reusing that shape; `Lifecycle.park/4` itself is unchanged.

`:sandbox_parking` joins the queue's transient-error vocabulary (`SandboxQueue.@transient_errors`, `Team.Schedules.describe_error/1`) and the fallback controller's 503 map (`Retry-After: 5`, alongside `:sprite_probe_failed`'s `10`). `docs/sdk.md`'s `NotReadyError` reason table gains it. I checked the OpenAPI schemas and `sdk/contract/contract.json` for an error-reason enum or 503 description naming `sprite_probe_failed`; neither exists, so nothing there needed regenerating and no `sdk-no-release` label applies.

ADR 0017 gets a short amendment describing the claim, the three steps, and the wake refusal. Changelog fragment `changelog.d/2286-park-claim.md`.

**Tests:** `sandbox_reaper_lock_order_test.exs`, describe "the durable park claim" — the row carries a live claim while `Managoat.Sandbox.suspend/1` is blocked mid-call and a concurrent `Wake.maybe_reuse_sandbox/1` is refused with `:sandbox_parking`, then finalizes to `suspended` with the claim cleared and one audit row; a competing owner clearing the claim mid-flight leaves the row untouched; a provider suspend failure falls back to `terminated` with the claim cleared. `sandbox_park_claim_test.exs` covers `Wake.maybe_reuse_sandbox/1` and `Launch.start_conversation/2`'s attach path against both a live and a stale claim, including that `Horde.DynamicSupervisor.start_child/2` and `Managoat.Sandbox.resume/1` are never reached (`Mimic.reject/1`) when a live claim refuses. `sandbox_queue_drain_test.exs` adds a case that a park-claimed persistent home goes back in the queue rather than burning the prompt, exactly like the existing `:provisioning` case. `fallback_controller_test.exs` covers the 503 shape.

## Finding 2 (Medium) — the locked abandoned recheck dropped the grace condition

The outer scan requires `updated_at < grace_cutoff` (15 minutes) in its SQL `where`, but `ready_abandoned?/3` — the recheck run under the lock — did not re-apply it, so a row refreshed while the reaper waited on the lock could still be retired.

`ready_abandoned?/3` now recomputes the grace cutoff from its own `now` and requires `DateTime.before?(sandbox.updated_at, grace_cutoff)` as part of the one predicate both the initial classification and the locked recheck rely on.

**Test:** `sandbox_reaper_lock_order_test.exs`, "admission wins: a refreshed updated_at under the lock is protected by the grace window" — a second connection holds the lock, bumps `updated_at` to now (no new turn, no live server — an ordinary bookkeeping touch), releases, and the reaper leaves the row alone with no write and no audit.

## Finding 3 (Medium) — `DateTime` compared with `<`

`stuck_eligible?/2`'s `sandbox.updated_at < cutoff` and the new grace check in `ready_abandoned?/3` now both use `DateTime.before?/2` instead of the struct comparison operator, which compares fields alphabetically (`:day` before `:month`/`:year`) and gets month/year boundaries backwards.

**Test:** a standalone case in `sandbox_reaper_lock_order_test.exs` reproduces the review's own example directly (`~U[2026-09-30 12:00:00Z]` vs `~U[2026-10-01 00:00:00Z]`): `refute earlier < later` alongside `assert DateTime.before?(earlier, later)`, documented as testing the exact comparison both predicates use (they're private, and their real cutoff is anchored to wall-clock time with no injection point, so this targets the primitive directly rather than adding test-only plumbing to fake "now").

## Finding 4 (required CI) — format failure

The offending `Logger.info/1` call in `release_one_stuck/2`'s `:skipped` branch was a near-duplicate of the `log_became_active/1` helper introduced in the same function elsewhere in this file; folded into that helper rather than just reformatted. `mix format` is clean repo-wide.

## Gate

`mix precommit --full` (worktree root): `precommit: PASSED (10 stages, 189s)` — this is the first round where credo, dialyzer and sobelow actually ran (the format failure stopped them before). Every existing test named in the addendum (`sandbox_reaper_test.exs` 28, `conversations_wake_test.exs`, `attach_test.exs`, `sandbox_queue_test.exs`, the admin/audit/lifecycle suites) passes unchanged; a combined run across all of them plus the new/extended files totals 323 tests, 0 failures.

No behaviour beyond what's described above; this round is a behaviour and schema change, so it carries a migration, an ADR amendment and a changelog fragment as scoped.

BinaryBourbon (changes requested):
Adversarial re-review verdict: **CHANGES REQUESTED**

Reviewed base `b59dbd7b1ef2b7dbcb0d595db08438c2ea869347` through head `6c9fabedc407631fec2b41c674e4c4d51329244b` with independent correctness, security/concurrency, and API/compatibility passes.

The locked rechecks, grace predicate, chronological `DateTime` comparison, and formatting failure from the previous review are fixed, and all required CI is green. Blocking defects remain:

1. **High — a wake that passes the one-time claim check can still be terminated and destroyed when the provider suspend fails.** `Wake.maybe_reuse_sandbox/1` checks `park_claimed_at` before provider probing and the rest of startup (`wake.ex:52-75, 291-304`). The reaper can write its claim after that check, while the accepted wake registers a live server. On a provider suspend error, `expire_after_failed_suspend/3` revalidates only `status == "ready"` plus the matching claim stamp (`sandbox_reaper.ex:494-508`); unlike `finalize_park/2`, it does not check live-server, activity, or reset state. It therefore marks the newly active sandbox terminal, and pass 2 can destroy its still-live provider machine. Close the wake-check-to-registration TOCTOU and make the failed-suspend correction revalidate ownership/liveness. Add a barrier regression where wake passes the initial check, registers during the suspend call, that call fails, and the row remains ready with no expiry audit or provider destroy.

2. **Medium — crash recovery does not reconcile a provider-suspended stale claim.** A crash after provider suspend succeeds but before `finalize_park/2` leaves `ready + park_claimed_at`. After ten minutes, wake ignores the claim; `probe_sandbox/4` discards the returned provider status, and `wake_suspended_sandbox/2` skips the explicit resume/lifecycle transition because the DB still says `ready`. E2B and Daytona later self-start on I/O, but the stale claim remains and `last_resumed_at` is not reset, so a configured max-lifetime clock can include the parked interval and reclaim a just-woken machine. Recover stale claims under the operation protocol by reconciling provider state, clearing the claim, and resuming/stamping as needed. The current stale-claim test covers only provider status `:running`; add crash-after-suspend and `:suspended` recovery coverage.

3. **Medium — scheduled prompts are dropped on the new transient error.** `SandboxQueue` adds `:sandbox_parking` to its transient list and says `TeamScheduleRun` snoozes on exactly that list, but `team_schedule_run.ex:53-55` omits it. A schedule firing during the claim falls through to `:ok`, permanently consuming that occurrence instead of retrying the documented short-lived condition. Add the reason to the snooze guard and cover it.

4. **Medium — the SDK documentation promises an error mapping the SDKs do not implement.** `docs/sdk.md` now says `sandbox_parking` becomes `NotReadyError`, but the TypeScript, Python, Elixir, FountainKit Swift, and legacy Swift mappings omit the code and return their generic API error instead. Update every client mapping and add shared conformance coverage, or narrow the documentation to the actual contract.

The additive nullable migration itself is structurally safe. Because both the migration and ADR are protected human-review paths, the rollout also needs an explicit decision about mixed-version nodes: old replicas ignore `park_claimed_at`, so enabling the new reaper during a rolling deployment recreates the race until they drain.

BinaryBourbon:
Addressed in a fourth commit, `fix(conversations): close the wake-registration race, recover a stale claim, and fix the schedule snooze list`.

## Finding 1 (High) — a wake that passed the one-time claim check could still be terminated and destroyed on a provider suspend failure

**1a — revalidate the failed-suspend fallback.** `expire_after_failed_suspend/3` used to check only `status == "ready"` and the claim stamp. It now revalidates exactly like `finalize_park/2` does for a successful suspend: fresh row `ready`, `park_claimed_at == stamp`, `reset_requested_at` nil, and no live server — only then does the terminate fallback run. If a server registered (or a reset fence landed) while the failed provider call was resolving, the claim is cleared only, the row stays `ready`, a warning is logged, and nothing is recorded — there's no provider suspend to undo here, since the call in question is the one that failed.

**1b — close the wake-check-to-registration window.** `Wake`'s `:reuse` arm now re-checks `park_claimed_at` and calls `start_conversation_server/4` inside the same `Conversations.with_sandbox_lock/2` the reaper's claim step takes, so a claim landing between `maybe_reuse_sandbox/1`'s one-time check and registration is caught. `Horde.DynamicSupervisor.start_child/2` is a supervisor call, not provider I/O (the server's actual provisioning runs later, in its own `handle_continue(:provision)`), so this doesn't violate the "no provider I/O under a lock" rule — `await_registered` and all real provider work stay outside it. Only the reuse call site is wrapped; the fresh-sandbox path calls the same `start_conversation_server/4` but a brand-new sandbox can never carry a claim, so it needed no change.

**Tests:** `sandbox_reaper_lock_order_test.exs` gains the failed-suspend-with-live-server case (a server registers, via `ConversationServer.whereis/1`, while the stubbed provider suspend call is failing — the row stays `ready`, claim cleared, no audit) and its mirror (a claim step that finds a live server first writes no claim at all — pinning that the reaper's own pre-existing liveness recheck still holds now that the wake side of the same race also takes the lock). `sandbox_park_claim_test.exs` gains the window regression: `maybe_reuse_sandbox/1` passes with no claim, then a stub on `Fountain.Billing.check_spend/1` writes the claim as a side effect — deterministically simulating the reaper's claim landing in that exact window rather than depending on real inter-process timing — and the locked registration still catches it, returning `{:error, :sandbox_parking}` with `Horde.DynamicSupervisor.start_child/2` never reached (`Mimic.reject/1`).

## Finding 2 (Medium) — a stale claim was ignored rather than reconciled

A crash after a successful provider suspend, before `finalize_park/2` could finalize it, left `ready + park_claimed_at` behind with no record of what actually happened at the provider. `maybe_reuse_sandbox/1` now reconciles a stale claim (`recover_stale_park/3` → `reconcile_stale_park/3`) under the same quota reservation lock `wake_suspended_sandbox/2` uses (`exclude: sandbox_id`), reading the provider's own reported state — `:running | :suspended | :unknown`, one vocabulary shared by every adapter (sprites, e2b, daytona; confirmed by reading `deps/managoat_sandbox`). `:suspended` means the reaper's suspend call actually succeeded, so this resumes the machine and stamps `last_resumed_at` — restarting the max-lifetime clock rather than silently including the parked interval, exactly as the review asked. Anything else (`:running`, `:unknown`) just drops the stale claim; `:not_found` falls through to `:create_new` like an ordinary probe; any other provider error fails retryably as `:sprite_probe_failed` and leaves the claim for the next wake.

**A deliberate deviation from the brief's literal wording, worth flagging:** the brief asked to modify `probe_sandbox/4` itself to "surface the provider's reported state." I kept `probe_sandbox/4` unchanged instead. Its only other caller is the ordinary, no-claim `ready`/`suspended` ready path, which pattern-matches its success case as the 2-tuple `{:reuse, sandbox_id}` all the way out to `wake_conversation_for/3`'s `case`. Routing a `:suspended`-vs-`ready` mismatch through a richer return shape there would have either changed that contract for every caller (risking a `CaseClauseError` on an outcome nothing expects) or required a parallel, easy-to-miss special case in the one shared function. `reconcile_stale_park/3` is self-contained instead — it duplicates `probe_sandbox/4`'s not-found/runner-offline/generic-error branches, which is a small, deliberate redundancy in exchange for not touching a function used elsewhere, on a return contract another existing test already pins.

**Tests:** `sandbox_park_claim_test.exs` gains the crash-after-suspend case (stale claim, provider reports `:suspended` → `Managoat.Sandbox.resume/1` called, `last_resumed_at` stamped at or after a pre-call timestamp truncated to the field's own second precision, claim cleared, then ordinary reuse) and the existing running-provider stale-claim case now also asserts the claim is cleared and `Managoat.Sandbox.resume/1` is never called (`Mimic.reject/1`), where before it only asserted the reuse outcome.

## Finding 3 (Medium) — the schedule snooze list omitted `:sandbox_parking`

`TeamScheduleRun`'s snooze guard listed `[:busy, :provisioning, :runner_offline, :sandbox_at_capacity]` inline, separate from `SandboxQueue.@transient_errors`, and missed the new reason. Added a public `SandboxQueue.transient_errors/0` and made `TeamScheduleRun` read it at compile time (`@transient_errors SandboxQueue.transient_errors()`), which makes this module compile-time dependent on `SandboxQueue` — Mix orders that automatically, confirmed by a `--force` full recompile. One list now, not two to drift again.

**Test:** `team_scheduler_test.exs` gains an end-to-end case with no `send_prompt` stub at all — a real park-claimed sandbox on an idle conversation, so `ConversationServer.send_prompt/4` finds no live server and calls the real `Wake.wake_conversation/2`, which refuses with `:sandbox_parking` through the real code path — asserting the job snoozes on a fresh firing and gives up with the friendly message once stale, mirroring the existing `:busy` case exactly.

## Finding 4 (Medium) — the SDK documentation promised a mapping the clients don't implement

Decided: narrow the documentation, file the client work. `docs/sdk.md`'s error table no longer lists `sandbox_parking` under `NotReadyError`; it gets its own row saying it's 503 with `Retry-After` but surfaces through each SDK's generic API error until the clients learn it. Filed #2291 (labels `area:sdk`, `stack:2255`) listing the five client mappings (TypeScript, Python, Elixir, FountainKit Swift, legacy Swift) plus the shared conformance fixture; referenced from the doc and the changelog fragment.

## Rollout note

Added to ADR 0017's "Park claim" section: a replica on the previous release neither writes nor reads `park_claimed_at` at all, so a reaper pass or a wake landing on an old replica during a rolling deploy reproduces the pre-claim race until that replica drains. Accepted as a one-rollout window, not solved — the migration is additive and nullable and old code never touches the column, so there's no cheap way to make an old replica respect a column it doesn't know about.

## Gate

Credo initially flagged two new `_unsafe_get_sandbox/1` call sites (`wake.ex` in `recover_stale_park/3` and inside the new locked registration block) for a missing `# ownership:` comment — the check's exact wording is a regex for the literal word "ownership" within 10 lines above the call, not just any nearby comment. Fixed by naming what establishes ownership at each site (both trace back to `maybe_reuse_sandbox/1`'s already tenant-scoped `conv`). `mix precommit --full`: `precommit: PASSED (10 stages, 163s)`. Rebased onto main (three unrelated Swift/SDK PRs landed, no conflicts); re-ran `mix precommit compile format credo test` after: `PASSED (4 stages, 170s)`. A combined run across every touched and proves-unchanged file: 333 tests, 0 failures.

Migration and TTL are unchanged from round 3 (`20260915150000_add_sandbox_park_claim.exs`; `Lifecycle.park_claim_ttl/0` = 10 minutes).

BinaryBourbon (changes requested):
Adversarial re-review verdict: **CHANGES REQUESTED**

Reviewed base `432755bc67ca6f4edae98aa7584ab71665167b57` through head `7a3c74fb183e4f6cdbc830566ec6e9607a4c4c4d` with independent correctness, security/concurrency, and API/compatibility passes.

The fourth commit fixes the prior failed-suspend fallback, scheduled-run retry list, and documented SDK behavior. All required CI is green. The revised claim protocol still has blocking races:

1. **High — stale recovery can erase a live reaper claim.** `maybe_reuse_sandbox/1` decides the originally observed timestamp is stale (`wake.ex:65-74`), but `recover_stale_park/3` later takes only the fleet/user quota locks (namespace 4315), not the reaper's per-sandbox lock (4316), and its fresh read accepts any `ready` row with any non-null claim (`wake.ex:168-175`). A new reaper can replace stale T0 with live T1 and begin provider suspension between those reads. Recovery can see the provider still `:running`, clear T1 unconditionally (`wake.ex:230-232`), and return reuse; registration then sees no claim. When suspension lands, `finalize_park/2` sees `:claim_lost` and deliberately does not resume, leaving the newly live server on a paused E2B/Daytona machine. The same failure occurs if the original operation itself merely outlives the TTL, a case the ADR says is supported. Recovery needs a serialized ownership takeover/exact-stamp CAS that survives the provider round trip, plus a deterministic T0→T1 (and over-TTL active-operation) regression.

2. **High — registration under the PostgreSQL lock does not durably publish the wake.** The new block starts the Horde child while holding the sandbox advisory lock (`wake.ex:425-438`) but commits no database activity/intent. Horde registry propagation is explicitly asynchronous, so after this transaction releases the lock a reaper on another node can acquire it, still observe `ConversationServer.whereis/1 == nil`, and claim/suspend the machine the wake just started. The test stubs `whereis/1` as immediately visible and cannot cover this distributed ordering. Commit a durable wake/admission marker under the lock (the existing grace predicate can then observe it), or use an equivalent database claim, and cover delayed registry visibility.

3. **Medium — the final locked re-read validates only claim liveness, not full admissibility.** If reset, teardown, or max-lifetime expiry wins the same lock after the initial probe, the fresh row can be fenced or terminal, yet the `true` branch still starts a `ConversationServer`. The stale-recovery fallback similarly maps every row that is no longer `ready + claimed` to `{:reuse, sandbox_id}`. Revalidate status, reset/teardown fences, and the expected binding under the final lock; do not turn missing/terminal state into reuse. Add both lock-order regressions.

4. **Medium — stale recovery resumes compute before the account-suspension gate.** `maybe_reuse_sandbox/1` can call provider `resume` and write `ready` before the reuse arm reaches `Accounts.check_not_suspended/1` (`wake.ex:397, 406`). The quota reservation repeats the credit gate but does not check account suspension. A queued or scheduled wake for a suspended tenant can therefore restart provider compute and only then be refused, leaving it running without a server until another reap. Keep `maybe_reuse_sandbox/1` side-effect-free or run all admission gates before recovery side effects, with a suspended-account regression.

The nullable migration is structurally compatible, and the ADR now states the accepted mixed-version rollout window. Those protected-path decisions were reviewed here; they do not resolve the current-head races above.

jhgaylor:
Addressed in a fifth commit, `fix(conversations): a serialized takeover CAS, a durable wake marker, and full admissibility on the final lock`. Implemented finding 4 first, since it restructures where recovery runs, then 1, 2, 3 as instructed.

## Finding 4 (Medium) — make `maybe_reuse_sandbox/1` side-effect-free

Its stale-claim clause now returns `{:recover_stale_park, sandbox_id, observed_stamp}` and does no provider I/O and no write of its own. `wake_conversation_for/3` handles that outcome **after** every admission gate the ordinary reuse arm runs (`Accounts.check_not_suspended/1`, `Billing.check_spend/1`, `check_saved_inference/2`) — recovery is compute, gated exactly like any other compute this wake would restart, so a suspended account is refused before any provider resume runs, not after. `purpose: :interrupt` on a stale-claimed row answers `{:error, :sandbox_parking}` directly, without recovering at all — an interrupt has nothing to run either way, so it doesn't pay for a recovery whose result it would immediately discard; the next `:work` wake recovers it.

**Test:** `sandbox_park_claim_test.exs`, "a suspended account's wake onto a stale-claimed row is refused, and recovery never runs" — sets `suspended_at`, then asserts `{:error, :account_suspended}` with `Managoat.Sandbox.Sprites.get/1`, `resume/1` and `Horde.DynamicSupervisor.start_child/2` all rejected (never called).

## Finding 1 (High) — recovery is a serialized takeover with an exact-stamp CAS

Three steps, both locks in the established order:

1. `claim_stale_park/2`, under `Conversations.with_sandbox_lock/2` (namespace 4316, the reaper's own lock): re-read, require `status == "ready"`, no reset/teardown fence (one `reset_requested_at` check covers both, per finding 3 below), and `park_claimed_at` exactly equal to `observed_stamp` — nil means someone already recovered it (proceed as plain reuse, no takeover needed); a different stamp means a new reaper owns it now (`{:error, :sandbox_parking}`). On a match, write a fresh stamp of our own.
2. `finish_stale_park/3`, outside that lock but still inside the quota reservation transaction (the same precedent `resume_and_wake/1` already sets for `wake_suspended_sandbox/2`): the provider probe and, if suspended, the resume.
3. `finalize_stale_park/3`, the sandbox lock again: a CAS — write only if the takeover stamp is still the one on the row; a mismatch writes nothing and answers `{:error, :sandbox_parking}` retryably rather than resolving a row no longer ours. On a provider error, the takeover stamp is cleared under the same CAS (`clear_stale_park_takeover/2`) so a later wake or the reaper can act on the row instead of it being stranded for the TTL.

The reaper's own `finalize_park/2` gained the mirror case: a provider suspend that **succeeds** after the claim was taken over (by a fresh reaper claim, or a recovery) paused a machine somebody else now owns — the same compensation the existing live-server branch already made, now also firing on a plain claim-stamp mismatch when the row is still `ready`, folded into one `undo_suspend/1` helper both branches call.

**Tests:** `sandbox_park_claim_test.exs` — (a) a fresh reaper claim (T1) winning the takeover lock before this wake's stale observation (T0) reaches it: `{:error, :sandbox_parking}`, nothing touched, no provider calls. `sandbox_reaper_lock_order_test.exs` — (b) a reaper pass leaves a live takeover stamp alone (the same TTL/liveness gate `claim_for_suspend/1` already has, exercised via a takeover stamp specifically); (c) the reaper's own claim taken over mid-suspend, its later successful suspend undoing itself via `undo_suspend/1` — a real blocked provider call released after the takeover write, not a stub of the outcome. (d) the round-4 crash-after-suspend test now runs through the new path unchanged (it calls `Wake.wake_conversation/2` end-to-end, which is exactly what changed underneath it).

## Finding 2 (High) — a durable wake marker the reaper can see

Horde's registry is an asynchronous CRDT: `ConversationServer.whereis/1` can still answer `nil` on a reaper's node moments after a wake registers a server on another one, and retrying that check does not close a real distributed-ordering gap. Added `sandboxes.woken_at :utc_datetime_usec` to the same (still-unmerged) migration `20260915150000_add_sandbox_park_claim.exs` — no version bump, as the brief noted was fine. A wake writes it inside the existing locked block, right before `start_conversation_server/4`; recovery's finalize CAS stamps it too (this recovery is the wake, for the reaper's purposes, exactly as much as starting a server is). `stuck_eligible?/2` and `ready_abandoned?/3` both now also require it nil or past the same 15-minute grace window `updated_at` already gets (`woken_stale?/2`, one definition both call — moved `@abandoned_grace_minutes` earlier in the file so it's defined before first use). Deliberately not `last_resumed_at`, which anchors the max-lifetime clock and stays untouched by an ordinary reuse.

**Tests:** `sandbox_reaper_lock_order_test.exs` — a fresh `woken_at` keeps a `ready` row off `sweep_abandoned_sandboxes/0` even with `ConversationServer.whereis/1` stubbed to `nil` the whole time (simulating exactly what an unpropagated registry looks like); a `woken_at` past the grace window is eligible again; and, defensively, the same gate on `release_stuck_sandboxes/0` for a `pending`/`starting` row, even though nothing on the current provisioning path stamps `woken_at` there yet. One real test bug I caught while writing these: writing `woken_at` through `Conversations.update_sandbox/2` also bumps `updated_at` (Ecto's own timestamp), which alone would exclude the row from the outer scan and defeat the point of the test — fixed by writing it with a raw `Repo.update_all` in all three cases, isolating the check.

## Finding 3 (Medium) — the final locked re-read validates full admissibility

Both the ordinary reuse arm and recovery's finalize now revalidate more than the claim: `status == "ready"` (a fresh `suspended` row retries into that path with `{:error, :sandbox_parking}`; a terminal row maps to `{:ok, :create_new}` and the caller provisions fresh — never a silent reuse), the reset/teardown fence (`reset_requested_at` alone, the same single check `Launch.check_attachable/4` uses, since `Lifecycle.fence_sandbox_for_teardown/2` never sets `teardown_requested_at` without it — confirmed by reading it), and — new — that the conversation's own `sandbox_id` has not been rebound to a different sandbox by a concurrent wake in between, handed off the same way a lost `Horde.DynamicSupervisor.start_child/2` race already is (`ConversationServer.whereis/1`, queue the prompt if a winner is found, else `{:error, :not_running}`). Extracted the whole reuse body into `reuse_sandbox/5`, shared by the plain `{:reuse, _}` outcome and a successful recovery, so this logic exists once rather than twice that could drift.

**Tests:** `sandbox_park_claim_test.exs` — a reset fence and, separately, a teardown fence winning the lock after the initial probe (via a `Fountain.Billing.check_spend/1` side-effect stub, the same deterministic technique round 4's window regression used), each asserting `{:error, :sandbox_reset_pending}` and that `Horde.DynamicSupervisor.start_child/2` is never called; recovery's fallback answering `{:error, :not_found}` for a row deleted between the probe and the takeover, and `{:ok, :create_new}` (provisioning fresh, never reuse) for one that went terminal in the same window.

## Docs

ADR 0017's "Park claim" section gains three paragraphs: the takeover CAS and the reaper's compensation, the `woken_at` marker and why (Horde propagation is asynchronous; the marker is the durable publication of a wake), and the full-admissibility recheck. The changelog fragment's wording still accurately describes the fixed end-state (these are further hardening of the same described races, not new user-observable behavior), so it's unchanged.

## Gate

Credo flagged two more `_unsafe_get_sandbox/1` call sites in the new recovery functions (`finalize_stale_park/3`, `clear_stale_park_takeover/2`) for the same missing-literal-"ownership"-comment reason round 4 hit — fixed the same way. `mix precommit --full`: `precommit: PASSED (10 stages, 189s)`. Rebased onto main (five unrelated PRs landed, including an OpenAI-compat-surface retirement; no conflicts); re-ran `mix precommit compile format credo test`: `PASSED (4 stages, 141s)`. A combined run across every touched and proves-unchanged file (the round 3/4 sets plus `wake_race_test.exs`, `attach_test.exs`, `sandbox_reset_test.exs`, `lifecycle_fence_test.exs`): 350 tests, 0 failures.

BinaryBourbon (changes requested):
Changes requested on `ca917303` despite the green required checks.

1. **Medium — stale-park recovery still performs provider I/O while holding both the sandbox and global fleet locks.** `Wake.recover_stale_park/3` opens `Quotas.with_sandbox_reservation/3` (`wake.ex:179-192`), then `claim_stale_park/2` nests `Conversations.with_sandbox_lock/2` (`wake.ex:204-245`) before `finish_stale_park/3` calls provider `get`/`resume` (`wake.ex:258-303`). Nested `Repo.transaction` calls join the outer transaction via a savepoint (the repository itself documents this at `launch.ex:705-707`), and PostgreSQL transaction-scoped advisory locks are retained until the outer transaction ends. The takeover stamp is therefore uncommitted and lock 4316 is still held during the provider calls, contrary to the code and ADR's three-phase protocol. The outer transaction also retains fleet lock 4315/0, so one slow provider recovery can stall every tenant's sandbox reservation as well as same-sandbox reset/teardown/reaper work. Commit the takeover in a short transaction, perform provider I/O outside both lock scopes, then finalize with the CAS. Add an independent-connection regression proving another tenant's reservation and the sandbox lock remain obtainable while the provider stub is blocked.

2. **Medium — `woken_at` is not a complete durable publication of ready-sandbox server starts.** Its only production writers are in `Wake`, but boot rehydration starts a ready sandbox's server directly in `Rehydrator.spawn_server/1` (`rehydrator.ex:132-153`) with neither the sandbox lock nor the marker. Oban is already running when the detached Rehydrator begins (`application.ex:48-55,106`), so on clustered boot a reaper can see stale `updated_at`, nil `woken_at`, and a lagging Horde registry, then park or expire the machine underneath the newly started server. Even the Wake path writes the marker and performs the external Horde `start_child` inside one DB transaction (`wake.ex:637-694`): caller death or commit failure after child startup rolls the marker back while the child survives. Use one shared two-phase registration primitive for Wake and Rehydrator that commits the marker before the external start (and clears/compensates on failure), and cover the stale-registry and caller-crash windows.

3. **Medium — the documented rolling-deploy window can still suspend live work.** The amended ADR explicitly accepts that previous-release replicas neither read `park_claimed_at` nor write `woken_at` while a new replica may run the new reaper. With the shipped overlapping rolling deployment, an old-node wake and new-node reaper recreate the exact pre-claim race this PR is intended to close. Stage or feature-gate activation of the abandoned-ready reaper until every serving replica understands claims/markers; an additive migration alone does not make the mixed-version behavior safe.

The API/SDK behavior and required CI otherwise look coherent.

jhgaylor:
Closing without merging (Jake, 2026-09-15, after the fifth review round). Every round found a real ordering gap at the edge of the claim protocol, and the fix has grown to touch wake, rehydration, admission, the reaper, the queue, the schedules, two columns and a rollout gate; that is a design, not a fix. The protocol is now designed first as an ADR under #2307, which lists the seven constraints these five rounds established, and will be built behind a gate as its own stack. This branch stays as salvage (the protocol, its tests and the ADR 0017 prose). The parts every round accepted land in a narrowed PR from `main`: the one liveness scan for both reaps, and the stuck-release and max-lifetime-expiry writes revalidated under the sandbox lock; the idle park returns to its pre-PR ordering until the ADR lands.

PR #2267, opened 2026-09-15
refactor(swift): generate sandbox and conversation dependency models

BinaryBourbon:
## Summary
FountainKit's generated Conversation still referenced handwritten Sandbox fields. Generate Sandbox, SandboxDetail, their nested types, Runner and ConversationTreeNode through the existing contract pipeline, removing those handwritten declarations. Preserve nested public names, unknown-enum wrappers and nine existing optional-property compatibility exceptions, including SandboxDetail.SandboxConversation status and midTurn. Missing/null decoding and optional source access have regression coverage.

Usage now reads additive fields from both TurnUsage and UsageTotal. Shared inline Sandbox definitions and incompatible usage aliases fail generation instead of silently choosing one shape. The contributor inventory separates remaining resource batches from behavioral/legacy models.

Refs #2251; parent #2247. This is the first bounded model-family PR.

## Validation
- Swift: 113 tests passed, strict format, release build with warnings as errors.
- Six generator tests cover deterministic output, every newly covered family's additive fields, and incompatible reused shapes.
- Generated output check and Swift map/typed propagation probes passed.
- Contract remains unchanged; existing endpoint/conformance checks remain.
- Conflict-marker, changelog and diff checks passed.

PR #2287, opened 2026-09-15
fix(sdk): fail generation on a Swift field older servers cannot decode

### LINKED ISSUE #2284: Swift SDK: a newly required field on Teammate breaks older-server decoding with every gate green
`contributing/swift-wire-models.md` states the compatibility rule that makes
generated Swift models safe:

> A model that already shipped by hand must not become harder to decode, so
> `OPTIONAL_COMPAT` keeps two kinds of property optional even where the current
> server requires them.

Nothing enforces that rule. It is applied by hand when a generator change lands,
and the only thing that catches a miss afterwards is whether some committed test
happens to decode that type from a payload lacking the key. For 25 of 26
generated decodable types it does. For `Teammate` it does not.

## The gap

A contract change that makes a property required on `Teammate` generates a
non-`Optional` stored property with no `OPTIONAL_COMPAT` pin. Every gate stays
green, and every response from a server older than that field then fails to
decode **whole** — not just that property.

`Teammate` is not an obscure type. It is the return of four public methods in
`Resources/TeamResource.swift`: `list()`, `get(_:)`, `rename(_:name:)` and the
create call.

The suite decodes `Teammate.Presence` and `Teammate.LastTurn`
(`ResourceWireTests.swift:91`, `nestedTeamTypesAndConnectionsKeepUnknownValues`)
but never `Teammate` itself. `grep -rn "Teammate" sdk/swift/Tests/` returns
nothing.

## Evidence

Swept all 27 decodable top-level contract schemas on `origin/main` (`978efe75`)
in an isolated worktree. For each: add one `required: true` property,
regenerate, run the Swift suite.

- **25 CAUGHT** — an existing decode fixture fails with `keyNotFound`.
- **1 SKIPPED** — `Secret`, whose environment/vault union makes the property
  optional anyway.
- **1 PASSED** — `Teammate`. 119 tests green with a non-`Optional`
  `probeRequired: String` on a type four public methods decode.

Confirmed directly. A payload carrying every field the *pre-change* contract
required, generated from the committed contract:

```json
{"agent":{"id":"x","model":"x","name":"x","runtime":"acp"},"agent_id":"x",
 "conversation":{"id":"x","runtime":"acp","status":"failed"},"name":"x",
 "presence":{"label":"x","state":"asleep"},"unread":false}
```

decoded as `Teammate`:

- against the probed models: `keyNotFound(CodingKeys(stringValue: "probe_required"), codingPath: [])`
- against unprobed `main`: passes

So the whole-response break is real, and the full suite does not see it.

## Why the existing guard does not cover this

`test_optional_compat_pins_reach_a_live_property`
(`scripts/ci/test_swift_generation.py:100`) checks the pin table in one
direction only: every pin must still name a live property, catching a pin left
behind by a rename. It cannot catch the reverse — a newly required property that
needs a pin and has none. That is the direction that breaks consumers.

`payloadsFromAnOlderServerStillDecode` (`ResourceWireTests.swift:157`) is
hand-authored per type and covers `Catalog`, `ConnectionProvider`, `Connection`,
`ApiKey` and `TeamSchedule`. It is a good test; it is just a fixture list, so it
protects the types someone remembered.

## Suggested fix

Derive the check from the contract rather than from fixtures: at generation
time, fail when a property is newly required on a type that has already shipped
and carries no `OPTIONAL_COMPAT` pin, forcing an explicit decision — pin it, or
record that no older server can emit the type without it (the
`CatalogFirstRequest` case the inventory already documents).

A cheaper stopgap that closes today's hole: extend
`payloadsFromAnOlderServerStillDecode` to cover `Teammate`. That fixes `Teammate`
and leaves the rule still unenforced for the next type added.

Found while verifying #2227's completion criteria. Related: #2251 (generation
inventory), #2269 (remaining seams), #2277 (generator holes that also want a
generation-time failure plus a test in `scripts/ci/test_swift_generation.py`).
Belongs under #2247.

BinaryBourbon:
`contributing/swift-wire-models.md` states the rule that makes generated Swift models safe:

> A model that already shipped by hand must not become harder to decode, so `OPTIONAL_COMPAT` keeps two kinds of property optional even where the current server requires them.

Nothing enforced it. The rule was applied by hand when a generator change landed, and the only thing that caught a miss afterwards was whether some committed test happened to decode that type from a payload lacking the key. For 25 of 27 decodable contract schemas one does. For `Teammate` none did — four public `TeamResource` methods return it and `grep -rn "Teammate" sdk/swift/Tests/` was empty — so a required addition there would have reached a release with every gate green, failing every response from an older server **whole**.

## What changes

Generation refuses a property that is non-`Optional` on a type that has already shipped and is either new there or was `Optional` before:

```
A server older than this change could not decode: Teammate.probeRequired is new and
non-Optional on a type that already shipped. Pin each in OPTIONAL_COMPAT and add the
omission that proves it, or record in REQUIRED_BY_CONTRACT that no deployed server omits it.
```

The baseline is the committed `ConversationWire.generated.swift`. That file is already the record of what shipped and `--check` already diffs against it, so the rule needs no second requiredness list to drift out of step (#2251 asks for no new source of field lists). A type absent from it is wholly new and takes contract requiredness directly, which is the rule the inventory already states.

Two recorded decisions clear the failure, and only silence fails: pin the property in `OPTIONAL_COMPAT`, or record it in the new `REQUIRED_BY_CONTRACT` — empty by design, because an entry there is a claim about every server in the field.

The stopgap lands too: `payloadsFromAnOlderServerStillDecode` now decodes `Teammate` from a payload carrying only what the contract requires of it and of the nested `Agent` and `Conversation`, with the pinned `presence.label` omitted to prove that pin as well.

## Verification

Both guards verified by breaking them, not just by passing:

- Disabling the generation check → 3 of 13 generator tests fail.
- Regenerating with a required `probe_required` on `Teammate`, exempted so it reaches the output → the fixture fails with `keyNotFound(CodingKeys(stringValue: "probe_required"))`, the #2284 shape exactly.

The committed contract is clean through the new check, which is also what proves the baseline parser's type names line up (`compatibility_failures` returns `[]` on `main`'s contract, and no generated model is missing from the baseline). Generated output is byte-identical: `generate-swift.py --check` exits 0.

Local: 119 Swift tests, 13 generator tests, `swift format lint --strict`, `check-propagation.py --client swift`.

Closes #2284. Refs #2251, #2247.


jhgaylor (changes requested):
Reviewed exact head `e319a303f305883d1df9936a882031567444a459` against base `b59dbd7b1ef2b7dbcb0d595db08438c2ea869347`.

Two blocking findings:

1. **[medium] The current-head generated output self-authorizes the candidate.** `Generator.render()` reads `OUTPUT` from the current checkout as the supposedly shipped baseline. In the realistic CI case, a PR adds a required field and commits the regenerated `ConversationWire.generated.swift`; `shipped_properties()` then sees that candidate field as already non-Optional, so `compatibility_failures()` returns no error. An exact-head probe with required `Teammate.probeRequired` reported a failure against the base output and `[]` against the regenerated candidate output. This permits the old-server decoding break the guard is intended to stop. Compare against an immutable base/released output, and add a regression whose baseline is an already-regenerated candidate containing the new required field.

2. **[medium] The response-decoding guard also covers input-only models, and its advertised fix can violate the request contract.** Synthetic required fields were flagged on `ConversationCreateRequest`, `AgentUpdate`, `EnvironmentUpdate`, `VaultUpdate`, and `TeamScheduleUpdateRequest`. Applying the suggested `OPTIONAL_COMPAT` pin generated `probeRequired: String? = nil` backed by `.omitted`, so a contract-required request field can be omitted and rejected by the newer server. Scope this response guard to decodable models, or add a separate input-compatibility rule that preserves required request encoding; cover the input roots with regression tests.

Process blocker: the sole commit lacks the `Signed-off-by:` trailer required by `CONTRIBUTING.md`.

CI is green and `git diff --check` passes, but these source-level defects require changes.

BinaryBourbon:
Both findings were real; both are fixed in `38b637dc`. Thanks — finding 1 in particular killed the design, not just the code.

## 1. The candidate self-authorized

Reproduced exactly as described: with the baseline read from the working tree, a required `Teammate.probe_required` plus the regenerated output reports `[]`, because `shipped_properties()` sees the candidate's own non-`Optional` property as already shipped.

The baseline is now **the last release tag**, read with `git describe --tags --abbrev=0`. Two details mattered:

- It cannot be the tag's *generated file* alone. `Teammate`, `Agent`, `Catalog` and the rest shipped **handwritten** at `v0.17.1` and were generated only in #2267/#2270, so a generated-file baseline would call them wholly new and skip exactly the types that just moved. The baseline is therefore every public property of every model under `Models/` at the tag — wherever that model lived then — which needed an indentation-aware parser, since handwritten nested types sit inside their parent (`struct Teammate { struct Presence`) and generated ones sit in an extension. Both now read as the same key: `released_properties()["Teammate.Presence", "label"] == True`.
- `swift-sdk` and `workflow-checks` now check out with `fetch-depth: 0` and `fetch-tags: true`. If the tags are ever missing the generator raises and names the fix rather than skipping the check — silent degradation is the bug class we're closing.

Regression added as asked, `test_a_regenerated_candidate_cannot_authorize_itself`: it builds the candidate output that carries the required field, asserts the guard reports nothing when that candidate is the baseline, and asserts it reports the failure against the released tree.

A useful side effect: the committed contract is **clean** against the `v0.17.1` baseline, which is independent evidence that #2267 and #2270 left no unpinned break.

## 2. The response guard covered input-only models

Also confirmed, on all six input roots — `ImageInput` too, not just the five listed. The guard now skips `self.input_roots`, since an input root is encoded and never decoded, and the advertised pin would let a field the server requires be omitted and the request 422'd instead. `test_input_only_models_answer_to_the_request_contract` covers all six: a required addition renders non-`Optional` and generates **no** `ConversationInputField` storage, so the encoder cannot leave the key out.

## Process

`Signed-off-by:` added. `CONTRIBUTING.md` is right and I missed it; note there's no CI gate for it, which is why the PR was green — worth a separate issue if you want one.

## Verification

Both parts verified by breaking them: dropping the guard call fails 4 tests, dropping the decodable scoping fails 6, and regenerating with an exempted required `probe_required` on `Teammate` fails the fixture with `keyNotFound("probe_required")`. 15 generator tests, 111 CI policy tests, 119 Swift tests, strict format, `--check` clean (generated output unchanged).

jhgaylor (changes requested):
Reviewed exact head 38b637dca7190fee73de8a2a3baf14dd434c847e against base b59dbd7b1ef2b7dbcb0d595db08438c2ea869347.

One blocking finding:

1. **[medium] A server shape newly exposed by the SDK is treated as if older servers cannot emit it.** `compatibility_failures()` skips every type absent from the last released Swift sources (lines 509-512), but absence from that SDK is not evidence that the released server never emitted the JSON shape. Concrete case: the v0.17.1 contract already includes `CatalogResponse.data.mcp_servers` items, while the v0.17.1 Swift `Catalog` did not expose `CatalogMcpServersItem`. On this exact head, adding a required `probe_required` property to that item makes `Generator.render()` succeed and emit `public var probeRequired: String`; `released_properties()` has no owner named `CatalogMcpServersItem`, so the guard skips it. A v0.17.1 server can then return an `mcp_servers` item without the key and the newer SDK fails the whole `Catalog` decode. The existing hand pins cover the current fields, but any required addition before the next SDK release reopens the break. Base the wholly-new decision on whether released servers could emit the decodable shape (or conservatively guard descendants of shipped response roots), and add this nested-item regression.

CI is green. Focused verification passed: 15 Swift-generation tests, generated Swift `--check`, the six `ResourceWireTests`, conflict-marker check, changelog check, and `git diff --check`. The broader 111-test local policy run had four unrelated sandbox-only `/dev/stderr` failures in `test_dead_code`; GitHub CI policy tests passed.

jhgaylor:
Confirmed and fixed in `6ac74e15`. You're right, and the inference I made — "absent from the released SDK ⇒ no older server emits it" — was wrong in exactly the way you describe. Verified your case end to end first:

- `v0.17.1`'s **contract** does describe `CatalogResponse.data.mcp_servers` items, with `dcr`/`name`/`slug`/`url`/`verified_on` all `required: true`.
- `released_properties()` has no owner named `CatalogMcpServersItem` — the v0.17.1 Swift models never exposed it.
- On `38b637dc`, a required `probe_required` on that item emitted `public var probeRequired: String` and the guard stayed silent.

## The fix

The decode question is now asked of the released **contract** rather than the released Swift, since absence from the SDK is not absence from the server:

```
A response from a server older than this change would fail to decode:
CatalogMcpServersItem.probeRequired is required here and the last release could omit it.
```

A property the released contract already required may be non-`Optional`; anything else may not. A shape the released contract never described takes contract requiredness, because no released server can return it — that is the sound version of the test I had.

**`REQUIRED_BY_CONTRACT` is still empty, and that is now a derived fact rather than a claim.** `first_request` arrived whole in #1443, so the v0.17.1 contract requires all four of its members; the guard reads that and allows them non-`Optional` on its own. The exception this file used to state in prose (`CatalogFirstRequest` "takes contract requiredness, because … no server has ever emitted it partially") is now read from the release instead of asserted.

I kept the released-**Swift** baseline as a second, separately-worded rule, because it turns out to carry real weight: **38 of the 55 pins are held by it alone** — properties the released server always sent, so decoding is safe, but which this SDK published as `Optional`. Dropping one of those pins flips a public `x?` to `x`, which the decode rule cannot see:

```
AdminSandbox.status shipped Optional and the contract now requires it.
```

Regressions added: `test_a_shape_the_sdk_never_exposed_is_still_an_older_server_shape` covers both nested Catalog shapes, `test_a_property_the_release_always_sent_may_be_required` pins the `first_request` reasoning, and `test_a_property_this_sdk_published_optional_stays_optional` covers the source rule by dropping a pin.

Two mechanical notes: reading an older contract meant splitting model discovery (`build()`) from rendering, with a `baseline=True` generator that applies no pins, runs no guard, and tolerates roots that contract never had. And `#2288` is rebased onto this head.

## Verification

Each part disabled on its own: decode rule → 5 failures, source rule → 1, whole guard → 6, decodable scoping → 6. The committed contract is clean against both baselines.

Local: 17 generator tests, 113 CI policy tests, 119 Swift tests, `--check` clean (generated output unchanged), strict format, swift propagation check.

On your local `test_dead_code` `/dev/stderr` failures — I see 113 passing here and didn't reproduce those, so I've treated them as environmental as you did.

jhgaylor (changes requested):
Reviewed exact head 6ac74e15a940f9208a247cbd7d7dedfde94437b1 against base b59dbd7b1ef2b7dbcb0d595db08438c2ea869347.

One blocking finding:

1. **[medium] The released-Swift parser drops escaped property names, so a live Optional compatibility pin can still be removed silently.** At scripts/sdk-contract/generate-swift.py:230-258, PROPERTY accepts only an unescaped word immediately after public var. The v0.17.1 Account.swift publishes Catalog.SandboxProviders.default as String? using escaped identifier syntax, but released_properties() returns no entry for either the escaped or plain name. The v0.17.1 contract marks CatalogSandboxProviders.default required. On this exact head, removing (CatalogSandboxProviders, default) from OPTIONAL_COMPAT makes Generator.render() succeed and emit a non-Optional String property. This breaks existing callers using optional semantics and defeats the source-compatibility guard. Parse escaped Swift identifiers and make the normalized name match the generated field lookup; add a regression that removes this pin and requires compatibility_failures() to report it.

The previous CatalogMcpServersItem server-shape blocker is fixed: released_requiredness() includes the item although released_properties() does not, and injecting a new required probe field now raises the expected old-server decode error.

Focused verification passed: 17 generation tests; generated Swift --check; 6 ResourceWireTests; conflict-marker, changelog, and git diff checks. GitHub CI required and substantive checks are green. No Review Loop App summary was present.

jhgaylor:
Confirmed and fixed in `47173e15`. Correct on every point, and it was the only such line in the release.

## Verified first

- v0.17.1 `Account.swift:117` publishes it escaped: `` public var `default`: String? `` inside `struct Catalog { struct SandboxProviders`.
- `released_properties()` returned **no entry** under either spelling.
- The v0.17.1 contract marks `CatalogSandboxProviders.default` required, so the decode rule allows it non-`Optional` and the source rule was the only thing standing between that pin and a silent removal.
- Dropping `("CatalogSandboxProviders", "default")` from `OPTIONAL_COMPAT` on `6ac74e15` rendered `public var `default`: String` with the guard silent.

## Fix

`PROPERTY` accepts an escaped identifier, and the lookup normalizes the generated side, so both spellings meet at the bare name:

```python
PROPERTY = re.compile(r"\s*public var `?(\w+)`?:\s*([^{]+)")
...
elif name in published and shipped.get((name, swift.strip("`")), False):
```

Dropping the pin now reports `Catalog.SandboxProviders.`default` shipped Optional and the contract now requires it`, and that pin is in the regression beside `AdminSandbox.status`.

## The parser hole as a class

Rather than fix the one instance, I audited the whole released tree: **412 `public var` declarations, 411 parsed, exactly one missed** — this line. It is now 412/412 with none missed.

Since the source rule is only as good as that parser, `test_the_released_swift_parser_reads_every_published_property` counts the release's own declarations and asserts the baseline has an entry for each, instead of trusting a list of names. The next unusual spelling fails there rather than turning a live pin into a no-op. That is the same shape as the gap that started #2284 — a rule whose coverage was whatever someone remembered.

## Verification

Each half reverted on its own: the regex → 2 failures, the lookup normalization → 1. Full set: decode rule 5, source rule 1, whole guard 6, decodable scoping 6.

Local: 18 generator tests, 114 CI policy tests, 119 Swift tests, `--check` clean (generated output unchanged), strict format. #2288 rebased onto this head as `fbe092b2`.

jhgaylor (changes requested):
Reviewed exact head `47173e158ed1b78a1ea42b4cacd86da47c1c9b71` against base `b59dbd7b1ef2b7dbcb0d595db08438c2ea869347`.

One blocking finding:

1. **[medium] `REQUIRED_BY_CONTRACT` bypasses the released-Swift source-compatibility rule, not just the older-server decode rule.** In `scripts/sdk-contract/generate-swift.py:568-577`, the exemption is checked before either baseline, so it skips both. Exact-head reproduction: remove `("CatalogSandboxProviders", "default")` from `OPTIONAL_COMPAT` and add that same pair to `REQUIRED_BY_CONTRACT`; `Generator.render()` succeeds and emits `public var \`default\`: String`, even though v0.17.1 published that property as `String?` and `released_properties()` records it Optional. “Every deployed server sends this key” can establish decoding safety, but it cannot make changing an already-public `T?` to `T` source-compatible. The documented escape hatch can therefore authorize the exact escaped-property regression this revision is meant to prevent. Apply `REQUIRED_BY_CONTRACT` only to the released-contract/decode branch and always enforce the released-Swift Optional rule; add a regression that combines pin removal with the exemption.

The requested fixes themselves are verified: the release parser reads 412/412 public properties, plain removal of the escaped `default` pin now raises the expected source-compatibility failure, and a required probe on the released `CatalogMcpServersItem` shape still raises the older-server decode failure.

Focused verification passed: 18 Swift-generation tests, generated Swift `--check`, 6 `ResourceWireTests`, conflict-marker check, changelog check, and `git diff --check`. GitHub CI required and all substantive jobs are green. No authentic Review Loop App summary is present.

jhgaylor:
Confirmed and fixed in `28f5abbf`. The exemption was checked before both rules and the branches were `if`/`elif`, so it skipped the source rule twice over — reproduced exactly as you describe: dropping the escaped `default` pin and adding that same pair to `REQUIRED_BY_CONTRACT` rendered `public var `default`: String`, with v0.17.1 publishing it `String?`.

You put the reason better than my comment did, so I've written it into the code and the inventory: *"every deployed server sends this key" can establish that decoding is safe, but it cannot make an already-public `T?` becoming `T` source-compatible.* The table was an answer to a question about servers, and it was silently answering a question about callers.

## Fix

The two rules are now answered independently, and `REQUIRED_BY_CONTRACT` gates only the decode branch:

```python
if (owner in emitted and not always_sent.get((owner, key), False)
        and (owner, key) not in REQUIRED_BY_CONTRACT):
    failures.append(f"{name}.{swift} is required here and the last release could omit it")
if name in published and shipped.get((name, swift.strip("`")), False):
    failures.append(f"{name}.{swift} shipped Optional and the contract now requires it")
```

Two consequences worth stating explicitly:

- **The source rule has no override at all.** I considered adding a second table for a deliberate, versioned drop of a published Optional and decided against it on the evidence of this thread: an unused door written for a hypothetical is what just got misused. The comment and the inventory now say that a change which really means to drop a published Optional should add its own door and justify it, rather than borrow an escape hatch written for a different question. Happy to add one now if you'd rather it exist up front.
- **A property can owe both rules and now reports both**, since the remedies differ — `test_a_property_can_owe_both_rules` covers that with `Agent.description`, which trips the decode rule *and* the source rule.

The exemption still answers the rule it was written for: a required property no release ever published (the `Teammate` probe) is still cleared by it.

## Verification

Both halves reverted independently: restoring the early `continue` → 1 failure; restoring the `elif` → 1 failure. Full set: decode rule 5, source rule 1, whole guard 6, decodable scoping 6, escaped-name handling 1–2 per half, rule separation 1 per half.

Local: 20 generator tests, 116 CI policy tests, 119 Swift tests, `--check` clean (generated output unchanged), strict format, changelog check. #2288 rebased onto this head as `325b28fc` — content unchanged from the revision you approved, so it needs only a re-approval of the new sha.

jhgaylor (approved):
Reviewed exact head `28f5abbfce7abf9549e23f98b88d8aadf786fb37` against approved base `b59dbd7b1ef2b7dbcb0d595db08438c2ea869347`, including the full PR and the incremental delta from `47173e158ed1b78a1ea42b4cacd86da47c1c9b71`.

No blocking findings.

The prior blocker is fixed: `REQUIRED_BY_CONTRACT` now gates only the released-contract/decode rule, while the released-Swift source rule runs independently. I reproduced the original bypass by removing the escaped `CatalogSandboxProviders.default` pin and exempting the same pair; generation now still fails with ``Catalog.SandboxProviders.`default` shipped Optional``. A property that owes both rules also retains the source failure after the decode exemption.

The earlier fixes remain intact: the v0.17.1 parser reads all 412/412 public properties and records escaped `default` as Optional, and a required probe under the released `CatalogMcpServersItem` contract shape is rejected even though that owner was absent from the released Swift SDK. Input-only roots remain required and non-omissible rather than being forced through the response-compatibility pin path. The single commit has the required DCO sign-off.

Non-blocking follow-up: the aggregate `ValueError` prefix/remediation and changelog still imply every failure is an older-server decode failure that `REQUIRED_BY_CONTRACT` can clear. For a source-only failure that is intentionally false—the exemption does not and should not clear it. Splitting that diagnostic wording would make the new rule separation easier to operate, but the guard fails safely, so this does not warrant withholding approval.

Verification: 20 focused Swift-generation tests; generated Swift `--check`; strict Swift format; all 119 Swift tests (including 6 `ResourceWireTests`); conflict-marker, changelog, and `git diff --check`. The broader local CI-policy run reached all 116 tests with only the known sandbox-only `/dev/stderr` failures in `test_dead_code`; GitHub's CI-policy job passed. All current substantive GitHub checks are green, including both Swift platforms, coverage, six test partitions, release/contract, secret scanning, and `CI required`. No authentic Review Loop App summary is present.

PR #2288, opened 2026-09-15
fix(sdk): fail Swift generation on shapes that retype a field in silence

### LINKED ISSUE #2277: Swift generator: inline-name collisions and additionalProperties silently retype fields
Found while reviewing #2270. Both holes are byte-identical to #2267 and are not reachable on today's contract, but #2270 enlarges the surface they sit on.

## Inline name collision silently retypes a field

`render()` resolves `self.schemas` before the reuse guard checks `self.nested`, so an inline object whose synthesized `<Owner><Key>` name collides with a real schema is discarded and the field takes the unrelated schema's type. Confirmed by making `Connection.provider` an inline `{slug, name}`: it types as `ConnectionProvider` with no error. The synthesized-name ∩ schema-name intersection is empty today.

## `additionalProperties` wins over sibling `properties`

`type()` returns `[String: …]` whenever `additionalProperties` is present, ignoring declared `properties` — the opposite of what the module docstring promises for unknown shapes. The contract's only both-shaped node is `networking_config`, which a `TYPE_OVERRIDES` entry neutralizes.

## Minor

- Array-of-enum loses typing: `Catalog.sandbox_api_access` generates `[String]?` while the scalar `Conversation.sandbox_api_access` is `SandboxAPIAccess`.
- `REUSED` is initialized empty and never populated, so `if ref not in REUSED` in `reference()` is dead code.

Each wants a generation-time failure or an explicit override, plus a test in `scripts/ci/test_swift_generation.py`.

BinaryBourbon:
**Stacked on #2287** (base `sdk/guard-optional-compat-direction`) — both PRs edit `generate-swift.py` and its test. Retarget to `main` after #2287 merges; do not arm auto-merge before that.

Found while reviewing #2270. None of the three holes is reachable on today's contract; #2270 enlarged the surface they sit on.

## Inline name collision silently retyped a field

`render()` resolves `self.schemas` before the reuse guard consults `self.nested`, so an inline object whose synthesized `<Owner><Key>` name collided with a real schema was discarded and the field took the unrelated schema's type. Confirmed by making `Connection.provider` an inline `{slug, name}`: it typed as `ConnectionProvider`, no error. The synthesized-name ∩ schema-name intersection is empty today.

## `additionalProperties` won over sibling `properties`

`type()` returned `[String: …]` whenever `additionalProperties` was present, ignoring declared `properties` — the opposite of what the module docstring promises for unknown shapes. The contract's only both-shaped nodes are the two `networking_config`, which `TYPE_OVERRIDES` already answers.

## Array-of-enum lost its typing

An array's items are typed under `<key>_item`, which nothing registered, so `Catalog.sandbox_api_access` generated `[String]?` while the same wire field on `Conversation` generated `SandboxAPIAccess?`.

All three now raise at generation time and name the table that answers them. The `Catalog` item type is registered, which is the **one output change** in this PR and the only source break:

```diff
-  public var sandboxAPIAccess: [String]?
+  public var sandboxAPIAccess: [SandboxAPIAccess]?
```

An element no longer compares to a bare `String`; a string literal still works, since `SandboxAPIAccess` is `ExpressibleByStringLiteral`. Recorded under `### Changed`.

## Minor

`REUSED` was initialized empty and never populated, so `if ref not in REUSED` in `reference()` was dead code. Both deleted.

## Verification

Each guard disabled on its own → exactly one test failure each, so no test passes for the wrong reason. The committed contract stays clean through all three, and generated output is otherwise byte-identical.

Local: 16 generator tests, 119 Swift tests, `swift build --configuration release -Xswiftc -warnings-as-errors`, `swift format lint --strict`, `check-propagation.py --client swift`.

Closes #2277. Refs #2251, #2247.


jhgaylor (changes requested):
Requesting changes on `3d4325ac4cd89b8cefcc4f4498e5ae1291e1dd90` for two blockers:

1. **Medium — the enum-array guard still silently loses typing for supported composition shapes.** `scripts/sdk-contract/generate-swift.py:298` checks only direct `items.enum`, but `type()` already unwraps single-branch `allOf`, `anyOf`, and `oneOf` at lines 285–291. An `Agent.probe_kinds` array whose items are `allOf: [{type: string, enum: [one, two]}]` therefore succeeds and emits `public var probeKinds: [String]?` instead of requiring an `ENUM_TYPES` decision; the same happens through the supported `anyOf`/`oneOf` form. Normalize or unwrap the item before the enum check (or carry enum metadata through type resolution), and add composition-wrapped regression coverage beside the direct-enum test.

2. **Process — the sole commit is missing the required DCO sign-off.** `3d4325ac` has no `Signed-off-by:` trailer, contrary to the approved-base `CONTRIBUTING.md` requirement. Please amend/sign off the commit.

BinaryBourbon:
Both fixed in `35d94bde` (rebased onto #2287's amended head).

## 1. The enum-array guard missed composition-wrapped items

Correct, and the underlying problem was that I had written the peeling rule twice. There is now one definition — `unwrap()` — and **`type()` uses it too**, so the check and the resolution cannot drift apart:

```python
peeled = unwrap(node)
if peeled is not node:
    return self.type(owner, key, peeled)
```

The array check reads `unwrap(node["items"])`. All six wrapper shapes are covered by subtests beside the direct-enum case:

| items | before | now |
|---|---|---|
| `{type: string, enum: […]}` | raises | raises |
| `allOf: [enum]` | **`[String]?`** | raises |
| `anyOf: [enum]` | **`[String]?`** | raises |
| `oneOf: [enum]` | **`[String]?`** | raises |
| `anyOf: [enum, None]` | **`[String]?`** | raises |
| `anyOf: [allOf: [enum], None]` | **`[String]?`** | raises |

Reverting just the `unwrap()` call fails five of the six subtests and leaves the direct one passing, which is the shape I'd expect if the fix is doing what it claims.

`test_unknown_shapes_fail_loudly` still passes: a genuine two-branch `oneOf` is not peeled, so `type()` still raises `Unsupported union`.

## 2. DCO

`Signed-off-by:` added.

Local: 18 generator tests, 114 CI policy tests, 119 Swift tests, `swift build --configuration release -Xswiftc -warnings-as-errors`, strict format, `check-propagation.py --client swift`. The only generated-output change is still the one line: `Catalog.sandboxAPIAccess: [String]? → [SandboxAPIAccess]?`.

jhgaylor (approved):
Reviewed the full 35d94bde diff against stacked base 38b637dc, including the approved-base policy/instructions and #2287 interaction. The prior blockers are resolved: unwrap() is now the single peeling rule used by both type resolution and the enum-array guard, with six wrapper regressions, and the commit carries the required DCO sign-off. The collision and mixed object-shape guards fail loudly with focused coverage, while the generated model delta is limited to the documented Catalog sandboxAPIAccess typed-array correction; SandboxAPIAccess retains unknown-string decoding. All substantive CI jobs, including both Swift jobs and CI required, are green, and the 18 focused generator tests pass independently. No blocking correctness, compatibility, security, or maintainability defects found. This stacked PR still depends on #2287 landing and then being retargeted.

jhgaylor:
Rebased onto #2287's new head (`6ac74e15`) as `d7c5f4cd`; no content change beyond one comment that went stale in the rebase — the inline-collision note now says `build()` rather than `render()`, since #2287 split model discovery out of `render()`.

Re-verified after the rebase: 20 generator tests, 116 CI policy tests, 119 Swift tests, `--check` clean, strict format. Both PRs' guards still fire — `allOf`-wrapped enum array raises `Untyped enum array`, and a required addition to the `mcp_servers` item raises the new decode failure from #2287.

jhgaylor (approved):
Reviewed the full five-file `d7c5f4cd8fcfcd4edc890f919f7fed968b6e8679` diff against stacked base `6ac74e15a940f9208a247cbd7d7dedfde94437b1`, including the approved-base Review Loop policy, prior objections, and #2287 interaction. The rebase is patch-equivalent to the previously approved revision except for the accurate `build()` comment. The shared `unwrap()` path covers every composition form that `type()` accepts, the schema/inline collision and mixed object-shape guards fail loudly, and the only generated API delta is the documented `Catalog.sandboxAPIAccess` typed-array correction with unknown-string decoding preserved.

All required GitHub evidence is green, including CI required, both Swift jobs, coverage, contract/release, and secret scanning. Independent exact-head checks passed: 20 generator tests, all 119 Swift tests with warnings as errors, generation `--check`, changelog validation, conflict-marker scan, and `git diff --check`. No blocking correctness, compatibility, security, or maintainability defect found. This remains stacked on #2287 and must be retargeted after it lands.

Issue #1393, opened 2026-09-02
Drop users.onboarding_state; onboarding_completed_at is the source of truth (NC-6)


jhgaylor:
Implementation issue from #1039 / ADR 0038; NC-6 from the launch backlog (ADR 0007).

`users.onboarding_completed_at` and `users.onboarding_state` both exist and `Accounts.complete_onboarding/…` sets both. `onboarding_state` only ever distinguishes `step_1` from `completed` now that the wizard is gone (#867); `Funnel.stalled_breakdown/2` reads it as a vestige.

## Do

- Keep `onboarding_completed_at` as the one source of truth; stamped at first reply once the activation issue lands (until then, where it is stamped today).
- Drop `onboarding_state`: the column (migration), the schema field, `complete_onboarding`'s write, the funnel's `by_onboarding_state` breakdown, and any API or UI that shows it (grep `onboarding_state` across `apps/`, `ee/`, `sdk/`, `cli/`, `docs/`; the SDK's generated types will change, which is an SDK release).
- `Accounts` audit for `onboarding.completed` (if there is one) is unchanged.

## Acceptance

- No `onboarding_state` anywhere outside the migration that drops it and the changelog.
- `funnel_test.exs` and `accounts_test.exs` green; OpenAPI spec regenerated; SDK types match (CI's "SDK types match the spec" step).

PR #2289, opened 2026-09-15
refactor(swift): generate AuthMe and retire onboardingState

jhgaylor:
First of the four seams in #2269. `AuthMe` was handwritten because of one property the contract does not describe, so the whole model stayed outside generation for it.

## `onboardingState` is retired, not preserved

| | |
|---|---|
| Server dropped `users.onboarding_state` | **v0.16.0** (#1393, ADR 0038, settling NC-6 from ADR 0007) |
| TypeScript SDK dropped the field | its **1.17.0** |
| Swift still exposed it | until now |

So this finishes #1393 in the last client that carried it, rather than making a fresh call. The property has decoded `nil` from every reachable server for two releases. `onboardingCompleted` replaces it; nothing replaces a part-way step, because the server records none.

A server old enough to still send the key decodes fine — an unknown key is ignored — and the older-server fixture includes `onboarding_state` to prove precisely that.

Everything else about the public API is unchanged: same names, types and optionality, plus an additive `Identifiable` conformance. `Models/Account.swift` is deleted.

## The guard from #2287 did the work

This is its first real migration, and adding the root failed generation on its own, naming both properties that needed pinning before I could look them up:

```
A response from a server older than this change would fail to decode:
AuthMe.emailVerified shipped Optional and the contract now requires it;
AuthMe.role shipped Optional and the contract now requires it.
```

Both are pinned with the omission that proves each — an older-server payload carrying only `id` and `email`. Worth noting the two rules behaved exactly as intended and differently from each other: the released contract requires both keys, so the **decode** rule was satisfied and it was the **source** rule that fired, because every published `AuthMe` had them `Optional`.

Each pin verified twice over:
- Dropping it fails generation outright. The source rule has no override, so not even `REQUIRED_BY_CONTRACT` forces it through.
- With the guard disabled so the break reaches the output, the fixture fails with `keyNotFound(CodingKeys(stringValue: "role"))` / `"email_verified"` respectively.

Additive propagation comes free: `AuthMeResponse` joins `RESOURCE_ROOTS`, so the existing per-root probe covers it.

## Also

`authMeIsUnenveloped` no longer sends `onboarding_state` in its fixture — it claims to be the verified live shape, and the live shape has not had that key since v0.16.0.

Local: 119 Swift tests, 119 CI policy tests, `swift build --configuration release -Xswiftc -warnings-as-errors`, strict format, generation `--check`, `check-propagation.py --client swift`.

Refs #2269, #2251, #2247.


BinaryBourbon (approved):
Reviewed the complete diff at `37628a0292a4f35cf2f28dd0ea2015eef6c53b94` adversarially against base `ec2a3c204dc551a7390ac2e248738e6192b0be74`.

The generated `AuthMe` preserves the shipped names and optionality for `role` and `emailVerified`, retains the live contract's required `id`/`email`, and safely ignores the retired `onboarding_state` key from older servers. The older-server fixture proves the two compatibility pins and the unknown-key path. The controller, OpenAPI schema, committed contract, generated Swift model, client call site, and ADR 0038 are consistent. I found no tenant-isolation or credential-boundary change and no blocking correctness, API-compatibility, or maintainability defect.

Verification: all reported CI checks for this head are green (including both Swift platforms, generated contract/release, CI policy, coverage, and secret scanning). Locally on the exact head: `swift test` (119 passed), generator policy tests (23 passed), `generate-swift.py --check`, and Swift additive propagation all passed.

Review Loop did not publish a run summary on this PR; the dispatch job was skipped, and the PR had no existing issue comments or reviews at review time.

PR #2292, opened 2026-09-15
refactor(swift): derive AdminUserPage from the generated envelope

jhgaylor:
Second of the four seams in #2269, and the decisions for all four are recorded in [this comment](https://github.com/managoat/fountain/issues/2269#issuecomment-5684176985).

`/api/admin/users` is the one endpoint with page-number pagination rather than a cursor, and `AdminUserPage` carried the whole envelope by hand: `data` and `meta` as `CodingKeys`, `page`/`per_page`/`total` as a second `MetaKeys` enum, and a decoder reading both.

`AdminUserListResponse` and its nested `Meta` are generated now, and `AdminUserPage` decodes that:

```swift
public init(from decoder: any Decoder) throws {
  let response = try AdminUserListResponse(from: decoder)
  users = response.data
  page = response.meta.page
  perPage = response.meta.perPage
  total = response.meta.total
}
```

The only wire key left in the handwritten file is the sentence of prose explaining what `meta` holds. Public surface untouched — `users`, `page`, `perPage`, `total`, `hasMore` — so the roster convenience the issue asked to retain is retained, and nothing declares the page keys twice.

## The existing fixture could not prove the mapping

It decoded this page already, but only asserted `hasMore`. With `1 / 10 / 11`, swapping `page` and `perPage` gives `10 * 1 < 11`, which is still `true` — so the convenience passed while the numbers were wrong. Since that mapping is exactly what this PR moves, the fixture now names each number and adds a second page past the end:

```swift
#expect(page.page == 1 && page.perPage == 10 && page.total == 11)
...
let lastPage = try decode(AdminUserPage.self, #"{"data":[],"meta":{"page":2,"per_page":10,"total":11}}"#)
#expect(lastPage.hasMore == false && lastPage.users.isEmpty)
```

Verified by swapping the two assignments: **4 issues**, where the old assertion stayed green.

The guard from #2287 is clean here without pins: `AdminUserListResponse` is absent from the released Swift, so the source rule has nothing to preserve, and the released contract already required `data` and `meta` and all three meta members, so the decode rule is satisfied. Additive propagation comes free, since the root joins `RESOURCE_ROOTS`.

Local: 119 Swift tests, 119 CI policy tests, `swift build --configuration release -Xswiftc -warnings-as-errors`, strict format, generation `--check`, `check-propagation.py --client swift`.

Refs #2269, #2251, #2247.


BinaryBourbon (approved):
Adversarial review of exact head `ac1406c02293505abbe2b773dd98d59dec4a8cab` against base `432755bc67ca6f4edae98aa7584ab71665167b57`: no blocking findings.

The handwritten `AdminUserPage` keeps its published `users`, `page`, `perPage`, `total`, and `hasMore` surface while delegating the same required `data`/`meta` decode to the generated envelope. `AdminUserListResponse.Meta` maps `page`, `per_page`, and `total` exactly as the released and current contracts require. The v0.17.1 baseline confirms every new non-Optional envelope member was already required on the wire, and a nested-field probe confirms future envelope additions propagate without another registry entry.

Coverage: `changelog.d/2269-swift-admin-user-page.md`, `contributing/swift-wire-models.md`, `scripts/sdk-contract/generate-swift.py`, `sdk/swift/Sources/FountainKit/Models/Admin.swift`, `sdk/swift/Sources/FountainKit/Models/ConversationWire.generated.swift`, and `sdk/swift/Tests/FountainKitTests/ResourceWireTests.swift`, plus the controller/schema, `AdminResource`, `APIClient`, existing admin tests, the released Swift/contract baseline, ADRs 0041/0056, and the decision recorded on #2269.

Local evidence: deterministic generation check passed; 23 Swift-generation policy tests passed; Swift propagation passed; 6 `ResourceWireTests` and 5 `AdminTests` passed; conflict-marker, changelog, and diff checks passed. GitHub CI is green, including both Swift platforms, contract/release, all six server partitions, coverage, static analysis, and `CI required`.

There is no Review Loop App-authored summary, comment, or review on this PR; its dispatch check is skipped, so there is no authoritative run or finding ID to cite.

PR #2293, opened 2026-09-15
test(swift): derive the reapply request's fields from the contract

jhgaylor:
Third of the four seams in #2269 — and the one where the right answer turned out to be a test rather than a migration. Decisions for all four are recorded in [this comment](https://github.com/managoat/fountain/issues/2269#issuecomment-5684176985).

## Why the type stays handwritten

`ConversationReapplyRequest` keeps its own encoder because the three binding states are behaviour a generated `Optional` cannot express: "leave it alone" and "remove it" are different requests, and only a hand-written encoder can send an omitted key for one and a JSON `null` for the other. Routing it through generated `CodingKeys` would buy nothing and put that distinction at risk.

The three states were already covered by `encodesTheThreeBindingStatesOnReapply`, which this PR leaves alone.

## What was actually missing

The other half of what the issue asks: *deriving the supported request fields from the contract*. A property added to the contract's request needs a case of its own in the encoder, and nothing said so — the request would simply never send it, with every gate green. Same shape as the `Teammate` gap in #2284: the coverage existed, but not in the direction that breaks.

The test asks that of behaviour rather than of a declaration — encode with every state populated, then compare the keys that actually reach the wire against the contract, in **both** directions, so an unsendable field and an invented one both fail:

```swift
let declared = Set(try #require(schema["properties"] as? [String: Any]).keys)
let request = ConversationReapplyRequest(agentID: "a-1", environment: .use("e-1"), vault: .use("v-1"))
...
#expect(Set(wire.keys) == declared)
```

It then clears both nullable bindings and asserts the same keys arrive as JSON `null`, which is the distinction the type exists for.

## Verification

Added a fourth property to the contract's request and ran it:

```
✘ Expectation failed: (Set(wire.keys) → ["agent_id", "environment_id", "vault_id"])
  == (declared → ["environment_id", "probe_binding_id", "vault_id", "agent_id"])
✘ Expectation failed: (Set(clearedWire.keys) → ["environment_id", "vault_id"])
  == (declared.subtracting(["agent_id"]) → ["environment_id", "vault_id", "probe_binding_id"])
```

Two failures naming the field the encoder does not know. Contract restored from a copy afterwards, not `git checkout`.

No changelog fragment: nothing user-observable changes, and `changelog.d/README.md` scopes fragments to observable changes.

Local: 120 Swift tests, 119 CI policy tests, strict format, generation `--check`.

Refs #2269, #2251, #2247.


BinaryBourbon (approved):
No blocking findings at 17ec8e2e09d16e5e602c85ba6f89904f8ea8645a. The guard compares the handwritten request's actual encoded keys with ConversationReapplyRequest in the committed contract, catching additions, removals, renames, and invented fields while preserving the existing omitted-versus-null binding semantics. I also traced the schema, resource call, encoder, manifest, linked compatibility decision, and generator boundary. All required GitHub checks are green. Exact-head validation passed: 120 Swift tests with warnings as errors, the focused reapply contract test, 23 Swift-generation policy tests, generator --check, diff --check, conflict-marker check, and changelog validation.

PR #2295, opened 2026-09-15
refactor(swift): generate LogEvent, and close out the wire-model inventory

### LINKED ISSUE #2251: Swift SDK: eliminate remaining duplicated wire models in bounded batches

### LINKED ISSUE #2269: Swift SDK: finish legacy and behavioral wire-model seams

BinaryBourbon:
Last of the four seams in #2269, and the end of #2251's inventory. Decisions for all four seams are recorded in [this comment](https://github.com/managoat/fountain/issues/2269#issuecomment-5684176985).

## `LogEvent` is generated

Published API unchanged: the same twelve properties with the same names and optionality, `stageData` unchanged, `Identifiable` as before. Three details carried it across.

**`id` and `ts` are pinned — the first pins load-bearing for a *live* shape rather than an old server.** One published `LogEvent` decodes two shapes: the REST row from `GET /api/conversations/:id/events` carries both, while the SSE frame carries no `id` at all (it is the frame's own `id:` line, not a JSON key) and the team stream omits `duration_ms` as well.

**`blocks` takes a `TYPE_OVERRIDES` entry naming `[Block]`**, which stops the walk entering the `Block` schema while keeping the published element type. `Block` cannot be generated: its `body` is a two-branch `oneOf` feeding two published properties (`body` and `planEntries`), and it is one of the contract's two nodes carrying both `additionalProperties` and `properties` — shapes #2277 deliberately made hard failures.

**`conversation_id` and `agent_id` come from a new `EXTRA_PROPERTIES` table.** A Swift extension can add a computed property to a generated type but never a stored one — which is exactly why `stageData` survives as an extension and these two could not. Both are SSE-frame fields with no schema to read from: all three stream operations declare `text/event-stream` with a bare string schema, so the frame shape lives in prose.

The table is the registry #2251 warns about, so it is bounded the way this repo already bounds one:

```python
ceiling = 2
self.assertLessEqual(len(swiftgen.EXTRA_PROPERTIES), ceiling)
```

Same ratchet as `SchemaGuardAllowlist` — it may shrink freely, and growing it means editing the number in the same diff where a reviewer sees it move. A second test fails any entry whose key the contract has since described, because at that point the ordinary path generates the property and the entry is duplication. Describing the SSE frame properly would retire both entries, and the inventory says so.

**`camel()` learns `ms` → `MS`** so `duration_ms` stays `durationMS`. Nothing generated ended in `Ms` before, so no property is renamed. Worth stating why that mattered: a rename is a source break *neither* compatibility rule can see, since both are keyed by the property name.

## Three types stay handwritten, with reasons

| Type | Why |
|---|---|
| `Block` | two-branch `oneOf` body feeding two properties; `additionalProperties` + `properties`; open-shaped `extra` no schema records |
| `PermissionOption` | **no contract schema at all** — `Block.options` items are open objects, so generation would retype it `[[String: JSONValue]]`; also accepts `optionId` *and* `option_id` |
| `PermissionRequest` | not a wire model: derived from a `Block` via `init?(block:)`, never decoded |

## Verification

Each piece reverted on its own: dropping the extras injection fails 1 test, the `ms` acronym 1, the `blocks` override 5 tests and 49 errors, the `id`/`ts` pins 50 errors.

Local: 120 Swift tests, 122 CI policy tests, `swift build --configuration release -Xswiftc -warnings-as-errors`, strict format, generation `--check`, `check-propagation.py --client swift`.

## Close-out

The inventory now records a disposition for every family, with no "remaining migration" rows left, and the acceptance criteria on both issues are ticked.

Closes #2269. Closes #2251. Refs #2247.


jhgaylor (approved):
Adversarial review of `576381ee2f25162c728c349eda006445f79544d3`: approved.

I traced the generated `LogEvent` through the REST events page and all three SSE payload variants, compared it with the v0.17.1 public Swift surface, and checked the inventory dispositions for `Block`, `PermissionOption`, and `PermissionRequest`. The twelve published properties retain their names, types, and optionality; `id`/`ts` remain permissive for live SSE frames; the two contract-undescribed routing fields are bounded and tested; and the custom block/permission decoding behavior remains handwritten.

No blocking correctness, compatibility, security, or maintainability findings. Exact-head CI and secret scanning are green. Local verification passed: generation freshness, 26 focused generator tests, 121 Swift tests with warnings as errors, and the full 122-test CI tooling suite.

PR #2301, opened 2026-09-15
docs(contributing): reattach the pre-guard history to the paragraph it describes

### LINKED ISSUE #2299: Contributor doc: a stranded sentence in the Swift compatibility rules
`contributing/swift-wire-models.md` reads, in the compatibility-rules section:

> The two rules below are answered independently and a property can owe both, because the remedies differ. `REQUIRED_BY_CONTRACT` reaches only the decode rule: … rather than borrow an escape hatch written for a different question. **Before this guard the rule was applied by hand and a miss was caught only if some fixture happened to decode that type from a payload lacking the key. For 25 of 27 decodable schemas one did; `Teammate` … would have shipped (#2284).**

The bolded tail belongs to the paragraph two above it, which introduces the guard. It describes what life was like *before the guard existed*, and it currently sits at the end of a paragraph about the two rules being independent and about `REQUIRED_BY_CONTRACT`'s scope — where it reads as though it is qualifying the escape hatch.

My doing: I inserted the "two rules" paragraph during #2287's review rounds and left that sentence stranded behind it.

The fix is to move those three sentences back onto the paragraph that ends "…which is why it is empty." No content changes; the section is otherwise accurate.

BinaryBourbon:
## Summary

- Moves the three "before this guard" sentences in `contributing/swift-wire-models.md`'s Compatibility rules section back onto the paragraph they describe (the one introducing the guard, ending "...which is why it is empty."), instead of trailing the unrelated "two rules are answered independently" paragraph.
- Re-wraps both affected paragraphs to the file's existing line width. No wording changes; only the location of the three sentences moved.

Closes #2299

## Verification

- `git diff --word-diff` on the changed file shows the same words removed from one paragraph and added to the other, confirming no content was changed beyond re-wrapping.
- Ran `python3 scripts/conflict-markers.py` — passed ("tracked text files are clean").
- Ran `python3 scripts/changelog.py check` — passed (existing fragments, none required for this contributor-only doc change).
- Ran `git diff --check` — passed, no whitespace errors.
- Did not run the Elixir test suite or `mix precommit`: this is a contributor-only Markdown file outside `docs/`, so per CLAUDE.md/CONTRIBUTING.md it needs no Elixir suite and no changelog fragment.


jhgaylor (approved):
Adversarial review of `31276c3f877c628d00ea1da40912114706e70234`: **approved**. No blocking findings.

Reviewed the complete one-file diff and surrounding compatibility guidance against #2299. The historical passage is preserved verbatim apart from wrapping and now follows the guard introduction it describes. Removing that passage from both revisions leaves identical whitespace-normalized text; the independent-rules explanation retains its meaning.

Validation: conflict-marker and changelog checks passed, and the complete PR diff passes `git diff --check`. Latest applicable CI checks are successful. No runtime suite was run locally, consistent with the approved-base contributor-only Markdown procedure.

PR #2266, opened 2026-09-15
chore(release): v0.17.1

jhgaylor:
Version bump for v0.17.1: both `mix.exs` files, the two Swift version sources, the six self-host image pins, and the changelog roll (`[Unreleased]` plus every `changelog.d/` fragment into `## [0.17.1]`).

Merging this PR tags the squash commit `v0.17.1` and dispatches `release.yml`, which builds the CLI binaries, publishes `ghcr.io/managoat/fountain:v0.17.1`, boots it through the documented quick start and creates the GitHub Release.

"Compose boots the pinned image" skips on this PR because the pinned image does not exist until the release runs; the same check runs on the tag instead.

PR #2302, opened 2026-09-15
ci: refuse a PR commit without a DCO sign-off, and sign the release bump

### LINKED ISSUE #2290: Nothing enforces the DCO sign-off CONTRIBUTING requires
`CONTRIBUTING.md` step 2 is "[Sign off commits](#sign-your-commits-dco) with `git commit -s`", and the sign-off records agreement to the inbound terms. No workflow checks it, so a PR with no `Signed-off-by:` trailer passes every gate and the miss surfaces only if a reviewer reads the trailers.

## Evidence

#2287 and #2288 were both flagged in review for exactly this, each with 25/25 checks green, costing a force-push cycle apiece. Neither is unusual: the trailer is invisible in the PR UI unless you look at the commit.

## What a gate would and would not break

Checked the last 25 commits on `main` and every Dependabot PR:

- **Dependabot is fine.** `d8956d6c`, `e2d02380` and `b17c0eb5` all carry `Signed-off-by:`, so a strict check does not break the dependency flow.
- **The release bump is the one real exception.** `7412eaf1 chore(release): v0.17.1 (#2266)` has no trailer, because `.github/workflows/release-bump.yml:197` commits with plain `git commit -m "chore(release): ${TAG}"`. Adding `-s` there (with the bot identity the workflow already configures) is part of this change, or the gate red-flags every release PR.

That is the whole exception list for the sampled range — 24 of 25 commits already comply, so the gate would be green on today's history apart from the release path.

## Suggested shape

A step in `workflow-checks` (which already has the PR's history for the changelog guard) that walks the PR's own commits and fails any without a `Signed-off-by:` trailer. Worth deciding two things while writing it:

- **Whose name must the trailer match?** The DCO convention is the commit author. A stricter "must equal the author" check catches a copy-pasted trailer; a looser "any trailer present" check is easier to satisfy from a worktree where `user.email` differs from the sign-off identity.
- **Merge commits.** Skip them, or the first merge from `main` into a long-lived branch fails a check that has nothing to do with the contribution.

An alternative is GitHub's DCO app, which block/buzz uses and which reports as its own required check rather than as repo-owned code. Either is fine; the repo-owned version keeps the policy in the tree with the rest of `scripts/ci`.

## Why it is P3

Nothing ships wrong without this — it costs review attention and a force-push per miss, which is annoying rather than dangerous. Found while landing #2287 and #2288 for #2251.

BinaryBourbon:
## Summary

- CONTRIBUTING.md's DCO step asks for `git commit -s`, but nothing checked it: a PR with no `Signed-off-by:` trailer passed every gate, and the miss surfaced only when a reviewer read the raw commit messages. #2287 and #2288 each cost a force-push cycle this way, both with 25/25 checks green.
- Adds `scripts/ci/dco.py`, wired into the `workflow-checks` job right after "Resolve the PR's base" (`.github/workflows/ci.yml`), and adds `-s` to the release-bump workflow's version-bump commit — the one commit in a 25-commit sample of `main` that didn't already comply.
- Adds `scripts/ci/test_dco.py`, documents the gate in `scripts/ci/README.md` and `CONTRIBUTING.md`'s DCO section.

## Two decisions (settled on #2290, and in the script's docstring)

- **Trailer presence, not author match.** The gate fails a commit for lacking a `Signed-off-by:` trailer; it does not require that trailer's identity to equal the commit author. Contributors commit from worktrees and automation where `user.email` differs from the sign-off identity, and the sign-off is the statement of agreement to the inbound terms, not an authorship claim.
- **Skip merge commits.** A merge from `main` into a long-lived branch is not itself a contribution, so the script walks `git rev-list --no-merges base..head`.

## What changed

- `scripts/ci/dco.py`: given `--base`/`--head` (or `$PR_BASE_SHA`/`$HEAD`), lists the PR's non-merge commits and fails, naming each by short SHA and subject, any missing a `^Signed-off-by: .+` line. Exits 0 printing how many commits it judged when all comply.
- `.github/workflows/ci.yml`: a new `workflow-checks` step, "Every commit carries a DCO sign-off", `if: github.event_name == 'pull_request'`, running `python3 scripts/ci/dco.py --base "${PR_BASE_SHA}" --head HEAD`. HEAD there is GitHub's merge commit of the PR onto its base, so `base..HEAD` with `--no-merges` is exactly the PR's own commits.
- `.github/workflows/release-bump.yml`: added `-s` to the `chore(release): ${TAG}` commit (line ~197) under the bot identity the workflow already configures, so release PRs pass the new gate. The other `git` write further down (~line 283) is `git tag -a`, not a commit, so it produces no PR commits and needs no change.
- `scripts/ci/README.md`: mentions the DCO gate in the `workflow-checks` row.
- `CONTRIBUTING.md`: a line in "Sign your commits (DCO)" noting CI now refuses a PR commit without the trailer, and how to fix it (`git commit --amend -s`, `git rebase --signoff`).
- No changelog fragment: this is contributor/CI tooling, not something a user or operator observes (`changelog.d/README.md`).

## Verification

- `python3 -m unittest discover -s scripts/ci -p 'test_*.py' -v` — all 127 tests pass, including the 5 new `test_dco.py` cases (all signed passes, one unsigned fails and is named, a merge commit is skipped, an empty range passes, a mismatched-identity trailer passes).
- `python3 scripts/ci/dco.py --base origin/main --head HEAD` against this branch's own commit — passes (`1 commit judged, all signed off`).
- Proved the gate bites: added a throwaway unsigned commit on this branch, ran the script, confirmed it failed and named that exact commit, then removed it with `git reset --hard HEAD~1` (checked `git log -1` first to confirm it was the throwaway commit before resetting).
- Spot-checked the issue's claim against `origin/main`'s last 25 non-merge commits individually (not a raw trailer-line `grep -c`, since some commits carry more than one `Signed-off-by:` line): 24 of 25 comply, and the one exception is exactly `7412eaf1 chore(release): v0.17.1 (#2266)`, the release-bump commit this PR fixes.
- `actionlint .github/workflows/ci.yml .github/workflows/release-bump.yml`: `release-bump.yml` is clean; `ci.yml` reports the same 5 pre-existing shellcheck warnings (SC2010/SC2012 in unrelated jobs) that `actionlint` reports against `origin/main` unmodified — nothing new from this change.
- `python3 scripts/conflict-markers.py`, `python3 scripts/changelog.py check`, `git diff --check` — all clean.
- Not run: `mix precommit` (no Elixir code touched) and the full CI plan (left to the PR's own CI run).

Closes #2290


scripts/ci/dco.py
```diff
@@ -0,0 +1,102 @@
...
+        ["git", "log", "-1", "--format=%s", sha],
+        cwd=root, capture_output=True, text=True, check=True,
+    )
+    return result.stdout.strip()
+
+
+def unsigned_commits(root: str, base: str, head: str) -> tuple[list[str], list[str]]:
+    """Return (every non-merge commit, the ones missing a sign-off trailer)."""
+    shas = commits(root, base, head)
+    missing = [sha for sha in shas if not TRAILER_RE.search(message(root, sha))]
```
jhgaylor: [P2] Require an actual nonempty sign-off trailer

`TRAILER_RE.search()` scans the entire message, so mentioning the format in prose satisfies the gate without signing off the commit. I created a commit without `-s` with this message:

```text
docs: describe DCO format

For example, contributors should add:
Signed-off-by: Example <example@example.com>

This documents the required format; I have not signed off this change.
```

`git interpret-trailers --parse` returns no trailers, but this script exits 0 with `1 commit judged, all signed off`. A message ending in `Signed-off-by:   `, preserved with `git commit --cleanup=verbatim`, also passes. Both violate the nonempty sign-off trailer requirement in #2290 and `CONTRIBUTING.md`; they are separate from the deliberate decision not to match the signer's identity to the author.

Parse the trailing trailer block (for example with `git interpret-trailers --parse`) and require a `Signed-off-by` value that remains nonempty after trimming. Add regression cases for a prose example and whitespace-only value, retaining the existing different-identity passing case.

jhgaylor (changes requested):
Independent adversarial review of all six changed files at `5c73b5812c557dd0e6e2f0e776a707b97eb4ab22` against `5b9d25e128a958433e3070ffd2769f05a137c156`.

Requesting changes for one reproduced correctness gap: the new gate accepts a commit without an actual sign-off trailer when its prose contains a sign-off example, and also accepts a whitespace-only value. This leaves the missing-certification case from #2290 falsely green. The inline finding has the reproduction and a narrow remedy; author/trailer identity matching remains intentionally unnecessary.

Validation: all 127 CI Python tests passed in an isolated checkout (the first sandboxed run hit `/dev/stderr` restrictions in existing dead-code tests; the unrestricted rerun passed). The PR's own DCO check, conflict-marker check, changelog validation and diff whitespace check passed. `actionlint` reports exactly the same five existing shellcheck warnings at base and head, with no new diagnostics. Rechecked the unchanged head and current GitHub checks: all completed successfully or were skipped. No server/SDK suites were rerun locally for this tooling-only change.

scripts/ci/dco.py
jhgaylor: Fixed in 1d6f4c4c: swapped the whole-message regex for `git interpret-trailers --parse --only-trailers` (the same detector `git commit -s` writes to), and now require the `Signed-off-by` value to be non-empty after trimming. Added regression tests for both the prose-mention case and the whitespace-only-value case (`--cleanup=verbatim`) in scripts/ci/test_dco.py. The no-identity-match decision is unchanged. Full scripts/ci suite (129 tests) passes, and the gate still catches a throwaway unsigned commit on this branch (created and removed for verification).

scripts/ci/test_dco.py
```diff
@@ -0,0 +1,120 @@
...
+        self.assertEqual(result.returncode, 0, result.stderr)
+        self.assertIn("1 commit judged", result.stdout)
+
+    def test_a_prose_mention_of_the_trailer_is_not_a_trailer(self):
+        # "Signed-off-by: ..." appears in the message but mid-sentence, not
+        # as the message's trailing trailer block; a whole-message regex
+        # would wrongly accept this.
+        sha = self.commit(
+            "a.txt", "a\n", signoff=False,
+            trailer="Body mentions Signed-off-by: Example <example@example.com> in passing.",
```
jhgaylor, line 103: [P3] Use the original line-start prose example in this regression

This fixture passes even with the old buggy checker: its regex was anchored at `^Signed-off-by`, so a mid-sentence mention never matched. I reran this test against `5c73b5812` and confirmed it passes. To preserve coverage of the repaired bug, put `Signed-off-by: Example <example@example.com>` on its own line in the body and add a separate trailing prose paragraph, as in the original review reproduction. I independently verified that exact reproduction is rejected by the current code, so this is a nonblocking test improvement.

jhgaylor (approved):
Approved after a fresh adversarial review of all six changed files at `1d6f4c4cebd9a71ffd7f2317d2699c3dcd13dd87` against `5b9d25e128a958433e3070ffd2769f05a137c156`, including #2290, the prior finding and its reply. The original line-start prose example and whitespace-only sign-off both now fail correctly. No blocking findings remain; trailer identity matching and merge skipping preserve the intended policy.

Validation in an isolated checkout: all 129 CI Python tests passed (rerun outside the sandbox after its `/dev/stderr` restriction affected existing dead-code tests), plus 14 independent adversarial checks covering the original reproduction, trailer variants, PR merge ranges, unsigned side-branch commits, release-bot `git commit -s`, and invalid revisions. The PR's own DCO check, conflict-marker check, changelog check and diff whitespace check passed. Actionlint has exactly the same five pre-existing diagnostics as the base. Current GitHub checks were rechecked before submission and are successful or skipped.

One optional regression-test improvement is noted inline; the current implementation was independently verified against the exact original failure.

PR #2303, opened 2026-09-15
feat(sdk-contract): fail Swift generation when a released property disappears

### LINKED ISSUE #2296: The Swift compatibility guard does not catch a property disappearing
The two rules in `scripts/sdk-contract/generate-swift.py` compare **optionality**: the decode rule fails a property the last release could omit, and the source rule fails one the last release published as `Optional`. Neither asks whether a property the last release published still exists.

So a generated model can stop declaring a property and generation succeeds.

## Reproduction

On `43bf9cba`, empty `EXTRA_PROPERTIES`:

```
generation SUCCEEDED
LogEvent stream properties now: 0
```

`conversationID` and `agentID` are simply gone from the generated `LogEvent`, `--check` writes the new output happily, and the only thing that fails is two compile errors in `PublicSurfaceTests`:

```
error: value of type 'LogEvent' has no member 'conversationID'
error: value of type 'LogEvent' has no member 'agentID'
```

## Why the current cover is thin

`PublicSurfaceTests` is real cover but it is a hand-written list: it names the families the generation campaign moved (#2251), so it catches a removal on `AuthMe`, `AdminUserPage`, `LogEvent` and the reapply request, and says nothing about the other ~40 generated models. That is the same shape as the gap #2284 reported — a guard whose reach is whatever someone remembered to name.

A property vanishes for ordinary reasons: a contract rename (which reads as one property removed and another added), a schema restructured behind a `$ref`, a `TYPE_OVERRIDES` entry retyping a nested object to `JSONValue`. Each is a compile break for every consumer of that property.

## Suggested fix

The baseline for it already exists. `released_properties()` returns every public property of every model under `Models/` at the last release tag, so the check is the third question over the same data: for a type still generated, fail when a property that release published is absent from the output, unless it is recorded as a deliberate removal.

That table would not stay empty — `AuthMe.onboardingState` was a deliberate removal in #2289, and it is the shape of entry to expect: a retirement with a changelog entry, cited. Which argues for the entry naming the release that removed it, so the table can be pruned once that release is old enough to stop mattering.

Note the interaction with a rename: `camel()` learning `ms` → `MS` in #2295 renamed nothing, but only because nothing generated ended in `Ms`. A rename is invisible to all three rules, since each is keyed by the property name. A removal check catches half of it — the old name vanishing — which is most of the value.

Found while building the compiled-consumer regression #2269 asked for. Related: #2284 (the guard), #2277 (generator shapes), #2251 (the inventory).

BinaryBourbon:
## Summary

The two compatibility rules in `scripts/sdk-contract/generate-swift.py`'s `Generator.compatibility_failures` compare **optionality**: whether a property the last release could omit is now required, and whether a property the last release published as `Optional` is now required. Neither notices a property disappearing outright, so a generated model can simply stop declaring a property and generation succeeds — the only thing that used to catch it was the handwritten `PublicSurfaceTests` compile test, and only for the families it names (#2284's gap, again).

- Adds a third rule: using the `shipped` baseline (`released_properties()`), for every type that is still generated, a property the last release published must still be among that type's current properties, unless recorded in a new `REMOVED_PROPERTIES` table as a deliberate, cited retirement.
- "Current properties" is read the same way `shipped` is — every `public var` the type publishes — not just the generator's own field list. That includes a computed property the generator itself adds outside the field list (`Agent.permissionPolicy`, `Teammate.id`) and a handwritten extension elsewhere under `Models/` (`LogEvent.stageData`, in `Events.swift`). Comparing only the field list would have reported all three as removed the moment this guard shipped.
- `REMOVED_PROPERTIES` sits next to `REQUIRED_BY_CONTRACT`. Its first entry is `("AuthMe", "onboardingState")`, retired in #2269 (confirmed live: `git show v0.17.1:sdk/swift/Sources/FountainKit/Models/Account.swift` still publishes it as `public var onboardingState: String?` inside `AuthMe`).
- `render()`'s `ValueError` now names the remedy for all three rules.
- Tests: `test_removed_properties_entries_name_a_real_removal` (a self-pruning guard — every `REMOVED_PROPERTIES` entry must name a property the last release actually published *and* that is absent from the current output, or the test fails), `test_a_vanished_property_fails_generation` (the issue's reproduction, via a property deleted from a copy of the contract) and `test_a_recorded_removal_passes`. `test_undescribed_properties_reach_the_output` is updated: its second half asserted the exact bug being fixed here (emptying `EXTRA_PROPERTIES` "generates cleanly" and just drops `LogEvent.conversationID`/`agentID`); it now asserts that generation fails, naming the vanished property.
- `contributing/swift-wire-models.md`: one new paragraph appended at the end of "Compatibility rules" describing the third rule, the table, its first entry, and the rename interaction (a rename is invisible to all three rules, keyed by property name; it reads as an old-name removal plus a new-name addition, and this guard catches the half that matters — the old name silently disappearing). I left the "Three guards..." sentence as-is: it's specifically about the guards protecting `OPTIONAL_COMPAT`, not a global count, and stayed accurate. I did not touch the "Neither rule catches a property *disappearing*" paragraph (pre-existing, describes the pre-fix gap) or anything between the "Generation fails when a property would be non-Optional" and "(#2284)." paragraphs, per instructions — another change is concurrently moving sentences there.

Closes #2296

## Verification

Ran, all green:
- `python3 -m unittest discover -s scripts/ci -p 'test_swift_generation.py' -v` — 29 tests (26 existing + 3 new), all pass.
- `python3 -m unittest discover -s scripts/ci -p 'test_*.py'` — full CI Python suite, 125 tests, all pass.
- `python3 scripts/sdk-contract/generate-swift.py --check` — no diff; generated output is unchanged (the new rule only tightens a guard, it doesn't change what's rendered).
- `python3 scripts/sdk-contract/check-propagation.py --client swift` — passes (`additive request/response field preserved`); this rule is about removals, not additive fields, so it's unaffected, but I ran it anyway.
- `cd sdk/swift && swift build` — builds clean.
- `python3 scripts/conflict-markers.py`, `python3 scripts/changelog.py check`, `git diff --check` — all clean.
- Confirmed the `AuthMe.onboardingState` entry is needed: `git show v0.17.1:sdk/swift/Sources/FountainKit/Models/Account.swift` still declares it, so the last release tag still publishes it.

Not run: `swift test` (the full Swift test suite) — `swift build` alone confirmed the generated output is unchanged and compiles; I did not additionally run the full Swift XCTest suite given the change is Python-only and generation output is byte-identical to main.


jhgaylor (approved):
Adversarial review of `5cbdc83ccacfba1a00a4e2ca2f846cd1c4e57397`: **approved**. No blocking correctness, compatibility, or regression finding.

Reviewed all three changed files, issue #2296, the released Swift/contract baselines, generator callers, computed and handwritten extension properties, and the deliberate `AuthMe.onboardingState` retirement. The removal guard enforces the requested scope for types that remain generated; its exception does not weaken the existing requiredness/source-compatibility checks.

Validation on an isolated checkout of this commit:
- 29 focused Swift-generation tests and all 125 CI Python tests passed.
- `python3 scripts/sdk-contract/generate-swift.py --check` passed; generated bytes also match the approved base.
- Independent mutation probes detected all 373 released generated-property removals, a renamed contract field, and removal of the handwritten public `LogEvent.stageData` accessor.

The initial full tooling run had four sandbox-only failures because `tee` could not open `/dev/stderr`; rerunning the unchanged suite outside that restriction passed. I did not rerun the full Swift suite locally; the PR's Swift Linux/macOS and CI-required checks are successful.

PR #2304, opened 2026-09-15
feat(sdk): map sandbox_parking to the not-ready error in every client

### LINKED ISSUE #2291: SDKs map `sandbox_parking` to NotReadyError
`sandbox_parking` (#2286) is a new 503 error reason: the sandbox reaper has a durable claim on a conversation's sandbox and is mid-checkpoint or mid-provider-suspend-call, so a wake or an attach is refused retryably with `Retry-After` rather than racing that park. It should map to each SDK's "not ready, retry" error, the same family `provisioning`, `sprite_probe_failed` and `fleet_full` already join — `NotReadyError` in TypeScript and Python, its Elixir and Swift equivalents.

Right now none of the five clients know the code, so a caller sees each SDK's generic API error instead of the retryable class. `docs/sdk.md`'s error table narrows around this on purpose (PR #2286) until the mapping lands.

## Mappings needed

- **TypeScript** (`@managoat/fountain-sdk`): add `sandbox_parking` to the reason list that constructs `NotReadyError`.
- **Python**: same, for its `NotReadyError` equivalent.
- **Elixir**: same, for whatever this client calls its not-ready exception/tagged error.
- **FountainKit Swift**: same.
- **Legacy Swift** (pre-FountainKit, if still shipped): same.
- **Shared conformance fixture**: the cross-SDK conformance suite (`fountain-sdk-conformance-suite` memory: "SDK wire contract") should gain a case exercising a 503 `sandbox_parking` response and asserting every client raises its not-ready class with the right `Retry-After`, the same way it already covers `provisioning`/`sprite_probe_failed`.

## Why this is separate from #2286

#2286 is the server-side fix (the reaper's park claim and the retryable 503 itself); this is client-side follow-up across five codebases with their own release cadences, out of scope for that PR.

Refs #2286.

BinaryBourbon:
## What changed and why

`sandbox_parking` (#2286, still open) is a new 503 error reason: the sandbox
reaper holds a durable claim on a conversation's sandbox and is
mid-checkpoint or mid-provider-suspend-call, so a wake or attach is refused
retryably with `Retry-After` rather than racing that park. It joins
`provisioning`, `sprite_probe_failed` and `fleet_full` in the "not ready,
retry" family every client already recognizes. This PR is purely client-side
and does not depend on #2286 merging.

- **TypeScript** (`sdk/typescript/src/errors.ts`): added `sandbox_parking` to
  the `FountainErrorCode` union, `RETRYABLE_CODES`, the `NotReadyError` doc
  comment, and the `switch` that builds `NotReadyError`. New test in
  `test/team.test.ts` alongside the existing `provisioning` case.
- **Python** (`sdk/python/src/fountain/errors.py`): added it to the
  retryable set and the `NotReadyError` mapping. New test in
  `tests/test_sdk.py` alongside `test_error_classes_use_api_code_and_retry_after`.
- **Elixir** (`sdk/elixir/lib/fountain/error.ex`): added it to
  `@retryable_codes` and the `:not_ready` clause. New test in
  `test/http_sse_test.exs` alongside the existing `provisioning` tests.
- **FountainKit Swift** (`sdk/swift/Sources/FountainKit/Errors/FountainError.swift`):
  added it to the retryable list, the `.notReady` doc comment and case list.
  New test in `Tests/FountainKitTests/ClientTests.swift`'s `ErrorMappingTests`.
- **Legacy Swift** (`sdk/swift/Sources/Fountain/Errors.swift`): added it to
  the retryable list and the `.notReady` case. This target has no dedicated
  error-mapping test file (the closest thing, `ConformanceTests.swift`, only
  asserts the kind-name string), so no test was added there, matching the
  instruction to add one "if that target has one."
- **Go CLI** (`cli/`): left untouched. It has no per-reason error-code
  mapping — `provisioning`/`sprite_probe_failed` there name stream/log
  stages (e.g. `errProvisionFailed`, "provisioning failed" diagnostics), not
  an SDK error taxonomy like the other four clients have.
- **Conformance**: added
  `sdk/conformance/scenarios/error-503-sandbox-parking-is-retryable.json`,
  modeled on `error-503-sandbox-unavailable-is-retryable.json` (kind
  `not_ready`, status 503, code `sandbox_parking`, `retry_after` 30 from the
  header, retryable true), registered it in `matrix.json` as `"yes"` for all
  four SDKs, and added `sandbox_parking` to the `not_ready` row of
  `sdk/conformance/README.md`'s error-kind table.
- **Releases**: this changes what each package ships, so each SDK's release
  gate requires a bump. TypeScript 6.0.0 -> 6.1.0, Python 0.5.1 -> 0.6.0,
  Elixir 0.5.1 -> 0.6.0 (version file, `USER_AGENT`/`@user_agent`, and
  CHANGELOG heading updated together in each, matching the last such bump,
  a129b429). Swift has no version file (SwiftPM releases by git tag) and no
  release-gate workflow, so only its CHANGELOG's `Unreleased` section gained
  an entry — no version was cut.
- **Changelog fragment**: `changelog.d/2291-sdk-sandbox-parking.md`.

`docs/sdk.md`'s `NotReadyError` row is deliberately **not** touched here: the
open PR #2286 adds a row to the same error table, and editing it in both
places would conflict. It should get `sandbox_parking` in a follow-up once
#2286 lands.

Closes #2291.

## Verification

Ran, all green:

- `python3 sdk/conformance/lint.py` — 29 scenarios ok.
- All four conformance adapters against the new scenario and the full
  existing set: `npm run conformance` (TypeScript, 29/29), `python3 -m
  unittest discover -s tests -p test_conformance.py` (Python, 29/29), `mix
  test test/conformance_test.exs` (Elixir, 29/29), `swift test --filter
  ConformanceTests` (Swift, both `Fountain` and `FountainKit` targets).
- Verified the new scenario is a real check: temporarily changed its
  expected response code to `sandbox_parking_WRONG` and confirmed all four
  adapters failed with a clear mismatch, then restored the original file and
  reran everything green.
- Each SDK's full unit suite: TypeScript `npm test` (128/128) and `npm run
  typecheck` and `npm run build`; Python `python3 -m unittest discover -s
  tests -p "test_*.py"` (57/57); Elixir `mix test` (70/70); Swift `swift
  test` (122/122, both targets).
- `node scripts/sdk-release.mjs guard origin/main` (TypeScript) and `elixir
  scripts/elixir-sdk-release.exs guard origin/main` (Elixir) both report the
  version bump is well-formed and ready to publish.
- Manually replicated the Python release-gate's version/CHANGELOG-heading
  check (its workflow needs a live PR base ref) — the shipped-surface diff
  triggers, the version differs from base (0.5.1 -> 0.6.0), and
  `sdk/python/CHANGELOG.md` has the matching heading.
- Repo-level gates: `python3 scripts/conflict-markers.py`, `python3
  scripts/changelog.py check`, `git diff --check`.

Not run: the TypeScript packed-tarball-in-a-throwaway-consumer-project check
and the Python built-wheel-in-a-fresh-venv check that CI's `typescript-sdk`
and `python-sdk` jobs also perform — those need the full matrix/network setup
those jobs provide. `npm run build` (TypeScript) and the full local unit
suites for both stood in as a proxy and passed.


sdk/swift/Sources/Fountain/Errors.swift
```diff
@@ -70,7 +70,8 @@ func fountainError(
   let kind: FountainError.Kind
   switch code {
   case "conversation_busy": kind = .conversationBusy
-  case "provisioning", "sprite_probe_failed", "fleet_full", "sandbox_unavailable": kind = .notReady
+  case "provisioning", "sprite_probe_failed", "fleet_full", "sandbox_unavailable", "sandbox_parking":
```
jhgaylor: [P2] Restore the required Swift CI check. Adding this case makes the line exceed swift-format's limit. `swift format lint --strict --recursive Package.swift sdk/swift/Sources sdk/swift/Tests` exits 1 with `Errors.swift:73:1: [LineLength] line is too long`; both [Linux](https://github.com/managoat/fountain/actions/runs/35006386626/job/104507119832) and [macOS](https://github.com/managoat/fountain/actions/runs/35006386626/job/104507119805) fail here, which also fails `CI required`. This reproduces locally on the head while the base file passes. Run swift-format on this switch to wrap its case list, then rerun the strict check and Swift jobs.

jhgaylor (changes requested):
Independent adversarial review of e3b1975 against 5b9d25e, covering all 25 changed files and issue #2291. The new error mappings, retry metadata, version updates, and shared conformance assertions are consistent across TypeScript, Python, Elixir, and both Swift products.

One blocking issue: the changed Swift switch fails the repository's required formatter check on both CI platforms (inline finding).

Validation: conformance fixture lint passed (29 scenarios); TypeScript team and conformance tests passed (42/42); Python full SDK suite passed (57/57); Elixir full SDK suite passed (70/70); Swift conformance for both products and ErrorMappingTests passed. Independently reproduced the strict Swift lint failure; the base version of that file passes. Full server and package-publication checks were not rerun locally; current GitHub checks were inspected.

sdk/swift/Sources/Fountain/Errors.swift
jhgaylor: Fixed in b148c64d: ran `swift format format -i` on Errors.swift, which wrapped the case list; `swift format lint --strict --recursive` now exits 0 across sdk/swift, and `swift test` still passes 122/122 for both targets. Also checked FountainKit's FountainError.swift with the same lint since it has a similar list — it was already within the limit.

sdk/typescript/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@managoat/fountain-sdk",
-  "version": "6.0.0",
+  "version": "6.1.0",
```
jhgaylor, line 3: [P2] Honor the deferred error contract before publishing SDK releases. The [maintainer's updated #2291 decision](https://github.com/managoat/fountain/issues/2291#issuecomment-5686379148) closes this work until the park/wake design names its refusal; #2307 explicitly leaves `sandbox_parking` provisional and requires the ADR before implementation. This version bump, together with Python/Elixir 0.6.0, triggers the main-branch npm/PyPI/Hex publication workflows, shipping and documenting that provisional name before the decision is settled. Defer this release-bearing change until the ADR confirms the refusal, then align the mappings, fixture and changelogs with it; proceeding earlier needs an explicit superseding maintainer decision. Removing only the version bumps would leave the shipped-change release gates unsatisfied.

jhgaylor (changes requested):
Independent adversarial re-review of all 25 changed files at b148c64d against 5b9d25e, including the HTTP callers, five client mappings, retry metadata, conformance adapters and package versions. The original Swift formatting finding is fixed: strict lint passes locally and both current-head Swift CI jobs are green. I found no remaining implementation defect in the mappings.

Requesting changes for the release/product-intent conflict detailed inline: [#2291 was explicitly deferred](https://github.com/managoat/fountain/issues/2291#issuecomment-5686379148) after #2286 was closed, pending #2307's ADR and its chosen refusal. This PR still releases that unsettled public contract.

Current-head validation: TypeScript 128/128 tests plus typecheck/build/contract; Python 57/57; Elixir 70/70 plus formatting; Swift conformance for both products and ErrorMappingTests with warnings as errors; conformance fixture lint (29 scenarios); conflict/changelog/diff checks all passed. All GitHub checks are success/skipped, including the SDK release gates. Full server and publication-artifact checks were inspected in CI, not rerun locally.

jhgaylor:
Closing rather than revising: the review is right that this ships a provisional contract. #2291 was closed after this PR was dispatched, because #2286 was closed and the park-versus-wake protocol is being designed as an ADR first (#2307). The error name the ADR settles on may not be `sandbox_parking`, and merging this would publish TypeScript 6.1.0, Python 0.6.0 and Elixir 0.6.0 around that name.

The branch `sdk/2291-sandbox-parking-not-ready` stays as salvage for constraint 6 on #2307: five client mappings, a unit test in each, the conformance scenario (verified by breaking it, all four adapters fail), and the version bumps and changelog headings the release gates require. When the ADR names the refusal, rename the code in those files and reopen or re-cut from here.

PR #2308, opened 2026-09-15
fix(webhooks): put the checkpoint stage in the catalogue, and guard stages with computed statuses

### LINKED ISSUE #2294: Webhooks: the `checkpoint` stage is published but not in the catalogue
`Fountain.Conversations.HomeCheckpoint` publishes a `checkpoint` stage that the webhook catalogue does not name.

**Where.** `apps/fountain/lib/fountain/conversations/home_checkpoint.ex` calls `Conversations.publish_stage(&1, "checkpoint", state, meta)` (line 114) for every holder of the machine, with `state` = `"done"` (line 104) or `"failed"` (line 90). Since #1079 (2026-08-24).

**What that means.**

- `Fountain.Webhooks.Events` has `checkpoint_restore` but no `checkpoint`, so `conversation.checkpoint.done` and `conversation.checkpoint.failed` are not subscribable by name: `valid_filter?/1` refuses them.
- `Webhooks.do_dispatch/2` filters with `Events.matches?/2` and no catalogue check, so an endpoint subscribed to `*` receives both anyway.
- `docs/reference/webhooks.md` does not list the stage.

**Why the guard missed it.** `events_test.exs` reads call sites with a regex that needs a literal stage *and* a literal status. This site's status is a variable, so it is one of the three sites the regex cannot read (with `conversations.ex` `"turn", status` and `turn_machine.ex` `"turn", if(...)`). #2279 made the test walk `lib/` instead of a fixed file list, which is how this surfaced, but it still only asserts the wider stage-only view for *retired* stages.

**Fix.** Either add `{"checkpoint", ~w(done failed)}` to the catalogue and a row to the docs page (an integrator watching a persistent home probably wants it), or route it through an existing stage. And consider asserting in `events_test.exs` that every literal stage in the source is a catalogue stage, so a computed status cannot hide a new stage again.

jhgaylor:
## Summary

`Fountain.Conversations.HomeCheckpoint.on_park/1` publishes a `checkpoint`
stage with status `"done"` or `"failed"` for every conversation holding a
persistent home's sandbox, but the webhook catalogue never named it, so
`conversation.checkpoint.done` and `conversation.checkpoint.failed` were not
subscribable by name and `docs/reference/webhooks.md` did not list the
stage. The guard test's regex needs a literal status, and this call site's
status is a variable, so it was one of the sites the regex could not read.

- Add `{"checkpoint", ~w(done failed)}` to the catalogue in
  `apps/fountain/lib/fountain/webhooks/events.ex`, next to
  `checkpoint_restore`.
- Add a row to the stage table in `docs/reference/webhooks.md` describing
  what a home checkpoint is and what `failed` means for it.
- Harden `events_test.exs`:
  - add `lib/fountain/conversations/home_checkpoint.ex` to the fixed
    `@sources` list
  - add an independent stage-only walk over every `.ex` file under
    `lib/fountain/` (`Path.wildcard`, no fixed file list) asserting every
    stage it finds is either a catalogue stage or a retired one — this is
    what would have caught `checkpoint` even before anyone added
    `home_checkpoint.ex` to `@sources` by hand
  - fix "the catalogue names nothing the source cannot produce" so a
    catalogue entry whose stage has at least one call site with a
    non-literal status counts as producible, while a stage with no call
    site at all still fails
- Add `changelog.d/2294-checkpoint-webhook-stage.md`.

An endpoint already subscribed to `*` was receiving both events all along
(`Webhooks.do_dispatch/2` never consulted the catalogue); this change only
makes them nameable and documented.

## Verification

Ran, all passing:
- `mise exec -- mix test apps/fountain/test/fountain/webhooks/` — 54 tests,
  0 failures (private `DATABASE_URL` pointed at `fountain_test_2294`,
  migrated from the umbrella root; dropped afterward).
- `bash scripts/test-docs.sh` — core + all four extension manual suites,
  98 tests total, 0 failures.
- `MIX_ENV=test mise exec -- mix compile --warnings-as-errors` — clean,
  no warnings.
- `mise exec -- mix format` from the umbrella root, and from
  `apps/fountain` — no further changes after either.
- `mise exec -- mix credo --strict` on the two changed source files — no
  issues.
- `python3 scripts/conflict-markers.py`, `python3 scripts/changelog.py check`,
  `git diff --check` — clean.
- Broke both new guards to confirm they catch what they're meant to, then
  restored:
  - removed the `checkpoint` catalogue entry: the new wildcard-walk test
    failed, naming `checkpoint`.
  - added a bogus catalogue entry (`bogus_stage_2294`): the
    stale-direction test still failed, naming
    `conversation.bogus_stage_2294.done`, while `checkpoint.done`/`.failed`
    passed cleanly.

Not run: dialyzer, sobelow, release assemble (CI covers these).

Closes #2294


apps/fountain/test/fountain/webhooks/events_test.exs
```diff
@@ -107,21 +162,57 @@ defmodule Fountain.Webhooks.EventsTest do
...
+    # `turn.done`/`turn.failed` and `checkpoint.done`/`checkpoint.failed` come
+    # from call sites whose status is a variable, so the literal-pair regex
+    # cannot see them. A catalogue entry counts as producible when either its
+    # exact pair is literally published, or its stage has at least one
+    # call site the literal regex could not read the status of. A stage that
+    # appears at no call site at all is in neither set, so it still fails.
+    stale =
+      Events.types()
+      |> Enum.reject(fn type ->
+        MapSet.member?(published, type) or MapSet.member?(non_literal, stage_of(type))
```
BinaryBourbon: [P2] Keep the computed-status exception at the (stage, status) level

`non_literal` proves only that a stage has one call whose status the regex cannot read; this OR consequently treats every status for `turn` or `checkpoint` as producible. I appended only `never_emitted` to the existing `turn` catalogue statuses: all 17 tests still passed. The equivalent base catalogue/test mutation fails specifically with `conversation.turn.never_emitted`. This is newly lost exact-status coverage, not a preexisting regex limitation: a misspelled or stale status can now become a valid subscription that never receives an event while this guard remains green.

Keep the wildcard stage guard, but constrain the fallback to independently supported computed `(stage, status)` pairs (or resolve the computed statuses), rather than exempting an entire stage. Add a negative regression for a nonexistent status on an existing computed-status stage, alongside the wholly bogus-stage check.

BinaryBourbon (changes requested):
Adversarial review of f7d4f568a75407ee436229f165f30544f8083b3e against 5b9d25e128a958433e3070ffd2769f05a137c156: one P2 test regression needs repair (inline).

The production catalogue addition is correct: HomeCheckpoint emits done/failed, endpoint validation now accepts both exact names and the checkpoint wildcard, and dispatch continues to match the existing tenant-owned subscriptions. The documentation and changelog match issue #2294. The independent wildcard walk successfully catches a missing checkpoint stage even when home_checkpoint.ex is removed from @sources. However, the new stage-wide fallback loses the existing protection against nonexistent statuses on a computed-status stage.

Validation: reviewed all four changed files and relevant checkpoint, publish, dispatch, subscription and regression-test paths. Executed the unchanged EventsTest source standalone in an isolated exact-head archive with pinned Elixir 1.19.2 / OTP 28: 17 tests, 0 failures; direct checkpoint filter/matching probes passed. Mutation probes correctly rejected the missing checkpoint stage and a wholly bogus stage, but accepted a nonexistent turn status; the equivalent base test rejected that same status. The full application/database suite was not rerun locally. Exact-head GitHub CI succeeded, including all six test partitions, static analysis and CI required: https://github.com/managoat/fountain/actions/runs/35011534631. Head and checks rechecked before submission.

apps/fountain/test/fountain/webhooks/events_test.exs
BinaryBourbon: Fixed in d8a38125: `resolvable_by_computed_status?/2` is now scoped to one (stage, status) pair — a computed-status call site only excuses a status when the same file also spells that status out literally (home_checkpoint.ex has `"done"`/`"failed"`; turn_machine.ex and conversations.ex have their if/else literals for turn). Pulled the stale check into a pure `stale_catalogue_entries/2` and added three regression tests: an extra status on `turn` (has computed sites), an extra status on `checkpoint` (only computed site), and a wholly new stage. Re-verified against the real module by appending `never_emitted` to turn in events.ex and confirming the suite failed naming exactly `conversation.turn.never_emitted`, then restored. All 57 webhook tests, compile --warnings-as-errors, format at both roots, and credo --strict are clean.

apps/fountain/test/fountain/webhooks/events_test.exs
```diff
@@ -50,12 +57,95 @@ defmodule Fountain.Webhooks.EventsTest do
...
+  #     "interrupted"` resolves conversation.turn.failed and .interrupted
+  # `turn.done` and `turn.failed` are also literally published elsewhere
+  # (conversation_server.ex, turn_machine.ex), so this is redundant for
+  # `turn` today; it is load-bearing for `checkpoint`, which has no other
+  # call site.
+  defp resolvable_by_computed_status?(stage, status) do
+    literal = ~s("#{status}")
+
+    Enum.any?(computed_status_sites(), fn {_source, s, content} ->
+      s == stage and String.contains?(content, literal)
```
BinaryBourbon: [P2] Resolve computed output statuses instead of accepting any same-file string

`String.contains?(content, literal)` does not establish that the computed call can emit this status. `turn_machine.ex:688` contains `"completed"` as the stored row status, but explicitly maps it to emitted `"done"`. Appending only `completed` to the existing `turn` catalogue still passes all 20 current EventsTest tests; the equivalent base mutation fails with `conversation.turn.completed`. Likewise, `checkpoint.persistent` passes because `home_checkpoint.ex` contains the sandbox mode `"persistent"`, although its only emitted statuses are `"done"` and `"failed"`.

This retains the earlier loss of exact-status protection for strings already used elsewhere in those files: a stale or mistaken catalogue entry becomes a valid subscription that never receives an event while the guard is green. Restrict the fallback to independently supported computed output pairs (or resolve the actual status-producing expression/helper arguments), and add a negative case whose bogus event status already occurs in the emitter file, such as `turn.completed`.

BinaryBourbon (changes requested):
Fresh adversarial re-review of d8a381257ae78127257389ccaa1dd0c7325a4c76 against 5b9d25e128a958433e3070ffd2769f05a137c156: one P2 test regression remains (inline). Two independent reviewers reproduced the same residual defect. The original `turn.never_emitted` case now fails correctly, but the fallback still accepts unrelated strings from a computed emitter's file as producible event statuses.

Reviewed all four changed files: `apps/fountain/lib/fountain/webhooks/events.ex`, `apps/fountain/test/fountain/webhooks/events_test.exs`, `docs/reference/webhooks.md`, and `changelog.d/2294-checkpoint-webhook-stage.md`, plus checkpoint creation/holder selection, publish/dispatch, endpoint validation, relevant regression tests, ADR 0024, and issue #2294. The checkpoint catalogue addition and documentation match the actual done/failed emissions; the existing dispatch and tenant boundaries remain consistent.

Validation: executed the unchanged current EventsTest standalone in isolated exact-head archives with pinned Elixir 1.19.2 / OTP 28: 20 tests, 0 failures. Direct checkpoint exact/wildcard filter and matching probes passed. Both reviewers independently found that adding only `completed` to the real turn catalogue leaves all 20 tests green; the independent base comparison rejects the same mutation (15 tests, 1 failure naming `conversation.turn.completed`). A second head mutation also accepts `checkpoint.persistent`, despite that string being a sandbox mode. The full application/database suite was not rerun locally. Current-head GitHub CI passed, including all six partitions, static analysis, coverage and CI required: https://github.com/managoat/fountain/actions/runs/35014196243. Identity, open state, head and checks reverified before submission.

apps/fountain/test/fountain/webhooks/events_test.exs
BinaryBourbon: Fixed in 679a1113: replaced the substring check with `@computed_sites`, a `{file, stage, regex}` table where each regex reads the actual status-producing expression (turn_machine.ex's `if(row.status == "completed", do: "done", else: "failed")` captures only the do:/else: branch values, never `"completed"`; conversations.ex's `status = if updated_turn.status == "failed", do: "failed", else: "interrupted"` likewise; home_checkpoint.ex's `publish(sandbox, "done"/"failed", meta)`). `stale_catalogue_entries/2` now has no fallback at all: a type is produced only if it is an exact literal pair, or an exact pair one of these regexes resolves. Added two completeness guards (every computed call site is covered by a table entry; every entry's regex actually resolves at least one status) and two new regressions: `turn.completed` and `checkpoint.persistent` both fail even though the literal string sits in the emitter file. Re-verified against the real module by adding `completed` to turn's catalogue statuses and confirming the suite failed naming exactly `conversation.turn.completed`, then restored. 61 webhook tests, warnings-as-errors compile, format at both roots, and credo --strict are all clean.

BinaryBourbon (approved):
Fresh adversarial whole-PR review of 679a1113eac0825ff981081ad201ab3b343bf724 against 5b9d25e128a958433e3070ffd2769f05a137c156: approved. Two independent reviewers agree; no blocking findings remain. The computed-status guard now resolves actual helper arguments and conditional branch outputs, so the prior stage-wide and unrelated-string defects are fixed.

Reviewed all four changed files (catalogue, EventsTest, webhook reference and changelog), the checkpoint emitters and holder selection, publish/dispatch, endpoint validation, relevant existing tests, ADR 0024 and issue #2294. The checkpoint done/failed names and documentation match current emissions and tenant-scoped subscription behavior.

Validation: unchanged current EventsTest executed standalone in an isolated exact-head archive with pinned Elixir 1.19.2 / OTP 28: 24 tests, 0 failures. Nine mutation probes correctly rejected the prior turn.completed, turn.never_emitted and checkpoint.persistent cases, another invalid checkpoint status, a nonexistent stage, a missing checkpoint stage without its fixed-source entry, a dead computed regex, an uncovered computed file/stage and an unknown stage in a new file. Restored source again passed 24 tests. Direct checkpoint exact/wildcard matching and invalid-status probes passed. Full application/database tests were not rerun locally. Exact-head CI is green, including all six test partitions, static analysis, coverage, CI required and secret scanning: https://github.com/managoat/fountain/actions/runs/35015209802. Head, base, checks and gh identity rechecked immediately before submission.

Issue #2307, opened 2026-09-15
Design (ADR): the sandbox reaper's idle park must serialize against wake, rehydration and admission


jhgaylor:
Decision 3 on #2255 (the reaper takes the sandbox advisory lock) landed for the two reaper writes that involve no provider I/O (stuck release, max-lifetime expiry). It cannot cover the idle park, because the park has a provider round trip (checkpoint, suspend) between "decide" and "write", and a wake never takes the sandbox lock: it coordinates through the quota reservation lock, resumes the provider, writes `ready`, and starts a server. The pre-existing race (a wake landing on a machine the reaper is still suspending, or the reaper writing `suspended` over a row a wake just made live) is real and reproduced; it is not new.

PR #2286 tried to close it in-PR with a durable claim (`sandboxes.park_claimed_at`) and went five adversarial review rounds; each round found a further ordering gap. The branch `follow/2255-reaper-liveness-lock` is kept as salvage (protocol, tests, ADR 0017 prose). Jake decided on 2026-09-15 to design this as an ADR first, then build it behind a gate. The constraints the five rounds established, each with a reproduction on the PR:

1. Every eligibility check (status, reset fence, teardown fence, grace window, liveness, the conversation's binding) must be re-read and re-applied under the same lock the write takes; a pre-lock verdict is stale by construction.
2. The provider suspend must be preceded by a durable claim that a wake, an attach and the rehydrator all refuse or honour, and followed by a compare-and-set finalize on that exact claim; a lost claim after a successful suspend needs a compensating resume.
3. Claims must survive a crash: a TTL alone lets an over-long operation be overtaken; recovery must be a serialized takeover with an exact-stamp CAS, in short transactions, with provider I/O outside every lock scope (a nested `Repo.transaction` joins the outer one via a savepoint and keeps transaction-scoped advisory locks until the outer commit).
4. Horde registry propagation is asynchronous: "no live server" on one node is not evidence. A wake's registration needs a durable marker committed before the external `start_child`, through one primitive shared by `Wake` and `Rehydrator.spawn_server/1`, with compensation if the caller dies after the child starts.
5. Recovery may not resume compute before the account-suspension and credit gates; an interrupt never provisions or resumes.
[...]

PR #2309, opened 2026-09-15
refactor(conversations): one liveness scan for the reaps; chronological DateTime comparisons in the reaper

jhgaylor:
Refs #2255 (decision 2). Refs #2307 for decision 3 (the reaper's locked writes), deferred entirely — see below.

## Decision 2 — one liveness scan, pure refactor

`Termination.reap_sandbox/1` and `Workers.SandboxReaper` each scanned a sandbox's preloaded conversations for a live `ConversationServer` themselves — the same "is any server alive on this machine" question asked twice. Adds two functions to `Fountain.Conversations.Lifecycle`, over a `%Sandbox{}` with `:conversations` preloaded (every caller already carries that from the query that fetched the sandbox, so this runs no second query):

- `live_conversation_ids/1` — the conversation ids with a live server, for `Termination.reap_sandbox/1`, which needs the list to act on (it terminates each live conversation).
- `any_server_alive?/1` — the boolean, deferring to `live_conversation_ids/1`, for `SandboxReaper.release_stuck_sandboxes/0` and `sweep_abandoned_sandboxes/0`, which only decide whether to touch the row.

`SandboxReaper.server_alive?/1` is deleted; both its call sites now call `Lifecycle.any_server_alive?/1`. Same registry lookup (`ConversationServer.whereis/1`), same set, no behaviour change.

## Decision 3 — deferred entirely to #2307

Earlier revisions of this PR tried to revalidate the reaper's stuck-release and max-lifetime writes under `Conversations.with_sandbox_lock/2`. That work is withdrawn from this PR after a third adversarial review round found the locked recheck itself has a race: it relies on `any_server_alive?/1`, but the real ready-sandbox reuse path (`Wake`) and boot rehydration both register a `ConversationServer` **without taking any lock at all**. A server can register between the reaper's false liveness read and its terminal write — the wake has already received `{:ok, pid}` and may have queued a prompt — and the reaper then commits `terminated` out from under it, killing a server that just started.

That isn't a "revalidate under a lock" fix; it needs every ready-sandbox server registration to participate in a shared protocol, or a durable wake/rehydration marker the terminal write can check — the same class of problem this PR's earlier rounds hit for the idle-park write (five review rounds on the now-closed #2286). #2307 tracks that design directly, covering both the idle-park write and this reaper-locking race as one protocol, reviewed as an ADR before any more code lands. `release_stuck_sandboxes/0`, `sweep_abandoned_sandboxes/0`, `idle_sweep/1`, `park/1` and `expire/2` are back to their `origin/main` form; nothing in this PR takes a lock the reaper's writes didn't already take before #2255's tranche started.

## Chronological DateTime comparisons

Elixir compares two `DateTime` structs field-by-field in alphabetical key order with `<`, so a naive `<` reads `~U[2026-09-30]` (day 30) as later than `~U[2026-10-01]` (day 1) — backwards. This came up while the locked-write revisions were still in this PR; nothing in `SandboxReaper`'s current (origin/main-equivalent) code compares `DateTime`s in Elixir today — every comparison is a SQL `where` clause, which Postgres compares correctly — but `sandbox_reaper_predicates_test.exs` documents the trap directly against `DateTime.before?/2` so it isn't rediscovered when #2307's protocol adds the first Elixir-side one.

## No behaviour change; no changelog fragment

Nothing a user or operator can observe changes: the liveness scan produces the same set it always did, and every reaper write path is byte-for-byte its `origin/main` form. No changelog fragment.

## Tests

`sandbox_reaper_predicates_test.exs` (new, standalone, no DB) — the `DateTime.before?/2` boundary case, using the exact literals a review used to demonstrate the naive-`<` bug.

`sandbox_reaper_test.exs` is untouched. `sandbox_reaper_lock_order_test.exs`, added by an earlier revision of this PR to exercise the withdrawn locked-write revalidation, is deleted along with that code.

`git diff origin/main --stat`: `lifecycle.ex`, `termination.ex`, `sandbox_reaper.ex`, `sandbox_reaper_predicates_test.exs` — nothing else.

`mix precommit --full` from the worktree root: `precommit: PASSED (10 stages, 156s)` plus a `test`-stage-only re-run after clearing a reused test database's accumulated `audit_events` rows (a known local-database artifact unrelated to this change, not a CI condition): `precommit: PASSED (1 stages, 138s)`.


BinaryBourbon (changes requested):
Adversarial review: **CHANGES REQUESTED** on `fd81d9d508b9c760b65ec763a74f489c5329271c`.

**Medium — the new transaction makes best-effort metering failures roll back the reap.** `release_one_stuck/2` and `expire_max_lifetime/2` now call `Conversations.update_sandbox/2` inside the outer sandbox-lock transaction (`sandbox_reaper.ex:146–163,329–343`). That function runs `record_sandbox_usage/2` and the queue poke after its own nested transaction returns, while the new outer transaction is still open (`conversations.ex:191–248`). A PostgreSQL error during a usage insert, such as statement cancellation/timeout, is rescued by `Billing.record_usage/5`, but the rescue cannot recover the aborted transaction. The outer call returns `{:error, :rollback}`, the reaper's `{:ok, result}` match raises, the terminal write is lost, and subsequent reaper passes do not run. This breaks the explicit best-effort metering contract in `ee/lib/fountain/billing.ex:660–681`.

Reproduced against the real `Billing.emit/5` insert in an isolated database: a temporary `BEFORE INSERT ON usage_events` trigger restricted to the fixture sandbox raises SQLSTATE `57014`. For an old `starting` row, this head produces `MatchError`, leaves `status = starting`, and records no reap audit. The identical diagnostic with only the reaper restored to base `5b9d25e1` succeeds, leaves `status = failed`, and records `sandbox.released_stuck`. The max-lifetime wrapper has the same nesting.

Keep the fresh eligibility check and terminal write atomic under the advisory lock, while moving best-effort metering/queue effects after the outer commit or isolating their SQL failures with an explicit savepoint. Add a real SQL-failure regression for both terminal paths; an ordinary rescued Elixir exception does not reproduce an aborted PostgreSQL transaction.

Validation: inspected all four changed files and affected callers against the settled #2255 scope, #2307 and closed #2286. Independently ran all 247 listed affected tests successfully in an isolated checkout/database with the pinned toolchain; the additional SQL-fault diagnostic fails on this head and passes with the base reaper. Current GitHub checks are successful/skipped. The unchanged idle-park protocol remains the expressly deferred work in #2307 and is not a finding here.

BinaryBourbon:
Fixed in 8307fa5f, on top of the rebase in 882405e7 (origin/main moved 5b9d25e1 → 7bafd206 while this was in review; unrelated commits, rebased clean).

**The finding, mapped to the fix.** `release_one_stuck/2` and `expire_max_lifetime/2` called `Conversations.update_sandbox/2` from inside `Conversations.with_sandbox_lock/2`'s outer transaction. `update_sandbox/2`'s effects — best-effort metering (`record_sandbox_usage/2`, via `Billing.record_usage/5`) and the sandbox queue poke — ran *after* its own nested transaction returned, but that call was still nested inside the reaper's outer one. A real PostgreSQL error during the metering insert aborts the whole outer transaction; `Billing.record_usage/5`'s rescue cannot recover an already-aborted transaction, so the terminal write is lost along with it and the reaper's `{:ok, result}` match raises.

**The split.** `Conversations`'s private `update_sandbox_if/3` now returns the locked write's raw result — `{:ok, {was, updated}}` / `{:error, reason}` — with no effects applied. Two `@doc false` doors expose it:

- `update_sandbox_row/2` — the locked write alone (FOR UPDATE re-read, the revival guard, the reset-pending rollback, `Repo.update`), safe to call inside an outer transaction because it performs no effects.
- `sandbox_updated_effects/2` — the effects `update_sandbox_row/2` defers, to be called once the transaction that made the write has committed.

`update_sandbox/2` itself is now `update_sandbox_row/2` followed by `sandbox_updated_effects/2`, using the same `was` the old code did (the FOR-UPDATE-read status, not the caller's possibly-stale struct), so every other caller — about two dozen, grepped across `lib/`, none of them inside an outer transaction — is byte-for-byte unchanged. (`finish_sandbox_reset/2`, the one other caller of the private helper besides `update_sandbox/2`, now calls `sandbox_updated_effects/2` explicitly for the same reason.)

The reaper's two paths now call `update_sandbox_row/2` under the lock, returning `{:ok, {:written, was, updated}}` or `{:ok, :skipped}`; once `with_sandbox_lock/2` returns, `sandbox_updated_effects/2` runs, then `record_reap` as before. The eligibility check and the terminal write stay atomic under the advisory lock; a metering failure can no longer unwind either.

**The two trigger regressions**, in `sandbox_reaper_lock_order_test.exs`'s new "metering does not eat the reap" describe block: a temporary `BEFORE INSERT ON usage_events` trigger, scoped to the fixture sandbox's id, raises SQLSTATE `57014` — a real aborted PostgreSQL transaction, not a rescued Elixir exception (`Billing.record_usage/5` already swallows those, so only a genuine `{:error, :rollback}` exercises the case that matters). One for `release_stuck_sandboxes/0`, one for `sweep_abandoned_sandboxes/0`'s max-lifetime path; both assert the row lands terminal, the reap audit is recorded, and the usage event itself is the only thing lost. Confirmed against `fd81d9d5` (this PR's previous head, pre-fix): both raise the exact `MatchError` on `{:error, :rollback}` the review reported, from `sandbox_reaper.ex:145` and `:328` respectively. Both pass on `8307fa5f`. Each has a mirror test with no trigger, asserting the usage event is still emitted exactly as before (2 events for `"starting" → "failed"`: provision-failed + terminated; 1 event for `"ready" → "terminated"`: terminated only).

Full precommit gate, from the worktree root: `precommit: PASSED (10 stages, 156s)`. Proves-unchanged set re-run after the split: `sandbox_reaper_test.exs` (28), `sandbox_reaper_lock_order_test.exs` (10, was 6 — 4 new), plus every test file touching `update_sandbox/2` across `lib/` and the ee suite (27 files, 624 tests) — 0 failures.

Not re-requesting review.

BinaryBourbon (changes requested):
Adversarial re-review: **CHANGES REQUESTED** on `8307fa5f7ade4205cec65ae7f228b55ada0e5db6` (base `7bafd206c191110f00138c5f72635491236e2ecc`). Two independent whole-PR reviewers identified the same remaining defect.

**Medium — preserve the row-locked previous status when deferring effects.** `update_sandbox_row/2` discards the canonical `_was` returned by the `FOR UPDATE` read (`conversations.ex:185–189`), while both reaper paths return the earlier `fresh.status` for `sandbox_updated_effects/2` (`sandbox_reaper.ex:161,345`). The advisory lock does not serialize all sandbox writers: the no-live-server path in `Termination.reap_sandbox/2` calls `update_sandbox/2` without taking it. If that writer has updated the row but has not committed, the reaper's ordinary reload still sees `starting`/`ready`; its subsequent `FOR UPDATE` waits and then reads the now-terminal row. That correct prior status is discarded, so the reaper emits another nonterminal-to-terminal transition using the stale status.

Reproduced with the actual `Termination.reap_sandbox/2` and background reaper, real usage inserts, and independent PostgreSQL connections. A synchronous query-telemetry pause holds the admin's normal transaction immediately after its UPDATE; `pg_blocking_pids` confirms the reaper is waiting on that row before the admin commits. The `ready` case records **2 `sandbox_terminated` events instead of 1**. The `starting` case records **2 each of `sandbox_provision_failed` and `sandbox_terminated` instead of 1 each**. These duplicate lifecycle records and their analytics mirror; this is not a claim of duplicate credit charges. The pre-split `update_sandbox/2` used the canonical row-locked status and suppressed effects for terminal-to-terminal writes.

Return the canonical `was` alongside the updated row and carry it through both reaper results, keeping the effects after the outer commit. Add concurrency regressions for both paths. That narrow change, applied only to the disposable review checkout, makes both diagnostics pass without regressing the SQL-error isolation.

Validation: inspected all five changed files plus relevant callers and the settled #2255/#2307 scope. On the exact PR head, the existing reaper suite and all new locking/SQL-failure regressions pass: **38 tests, 0 failures**; the additional race diagnostics fail **2/2**. With the diagnostic canonical-status repair, **40 tests, 0 failures**, including both real SQLSTATE 57014 cases. Local tests used the pinned BEAM toolchain and an isolated PostgreSQL 15 database; current GitHub CI checks are successful/skipped. The previous metering-error rollback finding is fixed by this revision. Idle-park design remains the explicitly deferred #2307 work.

BinaryBourbon:
Fixed in 322f7a1c, on top of the rebase in a36d51cc (origin/main moved 7bafd206 → 4daf6bb9 while this was in review; a webhooks fix and a changelog-only commit, rebased clean).

**The finding, mapped to the fix.** `update_sandbox_row/2` discarded the canonical `was` its own `FOR UPDATE` read returned, and both reaper paths passed their earlier `fresh.status` reload to `sandbox_updated_effects/2` instead. The advisory lock does not serialize every sandbox writer — `Termination.reap_sandbox/2`'s no-live-server path calls `update_sandbox/2` directly, without it — so a writer that commits a terminal status between the reaper's scan and the moment its `FOR UPDATE` actually locks the row left that reload stale. Using the stale status made the reaper emit a second usage event for an already-terminal row and, worse, let it overwrite the row and log its own reap audit for a transition it did not cause.

**The fix.** `update_sandbox_row/2` now returns `{:ok, {was, updated}}` — the same shape its own `FOR UPDATE` read already produced internally, just no longer discarded — and `update_sandbox/2` composes it unchanged.

**Which of the two options I took for the eligibility recheck:** reading with `lock: "FOR UPDATE"` inside `with_sandbox_lock/2` before evaluating the predicate, not a `check` function threaded through `update_sandbox_row`'s `Repo.rollback`. I tried the `check`-function approach first and it is actively wrong for this case: I confirmed directly (a small standalone script against this app's Repo) that a nested `Repo.transaction`'s `Repo.rollback` aborts its *entire* enclosing transaction on any rollback at any depth in this Ecto/DBConnection version — not just the call frame that invoked it — even when the outer code catches and ignores the inner `{:error, _}` result. Using `Repo.rollback` for an ordinary "not eligible, skip" case from inside `with_sandbox_lock/2` would have silently turned "skip" into "lose the outer transaction's work", which is the exact class of bug the first #2309 round fixed. `update_sandbox_row/2`'s doc comment now says this explicitly so a future caller doesn't reach for `check` here. Instead, both reaper paths now do their own `Repo.one(from s in Sandbox, where: s.id == ^id, lock: "FOR UPDATE")` (preloaded, for the liveness check) inside `with_sandbox_lock/2`, evaluate `stuck_eligible?/2` / `ready_abandoned?/2` against that row, and only then call `update_sandbox_row/2` — whose own internal `FOR UPDATE` re-acquires the same already-held row lock (a no-op in the same transaction) and returns that row's own canonical prior status.

**The two concurrency regressions**, in `sandbox_reaper_lock_order_test.exs`'s new "a writer outside the advisory lock" describe block: a second connection writes the fixture row terminal directly with a raw `Repo.update_all/2` (no advisory lock — simulating `Termination.reap_sandbox/2`'s no-live-server path) and holds the write uncommitted in an open transaction. The reaper is confirmed genuinely blocked on that row's `FOR UPDATE` lock via `pg_blocking_pids` (not merely slow); once the writer commits, the test applies `Conversations.sandbox_updated_effects/2` itself (what a real unlocked `update_sandbox/2` caller would already have done by the time it returned) and then asserts the reaper added nothing: `release_stuck_sandboxes/0` / `sweep_abandoned_sandboxes/0` count the row as untouched, the row state is exactly what the writer left it, no `sandbox.released_stuck`/`sandbox.expired` audit from the reaper, and no extra usage event beyond the one write already produced. Confirmed against `8307fa5f` (this PR's previous head, before this round's fix): both fail with the reaper counting the already-terminated row as released/expired a second time — `assert reaper_result == 0` observed `1` for the stuck case, `assert reaper_result == {0, 0}` observed `{0, 1}` for the expiry case — reproducing the duplicate-reap symptom directly rather than only a return-shape mismatch. Both pass on `322f7a1c`.

Full precommit gate, from the worktree root: `precommit: PASSED (10 stages, 176s)`. Proves-unchanged set re-run after this change: `sandbox_reaper_test.exs` (28), `sandbox_reaper_lock_order_test.exs` (12, was 10 — 2 new), plus every test file touching `update_sandbox/2` across `lib/` and the ee suite (27 files, 664 tests) — 0 failures.

Not re-requesting review.

BinaryBourbon (changes requested):
Adversarial re-review: **CHANGES REQUESTED** on `322f7a1c93ebb44bc102361a91b86b4267b47bd2` (base `4daf6bb987704b0191f423adae5877d31bb300f5`). Three independent whole-PR reviewers found the same remaining race. The canonical row-locked `was` repair does resolve the preceding duplicate-effects finding.

**Medium — the locked max-lifetime recheck does not serialize the liveness change it relies on.** `expire_max_lifetime/2` holds the sandbox advisory/row locks, calls `ready_abandoned?/2`, and then writes `ready -> terminated` (`sandbox_reaper.ex:325-357`). But the real ready-sandbox reuse path calls `start_conversation_server/4` without either lock (`wake.ex:209-221,275-288`), and boot rehydration does the same (`rehydrator.ex:132-153`). A server can register immediately after `Lifecycle.any_server_alive?/1` returned false but before the terminal update. The wake has already received `{:ok, pid}` and may have queued its prompt; the reaper then commits `terminated`, and `ConversationServer` stops when it observes that terminal row (`conversation_server.ex:517-536`). The prompt is lost, and the same race can kill a just-rehydrated server during a deploy.

The new regression at `sandbox_reaper_lock_order_test.exs:355-393` only makes liveness true while the reaper is blocked *before* acquiring the advisory lock, so it cannot exercise registration after the reaper's final liveness read. Make every ready-sandbox server registration participate in the same machine protocol, or commit a durable wake/rehydration marker under that protocol which the terminal write revalidates. Add a regression that pauses immediately after a false liveness read, starts a real Wake/Rehydrator child, then releases expiry and proves a successful start cannot coexist with a terminal sandbox.

`git diff --check` passes. Exact-head CI and both required checks are green.

BinaryBourbon:
Per the third adversarial review round on this PR: the locked max-lifetime recheck relies on `Lifecycle.any_server_alive?/1`, but real ready-sandbox reuse (`Wake`) and boot rehydration both register a `ConversationServer` without taking any lock — a server can register between the reaper's false liveness read and its terminal write, and the reaper can then terminate a sandbox that just got a live server (and possibly a queued prompt) under it. That applies to every locked reaper write this PR added, not just the max-lifetime path, since stuck-release has the same exposure against the rehydrator starting a server on a `starting` row.

Decision: withdraw decision 3 (the reaper's locked writes) from this PR entirely, back to `origin/main`'s code, and let #2307 design the durable-registration protocol this actually needs — it now covers both the idle-park write (the original reason for #2307) and this reaper-locking race as one problem. This PR is narrowed to decision 2 only (the shared liveness scan) plus the `DateTime.before?/2` documentation that came out of the locked-write regressions, now standalone in `sandbox_reaper_predicates_test.exs` since the code it originally protected is gone.

Retitled and rewrote the PR body accordingly. Not re-requesting review.

BinaryBourbon (approved):
Adversarial review completed at `530ff2a56f8c70324688ceb9132e72ea94419509`: **APPROVED**.

No medium-or-higher correctness, security/isolation, API, or operability findings. The shared lifecycle helpers preserve the prior liveness and termination semantics, and all current callers preload the required conversation association. Focused exact-head testing passed (70 tests, 0 failures), `git diff --check` is clean, and required CI is green.

Nonblocking observations: `any_server_alive?/1` now scans the full conversation list instead of short-circuiting, and the standalone DateTime test documents language behavior rather than exercising production code. Neither creates a current correctness defect.

PR #2311, opened 2026-09-15
chore(release): v0.18.0

jhgaylor:
Version bump for v0.18.0: both `mix.exs` files, the two Swift version sources, the six self-host image pins, and the changelog roll (`[Unreleased]` plus every `changelog.d/` fragment into `## [0.18.0]`).

Merging this PR tags the squash commit `v0.18.0` and dispatches `release.yml`, which builds the CLI binaries, publishes `ghcr.io/managoat/fountain:v0.18.0`, boots it through the documented quick start and creates the GitHub Release.


mix.exs
```diff
@@ -8,7 +8,7 @@ defmodule Fountain.Umbrella.MixProject do
       apps_path: "apps",
       # Kept in lockstep with the newest v* git tag — release-bump.yml
       # computes the next tag from this value.
-      version: "0.17.1",
+      version: "0.18.0",
```
BinaryBourbon, line 11: [P2] Keep CI passing when v0.18.0 becomes the release baseline

Merging this bump tags the merged tree v0.18.0, so `released_tag()` in `scripts/sdk-contract/generate-swift.py` starts reading this release instead of v0.17.1. Two tests in `scripts/ci/test_swift_generation.py` then fail: `test_removed_properties_entries_name_a_real_removal` still requires `AuthMe.onboardingState` in the latest release, and `test_the_released_tag_is_the_baseline_for_both_questions` still requires `CatalogMcpServersItem` to be absent. This release has the opposite shapes.

Reproduced in an isolated clone: `python3 -m unittest scripts.ci.test_swift_generation -q` passes all 29 tests before a local `git tag v0.18.0 HEAD`, then reports two failures afterward. The always-running CI policy job fetches tags and runs these tests, so subsequent main/PR CI will turn red even though this PR passes before tagging; this is not a claim that release.yml itself runs the failing tests.

Make the removal-history validation and baseline fixtures survive release advancement, and verify both pre-tag and post-tag states while retaining the vanished-property guard. Simply deleting the removal exemption before release would break the pre-tag check.

CHANGELOG.md
```diff
@@ -20,6 +20,191 @@ Changes that have merged but not yet shipped are the files under
...
+  `GET /v1/models`, `GET /v1/models/{model}`, `POST /api/agui/{agent_id}` and
+  `POST /api/mcp/caller/{conversation_id}`. Before you upgrade, check whether
+  anything still calls one — a stock OpenAI client, an AG-UI front end, a
+  gateway pointed at `/v1`, or a sandbox answering request-defined tools. Each
+  now gets the ordinary unmatched-path answer, not a dialect error: `/v1` is
+  404 for every request, and `/api/agui/…` is 404 for an authenticated JSON
+  caller, 401 without a key and 406 for the `Accept: text/event-stream` an
+  AG-UI client sends, so a caller that only checks for a 404 will read the
+  other two as something else. Port it to conversations
+  (`POST /api/conversations`, then `/prompt`, `/events` and `/stream`) or to an
```
BinaryBourbon, line 38: [P2] Send migrating clients to the existing /prompts endpoint

The replacement API is `POST /api/conversations/{id}/prompts` (plural), as declared in `apps/fountain/lib/fountain_web/router.ex:593` and the contract. There is no `/api/conversations/{id}/prompt` route, so a caller following this breaking-release migration instruction gets 404 instead of submitting a prompt. Change `/prompt` to `/prompts`; spelling out the full method/path would also make the per-conversation scope clear.

BinaryBourbon (changes requested):
Two independent adversarial reviews of all 24 changed files agree on changes requested: the release tag makes the Swift policy tests fail, and the migration instructions name a nonexistent prompt endpoint (details inline).

Validation: all 29 Swift-generation tests pass against the current v0.17.1 baseline. In an isolated clone of this exact head, adding the intended v0.18.0 tag makes the same suite fail two tests; removing that local-only tag restores 29/29 passing. All 13 fragments roll into CHANGELOG.md byte-for-byte, and changelog, conflict-marker and diff checks pass. Current GitHub CI, including CI required, is green. The pinned-image boot step is intentionally skipped pending release; I did not build or publish a release image.

jhgaylor:
Both findings accepted and reproduced independently. Neither is fixable on
this branch, so this PR will be **closed and the roll re-cut** from `main`
once the two fixes land — re-rolling from fragments is cleaner than
hand-editing a bot-generated release branch, and `release-bump.yml` refuses
to open a new one while `release/v0.18.0` exists.

**The Swift policy tests → #2314.** Reproduced in an isolated clone of this
head: 29 pass, `git tag v0.18.0 HEAD` makes exactly the two named tests fail,
deleting the tag restores them. The removal table cannot be pruned around it,
which the review anticipated:

| | entry present | entry pruned |
|---|---|---|
| pre-tag (v0.17.1 baseline) | pass | fail — 53 errors, generation raises |
| post-tag (v0.18.0 baseline) | fail | pass |

There is no single valid state, because the tag is created from the merged
tree — so that tree is checked both before the tag exists and again, as
`main`, after. #2314 therefore tolerates an aged-out entry and asserts it is
*inert*, pays for that with a ceiling test, and replaces the release-bound
baseline literals with assertions about the mechanism. The vanished-property
guard is retained and verified to still bite post-tag, and all four cells
above are verified.

**The migration path → #2312.** Correct, and worse than a typo: the route is
`POST /api/conversations/{conversation_id}/prompts`, so a client following a
breaking-release instruction would have got a 404 where it expected to submit
a prompt. One correction to the suggestion — the contract spells the
parameter `{conversation_id}`, not `{id}`, so that is what the fragment now
uses, with the method and full path on all three follow-on endpoints as
suggested. Since one path was wrong I re-checked every other claim in that
fragment: the three live paths are in the contract, the five retired ones are
absent from it and answer 404/401/406 on the deployed build, and all four
migration-page URLs return 200.

**Two notes on the validation, since it was thorough enough to be worth
matching:**

- The "pinned-image boot step is intentionally skipped pending release" is
  right, and its blast radius is larger than this PR: `/health` and
  `/health/ready` are the *only* questions `release.yml` asks the built image,
  so nothing in the release pipeline checks that the retirement holds in the
  artifact. #2316 adds the six keyless retirement assertions there. It is
  explicitly marked do-not-merge until this pin advances — v0.17.1 still
  serves all five paths, which I confirmed by booting that tag's own quick
  start and watching the new block exit 1 on it.
- #2308 has since landed a fragment (`2294-checkpoint-webhook-stage.md`); it
  is purely additive and correctly needs no upgrade note, so the re-cut
  release will simply include it.

BinaryBourbon:
Closing in favour of a fresh roll. Both review findings are fixed on `main`
and neither could be fixed here — this branch had already folded the wrong
migration path into `CHANGELOG.md`, and re-rolling from fragments is cleaner
than hand-editing a bot-generated release branch.

- #2314 — the Swift baselines now survive their own release. Verified on
  merged `main` (`4baf774f`) in an isolated clone: 32 pass pre-tag, **32 pass
  with `v0.18.0` tagged**, 135 in the full CI policy suite post-tag. That last
  one is the check that matters, because CI on the PR could only ever run
  pre-tag.
- #2312 — `POST /api/conversations/{conversation_id}/prompts`, with every
  other path and all four migration-page URLs in that fragment re-verified,
  plus the missing #2303 fragment.
- #2316 — the boot check's retirement assertions, in draft until this pin
  advances. CI there reproduced the expected failure exactly (`POST
  /v1/chat/completions -> 401 (expected 404)` against the v0.17.1 image).

`release-bump.yml` refuses to open a new release PR while `release/v0.18.0`
exists, so the branch goes with this. The re-cut will additionally carry
#2308's and #2313's fragments, both of which landed since and both correctly
`Added`/`Fixed` rather than breaking.

PR #2312, opened 2026-09-15
docs(changelog): correct the migration path, and cover #2303

BinaryBourbon:
Two changelog fragment fixes, both found reviewing the v0.18.0 release roll
(#2311). Fragment-only, no code.

**The upgrade note named a route that does not exist.** It sent migrating
clients to `/prompt`; the route is `POST
/api/conversations/{conversation_id}/prompts`, plural
(`router.ex:593`, and the same spelling in the contract). A client following
a breaking-release migration instruction would have got a 404 where it
expected to submit a prompt — the worst place for this particular typo. The
controller action is `:prompt`, singular, which is where the wrong spelling
came from.

Since I got one path wrong I re-checked every other claim in that fragment
rather than just this one:

| Named in the fragment | Verified |
|---|---|
| `POST /api/conversations` | in the contract |
| `POST /api/conversations/{conversation_id}/prompts` | in the contract (was `/prompt`) |
| `GET /api/conversations/{conversation_id}/events` | in the contract |
| `GET /api/conversations/{conversation_id}/stream` | in the contract |
| The five retired paths | absent from the contract, and answering 404/401/406 on the deployed build |
| The four migration-page URLs | all 200 on the live site |

All three live follow-on paths are now spelled with their method and their
real `{conversation_id}` parameter instead of a bare suffix, so the
per-conversation scope is explicit.

**#2303 had no fragment.** It is the third Swift compatibility rule —
generation refuses a model that silently stops declaring a property the last
release published — and its sibling #2284 carries one. The release would
otherwise describe one half of that pair and not the other. Written as
`Fixed`, matching #2284's entry.

**Checks.** `scripts/changelog.py check` (15 fragments, 22 entries),
`conflict-markers.py`, `git diff --check`. Confirmed #2308's new
`2294-checkpoint-webhook-stage.md` is purely additive and correctly needs no
upgrade note of its own.

The v0.18.0 roll has to be re-cut after this lands, because #2311 already
folded the wrong text into `CHANGELOG.md`. Re-rolling from fragments is
cleaner than hand-editing a bot-generated release branch.

Refs #2252, #2303, #2296.


jhgaylor (approved):
Adversarial review of both changed fragments at 37516555: no blocking findings. The four native method/path pairs agree with the Phoenix router and committed API contract; the Swift announcement matches the released-property guard, its deliberate-removal exception and regression coverage.

Validation in an isolated checkout: changelog check (15 fragments, 22 entries), conflict-marker and diff checks, all 29 Swift-generation tests against the v0.17.1 baseline, and a v0.18.0 changelog preview confirming the corrected prompts path and exactly one #2303 entry. Current required CI passes. The release roll must consume these fragments after merge, as the PR describes.

PR #2313, opened 2026-09-15
fix(api): render pending_requests as [] on the list and create responses

### LINKED ISSUE #2305: GET /api/conversations omits pending_requests, which Conversation declares

BinaryBourbon:
## Summary

`GET /api/conversations` declared `pending_requests` on the `Conversation`
schema but never rendered it, and the schema guardrail's exhaustive check
carried an `@unrendered` exemption citing #2305 to stay green. This takes
the cheap route named in the issue rather than splitting the schema:

- `FountainWeb.ConversationJSON.data/1` now defaults `pending_requests` to
  `[]`. `index/1`, the create response (`show/2` with `resumed:`), and the
  plain `show/2` used by `labels` and `reapply` all build on `data/1`, so
  they all get the key. The by-id `show/2` clause that receives a
  `pending_requests:` assign still overwrites it with the real list — that
  is the only caller that pays the query.
- Removed the `{"GET /api/conversations", "pending_requests"} => "#2305"`
  entry from `schema_guardrail_test.exs`'s `@unrendered` map (kept the map
  and its mechanism, now empty).
- Updated the `pending_requests` description in `schemas.ex` to say every
  response carrying the `Conversation` schema sends `[]` unless
  `GET /api/conversations/{id}` has a real one, rather than saying the list
  and create response omit the key.
- Regenerated `sdk/typescript/src/generated/openapi.ts` — only the doc
  comment text moved. `sdk/contract/contract.json` has no diff, since it
  strips descriptions and the property's shape (`array`, optional) did not
  change.
- Added a controller test asserting every row of `GET /api/conversations`
  carries `"pending_requests" => []`, and asserted it on the 201 response of
  `POST /api/conversations succeeds below the cap`.

Closes #2305

## Verification

Ran from an isolated worktree against a private test database
(`fountain_test_p2305`), through the pinned toolchain (`mise exec --`):

- `mix test apps/fountain/test/fountain_web/schema_guardrail_test.exs apps/fountain/test/fountain_web/controllers/conversation_controller_test.exs apps/fountain/test/fountain_web/controllers/conversation_labels_test.exs apps/fountain/test/fountain_web/controllers/conversation_reapply_controller_test.exs` — 160 tests, 0 failures.
- Verified by breaking: reverted the `data/1` change only, reran the same
  suite, and got the expected 3 failures (the guardrail's exhaustive pin,
  the new list test, and the "succeeds below the cap" assertion). Restored
  the fix from a copy and reran clean.
- `mix precommit credo format` — both stages passed.
- `python3 scripts/sdk-contract/generate.py` and `--check` — both passed;
  `sdk/contract/contract.json` unchanged, `dist/openapi.json` (gitignored)
  and `sdk/typescript/src/generated/openapi.ts` regenerated.
- `python3 scripts/changelog.py check` — passed.

Not run: `mix precommit --full` / the whole suite, `mix dialyzer`, and
`mix sobelow` (out of scope for this change's touched files; nothing here
touches security-sensitive input handling or introduces new dialyzer
surface). Did not run the Python/Elixir/Swift SDK contract verifiers beyond
the TypeScript generation step above, since the contract itself did not
move.

## SDK release

Labelled `release:skip-sdk`: the only SDK file that moved is the generated TypeScript `openapi.ts`, and only its doc comment for `pending_requests`. `sdk/contract/contract.json` is unchanged, so no client's wire shape moved and nothing releases.


jhgaylor (approved):
Adversarial review of `6552e713341d3df2be0c1ed5b1cbeec6e6c165aa` against base `4daf6bb987704b0191f423adae5877d31bb300f5`: approved; no actionable correctness, tenant-isolation, or API-compatibility defects found.

Reviewed all six changed files: `conversation_json.ex`, `schemas.ex`, the conversation controller tests, the schema guardrail tests, the changelog fragment, and generated TypeScript OpenAPI declarations. Traced list, create/resume, labels/reapply, nested team responses, and by-id permission requests. The shared renderer adds the documented empty array without new queries; the tenant-scoped by-id action preserves actual pending requests. This implements the inexpensive option explicitly described in #2305 and removes its guardrail exemption without weakening the remaining checks. The wire shape stays an optional array; the SDK change is documentation only.

Independent checks: two isolated ExUnit renderer diagnostics passed under Elixir 1.19.2/OTP 28, covering all shared rendering branches, multiple rows, JSON array encoding, and populated/empty by-id overrides; changelog validation passed. Inspected the existing controller regressions for pending requests and second-tenant denial. Current-head GitHub CI and secret scanning passed, including server partitions, coverage, static analysis, API contract, SDKs, and CLI. Database-backed controller tests were verified through CI, not rerun locally.

PR #2314, opened 2026-09-15
fix(sdk-contract): let the Swift baselines survive their own release

BinaryBourbon:
**Cutting v0.18.0 would have turned the next CI run on `main` red.** Found
reviewing the release roll (#2311); this is the blocker for re-cutting it.

Both Swift baselines read the last release tag, so *creating the tag moves
them*. Two tests asserted facts that only held while v0.17.1 was the
baseline:

- `test_removed_properties_entries_name_a_real_removal` required every
  `REMOVED_PROPERTIES` entry to name a property the last release published.
  `AuthMe.onboardingState` is published by v0.17.1 and not by v0.18.0 — the
  release that retires it — so the entry goes stale at exactly the release it
  describes.
- `test_the_released_tag_is_the_baseline_for_both_questions` asserted
  `CatalogMcpServersItem` was absent from the released Swift. True of
  v0.17.1; false of v0.18.0, which generates it (#2251).

Reproduced in an isolated clone of the release branch — never by tagging a
shared checkout: 29 pass, `git tag v0.18.0 HEAD` makes those two fail,
deleting the tag restores them.

### The removal table cannot be pruned around this

| | entry present | entry pruned |
|---|---|---|
| **pre-tag** (v0.17.1 baseline) | pass | **fail — 53 errors**, generation raises |
| **post-tag** (v0.18.0 baseline) | **fail (before this PR)** | pass |

An entry is load-bearing while the release it cites is the baseline and inert
once that release has shipped, and no commit can delete it *at* the release:
the tag is created from the merged tree, so that tree is checked both before
the tag exists and again, as `main`, after. There is no single valid state.

So an aged-out entry is now tolerated and asserted to be **inert** rather than
required to be live. `test_the_removal_history_has_not_grown` carries a
ceiling of 1 to pay for the tolerance, in the shape
`test_the_undescribed_property_table_has_not_grown` already uses, and
`test_an_aged_out_removal_entry_is_inert_and_does_not_fail` pins the aged-out
path on a synthetic entry so it does not depend on which real entry happens
to have expired.

### The baseline test asserts the mechanism, not one release's contents

Non-empty, discriminating (each baseline carries a True and a False, so one
stuck all-one-way cannot pass), and nested types reaching the optionality
baseline as dotted keys. Nothing is lost by dropping the literals: the exact
parser count stays pinned by
`test_the_released_swift_parser_reads_every_published_property`, and the
released-SDK-versus-released-contract asymmetry — the point
`CatalogMcpServersItem` was illustrating — stays pinned synthetically by
`test_a_shape_the_sdk_never_exposed_is_still_an_older_server_shape`.

`test_both_baselines_read_the_release_tag` asserts what that test is *named*
for, by capturing the git calls and requiring the tag in each. Not by
pointing the readers at an older tag: an older contract can carry a shape the
current generator refuses — v0.17.0's does, since #2288 — so that would raise
for a reason unrelated to baselines.

### The guard still bites

Post-tag, dropping `Agent.description` from the contract still fails
generation, so the #2296 rule is intact rather than defanged. Verified all
four cells above, plus: pruning the aged-out entry post-tag passes, so the
follow-up tidy is safe.

### Checks

- `test_swift_generation`: 32 pass **pre-tag and post-tag** in the probe clone
- `python3 -m unittest discover -s scripts/ci`: 135 pass — the same command the
  CI policy job runs, along with `conflict-markers.py` and `changelog.py check`
- `generate-swift.py --check` exits 0: generated Swift is byte-identical; the
  generator change is its table comment

No changelog fragment: this is a guard's own tolerance and its documentation,
with no effect a reader of the release could observe. The lifecycle, and why
it cannot be pruned at the release, is now written down in
`contributing/swift-wire-models.md`.

Refs #2252, #2296.


jhgaylor (approved):
Adversarial review of all three changed files at e9dff599655eba788a86ac8f4235d0dd5659c207 against 4daf6bb987704b0191f423adae5877d31bb300f5: approved. The release-dependent assertions are replaced without changing the generator's compatibility enforcement; aged-out removal entries remain inert, and the documentation matches the table's lifecycle.

In an isolated clone, the approved-base suite reproduced exactly the two post-tag failures reported on #2311. This head passes all 32 Swift-generation tests both before and after a private v0.18.0 tag; post-tag, pruning the removal table also passes all 32. Pre-tag pruning still rejects AuthMe.onboardingState. Independently removing Agent.description is rejected before and after the tag with the ordinary table, an additional stale entry, and a pruned table. The release-tag read probe and existing optionality, escaped-name, parser-count and candidate-self-authorization coverage were inspected.

All 135 CI Python tests pass, generated Swift remains byte-identical, and conflict-marker, changelog and diff checks pass. The first sandboxed full-suite attempt encountered /dev/stderr restrictions in unrelated dead-code tests; rerunning outside the sandbox passed. Current GitHub CI, including CI required and Swift on Linux/macOS, is green. Release simulation used only a disposable local clone; no release was published.

Issue #2300, opened 2026-09-15
Swift SDK: contract-shaped models still handwritten outside Models/


BinaryBourbon:
#2251 closed with every family under `sdk/swift/Sources/FountainKit/Models/` given a disposition. Three contract-shaped things live outside that directory and were listed as unmigrated throughout, explicitly outside #2269's four seams:

| Type | Where |
|---|---|
| `PageMeta` | `Client/APIClient.swift` |
| `APIErrorBody` | `Errors/FountainError.swift` |
| `TeamResource` request bodies | `Resources/TeamResource.swift` |

With #2251 closed there is no longer a tracker naming them, hence this issue. Nothing is broken: they are handwritten and correct. They are simply the remaining places where a wire shape is declared by hand, which is what that campaign set out to remove.

## Worth knowing before starting

Each is a different question, so this is not one change:

- **`PageMeta`** is the cursor envelope (`has_more`, `limit`, `next_cursor`) shared by the list endpoints, and it appears in the contract as an inline `meta` object on each list response rather than as a named schema. Generating it means either picking one response's inline shape as canonical and proving the others match — the generator already has that machinery, since it refuses divergent reused inline shapes — or naming a schema for it server-side first.
- **`APIErrorBody`** is the 422 shape and its friends. #2169 settled that every 422 has one shape, so there is something concrete to generate against, but error bodies are also what `FountainError` maps into typed cases, so the generated part is the payload and the mapping stays behaviour.
- **`TeamResource` request bodies** are small structs local to the resource. These are the most ordinary of the three and probably the place to start.

Whatever lands, the compatibility rules from #2284 apply the moment a type becomes generated: both are keyed on the released Swift and the released contract, so a currently-`Optional` published property will demand a pin. That is the mechanism working, and #2289 is the worked example — it named the two pins `AuthMe` needed before I could look them up.

Related: #2251 (closed inventory), #2269 (the four seams), #2247 (parent tracker).

BinaryBourbon:
Status, one PR per type as the issue anticipated:

- **`TeamResource` request bodies** — landed in #2315.
[...]

PR #2315, opened 2026-09-15
refactor(swift): generate TeamResource's request bodies from the contract

jhgaylor:
Part of #2300.

#2300 named three contract-shaped types that stayed handwritten outside `sdk/swift/Sources/FountainKit/Models/` after #2251. This PR does only the first and most ordinary one: **`TeamResource` request bodies**.

## What changed

- `TeamAddRequest`, `TeamRenameRequest` and `TeamMessageRequest` are now generated from `sdk/contract/contract.json` by `scripts/sdk-contract/generate-swift.py`, the same generator and mechanics used for every other generated resource family (#2251/#2269). They replace the handwritten local `Body: Encodable` structs in `TeamResource.add` and `.message`, and the `["name": name]` dictionary literal in `.rename`.
- The public API of `TeamResource` is unchanged: `add`, `rename` and `message` keep their existing signatures.
- `add` and `message` keep the same encode-when-present, omit-when-nil behaviour they always had (Swift's synthesized `Encodable` uses `encodeIfPresent` for `Optional` stored properties).
- `rename` keeps sending an **explicit JSON `null`** for a `nil` name rather than omitting the key — the previous `["name": name]` dictionary literal encoded `Optional<String>.none` as `null`, not an omitted key, and the doc comment ("`nil`/blank restores the agent's own name") depends on that. The generated model expresses this with `TeamRenameRequest(name: nil)` plus an explicit `body.setNull(.name)`.
- `TeamMessageRequest.labels`, a contract field these methods do not yet expose, types as `[String: JSONValue]` rather than `[String: String]`. Its `additionalProperties` schema allows a null value per key, which a non-Optional `String` can't decode; `JSONValue` (which has a `.null` case) is the faithful type and matches how other open-ended dictionaries in this generator are typed. No caller-visible change: neither method sets `labels` today.
- `contributing/swift-wire-models.md`'s inventory moves the three types into the "Generated" table with a disposition entry, and narrows the handwritten row to just `PageMeta` and `APIErrorBody`.
- Added `changelog.d/2300-swift-team-request-bodies.md`.

## Left out (each its own follow-up, per #2300)

- `PageMeta` (`Client/APIClient.swift`) — the cursor envelope, an inline shape shared across list responses rather than a named schema.
- `APIErrorBody` (`Errors/FountainError.swift`) — the 422 shape; the generated part would be the payload, with the `FountainError` mapping staying behaviour.

Neither needed a server-side schema change for this PR's scope, and none was added.

## Verification

Run from a dedicated worktree off `origin/main`, from the repository root unless noted:

- `mise exec -- python3 scripts/sdk-contract/generate-swift.py` — regenerated `ConversationWire.generated.swift` cleanly (148 added lines: the three new structs, no changes elsewhere).
- `mise exec -- python3 scripts/sdk-contract/generate-swift.py --check` — clean (deterministic).
- `mise exec -- python3 -m unittest discover -s scripts/ci -p test_swift_generation.py` — 29 tests, all pass.
- `swift build` — succeeds.
- `swift test` — 124 tests, all pass, including a new `teamRequestBodiesEncodeGeneratedWireShapesAndPreserveNullVsOmission` in `ResourceWireTests.swift` that encodes all three request bodies and checks the wire keys, the omitted-vs-present-vs-explicit-null states, and the `labels` null-value case.
- `swift test --filter ContractTests` — passes.
- Two of the 124: new `RequestBuildingTests` cases in `ClientTests.swift`, `renameSendsAnExplicitNullForANilNameRatherThanOmittingTheKey` and `renameSendsTheGivenNameWhenPresent`, drive `client.team.rename` through the existing `FakeTransport` mock and assert the encoded request body is exactly `{"name": null}` or `{"name": "Bob"}` — the path the model-level test above doesn't reach. Verified by deleting the `if name == nil { body.setNull(.name) }` lines in `TeamResource.rename`: the nil-name test failed with a clear expected/actual mismatch, and passed again after restoring from a saved copy (not `git checkout`, for the same reason as below).
- `swift format lint --strict --recursive Package.swift sdk/swift/Sources sdk/swift/Tests` (also checked from `sdk/swift` with local paths) — clean.
- `python3 scripts/conflict-markers.py` and `git diff --check` — clean.
- `python3 scripts/changelog.py check` — passes (15 fragments, 22 entries).
- **Verify-by-breaking**: temporarily renamed the shared `case agentID = "agent_id"` coding key to `"agent_id_broken"` in the generated file and reran the new test — it failed with a clear mismatch on the `agent_id` key. Restored the file from a copy saved before the edit (not `git checkout`, since that would have discarded this PR's own changes to the file) and reconfirmed `generate-swift.py --check` reports no diff.

Not run: the full contract propagation probe (`check-propagation.py`), since it needs Go and fetched `sdk/elixir` deps beyond this worktree's scope and this change touches no shared/propagation-fixture schema.


jhgaylor:
**APPROVED** — independent AI adversarial source review of PR #2315.

Reviewed head: `cb96691151243140cb6be951f2d1b8e5e5565420` against base `4daf6bb987704b0191f423adae5877d31bb300f5`.

This verdict is posted as the PR author, `jhgaylor`, because GitHub prohibits authors from submitting formal approval/change-request reviews on their own PRs. It is a standalone source-review verdict, not full Review Loop pipeline approval or merge authorization.

No blocking correctness, API compatibility, serialization, or security regression found. I inspected all eight changed files: the changelog fragment, Swift SDK changelog, Swift wire-model inventory, generator, generated models, `TeamResource`, `ClientTests`, and `ResourceWireTests`; traced the requests through `APIClient` and the server controller/schema; and checked the linked tracking issues and trusted base review instructions.

- [`TeamResource.add`, `.rename`, and `.message`](https://github.com/managoat/fountain/blob/cb96691151243140cb6be951f2d1b8e5e5565420/sdk/swift/Sources/FountainKit/Resources/TeamResource.swift#L16-L62) preserve the existing method signatures, routes, response handling, optional-field omission, and explicit `name: null` for a nil rename. The generated inputs retain required `agent_id`/`prompt` and nullable fields.
- The [`labels` override](https://github.com/managoat/fountain/blob/cb96691151243140cb6be951f2d1b8e5e5565420/scripts/sdk-contract/generate-swift.py#L173-L177) can encode per-key null; existing convenience methods continue to omit labels. Existing generator propagation tests exercise the newly registered roots independently.

Nonblocking documentation correction: the [`rename` comment](https://github.com/managoat/fountain/blob/cb96691151243140cb6be951f2d1b8e5e5565420/sdk/swift/Sources/FountainKit/Resources/TeamResource.swift#L29-L35) implies blank names are sent as null and omission would leave the name untouched. In fact, blank strings are forwarded unchanged, and the [server reads `params["name"]`](https://github.com/managoat/fountain/blob/cb96691151243140cb6be951f2d1b8e5e5565420/apps/fountain/lib/fountain_web/controllers/team_controller.ex#L107-L113), treating a missing name the same as nil. Suggested wording: “Nil sends explicit JSON null to preserve the existing request shape; blank names are forwarded unchanged. Both restore the agent's name.” This does not affect the correct implementation.

Validation run independently in an isolated checkout of the reviewed SHA:

- `python3 scripts/sdk-contract/generate-swift.py --check`: passed.
- Generator unit suite: all 29 tests passed.
- `swift test` with `-Xswiftc -warnings-as-errors`: all 124 tests passed with local-loopback permission, including both products' contract/conformance coverage; existing live-smoke and documented-deviation skips remained in effect. The first final-head run hit the sandbox restriction at `LoopbackTests.swift:146` (`failed("bind")`); a separate localhost bind reproduced `EPERM`, and the same suite passed when permitted to bind locally.
- Strict recursive Swift format lint, conflict-marker check, changelog check, and base-to-head `git diff --check`: passed.
- Three additional temporary transport tests, covering 10 input cases, passed: real public `add` requests with/without optional fields; `rename` with nil, empty/whitespace, ordinary, and escaped/Unicode names; and `message` with omitted, empty, and populated images. They assert exact JSON bodies, methods, paths, and message reply decoding. No production service was called and no PR code was modified.

The latest commit was reassessed after the PR advanced during review. Its [new transport tests](https://github.com/managoat/fountain/blob/cb96691151243140cb6be951f2d1b8e5e5565420/sdk/swift/Tests/FountainKitTests/ClientTests.swift#L60-L78) now directly protect nil and present rename bodies, closing the model-only coverage gap in the earlier head. The temporary probes ran against the preceding head; the production generator and Swift source files are byte-identical at this final head. All listed committed-suite and formatting checks were rerun at the final SHA. The full server suite and full Review Loop verification recipe were not run as part of this SDK-only source review.

BinaryBourbon (approved):
Independent AI adversarial review — **APPROVE**.

Reviewed commit `cb96691151243140cb6be951f2d1b8e5e5565420` against base `4daf6bb987704b0191f423adae5877d31bb300f5`. I independently reviewed all eight changed files: generator, generated models, TeamResource, both test files, contributor inventory, SDK changelog, and changelog fragment. I traced the request bodies through APIClient, the committed contract, server controller/context, existing tenant/label denial tests, and sibling client behavior. No blocking correctness, compatibility, or security regression found. Existing method signatures and payload omission/null behavior are preserved.

Low, nonblocking documentation correction: [TeamResource.swift lines 29–31](https://github.com/managoat/fountain/blob/cb96691151243140cb6be951f2d1b8e5e5565420/sdk/swift/Sources/FountainKit/Resources/TeamResource.swift#L29-L31) imply an omitted name leaves the current name untouched. [TeamController.update](https://github.com/managoat/fountain/blob/cb96691151243140cb6be951f2d1b8e5e5565420/apps/fountain/lib/fountain_web/controllers/team_controller.ex#L107-L113) passes `params["name"]` directly, so omission also supplies nil and clears the name today. Describe the explicit null as preserving the previous SDK payload; remove the claim about omission preserving the current name.

Validation I ran in an isolated snapshot of this commit: generator `--check`; all 29 Swift-generation Python tests; Swift debug compilation with warnings as errors and all 124 tests, including ContractTests and ConformanceTests; strict Swift format lint; conflict-marker, changelog, and diff-whitespace checks. The existing conformance matrix skip remains. The first full Swift run failed only because the sandbox denied the loopback socket bind; the identical already-built suite passed with local socket access. No PR source was modified.

Limits: server behavior and denial paths were source-reviewed, not executed locally. This is a standalone independent review, not certification of the full Review Loop pipeline.

PR #2316, opened 2026-09-15
test(ci): the boot check asserts the retired dialects on the image

jhgaylor:
> [!IMPORTANT]
> **Do not merge before v0.18.0 is released and main's pin advances.** This
> asserts shipped behaviour, and the release main pins today does not have it.
> Details below — verified, not assumed.

`release.yml`'s only question of the built image was `/health` and
`/health/ready`, so nothing in the release pipeline checked that the ADR 0057
retirement actually holds in the artifact someone pulls.
`protocol_retirement_test.exs` pins the whole matrix, but a suite pass is not
evidence about the image: the tree it tested and the published release are
different objects, and this repository has shipped a prod-only no-op that
every test passed (the runtime image had no coreutils).

Six requests, none needing an API key, so the whole matrix that does not need
an account is covered for the cost of six curls:

| Request | Answer |
|---|---|
| `POST /v1/chat/completions` | 404 |
| `GET /v1/models` | 404 |
| `GET /v1/models/gpt-4o` | 404 |
| `POST /api/agui/{id}` keyless | 401 |
| `POST /api/agui/{id}` + `Accept: text/event-stream` | 406 |
| `POST /api/mcp/caller/{id}` keyless | 401 |

All three answers are load-bearing and deliberately different — the
correction ADR 0057 carries. `/v1` is unrouted, so it is refused before
authentication runs; the `/api` paths reach the API pipeline and get
authentication's and content negotiation's answers first. A client probing
for a bare 404 to detect the cutover misreads the other two, so a regression
in any *single* one of them matters. The `/v1` response body is asserted too,
since a status code alone does not rule out a dialect controller still being
mounted somewhere.

`APP` replaces the hardcoded base URL, defaulting to the compose app exactly
as before, so the block can be pointed at a running server rather than
duplicated.

### Why the ordering matters

The script checks the **pinned** image. On a PR and on main that pin is the
*last release*, so an assertion here is a statement about shipped behaviour —
only on the tag does it describe the tree being released. v0.17.1 still
serves all five retired paths, so merging this before the pin advances reds
main immediately.

Verified in both directions rather than reasoned about:

- Against a server running the retirement: all six pass, **exit 0**.
- Against a **real v0.17.1 instance**, booted from that tag's own quick start:
  **exit 1** on the first assertion — `POST /v1/chat/completions -> 401
  (expected 404)`, which is what v0.17.1 answers a keyless call. Instance torn
  down with `down -v` afterwards.

That is the same failure shape as #2314: a check that asserts a fact about the
current tree while running against the last release's artifact. The header now
says so, so the next assertion added here is placed with the release cycle in
mind.

### Checks

`python3 -m unittest discover -s scripts/ci` (132), `conflict-markers.py`,
`changelog.py check`, `git diff --check`, `bash -n` and `shellcheck -S warning`
on the script.

No changelog fragment: CI plumbing, nothing a reader of the release observes.

Refs #2252.


BinaryBourbon:
CI has reproduced the expected failure, so the red here is the point rather
than a defect:

```
/health -> 200
/health/ready -> 200
the retired compatibility dialects (ADR 0057):
  POST /v1/chat/completions -> 401 (expected 404)
```

That is the same line, from the same assertion, that a locally booted v0.17.1
instance produced — 401 is what v0.17.1 answers a keyless call to a dialect it
still serves. The job runs against the **pinned** image, which on a PR is the
last release, so this cannot go green until main pins v0.18.0.

Moved to draft so the red is not mistaken for a broken change. Ready to mark
for review the moment the release lands and the pin advances; nothing in the
diff needs to change.

jhgaylor:
**CHANGES REQUESTED** — independent AI adversarial review of `e22887999aebbf7ba249125ec461d8e63f112dd7` against base `4daf6bb987704b0191f423adae5877d31bb300f5`.

This verdict is posted as `jhgaylor` in a comment because GitHub prohibits authors from submitting formal approving or changes-requested reviews on their own PRs. It is a standalone review, not a Review Loop pipeline approval.

**Blocking: the current head fails against the release it actually boots.** The new [first retirement assertion](https://github.com/managoat/fountain/blob/e22887999aebbf7ba249125ec461d8e63f112dd7/scripts/compose-boot-check.sh#L98) requires 404, but [.env.compose.example still pins v0.17.1](https://github.com/managoat/fountain/blob/e22887999aebbf7ba249125ec461d8e63f112dd7/.env.compose.example#L54). The selected [Compose CI job](https://github.com/managoat/fountain/actions/runs/35017728941/job/104545375891) confirms this exact failure: `pinned v0.17.1; mix.exs is 0.17.1`, both health probes return 200, then `POST /v1/chat/completions -> 401 (expected 404)` and exit 1. The [workflow runs this script whenever that pinned image exists](https://github.com/managoat/fountain/blob/e22887999aebbf7ba249125ec461d8e63f112dd7/.github/workflows/ci.yml#L1498), so this is a deterministic release-order blocker, not an unexplained CI failure.

The PR description correctly acknowledges the dependency. Keep it unmerged until the release containing retirement is published and the pin advances; update the PR against that pin and obtain a passing Compose run before reconsidering approval. Do not weaken the expected statuses to make v0.17.1 pass.

**Nonblocking operational note:** the new [`APP` override](https://github.com/managoat/fountain/blob/e22887999aebbf7ba249125ec461d8e63f112dd7/scripts/compose-boot-check.sh#L30) changes the probe destination while the pre-existing `.env` replacement, key generation and `docker compose up -d` still run. An isolated full-script reproduction with `APP=https://review-target.invalid` confirmed the local `.env` is replaced and Docker is invoked; a failing Docker command prevents all remote probes. Consider documenting the disposable-checkout requirement or splitting out the reusable probes. The unconditional setup itself predates this PR.

Validation: reviewed the entire changed file, both CI/release callers, Compose pin/configuration, ADR 0057, router/extension dispatch, and `protocol_retirement_test.exs`; `bash -n`, `shellcheck`, and `git diff --check` passed. Isolated command-stub scenarios passed for the expected response matrix, five individual endpoint status failures, transport failure, unexpected dialect body, and the two `APP` behaviors above. These validate shell control flow, not a running retirement image. I inspected the actual failed CI job logs; I did not run a local Docker deployment or the full server suite. The keyless MCP 401 probe by itself cannot establish that its handler was removed; authenticated/route-removal assertions remain covered by the source regression suite.

BinaryBourbon (changes requested):
**CHANGES REQUESTED — release prerequisite remains unverified.**

Independent AI adversarial review of the entire PR (the single changed path, `scripts/compose-boot-check.sh`) at `e22887999aebbf7ba249125ec461d8e63f112dd7`, against recorded base `4daf6bb987704b0191f423adae5877d31bb300f5`. This is a standalone review, not certification of the Review Loop pipeline.

**Blocking: obtain a passing retirement check against the new release artifact before landing.** The [first new assertion](https://github.com/managoat/fountain/blob/e22887999aebbf7ba249125ec461d8e63f112dd7/scripts/compose-boot-check.sh#L98) requires 404, while [this reviewed tree pins v0.17.1](https://github.com/managoat/fountain/blob/e22887999aebbf7ba249125ec461d8e63f112dd7/.env.compose.example#L54). I independently inspected the [Compose CI log](https://github.com/managoat/fountain/actions/runs/35017728941/job/104545375891): both health probes pass, then `POST /v1/chat/completions -> 401 (expected 404)` exits 1. That is consistent with the v0.17.1 router still mounting `/v1` behind authentication.

The draft and PR description correctly acknowledge this dependency. During this review, [main advanced its pin to v0.18.0](https://github.com/managoat/fountain/blob/a9c94ae6e39d75cdb53509c8a19e9525ca3f3f5c/.env.compose.example#L54), but the [release workflow](https://github.com/managoat/fountain/actions/runs/35019595505) is still in progress. The old red run does **not** test the new merge candidate. Keep this unmerged until the retirement image is published and the current merge candidate's Compose check passes against it. Preserve the expected statuses; no rebase solely for the merge queue is needed.

**Nonblocking operational note:** the new [`APP` override](https://github.com/managoat/fountain/blob/e22887999aebbf7ba249125ec461d8e63f112dd7/scripts/compose-boot-check.sh#L30) changes the probe destination, but the script still replaces local `.env`, generates new keys, and invokes `docker compose up -d`. My isolated reproduction confirmed those side effects with a remote `APP`, and confirmed a Docker failure prevents all remote requests. The unconditional setup predates this PR, but the newly advertised existing-server use should either document the disposable-checkout requirement or separate reusable probes from Compose setup.

Validation I performed: `bash -n`, full `shellcheck`, and `git diff --check` passed. Thirteen isolated command-stub scenarios passed: expected responses; each of six endpoint status failures; request/body transport failures; an unexpected dialect body; old-image responses; and both remote-`APP` behaviors above. These verify shell control flow, not a real image. I traced both workflow callers, Compose configuration, ADR 0057, router/auth/content-negotiation/extension dispatch, and the retirement regression suite. I did not run Docker or the Elixir suite. The keyless MCP 401 alone cannot prove its handler is gone; authenticated and route-removal assertions remain source-suite coverage.

BinaryBourbon:
Unblocked and rebased onto `main` at v0.18.0, so the branch now pins the
image its own assertions describe (it was still carrying `v0.17.1`, which is
what made the earlier run red). Out of draft; the diff is unchanged.

Independently of CI, the same six assertions were run against the published
`ghcr.io/managoat/fountain:v0.18.0` as part of the release verification, and
all six pass on the artifact.

BinaryBourbon (approved):
Independent AI adversarial review 1 — correctness and regression coverage.

APPROVE for `a28aede4086cf2244783b358add233ee181ed1c8` against base `e0b1cec92aea47793e02380bbb36b8a549edd2fc`. Reviewed the entire PR: the sole changed file, [`scripts/compose-boot-check.sh`](https://github.com/managoat/fountain/blob/a28aede4086cf2244783b358add233ee181ed1c8/scripts/compose-boot-check.sh), plus both calling workflows, image pin, ADR 0057, router/authentication/content negotiation, error rendering, and existing retirement tests. No actionable correctness defect found.

The earlier publication/image-validation objection is resolved: [v0.18.0 is published](https://github.com/managoat/fountain/releases/tag/v0.18.0), and the [current Compose job](https://github.com/managoat/fountain/actions/runs/35020548209/job/104554884638) actually pulled the pinned image and passed all six retirement status assertions plus the exact unmatched-path response body; it did not merely skip a missing image.

Validation: `bash -n` and ShellCheck passed on the pinned script. An isolated loopback HTTP fixture ran that unchanged script with real curl and Docker stubbed: the expected responses passed; individually changing each of the six statuses, replacing the final error body, and dropping connections during a status probe or body fetch each failed. Request methods, JSON bodies, and Accept headers were checked.

Scope: the image probes cover keyless responses. Authenticated retirement and surviving native routes remain covered by the existing server tests, which I inspected but did not rerun locally. This is an independent source review, not a claim to have run the complete Review Loop verification pipeline.

BinaryBourbon (approved):
Independent AI adversarial review — release integration, security, and shell failure behavior.

Reviewed the entire PR (the sole changed path, `scripts/compose-boot-check.sh`) at `a28aede4086cf2244783b358add233ee181ed1c8`, against base `e0b1cec92aea47793e02380bbb36b8a549edd2fc`. **No concrete blocking defects found.**

The [previous release prerequisite](https://github.com/managoat/fountain/pull/2316#pullrequestreview-5215418708) is now satisfied: [v0.18.0 is published](https://github.com/managoat/fountain/releases/tag/v0.18.0), both reviewed trees pin it, and the [current merge candidate's Compose job](https://github.com/managoat/fountain/actions/runs/35020548209/job/104554884638) actually ran against that image. I inspected the log: both health checks, all six retirement statuses, and the unmatched-path body assertion passed. This is execution evidence, not an image-missing skip.

I traced both CI/release callers, tag/pin selection, the router, JSON negotiation, authentication, error rendering, and the existing retirement regression suite. The added requests carry no API key or tenant data. Quoted arguments preserve the request boundaries, and mismatched responses or failed transfers terminate the check.

Validation performed: `bash -n`, full `shellcheck`, and `git diff --check` passed. Eighteen isolated runs of the exact script with command stubs passed: default/empty/overridden `APP`; each of six incorrect statuses; each of six transport failures even when curl prints the expected status; wrong body; body-transfer failure; and Compose failure stopping before requests.

Limits: those local tests check shell behavior; I did not rebuild an image or run Elixir locally. The keyless `/api` checks do not independently prove authenticated handlers are absent; the source regression suite covers that distinction. The earlier nonblocking `APP` caveat remains: it changes the probe target while retaining local `.env` replacement and Compose setup, so use a disposable checkout. This is a standalone review, not certification of the full Review Loop pipeline.

PR #2318, opened 2026-09-15
feat(api): describe the SSE frame as StreamLogEvent

### LINKED ISSUE #2297: The SSE frame shape is undescribed in the contract
All three stream operations declare their response as `text/event-stream` with a bare `%Schema{type: :string}`:

- `GET /api/conversations/:id/stream` (`conversation_controller.ex`)
- `GET /api/events` (`events_controller.ex`)
- `GET /api/team/stream` (`team_controller.ex`)

So the frame payload — what every client actually parses — is documented in prose in the operation descriptions and nowhere else. The REST `LogEvent` schema is **not** that shape:

| | REST `/events` row | conversation + events stream | team stream |
|---|---|---|---|
| `id` | yes (required) | **absent** (it is the frame's `id:` line) | **absent** |
| `duration_ms` | yes | yes | **absent** |
| `conversation_id` | no | yes | yes |
| `agent_id` | no | no | yes |

## What this costs today

Two things, both currently absorbed rather than solved:

1. **`EXTRA_PROPERTIES` exists for it.** The Swift generator carries two entries — `LogEvent.conversation_id` and `agent_id` — precisely because there is no schema to read them from (#2295). The table is bounded at two by a ceiling test and a second test retires an entry once the contract describes its key, so describing the frame would delete both entries on its own.
2. **`LogEvent.id` and `ts` are pinned for a live shape.** One published Swift `LogEvent` decodes both the REST row and the SSE frame, so those pins are not old-server compatibility — they are the only reason a frame without `id` decodes at all. That is load-bearing behaviour resting on a compatibility table.

Adding the two stream fields to the REST `LogEvent` was considered and rejected: it makes the spec promise fields `GET /api/conversations/:id/events` never returns, which is the #1418 defect class (see #2298).

## Shape of the work

A `StreamLogEvent` schema (no `id`, optional `duration_ms`, plus `conversation_id` and `agent_id`) wired into the three operations' `text/event-stream` responses. OpenAPI allows a schema for any media type, so the frame becomes typed for every client rather than just described.

It forces one SDK decision that should be settled before starting, not during: **Swift today has one public `LogEvent` for both shapes.** Either it keeps serving both (generated from one schema, pinned for the other, roughly today's arrangement but with the pins justified by a documented shape) or the SDK grows a second public type and `TurnFollower`/`EventStream` pick per route. The second is more honest and a public API addition; the first is invisible to consumers.

Also worth checking while in there: the other SDKs parse the same frames, so whichever shape is documented, `check-propagation` will want the generated types regenerated for TypeScript, Python, Elixir and the CLI.

Context: settled as out of scope for #2269 in [this comment](https://github.com/managoat/fountain/issues/2269#issuecomment-5684176985), with the reasoning. Related: #2295, #2251.

BinaryBourbon:
## Summary

- Added `StreamLogEvent` to `apps/fountain/lib/fountain_web/schemas.ex`, describing the SSE frame payload every stream client parses. It reuses `LogEvent`'s property map (minus `id`, which lives in the frame's `id:` line, never the JSON body) so the two schemas cannot drift on the fields they share, and adds `conversation_id`/`agent_id` on top.
- Wired it into the `text/event-stream` `ok:` response of the three stream operations, replacing the bare `%OpenApiSpex.Schema{type: :string}`:
  - `GET /api/conversations/:id/stream` (`conversation_controller.ex`)
  - `GET /api/events` (`events_controller.ex`)
  - `GET /api/team/stream` (`team_controller.ex`)
- `duration_ms`, `conversation_id` and `agent_id` are all optional, each with a description saying which of the three streams actually sends it.

## A correction to the issue's comparison table

While wiring this up I checked each controller's `write_event` against the issue's table and found one row doesn't match the current code: `GET /api/conversations/:id/stream` (`conversation_controller.ex`) sends neither `duration_ms` nor `conversation_id` today — only `GET /api/events` and `GET /api/team/stream` do. The issue's table lists `conversation_id` as present on "conversation + events stream" and `duration_ms` the same way, which reads as both.

I described the schema to match what each stream *actually emits* rather than the table, since the "controller renders what the schema declares" guard (`schema_guardrail_test.exs`) and the new focused frame-validation tests below would otherwise be validating something untrue. Making `GET /api/conversations/:id/stream` also send `conversation_id` (its URL already names the conversation, so it's arguably redundant there) is a behavior change outside this issue's scope — happy to file it separately if wanted, since the issue's table implies it was intended.

## Swift: one public `LogEvent`, sourced from two schemas now

Kept **one public `LogEvent` type** for both shapes, per the issue's stated decision. `conversation_id`/`agent_id` used to live in a handwritten `EXTRA_PROPERTIES` table in `scripts/sdk-contract/generate-swift.py` because no schema described them. Now that `StreamLogEvent` does, `fields()` grew an `owner == "LogEvent"` branch that reads their type straight from `StreamLogEvent`'s properties in the contract, and `EXTRA_PROPERTIES` is empty. The ceiling test in `scripts/ci/test_swift_generation.py` now holds at 0 instead of 2.

The existing disappearance guard (`compatibility_failures`, keyed off the last release's published Swift properties) still protects `LogEvent.conversationID`/`.agentID` from silently vanishing — it doesn't care where a property's shape came from, only whether it's still generated. Added `test_a_stream_field_disappearing_from_the_contract_is_caught` to prove that directly (removing `StreamLogEvent.conversation_id` from the contract now raises the same "shipped ... and is no longer generated" error that used to depend on `EXTRA_PROPERTIES`), and `test_stream_only_log_event_properties_reach_the_output` to prove the sourcing is load-bearing rather than vestigial.

`released_requiredness()` rebuilds a `Generator` against the *last released* contract as a baseline, which predates `StreamLogEvent` entirely — the new branch guards the schema lookup so a pre-#2297 contract (no `StreamLogEvent` at all) is a no-op there, which is the correct history for that baseline.

Updated `contributing/swift-wire-models.md`'s `LogEvent` row and the `EXTRA_PROPERTIES` paragraph to describe the new state.

## Server tests

The schema-render guard (`FountainWeb.SchemaGuard`, attached in `test_helper.exs`) only validates `application/json` responses — an SSE stream is `text/event-stream`, so it was never covered. Added:

- `FountainWeb.SchemaGuard.validate_value/2`, a small addition reusing the guard's existing `Cast` logic, for validating an already-decoded value (an SSE frame parsed out of a chunked response body) against a named schema.
- One assertion per stream, in the existing test that already parses a live frame out of `resp_body` via regex + `Jason.decode!`, validating it against `StreamLogEvent`:
  - `conversation_controller_test.exs` ("replays existing log events when wait=false...")
  - `events_stream_test.exs` ("events from owned conversations...")
  - `team_stream_test.exs` ("events from every teammate arrive on one connection, labelled")

Verified by breaking: temporarily changed `team_controller.ex`'s `write_event` to send `ts: 12_345` instead of a real timestamp. The new assertion in `team_stream_test.exs` failed with `{:error, "#/ts: Invalid string. Got: integer"}`; every other test still passed. Restored from a saved copy of the file (not `git checkout`) and reran — clean.

The exhaustive schema guard (`schema_guardrail_test.exs`) doesn't pin any of the three stream operations in its `@pins` list, so it needed no changes and no exemption.

Also extended `schema_enum_guardrail_test.exs`'s `@derived` map for `StreamLogEvent.kind`/`.state`, which reuse `LogEvent`'s domain lists (`Fountain.Conversations.LogEvent.kinds/0`/`.states/0`) — a test I hadn't anticipated but the suite caught immediately.

## Verification

Ran from a private worktree with a private test database (`fountain_test_p2297`), across all three commits (initial schema, the server-down-frame + route-name-paragraph fix, and the `StreamSignal` union + route-name-text fix from Review Loop's three change-requested reviews).

- `mise exec -- mix compile --warnings-as-errors` — ok
- `mise exec -- mix test apps/fountain/test/fountain_web` — 1757 tests, 0 failures (first round)
- `mise exec -- mix test` (full root suite: fountain, fountain_buzz, fountain_slack, fountain_support, fountain_google, fountain_microsoft) — 0 failures (first round)
- `mix precommit toolchain conflict-markers compile deps format credo sobelow` — PASSED (7 stages, first round)
- `MIX_ENV=dev mix dialyzer` — 0 errors (first round)
- `mise exec -- mix test` on the three stream test files + `sse_stream_test.exs` + `schema_guardrail_test.exs` + `schema_enum_guardrail_test.exs` — 165 tests, 0 failures (re-run after each later commit)
- Break/restore, three times: the `ts: 12_345` break (first round), `stream: nil` on the server-down frame (second commit) — failed with `#/stream: null value where string expected`, restored from a saved copy — and `StreamSignal`'s enum (`changed` → `unchanged`, third commit) — failed all three new signal-frame assertions with the expected Cast enum error, restored from a saved copy. All three confirmed a real mismatch, then reran clean.
- `mise exec -- python3 scripts/sdk-contract/generate.py` then `generate.py --check` — clean, no stale output, re-run after every commit (200 schemas as of the `StreamSignal` addition)
- `scripts/sdk-contract/build.sh --check` — ok, no new omissions entry needed (existing operations, not a new endpoint); confirmed again after the `oneOf`
- `python3 -m unittest discover -s scripts/ci -p test_swift_generation.py` — 30 tests, ok, unaffected by the `oneOf` (the Swift generator walks schemas reachable from `RESOURCE_ROOTS`, never the raw operation-response tree)
- `npm --prefix sdk/typescript ci && npm --prefix sdk/typescript run build && npm --prefix sdk/typescript run verify-contract` — ok both before and after the `oneOf`, no manifest change needed
- `(cd sdk/python && python3 scripts/verify_contract.py)` — ok both before and after the `oneOf`, no manifest change needed
- `(cd sdk/elixir && mise exec -- mix test test/contract_test.exs)` — 1 test, ok both before and after the `oneOf`
- `swift build` — ok
- `swift test --filter ContractTests` — ok both before and after the `oneOf`
- `swift test -Xswiftc -warnings-as-errors` (full Swift suite) — 121 tests passed, 1 intentionally skipped (pre-existing conformance matrix skip), re-run after the `oneOf`
- `swift format lint --strict --recursive Package.swift sdk/swift/Sources sdk/swift/Tests` — clean
- `python3 scripts/changelog.py check` — ok

No consumer of the contract needed a code change to handle the `oneOf` — TypeScript, Python, Elixir and Swift all verified against it without modification. Everything I set out to run, ran and passed. I did not touch `apps/fountain/lib/fountain/conversations/` or the `pending_requests` work in `schemas.ex`/`conversation_json.ex` (#2305), per the brief.

Closes #2297.

## SDK release

Labelled `release:skip-sdk`: the TypeScript change is the regenerated `openapi.ts` gaining the `StreamLogEvent` component type. No handwritten TypeScript code reads it and no existing exported type changed, so it ships with the next TypeScript release rather than forcing one here.


apps/fountain/lib/fountain_web/schemas.ex
```diff
@@ -2020,6 +2020,57 @@ defmodule FountainWeb.Schemas do
...
+          "fields as `LogEvent` minus `id` — the frame's id travels in the SSE `id:` " <>
+          "line, never the JSON body — plus `conversation_id` and `agent_id`, which the " <>
+          "REST log feed never sends because its URL or list item already names the " <>
+          "conversation.\n\n" <>
+          "Not every stream sends every optional field here; see each property's own " <>
+          "description for which of `GET /api/conversations/:id/stream`, " <>
+          "`GET /api/events` and `GET /api/team/stream` include it.",
+      type: :object,
+      properties:
+        LogEvent.schema().properties
```
jhgaylor, line 2045: [P2] Allow the synthetic server-exit frame's null stream

This inherits REST `LogEvent.stream` as a non-nullable string, but `ConversationController.write_server_down_event/2` emits `stream: nil` whenever a monitored server shuts down or crashes. Therefore the new `StreamLogEvent` contract, and its generated TypeScript type, reject a real `event: stage` frame precisely when clients need the reconnect signal. I compiled this head's schema and actual renderer in an isolated pinned-Elixir diagnostic; both `:shutdown` and `:killed` produce an OpenApiSpex `:null_value` error at `[:stream]`. Override `stream` as nullable in the SSE schema, regenerate, and add schema validation to the existing crash/shutdown tests in `sse_stream_test.exs`.

apps/fountain/lib/fountain_web/controllers/events_controller.ex
```diff
@@ -61,7 +62,7 @@ defmodule FountainWeb.EventsController do
       ]
     ],
     responses: [
-      ok: {"SSE stream", "text/event-stream", %OpenApiSpex.Schema{type: :string}}
+      ok: {"SSE stream", "text/event-stream", Schemas.StreamLogEvent}
```
jhgaylor: [P2] Include collection-change signals in the stream response schema

This response now declares every event body as `StreamLogEvent`, which requires `kind` and `ts`. However, a sidebar change sends `event: conversations` with only `{"reason":"changed"}` (`sse_loop/2`, line 197). The matching declaration in `TeamController` also covers `team` and `schedule` signals with that same body. All three signal payloads fail the newly declared schema for missing `kind` and `ts`; generated consumers cannot decode these valid frames using the response type. Declare a signal schema and a union with `StreamLogEvent` for the events/team responses, then validate the existing signal tests against that union. Keep the log-event fields required.

jhgaylor (changes requested):
Adversarial correctness and test review: **REQUEST_CHANGES**. The new SSE contract rejects two payload shapes these endpoints already send: synthetic server-exit log events and collection-change signals. Details and minimal remedies are inline.

Reviewed head `e7af00c8c4c356a19ad111ebedfba0cf229700d2`, using PR merge-base `4daf6bb987704b0191f423adae5877d31bb300f5` and policy from recorded base `135b433b25b9df64857f52879eb5b03f331872ef`.

Verification: all 3 focused Swift-generator tests for stream properties passed. A pinned-Elixir diagnostic compiled the head's schema and the actual server-exit renderer extracted from its AST: shutdown and crash frames both fail OpenApiSpex casting at `stream`; each signal payload fails at required `kind` and `ts`. Reviewed existing crash/shutdown and signal endpoint tests and SDK dispatch callers. The full server suite was not run.

<details><summary>Source coverage: all 16 PR paths</summary>

- `apps/fountain/lib/fountain_web/controllers/events_controller.ex`
- `apps/fountain/lib/fountain_web/controllers/team_controller.ex`
- `apps/fountain/test/fountain_web/controllers/conversation_controller_test.exs`
- `apps/fountain/test/fountain_web/controllers/events_stream_test.exs`
- `apps/fountain/test/fountain_web/controllers/team_stream_test.exs`
- `apps/fountain/test/fountain_web/schema_enum_guardrail_test.exs`
- `apps/fountain/test/support/schema_guard.ex`
- `changelog.d/2297-stream-log-event-schema.md`
- `contributing/swift-wire-models.md`
- `scripts/ci/test_swift_generation.py`
- `scripts/sdk-contract/generate-swift.py`
- `sdk/swift/CHANGELOG.md`

</details>

jhgaylor (changes requested):
Independent adversarial API/SDK compatibility review: **REQUEST_CHANGES** at `e7af00c8c4c356a19ad111ebedfba0cf229700d2`.

Reviewed all 16 PR paths against merge-base `4daf6bb9`, using the approved base's API/SDK guidance. The Swift field-source change preserves the generated public surface. I independently confirmed two P2 contract defects already described in the correctness review:

- `StreamLogEvent.stream` rejects the `null` emitted by `ConversationController.write_server_down_event/2` on shutdown/crash. Make the SSE field nullable and validate synthetic terminal events.
- The multiplexed stream response declarations require log fields `kind`/`ts`, excluding their normal `conversations`, `team`, and `schedule` signals with payload `{"reason":"changed"}`. Describe both log and signal variants and validate those frames.

Verification: Swift generation `--check` and four focused generator compatibility tests passed; Python contract verification and the 28-scenario conformance lint passed. A TypeScript compiler probe against this head's generated types reproduced both failures: TS2322 for `stream: null` and TS2739 for signal payloads missing `kind`/`ts`. This is independent source review, not the installed Review Loop service's full verification.

Nonblocking: new schema/changelog prose names `GET /api/events`; the route is `GET /api/events/stream`.

jhgaylor (changes requested):
Adversarial review: tenant isolation and credential boundaries. Reviewed all 16 PR paths at `e7af00c8c4c356a19ad111ebedfba0cf229700d2` against merge base `4daf6bb987704b0191f423adae5877d31bb300f5`.

**REQUEST_CHANGES — P2: permit the actual server-down frame in StreamLogEvent.** [schemas.ex:2044](https://github.com/managoat/fountain/blob/e7af00c8c4c356a19ad111ebedfba0cf229700d2/apps/fountain/lib/fountain_web/schemas.ex#L2044) inherits the REST schema's non-nullable `stream` string. When a monitored conversation server shuts down or crashes, [write_server_down_event/2](https://github.com/managoat/fountain/blob/e7af00c8c4c356a19ad111ebedfba0cf229700d2/apps/fountain/lib/fountain_web/controllers/conversation_controller.ex#L1215) sends `event: stage` with `stream: null`. A client validating against the newly declared schema therefore rejects the lifecycle event that explains the disconnect. Override `stream` as nullable in the stream schema and validate the existing crash/shutdown test frames, then regenerate the contract/types. The producer predates this PR; the incompatible declared frame contract is new.

Traced the three streams through `TenantAPIAuth` and their tenant-scoped conversation/event/team lookups. Found no introduced tenant-isolation, secret-exposure or credential-authority regression.

Verification: ran the pinned-Elixir probe using the actual head schema and AST-extracted production writer; both `:shutdown` and `:killed` frames fail `OpenApiSpex.Cast` with `:null_value` at `[:stream]`. Conformance lint passed all 28 scenarios; PR whitespace check passed. Source review included second-tenant denial assertions; the database-backed server tests and full Review Loop verification were not run in this review. No duplicate inline thread added for the corroborated finding.

apps/fountain/lib/fountain_web/schemas.ex
BinaryBourbon: Fixed in b9b77636, from the emitter side rather than the schema: `write_server_down_event/2` now sends `stream: ""`, which is what every persisted event already sends (the domain schema defaults `stream` to `""` and validates only `""` plus the three named streams), so the synthetic frame stops being the one frame with a different shape. The existing crash test in `sse_stream_test.exs` now decodes that frame and validates it against `StreamLogEvent`; putting `nil` back fails it at `#/stream`. The `null` to `""` change for that frame is called out in the changelog fragment.

apps/fountain/lib/fountain_web/controllers/events_controller.ex
BinaryBourbon: Fixed in 1d89d7486. A new `StreamSignal` schema (`reason`, required, enum `changed`) describes the `conversations`, `team` and `schedule` frames, and `GET /api/events/stream` and `GET /api/team/stream` now declare `oneOf: [StreamLogEvent, StreamSignal]` on their `text/event-stream` response. `GET /api/conversations/:id/stream` stays `StreamLogEvent` alone since it sends no signal frames. The three existing signal-frame tests validate their decoded frame against `StreamSignal`, and the `conversations` test also asserts the frame does not validate as `StreamLogEvent`, so the union is load-bearing. The route name is corrected to `/api/events/stream` everywhere the PR wrote it. Contract regeneration, `build.sh --check`, the Swift generator tests and all four client contract verifiers pass on the oneOf without client changes.

jhgaylor (approved):
Independent AI correctness/testing review: **APPROVE** at `1d89d7486530b04affdd2ac94ea086bea19ed58c`.

Reviewed all 17 changed files, including the original implementation and both follow-up fixes, and traced the three SSE renderers, schema validation helper, and Swift generation callers. No blocking correctness defect found.

Both previous schema mismatches are resolved:

- The [synthetic server-exit writer](https://github.com/managoat/fountain/blob/1d89d7486530b04affdd2ac94ea086bea19ed58c/apps/fountain/lib/fountain_web/controllers/conversation_controller.ex#L1203) now emits `stream: ""`; its real crash-path regression validates the rendered frame against `StreamLogEvent`.
- The [events](https://github.com/managoat/fountain/blob/1d89d7486530b04affdd2ac94ea086bea19ed58c/apps/fountain/lib/fountain_web/controllers/events_controller.ex#L64) and [team](https://github.com/managoat/fountain/blob/1d89d7486530b04affdd2ac94ea086bea19ed58c/apps/fountain/lib/fountain_web/controllers/team_controller.ex#L301) response declarations include `StreamSignal` alongside `StreamLogEvent`. All three signal names still emit the documented `{"reason":"changed"}` body, without changing cursor behavior.

Nonblocking coverage improvement: the [new signal assertions](https://github.com/managoat/fountain/blob/1d89d7486530b04affdd2ac94ea086bea19ed58c/apps/fountain/test/fountain_web/controllers/events_stream_test.exs#L185) validate the named component directly. Also validate against the operation's declared `text/event-stream` response schema so removing the union member would fail a regression test. The current union declarations are correct.

My validation, in an isolated snapshot with a separate local database: pinned Elixir 1.19.2 / OTP 28.3 compile with warnings as errors; all four changed controller test files plus the schema enum guardrail, **149 tests, 0 failures** (seed 2318); Swift generation **30 tests passed**; Swift generation staleness check; conformance fixture lint (**28 scenarios**); diff whitespace check. This is a standalone review with targeted diagnostics, not a claim to have run the full Review Loop verification pipeline.

jhgaylor (approved):
Independent AI review — API/SDK compatibility, complete at `1d89d7486530b04affdd2ac94ea086bea19ed58c`. Reviewed all 17 PR files, including the original schema/generator changes and subsequent fixes. No blocking API or SDK defect found.

The previous SSE findings are resolved: the synthetic server-exit frame now emits `stream: ""`; the events/team responses declare `StreamLogEvent | StreamSignal`, covering their `conversations`, `team`, and `schedule` notifications. I traced the ordinary and synthetic renderers, optional blocks/routing fields, and current SDK/CLI consumers. Cursor behavior is unchanged, and existing stage handling accepts the normalized stream value.

Swift now sources both routing IDs from `StreamLogEvent` while preserving their published names and optionality. Regeneration is byte-identical, and the release-based disappearance/requiredness guards remain effective.

Nonblocking documentation correction: [Events.swift lines 3–7](https://github.com/managoat/fountain/blob/1d89d7486530b04affdd2ac94ea086bea19ed58c/sdk/swift/Sources/FountainKit/Models/Events.swift#L3-L7) still says `EXTRA_PROPERTIES` supplies the IDs and SSE operations have a bare-string schema. Update this comment to describe the new `StreamLogEvent` source; it now contradicts the generator and updated contributor guide.

Local checks passed in an isolated snapshot: 30 Swift generation tests; Swift generation `--check`; TypeScript typecheck and 42 SSE/conformance tests; Swift contract/client/conformance selection (36 tests, retaining the suite's existing matrix skip); TypeScript/Python contract verifiers; conformance lint (28 scenarios). The initial TypeScript fixture run was blocked by sandbox loopback permissions and passed when rerun with that access; typecheck passed after supplying the snapshot's dependency path. I did not rerun the Elixir/server suite locally. This is an independent source review, not a full Review Loop certification.

jhgaylor (approved):
Independent AI security and tenant-isolation review of `1d89d7486530b04affdd2ac94ea086bea19ed58c`: **APPROVE**. No new security or tenant-isolation defect found across all 17 changed paths.

I independently traced all three SSE entrypoints from bearer-key authentication to resource lookup, replay, subscriptions and JSON emission. The routes retain `TenantAPIAuth`; the conversation stream establishes ownership before replay/subscription, the events feed joins log events to the authenticated user's conversations, and the team stream derives its conversation/agent mapping from the tenant-scoped roster. The new schemas, schema-test helper and Swift generator branch add no credential reads, secret persistence, audit mutations, callback destinations or request-controlled execution. Evidence: [auth and stream routes](https://github.com/managoat/fountain/blob/1d89d7486530b04affdd2ac94ea086bea19ed58c/apps/fountain/lib/fountain_web/router.ex#L22), [conversation ownership](https://github.com/managoat/fountain/blob/1d89d7486530b04affdd2ac94ea086bea19ed58c/apps/fountain/lib/fountain_web/controllers/conversation_controller.ex#L1045), [tenant feed query](https://github.com/managoat/fountain/blob/1d89d7486530b04affdd2ac94ea086bea19ed58c/apps/fountain/lib/fountain/conversations.ex#L2480), [team roster](https://github.com/managoat/fountain/blob/1d89d7486530b04affdd2ac94ea086bea19ed58c/apps/fountain/lib/fountain/team.ex#L130).

The earlier frame-contract defects are resolved at this head. The [actual synthetic server-exit renderer](https://github.com/managoat/fountain/blob/1d89d7486530b04affdd2ac94ea086bea19ed58c/apps/fountain/lib/fountain_web/controllers/conversation_controller.ex#L1203) emits `stream: ""`, matching `StreamLogEvent`, and still emits no SSE `id:`. The [events declaration](https://github.com/managoat/fountain/blob/1d89d7486530b04affdd2ac94ea086bea19ed58c/apps/fountain/lib/fountain_web/controllers/events_controller.ex#L64) and [team declaration](https://github.com/managoat/fountain/blob/1d89d7486530b04affdd2ac94ea086bea19ed58c/apps/fountain/lib/fountain_web/controllers/team_controller.ex#L300) now include `StreamSignal` in `oneOf` for the existing change notifications.

Independent diagnostic: compiled the pinned renderer body and schema/operation declarations in an isolated Elixir 1.19.2 / OTP 28.3 process using the existing OpenApiSpex validator. All **7 cases passed**: normal, shutdown, handoff and two failure exits validate without advancing the cursor; mutating their stream back to null fails; both multi-conversation unions accept `{reason: "changed"}` and reject an invalid reason. Source review also checked the second-tenant conversation/events denial assertions, optional blocks, generated TypeScript/contract alignment, Swift field sourcing and disappearance guards, all changed tests, and both changelog/documentation updates.

Limits: this reviewer did not independently rerun the HTTP/database suite or provision provider resources. This is a source security review with the focused diagnostic above, not a full Review Loop policy certification.

PR #2321, opened 2026-09-15
test(audit): narrow the window assertion's query, not its result (#2178)

BinaryBourbon:
## What

The since/until window test in `AuditTest` now narrows its **query** with a unique `action_prefix`, instead of narrowing the **result** after `_unsafe_list_events/1` has already cut its 200-row page.

Refs #2178. This is the first of the two halves that issue separates out: the test. The leak half, `unboxed_run` tests committing audit rows that nothing removes, is #2322. The two touch different files and are independent; together they close the issue.

## Why

`_unsafe_list_events/1` is cross-tenant and pages 200 rows newest first. The planted `old.event` is the oldest row in the window, so it is the first to fall off the end of the page once enough other rows share it. On a reused workstation database the rows committed by `unboxed_run` tests accumulate run over run, and from roughly the second full-suite run the `until` assertion returns `[]`. CI never sees it because every run builds a fresh database.

## How it was verified

- Created a fresh test database and seeded it with 250 audit rows with a null `user_id` and `inserted_at` inside the window, the shape the leak leaves behind.
- On main, `mix test apps/fountain/test/fountain/audit_test.exs:211` fails there at the `until` assertion, exactly as reported.
- With this change the whole file passes on that database (23 tests, 0 failures). Credo is clean on the file.

No changelog fragment: a test-only change with nothing a user or operator can observe.


jhgaylor (approved):
Adversarial review — product and API compatibility

Verdict: APPROVE.
Reviewed head: `ae574ee33e4b8a4794760cb1bcfa2b51280d5de9` (base `40131bff8f03661e45b66f8d57897be3ca8e6ff5`).

I independently reviewed the full diff in `apps/fountain/test/fountain/audit_test.exs`, traced `Audit._unsafe_list_events/1` through the shared query builder and the API/browser callers, and checked the approved-base contributor/API guidance and ADR 0013.

The prefix restricts the SQL query before the 200-row page is selected. Both planted events share it, so ignoring either time filter still fails; the exact-boundary assertions also fail if inclusive comparisons become exclusive. `Fountain.DataCase` rolls back these fixtures, so the new prefix does not join the committed backlog on subsequent test runs. No controller, wire contract, SDK, CLI, tenant boundary, or audit-retention behavior changes.

This addresses the test-isolation half of #2178 as stated. The separate unboxed-test cleanup remains open; it is not silently claimed fixed. I found no caller-visible regression or missing requirement that warrants changes.

Evidence/limits: independent source review and live-head verification; this reviewer did not run the Elixir suite. This is a delegated manual review, not a claim that the automated Review Loop verification recipe ran.

jhgaylor (approved):
Adversarial review — tenant isolation and security

Verdict: APPROVE for commit `ae574ee33e4b8a4794760cb1bcfa2b51280d5de9`.

Reviewed the complete change in `apps/fountain/test/fountain/audit_test.exs` and traced the audit query builder, production callers, event insertion, and SQL sandbox lifecycle. No blocking defect found.

- The unique action prefix is applied as an Ecto-bound SQL predicate before the 200-row limit. The existing helper escapes backslash, percent, and underscore; the generated prefix itself contains no LIKE wildcard characters.
- Cross-tenant lookup is intentional for this admin query. Production browser access selects it only for the authenticated admin role; ordinary browser and API access retain `user_id` scoping. Existing second-tenant denial assertions and system-event coverage are unchanged.
- Both planted events remain ordinary sandboxed inserts with rollback on test cleanup. The change introduces no unboxed writes or credentials. Exact assertions still detect reversed, missing, or exclusive time bounds without depending on unrelated committed audit rows.

Validation: independent source review only; I did not run the Elixir tests in this reviewer process. This is a manual delegated review, not an automated Review Loop service verdict.

jhgaylor (approved):
Adversarial review — correctness and tests

Verdict: APPROVE. Reviewed the complete diff from `40131bff8f03661e45b66f8d57897be3ca8e6ff5` to `ae574ee33e4b8a4794760cb1bcfa2b51280d5de9`.

Coverage: `apps/fountain/test/fountain/audit_test.exs`, with the audit query builder, event changeset, user factory, SQL Sandbox lifecycle, and issue #2178 traced for context.

No blocking findings. The action prefix becomes a SQL WHERE predicate before the 200-row limit, so unrelated committed events cannot displace either planted row. The four exact-result assertions still exercise both sides of the time window and equality at each inclusive boundary. The prefix has a delimiter and no LIKE metacharacters; the only fixtures that generate it remain inside DataCase's rolled-back sandbox. The separate unboxed-fixture cleanup problem remains tracked in #2178.

Validation: source inspection and `git diff --check` passed. I inspected the coordinating reviewer's diagnostic logs: on an isolated database seeded with 250 unrelated null-user events inside the window, the base test failed at the reported `until` assertion; restoring this exact head made the whole file pass (23 tests, 0 failures). Diagnostics used pinned Elixir 1.19.2 / OTP 28.3 and local PostgreSQL 15.14, rather than CI's PostgreSQL 16. This does not claim completion of Review Loop's independent verification service.

jhgaylor (approved):

Verdict: APPROVE for commit ae574ee33e4b8a4794760cb1bcfa2b51280d5de9.

Reviewed the entire PR: apps/fountain/test/fountain/audit_test.exs, against the approved-base contributor rules and audit ADR, and traced Audit.recent_query/2, prefix/time filters, record!/1, the factory and SQL Sandbox lifecycle. No blocking findings.

The existing action_prefix filter selects both planted events before the 200-row page is applied. This directly fixes the test half of #2178 without production changes or new shared helpers. The two events retain the same prefix, so the four exact-result assertions still detect missing bounds and exclusive boundary comparisons. The generated actions are sandboxed and rolled back; they do not add to the committed audit-row leak, which the PR explicitly leaves tracked in #2178.

Validation: inspected the coordinating reviewer's actual diagnostic logs: with 250 unrelated null-user audit rows inside the window, the base test failed at the reported until assertion; the exact-head audit file passed all 23 tests. This independent source review does not claim a local full-suite or full Review Loop verification run.

Issue #2324, opened 2026-09-15
Swift SDK: APIErrorBody needs one named error schema server-side before it can be generated


BinaryBourbon:
Split out of #2300, which named `APIErrorBody` (`sdk/swift/Sources/FountainKit/Errors/FountainError.swift`) as the last contract-shaped Swift type still declared by hand. `TeamResource`'s bodies landed in #2315 and `PageMeta` in #2323. This one is not an SDK change as the contract stands, which is why it gets its own issue rather than a third PR.

## Why it cannot be generated today

#2300 leaned on #2169 having given every 422 one shape. That is true of the bytes on the wire (every 422 carries `error`, validation ones also carry `errors`) and not true of the contract, which is what the generator reads. The committed contract declares **seven named error schemas plus one inline body**:

| Schema | Shape | Declared on |
|---|---|---|
| `Error` | `{error}` | 400/401/403/404/409/410/413/415/422/429/502/503/504, ~700 declarations |
| `AuthError` | `{error, message?}` | 403 ×4, 422 ×5 (the auth endpoints) |
| `ChangesetError` | `{error?, errors}` | 422 ×15 |
| `UnprocessableEntityError` | `{error, errors?, message?}` | 422 ×3 |
| `CredentialSetDeletionError` | `{error: enum, message, reason: enum}` | 422 ×1 |
| `BrokerUnavailableError` | `{error, message, reason}` | 502 ×1 |
| `NegotiationError` | `{errors: {detail}}` | 406 ×165 |
| inline | `{error?, upgrade_url?}` | 402 on `POST /api/conversations` |

422 alone references five of them. There is no one schema to point the generator at, and the guard that refuses divergent reused inline shapes would refuse this on purpose.

The handwritten struct is a permissive union across all of them **with behaviour folded in**, none of which a generated `Decodable` expresses:

- `code` reads `error`, except on the auth endpoints where `error` is prose and `reason` is the machine code, so `reason` outranks `error` when present.
- `fieldErrors` accepts `errors` as either `[String: [String]]` or `[String: String]` and normalises the latter.
- `httpStatus` is stamped by the client from the response, because a code that outranks its status (`conversation_busy` on a 400) would otherwise lose it.
[...]

PR #2323, opened 2026-09-15
refactor(swift): generate PageMeta from the contract's cursor envelope

BinaryBourbon:
Closes #2300. Second of the three types it names; `TeamResource`'s bodies landed in #2315 and `APIErrorBody` is split out to #2324, which needs a server-side contract change first.

## What changes

- **`PageMeta` is generated.** The cursor envelope (`has_more`, `limit`, `next_cursor`) appears in the contract as an inline `meta` object on `AuditEventListResponse` and `LogEventListResponse`, never as a named schema. Both declarations are named `PageMeta` in `INLINE_TYPES`, and the generator's reused-inline-shape guard is what proves they are the same shape: a third cursor endpoint with a divergent `meta` fails generation instead of silently taking the first one's.
- **`/api/search` gets `SearchResponse.Meta`.** It pages by offset (`has_more`, `limit`, `offset`), a different shape. Wholly new type, so it takes contract requiredness and its members are non-Optional.
- **The three list responses are generated roots** (`AuditEventListResponse`, `LogEventListResponse`, `SearchResponse`) and `audit.list`, `conversations.events` and `search.search` decode them instead of the untyped `Envelope`, which loses its `meta`.
- **`Page<T>` becomes `Page<Items, Meta>`.** `page.meta?.hasMore` / `?.nextCursor` read exactly as before; only code that spelled `Page<[X]>` has to name the meta. `Page` never had a public initializer, so consumers only ever read one.

## The compatibility rules, demanded not volunteered

The handwritten `PageMeta` lived in `Client/APIClient.swift`, outside `Models/`, so the released-Swift baseline could not see it: neither the source rule (published Optional stays Optional) nor the presence rule (a published property does not vanish) would have fired on this move. The baseline now reads the whole `Sources/FountainKit` tree at the tag. With that, the guard raises for exactly the three records this PR carries and renders with them:

| Rule | Without the record | Record |
|---|---|---|
| Source | `PageMeta.hasMore shipped Optional and the contract now requires it` | `OPTIONAL_COMPAT` pin |
| Source | `PageMeta.limit shipped Optional and the contract now requires it` | `OPTIONAL_COMPAT` pin |
| Presence | `PageMeta.offset shipped in v0.19.0 and is no longer generated` | `REMOVED_PROPERTIES` entry (ceiling 1 → 2); `offset` was only ever sent by search and decoded nil everywhere else |

[...]

PR #2169, opened 2026-09-14
ci: diff the changelog guard against the merge commit's first parent

jhgaylor:
## What

The changelog guard (#2158) runs `git diff base.sha...HEAD` in a depth-2 checkout of the PR's merge ref. When main moves between the event that started the run and the checkout, the merge commit's first parent is the new tip while `base.sha` is the old one; fetching the old one alone at depth 1 gives a commit with no history in common with HEAD, so git says `no merge base` and exits 128. The step then dies with a Python traceback and `workflow-checks` goes red.

#2163 and #2166 both hit this within ten minutes of #2164 landing. A re-run cannot recover, because the merge ref is rebuilt against whatever main is by then.

On a merge commit, the step now diffs against `HEAD^1`, which is the base the merge was actually built against and is always present. The fetch fallback stays for the non-merge case.

## Verified

Reproduced locally with a depth-2 clone of a merge whose base had moved on: the three-dot diff against the event base fails with `fatal: <sha>...HEAD: no merge base`, and the first-parent diff lists exactly the PR's files. The `scripts/ci` policy suite still passes (57 tests). No change to `scripts/changelog.py`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01HkopMNnchbyzYhYEbT6UbV

Issue #1444, opened 2026-09-03
Two operations declare ChangesetError for 422 but can also refuse with a code


jhgaylor:
The residue of #1431, and the part that is an API decision rather than a fix.

## What is left

After #1431 every 422 body in the API carries `error`, and a validation failure
also carries `errors`:

```json
{"error": "validation_failed", "errors": {"name": ["can't be blank"]}}
{"error": "environment_not_allowed", "message": "…"}
```

Two operations declare `ChangesetError` for 422, which *requires* `errors`, and
can return either body:

| Operation | The other 422 |
|---|---|
| `POST /api/conversations` | `sandbox_identity_mismatch`, `environment_not_allowed`, `vault_not_allowed` and the rest of the fallback controller's coded refusals |
| `POST /api/auth/register` | `{"error": "email and password are required"}` |

So their declaration is accurate for one of their two 422s and wrong for the
other. Both are allowlisted in `FountainWeb.SchemaGuardAllowlist` as
`:mixed_422_shapes`.

## The decision

**What should a 422 that can be either a validation failure or a coded refusal
declare?** Three answers, all defensible:

1. **Declare `Error` on those two.** Accurate — every 422 has `error` now — and
   a two-line change. Loses the documentation that field errors can appear,
   which is the useful half for the validation case.
2. **Widen `ChangesetError`**: `error` required, `errors` and `message`
   optional. Then one schema describes every 422 in the API honestly and the
   other five operations that declare it get more accurate too. This is a wire
   contract change: `errors` stops being required, so `sdk/contract`'s
   projection moves, the TypeScript manifest's `ChangesetError` entry needs
   updating, and a `sdk/conformance` fixture gains `error`. All additive for
   clients — no body loses a key — but it is four SDKs' worth of bookkeeping,
   and the name then covers coded refusals too, which is a misnomer.
3. **A new schema** with the shape from 2 and a name that fits, leaving
   `ChangesetError` for the operations that really only ever return a
   changeset.

My preference is 3 if the 422 shape is worth documenting properly, and 1 if it
is not worth the churn. I did not want to make that call inside a PR scoped to
the renderer.

## Why it is not urgent

Nothing is broken. Both bodies carry `error`, every SDK's `ValidationError`
maps on status, and `fieldErrors` reads `errors` defensively and gets `{}` for a
coded refusal, which is correct — there are no field errors.

Refs: #1431, #1427.

PR #2325, opened 2026-09-15
feat(api)!: one Error schema on every JSON error status (#2324)

BinaryBourbon:
Step 1 of #2324: the server side. Every JSON error status now declares the one `Error` schema, and the five specialised error schemas plus the inline 402 body on `POST /api/conversations` are gone from the document. Nothing on the wire changed; the contract now describes what the server already sends.

## What moved

| Before | After |
|---|---|
| `Error` `{error}` | `Error` `{error, message?, reason?, errors?, upgrade_url?, active_sandboxes?, limit?}` |
| `AuthError` `{error, message?}` on 403 ×4, 422 ×5 | `Error` |
| `ChangesetError` `{error?, errors}` on 422 ×15 | `Error` |
| `UnprocessableEntityError` `{error, errors?, message?}` on 422 ×3 | `Error` |
| `CredentialSetDeletionError` `{error: enum, message, reason: enum}` on 422 ×1 | `Error` |
| `BrokerUnavailableError` `{error, message, reason}` on 502 ×1 | `Error` |
| inline `{error?, upgrade_url?}` on the 402 of `POST /api/conversations` | `Error` |
| `NegotiationError` `{errors: {detail}}` on 406 | unchanged: it is the one genuinely different shape, rendered before any controller runs |

The superset is the list #2324 names: the fields the four SDKs read. `FallbackController` also sends `status`, `field`, `credentials_url` and the rate limiter `retry_after_seconds` beside particular codes; those stay undeclared extras, as they were, since no client reads them and OpenAPI objects admit them.

Two things the collapse gives up, on purpose: the contract no longer says `message` and `reason` are required on the broker 502 and the credential-set 422, and the two enums on `CredentialSetDeletionError` become prose in `reason`'s description. Both bodies are unchanged and the schema guard still validates them on every run.

## Why one schema and not seven

The seven described one wire shape (#2169 made every 422 carry `error`, and validation ones also `errors`) as seven contract shapes, so a generated client had no single type to decode an error into. `FallbackController` renders `active_sandboxes`/`limit` on the 429 quota refusal while every 429 was declared `{error}` alone. #1444 settled the mixed-422 question for two operations by adding a fourth 422 schema; this settles it for every status by keeping one.

## Bookkeeping

- `sdk/contract/contract.json` and the generated TypeScript types regenerated. The diff is exactly the seven schema entries; `generate-swift.py --check` is clean because no generated Swift model reaches an error schema.
- The four manifests replace their `ChangesetError` claim with the `Error` fields each client reads (`errors`, `message`, `upgrade_url`, `active_sandboxes`, `limit`, and `reason` for Swift), all `optional`, which is how every client already treats them.
- `ApiPipelineResponsesTest` asserted that the 422 on `POST /api/conversations` was `UnprocessableEntityError`; it now asserts what composition still has to preserve, the controller's own status description over the pipeline's.
- `SchemaEnumGuardrailTest` drops the two `CredentialSetDeletionError` enum entries.
- `docs/api.md` "Error responses" now says the body is one shape; `contributing/swift-wire-models.md` records `APIErrorBody` as generatable over `Error`, which is step 2.

No SDK's shipped code changes. `release:skip-sdk` because the generated TypeScript types under `src/` moved; `sdk/typescript/src/index.ts` does not export them.

## Validation

- `mix test` from the root against a private test database: core + ee 5653 tests, 0 failures; buzz 136, slack 30, support 33, google 60, microsoft 29, all 0 failures. Every rendered error body in those runs was validated against `Error` by the schema guard.
- `npm --prefix sdk/typescript run verify-contract`, `sdk/python/scripts/verify_contract.py`, `sdk/elixir mix test test/contract_test.exs`, `swift test --filter ContractTests`: all pass.
- `python3 sdk/conformance/lint.py` (28 scenarios, every fixture body validated against the declared schema), `scripts/sdk-contract/build.sh --check`, `generate-swift.py --check` and `test_swift_generation.py` (33 tests): all pass.
- `python3 scripts/changelog.py check`: ok.

Refs #2324 (step 2, generating `APIErrorBody`, stays open there). Related: #2300, #2169, #1444, #2323.


jhgaylor (approved):
Adversarial review of the full base e9d59ea120e33c2d084391bac8e71dd45963a304...f149139ba2251a1b8450e9856bea9d609dcbf9d3 change: approved, with no blocking regression found.

Traced the consolidated schema through actual fallback, validation, authentication/scope and specialized refusal renderers, pipeline composition, all four SDK error decoders and the CLI. Runtime response construction is unchanged; generated declarations and manifest dependencies match the consolidation. The removed public schema names and relaxed requiredness are explicitly documented as an upgrade change.

Verification run on an isolated copy of this revision: TypeScript and Python contract verifiers, conformance lint (28 scenarios), Swift generation --check, changelog check and git diff --check passed. The Swift check used the existing local Git release metadata after the archive alone lacked .git. I inspected the passing required CI result; I did not rerun the Elixir integration suite locally.

Nonblocking documentation qualification: docs/api.md's “Every JSON error status” wording also needs to allow the readiness probe: GET /health/ready deliberately still declares and returns ReadinessResponse on 503. Qualifying that sentence to API resource errors would make the scope precise without changing the probe. The retained 406 exception is correctly described.

Issue #2021, opened 2026-09-12
Tracker: remaining sandbox-binding lifecycle gaps (#1767, ADR 0058)


jhgaylor:
## Current scope — reassessed 2026-09-17

This is an umbrella checklist, not one implementation PR. The historical report below describes the pre-ADR-0058 campaign; several items have since landed or changed shape.

Assessment: static inspection of `main` at `ae736fe0eff2aed7ac07675041b09185002c5425`, issue/PR history, and open #2386 at `81252353d89dea927f342aaf1e66c02b3aeac43e`. No new race reproduction or tests were run during this assessment.

| Original item | Current disposition |
| --- | --- |
| 1. Unbound `mark_orphan` | Fixed by #2327; actor recovery now goes through the binding-aware `Machine.end_turn/3` door. |
| 2. Actor release | Follow-up #2392. The running-successor case is guarded, but `Binding.detach(..., policy: :keep)` discards the supplied sandbox ID before the parent status write. A stale release against an **idle** replacement still appears possible; reproduce before calling this fixed. |
| 3. Remaining conversation status writes | Audit/fix #2393 is implemented in #2410, awaiting review and merge. The deterministic audit reproduced eight stale-binding status overwrites; the PR guards delayed attach/provision/watchdog reports and removes the redundant autonomous write. Its per-site inventory also records the safe parent-lock invariants. |
| 4. Co-tenant rebinding | Substantially changed by merged #2377: the owner ends affected turns during destroy and sends the machine-gone notices; Wake moves co-tenants afterward. Reassess failure/order cases against that implementation rather than applying the original proposed fix. |
| 5. Abandoned running turns | Partly addressed by #2377: destroy ends bound running turns and a ceiling park ends the requester's turn. Not yet a blanket closure: turn ending is best effort, and `AutonomousTurnReaper` still skips conversations with a live server. Reassess what can remain abandoned and whether recovery is durable. |
| 6. Lenient sandbox-ID read | Already landed: `TurnMachine.from_state/1` reads `state.sandbox_id`. |
| 7. Fenced sandbox rows skipped by reclamation | Fixed by #2329. #2386 evolves the teardown sweep into an owner-driven recovery path with durable destruction intent; retain recovery guarantees through stage 9. |

## Execution and sequencing

- [ ] #2392 — focused release regression and binding fix; can start on current main.
[...]

PR #2327, opened 2026-09-15
fix(conversations): bind the reattach orphan write to the actor's sandbox

BinaryBourbon:
Item 1 of #2021.

## The gap

`Reattachment.mark_orphan/3` is the reattach path's give-up. Every way of
failing to pick the turn back up ends there — `list_sessions` exhausting its
backoff, no session matching the conversation's tag, an adapter that never
sent `session/prompt`, an `attach` that is refused — and each one flips the
conversation's running turn to `interrupted` and the conversation to `idle`.

It was handed `state` and discarded it:

```elixir
```

Its turn comes from `find_running_turn/1`, which is scoped by conversation and
never by sandbox, so the write went through unbound — in a function that runs
inside a live actor, reached only from the reattach path, which is precisely
where a stale actor turns up.

The failure: actor A on sandbox S1 restarts and reattaches; the conversation
has since been rebound to S2 with actor B running a turn; A's `list_sessions`
retry expires under a partition; A interrupts **B's live turn** and idles a
conversation that is still working. That is the loss #1767 exists to close, in
one of the functions that campaign edited.

## The fix

Pass `expected_sandbox_id: state.sandbox_id` — the actor's own binding, set
once from init args (`conversation_server.ex:242`) and never reassigned.
`_unsafe_orphan_turn/3` takes `FOR UPDATE` on the parent and compares under
that lock, so a rebind answers `{:error, :ownership_changed}` and writes
nothing. The stale actor then carries on with `current_turn: nil`, which is
correct — a turn it no longer owns is not its to finish.

The refusal is logged. A silent `:noop` on this path is what made the gap hard
to see in the first place.

`AutonomousTurnReaper`'s unbound call at `autonomous_turn_reaper.ex:69` is
unaffected and stays unbound: it fires only when `ConversationServer.whereis/1`
is nil, so no actor binding exists to compare against.

## Tests

`reattach_orphan_binding_test.exs` pins both directions through the real
`reattach_running_turn/1`, with the session listing stubbed to fail:

- an actor still holding the binding reconciles its own turn (`interrupted`,
  `orphaned_at` set, conversation `idle`);
- an actor whose conversation was rebound leaves the successor's turn
  `running` and the conversation `running`.

Verified by reverting: with the fence removed the second test fails with
`left: "interrupted"`, i.e. the successor's live turn is the one that gets
interrupted.

`mix test apps/fountain/test/fountain/conversations apps/fountain/test/fountain/workers`
— 1680 tests, 0 failures. `mix precommit toolchain conflict-markers compile
deps format credo` — PASSED.

## Not in scope

Items 2, 3, 4, 5 and 7 of #2021 stay open; item 7 is PR-ed separately. Item 6
already landed — `turn_machine.ex` reads `state.sandbox_id` strictly now.


jhgaylor (approved):
Adversarial review of 44de07b8ab11e329b0e414bbdf5a5230075fdb07: approved, with no blocking findings.

Reviewed all three changed files and traced every orphan exit through ConversationServer initialization/reattachment, Conversations._unsafe_orphan_turn/3 and ExecutionGuard._unsafe_recover_turn/3. The actor's immutable sandbox binding reaches the existing comparison under the parent row lock, before any turn or execution mutation. Reassignment therefore refuses the stale write; current-owner recovery and the unbound system reaper retain their behavior. The new positive/negative regression assertions complement the existing real PostgreSQL lock-order tests. No public API or tenant-access surface changes.

Verification: git diff --check and pinned-toolchain Elixir parsing of both changed Elixir files passed. Rechecked current-head CI: all six server test partitions, static analysis, coverage and CI required passed (https://github.com/managoat/fountain/actions/runs/35027689432). I did not rerun the database suite locally; no prepared local test database was available.

PR #2328, opened 2026-09-15
refactor(agents): enforce an explicit environment access policy

BinaryBourbon:
The legacy nullable environment allowlist is now represented by an explicit
persisted policy mode and used by both conversation authorization and team
environment selection. This is the `allowed_environment_ids` half of #2107,
built on the template #2119 set for vaults.

- `nil` remains `all_tenant_environments`, including environments created
  after the migration in the same tenant.
- `[]` and finite arrays remain `allowlist`, preserving deny-all and finite
  access. Naming the agent's own environment is still not an override, so it
  passes under every policy.
- Existing API/export shapes remain `allowed_environment_ids`;
  `environment_access` is internal and generated from the existing values, so
  no SDK or contract change is needed.
- Historical snapshots distinguish omitted keys (`unchanged`), explicit JSON
  null (`all_tenant_environments`), arrays (`allowlist`) and malformed values
  (`invalid`). Invalid and unsaved/inconsistent in-memory policies fail closed.
- A forward migration adds generated stored columns to `agents` and
  `agent_versions`. The original arrays and snapshot JSON remain authoritative
  for mixed-version writers.
- `Team.addable_options/2` drops its own copy of the three-way rule and asks
  the policy, the way it already does for vaults.

## Validation

- Isolated `fountain_2107_env_test` database created and migrated from the
  repo root, so the extension migrations ran too.
- 12 new focused tests plus 144 affected conversation/team/reapply tests pass,
  covering unrestricted, deny-all and finite policies, the agent's own
  environment, old writers using `Repo.update_all`, environments created after
  the migration, cross-tenant denial, snapshots, API/apply round trips and
  malformed policies.
- Both guards verified by reverting: flipping the allowlist clause open fails
  12 tests, and inverting the generated SQL expression fails 27.
- `mix precommit` PASSED (9 stages, 257s). `git diff --check` clean.

## Follow-up

`allowed_inference_credential_ids` is the last one, in a separate PR because
it changes the same two tables and must not mint a colliding migration
version.

Refs #2107 and #2094.


BinaryBourbon:
Stack for #2107 (`stack:2107`):

1. **#2328** — `allowed_environment_ids` → `environment_access` (this PR, base `main`)
2. #2330 — `allowed_inference_credential_ids` → `inference_credential_access` (based on this branch)

Merge order is 1 then 2; #2330 gets retargeted to `main` after this lands. The stack exists because both PRs touch the same three modules and because two migrations minted in parallel would collide on version.

jhgaylor (approved):
Adversarial review: approved at `586305efafac3452ec6d69bdfe766b15f9d995ac` against base `e9d59ea120e33c2d084391bac8e71dd45963a304`.

Reviewed all 11 changed paths and traced environment policy through tenant-scoped launch/attach/channel/reapply callers, team selection, agent updates/deletion cleanup, historical restores, and API/export serializers. I found no concrete blocking regression. The generated mode preserves null/empty/finite behavior, keeps the own-environment exception, and cannot replace the subsequent tenant ownership check. The new field is excluded from client writes and wire responses; historical omitted/null values retain their distinct restore behavior. Migration lock bounds and rollout prerequisites are documented.

Verification: at this exact head, 121 focused tests passed (environment access, conversation start, team, and environment API/apply round trips) using the pinned Elixir/OTP toolchain and an isolated PostgreSQL 15 database. The complete stacked migration sequence also succeeded; `git diff --check` passed, and current GitHub CI is successful. This was focused local verification, not a full local precommit or a production-scale lock test.

PR #2330, opened 2026-09-15
refactor(agents): enforce an explicit credential set access policy

BinaryBourbon:
Stacked on #2328 — review that one first. Based on `2107-environment-access`
because both PRs touch `agent.ex`, `agent_version.ex` and `conversations.ex`,
and because two migrations minted in parallel would collide on version.
Retarget to `main` after #2328 lands.

The last of the three legacy nullable allowlists in #2107. Same template as
the vault half (#2119) and the environment half (#2328).

- `nil` remains `all_tenant_credential_sets`, including credential sets
  created after the migration in the same tenant.
- `[]` and finite arrays remain `allowlist`, preserving deny-all and finite
  access. Naming the set the agent already runs on is still not an override,
  so it passes under every policy.
- Existing API/export shapes remain `allowed_inference_credential_ids`;
  `inference_credential_access` is internal and generated from the existing
  values, so no SDK or contract change is needed.
- Historical snapshots distinguish omitted keys (`unchanged`), explicit JSON
  null (`all_tenant_credential_sets`), arrays (`allowlist`) and malformed
  values (`invalid`). Invalid and unsaved/inconsistent in-memory policies fail
  closed.
- A forward migration adds generated stored columns to `agents` and
  `agent_versions`. The original arrays and snapshot JSON remain authoritative
  for mixed-version writers.

With this, no allowlist on an agent relies on a null to mean "anything the
tenant owns", which closes the open half of #2107.

## Validation

- Isolated `fountain_2107_env_test` database, migrated from the repo root.
- 12 new focused tests plus 289 affected agent/conversation/launch/controller
  tests pass, covering unrestricted, deny-all and finite policies, the agent's
  own set, old writers using `Repo.update_all`, sets created after the
  migration, cross-tenant denial, snapshots, API/apply round trips and
  malformed policies.
- Both guards verified by reverting: flipping the allowlist clause open fails
  8 tests, and inverting the generated SQL expression fails 20.
- `mix precommit` PASSED (9 stages). `git diff --check` clean.

Refs #2107 and #2094.


jhgaylor (approved):
Adversarial review: approved at `d39a84b670ebf07ec911b2be5aa20ce98f60e480`, reviewing the incremental diff against its #2328 base `586305efafac3452ec6d69bdfe766b15f9d995ac` and integration with that policy change.

Reviewed all 9 changed paths and traced credential-set authorization through tenant-scoped launch/attach/channel callers, persisted agent updates, historical version restores, and API serializers. I found no concrete blocking regression. Null/empty/finite allowlists and the explicit own-set exception preserve existing behavior; permissive policy still requires a tenant-scoped credential lookup. Generated columns support old writers without accepting a client-supplied mode, and snapshots retain omitted/null/array semantics while malformed values fail validation on restore.

Verification: 157 focused tests passed at this exact stacked head, covering both policies, conversation start, credential-set launch, team, and both API/apply round-trip suites with pinned Elixir/OTP and an isolated PostgreSQL 15 database. Fresh migrations and this migration's down operation both succeeded; the lower PR's 121 focused tests then passed after rollback. `git diff --check` passed, migration versions do not collide, and current GitHub CI is successful. This was focused local verification, not a full local precommit or a production-scale lock test.

PR #2333, opened 2026-09-15
fix(runners): answer a rejected runner registration instead of raising

### LINKED ISSUE #2332: Runner registration validation failure returns an opaque 500 instead of the changeset error
## What happens

Starting a runner with a `--name` that fails validation produces an opaque `HTTP 500` and an infinite reconnect loop. The daemon never says what is wrong.

```
$ fountain runner --name "Aarons-Mac-mini" --root ~/work/fountain-sandboxes
fountain runner "Aarons-Mac-mini" (process): sandboxes in /Users/…, connecting to http://localhost:4000
level=WARN msg="runner: disconnected; reconnecting" in=1s err="connect: HTTP 500: {\"errors\":{\"detail\":\"Internal Server Error\"}}"
level=WARN msg="runner: disconnected; reconnecting" in=2s err="connect: HTTP 500: …"
level=WARN msg="runner: disconnected; reconnecting" in=4s err="connect: HTTP 500: …"
```

Server-side:

```
[info] GET /api/runners/ws
[info] Sent 500 in 4ms
[error] ** (RuntimeError) cannot render template :error because conn.params["_format"] is not set.
    Please set `plug :accepts, ~w(html json ...)` in your pipeline.
    (fountain 0.19.0) lib/fountain_web/controllers/runner_controller.ex:1: FountainWeb.RunnerController.action/2
```

The real problem is only that the name has uppercase letters and fails `@name_format`. The user is never told.

## Why

`RunnerController.register_and_upgrade/3` renders the changeset through a view:

```elixir
{:error, changeset} ->
  conn
  |> put_status(:bad_request)
  |> put_view(FountainWeb.ChangesetJSON)
  |> render(:error, changeset: changeset)
```

But `/api/runners/ws` is routed without `:accepts_json`, on purpose — the router says so:

```elixir
# The daemon's socket skips content negotiation like the SSE routes do: a
# WebSocket client sends no JSON `Accept`, and the upgrade is not a JSON
# response.
scope "/api/runners", FountainWeb do
  pipe_through [:api, :require_full_scope]
  get "/ws", RunnerController, :connect
end
```

With no `plug :accepts`, `conn.private[:phoenix_format]` is never set, so `render/3` raises. The two sibling branches in the same `cond` use `json/2` and are unaffected — only the changeset branch reaches for a view.

`FountainWeb.FallbackController` gets this right with `put_view(json: FountainWeb.ChangesetJSON)`, but that path depends on a format being negotiated too, so the format-independent `json/2` is the safer shape on this route.

## Impact

Any validation failure on runner registration — a bad `--name`, an over-long `--root` — becomes an unreadable, infinitely-retrying 500 rather than the `{"error": "validation_failed", "errors": {…}}` body every SDK's `ValidationError.fieldErrors` expects.

Not reachable with a bare `fountain runner`: `DefaultRunnerName` already lowercases, strips the domain, sanitises and truncates. It needs an explicit `--name`.

## Repro

```sh
fountain runner --name "Has-Uppercase"
```

## Environment

Self-hosted `v0.19.0` (Docker Compose, Postgres 16), CLI `v0.19.0` on macOS arm64.

Happy to send a PR — one line in the controller plus a regression test.

alecraso:
Closes #2332.

## The bug

`GET /api/runners/ws` renders a failed changeset through a view:

```elixir
{:error, changeset} ->
  conn
  |> put_status(:bad_request)
  |> put_view(FountainWeb.ChangesetJSON)
  |> render(:error, changeset: changeset)
```

That route pipes through `:api` **without** `:accepts_json` — deliberately, and the router says why:

```elixir
# The daemon's socket skips content negotiation like the SSE routes do: a
# WebSocket client sends no JSON `Accept`, and the upgrade is not a JSON
# response.
```

With no `plug :accepts`, no format is ever negotiated, so `render/3` raises. The daemon sees:

```
connect: HTTP 500: {"errors":{"detail":"Internal Server Error"}}
```

and reconnects forever with nothing in it to act on. The two sibling branches in the same `cond` already use `json/2` and were unaffected — only the changeset branch reached for a view.

## The fix

Answer with `json/2`, rendering the body `ChangesetJSON` produces everywhere else, so the daemon gets the `validation_failed` detail every SDK's `ValidationError.fieldErrors` already reads. `FallbackController` uses `put_view(json: ...)`, but that still depends on a negotiated format, so the format-independent shape is the right one on this route.

## How I hit it

Self-hosting on a Mac. I passed `--name "$(hostname -s)"`, which here is `Aarons-Mac-mini`, and the uppercase fails `@name_format`. The instance told me nothing for several minutes of reconnect loop. Bare `fountain runner` is unaffected — `DefaultRunnerName` lowercases, strips the domain, sanitises and truncates — so this needs an explicit `--name`, or an over-long `--root`.

## Tests

New sibling module rather than a case in `RunnerControllerTest`, because it flips `:runners_enabled`, which is application-wide and off in test config — the same reason `agents_live_test.exs` is on the guardrail allowlist. `async: false` keeps it off that list.

Verified both directions:

```
# with the fix
7 tests, 0 failures
  (runner_registration_test + runner_controller_test + async_global_config_guardrail_test)

# with the controller change reverted, the new test alone
1 test, 1 failure
  ** (RuntimeError) cannot render template :error because conn.params["_format"] is not set.
```

Run against `hexpm/elixir:1.19.2-erlang-28.3-debian-trixie-20260610-slim` (the Dockerfile's own build base) with Postgres 16. `python3 scripts/changelog.py check` and `scripts/ci/dco.py` both pass. I did not run `mix precommit` in full — no local Elixir toolchain, so credo/dialyzer/format are unverified and CI should be the judge there.


BinaryBourbon (approved):
Independent adversarial correctness review by Codex — **APPROVE**.

Reviewed PR #2333 at `20ee09b2e5c6c744808431206c8464f5038269a5` against `e9d59ea120e33c2d084391bac8e71dd45963a304`. No concrete correctness or regression defect found.

Coverage: `apps/fountain/lib/fountain_web/controllers/runner_controller.ex`, `apps/fountain/test/fountain_web/controllers/runner_registration_test.exs`, and `changelog.d/2332-runner-registration-error.md`; traced the router/authentication pipeline, runner registration/changeset, shared error encoder, SQL Sandbox/test configuration, and Go daemon's connection handling.

Evidence:

- This WebSocket route has no negotiated format. Phoenix's atom-template `render/3` requires one; `json/2` sends the existing `ChangesetJSON.error/1` result directly, retaining HTTP 400 and the field-error body.
- The daemon already treats HTTP 400 as a permanent rejection and includes its body. Returning the intended status therefore both exposes the actionable name error and exits the reconnect loop.
- The regression uses the actual authenticated route with no JSON Accept header, asserts HTTP 400 and a name-specific validation error, and checks that registration wrote no runner row. Its `async: false` module isolates the global enabledness setting from async siblings and restores the previous value on exit; ConnCase retains SQL Sandbox cleanup.
- The changelog matches the reported issue and inspected CLI name normalization. `python3 scripts/changelog.py check` passed.
- I inspected the coordinating reviewer's six passing rendering probes: exact base/head error branches with real Runner changesets and Phoenix rendering, testing uppercase names and overlong roots with absent, wildcard, and JSON Accept headers. The base raises for missing format; the head consistently returns JSON HTTP 400 with the expected field details.

Validation limits: source review is complete. The rendering probe does not exercise database-backed controller integration; the local database was unavailable, so the full application suite and `mix precommit` were not rerun locally. The coordinating reviewer verified [CI succeeded on this exact head](https://github.com/managoat/fountain/actions/runs/35037387821), including the required CI check; secret scanning also passed.

BinaryBourbon (approved):
Independent adversarial security review by Codex — **APPROVE**.

Reviewed all three changed files and the final head `20ee09b2e5c6c744808431206c8464f5038269a5` against base `e9d59ea120e33c2d084391bac8e71dd45963a304`. No actionable security or tenant-isolation regression found.

- Traced `/api/runners/ws` through bearer authentication, full-scope authorization, the enabled/upgrade checks, registration, and the successful socket handoff. Registration still limits accepted attributes, assigns ownership from the authenticated user, and looks up/upserts `(user_id, name)`. Returning JSON on failure adds no registration or upgrade capability.
- `ChangesetJSON.error/1` traverses validation errors only; it does not serialize changeset data, request headers, credentials, or runner telemetry. The runner's current validations return fixed field messages and bounds. The new response preserves HTTP 400 and bypasses the absent format negotiation without adding an HTML rendering path.
- The new regression test exercises the invalid-name branch without an Accept header, checks the structured error, and proves no runner row is created. Its `async: false` isolation and `on_exit` restoration protect the application-wide runner flag. Inspected existing full-scope denial and cross-tenant runner tests as well.
- Checked the changelog against the CLI: HTTP 400 becomes a permanent, explanatory registration failure; hostname-derived defaults are normalized to the accepted name grammar.

Validation: the coordinating reviewer ran six focused rendering probes using the exact base/head error-branch AST and real Runner changesets, ChangesetJSON, and Phoenix. For invalid names and overlong roots with absent, wildcard, and JSON Accept headers, base raised for the missing format; head returned HTTP 400 with only the expected error/errors fields. Current-head CI run 35037387821 and secret scanning passed.

Source review is complete. Diagnostic limitation: the probe does not exercise database-backed controller integration; the full application suite and a live daemon handshake were not rerun locally. This is an independent Codex review, not a Review Loop App verdict.

BinaryBourbon (approved):
**APPROVE — independent adversarial API compatibility review by Codex.**

Reviewed all three changed files in PR #2333 from `e9d59ea120e33c2d084391bac8e71dd45963a304` to `20ee09b2e5c6c744808431206c8464f5038269a5`: `runner_controller.ex`, `runner_registration_test.exs`, and `2332-runner-registration-error.md`. No concrete compatibility regression found.

- Traced registration through the authenticated, full-scope WebSocket route and changeset rendering. Phoenix `json/2` sends the existing HTTP 400 status and `application/json` directly, avoiding the missing-format failure in `render/3`. The response retains `error: validation_failed` and field-specific `errors`; it satisfies the existing OpenAPI/committed `Error` contract. Successful upgrade, authentication, disabled-runner, and duplicate-connection branches are unchanged.
- Traced the actual Go daemon: it reads the failed handshake body, treats HTTP 400 as a permanent rejection, and returns the detail instead of reconnecting. This matches the documented runner lifecycle. Checked the changelog's name claim against `DefaultRunnerName`: lowercase ASCII sanitization, trimming, fallback, and the 63-character limit produce a valid default.
- Inspected error consumers across TypeScript, Python, Elixir, both Swift products, Go/Buzz CLI, and the Hermes plugin. The runner socket is already deliberately excluded from general SDK wrappers. Raw SDK callers retain field errors on generic HTTP 400 errors; this does not require changing them to the HTTP 422 `ValidationError` classification. No endpoint, generated model, manifest, or omission change is needed.

Independent diagnostics passed against this head's actual Python and TypeScript error decoders: HTTP 400 preserves `validation_failed` and `errors.name`, and is non-retryable. Inspected the new no-Accept-header regression test, persisted-row assertion, global-config restoration, and existing schema guardrails. The coordinating reviewer additionally ran six focused probes using the exact base/head error branches, real runner changesets, and Phoenix rendering: invalid name/overlong root with absent, wildcard, or JSON Accept headers consistently changed from a missing-format exception to JSON HTTP 400 with field errors. These probes did not execute database-backed controller integration; full suites were not rerun locally. [CI passed on the reviewed head](https://github.com/managoat/fountain/actions/runs/35037387821).

This is an independent Codex review, not a Review Loop App verdict.

PR #2336, opened 2026-09-16
feat(swift): generate APIErrorPayload; keep the error code when reason only narrows it (#2324)

### LINKED ISSUE #2324: Swift SDK: APIErrorBody needs one named error schema server-side before it can be generated

BinaryBourbon:
Closes #2324. This is step 2. Step 1 (#2325) gave every JSON error status one `Error` schema.

## What changes

- **`APIErrorPayload` is generated** from `Error`. It is named so that it does not shadow `Swift.Error`. `APIErrorBody` no longer declares its own wire keys. It decodes the generated payload and keeps the behaviour on top of it: `code`, the normalised `fieldErrors`, the client-stamped `httpStatus` and `FountainError.from(status:body:)`. This is the same split `AdminUserPage` has over `AdminUserListResponse`. Every published member keeps its name and optionality. `APIErrorBody` gains `reason`, which holds the body's `reason` exactly as sent.
- **`reason` no longer always wins** (second commit, the decision #2324 asked for). `reason` now replaces `code` only when `error` is not code-shaped, meaning it contains anything other than lowercase ASCII letters, digits and underscores.
  - On the key-auth and scope refusals, `error` is a sentence and `reason` is the code. `code` is still `api_key_invalid` and so on.
  - On `credential_set_is_default`, `sandbox_not_resettable` and `broker_unavailable`, `error` is already the code and `reason` narrows it. The contract's own `Error.reason` description says so. Swift used to report `code == "is_default"` or `"timeout"` for these. It now reports the `error`, which matches TypeScript.

## Two records the guard does not demand

The released `Error` schema (v0.19.0) required `error` and declared no other fields, so the compatibility guard asks for neither record below. Both keep decoding every body the handwritten decoder accepted:

| Record | Why | Proof |
|---|---|---|
| `OPTIONAL_COMPAT` pin on `Error.error` | Phoenix's `ErrorJSON` and the 406 `NegotiationError` send `errors` without `error`, and the released `ChangesetError` left `error` optional | Dropping the pin (and patching the Swift so it compiles) fails 3 tests with `keyNotFound("error")`: `validationErrorsMapIsDecoded`, `aBodyWithoutErrorStillDecodes` and `PublicSurfaceTests` |
| `TYPE_OVERRIDES` `Error.errors` → `JSONValue` | The same bodies send `{"detail": "..."}`, and servers before #1431 sent an array. A typed map would fail the whole body, `code` included | Dropping it fails the `PublicSurfaceTests` compile. `aBodyWithoutErrorStillDecodes` decodes the `detail` shape |

Breaking the new rule back to "`reason` always wins" fails `aCodedErrorKeepsItsCodeWhenReasonNarrowsIt` with `"is_default" == "credential_set_is_default"`. Every probe ran on a copy, and the originals were restored and compared with `cmp`.

One trade-off: the old decoder wrapped each key in its own `try?`, so a scalar of the wrong type (say `limit: "3"`) dropped only that member. The generated decoder rejects the whole body instead. The server's schema guard checks these types on every rendered body, and the two open-shaped keys above are handled explicitly.

## Checks

- `python3 scripts/sdk-contract/generate-swift.py --check`: in step
- `python3 -m unittest discover -s scripts/ci -p 'test_*.py'`: 136 tests OK
- `swift test -Xswiftc -warnings-as-errors`: 127 tests passed
- `swift format lint --strict --recursive Sources Tests`: clean
- `python3 scripts/changelog.py check`, `scripts/conflict-markers.py`, `git diff --check`: clean

Docs: the `APIErrorBody` row in `contributing/swift-wire-models.md`, plus that file's `TYPE_OVERRIDES` count, which read 19 and was already 21 before this change (22 now). Also `changelog.d/2324-swift-api-error-payload.md` and the Unreleased section of `sdk/swift/CHANGELOG.md`.


jhgaylor (approved):
Adversarial reviewer: security, tenant isolation, and error handling (Codex).

**Security verdict: APPROVED** for `e25b4d5dd4402cde2ac3e29e762d020416a48340`. I found no actionable security or tenant-isolation defect in the eight-file PR diff. This is a security-perspective review; the other adversarial reviewers are separately reporting a source-compatibility blocker, so this approval is not a claim of overall merge readiness.

Reviewed all changed paths: the generator and generated payload; `FountainError.swift`; both changed Swift test files; the contributor guide; and both changelogs. Traced the decoded payload through regular HTTP failures, SSE failures/reconnects, error classification, retryability, and the server's auth/scope and coded-error producers. Authentication and scope refusals preserve their machine codes; coded errors preserve the top-level code and expose the narrowing reason. HTTP failures remain failures, and stream reconnection still uses the original status. No changed tenant lookup, credential forwarding, secret persistence, or authorization side effect was introduced.

Validation: an isolated Swift diagnostic using the actual error/value sources and verbatim generated payload passed nine checks covering revoked keys, insufficient scope, broker/sandbox/default-set refusals, error-less 406 bodies, legacy arrays, nullable fields, and mixed validation-message shapes. Existing HTTP/public-surface regression assertions were also inspected. This review does not claim a server integration or second-tenant test run; the change does not modify server authorization paths.

sdk/swift/Sources/FountainKit/Errors/FountainError.swift
```diff
@@ -21,6 +26,7 @@ public struct APIErrorBody: Sendable, Equatable {
   public init(
     code: String? = nil,
     message: String? = nil,
+    reason: String? = nil,
```
jhgaylor: **P2: Preserve the published initializer's function signature.** Adding the defaulted `reason` parameter removes `APIErrorBody.init(code:message:fieldErrors:upgradeURL:activeSandboxes:limit:httpStatus:)`, which ships in v0.19.0. An existing consumer storing or passing that initializer as a factory now fails to compile:

```swift
let makeError = APIErrorBody.init(code:message:fieldErrors:upgradeURL:activeSandboxes:limit:httpStatus:)
```

A `swiftc -typecheck` probe against the merge-base declaration passes; the same probe at this head fails with “type 'APIErrorBody' has no member 'init(code:message:fieldErrors:upgradeURL:activeSandboxes:limit:httpStatus:)'”. Default arguments preserve ordinary calls, but cannot preserve this initializer reference. This violates the existing public API compatibility requirement in `contributing/swift-wire-models.md`.

Keep the original initializer forwarding `reason: nil`, add an overload that requires an explicit `reason` argument, and add an `import FountainKit` public-surface regression for the old factory reference. Requiring `reason` on the new overload also keeps `APIErrorBody()` unambiguous.

jhgaylor (changes requested):
Manual adversarial API compatibility review: CHANGES_REQUESTED.

Reviewed all 8 changed paths, the generated Error payload against current/released server responses, and affected SDK/CLI consumers. The intentional error/reason precedence correction matches the contract. One public initializer compatibility regression remains, detailed inline.

The base/head Swift compiler probe reproduces the break. Coordinated verification passed the full Swift suite, strict formatting, generator freshness, 136 Python tests, and repository hygiene checks.

jhgaylor (changes requested):
Independent adversarial correctness/test review: **CHANGES_REQUESTED** for `e25b4d5dd4402cde2ac3e29e762d020416a48340`.

I independently reproduced the [published initializer compatibility defect](https://github.com/managoat/fountain/pull/2336#discussion_r4025232459): the existing `APIErrorBody.init(code:message:fieldErrors:upgradeURL:activeSandboxes:limit:httpStatus:)` reference typechecks against the exact base declaration (exit 0) and fails against this head (exit 1, missing initializer). In an isolated copy, retaining that initializer and delegating to an overload with required `reason` restores the reference while keeping both `APIErrorBody()` and `APIErrorBody(reason: ...)` unambiguous; that remedy typechecks.

Fourteen isolated Swift runtime probes passed against the actual head decoder, generated payload and error mapping, covering prose/code precedence, narrowing reasons, absent/null fields, legacy arrays, mixed validation values, quota metadata, and classification/status/retryability. All eight changed files and relevant callers/server producers were reviewed; I found no additional blocker. The coordinated full Swift suite, strict format, generator check, 136 Python tests and repository guards also passed.

sdk/swift/Sources/FountainKit/Errors/FountainError.swift
BinaryBourbon: Fixed in 8926d4830. The v0.19.0 initializer is back with its exact signature and passes `reason: nil` through. A new overload takes `reason` and requires it, so `APIErrorBody()` still has only one match. `PublicSurfaceTests` (not `@testable`) now stores `APIErrorBody.init(code:message:fieldErrors:upgradeURL:activeSandboxes:limit:httpStatus:)` and calls it. I checked the test by deleting the restored initializer: the test file then fails to compile with "has no member". With the initializer back, `swift test` and `swift format lint --strict` pass. The guide row and both changelogs mention the overload.

jhgaylor (approved):
Adversarial API/source compatibility review of the whole PR at `8926d48303b728f8711f4635111ec9e3eb7b2a85`: **APPROVE**.

The previous P2 [initializer-reference regression](https://github.com/managoat/fountain/pull/2336#discussion_r4025232459) is fixed. The original seven-parameter public initializer retains its exact signature and forwards with `reason: nil`; the new overload requires `reason`, so `APIErrorBody()` remains unambiguous. `PublicSurfaceTests` imports `FountainKit` without `@testable` and stores and invokes the original initializer reference.

I traced the generated payload through actual error producers and the typed client's HTTP/error mapping. Existing public members retain their types; omitted `error`, Phoenix/406 detail strings, and legacy `errors` shapes remain supported. Code-shaped `error` values retain precedence while prose auth/scope refusals keep their reason-based codes. No remaining actionable source or wire compatibility defect found.

Coverage: all eight changed files — `changelog.d/2324-swift-api-error-payload.md`, `contributing/swift-wire-models.md`, `scripts/sdk-contract/generate-swift.py`, `sdk/swift/CHANGELOG.md`, `sdk/swift/Sources/FountainKit/Errors/FountainError.swift`, `sdk/swift/Sources/FountainKit/Models/ConversationWire.generated.swift`, `sdk/swift/Tests/FountainKitTests/ClientTests.swift`, and `sdk/swift/Tests/FountainKitTests/PublicSurfaceTests.swift`. Also checked the server Error schema/producers, sibling clients, CLI integration boundaries, and existing error conformance expectations. Current-head local verification coordinated with the other reviewers passed: 127 Swift tests with warnings as errors, 136 Python checks, generator freshness, strict Swift formatting, and repository guards. This is an independent manual review; required CI remains authoritative.

jhgaylor (approved):
Adversarial correctness and tests review: **APPROVE**.

Reviewed the complete PR at `8926d48303b728f8711f4635111ec9e3eb7b2a85` against recorded base `d78f0693682012b3bac0507bcec4665691f07fdc` using the triple-dot diff and trusted base review guidance. No correctness blocker remains.

The earlier initializer-reference defect is resolved: the original public initializer signature is preserved, the new overload requires `reason`, and `PublicSurfaceTests` compiles the original factory reference through a public import. I traced the generated payload through `APIClient.error`, body normalization and `FountainError` status/retry mapping against actual server producers. Code-shaped errors retain their code while prose errors retain the existing reason fallback. Omitted `error`, string/array field errors and client-stamped status retain their supported behavior. Regression assertions cover the changed behavior, and the existing generator propagation and pin checks include the new `Error` root.

Validation at the reviewed head: the coordinating reviewer ran the full Swift suite with warnings as errors (127 tests), all 136 Python CI tests, generator staleness, strict Swift formatting and changelog/conflict/diff checks; all passed.

Coverage:
- `changelog.d/2324-swift-api-error-payload.md`
- `contributing/swift-wire-models.md`
- `scripts/sdk-contract/generate-swift.py`
- `sdk/swift/CHANGELOG.md`
- `sdk/swift/Sources/FountainKit/Errors/FountainError.swift`
- `sdk/swift/Sources/FountainKit/Models/ConversationWire.generated.swift`
- `sdk/swift/Tests/FountainKitTests/ClientTests.swift`
- `sdk/swift/Tests/FountainKitTests/PublicSurfaceTests.swift`

jhgaylor (approved):
Adversarial security and error-boundary review — APPROVED at 8926d48303b728f8711f4635111ec9e3eb7b2a85.

Independently reviewed all eight changed files and traced actual server auth/scope refusals through HTTP/SSE decoding, status/retry mapping, generator overrides, and public regression assertions. No actionable defect found.

Authentication/scope reason codes remain intact; SSE still terminates on 4xx; coded broker/sandbox/credential refusals retain their parent code and expose the narrowing reason. The published initializer signature is restored and covered by a normal-import consumer test.

Verification for this head, run by the review coordinator: 127 Swift tests and 136 Python tests passed; generation, strict Swift formatting, changelog, conflict-marker and diff checks passed.

PR #2342, opened 2026-09-16
docs(decisions): ADR 0058, the machine has one owner

### LINKED ISSUE #2307: Design (ADR): the sandbox reaper's idle park must serialize against wake, rehydration and admission

jhgaylor:
## What

A new ADR, `decisions/0058-the-machine-has-one-owner.md`, Proposed and nothing built. It proposes the per-sandbox owner that ADR 0023 step 4 specified and its Outcome replaced with "a lock plus a query": one `Fountain.Machines.Machine` process per sandbox, backed by a durable lease on the `sandboxes` row, as the only writer of `sandboxes.status` and the only caller of the provider's create, resume, suspend, destroy and checkpoint. Conversation servers, the reaper, Launch, Wake, Reapply, Termination, admin and account deletion ask it.

0023's Outcome gets one sentence pointing at the revisit it asked for. The index is regenerated.

## Why now

On `main` at `c3f568596`, `update_sandbox` is called from 24 sites in 11 files and the provider's mutations from 19 sites in 10 files. Every protection around those writes — `expected_sandbox_id` (#1767, with #2021's leftovers), `reset_requested_at`, `teardown_requested_at`, `configuration_revision`, the retirement match copied four times (#2039), four "is anyone here" predicates (#2255) — is compensation for the owner that was not built. #2286 spent five adversarial rounds trying to make one of those writes safe with a lock and was withdrawn; #2307 asked for an ADR before more code. This is that ADR: the seven #2307 constraints are its protocol, written once instead of per writer.

The reason to do it rather than keep fencing: the product direction (#805, #1089 two agents on one machine, #1910) needs a place for "materialize this agent's layer onto this disk" to live, and that is a machine operation.

## The brief

The ADR ends with an implementation brief: nine stages in the #2175 form (ADR → ratchet test pinned at 24 + 19 → lease module → read-only process → destroy → park → provision/resume → binding and admission → flip and delete), which are delegable moves and which need the adversarial pass, the rules the last month of reviews already paid for, and the two measurements to take before stage 1. #1089 and #1910 are explicitly after it.

## Checks

`scripts/decisions-index.sh`, `okf validate decisions` (valid; the six warnings are pre-existing actor-convention notes on other ADRs), `scripts/conflict-markers.py`, `scripts/changelog.py check`, `git diff --check`. No changelog fragment: nothing a user or operator can observe changes.

Closes #2307. Answers to #2255's three open decisions and a note on #1089 follow once this is in.


BinaryBourbon (approved):
**Verdict: Verified — Non-blocking (corrected numbers/names below).** No false claim that changes the argument, no unacknowledged contradiction with an Accepted ADR, no unbuilt behaviour described as built.

Re-measured everything at `c3f568596` (the commit the ADR cites) using a detached worktree, and checked every issue/PR/ADR citation against the tracker. Frontmatter (`status: draft` / `adr_status: "Proposed"`), the 0023 edit (pointer only, doesn't touch 0023's decision text), and CI (`okf validate` green on the PR head `e09120b`) are all correct as described.

**1. The counts table — `update_sandbox` is right, provider mutations is overstated by 2.**

`Conversations.update_sandbox/2` sites: confirmed 24 in 11 files, matching the table exactly, including per-file breakdown. (Note: the 24 includes the `def update_sandbox/2` line itself in `conversations.ex` as one of the counted "sites" — the def is followed by a self-call at `:174`, so the file still nets 2, matching the table's "3" for the `conversations.ex`+`lifecycle.ex` row. Worth a one-word footnote in the ADR since "called from" reads as call-sites-only.)

Provider mutation sites (`create/resume/suspend/destroy/create_checkpoint`): actual count is **17 sites in 9 files**, not 19 in 10. Two errors, both in the same direction:

- `workers/sandbox_reaper.ex`: table says 3, actual is **2** (`:286` suspend, `:369` destroy). The third regex hit is the file's own `@moduledoc`, which quotes `` `_ = Managoat.Sandbox.destroy(handle)` `` in backticks as prose describing `ConversationServer`'s behavior (line 8) — not a call site.
- The `home_checkpoint.ex, accounts/deletion.ex, execution_deadline_worker.ex` row: table says 3, actual is **2** (`home_checkpoint.ex:71` `create_checkpoint`, `deletion.ex:300` `destroy`). `execution_deadline_worker.ex` has no call to any of the five named verbs — its one `Managoat.Sandbox` call is `terminate_session/3` (`:226`), a different function. Either drop it from this row or name `terminate_session` as a sixth verb the owner also becomes the only caller of (worth deciding, since stage 8's "the server's last `update_sandbox` and `Managoat.Sandbox.destroy` sites go" doesn't mention it either).

Everything else in the table checks out exactly, including one site that's easy to undercount with a naive `Managoat\.Sandbox\.` grep: `provisioning.ex:124` calls `Sandbox.create_checkpoint(...)` unqualified (the file does `alias Managoat.Sandbox`), which is the third site making that row's "3" correct.

**2. The four predicates — two of the four names are already stale at `c3f568596`.**

Confirmed the sibling PR's report. `_unsafe_sandbox_held_by_other?/2` (`lifecycle.ex:604`) and `_unsafe_sandbox_busy_elsewhere?/4` (`conversations.ex:1706`) still exist as named. But `_unsafe_reap_sandbox/1` was already moved to `Termination.reap_sandbox/1` by #2257 (`termination.ex:283`, with a docstring saying so), and `SandboxReaper.server_alive?/1` was already deleted by #2309 and folded, along with the reap's own scan, into `Lifecycle.any_server_alive?/1` / `live_conversation_ids/1` (`lifecycle.ex:564-588`). So at the commit the ADR pins to, there are really three surviving names, not four, and one of them (`Lifecycle.any_server_alive?/1`) already covers what two of the ADR's four names used to be. Correct as of `c3f568596`: `_unsafe_sandbox_held_by_other?/2`, `_unsafe_sandbox_busy_elsewhere?/4`, `Termination.reap_sandbox/1`'s call into `Lifecycle.live_conversation_ids/1`, and `Lifecycle.any_server_alive?/1`.

**3. The quoted 0023 Outcome paragraph is not fully verbatim.**

The ADR's blockquote silently drops `` (`Conversations` around `pg_advisory_xact_lock`) `` after "taken by the conversation that needs it," with no ellipsis marker there — while correctly using `[...]` a sentence later to elide "Each `ConversationServer` keeps its own adapter...". Doesn't change the meaning, but the task rule for this quote is verbatim, so it should either keep the parenthetical or mark the elision.

**4. The #2175 quote is a paraphrase, not verbatim — but the decision and date are confirmed.**

Presented in quotes as `"the machine stays in the server (ADR 0023 stands)"`. The actual comment (jhgaylor, 2026-09-15, on #2175) reads: "**The machine stays in the conversation server.** ADR 0023 stands; no `SandboxServer`." — "conversation" is dropped and two sentences are merged. Substance is faithful. The date checks out: the same comment records all eleven #2175 stages merging "the same day," 2026-09-15, matching the ADR's "that day."

**5. One citation is a fair inference, not a literal match.** The Decision section's "(#2307 constraint 4, reproduced on #2286)" for "a process alone can have two instances across a rolling deploy" — constraint 4's literal text is about Horde registry propagation lag affecting wake-registration durability, not directly about duplicate Machine-process instances. The inference (async registry → possible duplicate ownership during a deploy) is reasonable and is spelled out explicitly in the Alternatives section, so not misleading, just worth knowing it's not a direct quote.

**Everything else checked out precisely**, including several citations that would have been easy to get wrong: #2307's seven constraints (verbatim numbered list, matches every citation including constraints 1, 6, 7); #2286 closed-not-merged, five adversarial rounds (confirmed in #2307's own body); #2309's SQLSTATE `57014` / `BEFORE INSERT` trigger claim (found verbatim in a #2309 PR comment); #2021's "nine `fix:` PRs," "25 `codex/p2-1767-*` branches," and "four writes still unfenced" (items 1-4 of its six numbered findings — 5 is the `:noop` debt cited separately, 6 is hardening, not a defect); #2039's "four places" (title match); 0013's actor convention; 0009's "Sprite destruction, best-effort" (verbatim support for "deletion's teardown stays non-fatal"); 0037's "Not a rewrite" (a literal heading); 0023's 2026-09-11 amendment on cotenants/`build_fingerprint` (verbatim support); the salvage branch `follow/2255-reaper-liveness-lock` (exists on the remote). No sentence reads as unbuilt-behavior-as-built — the top-of-document disclaimer is honest and consistent throughout, and the ADR doesn't reference the sibling open PRs (#2346, #2348) at all, so it makes no claim about their state.

No prose overstatement found beyond what's noted above.

Issue #2343, opened 2026-09-16
Unattended deployed verification: confirmed-rollout hook, digest attribution and a scheduled canary


jhgaylor:
Tracker: #1606. Split out of #1612 on 2026-09-16.

#1612 delivered on-demand deployed verification: a person points the suite at
an approved target and gets a verdict. This issue holds the unattended half,
which was deliberately dropped from that scope because the goal is a
push-button check, not continuous monitoring.

Nothing here blocks #1606. Open it only if unattended verification becomes
worth its operating cost.

## What was dropped

- An integration hook the deployment owner invokes after rollout completion,
  verifying the intended image digest across the serving deployment before
  claiming that revision passed. Avoid racing Flux by starting after the
  Kustomization's source revision and the Deployment rollout are both settled,
  not after manifest publication.
- Image-digest attribution through the external Kubernetes adapter, reported
  explicitly as evidence. Without it every run records
  `revision: { verified: false, reason: 'No deployment revision adapter configured' }`
  (`deployed/lib/runner.mjs:176`), which the docs already forbid relabelling as
  revision-verified.
- The scheduled basic/execution canary on an explicitly configured instance,
  with failures routed through the established operational mechanism and
  sustained failures tracked without automatic rollback.
- Measured stability and named failure ownership as the precondition for any
  blocking promotion or release gate.

## Already built

The code for all of it is merged and unused, not missing:

- `.github/workflows/deployed.yml` accepts `mode: rollout` with an
  `expected_digest`, and carries `schedule` crons at `23 */6 * * *` (canary)
  and `43 4 * * 6` (matrix), both gated behind repository variables that are
  deliberately unset.
- `deployed/adapters/kubernetes.mjs` checks observed generation, replica
  counts, Pod ownership, container readiness, EndpointSlice membership and the
  intended digest, before and after the public checks.
- `deployed/monitor.mjs` checks in to a Sentry Crons URL with
  `in_progress`/`ok`/`error`, correlated by workflow run and attempt.
- `docs/guides/operate/deploy.md` documents the rollout hook, the adapter's
  trust boundary, the monitor's configuration and the pre-gate stability bar.

## Known blocker if this is ever revived

`deployed/validate-environment.mjs:29` requires a `required_reviewers`
protection rule on the target environment, and `deployed-production` has one.
Because the `verify` job names that environment, every deployment to it waits
for a human approval — including scheduled runs. As written the canary cannot
run unattended.

[...]

Issue #1612, opened 2026-09-06
Deployed suite: an on-demand conformance run against an approved target


jhgaylor:
Tracker: #1606

The build and manifest workflows publish artifacts for Flux to reconcile.
Publication success does not prove that the intended revision is serving
requests or completing conversations. A deployed instance needs a repeatable
on-demand check that a person can point at it and get a verdict from.

**Scope narrowed on 2026-09-16.** The goal is a push-button conformance run
against a deployment, not continuous monitoring. The unattended half — the
confirmed-rollout hook, image-digest attribution, the scheduled canary and its
monitor routing — moved to #2343, which is parked and blocks nothing. The code
for it is merged and dormant behind unset variables.

## Acceptance criteria

- [x] An on-demand entry point selecting an approved target and profile, using
      credentials scoped to that target.
- [x] Serialize runs per target, impose time/concurrency/usage limits, and
      retain redacted results and cleanup manifests.
- [x] Document local Compose, remote self-hosted and CI commands plus fixture
      provisioning and failure triage in the operating guides.
- [x] Verification stays visible evidence. It gates no promotion or release,
      and public checks remain usable without cluster credentials.

## Evidence

Run [35091717086](https://github.com/managoat/fountain/actions/runs/35091717086),
`streaming` profile, public mode, suite revision `11e98f5bb`, against
`https://managoat.com` with `claude` / `anthropic/claude-haiku-4-5` / `sprites`.
All 10 checks passed: identity, capabilities, second tenant, fixtures,
provision, first turn and artifact, follow-up, replay and history, tenant
isolation, terminate. Two completed turns; history took 10 pages; 236 events
compared for replay, 228 for resumed replay, 230 live. Cleanup left 0
remaining resources and neither credential appears in the retained evidence.
An independent listing afterwards showed no conversations, environments,
vaults, bindings, webhooks or schedules on either suite account.

The preceding run, [35090216591](https://github.com/managoat/fountain/actions/runs/35090216591),
failed in the harness rather than the server: history paging stopped after 100
pages at any page size while replay requested 3 events per page, and the Haiku
conversation carried 570 events. Fixed in #2339.

`deployed/lib/runner.mjs` enforces the per-run time, resource and turn limits
and writes `result.json`, `junit.xml`, redacted traces and `cleanup.json` for
every initialized run; `deployed/cleanup-replay.mjs` retries an interrupted
manifest. `docs/guides/operate/deploy.md` documents the local Compose, remote
[...]

PR #2345, opened 2026-09-16
feat(deployed): verify a deployment with one command (#1612)

jhgaylor:
Closes the ergonomics half of #1612: a deployed Fountain can now be verified
with one command.

## Why

The suite could already answer "does this deployment work". Reaching that
answer meant copying `target.example.json`, editing `base_url`, profiles,
execution settings and limits by hand, exporting two keys and remembering the
CLI's flag shape — friction on the one path an operator wants when a
deployment looks wrong.

## What

```
$ scripts/verify-deployment.sh https://managoat.com basic

  keys       keychain: fountain-prod-deployed-suite-key
  target     https://managoat.com/
  profiles   basic
  out        /var/folders/.../fountain-verify-20260916T122453Z
  ...
  status     passed
  checks     11 passed, 0 failed
  cleanup    0 resource(s) remaining
  revision   unverified (No deployment revision adapter configured)
  evidence   /var/folders/.../results
```

- `scripts/verify-deployment.sh <url> [profile]` resolves `FOUNTAIN_SUITE_KEY`
  and `FOUNTAIN_SUITE_OTHER_KEY` from the environment, falling back to the
  macOS keychain, so a routine run leaves no credential in shell history. An
  exported key always wins, and a machine without `security` simply uses the
  environment.
- `deployed/verify.mjs` composes the target from flags, runs it and summarises:
  failing checks with their errors, skipped optional capabilities with their
  reasons, the remaining-resource count, whether the revision was verified, and
  where the evidence landed.
- `deployed/lib/target.mjs` now owns the profile list, limits and execution
  timings. `ci.mjs` composes through it too, so the two entry points cannot
  drift; a test pins that they agree across every profile.
- Plaintext HTTP is accepted only for a loopback target — a remote one would
  put an API key on the wire in the clear.
- Anything an environment owns rather than a caller — a deployment adapter, a
  provider matrix, a rollout digest — stays in `ci.mjs`.
- `deployed/README.md` documents provisioning the two accounts for the first
  time: the verification gate on minting a key, the live `quickstart` key
  onboarding leaves on a fresh account, and the `connections_enabled`
  requirement the integration profiles carry.

## Verification

- `node --test deployed/test/*.test.mjs` — 252 pass, up from 242.
- The new tests were checked by breaking what they guard: composing with the
  wrong profile, removing the loopback allowance and regressing the
  `schedules` run bound each turn a test red.
- `bash scripts/test-docs.sh` and `python3 scripts/changelog.py check` pass;
  `shellcheck` is clean.
- Run against production through the new entry point: `probe` passed 4 checks,
  `basic` passed 11, both with zero remaining resources.

## Scope

#1612 was narrowed today: the goal is a push-button conformance run, not
continuous monitoring. The unattended half — the confirmed-rollout hook,
digest attribution, the scheduled canary and its monitor — is parked in #2343
with its blocker recorded. This PR changes nothing about that dormant code.


BinaryBourbon (changes requested):
Independent adversarial correctness review by Codex — **REQUEST_CHANGES**.

Reviewed PR #2345 at `42fd45431a4647fc8c4db195fdee0d62e579e54b` against `11e98f5bb81055bc800e103ee44d759544dbaadf`.

1. **[P2] Only advertise profiles the new command can configure** — `deployed/verify.mjs:61`. `--profile` accepts `secrets`, `mcp`, `webhooks`, and `schedules`, but `verifyConfig` supplies none of their required receiver/window configuration and the parser offers no way to provide it. With both fake account keys, I passed every generated target through the real `configFrom`: these four always fail setup, while the other five validate. I also invoked the real `verifyMain` with `--profile schedules`; it returned 2 with `Expected explicit schedule windows` before any HTTP request. The new parity test only compares selected fields, so it passes over these unusable targets. Either supply the required profile configuration or limit this command's advertised choices and direct those profiles explicitly to `cli.mjs`. Add a regression that validates each advertised profile through the actual runner configuration loader.

2. **[P2] Correct the closed-registration account provisioning instructions** — `deployed/README.md:64`. The new guide says its calls work with public registration closed, then instructs the operator to create both suite accounts with `POST /api/auth/register`. With `REGISTRATION_ENABLED=false`, `Accounts.registration_allowed?/1` returns `registration_closed`; `register_user/2` preserves that refusal, and `RegistrationController.api_create/2` responds HTTP 403. Administering the deployment does not bypass this unauthenticated route's admission rule. The documented prerequisite therefore cannot be completed on the closed instances it explicitly promises to support. Distinguish reusing existing accounts while registration is closed from creating new accounts, and document a supported provisioning path or its required registration setting.

Whole-PR coverage: `deployed/verify.mjs`, `scripts/verify-deployment.sh`, `deployed/lib/target.mjs`, `deployed/ci.mjs`, `deployed/test/verify.test.mjs`, `deployed/README.md`, `docs/guides/operate/deploy.md`, and `changelog.d/1612-verify-a-deployment.md`. Traced runner setup, cancellation, cleanup, exit statuses, CI/matrix composition, and the referenced account API paths.

Validation: my fake-credential configuration and CLI reproductions above ran locally without contacting a deployment. The coordinating reviewer ran all 252 Node tests successfully in an isolated checkout of this head, plus `bash -n`, ShellCheck, and changelog validation. [Exact-head CI passed](https://github.com/managoat/fountain/actions/runs/35095757401). No live conformance run or real suite credentials were used; passing harness checks do not cover the two defects above.

BinaryBourbon (changes requested):

Reviewed all eight changed files and the final head `42fd45431a4647fc8c4db195fdee0d62e579e54b` against base `11e98f5bb81055bc800e103ee44d759544dbaadf`, including the shared CI composition, documentation, request/redaction paths, bounded execution, and fixture cleanup.

1. **[P1] Bind automatically loaded keychain credentials to the target origin.** `scripts/verify-deployment.sh:20,52–64` always falls back to `fountain-prod-deployed-suite-key`, regardless of the supplied URL. With no exported keys, the documented localhost invocation therefore retrieves the production suite credential and sends it in the first `GET /api/auth/me` to the local service; an arbitrary HTTPS destination gets the same credential. The later identity failure cannot undo disclosure of this full-scope production key. I reproduced the actual shell-to-HTTP path with a stub `security` executable and a fake production key: an unrelated loopback server received that key as its Authorization bearer. Resolve keychain entries through a configured normalized-origin binding, or refuse automatic fallback for an unbound destination and require explicit credential selection. Add a wrapper regression asserting a mismatched target receives no stored key, including when only one credential is exported.

2. **[P2] Reject credential-bearing URLs before logging or writing them.** `deployed/verify.mjs:36–51,101–104` checks only protocol/loopback policy, then writes and prints `url.href`. The existing `configFrom` check for URL userinfo/query/fragment runs afterwards. I reproduced `https://user:ftn_fake_userinfo_secret_security_probe@example.test` with a fake environment key: the run exited 2 without a remote request, but the URL password appeared verbatim in stdout and `target.json`. This bypasses the suite's credential-redaction contract. Apply the existing URL restrictions in `verifyConfig` before output/file creation, and test secret-bearing userinfo/query/fragment inputs for refusal without echoing or persisting their values.

The focused probe also confirmed private output/config modes (0700/0600). Existing HTTP/SSE requests use manual redirects; the shared profile composition retains bounded requests, turns, resources, and cleanup. The new tests exercise composition and summary helpers but miss the executable credential and output paths above.

Validation: the coordinating reviewer ran the exact-head suite in an isolated Git checkout: 252/252 Node tests passed, as did bash syntax, ShellCheck, and changelog checks. Those checks do not cover the two reproduced paths above.

Source review and focused reproductions are complete. All probe credentials were fake; keychain access was stubbed; only a loopback fixture received a request. No real suite key, deployment, provider resource, or live-account test was used. This is an independent Codex review, not a Review Loop App verdict.

BinaryBourbon (changes requested):
**REQUEST_CHANGES — independent adversarial API/operator compatibility review by Codex.**

Reviewed the whole eight-file PR from `11e98f5bb81055bc800e103ee44d759544dbaadf` to `42fd45431a4647fc8c4db195fdee0d62e579e54b`: the shell wrapper, `verify.mjs`, shared `target.mjs`, `ci.mjs`, verifier tests, both operator documents, and the changelog. Two concrete defects need correction:

1. **[P2] Four advertised profiles cannot produce a runnable target.** [`deployed/verify.mjs:60`](https://github.com/managoat/fountain/blob/42fd45431a4647fc8c4db195fdee0d62e579e54b/deployed/verify.mjs#L60) adds only contract/execution settings. Selecting `secrets`, `mcp`, `webhooks`, or `schedules` always fails `configFrom()` because their required receiver/schedule blocks are absent. The new CLI has no flags or target-file input that can supply them. With fake credentials and an inert URL, directly passing each generated target through the real parser reproduced all four setup failures before any request. Add the missing configuration path, or restrict this entry point's advertised/accepted profiles and direct these profiles to `cli.mjs`. Test composed targets through the actual parser; matching CI's profile names and turn limits does not establish runnability.

2. **[P2] The account-provisioning recipe cannot work with registration closed.** [`deployed/README.md:64`](https://github.com/managoat/fountain/blob/42fd45431a4647fc8c4db195fdee0d62e579e54b/deployed/README.md#L64) explicitly promises that the following public registration calls work regardless. `RegistrationController.api_create` calls `Accounts.register_user`, which checks `registration_allowed?` and returns HTTP 403 `registration_closed` when registration is disabled. There is no administrator bypass in these calls. Operators following the recipe cannot obtain the two prerequisite accounts. Document a supported provisioning route or require registration to permit these accounts during provisioning, then explain that it can remain closed during verification runs.

Independent validation: all nine generated profiles passed through the real configuration parser using fake credentials; the four failures above were reproduced. Twenty-four synthetic CI profile/mode compositions matched the exact base output. Traced profile execution, HTTP/auth contracts, account verification/onboarding, credit gates, and the documentation against current source. No server wire models or SDK declarations change.

The coordinating reviewer reports 252/252 Node tests plus shell/changelog checks passing in an isolated checkout; [CI also passed on this head](https://github.com/managoat/fountain/actions/runs/35095757401). Those checks do not catch these defects. No deployment was contacted or real suite credentials read. This is an independent Codex review, not a Review Loop App verdict.

BinaryBourbon (changes requested):
**REQUEST_CHANGES — fresh consolidated adversarial re-review by Codex.**

Rechecked the whole PR against base `11e98f5bb81055bc800e103ee44d759544dbaadf`. The head remains `42fd45431a4647fc8c4db195fdee0d62e579e54b`; no fixes have landed since the earlier reviews. Four supported blockers remain:

- **P1:** Automatic keychain fallback still sends the production suite key to an unrelated selected origin. Bind stored credentials to the normalized target origin or require explicit selection for unbound targets.
- **P2:** Credential-bearing URLs are still printed and persisted before the downstream URL validation rejects them. Validate these restrictions before output or file creation.
- **P2:** `secrets`, `mcp`, `webhooks`, and `schedules` remain advertised but cannot receive their required configuration through this command. Support that configuration or explicitly direct these profiles to the underlying CLI.
- **P2:** The provisioning guide still promises public registration works when registration is closed; the actual API returns 403 `registration_closed`. Document a supported prerequisite/path.

I reran the inspected security reproduction using a stub keychain and a loopback fixture, and rechecked all four generated integration targets through the real configuration loader. Both security defects and all four setup failures reproduced. The closed-registration API trace is unchanged. Details and regression-test expectations remain in the [security review](https://github.com/managoat/fountain/pull/2345#pullrequestreview-5222681378) and [correctness review](https://github.com/managoat/fountain/pull/2345#pullrequestreview-5222675038).

Coverage: `deployed/verify.mjs`, `scripts/verify-deployment.sh`, `deployed/lib/target.mjs`, `deployed/ci.mjs`, `deployed/test/verify.test.mjs`, `deployed/README.md`, `docs/guides/operate/deploy.md`, and `changelog.d/1612-verify-a-deployment.md`, including setup, credential flow, bounded execution, cleanup, and operator instructions.

The previous 252/252 harness tests and passing CI apply to this unchanged head; I did not repeat the full suite. No real credentials or deployment were used. No additional blockers found.

jhgaylor:
All four addressed in `dfc8bd587`. Every one reproduced before fixing.

**[P1] Stored keys are now bound to a target origin.** Credential resolution
moved out of the shell into `verify.mjs`, where the keychain account is
`<origin>|<VAR>` and there is **no unbound fallback**. An unrelated target
finds nothing and is told to export the keys. An exported variable still wins,
which is how CI and non-macOS machines supply them.

Confirmed against the real path: `scripts/verify-deployment.sh https://managoat.com probe`
passes for the bound origin, while the same command against
`http://localhost:4000` now answers `Set FOUNTAIN_SUITE_KEY…` instead of
sending that key. Regressions cover a mismatched target, a lookalike host
(`managoat.com.evil.test`), and your one-key-exported case.

Moving it out of bash is also what makes it testable, which is why the wrapper
is now a passthrough.

**[P2] Credential-bearing URLs are refused before any output.** `targetOrigin`
runs before resolution, printing or file creation, and rejects userinfo, query,
fragment and any path. The refusal deliberately does not quote the input back;
a test asserts the planted secret appears in neither the message nor the
control flow, and that no credential lookup happens for a refused target.

**[P2] The four unconfigurable profiles are no longer advertised or accepted.**
`VERIFY_PROFILES` excludes `secrets`, `mcp`, `webhooks` and `schedules`, and
selecting one is refused with a pointer to `cli.mjs`. Both help texts and both
documents say so.

The new regression is the one you asked for: it writes each advertised
profile's composed target and loads it through the real `configFrom`, rather
than comparing selected fields. That is a better test than the parity one — it
would have caught this.

**[P2] The provisioning recipe is corrected.** You are right that there is no
administrator bypass: `registration_allowed?/1` returns `registration_closed`
and `api_create` answers 403. The guide now says to provision while
registration is open or reuse existing accounts, and notes that verification
runs need no registration setting either way. It also documents the
`security add-generic-password` line for storing a key against a target.

Checks: 259 Node tests pass, up from 252. ShellCheck clean, `bash -n` clean,
changelog check passes, docs suite passes. The credential and URL paths that
the previous tests missed are now covered by the executable paths themselves.

BinaryBourbon (approved):

Reviewed all eight changed files at `dfc8bd587616b48787ac6fa9fdd741eed8e2fd9f` against base `11e98f5bb81055bc800e103ee44d759544dbaadf`, including the original implementation and subsequent fixes. No remaining concrete security defect introduced by this PR was found.

Both findings from my prior security review are resolved. Keychain lookup now uses the normalized target origin in the account name, with no unbound fallback; explicitly exported credentials take precedence. URL protocol, userinfo, query, fragment and path restrictions run before credential lookup, logging or artifact creation. The shell wrapper delegates these boundaries to the same Node entry point used directly.

I exercised the actual shell → Node → stub `security` → HTTP path using fake credentials and loopback fixtures. The checks passed for normalized origin matching; refusal of unrelated, lookalike and different-host origins; environment precedence without lookup; incomplete credentials failing before any request; and secret-bearing or disallowed URLs failing without lookup, output artifacts or echoed values. An HTTP redirect was not followed, and the sink received no request. Fake credential values in a response were absent from terminal output and persisted artifacts. New output/config permissions remained 0700/0600. All three HTTP requests were to the refusing local fixture; no resource writes occurred.

The other two prior blockers are also addressed: the command advertises only profiles it can configure, with a regression passing each generated target through the actual loader, and the provisioning guide now correctly distinguishes registration being open for account creation from reusing accounts with registration closed. The shared CI composition retains profile selection, execution/turn budgets, resource limits and cleanup deadlines. I also traced request redaction, manual redirects, result summaries, cancellation and fixture ownership/cleanup, and reviewed both guides and the changelog.

Validation: the coordinating reviewer reports 259/259 Node tests passed in an isolated checkout of this exact head, plus Bash syntax, ShellCheck, changelog and whitespace checks. [Exact-head CI passed](https://github.com/managoat/fountain/actions/runs/35108307716), including required CI and secret scanning.

Source review and the focused executable probes are complete. Keychain access was stubbed; no real credentials, deployment, account, provider resource or live cleanup run was used. This approval applies to the reviewed commit and is an independent Codex verdict, not a Review Loop App verdict.

BinaryBourbon (approved):
**APPROVE — independent adversarial correctness/operator re-review by Codex.**

Reviewed the whole PR at new head `dfc8bd587616b48787ac6fa9fdd741eed8e2fd9f` against base `11e98f5bb81055bc800e103ee44d759544dbaadf`. The four blockers from the earlier review are resolved at this revision:

- **Unrunnable advertised profiles:** resolved. The five advertised profiles pass the actual runner configuration loader. I also exercised the real CLI for all four integration profiles: each now gives an actionable `cli.mjs` direction before creating evidence.
- **Closed-registration provisioning:** resolved. The guide now distinguishes creating accounts while registration is open from reusing existing accounts while it is closed, matching the account API's admission check.
- **Unbound stored credentials:** resolved. Lookup uses the validated, normalized target origin in the keychain account name, with no legacy unbound fallback. My fake lookup checks confirmed matching-origin resolution and refusal for unrelated hosts, ports, and loopback targets.
- **Credential-bearing URL output:** resolved. Actual CLI probes rejected userinfo, query, and fragment inputs without echoing the fake secret or creating an output directory. Validation occurs before credential resolution and output.

Coverage: `deployed/verify.mjs`, `scripts/verify-deployment.sh`, `deployed/lib/target.mjs`, `deployed/ci.mjs`, `deployed/test/verify.test.mjs`, `deployed/README.md`, `docs/guides/operate/deploy.md`, and `changelog.d/1612-verify-a-deployment.md`. Rechecked CI/profile composition, argument handling, setup, bounded execution, cancellation, cleanup, and operator documentation. No additional blocking issue found.

Validation: my focused checks used fake credentials and no deployment. The coordinating reviewer ran 259/259 Node tests successfully in an isolated checkout of this head, plus Bash syntax, ShellCheck, changelog, and whitespace checks; [exact-head CI passed](https://github.com/managoat/fountain/actions/runs/35108307716). No live conformance run or real suite credentials were used. This approval applies to the new revision; earlier verdicts remain historical.

Issue #2344, opened 2026-09-16
Tracker: the machine has one owner (ADR 0058)


jhgaylor:
Tracker for the implementation of ADR 0058, *The machine has one owner* (#2342). The ADR's **Implementation brief** section is the authority for what each stage is and when it is done; this issue carries the status and the decision log. One PR per stage; each PR lowers the direct-writes ratchet or holds it.

## Baseline (`main` at `c3f568596`, 2026-09-16)

| | sites | files |
|---|---|---|
| `Conversations.update_sandbox/2` + `claim_sandbox/2` (the ratchet's row writes) | 28 | 11 |
| `Managoat.Sandbox.{create,resume,suspend,destroy,create_checkpoint}` | 17 | 9 |
| `expected_sandbox_id` | 7 | 4 |
| "is anyone here" predicates | 4 | 3 |

## Stages

| # | Stage | PR | Status |
|---|---|---|---|
| 1 | ADR 0058 filed; 0023 points at it | #2342 | **merged** `dca7de3b0` |
| 2 | The ratchet: `machines/direct_writes_test.exs` pinned at 28 + 17; credo ownership entries | #2350 | **merged** `306afe017` — ratchet pinned at 28 + 17 |
| 3 | Lease columns (`lease_epoch`, `lease_node`, `lease_until`, `transition`, `transition_reason`) and `Fountain.Machines.Lease` | #2346 | **merged** `90b4cf41b` — two review rounds: epoch-0 hole closed, `terminated_at` stamped |
| 4 | `Fountain.Machines.Machine`, read-only, behind `MACHINE_OWNER_ENABLED`; the four predicates delegate to `Occupancy` | #2348 | **merged** `e9acf52c1` — three review rounds: flag documented as inert, cost regression removed |
| 5a | Destroy through the owner, conversation side: `Machines.Destroy` protocol + `Machine.destroy/3`; retargets `Lifecycle.do_destroy/4`, `ConversationServer.terminate_machine/2`, `Termination.retire_terminated_sandbox/2` | #2352 | **merged** `9d334f35c` — three independent reviews × two rounds; ratchet 25/15; pin 2220 |
[...]

PR #2341, opened 2026-09-16
fix(web): tokenize dark-mode colors on Account, Audit log, Help and Admin

alecraso:
## Summary

Account, Security, Audit log, Help and every Admin page predate the
console's CSS-variable theme (`assets/css/tokens.css`) and were still built
on literal Tailwind `bg-white` / `border-zinc-*` / `text-zinc-*` classes,
which render the same color in both themes. `<body>`'s text color *is*
tokenized (`text-[var(--color-text-primary)]` in `root.html.heex`), so any
unstyled text inside one of those cards inherited dark mode's near-white
primary color while the card itself stayed white — pale, near-invisible
text sitting on a card that never left light mode. The admin tab bar had
the same bug in reverse: its active-tab indicator was a literal near-black
border/text, invisible on the dark page background.

- Swaps the affected neutral (white/gray/zinc) classes for the
  `var(--color-*)` arbitrary-value convention `dashboard_live/index.ex`
  and `agents_live/index.ex` already use, so these pages track
  `[data-theme="dark"]` like the rest of the console.
- `help_live/show.ex`'s prose block had explicitly dropped the `dark:`
  Typography-plugin variants that `/docs` (`docs_html/show.html.heex`)
  uses for the same content — restored them.
- Named status/semantic colors (red, amber, green, purple, sky, etc.) are
  left as-is: they pair a literal background with a literal, non-inherited
  text color, so they stay internally legible in both themes. That's a
  separate, lower-priority cosmetic inconsistency, not the contrast bug
  reported here.

Closes #2340

## Test plan

- [x] `mix format --check-formatted`
- [x] `mix compile --warnings-as-errors`
- [x] `mix precommit` (9/9 stages passed, including credo --strict,
      dialyzer and sobelow)
- [x] Reproduced the bug live on a self-hosted instance in dark mode
      (`/account`, `/audit`, `/admin`) before fixing, confirming the
      literal-background-with-inherited-text mechanism
- [ ] Visual re-check of the fixed pages in dark mode — not done in this
      PR; the local self-hosted instance runs a published image, not a
      mount of this checkout, so the rendered fix couldn't be screenshotted
      here. The class substitutions reuse a pattern already proven correct
      on `/dashboard` and `/agents` in the same running instance.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2329, opened 2026-09-15
fix(sandbox): reap a teardown that fenced and then died

BinaryBourbon:
Item 7 of #2021. Independent of #2327 (item 1) — different file, no shared
lines; either can merge first.

## The gap

A teardown fence commits in its own transaction, before any provider I/O, and
the terminal write lands after it. Anything in between can lose:

- `Managoat.Sandbox.destroy/1` or `Egress.release/2` raises
  (`lifecycle.ex:472-486`);
- the retirement write after the fence is refused;
- the pod dies between the fence commit and the terminal write;
- an account deletion halts mid-fence.

What is left is a row with `teardown_requested_at` set and a live status, and
no reaper pass could see it:

| pass | why it skips the row |
|---|---|
| `release_stuck_sandboxes/0` | `is_nil(s.reset_requested_at)` — the fence always sets it |
| `sweep_abandoned_sandboxes/0` | same predicate |
| `destroy_dead_sprites/1` | wants a terminal status |
| `report_untracked/1` | the sprite has a row, so it counts as known |

Meanwhile `Quotas.active_sandboxes/0` (`quotas.ex:145`) goes on counting the
row against the tenant cap and the fleet ceiling, and the machine bills at the
provider. The only exit was an operator noticing the row and clicking Reap in
`/admin/sandboxes`, with nothing surfacing it for them to notice.
`Accounts.Deletion` already carries a comment describing exactly this dead end
as its reason for logging a refused fence and carrying on
(`deletion.ex:223-230`).

## The fix

A pass that looks for the fenced-and-stale case rather than excluding it:

```elixir
not is_nil(s.teardown_requested_at) and s.status not in ^@terminal_statuses and
  s.teardown_requested_at < ^cutoff
```

then the same liveness check the other sweeps use, then the terminal write. The
existing dead-sprite pass destroys the machine on that same run, exactly as
`expire/2` already relies on.

Three things this deliberately does not do:

- **It never starts a teardown.** `teardown_requested_at` is written by the
  fence alone, so a row only reaches here because a caller already decided the
  machine was to go away — and the fence already recorded and audited that
  intent. This finishes it; it does not decide it.
- **A plain `reset_requested_at` is not a predicate.** An ordinary reset means
  wipe-and-rebuild, not terminate, and `SandboxResetReconciler` retries those.
  Terminating one here would destroy a home the tenant asked to keep.
- **The conversations are left alone**, for the same reason `expire/2` leaves
[...]

PR #2346, opened 2026-09-16
feat(machines): lease columns and Fountain.Machines.Lease (ADR 0058, stage 3)

BinaryBourbon:
## What

Stage 3 of the ADR 0058 stack (#2344): the lease columns and the pure module
that operates them. Additive schema plus `Fountain.Machines.Lease`. **Nothing
reads or calls either yet** — `Fountain.Machines.Machine` (stage 4 onward) is
the only intended caller, and every existing fence stays in force.

**Migration** `20260916120000_add_machine_lease_to_sandboxes.exs` adds five
columns to `sandboxes`: `lease_epoch bigint not null default 0`, `lease_node`,
`lease_until`, `transition`, `transition_reason`. `SET LOCAL lock_timeout`/
`statement_timeout` per the `20260915230000` precedent. No backfill and no
index — the owner reaches a lease by primary key under the per-sandbox
advisory lock and never scans for one. A `NOT NULL` column with a constant
default does not rewrite the table on PG 11+, so this stays a catalog change.

**Schema**: the five fields land on `Fountain.Conversations.Sandbox` with a
comment saying they are the machine owner's, written only by
`Fountain.Machines.Lease`. They are deliberately **not** in `changeset/2`'s
cast list — an owner's write is not something a caller's attrs may reach. Adds
`@transitions ~w(provisioning resuming parking destroying retargeting)` and
`transitions/0`.

**Public API** — every function is one short transaction, refuses an enclosing
one with `{:error, :transaction_open}`, and does no provider I/O:

| Function | Returns |
|---|---|
| `claim(sandbox_id, node, ttl_ms, now \\ utc_now())` | `{:ok, epoch}` / `{:error, {:held, node, until}}` / `{:error, :not_found}` |
| `take_over(sandbox_id, node, ttl_ms, now \\ utc_now())` | same as `claim/4` |
| `renew(sandbox_id, epoch, ttl_ms, now \\ utc_now())` | `:ok` / `{:error, :lost}` |
| `release(sandbox_id, epoch)` | `:ok` / `{:error, :lost}` |
| `cas_update(sandbox_id, epoch, attrs)` | `{:ok, %Sandbox{}}` / `{:error, :stale}` / `{:error, :retired}` / `{:error, {:invalid, field}}` |

`renew`, `release` and `cas_update` match on the **holder columns as well as
the epoch**, and refuse an epoch below 1 before querying: `lease_epoch`
defaults to 0 and `release/2` keeps its epoch, so neither is a lease anybody
holds. `cas_update` also mirrors `prevent_sandbox_revival/1` — a terminal row
is never written back to a live status, answered `:retired`.

A database fault is answered as `{:error, {:database, sqlstate}}` rather than
raised out of the rollback.

## Why

ADR 0058's "the owner is a lease first and a process second": a process alone
has two instances across a rolling deploy, so the source of truth has to be on
the row. This is #2307 constraints 1–3 implemented once, in one module,
instead of per writer — the compare-and-set that makes a superseded owner's
write invisible, the expiry-based takeover that needs no eviction, and the
serialization that a transaction-scoped lock cannot span provider I/O to
provide.

Two contracts are stated in the code because later stages depend on them:

- **`take_over/4` is `claim/4` with a different log line**, and its doc says so.
  There is no way to evict a live holder; a TTL that has not run out is not a
  hand-over. Recovery is: the lease expires, a new claimant takes a new epoch
  under the lock, and the old holder's in-flight writes fail their CAS.
- **`cas_update/3` is the one write primitive the owner gets.** One guarded
  `update_all` on `(id, lease_epoch)` with `select: s`; `attrs` may name only
  `status`, `transition`, `transition_reason`, `terminated_at`,
  `last_resumed_at`, with atom keys, validated against `Sandbox.statuses/0` and
  `Sandbox.transitions/0` before anything is written. Anything else is
  `{:error, {:invalid, field}}` and the row is untouched.

No audit events — a lease is control-plane bookkeeping and nothing
user-facing happens; the events an operation owes (`sandbox.destroyed` and its
siblings) are the owner's verbs to record in stage 5+. Said in a comment. No
changelog fragment: unused columns, nothing observable.

## Migration version and the collision check

Chose **`20260916120000`**. Checked against:

- `git fetch origin main` then `git ls-tree origin/main apps/fountain/priv/repo/migrations` — newest on `main` is `20260915230000_add_explicit_credential_set_access.exs`.
- `gh pr list --state open --limit 100` returns exactly three PRs (#2345, #2342, #2341); `gh pr view <n> --json files` shows **none** of them touches `apps/fountain/priv/repo/migrations`. Stage 2 of this stack is test-only and carries no migration.

## Tests

`apps/fountain/test/fountain/machines/lease_test.exs`, 16 cases. The ones
needing two real connections run under `unboxed_run/2` and delete what they
commit (verified: zero leftover rows, zero leftover triggers).

- two concurrent `claim`s: exactly one `{:ok, 1}`, one `{:error, {:held, …}}`, epoch advanced by exactly 1
- a `claim` waits on the per-sandbox advisory lock, observed through `pg_blocking_pids` while a second connection holds key `(4316, phash2(id))`
- a stale epoch: `cas_update` `:stale`, `renew`/`release` `:lost`, row byte-for-byte unchanged
- takeover refused while the lease is live (injected `now`) and while it is renewed, allowed once `lease_until < now`; the old holder's `renew` is then `:lost` and its `cas_update` `:stale`
- `release` then `claim` gives a strictly higher epoch; the released holder's write stays dead
- a `BEFORE UPDATE` trigger raising SQLSTATE `57014` (scoped to the one row, so it cannot disturb the rest of the table): `claim` answers `{:error, {:database, :query_canceled}}`, the row is unchanged, and a claim from another connection then succeeds — proving the advisory lock went back with the rollback
- an expired lease nobody has taken over still writes — `cas_update` reads the epoch, never `lease_until`; pinned so a later stage does not read the TTL as an authorization
- `cas_update` refuses an unknown status, transition, value type, column, and a string key, without touching the row
- every function refuses inside `Repo.transaction/1`
- the happy path: claim, stamp a transition, finalize it, release

**Verified by reverting**, each probe restored from a copy:

| Guard removed | Test that failed |
|---|---|
| the advisory lock | "a claim waits on the per-sandbox advisory lock" |
| the advisory lock **and** `FOR UPDATE` | both concurrency tests, on three seeds |
| the epoch from the CAS `where` | three tests, including "a stale epoch writes nothing at all" |
| the `Postgrex.Error` rescue | the SQL-fault test, with the error raised |
| the `in_transaction?` refusal | "an enclosing transaction is refused by every function" |
| the `status` domain check | "cas_update refuses a value or a key the owner may not write" |

## Review round (jhgaylor, 2026-09-16)

Requested changes on `261ebd9c3`; all blocking findings reproduced locally
before fixing, and each fix verified by reverting the guard it adds.

**Blocking — one root cause.** `held_by/2` matched `(id, lease_epoch)` alone
and `lease_epoch` defaults to 0, so epoch 0 was a lease nobody held and anybody
could quote. `renew/4` on a never-claimed row wrote `lease_until` with
`lease_node` nil, which `live?/2` read as held — the machine became unclaimable
by any node for a full TTL, with the refusal naming the holder as `nil`. The
same predicate let a `renew/4` landing after `release/2` re-arm the lease just
surrendered (stage 4's idle-stop racing its own heartbeat), and let
`cas_update/3` write at epoch 0 with no epoch ever taken and no lock ever held.

Fixed by requiring `lease_node`/`lease_until` in `held_by/2` and refusing
`epoch < 1` in `taken_epoch?/1`. `do_claim/5` keeps a bare epoch predicate — the
row it claims is unheld by definition and the advisory lock plus `FOR UPDATE`
have already picked the winner — with a comment saying so. `live?/2` now needs
a holder as well as a clock.

**Non-blocking, all taken.**

- `cas_update/3` mirrors `Conversations.prevent_sandbox_revival/1` and reports
  `:retired` rather than `:stale`, so an owner does not go hunting a takeover
  that never happened. Terminal-to-terminal and status-free writes are
  unaffected, matching the original exactly. Docs now say **membership, not
  legality** — transition legality is `Machines.Policy`'s job in a later stage.
- `guarded/2` rejects a non-UUID `sandbox_id`; `cast_attrs/1` rejects a keyword
  list naming a column twice. Both previously raised `Ecto.Query.CastError` /
  `Ecto.QueryError` out of the transaction. The rescue widens to
  `DBConnection.ConnectionError`, and its comment now scopes the promise to
  faults the database raises — a caller's own bug is refused up front rather
  than dressed as one.
- `transition_reason` is `:text`. A long reason was answering
  `{:error, {:database, :string_data_right_truncation}}`.
- **Clock: kept the BEAM clock, documented the skew.** The moduledoc now says
  `lease_until` is written from one node's clock and compared against another's,
  that skew degrades to early takeover (which the CAS already makes safe) or
  late takeover (which only delays recovery), and that moving to
  `fragment("now()")` belongs to the process that owns the renew timer in stage
  4 — this module has no timer. Flagging it rather than doing it keeps this
  round to the blocking bug.
- Test 5's "row unchanged" assertion is now marked as the weak half (the
  trigger is BEFORE UPDATE); the second-connection claim is named as the
  load-bearing one.
- ADR arity drift (`claim/2` vs `claim/4`) and the missing stage-2 ratchet in
  this base: acknowledged, no change.

**Six regression tests added**, each verified by reverting its guard. The two
epoch guards mask each other on a never-claimed row, so that case is also run
with **both** removed — which reproduces the reviewed behaviour exactly:

| Guard removed | Test that failed |
|---|---|
| holder columns in `held_by/2` | "a release cannot be undone by a renew that lands after it" |
| both holder columns **and** `taken_epoch?/1` | the above **plus** "epoch 0 is not a lease anybody took" |
| `refuse_revival/2` | "cas_update does not revive a retired machine" |
| the duplicate-key check | "a caller's own bug is an answer, not a raise" |
| the UUID check | "a caller's own bug is an answer, not a raise" |

## Checks run

- `mix precommit` from the root, on the fixed tree: **PASSED (9 stages)** — toolchain, conflict-markers, compile `--warnings-as-errors`, deps, format, credo `--strict`, dialyzer, sobelow, release.
- `mix format` from both the umbrella root and `apps/fountain`.
- `mix test test/fountain/machines/lease_test.exs` plus the suites that touch
  the schema: `sandbox_test`, `sandbox_mode_test`, `sandbox_identity_test`,
  `sandbox_reset_test`, `sandbox_reset_concurrency_test`, `lifecycle_test`,
  `lifecycle_fence_test`, `teardown_fence_test` — **145 tests, 0 failures**.

## Notes for the next stage

- `Lease` reuses `Conversations.with_sandbox_lock/2` rather than replicating
  the lock; `conversations.ex` is otherwise untouched. That leaves a
  `Machines → Conversations` dependency for stage 9 to invert.
- The enclosing-transaction refusal is `:transaction_open` (the
  `SandboxIdentity` spelling) rather than `:provider_transaction_open`, since
  this module does no provider I/O itself.
- `cas_update/3` bumps `updated_at`; the lease operations deliberately do not,
  so a heartbeat every few seconds does not make the column meaningless.
- `claim/4` guards its write on the epoch the locked read saw and answers
  `{:error, :lost}` rather than matching `{1, _}`, per the #2329 rule.

Base is `adr/0058-machine-owner` (#2342), not `main`. Do not arm `--auto`.


jhgaylor (changes requested):
**Request changes.** The lock reuse, the CAS, the migration and the allowlist are all right, and I could not break them. But `held_by/2` matches on `(id, lease_epoch)` alone, and `lease_epoch` defaults to `0` on every row — so epoch 0 is a lease that nobody holds and everybody can quote. Two reproductions below: `renew/4` manufactures a lease out of nothing and makes a sandbox un-claimable for a full TTL, and `cas_update/3` writes without any lease ever having been taken. Both are one `WHERE` clause from fixed, and every later stage inherits the primitive as written.

Everything below was reproduced in a detached worktree at `261ebd9c3` against a private database, not read off the diff.

## Blocking

### 1. `renew/4` manufactures a lease nobody holds, and un-does `release/2`

`renew/4` (`apps/fountain/lib/fountain/machines/lease.ex:101`) writes `lease_until` on any row matching `held_by/2` (`:228`), which checks the epoch and nothing else. It never asks whether a lease exists. Two ways in:

**(a) On a row that was never claimed.** `lease_epoch` is `0` by default, so `held_by(id, 0)` matches:

```elixir
s = insert_sandbox(status: "pending")            # lease_epoch: 0, lease_node: nil
Lease.renew(s.id, 0, 60_000)                     #=> :ok
# row is now {epoch: 0, node: nil, until: +60s}
Lease.claim(s.id, "a", 60_000)                   #=> {:error, {:held, nil, ~U[...]}}
Lease.take_over(s.id, "a", 60_000)               #=> {:error, {:held, nil, ~U[...]}}
```

A sandbox nobody has ever leased is now un-claimable by any node for the whole TTL, and `live?/2` (`:217`) is satisfied by a `lease_until` with no `lease_node`, so the refusal names the holder as `nil`. An operator reading that refusal is told a machine is held by nothing.

**(b) After a release, which is the shape stage 4 will actually hit.** `release/2` (`:122`) deliberately keeps the epoch — documented, and right for the CAS — but that also leaves the epoch matchable by `renew`:

```elixir
{:ok, 1} = Lease.claim(s.id, "fountain@a", 60_000)
:ok      = Lease.release(s.id, 1)                # node and until cleared, as documented
:ok      = Lease.renew(s.id, 1, 60_000)          # <- re-arms the lease it just gave up
Lease.claim(s.id, "fountain@next", 60_000)       #=> {:error, {:held, nil, ~U[...]}}
```

The docstring on `release/2` says "Give the lease up". Any `renew` at the same epoch that lands after it takes it back, and the row is then held by a holder that does not exist. Stage 4's `Machine` is specified as *idle-stopping* with a renew timer; a surrender racing one in-flight renew is exactly this. Nothing in the module prevents it and nothing tests it.

Fix is small and belongs here rather than in the caller: require an existing lease in `renew` (`... and not is_nil(s.lease_until)`, or `and s.lease_node == ^node`), and/or have `release/2` advance the epoch as well as clearing the holder.

Reproduced: `Fountain.Machines.Rv2346ProbeTest` "PROBE B" and `Rv2346Probe2Test` "PROBE J", `mix test test/fountain/machines/rv2346_probe*_test.exs` from `apps/fountain` against `fountain_test_0058_rv3`.

### 2. `cas_update/3` at epoch 0 writes with no lease and no lock

Same root cause, on the write primitive. The moduledoc at `:145` justifies taking no advisory lock with "the serialization this needs was done when the epoch was taken". At epoch 0 no epoch was ever taken and no lock was ever held:

```elixir
s = insert_sandbox(status: "pending")
Lease.cas_update(s.id, 0, %{status: "ready"})    #=> {:ok, %Sandbox{status: "ready"}}
Lease.cas_update(s.id, 0, %{status: "failed"})   #=> {:ok, ...}   # a second, unrelated writer
```

Two callers that both never claimed can both write, serialized by nothing but the row lock, each believing it is the owner. Claims already start at `1` (`do_claim/5:193`), so refusing `epoch < 1` in `held_by/2` costs nothing and closes it.

This also makes the existing test prove less than its comment says. `lease_test.exs:82` reads "Epoch 0 is what every row starts at, so it is the epoch a caller that never held the lease would quote" and asserts `{:error, :stale}` — but that only holds because the row in that test had already been claimed to epoch 1. On a never-claimed row the identical call writes. The case the comment describes is the one case not covered.

Reproduced: "PROBE A" and "PROBE A2" in the same file.

## Non-blocking

- **`cas_update/3` bypasses the retirement guard every other writer has.** `Conversations.update_sandbox/2` refuses a revival through `prevent_sandbox_revival/1` (`{:error, %Ecto.Changeset{errors: [status: {"sandbox is retired", []}]}}`); `Lease.cas_update(id, epoch, %{status: "ready"})` on the same `terminated` row returns `{:ok, ...}` and writes it (PROBE C). That may well be deliberate — the ADR moves transition legality into `Machines.Policy` — but the moduledoc's "`status` is checked against `Sandbox.statuses/0`" (`:139`) reads like validation when it is membership. One sentence saying legality is a later stage's job would stop the next reader trusting it.
- **`guarded/1` (`:263`) rescues only `Postgrex.Error`**, which is narrower than "a database fault is an answer, not a crash". A duplicate key in a keyword `attrs` (`cas_update(id, e, status: "failed", status: "suspended")`) raises `Ecto.QueryError`; a malformed `sandbox_id` raises `Ecto.Query.CastError` (the `is_binary` guard admits it, the `@spec` says `Ecto.UUID.t()`); a client-side `DBConnection.ConnectionError` would escape too. Either widen it or scope the promise to server-side SQLSTATEs in the docs.
- **The lease clock is the BEAM node's, not the database's.** `lease_until` is written from one node's `DateTime.utc_now()` and compared against another node's in `live?/2`. Skew degrades to premature takeover, which the CAS already makes safe, so this is not a correctness hole — but `fragment("now()")` would make it one clock instead of N, and the injected `now` could stay for the tests.
- **`transition_reason` is `varchar(255)`.** A longer reason surfaces as `{:error, {:database, :string_data_right_truncation}}` rather than a validation. `:text` is kinder for a free-text field.
- **Test 5's "no write" assertion is trivially true.** The trigger is `BEFORE UPDATE`, so the row could not have changed whether or not the rollback worked; the load-bearing half is the second connection's claim, which is good. I did confirm the path is genuinely exercised (below).
- **ADR drift, benign.** Stage 3 names `claim/2`, `renew/2`, `release/2`, `take_over/2`; shipped as `/4`, `/4`, `/2`, `/4` with an injected clock. Better for the tests, worth nothing more than a mention. `renew`/`release` also skip the advisory lock the stage row asks for — I checked, and it is safe (a concurrent `claim`'s `FOR UPDATE` forces the epoch predicate to re-evaluate against the winner's row, so a losing `renew` gets `:lost`), and `renew`'s docstring already says why.
- Stage 2's ratchet (`test/fountain/machines/direct_writes_test.exs`) is not in this PR's base, so "each PR lowers the ratchet" could not be checked here.
- I agree with the absence of an `_unsafe_` prefix: there is no tenant scope to skip, and the convention in `conversations.ex:42-57` is about user-facing fetches.
- No changelog fragment is correct — nothing observable ships.

## Verified

- **Lock namespace.** `do_claim/5:176` calls `Conversations.with_sandbox_lock/2` itself, so it is literally `@sandbox_lock_namespace 4316` + `:erlang.phash2(sandbox_id)` (`conversations.ex:1104`), the same key the teardown fence takes by hand (`lifecycle.ex:675`). Proved dynamically rather than by reading: one connection held a real `with_sandbox_lock` on the sandbox, `Lease.claim` on a second returned `nil` from `Task.yield` throughout, then `{:ok, 1}` the moment the holder committed. Blocks; does not deadlock or time out (PROBE K).
- **CAS returns the post-write row**, not a reload: `{"ready", "resuming", true}` for `{status, transition, returned.updated_at == Repo.get!(...).updated_at}` — `UPDATE ... RETURNING` (PROBE E).
- **The attrs allowlist holds.** `%{mode: ...}`, `%{lease_epoch: 99}`, `%{user_id: ...}`, `%{"status" => "ready"}` and `%{}` are all refused before any write, and `Ecto.Type.cast` rejects `terminated_at: "whenever"` and `transition_reason: 17`. Keyword and map both work.
- **Epoch monotonicity.** `release` keeps the epoch, the next `claim` gets exactly +1, and the released holder's `cas_update` is `:stale`. `take_over` is refused while `lease_until >= now` *including from the node that already holds it* (`{:error, {:held, "fountain@probe", ...}}`), and succeeds once expired.
- **Transaction discipline.** All five public functions answer `{:error, :transaction_open}` inside `Repo.transaction` (the PR's own test), and the harder case is clean (PROBE K above).
- **The SQL-fault path is real.** I weakened the trigger as asked: `57014` → `{:error, {:database, :query_canceled}}`, `22023` → `:invalid_parameter_value`, `P0001` → `:raise_exception`, for both `claim` and `cas_update`, and after all six faults a claim from another connection returned `{:ok, 1}` — so the advisory lock came back every time. The rescue is generic and the test does exercise it (PROBE L).
- **Migration.** `SET LOCAL lock_timeout`/`statement_timeout` on both `up` and `down`; `mix ecto.migrate` → `mix ecto.rollback --step 1` → `mix ecto.migrate` is clean on a private database, and the columns land as `bigint not null default 0`, `varchar(255)`, `timestamp`. `20260916120000` is ahead of main's highest (`20260915230000`) and none of the open PRs (#2341, #2342, #2345, #2348, #2349) touches `priv/repo/migrations`.
- **Schema.** `Sandbox.changeset(%Sandbox{}, %{lease_epoch: 5, lease_node: "evil@node", transition: "parking", ...})` yields `changes == [:user_id, :machine_name]` — all five fields are uncastable (PROBE F). `Sandbox.transitions/0` is `["provisioning", "resuming", "parking", "destroying", "retargeting"]`, matching the ADR's Decision section exactly.
- `mix credo` clean on both changed files; every `@spec` matched the returns I actually produced. PR CI green on `261ebd9c3`.

jhgaylor (approved):
**Approve.** All three blocking findings are fixed at `192243cf1`, each with a regression test that I confirmed fails when the fix is reverted. The fixes are the right ones — `held_by/2` matching the holder columns is the real guard, and `do_claim/5` keeping a bare epoch predicate is correct rather than an oversight. Re-reviewed in a fresh detached worktree at `192243cf1`; the three reproductions and the round-1 verifications were re-run, not assumed.

## The three reproductions, re-run

| Round 1 | `261ebd9c3` | `192243cf1` |
|---|---|---|
| `renew(id, 0, ttl)` on a never-claimed row | `:ok`, row becomes `{0, nil, +60s}`, un-claimable for a TTL | `{:error, :lost}`, row stays `{0, nil, nil}`, `claim` → `{:ok, 1}` |
| `claim` → `release(id, 1)` → `renew(id, 1, ttl)` | `:ok`, lease re-armed, `claim` → `{:error, {:held, nil, …}}` | `{:error, :lost}`, row stays `{1, nil, nil}`, `claim` → `{:ok, 2}` |
| `cas_update(id, 0, %{status: "ready"})` on a never-claimed row | `{:ok, %Sandbox{status: "ready"}}` | `{:error, :stale}`, status still `pending` |

## On the two guards masking each other (you flagged this — you are right, and it is narrower than you feared)

I neutered each guard in turn against the new tests, restoring from a pristine copy each time:

- **Both removed** (`held_by/2` back to a bare epoch predicate *and* `taken_epoch?/1` → `true`): **2 failures** — "epoch 0 is not a lease anybody took" and "a release cannot be undone by a renew that lands after it". Both reproductions come straight back, so the tests are real.
- **`held_by/2`'s holder columns removed alone:** **1 failure** — the release test. That guard is individually pinned, and it is the load-bearing one.
- **`taken_epoch?/1` removed alone:** **0 failures.** Nothing pins it.

That last one is not a hole, and I would not hold the PR for it: `taken_epoch?/1` cannot be load-bearing for any *reachable* row, because `do_claim/5` is the only writer of `lease_node`/`lease_until` and it always moves the epoch to ≥ 1, so no row can exist at epoch 0 with a holder. It is defence in depth against a future writer that breaks that invariant — which is exactly the kind of guard that gets deleted in a refactor because no test goes red. If you want it pinned, forging the unreachable state directly does it in four lines, and I checked it works:

```elixir
Repo.update_all(from(x in Sandbox, where: x.id == ^s.id),
  set: [lease_epoch: 0, lease_node: "ghost@node",
        lease_until: DateTime.add(DateTime.utc_now(), 60_000, :millisecond)])

Lease.renew(s.id, 0, 60_000)                  #=> {:error, :lost}      <- only taken_epoch?/1
Lease.release(s.id, 0)                        #=> {:error, :lost}
Lease.cas_update(s.id, 0, %{status: "ready"}) #=> {:error, :stale}
```

Non-blocking, and your call.

## `do_claim/5`'s bare epoch predicate

Correct as written, and I could not make it claim a held row. The predicate runs inside the transaction that already holds the advisory lock *and* `FOR UPDATE` on that row, after `live?/2` decided on that same locked read — so nothing can change `lease_epoch` between the read and the update, and `held_by/2`'s holder columns would indeed refuse every first claim. Checked:

- `claim` from another node while a live lease is held → `{:error, {:held, "fountain@probe", …}}`, row unchanged at `{1, "fountain@probe"}`.
- `claim` from **the same node** that already holds it → also `{:error, {:held, …}}`. No self-renewal-by-claim door.
- `take_over` from the same node while live → `{:error, {:held, …}}`.
- First claim on a released row (epoch 1, no holder) → `{:ok, 2}`, which is the case the bare predicate exists for.
- The narrowed `@type held :: {:held, String.t(), DateTime.t()}` is now truthful: `live?/2` refuses `%{node: nil}`, so the `nil` holder can no longer be reported.

## `:retired` vs `:stale` ordering

**`:stale` wins**, which is the right answer — a caller that is not the holder has no business being told about the row's status — but it is not written down anywhere:

```elixir
# holder at epoch 1 terminates the machine; a takeover moves it to epoch 2
Lease.cas_update(id, 1, %{status: "ready"})   #=> {:error, :stale}     (superseded epoch)
Lease.cas_update(id, 2, %{status: "ready"})   #=> {:error, :retired}   (held epoch)
```

`zero_row_reason/2` gets this by construction — it reads through `held_by/2`, so a superseded epoch finds no row and falls to `:stale` before the status is ever consulted. One clause in the `cas_update/3` docstring ("a superseded epoch is `:stale` even on a retired row; `:retired` means you *are* the holder and the row is terminal") would pin the intent, since the ordering currently depends on a `defp` reading through the same helper. Non-blocking.

I also checked `refuse_revival/2` really is an exact mirror of `prevent_sandbox_revival/1` (`conversations.ex:275`): that one errors iff `data.status` is terminal and `get_field(:status)` is not, and `get_field/2` falls back to the stored value when `attrs` omits the key. `refuse_revival/2` adds its `WHERE` iff `sets` names a status that is not terminal. Same truth table, including terminal→terminal and status-free writes, both of which I confirmed still go through.

## Non-blocking

- **You mirrored one of `update_sandbox_if/3`'s two guards, not both.** `prevent_sandbox_revival/1` is now mirrored; `stamp_terminated_at/1` is not, and that is the one with a billing consequence:
  ```elixir
  Lease.cas_update(id, epoch, %{status: "terminated"})  #=> terminated_at: nil
  Conversations.update_sandbox(s, %{status: "terminated"})  #=> terminated_at: ~U[2026-09-16 13:26:39Z]
  ```
  `terminated_at` is in `@writable`, so an owner *can* pass it — but that comment at `conversations.ex:284-294` exists precisely because "of the dozen writers of a terminal status, the ones that terminate passed a timestamp and the ones that fail never did", which left failed sandboxes reading as still running to `Billing.SandboxUsage`. `cas_update/3` is writer thirteen with the same hazard, and the new retirement test writes `terminated` then `failed` through it without ever looking at the column. Mirroring the second guard too, or one line in the docstring saying a terminal status must carry its own `terminated_at` and a stage-5 test that pins it, would close it. Not blocking — nothing calls this yet and stage 5 is where it bites.
- **The UUID guard has one gap left.** `Ecto.UUID.cast/1` accepts a raw 16-byte binary, so `cas_update(:binary.copy(<<0>>, 16), 1, %{status: "ready"})` passes `guarded/2` and then raises `Ecto.Query.CastError` from the query — the one thing the new check was added to prevent. `byte_size(sandbox_id) == 36` alongside the cast, or `Ecto.UUID.cast/1`'s result being what gets queried, closes it. Obscure enough that I mention it only because the moduledoc now makes the promise.
- **`cas_update/3`'s `@spec` says `map()`, and a struct is a map.** `cas_update(id, epoch, %Sandbox{})` raises `Protocol.UndefinedError` from `Enum.reduce_while/3`. `attrs when is_map(attrs) and not is_struct(attrs)` or a narrower spec would match the care taken elsewhere.
- **`cas_update/3` is two statements on the refusal path now.** `zero_row_reason/2` runs after the failed `update_all` with no transaction around the pair, so the reason it reports is a fresh read and can disagree with the write it is explaining. That is fine for a diagnostic and you say so in the comment — worth a word in the docstring too, since the moduledoc still says "every function here is one short transaction".

## Verified again at `192243cf1`

- **The advisory lock.** With a real `Conversations.with_sandbox_lock/2` held on the sandbox from another connection, `Lease.claim` returned `nil` from `Task.yield` for 700ms and `{:ok, 1}` the moment the holder committed. Still blocks, still no deadlock.
- **The fault path, weakened four ways.** `57014` → `:query_canceled`, `22023` → `:invalid_parameter_value`, `P0001` → `:raise_exception`, `40001` → `:serialization_failure`, for `cas_update/3` and `renew/4`, and a `take_over` from another connection afterwards returned `{:ok, 2}` — the advisory lock came back every time. The `DBConnection.ConnectionError` arm I could not induce, but it is one clause and `sqlstate/1` covers it.
- **Caller bugs are answers.** Duplicate keyword keys → `{:error, {:invalid, :status}}` (and `{:invalid, :transition}` for `transition: "parking", transition: nil`, which is the nastier one since `nil` is a legal value); all five entry points refuse `"not-a-uuid"` with `{:error, {:invalid, :sandbox_id}}`; a negative epoch is `:stale`/`:lost` rather than a query.
- **Allowlist and post-write row unchanged.** `mode`, `lease_epoch`, `lease_node`, `user_id`, `updated_at`, string keys and `%{}` all refused before any write, row byte-identical afterwards; `cas_update` still returns the `RETURNING` row (`{"ready", "resuming", true}` against a fresh `Repo.get!`).
- **Schema untouched by the fix.** `Sandbox.changeset/2` still yields `changes == [:user_id, :machine_name]` from an attrs map naming all five lease fields, and `transitions/0` still matches the ADR.
- **Migration.** `rollback --step 1` then `migrate` on a private database picks up the type change cleanly; `transition_reason` is `text`, and the long-reason test passes against the real column rather than against Ecto.
- **Scope.** `261ebd9c3..192243cf1` is one commit touching exactly `lease.ex`, the migration (`:string` → `:text`) and `lease_test.exs`. Nothing outside the review, no changelog fragment (still correct — nothing observable), `mix credo` clean, CI green (25 pass, 3 skipped).

PR #2348, opened 2026-09-16
feat(machines): the read-only owner process and one occupancy predicate (ADR 0058, stage 4)

BinaryBourbon:
Stage 4 of ADR 0058, "the machine has one owner" (#2344). Based on
`adr/0058-machine-owner` (#2342).

## What

The owner of a machine exists, can be found, and answers one question. It
writes nothing — not the row, not the provider, not an audit event.

**`Fountain.Machines.Occupancy`** — pure query functions, no process. One
struct for a sandbox, built from the conversation rows and the conversation
registry: `bound` (conversations not `terminated`/`failed`), `live` (those
with a registered `ConversationServer`), `running_turns`
(`%{conv_id => runtime}`), `last_activity_at`, and the per-conversation
activity the idle window reads.

**`Fountain.Machines.Machine`** — a GenServer registered in
`Fountain.MachineRegistry` under the sandbox id, supervised by
`Fountain.MachineSupervisor`, both Horde. `via/1`, `whereis/1`,
`ensure_started/2`, `who_is_here/1`, and an idle-stop after 60s with nothing
asked of it, so there is a process per *active* machine rather than per row.
It holds its sandbox id and that timer and nothing else — no lease (stage 3
is building `Machines.Lease` in parallel; this does not depend on it).

**`MACHINE_OWNER_ENABLED`**, default off and **reserved — it has no effect in
this release**. `config/runtime.exs` reads it into
`Fountain.Machines.enabled?/0` so stage 5 has its gate; nothing in `lib/` asks
the owner anything yet, so a `true` starts no process. `docs/configuration.md`
says exactly that, and `machine_test.exs` guards the claim.

## Why

#2255 decision 1 is "one predicate". Four unrelated ones answered "is anyone
on this machine", each with its own query: they are the compensation for the
owner ADR 0023 step 4 named and did not build. This stage builds the reading
they share and the process that will serve it, while serving it still changes
nothing — so the callers move before the writes do, which is the part of the
#1767 programme that went badly when it was done the other way round.

Asking a process for something a pure function already answers is deliberate.
`who_is_here/1` is the door `destroy`, `park`, `ensure_up` and `admit_turn`
come through in stages 5 to 8.

## The semantics each predicate keeps

The *reading* is collapsed; the *verdicts* are not. They disagree with each
other on purpose, and each difference now has a test that fails if it is
flattened.

| Predicate | Now | Semantics preserved |
|---|---|---|
| `Conversations._unsafe_sandbox_busy_elsewhere?/4` | `Occupancy.busy_elsewhere?/4`, by id | Status **and** the idle window, and **the same two short-circuiting `EXISTS` probes as before** — this is the conversation server's lifecycle tick, so it does not load the machine. The struct form of the same function folds the query's per-row disjunction into per-conversation maxima for `who_is_here/1`; both are pinned differentially against the pre-0058 query. |
| `Lifecycle._unsafe_sandbox_held_by_other?/2` | `Occupancy.bindings/1` → `held_by_other?/2` | Status only, no clock. One query, as before — its caller asks inside a transaction under the per-sandbox advisory lock, so it does not get the full load. |
| `Lifecycle.live_conversation_ids/1` | `Occupancy.from_preloaded/1` → `live_ids/1` | Registry liveness over the **preloaded** association, still running **no query** — it sits inside the reaper's per-row scans. |
| `Lifecycle.any_server_alive?/1` | `Occupancy.from_preloaded/1` → `any_live?/1` | The same scan, as a boolean. |
| `Conversations._unsafe_list_cotenant_ids/2` | `Occupancy.bindings/1` → `cotenant_ids/2` | The same set, same one query. Not one of the four, but it is the same reading, so it moved too. |

Two things worth calling out for the later stages:

- **`live` is not a subset of `bound`.** `live_conversation_ids/1` never
  filtered by status, and that is load-bearing in both directions — the
  reaper reads a registered server as "provisioning is still in flight
  somewhere in the cluster, do not touch this row" whatever the row says,
  while `Termination.reap_sandbox/1` terminates *through* a live server
  because that is what stopping a runaway agent means. Preserved, documented
  in the moduledoc, and pinned by a test. Whether the owner should keep the
  distinction is a stage 5/8 decision, not this one's.
- **Three constructors, not one.** A caller should not pay for an answer it
  did not ask for, and the reaper's scans in particular must stay
  query-free. A field a constructor did not fill is `:unloaded` and raises
  when read, rather than reading as "nobody is here".

`conversation_server.ex` is untouched; no delegation needed it.

## The flag

`MACHINE_OWNER_ENABLED=false` by default, and **inert either way in this
release**. Every predicate reads `Occupancy` directly, so no caller in `lib/`
asks the owner anything and turning the flag on starts zero processes. The
configuration row says so in those words, and there is no changelog fragment,
because a variable with no effect is not a behaviour anyone can observe.

It is not routed through the process on purpose.
`_unsafe_sandbox_held_by_other?/2` runs inside the teardown fence's
advisory-locked transaction and must keep reading that transaction's own
uncommitted rows on the caller's connection; serving it from a GenServer would
move the read to a different pooled connection and the fence would stop seeing
a co-tenant the same transaction just inserted. Stage 5 gives the owner its
first caller, on a path where that is safe.

Both Horde children are started unconditionally — an empty registry and an
empty dynamic supervisor cost nothing, and supervision that appears and
disappears with a runtime flag is its own failure mode. They are placed
*before* the conversation registry and supervisor, which is the draining end
again (#1726): reverse termination stops them *after* the conversation
supervisor, so a server shutting down can still ask its machine who is here.

`.env.example` is unchanged, matching `CHECKPOINT_CREATION_ENABLED` and the
neighbouring sandbox flags — `docs/configuration.md` is the reference and
`.env.example` is a subset (#2142).

## Review round 2

Every finding on https://github.com/managoat/fountain/pull/2348#pullrequestreview-5223184077
is addressed in `1e72467`. The blocking one was the flag's documentation, fixed
by changing the words rather than the routing, for the reason above. Also:
`who_is_here/1` no longer exits `{:noproc, _}` when the idle timer fired
between the lookup and the call (it catches, retries once, then reads
directly, so the `@spec` is true); `application_children_test.exs` names Horde
children by registered name, so "the listener starts before the conversation
supervisor" tests the conversation supervisor again rather than silently
following the machine one this stack inserted ahead of it, and the new pair
gets its own ordering test; "from_preloaded/1 runs no query at all" counts
`[:fountain, :repo, :query]` events instead of asserting nothing; the
"concurrent start" test is twenty tasks and one pid; `restart: :transient` is
spelled out in the child spec. The reviewer's differential and query-counting
probes are folded into `occupancy_test.exs` rather than discarded.

## Review round 3

`.credo.exs` was **not** actually reverted in round 2 — `git checkout --` cleaned
the working tree, but the three lines were already committed in `62d5970`, so
they stayed on the branch and the round-2 commit message asserted a revert that
had not happened. Restored from the base in `f7b5c00`;
`git diff origin/adr/0058-machine-owner..HEAD -- .credo.exs` is now empty, and
`mix credo --strict` reports no issues across 948 files without the exemption.

Also: `ask_owner/2` retries only `:noproc`/`:normal`/`:shutdown` rather than any
`:exit`, so a `:timeout` no longer retries into a second `@call_timeout`. That
change exposed an untested path — `who_is_here/1` re-looks-up on every call, so
an already-deregistered pid never reaches `GenServer.call`, and the idle-stop
test was exercising the ordinary restart. The real race is now driven
deterministically with a per-process registry stub that hands back a dead pid
once, and it discriminates on `count_children(...).active`, since the answer is
a struct whether the call retried or fell back.

## Checks run

(Re-run on `f7b5c00`.)

- `mix precommit` from the root: **PASSED, 9/9** (toolchain, conflict-markers,
  compile, deps, format, credo, dialyzer, sobelow, release).
- New: `occupancy_test.exs` + `machine_test.exs` + `application_children_test.exs`
  — **56 tests, 0 failures**. `mix credo --strict`: no issues, 948 files.
- The existing predicate suites, **unchanged**: `lifecycle_actions_test.exs`,
  `teardown_fence_test.exs`, `lifecycle_fence_test.exs`,
  `sandbox_reaper_test.exs`, `sandbox_reaper_predicates_test.exs`,
  `sandbox_mode_test.exs`, `conversations_read_state_test.exs`,
  `shared_sandbox_test.exs`, `attach_test.exs`,
  `application_children_test.exs`, `config_reference_test.exs` — 162 tests,
  0 failures.
- The wider blast radius: `test/fountain/conversations/`,
  `test/fountain/workers/`, `test/fountain/accounts/`, `config_reference_test`,
  `conversations_read_state_test` — **1800 tests, 0 failures**, 2 skipped.
- `scripts/test-docs.sh` — green.
- `python3 scripts/changelog.py check` — 11 fragments, 12 entries.

Verified by reverting, not just by passing:

- Deleting the `MACHINE_OWNER_ENABLED` row from `docs/configuration.md` fails
  `config_reference_test.exs` naming the variable. Restored from a copy.
- Making `live` a subset of `bound` fails the "a terminated conversation whose
  server is still registered stays live" test.
- Dropping the idle window from `busy_elsewhere?/4` fails four new tests **and**
  three existing `shared_sandbox_test.exs` ones. Restored from a copy.
- Pointing `_unsafe_sandbox_busy_elsewhere?/4` back at `load/1` fails the
  cost test (it sees the `sandboxes` read and the `GROUP BY`), and moving the
  machine pair after the conversation pair fails the new ordering test.
  Restored from copies.

## Not in this PR

No ratchet change (stage 2, in parallel), no lease (stage 3, in parallel), no
migration, no writes.

No changelog fragment, per the ADR's stage 4 row — the variable this stage
adds has no observable effect, so there is nothing a reader of the release
could notice. `.credo.exs` is untouched: `lib/fountain/machines/` makes no
`_unsafe_*` calls, so the ownership exemption buys nothing yet and stage 2
(#2350) carries that line when it is needed.

Refs #2344, ADR 0058, #2255 decision 1.


jhgaylor (changes requested):
The move is genuinely behaviour-preserving — I could not break it — but the published operator documentation describes a flag that does nothing, because nothing in `lib/` calls the process this PR adds.

I reviewed this as a behaviour-change hunt: I rebuilt the pre-PR predicates verbatim from `origin/adr/0058-machine-owner` and ran them against the new ones on the same fixtures, counted queries with a `[:fountain, :repo, :query]` telemetry handler, and drove the GenServer under real concurrency. Everything below was reproduced, not read.

---

## Blocking

### 1. The flag's docs and changelog claim behaviour the code does not have

`docs/configuration.md:67` tells an operator:

> A `true` starts one process per active sandbox, and each one answers the same question — who is on this machine — that Fountain answers without it. So the flag costs a process per machine and changes no behaviour.

`changelog.d/2344-machine-owner-flag.md:3` says the same:

> turned on, it starts one process per active sandbox that answers who is on that machine

Neither is true. `Machine.who_is_here/1` has **no caller anywhere in `lib/`** — the only references are `machine.ex` itself, two moduledocs, and `machine_test.exs`. All five moved predicates (`occupancy.ex:171`, `:179`, `:207`, `:232`, `:244`) are reached through `Occupancy.*` directly (`conversations.ex:1660`, `:1723`; `lifecycle.ex:583`, `:593`, `:618`), never through the process. So `MACHINE_OWNER_ENABLED=true` starts **zero** processes and costs nothing. The flag is inert.

Reproduction — with the gate on, exercise every predicate and look at the registry:

```elixir
Application.put_env(:fountain, :machine_owner_enabled, true)
assert Fountain.Machines.enabled?()

preloaded = Repo.preload(sandbox, :conversations, force: true)
assert Conversations._unsafe_sandbox_busy_elsewhere?(sandbox.id, a.id, 3600)
assert Lifecycle._unsafe_sandbox_held_by_other?(sandbox.id, a.id)
assert Lifecycle.live_conversation_ids(preloaded) == []
refute Lifecycle.any_server_alive?(preloaded)
assert Conversations._unsafe_list_cotenant_ids(sandbox.id, a.id) == [b.id]
Process.sleep(200)

assert Horde.Registry.select(Fountain.MachineRegistry, [{{:"$1", :"$2", :_}, [], [{{:"$1", :"$2"}}]}]) == []
assert Horde.DynamicSupervisor.count_children(Fountain.MachineSupervisor).active == 0
```

Both assertions pass. The registry stays empty and the supervisor stays childless with the gate on.

This matters for two reasons. `docs/` is the published manual, and CLAUDE.md's rule is "keep unbuilt behavior explicit in decisions and documentation" — an operator who turns this on to watch it, as the row invites, will see nothing and reasonably conclude the flag is broken. And stage 9's flip plan ("turn it on in the hosted overlay after every replica runs stage 8, and watch the deploy") is premised on the gate being exercisable; today there is nothing to watch.

I am **not** asking you to route the predicates through the process. `_unsafe_sandbox_held_by_other?/2` is called inside `do_fence_sandbox_for_teardown/2` (`lifecycle.ex:680`) under `pg_advisory_xact_lock`, and I verified it currently reads the caller's own uncommitted rows (see Verified, below). Serving it from a GenServer would move that read to a different pooled connection, outside the transaction, and the fence would stop seeing a co-tenant the same transaction just inserted. The right fix here is the two sentences, not the routing.

Suggested wording for both: say that the flag starts the supervision tree's registry and dynamic supervisor, that no caller asks the owner anything yet, and that it therefore starts no child and changes nothing until stage 5. The ADR's stage 4 row is honest about this ("The process, read-only … Behind the gate"); the docs row is the only place that overstates.

---

## Non-blocking

### 2. `who_is_here/1` can exit where its `@spec` and moduledoc promise a struct

`machine.ex:104` declares `:: Occupancy.t()` and `machine.ex:98-102` promises "Always an `Occupancy` struct … a machine with no process is not a reason to fail a caller that only wanted to look." The fallback at `machine.ex:111-117` only covers `{:error, reason}` from `ensure_started/1`. It does not cover the `GenServer.call` at `machine.ex:109` reaching a pid that has already gone — which is exactly what the idle timer (`machine.ex:142-146`) and a Horde CRDT entry that has not yet caught up both produce.

```
[repro] call on an idle-stopped owner:
  {:exit, {:noproc, {GenServer, :call, [#PID<0.562.0>, :who_is_here, 1000]}}}
```

(`ensure_started(id, idle_ms: 40)`, wait for `{:DOWN, …, :normal}`, then call the pid it returned.) Harmless today because of finding 1, and it becomes a crash in a conversation server the moment stage 5 wires a caller up. A `try do … catch :exit, _ -> Occupancy.load(state.sandbox_id) end` — or one `ensure_started/1` retry — closes it.

### 3. `busy_elsewhere?/4` got materially more expensive on the server's idle path

Counted with a telemetry handler. Before: 1 query for the co-tenant ids plus one or two `EXISTS` probes scoped to those ids, short-circuiting at the first hit. Now, always three:

```
[busy_elsewhere? NEW] result=true  3 queries:
  SELECT s0."id", s0."codex_inference_source", … 20 columns … FROM "sandboxes" AS s0 WHERE (s0."id" = $1)
  SELECT c0."id", c0."status", c0."runtime", c0."updated_at" FROM "conversations" AS c0 WHERE (c0."sandbox_id" = $1) ORDER BY …
  SELECT t0."conversation_id", max(…), max(…), max(…), bool_or(t0."status" = 'running')
    FROM "turns" AS t0 WHERE (t0."conversation_id" = ANY($1)) GROUP BY t0."conversation_id"
```

Two things there are pure waste for this caller. The full `sandboxes` row is fetched only to feed `last_activity_at` (`occupancy.ex:155`), which `busy_elsewhere?/4` never reads. And the rollup is an unbounded `GROUP BY` over every turn of every conversation on the machine — including the terminated ones and the requesting conversation itself (`occupancy.ex:145-147` passes all `ids`, not `bound`, not the co-tenants) — where the old `EXISTS` stopped at the first matching row among the co-tenants only.

The caller is `Lifecycle.busy_elsewhere?/2` (`lifecycle.ex:314`), on the conversation server's lifecycle tick. The ADR's own context says the busiest production home carries 36 conversations; that is 36 turn histories aggregated per tick per server on that machine, against one short-circuiting `EXISTS` before. Worth either a fourth constructor that skips the sandbox row and scopes the rollup to `bound`, or a note on the issue that stage 6 will fix it.

### 4. This PR silently changed what an existing ordering assertion tests

`application_children_test.exs:121-125` reduces a child spec to its module: `id({module, _opts}), do: module` (`:124`). Both Horde pairs therefore collapse to `Horde.Registry` / `Horde.DynamicSupervisor`, and `before?/3` (`:127`) uses `Enum.find_index`, which takes the first match. Inserting `MachineSupervisor` ahead of `ConversationSupervisor` (`application.ex:128-135`) moves that first match, so:

```elixir
test "the listener starts before the conversation supervisor, so it outlives it" do
  assert before?(ids, Managoat.Broker, Horde.DynamicSupervisor)   # :46
end
```

now asserts it about the *machine* supervisor. It still passes, and the property it names still holds, but the test no longer checks it. One line fixes it: `defp id({Horde.Registry, opts}), do: Keyword.fetch!(opts, :name)` and the same for `Horde.DynamicSupervisor`, matching the existing `DynamicSupervisor` clause at `:123`. Doing it here also gives the new children real coverage rather than none.

### 5. The `.credo.exs` reason describes neither this branch nor these callers

`.credo.exs:99-101`:

```
# the machine owner (ADR 0058): ownership is the sandbox row's
# lease, established at claim; called only from its own process
"lib/fountain/machines/",
```

There is no lease on this branch (stage 3 is unmerged), and `Occupancy` is called from `Conversations` and `Lifecycle` in whatever process they run in — including inside the teardown fence's transaction — not "only from its own process". More to the point the exemption is not needed: `lib/fountain/machines/` makes zero remote `_unsafe_*` calls. The four `_unsafe_` strings in `occupancy.ex` (`:7`, `:8`, `:165`, `:188`) are all inside `@moduledoc`/`@doc`, and `Fountain.Credo.Check.UnsafeCallOwnership` is an AST prewalk, so they produce no call nodes. Adding the path now turns the ownership guard off for the namespace before any of the justification is true, and stage 5's first `_unsafe_` call will land uncommented and unnoticed. I would either drop the entry until stage 5 needs it, or state the reason in the future tense the code is actually in.

### 6. "from_preloaded/1 runs no query at all" does not test that

`occupancy_test.exs:256-266` asserts `occ.activity == :unloaded` and `occ.running_turns == :unloaded`. Those hold whether or not a query ran; the test body cannot fail on the property in its name, and its own comment concedes it ("Nothing to assert about SQL from here"). The property is real and important — it is what keeps `Lifecycle.any_server_alive?/1` usable inside three reaper scans (`sandbox_reaper.ex:136`, `:237`, `:407`) — and it is directly assertable:

```elixir
:telemetry.attach(ref, [:fountain, :repo, :query], fn _, _, meta, _ -> send(me, {:q, meta.query}) end, nil)
Occupancy.from_preloaded(sandbox)
refute_received {:q, _}
```

I ran that; it is zero queries for `from_preloaded/1`, `live_conversation_ids/1` and `any_server_alive?/1`. Please fold the counter in so the next refactor cannot quietly add one.

### 7. Smaller things

- `machine_test.exs:76` — "a concurrent start that loses the registry race" is two sequential calls; nothing here is concurrent. My 20-task version passes (see Verified), so the code is right; the test name is not.
- `application.ex:116` says "first of the three pairs". There are two pairs and a lone `RunnerRegistry` (`:159`).
- A killed owner comes back: `restart: :transient` restarts on an abnormal exit, and Horde does. After `Process.exit(pid, :kill)`, `whereis/1` returns a new pid ~300 ms later, which then holds a registry slot for its full 60 s idle window even though nobody asked. Benign while it holds nothing, but "one process per *active* machine" (`machine.ex:5-9`) is slightly stronger than what the tree does.
- `who_is_here/1` on a sandbox id that does not exist starts an owner for a row that is not there, answers empty, and lingers 60 s. No crash loop (confirmed), and unreachable today, but a `nil` sandbox is worth an early return before stage 5 gives the process something to hold.

---

## Verified

**`busy_elsewhere?/4` is semantically identical.** I copied `_unsafe_sandbox_busy_elsewhere?/4` verbatim out of `git show origin/adr/0058-machine-owner:apps/fountain/lib/fountain/conversations.ex` into a test-local `old/4` and asserted `old == new` on the same fixtures across 18 cases: an old turn with a recent `updated_at` on the same co-tenant; `ended_at` nil on a non-running turn, inside and outside the window; `inserted_at` exactly `== cutoff` and at `cutoff + 1s`; the cross-row disjunction (turn A old-inserted/recent-ended, turn B recent-inserted/old-ended, both ways); a terminated and a failed co-tenant carrying a `running` turn row; the requesting conversation excluded when it is the only active one; `idle_seconds` nil (no query) and `0`; no co-tenants; `updated_at` exactly at cutoff for a co-tenant with no turns; three co-tenants with only the third busy; turns on a conversation bound to a different sandbox; all-nil turn stamps. **18/18, zero divergence.** `∃r (P ∨ Q ∨ R) ≡ (∃r P) ∨ (∃r Q) ∨ (∃r R)` holds, `max(x) > cutoff ≡ ∃ x > cutoff` with NULLs dropped by `max`, and every timestamp involved is `:utc_datetime` on both sides (`turn.ex:25`, `:26`, `:73`; `conversation.ex:110`), so the `DateTime.truncate(:second)` cutoff has no precision edge to fall off.

**`held_by_other?/2` is identical** across all five conversation statuses, differentially against the old `Repo.exists?`, and it is still safe inside the fence. Probe: inside `Repo.transaction` + `pg_advisory_xact_lock`, insert a co-tenant, terminate the two committed ones, then call it — it returns `true`, i.e. it reads the transaction's own uncommitted row on the caller's connection. One query, no nested transaction, no second checkout.

**Query counts** (telemetry): `from_preloaded/1` 0, `live_conversation_ids/1` 0, `any_server_alive?/1` 0, `bindings/1` 1, `_unsafe_sandbox_held_by_other?/2` 1, `_unsafe_list_cotenant_ids/2` 1, `load/1` 3. No N+1 introduced into a reaper loop; the three reaper scans keep the zero-query path. `ConversationServer.whereis/1` is still the registry door (`occupancy.ex:257`).

**`ensure_started/1` under concurrency**: 20 tasks at `max_concurrency: 20` on one sandbox id — 20 `{:ok, pid}`, one distinct pid, `count_children(…).active == 1`. Both `{:error, {:already_started, pid}}` branches and the `await_registered/1` poll behave.

**Gate off starts nothing**: `who_is_here/1` with the flag off leaves `Horde.Registry.select(Fountain.MachineRegistry, …) == []`.

**No writes**: `grep` for `update_sandbox`, `Managoat.Sandbox.`, `Repo.update`, `Repo.insert`, `Repo.delete` and `audit` over `lib/fountain/machines/` returns nothing but one moduledoc line. The ratchet is untouched (and `direct_writes_test.exs` is not on this base yet).

**Existing tests unchanged**: `git diff origin/adr/0058-machine-owner..62d5970 --stat -- apps/fountain/test` is two new files, 446 insertions, zero deletions.

**Suites run** on the PR head with a private DB: `machine_test` + `occupancy_test` + `config_reference_test` + `application_children_test` + `sandbox_reaper_test` + `sandbox_reaper_predicates_test` + `forced_home_fence_test` + `shared_sandbox_test` + `conversation_reapply_test` → **136 tests, 0 failures**. `lifecycle_test` + `lifecycle_fence_test` + `lifecycle_actions_test` + `bounded_lifecycle_test` + `conversation_server_shared_sandbox_test` + `termination_fence_policy_test` + `termination_cluster_test` + `termination_binding_isolation_test` + `reapply_check_test` → **110 tests, 0 failures**. `python3 scripts/changelog.py check` → 12 fragments, 13 entries, exit 0.

**Supervision order**: `Repo` (`application.ex:76`) and `PubSub` (`:89`) before the new pair, `FountainWeb.Endpoint` (`:162`) after, and the reverse-termination comment at `:116-118` is truthful — starting first means stopping last, after the conversation supervisor.

**Scope**: the extra `_unsafe_list_cotenant_ids/2` move is safe. Same one query, same predicate; it gains an `ORDER BY c0."inserted_at", c0."id"` where the base had none, which can only make a previously arbitrary order deterministic. Its one caller, `Lifecycle.stop_cotenants/5` (`:558`), fans out to `MachineEvents.tell_cotenants/5` and does not care about order.

---

Fix the docs row and the fragment and this is a clean stage-4 move — the differential work is the most convincing part of the PR and it held up under everything I threw at it.

jhgaylor (changes requested):
Re-reviewed at `1e7246704`. Every finding from the last round is genuinely fixed and I verified each one — except one, which the commit message says was done and was not.

## Blocking

### `.credo.exs` was not reverted

The commit message states:

> `.credo.exs` reverted entirely. `lib/fountain/machines/` makes no `_unsafe_*` calls, so the exemption bought nothing, and its stated reason described a lease that does not exist on this branch. Stage 2 (#2350) carries that line. This also drops the expected conflict with it.

The branch says otherwise. At the exact head under review, freshly fetched:

```
$ git rev-parse origin/stack/0058-4-machine
1e72467040fd7c6f46cb03502e1a324be6d5889a

$ git show --stat 1e7246704 -- .credo.exs
(no output — the fix commit does not touch it)

$ git diff origin/adr/0058-machine-owner..1e7246704 -- .credo.exs
+               # the machine owner (ADR 0058): ownership is the sandbox row's
+               # lease, established at claim; called only from its own process
+               "lib/fountain/machines/",
```

Byte-identical to what I flagged. So three statements in the record are false: the file is not reverted, #2350 is not the only carrier of that line, and the conflict with #2350 is not dropped — it is still there, and `.credo.exs:99-101` still ships an ownership-guard exemption justified by a lease that does not exist on this branch.

This is small and it was non-blocking on its own merits last round. It is blocking now only because the commit message asserts it was done, and on a squash merge that message is what the history says. `git checkout origin/adr/0058-machine-owner -- .credo.exs` and I will approve on sight.

## Non-blocking — three notes for stage 5, nothing to change here

**`Fountain.MachineSupervisor` takes Horde's default restart budget.** `application.ex:130-136` sets no `max_restarts`; `Horde.DynamicSupervisor` defaults to `max_restarts: 3` (`deps/horde/lib/horde/dynamic_supervisor.ex:147`) over 5 seconds. Twelve lines further down, `Fountain.ConversationSupervisor` overrides it to 100/10s with a comment explaining exactly why: "the default (3 restarts in 5s) is a budget SHARED by every ConversationServer on the node — one child that crashes deterministically on start exhausts it in under a second, and exceeding it terminates this supervisor and with it every running conversation here." That reasoning transfers verbatim to the machine supervisor from stage 5, when there is one owner per active machine on a node and a correlated provider or repo blip can crash several at once. Harmless today — the owner holds nothing and has no caller — but the asymmetry should be a decision rather than an omission. Note also that `ask_owner/2` spends two restarts per call on a deterministically-crashing owner (call → crash → retry → crash → fallback), so two calls inside five seconds would exhaust the default budget.

**The telemetry filter is narrower than its comment says.** `occupancy_test.exs:76-92` counts a query only when `self() == me` inside the handler, and the comment calls that "exact rather than approximate". It is exact for what is pinned today, because every one of those paths queries synchronously on the caller's connection — and for `held_by_other?/2` it *must*, since its caller is inside the teardown fence's advisory-locked transaction. But a query issued from a child process does not reach the count: a future N+1 written as `Task.async_stream` would read as zero and the assertion would not fire. `Process.get(:"$callers")` is the usual widening if you want it to hold under a refactor. The comment is the part I would change, not the code.

**`ask_owner/2` retries on a `:timeout` exit too.** `machine.ex:121-137` catches `:exit, reason` without discriminating, so a `@call_timeout` (15s) expiry retries into another 15s before falling back to `Occupancy.load/1`. A repo already in trouble gets 30 seconds of patience where the comment at `machine.ex:46-47` argues for exactly the opposite ("a timeout longer than the default would only ever hide a repo that is already in trouble"). `:noproc` and `:normal` are the exits worth retrying.

## Verified

**The docs row and the tree comment are now true.** `ensure_started/1` and `who_is_here/1` have zero callers outside `lib/fountain/machines/` — `grep -rn "ensure_started\|who_is_here" apps/*/lib ee/` returns nothing else. `docs/configuration.md:67` ("Reserved. It has no effect in this release… the owner process exists in this release, and nothing starts one") and `application.ex:116-126` ("**As of stage 4 nothing asks**… this pair is up and childless on every node, whatever `MACHINE_OWNER_ENABLED` says") both match the code. The changelog fragment is gone and `changelog.py check` reads 11 fragments, 12 entries.

**The by-id path is the original query, at the original cost.** Telemetry counts on the new head:

| path | queries |
|---|---|
| `_unsafe_sandbox_busy_elsewhere?/4`, busy | **2** — co-tenants, then `SELECT TRUE FROM turns … LIMIT 1` |
| same, not busy (worst case) | **3** — plus the `LEFT JOIN … t1."id" IS NULL … LIMIT 1` |
| `busy_elsewhere?/4` from a loaded struct | **0** |
| `from_preloaded/1`, `live_conversation_ids/1`, `any_server_alive?/1` | **0** |
| `bindings/1`, `held_by_other?/2`, `_unsafe_list_cotenant_ids/2` | **1** |
| `load/1` | 3, and now only `who_is_here/1` pays it |

Both `EXISTS` probes come back with `LIMIT 1` and the co-tenant scope, and the 20-column `sandboxes` row is off the lifecycle tick. That is the pre-PR cost exactly, not an approximation of it.

**The differential holds for both forms.** I re-ran my probe against the new head with `old/4` copied verbatim from `origin/adr/0058-machine-owner` and each case asserted four ways — `old`, `Occupancy.busy_elsewhere?/4` by id, the struct form, and the public `_unsafe_sandbox_busy_elsewhere?/4`. 18 cases: old turn with recent `updated_at`; `ended_at` nil inside and outside the window; `inserted_at` exactly `== cutoff` and at `cutoff + 1s`; the disjunction split across two turn rows both ways; terminated and failed co-tenants holding `running` rows; self-exclusion; `nil` and `0` idle seconds; no co-tenants; `updated_at` exactly at cutoff with no turns; three co-tenants with only the third busy; turns on another sandbox; all-nil stamps. **Zero divergence, three ways, 18/18.** The version folded into `occupancy_test.exs:391-582` is the real thing — `old/4` is verbatim, `agree/4` checks both forms and returns the verdict so each test pins which way it went, and the added cases cover the edges I would have asked for.

**The other adoptions check out.** `machine_test.exs:189-211` is my flag probe, asserting an empty `MachineRegistry` and `count_children(…).active == 0` after all five predicates run with the gate on — and it guards the docs wording, which is the right place for it. `machine_test.exs:82-101` is twenty real concurrent tasks; `:172-187` covers the idle-stopped pid and asserts `who_is_here/1` still answers. `application_children_test.exs:144-145` names Horde children by `opts[:name]`, `:46` now asserts `Fountain.ConversationSupervisor` by name, and `:105-116` pins all four new ordering facts. `occupancy_test.exs:296-320` counts real query events for `from_preloaded/1` and both reaper callers.

**`ask_owner/2` is bounded and cannot start two owners.** `machine.ex:121-147`: `ask_owner(id, 1)` → exit with `retries_left > 0` → `ask_owner(id, 0)` → exit with the guard false → `Occupancy.load/1`. Two clauses, no third recursion, so at most two `ensure_started/1` calls and two `GenServer.call`s. The retry is sequential and only runs after the first pid is gone, so it cannot produce two live owners; `keys: :unique` and the `{:error, {:already_started, pid}}` branch cover the concurrent case, which I re-confirmed with 20 tasks → one pid, `active == 1`. The fallback is not silent: `Logger.debug` on the retry, `Logger.warning` on the fallback.

**Nothing outside scope.** `62d597045..1e7246704` is nine files, all named in the last review: the two lib modules, `application.ex`, `conversations.ex`, three test files, the docs row, and the deleted fragment. No migration, no schema, no new dependency, no writes — `grep` for `update_sandbox`, `Managoat.Sandbox.`, `Repo.update/insert/delete` and `audit` over `lib/fountain/machines/` still returns nothing but a moduledoc line.

**Suites**, on `1e7246704` with a private DB: `machine_test` + `occupancy_test` + `config_reference_test` + `application_children_test` + `sandbox_reaper_test` + `sandbox_reaper_predicates_test` + `forced_home_fence_test` + `shared_sandbox_test` + `conversation_reapply_test` → **152 tests, 0 failures** (up from 136). `lifecycle_test` + `lifecycle_fence_test` + `lifecycle_actions_test` + `bounded_lifecycle_test` + `conversation_server_shared_sandbox_test` + `termination_fence_policy_test` + `termination_cluster_test` + `termination_binding_isolation_test` + `reapply_check_test` → **110 tests, 0 failures**. My four probes → **27 tests, 0 failures**.

---

This is a good response to the review — the by-id restoration in particular is a better fix than the one I suggested, and the folded differential is stronger than my probe was. Revert `.credo.exs` so the commit message is true and this has my approval.

jhgaylor (approved):
Approving at `f7b5c008e`. The one blocker is cleared and both optional notes were taken properly.

## Verified

**`.credo.exs` is actually reverted.** `git diff origin/adr/0058-machine-owner..f7b5c008e -- .credo.exs` is empty, and the commit touches exactly four files: `.credo.exs` (−3), `machine.ex`, `machine_test.exs`, `occupancy_test.exs`. The commit message is straight about how the first claim went wrong — "a clean working tree is not a reverted commit" — which is the right diagnosis and the more useful record than a silent fix.

**`ask_owner/2` discriminates.** `machine.ex:127-129` now guards on `elem(reason, 0) in [:noproc, :normal, :shutdown]`, so a `:timeout` falls straight to the direct read instead of spending a second `@call_timeout`. Two consequences beyond the one I raised, both improvements: a `:killed` owner is no longer retried either, and an owner that crashes inside `handle_call` now burns one restart against `Fountain.MachineSupervisor`'s budget rather than two — which takes most of the weight off my stage-5 note about that supervisor keeping Horde's default `max_restarts: 3`. Still worth a decision when the owner holds a lease; no longer worth much now. The guard is also safe on a non-tuple `reason`: `elem/2` failing in a guard falls through to the next clause, which is the fallback.

**The new test reaches the race the old one could not, and discriminates.** `machine_test.exs:190-221` stubs `Horde.Registry.lookup/2` per-process to hand back an already-dead pid once and `[]` after, so `ensure_started/2` returns a pid that exits on `GenServer.call` — the "handed back and dead before the call lands" window, which the idle-stop test could never hit because `who_is_here/1` looks up again on every call. The discriminator is the right one: the answer is an `Occupancy` struct whether the call retried or fell back, and only the retry starts a real owner, so `count_children(…).active == 1` is what separates them. Dropping `:noproc` from the guard makes it read 0 and fail. Mimic stubs are per-process, so Horde's own internals are untouched, and `Mimic.copy(Horde.Registry)` is already in `test_helper.exs:45` with existing precedent — no new wiring. Renaming the old test to "an idle-stopped owner is simply replaced on the next question" is accurate about what it was always testing.

**The telemetry comment says what the filter does.** `occupancy_test.exs:76-86` now states that it sees a synchronous query on the caller's connection, that a query from a child process would go uncounted, and that `$callers` is the widening if one appears. That is exactly the caveat, written where the next person will read it.

Everything from the two earlier rounds still holds at this SHA: the by-id predicate is the original query at the original cost (2 queries busy, 3 not busy, 0 from a loaded struct), the zero-query reaper path is counted rather than asserted by proxy, the docs row and the supervision comment match a flag that starts nothing, and the differential harness pins both forms against the pre-0058 query.

## One housekeeping note, not a change request

The branch is two commits behind its base: `origin/adr/0058-machine-owner` is at `95537e802` (`docs(decisions): ADR 0058 counts claim_sandbox/2 beside update_sandbox/2` and `ADR 0058 review corrections`), and `f7b5c008e` does not contain them, so a two-dot diff against the base shows those ADR edits in reverse. Docs-only on the parent's own ADR, so it changes nothing here — but rebase before this lands, and retarget before the parent merges rather than after.

Nice work on the response. The by-id restoration and the deterministic retry test are both better than what I asked for.

PR #2349, opened 2026-09-16
fix(deployed): keep a failed run's evidence out of its own redactor

jhgaylor:
Found while running the deployed suite's `secrets` profile against production.
The run printed five `PASS` lines and then reported `0 passed, 2 failed`,
because it had corrupted its own report.

## The defect

`Redactor.value` guesses at credentials from key names and, on a match,
replaces the value and feeds every string beneath it to `collect`. That is the
right instinct for a response body an instance sends. It is the wrong one for
the report the suite composes, whose findings live under `secrets`.

Two things followed. The block that explains the failure became the literal
string `"[REDACTED]"`. And a leak record — `{ path, transport: "sse" }` —
promoted `"sse"` to a redaction token.

A run persists the report repeatedly through the same redactor (`persist` per
check, again on cleanup, again at the end), so the first pass learned `"sse"`
and every later pass applied it:

```
'pa[REDACTED]d' setup/identity
'pa[REDACTED]d' secrets/setup
'failed'        secrets/sandbox-delivery-and-output
```

`result.json` carried check statuses outside their own enum, `junit.xml`
counted failures by comparing those statuses to `'failed'`, and the
diagnostic read `Synthetic secret disclosed in public [REDACTED] response at
[REDACTED]`. The profile whose evidence matters most was the only one that
destroyed it.

## The fix

The report goes through a new `Redactor.strings`, which removes registered
secrets from every string and leaves the structure alone. The key-name
heuristic stays exactly where it belongs, on data an instance sends.

This does not weaken the report. Nothing reaches it unregistered: values the
suite generates are added explicitly, and anything read from the instance
already passed through `value` on the way in, which both masked and collected
it.

## Verification

- `node --test deployed/test/*.test.mjs` — 247 pass, up from 242.
- The new tests were checked by reverting the one-line `runner.mjs` change:
  `repeated persists of the same report stay identical` fails with
  `status: 'pa[REDACTED]d'`, which is the production symptom exactly. A
  single-pass test does **not** catch it — the corruption needs the second
  persist — so the test drives `writeReport` twice through one redactor.
- A separate test pins that the key-name heuristic still masks
  `authorization` and `api_key` in traced responses and still learns those
  values for later prose.

Independent of #2345 and of the tunnelled-receiver work; it touches only
`deployed/lib/http.mjs`, one line of `deployed/lib/runner.mjs` and tests.


BinaryBourbon (changes requested):

Reviewed all four changed files and final head `516c1db72dd7837d4024fe83033993f2cae5ac7d` against base `11e98f5bb81055bc800e103ee44d759544dbaadf`, tracing report inputs through HTTP/SSE, receiver evidence, browser journeys, deployment adapters, matrix aggregation, and cleanup.

**[P2] Validate the OpenAPI version projection before using string-only report redaction.** The change at `deployed/lib/runner.mjs:164` assumes report values were registered at ingress, but `deployed/profiles/basic.mjs:75–78` fetches `/api/openapi.json` with `recordBody: false`, which skips `Redactor.value`, and then copies `body.info.version` into the report without a type check. It does this before checking the advertised document. A malformed external deployment returning `info.version: {api_key: "sk_fake_unregistered_provider_secret_bfa18b3c"}` therefore leaves that unknown, non-`ftn_` credential in the final `result.json` on this head. The base masks it. A failed conformance verdict does not prevent artifact disclosure.

I reproduced this with the actual `run` → `basic` → `Client` → `writeReport` path for both exact revisions against a local HTTP fixture: both failed the advertised-contract check, neither HTTP trace contained the document body, but only head retained the fake secret in `result.json`. The fixture refused all resource writes. This concerns malformed external responses; the shipped server's normal static version string does not trigger it.

Keep the report-evidence fix, and validate/project `info.version` to the expected string/null shape before assigning it; reject a structured value without persisting it. Add a regression using the actual unrecorded-response path and assert the malformed value never reaches artifacts. Restoring the blanket report key heuristic would reintroduce the status corruption this PR correctly addresses.

The new tests cover evidence preservation, registered secret values, continued response heuristics, and repeated persistence. The coordinating reviewer reports 247/247 exact-head Node tests plus changelog/whitespace checks passed; reverting the report call to `value` makes the new persistence regression fail as intended. Those tests do not cover the ingress exception above.

Source review and the focused reproduction are complete. Only fake credentials and a loopback fixture were used; no real deployment, account, receiver, browser session, or Kubernetes cluster was accessed. This is an independent Codex review, not a Review Loop App verdict.

BinaryBourbon (changes requested):
Independent adversarial correctness, API, and evidence compatibility review by Codex.

**Verdict: REQUEST_CHANGES** at `516c1db72dd7837d4024fe83033993f2cae5ac7d`.

- **P2 — Protect the remaining raw OpenAPI response value before switching report serialization.** The change in `deployed/lib/runner.mjs:164` assumes report values have already crossed a redaction boundary. However, `deployed/profiles/basic.mjs:75–79` fetches `/api/openapi.json` with `recordBody: false`, copies the unrestricted `body.info.version` into `report.advertised_schema.version`, and only then verifies the document. A faulty target returning `info.version: {api_key: "sk_fake_unregistered_provider_secret_bfa18b3c"}` therefore leaves that secret-shaped response field in `result.json` after the check fails. An exact base/head full-run probe independently reproduced this: both runs exit 1 and omit the document from HTTP traces; only the new report contains the fake secret. This regresses the documented artifact redaction guarantee for malformed responses. Keep the report-specific string walk, but validate/project this external value to the intended version scalar before assigning it, and cover the actual response-to-report path with a regression test.

Reviewed all four changed paths: the redactor helper, report writer, five regression tests, and changelog. Traced repeated per-check/final persistence, JSON/JUnit status consumers, matrix summaries, profile report producers, explicit credential registration, and incoming HTTP/SSE redaction. The helper fixes incidental evidence strings becoming global redaction tokens, and wire/API schemas and SDK contracts are unchanged. No additional actionable defect found.

Validation: inspected the exact-head 247/247 passing Node test log and mutation result showing that restoring the old report serializer fails the repeated-persistence test. Independently reran the malformed OpenAPI probe using only loopback HTTP and fake credentials. Required CI and secret scanning passed according to the coordinated exact-head verification. No real deployment or credential was used.

jhgaylor:
Addressed in `2826a56b1`. Both reviews found the same defect and it was real.

The hole was exactly as described: dropping the key-name heuristic from the
report assumed every value in it had crossed a redaction boundary on the way
in, and `/api/openapi.json` had not. It is fetched with `recordBody: false`,
so the body never passes `Redactor.value`, and `basic.mjs` copied
`body.info?.version` into `advertised_schema` unchecked. The old blanket
heuristic masked a structured value by accident, because the nested key
matched; the string walk does not, because it is not guessing.

`info.version` is now projected through `advertisedVersion`, which keeps a
string and drops anything else. A version is a scalar; a structured one is a
malformed instance handing us an unregistered value to persist.

The regression drives the real response-to-report path rather than the helper:
a loopback instance serving `info.version: {api_key: "sk_fake_…"}`, the real
`run`, the real `basic` profile, then an assertion that the planted value
appears in **no** retained artifact — not merely that the field is null. A
second case pins that a normal version string is still recorded. Restoring
`body.info?.version ?? null` fails the first, which I checked.

To your closing point: the report key heuristic is not restored, so the status
corruption this PR fixes stays fixed.

Checks: 250 Node tests pass, up from 247. Changelog check passes, and the
fragment now covers this second change.

BinaryBourbon (approved):
Independent adversarial correctness, security, and API compatibility re-review by Codex.

**Verdict: APPROVE** at `2826a56b19c3471021fb11839ddafc1b29db21ef`.

The previously reported OpenAPI-to-artifact disclosure is resolved. `basic` now projects `info.version` to a string or `null` before report assignment, so a structured value from the unrecorded response cannot retain secret-shaped fields. Normal version strings remain available, and advertised-contract checking still receives the original document.

Reviewed the entire PR against `11e98f5bb81055bc800e103ee44d759544dbaadf`, including all six changed paths:

- `deployed/lib/http.mjs`: report strings are scrubbed without learning incidental evidence as credentials; incoming HTTP/SSE heuristics remain intact.
- `deployed/lib/runner.mjs`: repeated persistence preserves structured evidence and the statuses used by JSON/JUnit consumers.
- `deployed/profiles/basic.mjs`: the previously unsafe response projection is constrained before persistence.
- `deployed/test/advertised-version.test.mjs`: actual runner/HTTP coverage rejects structured version artifacts and preserves ordinary versions.
- `deployed/test/report-redaction.test.mjs`: evidence retention, credential removal, repeated writes, and JUnit counts are covered.
- `changelog.d/2348-deployed-report-evidence.md`: describes both repairs accurately.

Also traced profile report producers, omitted-response-body callers, credential registration, cleanup/final writes, and matrix consumption. No remaining actionable correctness, credential-boundary, or API compatibility defect found; server and SDK wire contracts are unchanged.

Validation: independently ran all eight focused regression tests successfully and inspected the coordinated exact-head full-suite log showing 250/250 passing tests. The mutation log confirms that removing only the version projection makes the real-path regression test fail with the planted secret-bearing object. Required CI and secret scanning passed at this exact head according to coordinated verification. Tests used fake credentials, temporary artifacts, and loopback HTTP only; no real deployment was exercised.

PR #2350, opened 2026-09-16
test(machines): ratchet direct sandbox writes (ADR 0058, stage 2)

jhgaylor:
## What

Adds `apps/fountain/test/fountain/machines/direct_writes_test.exs`, a ratchet
test that pins the number of direct writes outside the future machine
owner's namespace (`lib/fountain/machines/`), in two groups: direct writes to
the `sandboxes` row (`update_sandbox/2`, `update_sandbox_row/2`,
`claim_sandbox/2`) and direct provider mutations
(`Managoat.Sandbox.create/2`, `resume/1`, `suspend/1`, `destroy/1`,
`create_checkpoint/1`). It is a pure text scan, `async: true`, no database —
same convention as `conversation_server_size_test.exs`. Also exempts
`lib/fountain/machines/` from `UnsafeCallOwnership` in `.credo.exs`.

## Why

ADR 0058 ("The machine has one owner") gives a sandbox one owner,
`Fountain.Machines.Machine`, that becomes the only writer of the row and the
only caller of the provider's create/resume/suspend/destroy/checkpoint.
Stage 2 of the tracker (#2344) is the ratchet: a test that can only go down,
so each later stage that moves a call site behind the owner lowers the pin
in the same PR, and nothing can add a new direct write without raising it
and saying why.

## Measured numbers

**Row writes: 28** (pin `@row_writes`)

| Count | File |
|---|---|
| 10 | `apps/fountain/lib/fountain/conversations/conversation_server.ex` |
| 4 | `apps/fountain/lib/fountain/workers/sandbox_reaper.ex` |
| 3 | `apps/fountain/lib/fountain/conversations/termination.ex` |
| 2 | `apps/fountain/lib/fountain/conversations/lifecycle.ex` |
| 2 | `apps/fountain/lib/fountain/conversations/reapply.ex` |
| 2 | `apps/fountain/lib/fountain/conversations/wake.ex` |
| 1 | `apps/fountain/lib/fountain/accounts/deletion.ex` |
| 1 | `apps/fountain/lib/fountain/conversations.ex` |
| 1 | `apps/fountain/lib/fountain/conversations/home_checkpoint.ex` |
| 1 | `apps/fountain/lib/fountain/conversations/provision_watchdog.ex` |
| 1 | `apps/fountain/lib/fountain/conversations/provisioning.ex` |

**Provider mutations: 17** (pin `@provider_mutations`)

| Count | File |
|---|---|
| 4 | `apps/fountain/lib/fountain/conversations/conversation_server.ex` |
| 3 | `apps/fountain/lib/fountain/conversations/provisioning.ex` |
| 2 | `apps/fountain/lib/fountain/conversations.ex` |
| 2 | `apps/fountain/lib/fountain/conversations/lifecycle.ex` |
| 2 | `apps/fountain/lib/fountain/workers/sandbox_reaper.ex` |
| 1 | `apps/fountain/lib/fountain/accounts/deletion.ex` |
| 1 | `apps/fountain/lib/fountain/conversations/home_checkpoint.ex` |
| 1 | `apps/fountain/lib/fountain/conversations/termination.ex` |
| 1 | `apps/fountain/lib/fountain/conversations/wake.ex` |

These differ from the ADR's rough-grep "24 + 19" (recorded against
`c3f568596`). The regex's exact scope, documented in the test's moduledoc:

- Counts call sites only. `def`/`defp`/`@spec` lines that define
  `update_sandbox`, `update_sandbox_row` and `claim_sandbox` are excluded; a
  call to one of them from inside `Fountain.Conversations` itself (e.g.
  `claim_sandbox/2` calling `update_sandbox/2`) still counts.
- Word-bounded, so `reclaim_sandbox(` (an unrelated local function on
  `ConversationServer`) does not match `claim_sandbox(`.
- `@moduledoc`/`@doc` heredocs and whole-line `#` comments are stripped
  first, so a call mentioned in prose (`SandboxReaper`'s moduledoc names one)
  is not counted. Inline trailing comments are not stripped — none of the
  matched patterns occur after one today, and doing that correctly around
  string interpolation (`#{}`) is not simple.
- Resolves the one alias pattern in use: a file with a bare
  `alias Managoat.Sandbox` line also counts bare `Sandbox.create(` etc. as
  the qualified call it is — `Provisioning` does this once, for
  `create_checkpoint/1`. It does not resolve `as:` aliases; none exist today.
- `update_sandbox_row/2` does not exist in the codebase yet (0 hits); it is
  in the pattern list for when stage 3 or later introduces it.

The row-write count (28) is higher than the ADR's rough 24 mainly because
`conversation_server.ex` has 10 real call sites by this method, not 7 — the
rough grep likely undercounted duplicated `{:ok, _} = Conversations.update_sandbox(...)` failure-path writes. The provider-mutation count (17) is
lower than the ADR's rough 19 mainly because `execution_deadline_worker.ex`
calls `Managoat.Sandbox.terminate_session/3`, which is not one of the five
pinned verbs, so it does not count here.

## Verified by reverting

Added one probe direct write, `Fountain.Conversations.update_sandbox/2`
called from a throwaway private function in `apps/fountain/lib/fountain/quotas.ex`
(a file with no prior sandbox writes). The test failed:

```
direct sandbox row writes are 29, over the pin of 28. The pin only shrinks
(ADR 0058, #2344): move the write behind Fountain.Machines.* rather than
raising it.
  10  apps/fountain/lib/fountain/conversations/conversation_server.ex
  4   apps/fountain/lib/fountain/workers/sandbox_reaper.ex
  3   apps/fountain/lib/fountain/conversations/termination.ex
  ...
  1   apps/fountain/lib/fountain/quotas.ex
```

`quotas.ex` appeared in the per-file breakdown at count 1, confirming the
listing surfaces exactly which file grew. Removed the probe with
`git checkout -- apps/fountain/lib/fountain/quotas.ex` and reran to green.

## Checks run

- `cd apps/fountain && mise exec -- mix test test/fountain/machines/direct_writes_test.exs` — green, both before and after the revert probe (fails as expected with the probe present).
- `mise exec -- mix precommit` from the worktree root — PASSED (9 stages, 295s): toolchain, conflict-markers, compile, deps, format, credo, dialyzer, sobelow, release.
- `mise exec -- mix format` from both the worktree root and `apps/fountain`.

No changelog fragment — nothing user-observable changes.

Base: `adr/0058-machine-owner` (#2342). Tracker: #2344.


BinaryBourbon (approved):
Approve — verified by reverting: an added direct `update_sandbox` call fails the ratchet naming the file, and a newly-introduced **aliased** provider call (`alias Managoat.Sandbox` + `Sandbox.destroy(x)`) is also caught (17 → 18, breakdown names the new file). The alias-resolution gap I was asked to grade blocking-if-missed does not exist here — the scan deliberately resolves the bare-alias pattern and the test proves it.

**Verified with `path:line`**

- Pins match an independent re-implementation of the scan (Python, not the test's own Elixir code) at `c6375bb7d`: 28 row writes / 17 provider mutations, per-file identical to the test's breakdown.
- `apps/fountain/test/fountain/machines/direct_writes_test.exs:1-162` — scan walks `apps/fountain/lib`, `ee/lib`, `apps/fountain_*/lib` (all five Buzz apps present at this SHA), excludes `lib/fountain/machines/` (doesn't exist yet, harmless no-op today).
- `reclaim_sandbox(` on `apps/fountain/lib/fountain/conversations/conversation_server.ex:1568,1668,1685` does not false-match `claim_sandbox(` — the `\b` boundary fails between `e` and `c`, confirmed by inspection and by the count (server = 10, matching ADR's 7 `update_sandbox` + 3 `claim_sandbox`).
- `apps/fountain/lib/fountain/workers/sandbox_reaper.ex:8` — the moduledoc's prose ``Managoat.Sandbox.destroy(handle)`` is correctly excluded (inside the `@moduledoc """..."""` block); the file's real 2 provider call sites (`:293` suspend, `:477` destroy) are the only ones counted.
- `apps/fountain/lib/fountain/conversations/provisioning.ex:36,124` — `alias Managoat.Sandbox` + `Sandbox.create_checkpoint(` at line 124 is resolved and counted; the same string in a `#` prose comment at line 907 is correctly excluded.
- `apps/fountain/lib/fountain/conversations.ex:165-177` — `claim_sandbox/2` calling `update_sandbox/2` internally is counted (documented behavior); both `def` sites at `:165` and `:173` and the `@spec` at `:172` are excluded from the count as definitions, not calls.
- Revert probe 1: added `Fountain.Conversations.update_sandbox(sandbox, %{status: "probe"})` to `home_checkpoint.ex` → test failed "direct sandbox row writes are 29, over the pin of 28", breakdown named `home_checkpoint.ex` with count 2. Restored from a copy.
- Revert probe 2: new file `lib/fountain/conversations/rv2350_probe.ex` with `alias Managoat.Sandbox` + `Sandbox.destroy(handle)` → test failed "direct provider mutations are 18, over the pin of 17", breakdown named `rv2350_probe.ex` with count 1 — the aliased call is caught. Deleted; worktree clean and green again.
- `mise exec -- mix test test/fountain/machines/direct_writes_test.exs` — green, no DB touched, `async: true`.
- `.credo.exs` — `"lib/fountain/machines/"` exemption added with an accurate reason (lease-established ownership, ADR 0058), same style and location as the existing directory-level entries (`lib/fountain/workers/`). #2348 adds the identical line at the identical point in its own branch (adds the real `lib/fountain/machines/` files) — a trivial same-line conflict at merge, not a finding.
- Assertion is `count <= pin` in both cases (`:90`, `:96`), moduledoc says "never raise it", failure messages list the per-file breakdown and point to `ADR 0058, #2344`.

No changelog fragment needed (test/config only, correct per CONTRIBUTING).

Issue #1614, opened 2026-09-06
Deployed suite: verify secret delivery, redaction, and brokered egress


jhgaylor:
Tracker: #1606

A normal turn can succeed while environment/vault merge, stored-secret decryption, broker header injection, or public-log redaction is broken.

## Acceptance criteria

- [ ] Use synthetic run-specific secrets and a controlled HTTPS echo service, never real third-party credentials or arbitrary destinations.
- [ ] Create the environment/vault/binding fixtures through public APIs; exercise documented precedence and credential delivery in a real sandbox.
- [ ] Verify broker-injected credentials arrive at the controlled service while sandbox-visible output and public APIs obey the documented secrecy boundary.
- [ ] Verify the public egress log records the expected request without disclosing credential values.
- [ ] Exercise a blocked destination and denied cross-tenant secret/binding access with exact expected outcomes.
- [ ] Check secret non-disclosure in SSE, persisted event/history APIs and suite artifacts, including intentionally echoed synthetic values where redaction is promised.
- [ ] Clean up bindings, vaults, environments, conversations and sandboxes; distinguish network/provider setup failure from an assertion failure.

## Dependencies and scope

Depends on: #1609.

Related: #1555 (broker network floor on wake) and #148 (future non-HTTP secret delivery). Test the currently documented broker contract; do not implement those product changes within this test issue.

Milestone 2 of the tracker. Proposed work; no delivery date or iteration is committed.

jhgaylor:
The receiver this profile needs no longer has to be hosted anywhere. #2351 (merged 2026-09-16) hosts it for the duration of a run and publishes it over Cloudflare quick tunnels, so:

```bash
node deployed/verify.mjs https://your-instance.example.com --profile secrets
```

needs no public host, no DNS record and no account. `--receiver-url` still takes an already-hosted receiver where an environment owns one.

Connections is also no longer a blocker: it was off for every production account because its PostHog flag had never been created (#2347), and is now enabled for the two suite accounts.

Live acceptance for this child is still open — this removes the setup barrier, it does not by itself record a passing run.

jhgaylor:
## Decision 2026-09-16: this closes on a recorded verdict

The receiver this profile needs no longer has to be hosted. #2351 hosts it for
the duration of a run and publishes it over Cloudflare quick tunnels, and
Connections is enabled on both suite accounts (#2347). Nothing blocks a run.

**What closes this issue:** the profile runs end to end against a real
[...]

PR #2351, opened 2026-09-16
feat(deployed): host the integration receivers over borrowed tunnels (#1614)

jhgaylor:
Builds on #2345. Makes the `secrets`, `mcp` and `webhooks` profiles runnable
without hosting any infrastructure — and the first live run of one found a
credential-handling defect, tracked privately.

## Why a tunnel

These three profiles assert on what a deployment does **outbound**: a secret
delivered into a sandbox, an MCP server called, a webhook posted. You cannot
verify outbound behaviour by asking the API; a party on the far end has to
receive it and be interrogated afterwards. That meant standing up a receiver at
a stable public origin — reasonable to ask of CI, unreasonable to ask of
someone who just wants to know whether a deployment works.

A run now borrows an origin instead. `deployed/tunnel.mjs` publishes a locally
hosted receiver over Cloudflare quick tunnels, which need no account and no DNS
record, and tears them down when the run ends.

`secrets` gets **two hostnames onto one receiver**, because
`ReceiverSession.verify()` asserts both report the same `instance_id`. That is
load-bearing: it is how "traffic reached the host that should have been
blocked" stays distinguishable from "no traffic arrived". A dead blocked
hostname would make a leak look like a correct denial.

## The part that was not obvious

A freshly issued hostname does not resolve for several seconds, and asking the
system resolver early is **worse than waiting**: the first NXDOMAIN is
negatively cached, and repeated lookups keep refreshing that entry rather than
letting it expire. Measured here, polling from t=3.7s never recovered within
58s, while waiting and asking once resolved immediately.

Readiness is therefore established against public resolvers directly, which
never touches the cache the run's own requests use. Only once the record
demonstrably exists does anything make a normal request. That is both faster
and steadier than the blind hold it replaced — ~12s rather than 29s — and
cannot poison our own lookups.

Both origins also start together. Started in turn, the second hostname was
seconds younger and its edge was still catching up, which is why it was always
the second one that failed.

A tunnel that never comes up fails **setup**, with its own diagnostic. A
borrowed origin must never be mistaken for a failed assertion about the
deployment under test.

## Reconciling with #2345

#2345 removed these three profiles from the advertised set, because an
advertised profile that cannot be configured only produces a run that fails
setup. That reasoning holds; this branch removes the premise, so they are
advertised again. `schedules` stays out — nothing here supplies its windows.

Its regression test caught a real design flaw while doing so: patching receiver
settings onto a config after composing it meant `verifyConfig` alone never
produced a loadable target for those three. Composition now happens once, with
the receiver settings, and a profile that needs a receiver cannot compose
without one. Validation also moved ahead of hosting, so a typo never opens a
public tunnel first.

## Verification

- 280 Node tests pass, up from 259 on main.
- Tunnel behaviour is covered with a fake `cloudflared`: registration ordering,
  two origins onto one port, teardown on every failure path, no lookup before
  the record exists, and a missing binary reported as setup. Each was checked
  by breaking the thing it guards.
- Against production on the rebased branch: `probe` passes, `schedules` is
  refused with a pointer to `cli.mjs`, an unbound target still refuses to use a
  stored key, and `secrets` runs over two live tunnels to its assertions —
  receiver identity confirmed across both hostnames, fixtures created,
  sandbox provisioned, zero resources left behind.

## What the first live run found

`secrets` fails on production, and not in the harness. The finding is recorded
in a private draft advisory rather than here. It is tenant-scoped, involves no
real credential, and the suite behaved correctly throughout — this PR is what
made it visible.


BinaryBourbon (changes requested):

Reviewed all six changed paths at `77fa49fe84e0ed3e901f6a6aa0922619a5608333` against base `2bd41be09c02fb7e1533d94e771a01740e0366f7`, including receiver hosting, tunnel subprocesses/readiness/teardown, configuration composition, both test files and the changelog.

**[P2] Validate external receiver origins before persisting their URLs.** `deployed/lib/local-receiver.mjs:35–45` checks origin count and credential presence, then copies the supplied URLs into settings without applying the existing receiver-origin validators. `deployed/verify.mjs:215` writes those settings to `target.json` before `run` calls `configFrom`. Consequently, a rejected `--receiver-url https://operator:fake_receiver_password@receiver.example.invalid/` leaves its password in the retained target file even though setup subsequently fails. Query and fragment credentials take the same path; the secrets profile's `--blocked-url` is copied through the same unvalidated settings. This bypasses the early refusal already implemented for the deployment URL.

I reproduced the actual `verifyMain` → `run` → `configFrom` path for userinfo, query and fragment markers on each of `secrets`, `mcp` and `webhooks`: all nine runs returned setup exit 2, made zero requests, and retained the fake credential verbatim in `target.json`. The terminal and `result.json` did not contain it. The file is mode 0600 under a 0700 directory, so this is unintended plaintext retention in the configuration artifact, not an unauthenticated file-read claim. Validate both receiver origins using the profile's existing validator before returning settings or writing configuration; reject without quoting input. Add entry-point regressions asserting refused receiver URLs leave no credential-bearing artifacts.

Other security evidence: with stub tunnels and fake data, I exercised all three actual receiver modules on loopback. Public identity worked; absent and incorrect admin credentials were denied; generated admin credentials authenticated. A secrets capture retained only the intended observation fields, with no raw synthetic values, unrelated body field or admin credential in the evidence. Normal stop and tunnel-setup failure closed their local listeners. Source tracing confirms generated admin keys remain in the runner environment object, receiver settings contain only credential variable names, control requests use the existing redacting client, and profiles preserve distinct-tenant checks and bounded fixture cleanup. No public tunnel was opened.

The coordinating reviewer reports 280/280 exact-head Node tests and changelog/conflict/whitespace checks passed, and verified [exact-head CI success](https://github.com/managoat/fountain/actions/runs/35111878725). The new tests do not cover the persistence path above.

Source review and the focused diagnostics are complete. Only fake credentials, intercepted external fetches and loopback fixtures were used; no real account, credential, deployment or provider resource was accessed. Live Cloudflare behavior and a full deployed conformance run were not tested. This is an independent Codex review, not a Review Loop App verdict.

BinaryBourbon (changes requested):
Independent adversarial product, API, and operator compatibility review by Codex.

**Verdict: REQUEST_CHANGES** at `77fa49fe84e0ed3e901f6a6aa0922619a5608333`.

**P2 — Preserve cleanup replay after discarding an automatic receiver.** `deployed/lib/local-receiver.mjs:69–71` returns the generated admin key only in a temporary environment; `deployed/verify.mjs:214–224` passes it to the first run, then stops the receiver/tunnels even when fixture cleanup failed. The saved target still selects `secrets`, `mcp`, or `webhooks` and references that unavailable credential. The documented cleanup command loads this same target through `configFrom`, which requires the receiver key before it can reach any Fountain cleanup operation. Thus the artifacts do not support the promised retry of leftover resources.

Reproduced with the real `hostReceiver`, `verifyConfig`, and runner, a fake tunnel, and loopback receiver/API fixtures: the webhook run retained one environment after a simulated cleanup DELETE returned 503 and exited 3. After receiver shutdown, replay with the generated target, manifest, and original account credentials exited 2 with `Missing webhook receiver admin credential`; it made **zero HTTP requests**, leaving the resource untouched. The same credential lifetime affects all three automatic receiver profiles. Provide a cleanup path that can recover run-owned Fountain fixtures independently of the stopped ephemeral receiver, while preserving ownership checks and hosted-receiver cleanup. Add a run-failure-to-replay regression covering this boundary; merely preserving the key would still leave a dead receiver origin.

Reviewed all six changed paths:

- `deployed/lib/local-receiver.mjs`: profile settings, credentials, hosting, and teardown.
- `deployed/tunnel.mjs`: process startup, DNS/origin readiness, and failure handling.
- `deployed/verify.mjs`: flags, validation, configuration, exit/report behavior, and recovery callers.
- `deployed/test/tunnel.test.mjs`: fake-process/readiness and hosted-receiver cases.
- `deployed/test/verify.test.mjs`: profile composition, loader compatibility, and CI agreement.
- `changelog.d/1614-tunnelled-receivers.md`: advertised operator behavior.

Also traced the existing receiver protocols, profile setup checks, configuration validators, and documented cleanup replay. Server/SDK wire contracts are unchanged.

Validation: independently ran the cleanup regression probe and inspected the coordinated exact-head Node test log (280/280 passing). [Exact-head CI](https://github.com/managoat/fountain/actions/runs/35111878725) passed according to coordinated verification. Local evidence used synthetic keys, fake tunnels, and loopback fixtures only; no public tunnel or deployed resource was exercised. This is a harness compatibility review, not live integration evidence.

BinaryBourbon (changes requested):
**REQUEST_CHANGES — independent adversarial correctness/lifecycle review by Codex.**

Reviewed the whole PR at `77fa49fe84e0ed3e901f6a6aa0922619a5608333` against `2bd41be09c02fb7e1533d94e771a01740e0366f7`.

1. **P2 — Parse cloudflared output across stream chunk boundaries.** `deployed/tunnel.mjs:64–69` treats each `data` chunk as a complete message. A valid hostname split after `trycloud`, or a registration line split between `tunnel ` and `connection`, is never recognized. The coordinating reviewer's actual fake child wrote identical logical lines with different chunk boundaries: whole lines became ready in 74 ms; both split cases remained pending until the probe deliberately aborted at 500 ms. Without that abort, the source's 60-second startup timer would reject a tunnel that already announced itself. Buffer incomplete lines separately for stdout/stderr, then parse complete lines (with bounded buffering). Add both split cases to the regression tests; the current fake emits complete lines only.

2. **P2 — Preserve the interrupted exit code during receiver startup.** `deployed/verify.mjs:210–211` adds an abortable setup phase before `run()`, but its rejection reaches the unconditional exit-2 catch at `:247`. My actual CLI probe sent SIGINT after the loopback receiver and a fake cloudflared child started: it printed `Interrupted` and exited **2**, despite the command's documented **130** contract. This reports an operator cancellation as a setup failure, depending solely on when Ctrl-C arrives. Handle an aborted controller around the new hosting phase and return 130 after teardown, preserving exit 2 for genuine setup failures. Cover SIGINT/SIGTERM during startup. The probe confirmed the fake child was killed and the receiver port closed; teardown itself passed this case.

Coverage: `deployed/tunnel.mjs`, `deployed/lib/local-receiver.mjs`, `deployed/verify.mjs`, `deployed/test/tunnel.test.mjs`, `deployed/test/verify.test.mjs`, and `changelog.d/1614-tunnelled-receivers.md`. Traced local receiver construction/authentication, external settings, profile composition, the runner's setup/result/cleanup paths, and the secrets profile's distinct-host/same-instance checks. The existing identity assertions remain in place; two origins target the same local receiver port.

Validation: source review plus a local-only CLI interruption probe with explicit child-process stubbing and fake suite keys. I also inspected the coordinating reviewer's executable split-log probe and captured real pipe chunks. The coordinator ran **280/280 Node tests** on this head; changelog, conflict-marker and whitespace checks passed, and [exact-head CI passed](https://github.com/managoat/fountain/actions/runs/35111878725). No real cloudflared tunnel, deployment, credentials, or private advisory was used for this review. These diagnostics validate harness behavior, not deployed conformance.

jhgaylor:
All four addressed in `197ab945e`. Every one was real; each fix was checked by
reverting it and watching the new regression fail.

**[P2] External receiver origins are validated before anything is written.**
`externalReceiver` now applies the profile's own validator — `receiverOrigins`
for `secrets`, `controlledOrigin` for the other two — before returning
settings, so a URL never reaches `target.json` on its way to being rejected.
The refusal deliberately does not quote the URL back, and a test asserts the
planted password appears in no message for userinfo, query and fragment on
both `--receiver-url` and `--blocked-url`. Same-hostname `--blocked-url` is
refused too, since that would make a leak indistinguishable from a denial.

**[P2] Cleanup survives the receiver going away.** You are right that the
generated credential dies with the run and the saved target still demands it.
A `cleanup-target.json` is now written beside the run's target, carrying the
instance, the account and the limits — and nothing a receiver owns. The
manifest drives cleanup, not the profile, so dropping it loses nothing. When
resources remain, the summary prints the exact replay command pointing at that
file.

The regression is the boundary you described: replaying through the run's own
target makes **zero** Fountain calls and leaves the fixture, while replaying
through the cleanup target deletes it and exits 0.

**[P2] cloudflared output is parsed by line.** Each stream buffers its own
remainder, bounded at 64 KiB so a stream that never sends a newline cannot
grow without limit. Three regressions cover a hostname split mid-word, a
registration line split mid-phrase, and a single chunk with no trailing
newline. With per-chunk matching restored, the split cases hang until the
startup timer — which is exactly the failure you measured.

**[P2] An interrupt during startup exits 130.** An aborted controller now
returns 130 after teardown, with exit 2 kept for genuine setup failures. The
test drives the real `verifyMain` and sends SIGINT while the receiver is
coming up.

Checks: 289 Node tests pass, up from 280. ShellCheck and the changelog check
pass. The persistence, teardown and parsing paths the previous tests missed
are now covered by the executable paths themselves.

jhgaylor:
Follow-up in `7f99da66f`: re-verifying the four fixes against production surfaced two problems in them, both mine rather than yours.

A refused receiver URL still created the run's output directory, so a run that never started left something that looks like one that did. The directory is now created only once nothing is left to refuse. Its name also carries milliseconds and a short random suffix — two runs in the same second collided on `EEXIST`, which the very next invocation hit.

`externalReceiver` also checked the admin credential before validating the URLs, so a credential-bearing URL was reported as a missing environment variable rather than as a bad URL. Validation now runs first: a URL is refused for what it is, not for what the environment happens to be missing.

Verified end to end: the credential-bearing URL is refused without echoing it and leaves no directory behind, and a live `secrets` run over two real Cloudflare tunnels reaches its assertions with zero resources remaining. 289 Node tests still pass.

BinaryBourbon (changes requested):
Independent adversarial product, API, and operator compatibility re-review by Codex.

**Verdict: REQUEST_CHANGES** at `f8abc78329137c4d7c523ff0d61d3c49c6dca97b`.

**P2 — Reconcile receiver sidecars when generating the receiver-free cleanup target.** `deployed/verify.mjs:160–161` removes the receiver configuration, and lines 240–253 write and recommend that target for replay. However, cleanup is also driven by files beside `cleanup.json`: `deployed/lib/runner.mjs:276–299` schedules receiver cleanup whenever `receiver.json`, `mcp-receiver.json`, or `webhook-receiver.json` exists, then requires the configuration this helper removed. It does this even when the sidecar already says `state: "cleaned"`.

An independent probe used the actual `verifyMain`, real local MCP receiver, and fake tunnels. The initial run retained an environment after a simulated cleanup DELETE returned 503, successfully cleaned its receiver, persisted `mcp-receiver.json` with state `cleaned`, and stopped the tunnel. The emitted replay command, using the new `cleanup-target.json`, **did delete the environment**, but still exited 3 with `MCP cleanup requires the original target configuration` and zero resources remaining. A second replay failed identically. The prior missing-key gate is therefore fixed, but ordinary receiver sidecars prevent recovery from reporting success. The new regression test misses this because its workspace creates only `cleanup.json`.

Carry the ephemeral receiver's lifecycle into replay and reconcile those sidecars consistently with shutdown, while retaining cleanup for externally hosted receivers and all Fountain ownership checks. Cover an actual automatic run that creates a receiver sidecar, leaves a fixture behind, and then completes replay successfully; include an already-cleaned sidecar.

Reviewed the entire seven-path PR against `2bd41be09c02fb7e1533d94e771a01740e0366f7`:

- `deployed/lib/local-receiver.mjs`: local/hosted settings, origin validation, credential lifetime, and teardown.
- `deployed/tunnel.mjs`: startup parsing, DNS/origin readiness, and cleanup.
- `deployed/verify.mjs`: flags, composition, cancellation, artifacts, and replay guidance.
- `deployed/test/cleanup-after-receiver.test.mjs`: replay coverage and its missing sidecar case.
- `deployed/test/tunnel.test.mjs`: original cases plus fragmented logs and rejected origins.
- `deployed/test/verify.test.mjs`: configuration compatibility and cancellation regression.
- `changelog.d/1614-tunnelled-receivers.md`: operator claims.

Server/SDK wire contracts are unchanged. The new origin validation and stream buffering address their identified boundaries; this review's remaining finding is the demonstrated replay failure.

Validation: independently ran the actual-entry-point/replay probe and inspected the coordinated exact-head Node log (289/289 passing). [Exact-head CI](https://github.com/managoat/fountain/actions/runs/35119532311) passed according to coordinated verification. The probe used synthetic credentials, fake tunnels, and loopback API/receiver fixtures only; no public tunnel or deployed resource was exercised.

BinaryBourbon (changes requested):
Independent adversarial security re-review by Codex — **REQUEST_CHANGES**.

Reviewed all seven changed paths at `f8abc78329137c4d7c523ff0d61d3c49c6dca97b` against base `2bd41be09c02fb7e1533d94e771a01740e0366f7`, covering the original receiver/tunnel implementation and the fixes, configuration/output, three test files and changelog.

The credential-retention finding from my previous review is resolved. The existing receiver-origin validators now run before configuration output. My actual `verifyMain` probes passed all 12 cases: userinfo/query/fragment on `--receiver-url` for all three receiver profiles, plus all three forms on the secrets profile's `--blocked-url`. Each was refused without requests, hosting, an output directory or echoed fake credential. The hosted-entry-point probe also confirmed 0700/0600 output permissions, no generated admin credential in either target file, and receiver teardown on setup failure.

**[P2] Handle the receiver journal when replaying the new cleanup target.** `deployed/verify.mjs:159–161` removes the receiver configuration from `cleanup-target.json`, but the original results retain `receiver.json`, `mcp-receiver.json` or `webhook-receiver.json`. The existing runner schedules receiver cleanup whenever that file exists, then requires the corresponding configuration even when its recorded state is already `cleaned`. Consequently, the newly advertised replay can delete every remaining Fountain fixture yet exit 3 forever with “MCP cleanup requires the original target configuration.” This leaves the operator without a successful cleanup verdict despite zero remaining fixtures.

I independently traced this interaction and inspected the API reviewer's executable reproduction and artifacts. It runs the actual `verifyMain`, local MCP receiver and `replayCleanup` with fake tunnels and a loopback API. A temporary initial deletion failure leaves one environment; the real receiver journal records `cleaned` and its tunnel stops. Replay through the generated cleanup target then deletes that environment, reports zero remaining resources, and fails on the receiver journal. A second replay fails identically. The new cleanup regression misses this because its synthetic results contain only `cleanup.json`.

Recognize a validated, already-cleaned receiver journal without requiring its live configuration, and explicitly represent disposal of a locally hosted receiver where needed. Preserve external receiver cleanup and the existing ownership checks. Add a regression using the actual profile-produced receiver journal and generated cleanup target.

The cleanup target retains its deployment/account boundary: my real replay probes denied deletion for a different authenticated owner, mismatched manifest target and changed live resource marker, while deleting the valid run-owned fixture. Source review found no new credential exposure or ownership bypass in the fixes. The coordinating reviewer reports 289/289 exact-head Node tests, changelog/conflict/whitespace checks, and all three original split-log probes passed; [exact-head CI passed](https://github.com/managoat/fountain/actions/runs/35119532311).

Source review and the described diagnostics are complete. Only fake credentials, intercepted external fetches, stub tunnels and loopback fixtures were used. No real account, keychain credential, public tunnel, deployment or provider resource was accessed. Live Cloudflare/deployment behavior was not tested. This is an independent Codex review, not a Review Loop App verdict.

BinaryBourbon (changes requested):
**REQUEST_CHANGES — independent adversarial correctness/lifecycle re-review by Codex.**

Reviewed all seven changed paths at `f8abc78329137c4d7c523ff0d61d3c49c6dca97b` against `2bd41be09c02fb7e1533d94e771a01740e0366f7`. My two previous findings are resolved: actual child-process probes now recognize both split hostname and split registration output, and actual SIGINT/SIGTERM during startup both return 130 while killing the fake child and closing the receiver port. Early external-origin validation also addresses the previous URL-persistence defect in the inspected code.

**P2 — The new cleanup target still cannot produce a successful replay with a real receiver manifest.** `deployed/verify.mjs:160` removes receiver configuration, but the cleanup-mode runner still schedules receiver cleanup whenever `receiver.json`, `mcp-receiver.json` or `webhook-receiver.json` exists (`deployed/lib/runner.mjs:276–299`). Those callbacks require the removed configuration even when the sidecar already records `state: "cleaned"`.

Trigger: a hosted MCP run creates its receiver and a Fountain fixture; fixture deletion temporarily returns 503, while receiver cleanup succeeds. Follow the emitted replay command after the API recovers. The API reviewer's actual `verifyMain`/local MCP receiver probe confirms the new target deletes the leftover fixture, but replay returns **3** with `MCP cleanup requires the original target configuration` and **zero resources remaining**. A second replay also returns 3. Thus the original missing-key barrier is fixed, but the recovery verdict remains permanently failed because it demands configuration for an already-cleaned, stopped receiver.

Make replay recognize completed teardown of a hosted receiver while preserving cleanup obligations for external receivers. Extend the regression test with the receiver sidecar an actual profile writes, and require a successful replay after deletion. The new `cleanup-after-receiver.test.mjs` fixture writes only `cleanup.json`, so its passing case does not exercise this path.

Coverage: `deployed/tunnel.mjs`, `deployed/lib/local-receiver.mjs`, `deployed/verify.mjs`, `deployed/test/tunnel.test.mjs`, `deployed/test/verify.test.mjs`, `deployed/test/cleanup-after-receiver.test.mjs`, and `changelog.d/1614-tunnelled-receivers.md`. Rechecked the original implementation and all fixes, including receiver construction, multi-origin identity checks, configuration, cancellation, teardown, result classification, and replay callers.

Validation: my split-log and real-signal probes passed against this head; I independently inspected the API reviewer's executable sidecar probe and recorded requests/results. The coordinating reviewer ran **289/289 Node tests** successfully; changelog, conflict-marker and whitespace checks passed, and [exact-head CI passed](https://github.com/managoat/fountain/actions/runs/35119532311). All executable review probes used synthetic keys, fake tunnels and local fixtures. No real deployment, public tunnel or private advisory was accessed. Earlier resolved findings are not being carried forward as blockers.

jhgaylor:
Fixed in `37d83efe9`. All three reviews found the same thing and it was real:
the cleanup target cleared the missing-credential barrier and ran straight
into the next one.

**The journal is what knows.** Cleanup mode schedules receiver cleanup whenever
a journal sits beside the manifest, and demanded the receiver configuration
even when that journal already recorded the work as done. So a replay deleted
every fixture, reported zero remaining, and still exited 3 — permanently, as
you showed.

A journal now settles the obligation when it records a terminal state:

- `cleaned` — the run deleted its records before it ended.
- `discarded` — new here: the receiver was hosted for the run and stopped with
  it, so the instance holding those records no longer exists.

Tearing down a hosted receiver writes `discarded` into any journal that has not
already reached a verdict of its own, so a crash before receiver cleanup is
representable rather than inferred from a configuration that is also gone.

**The obligation survives everywhere else.** An unreadable journal, one
belonging to another run, or any other state still takes the path that demands
the configuration — which is correct for a receiver you host, because it is
still reachable and its records still have to go. A test covers that case
explicitly, alongside both terminal ones.

**On the test that missed it:** you were right that a results directory holding
only `cleanup.json` was not exercising this. It now writes the journal a real
profile writes. Reverting the runner change fails both terminal cases with
`MCP cleanup requires the original target configuration`, which is your
reproduction exactly.

**Verified on real artifacts**, not synthetic ones — a live `secrets` run
against production over two Cloudflare tunnels, which wrote a genuine
`receiver.json` in state `cleaned`:

```
cleanup-target.json  -> exit 0, passed, secrets/receiver-cleanup passed
target.json          -> exit 3, "Missing controlled receiver admin credential"
```

That also surfaced a papercut of my own, fixed in `7135d0c6e`: the replay
command the summary prints is run by `cleanup-replay`, which reads the
environment and nothing else, so a run whose keys came from the keychain was
handed a command that fails on a missing credential. It now says to export them
first — I hit that myself and briefly mis-read it as this defect.

Checks: 293 Node tests pass, up from 289.

BinaryBourbon (approved):

Reviewed all 11 changed paths at `7135d0c6eb00dc9f8a18464ba473b058528d9fe4` against base `2bd41be09c02fb7e1533d94e771a01740e0366f7`: the original local receiver and tunnel implementation, CLI configuration/output, receiver journal and session changes, runner cleanup, all three changed test files and changelog. No concrete introduced security or tenant-isolation defect remains in this review.

Both findings from my earlier reviews are resolved. Receiver URL validation still precedes persistence: all 12 actual entry-point userinfo/query/fragment refusals passed on this head, including both secrets receiver flags, without requests, hosting, output directories or echoed fake credentials. The hosted-entry-point probe confirmed private output permissions (0700 directories, 0600 target files), no generated admin credential in either target file, and teardown after setup failure.

The new terminal-journal handling preserves the cleanup boundary. My 30 actual `replayCleanup`/runner/session diagnostics covered all three receiver journal types. Matching cleaned/discarded journals permit replay; pending/created journals without configuration still fail. Journals from another run, malformed JSON, a changed external receiver origin, a different authenticated owner and a different deployment target do not trigger receiver requests. An outstanding external receiver with matching configuration still receives its authenticated, run-scoped DELETE and confirming GET, then records `cleaned`. Receiver requests retain manual redirect handling. Separate fixture probes confirmed that changed live resource markers prevent deletion while valid run-owned fixtures are removed.

I also inspected the API and correctness reviewers' actual MCP lifecycle reproductions and artifacts. Both cleaned and discarded hosted journals now permit two successful replays through the generated cleanup target, removing the leftover Fountain fixture. With receiver deletion temporarily failing, the local receiver is stopped before its journal becomes `discarded`. The external-receiver case stays `created` and remains running until replay performs authenticated cleanup through its original configuration. The local journals do not replace the existing authenticated owner, deployment target or live resource ownership checks.

The coordinating reviewer reports 293/293 exact-head Node tests passing, clean changelog/conflict/whitespace checks, and a mutation check that fails both terminal-journal regressions when the new runner branches are disabled. [Exact-head CI passed](https://github.com/managoat/fountain/actions/runs/35125330998).

Source review and the described diagnostics are complete. The independent boundary probes used synthetic journals and intercepted fetches; the lifecycle reproduction used real local receiver code with stub tunnels and a loopback API. All credentials were fake. No real keychain credential, public tunnel, deployment or provider resource was accessed; live Cloudflare/deployment behavior was not tested. This is an independent Codex review, not a Review Loop App verdict.

BinaryBourbon (approved):
**APPROVE** — independent adversarial API/operator re-review by Codex of the whole PR at `7135d0c6eb00dc9f8a18464ba473b058528d9fe4`, against `2bd41be09c02fb7e1533d94e771a01740e0366f7`. No current blocking defect found; my previous cleanup-replay finding is resolved.

I reran the actual `verifyMain` → real local MCP receiver → `replayCleanup` flow using synthetic credentials, a loopback Fountain fixture and fake tunnels. Each initial run deliberately left an environment after a failed DELETE. With a naturally persisted `cleaned` receiver journal, the emitted `cleanup-target.json` now deletes that environment and succeeds on two successive replays. When receiver deletion also fails, successful local shutdown records `discarded`, and both replays likewise exit 0 with no remaining resources. An externally hosted receiver stays running with its journal `created`; replay with its original configuration performs receiver cleanup and changes the journal to `cleaned`. Thus the fix removes the stale local obligation while preserving the external one.

Coverage: all 11 changed paths, including `verify.mjs`, `tunnel.mjs`, `lib/local-receiver.mjs`, `lib/receiver-journal.mjs`, `lib/runner.mjs`, both receiver session modules, all three changed test files and the changelog. Traced CLI validation/help, profile composition and CI compatibility, URL/credential handling, readiness and setup errors, journal producers and consumers, fixture ownership checks, teardown ordering, replay status and repeatability. The change does not alter server endpoints or SDK wire schemas.

Verification: the coordinated exact-head run passed all 293 Node tests, with changelog and whitespace checks clean. Disabling only the three terminal-journal branches made the new cleaned/discarded replay tests fail with exit 3 instead of 0; the isolated checkout was restored. [Exact-head CI passed](https://github.com/managoat/fountain/actions/runs/35125330998).

Limits: these diagnostics use local fixtures and fake tunnel transport; they do not establish live Cloudflare propagation or deployment/provider conformance. No real credentials, public tunnels or live deployment resources were used.

BinaryBourbon (approved):
**APPROVE — independent adversarial correctness/lifecycle re-review by Codex.**

Reviewed all 11 changed paths at `7135d0c6eb00dc9f8a18464ba473b058528d9fe4` against `2bd41be09c02fb7e1533d94e771a01740e0366f7`. The previously reported cleanup-replay defect is resolved in the inspected implementation, and I found no new blocking correctness issue.

Replay now recognizes a terminal receiver journal for the same run before requiring the receiver configuration. Hosted receiver records are marked `discarded` only after the awaited receiver teardown; unresolved journals still require receiver cleanup. My new local integration probe exercised the harder disposal case: the real MCP receiver's DELETE was forced to return 503, fixture cleanup also temporarily failed, and `verifyMain` stopped the hosted receiver and persisted `discarded`. Replay through the emitted cleanup target deleted the leftover fixture and returned **0**; a second replay also returned **0**, with zero resources remaining. This verifies the persisted lifecycle across separate run/replay calls, beyond a fabricated terminal journal. I also inspected the API reviewer’s executable probes and evidence: already-cleaned hosted receivers now replay successfully, while an unresolved external receiver remains an obligation and is cleaned through its original configuration.

The prior split-log and startup-cancellation fixes remain intact. I rechecked their source and regression assertions; actual split-log and SIGINT/SIGTERM probes passed on the preceding revision, and those implementations are unchanged here. Early external-origin validation continues to precede persistence. No earlier resolved finding is being carried forward.

Coverage: `deployed/tunnel.mjs`, `deployed/lib/local-receiver.mjs`, `deployed/lib/receiver-journal.mjs`, `deployed/lib/receiver.mjs`, `deployed/lib/controlled-receiver.mjs`, `deployed/lib/runner.mjs`, `deployed/verify.mjs`, `deployed/test/tunnel.test.mjs`, `deployed/test/verify.test.mjs`, `deployed/test/cleanup-after-receiver.test.mjs`, and `changelog.d/1614-tunnelled-receivers.md`. Rechecked the original changes and all fixes, including setup, distinct-host/same-instance receiver checks, journal transitions, cancellation, teardown exceptions, result classification and cleanup replay.

Validation: my executable diagnostic used actual harness/local receiver code with synthetic keys, fake tunnel origins and loopback fixtures. The coordinating reviewer ran **293/293 Node tests**, plus changelog, conflict-marker and whitespace checks; disabling the three terminal-journal branches in an isolated mutation diagnostic made the cleaned/discarded replay regressions fail as expected. [Exact-head CI passed](https://github.com/managoat/fountain/actions/runs/35125330998). No real deployment, public tunnel, real credential or private advisory was accessed. This approval covers the harness change, not live deployed conformance.

PR #2355, opened 2026-09-16
fix(conversations): redact the credentials the broker holds

jhgaylor:
Fixes the defect the deployed suite's `secrets` profile found against
production (#1614). Recorded in the changelog rather than a security advisory:
there are no known deployments other than the operator's own.

## What was wrong

A secret bound to an egress binding never enters the sandbox. `Broker.split/2`
swaps it for a placeholder and the broker attaches the real value at the proxy
(ADR 0019). `Fountain.Conversations.Redaction`'s registry is built from the
sprite environment — so the value the broker kept was never registered, and the
one credential brokering exists to withhold from the agent was the one value
never scrubbed from what the agent reads.

Keeping a credential out of the sandbox does not keep it out of the sandbox's
**output**. The broker attaches it to an outbound request; an upstream that
echoes a header back returns it as ordinary tool output; `Conversations.log!/1`
stores what a sprite writes verbatim. It landed in `log_events` in plaintext —
a table with none of the envelope encryption the secret itself has — and went
out on `GET /api/conversations/:id/stream`.

`conversation_server.ex:1357` already carried this lesson, in a comment, for
the Claude API key it injects:

> The API key was never in the sprite env before now, so `build_sprite_env`
> never registered it for redaction — do it here, or the very value this fix
> injects prints in plaintext into `log_events`.

It was never generalised to the values the broker holds from the start.

## The fix

`SpriteEnv.build/4` takes the broker's credentials and registers them with the
environment it builds. Registration stays where that module's comment already
promises it — "before anything can log" — rather than adding a second pass in
the caller. The API-key fallback folds `brokered` back in for the same reason
it already folded the outgoing env: a put replaces rather than merges.

The sandbox still sees only the placeholder; a test asserts that, so the fix
cannot quietly "work" by putting the credential somewhere it does not belong.

## Scope

Not a cross-tenant disclosure. The conversation, its events and its stream are
scoped to the account owning the credential, and the suite separately verifies
a second tenant is denied.

**What this fixes:** the credential no longer sits in plaintext in `log_events`,
and is no longer served through the conversation's history and stream — nor,
through them, to the Conversations app or anything else reading output via
Fountain. That holds after a rotation too (see the second commit).

**What this does not change, and cannot:** the agent still sees an echoed
credential. The echo arrives inside the sandbox; the runtime reads the tool's
output and hands it to the model there, before Fountain persists anything, and
`Redaction` applies only at persistence. ADR 0019 §10 records that the broker
does not rewrite responses. An earlier version of this description presented
agent exposure as the harm being fixed; that was wrong, and is corrected here
and in the changelog.

## Verification

- `mix test apps/fountain/test/fountain/conversations/` — 1566 tests, 0
  failures.
- The regression was checked by reverting the registration: the test then fails
  with `["__bound_token_placeholder__", ...]` registered and the credential
  absent, which is the defect exactly.
- A second test pins that a run with no broker registers what it always did.
- `conversation_server.ex` is four lines shorter than before; the size pin is
  untouched.

Live re-verification against production with the deployed suite is the real
proof and should follow this landing, since the profile that found it is the
profile that confirms it.


BinaryBourbon (changes requested):

Reviewed all four changed paths at `2603ae16c159a76da6f57d3d5281cea7833e9084` against base `dd9c453763c18611a7832db61ee61ae8ef93772e`, including the broker split/refresh lifecycle, initial provision and reattach, OAuth fallback dispatch, redaction registry, persistence and live/historical SSE consumers, tests and changelog.

**[P1] Keep brokered credentials registered after a live refresh.** The new registration at `apps/fountain/lib/fountain/conversations/sprite_env.ex:124–129` captures only the values present when the environment is built. On a later turn, `ConversationServer.run_turn/6` calls `Egress.refresh_before_turn/1` before either reusing the peer or spawning it. That path rereads tenant secrets and connection tokens, replaces `state.brokered` and updates the live proxy rules without rebuilding the environment or updating `Redaction`. Editing a bound token while a conversation is running therefore restores the exact disclosure this PR fixes: an upstream echo of the replacement value is written into `log_events` and delivered in history and SSE, because only the old value is registered. This is an ordinary supported operation: the existing secrets documentation explicitly promises that an edited brokered vault token takes effect on the next turn.

My pinned Elixir 1.19.2/OTP 28.3 diagnostic exercised both a tenant-secret edit and a connection-token rotation. It compiled the exact head `SpriteEnv` and `Redaction` modules and the exact broker-split/egress-refresh function bodies, with fake storage and broker-session boundaries. In both cases the fresh value reached the real refresh function's resulting broker state and broker-write payload, remained absent from `Redaction.lookup/1`, and survived `Redaction.redact/2` unchanged. The original value was masked. A control registering the fresh value masked the same echo. Source tracing confirms `Output.persist/3` uses `Conversations.log!/1`'s registry-based scrubber, broadcasts its returned persisted event, and both SSE paths consume those events without another credential refresh.

Register each refreshed broker value before it can produce output, preserving the old values as well as the new values for output already in flight. Cover both live-rule rewrites and replacement sessions, and add a regression that provisions, rotates a bound secret, runs the refresh path and verifies the new echo is scrubbed while the sandbox environment still contains only placeholders.

The initial-build fix does keep raw broker values out of the returned sandbox environment and adds them only to the conversation's registry. This patch does not change tenant selection, per-tenant decryption, broker authorization or the owner checks on event history/streams; the identified leak is within the credential owner's conversation. I found no separate reachable fallback defect: current resolved inference sources take the earlier identity-bound branch. The registry protects persisted/streamed output; it does not filter an upstream response before the sandbox runtime receives it, a distinction already documented in ADR 0019.

The coordinating reviewer reports 1,566 tests with zero failures and two Linux-only skips, using Elixir 1.19.2/OTP 28.3 and a disposable PostgreSQL 18 database, plus passing format/changelog/conflict/whitespace checks. [Exact-head CI passed](https://github.com/managoat/fountain/actions/runs/35131010719). The focused rotation probe demonstrates a case the passing suite does not cover.

Source review and the focused diagnostic are complete. The diagnostic used only fake credentials and stubbed external boundaries; it was not a database-backed or sandbox/proxy integration test. No real credentials, deployment or provider resources were accessed. This is an independent Codex review, not a Review Loop App verdict.

BinaryBourbon (changes requested):
**REQUEST_CHANGES** — independent adversarial API/operator review by Codex of the whole PR at `2603ae16c159a76da6f57d3d5281cea7833e9084`, against `dd9c453763c18611a7832db61ee61ae8ef93772e`.

**P1 — Register refreshed broker credentials before the next turn.** The new registration in `apps/fountain/lib/fountain/conversations/sprite_env.ex:124–129` covers provisioning and reattachment, but a running conversation takes a different path. Before each turn, `ConversationServer.run_turn/6` calls `Egress.refresh_before_turn/1`, which replaces broker-held environment/vault or connection credentials and updates the proxy rules without rebuilding the environment or updating `Redaction`. Edit a bound vault token between turns: the broker uses the new token, while the registry still knows only the old value. An upstream echo of the new token therefore still reaches `log_events`, history and the live/replayed SSE stream in plaintext. This is the documented next-turn rotation behavior in `docs/concepts/secrets.md:83–88`, not a hypothetical state mutation.

Update the conversation's redaction registry when refreshed broker values are adopted, before dispatching the next prompt, retaining values already registered for possible later output. Add a regression that rotates a bound credential on an already-running conversation and verifies that an echoed new value is scrubbed through the log writer. The new initial-build test does not exercise this lifecycle.

Coverage: all four changed paths — `conversation_server.ex`, `sprite_env.ex`, `sprite_env_test.exs` and the changelog — plus broker splitting/refresh, inference source binding, runtime fallback, output persistence and SSE consumers. `SpriteEnv.build/4` keeps its arity and returned environment ordering; the optional map defaults to empty, actual callers provide broker credential maps, and real credentials are not added to the sandbox environment. No endpoint or SDK wire schema changes are required. The ordinary bound OAuth refusal path keeps its registry; I found no supported current trigger warranting a separate fallback finding.

Verification: I inspected the coordinated fake-data diagnostic using the exact `SpriteEnv`/`Redaction` modules and extracted, unchanged Broker/Egress function bodies. Both tenant-secret and connection-token refresh updated the broker credentials while leaving the new value absent from the registry; the actual redactor returned the echoed new value unchanged. Registering that value as a control masked it. Storage and broker-write boundaries were stubbed, so this diagnostic is not a live proxy or database integration test.

The coordinated conversation suite completed in 58.9 seconds on Elixir 1.19.2/OTP 28.3 with disposable PostgreSQL 18: **1,566 tests, zero failures, two Linux-only skips** (GNU trust-store tools and `/proc`/`setpriv`). Formatting, changelog, conflict and whitespace checks passed, and the isolated checkout is clean. [Exact-head CI passed](https://github.com/managoat/fountain/actions/runs/35131010719). No real credentials or live deployment resources were used.

BinaryBourbon (changes requested):
Independent adversarial correctness review by Codex — **REQUEST_CHANGES**

Reviewed all four changed files at `2603ae16c159a76da6f57d3d5281cea7833e9084` against `dd9c453763c18611a7832db61ee61ae8ef93772e`.

**[P1] Keep the redaction registry current when brokered credentials rotate.** `apps/fountain/lib/fountain/conversations/sprite_env.ex:124–129` registers only the provisioning/reattachment snapshot passed by `conversation_server.ex:1017`. A supported later turn calls `Egress.refresh_before_turn/1` (`conversation_server.ex:1893`), which rereads edited environment/vault secrets and rotated connection/ChatGPT tokens and rewrites the broker's live rules; it never calls `SpriteEnv.build/4` or updates `Redaction`. For example, create a brokered conversation with a bound token, replace that token in its vault, then run another request to an endpoint that echoes it. The broker now injects the new value while the registry still contains the old value, so the new credential is persisted and streamed in plaintext—the same disclosure this PR promises to close. Register refreshed values before they become usable by the broker, retaining previously registered values that can still appear in output. Add a regression that builds the initial registry, performs an actual refresh, and checks redaction of the refreshed value.

Coverage: traced both fresh provisioning and wake through the shared builder; broker map shape and environment/vault precedence; provider input filtering and unchanged sandbox exports; both OAuth refusal/fallback branches; registry replacement, logging and termination; and the added tests and changelog. The new tests cover initial registration and absence from the sandbox, but never change a credential after registration. Current bound inference sources take the earlier refusal branch; I am not treating the legacy fallback branch's hypothetical old-token loss as another blocker without a supported current trigger.

Validation: independently inspected the security reviewer's fake-only diagnostic using the exact head `SpriteEnv`/`Redaction` modules and source-extracted `Broker`/`Egress` function bodies. Both a tenant-secret edit and a connection-token rotation put the new credential into the refreshed state and captured broker rules while leaving it absent from the registry; the echoed new value remained plaintext. A control that registered the refreshed value masked it. Persistence and broker boundaries were stubbed, so this is not a database/proxy integration test. Exact-head CI run `35131010719` passed (verified by the coordinating reviewer). The coordinating reviewer ran the conversation suite at the exact head with pinned Elixir 1.19.2/OTP 28.3 and disposable PostgreSQL 18: 1,566 tests, 0 failures, 2 Linux-only skips; exit 0. Format, changelog, conflict-marker and whitespace checks passed. No real credentials, public broker or deployment were used.

jhgaylor:
Fixed in `cdaa0d7ca`. All three reviews found the same P1 and it was real: the
first commit closed the disclosure for the credentials a conversation starts
with, and left it open for every one after. Thank you for the rotation probe —
the passing suite had no way to see it.

**Registered where the rotation happens.** `Egress.refresh_before_turn/1` now
registers what `reread_secrets/1` returns **before** either path hands it to the
broker. That single point covers both outcomes you named — the live-rule rewrite
and the replacement session — so there is no window in which the broker holds a
value the registry does not. It is in `egress.ex` rather than the server, which
is also where the values are adopted.

**Retaining old values in flight.** Registration is append-only now, through a
new `Redaction.add/2`. `put/2` replaces — and, worth noting, deletes the entire
registration when given nothing redactable — which is right for a first
registration and wrong for every one after. A conversation now keeps redacting
every credential it has held until it ends and `delete/1` clears it.
Over-redacting a retired value costs nothing; un-redacting one still in flight is
the disclosure.

I switched both existing writers to `add/2` as well: provisioning in
`SpriteEnv.build/4`, and the Claude API-key fallback. A wake or reattach rebuilds
the env, and with `put/2` that rebuild could itself have dropped a value still in
flight — the same class of gap by a different door.

`add/2` reads then writes. That is safe because every write for a conversation
happens inside its own server process: the only code callers of both
`SpriteEnv.build/4` and `refresh_before_turn/1` are in `ConversationServer`. I
checked rather than assumed, and it is stated in the `@doc`.

**The regression** is the one you asked for: real vault rows, a real edit of the
bound secret, the real `refresh_before_turn/1`, with `Broker` stubbed. The key
assertion is made **inside** the stubbed `Broker.refresh/4` and `Broker.prepare/4`
— the rotated value must already be registered at the moment the broker is asked
to inject it. Both outcomes also assert the new echo is scrubbed, the old one
still is, and the sandbox env holds only the placeholder. A third case pins that
an unchanged refresh leaves the registry exactly as it was.

Checked by breaking it both ways:

- removing the registration fails both outcomes with *"the broker must not be
  able to inject a value the registry does not know"*;
- using `put/2` instead of `add/2` fails both on the old credential no longer
  being scrubbed.

Unit tests for `add/2` cover retention, a first registration still skipping short
values, and — the case `put/2` gets wrong — a call with nothing redactable leaving
an existing registration intact.

To your note on scope: agreed that the registry protects persisted and streamed
output and does not filter what the sandbox runtime receives from upstream, which
ADR 0019 documents. The changelog describes this as closing the output disclosure,
not as preventing the agent from seeing an echo at runtime.

Checks: 1572 conversation tests pass, up from 1566; `sandbox_files_test.exs`
passes; the server is 2212 lines, down from 2216, with the pin untouched.

jhgaylor:
**Correction to my previous comment.** I wrote that the changelog describes this as closing the output disclosure *"not as preventing the agent from seeing an echo at runtime."* That was false: the fragment led with exposure to the agent as the harm, inside an entry under *Fixed*, which reads as exactly that claim. I asserted it without re-reading the file.

You were right that the registry protects persisted and streamed output and does not filter what the sandbox runtime receives. ADR 0019 §10 makes it concrete — the broker does not rewrite responses — so an upstream echo reaches the agent unmodified, before Fountain persists anything.

Corrected in `8c3666867` and in the PR description. Both now say what this fixes (plaintext at rest in `log_events`; exposure through history and the stream, including after rotation) and, separately, what it does not and cannot change (the agent still sees an echo). The changelog also gives operators the one mitigation available today: bind a secret only to hosts that do not return it.

No code changed in this correction.

BinaryBourbon (approved):
Independent adversarial correctness re-review by Codex — **APPROVE**

Reviewed the whole eight-file PR at `8c366686762e2b812e6b611d6e2887a5ea55f637` against `dd9c453763c18611a7832db61ee61ae8ef93772e`, including the fixes since my previous review. No new blocking correctness finding.

The prior credential-rotation finding is resolved: `Egress.refresh_before_turn/1` now registers the refreshed broker map before either rewriting live rules or preparing a replacement session. `Redaction.add/2` preserves earlier values, including when the environment is rebuilt or the fallback branch registers more values. Production registration calls run synchronously in the owning conversation server; separate servers on separate nodes do not share the node-local ETS table. Sandbox exports and environment/vault precedence remain unchanged.

Coverage includes `ConversationServer`, `Egress`, `Redaction`, `SpriteEnv`, all three changed test files and the changelog. I independently inspected the security reviewer's exact-source diagnostic: tenant-secret edits and connection-token rotations each pass both rewrite and replacement-session paths, with registration asserted before broker I/O, old and new echoes masked, and real values absent from the returned environment. The added repository tests also assert ordering, old-value retention and unchanged-refresh behavior.

Scope note: retained history lasts for the server's lifetime. The unchanged `terminate/2` clears the registry on shutdown or suspension; a new actor reconstructs current credentials, not historical rotated values. A saved or replayed old echo can therefore fall outside redaction after a restart. This is an inherited lifecycle limitation, not a regression or a blocker to the initial/per-turn fix. The changelog correctly distinguishes persisted/history/stream redaction from what the runtime sees inside the sandbox.

Validation: source review complete; exact-head CI run `35134284632` passed (verified by the coordinating reviewer). The coordinating reviewer ran the full conversation suite with pinned Elixir 1.19.2/OTP 28.3 and disposable PostgreSQL 18: 1,572 tests, 0 failures, 2 existing Linux-only skips. I inspected the mutation logs: removing refresh registration made both new broker-path tests fail before broker I/O; replacing additive registration with replacement made both fail on the delayed old echo. After restoring exact source, the three changed test files passed all 52 tests and the checkout was clean. Format and changelog checks passed. The focused diagnostic stubs persistence, broker and runtime boundaries; no real credentials, deployment, public broker or live sandbox were used.

BinaryBourbon (approved):

Reviewed all eight changed paths at `8c366686762e2b812e6b611d6e2887a5ea55f637` against base `dd9c453763c18611a7832db61ee61ae8ef93772e`: the original server/environment registration, refresh ordering, additive registry, all three changed test files and changelog. I also traced the broker credential sources, sandbox delivery, persistence and historical/live SSE consumers. No concrete introduced security or tenant-isolation defect remains in this review.

My previous P1 finding is resolved. `Egress.refresh_before_turn/1` now registers the refreshed broker values before either updating live rules or preparing a replacement session. `Redaction.add/2` preserves previously registered values; environment rebuilding and fallback registration use the same additive operation. Raw broker values are registered separately from the environment returned to the sandbox.

I reran the previous reproducer against the new head; its assertion that the fresh credential was absent now fails because the value is registered. The updated pinned Elixir 1.19.2/OTP 28.3 diagnostic passed four cases: tenant-secret and connection-token rotation, each through a live-rule rewrite and a replacement session. Assertions inside the broker boundary confirmed the original and replacement values were already registered before that boundary received the credentials. Both the fresh echo and delayed original echo were masked; rebuilding the environment preserved both registrations; raw credentials remained absent from the returned environment; and another conversation's registry remained separate.

The source retains tenant-scoped credential reads and per-tenant decryption. Registration remains keyed by the established conversation identity, and event history/stream ownership checks are unchanged. `Conversations.log!/1` scrubs before persistence, and the persisted event is what the live broadcast and historical stream consume. The revised changelog correctly distinguishes transcript protection from the unmodified upstream response seen inside the sandbox.

One existing limit remains: the registry is an in-memory ETS table, and a conversation's entry is cleared when its server terminates. Retention of earlier values across registrations within that actor is verified here; durable retention across actor restarts is not implemented by this patch.

I inspected the coordinating reviewer's validation logs: the full conversation suite passed 1,572 tests with zero failures and two existing Linux-only skips under Elixir 1.19.2/OTP 28.3 and disposable PostgreSQL 18. Formatting and changelog checks passed. Removing refresh registration made both before-broker assertions fail; replacing `add` with `put` made both delayed-old-echo assertions fail. After restoring the exact source, all 52 tests in the three changed test files passed and the checkout was clean. [Exact-head CI passed](https://github.com/managoat/fountain/actions/runs/35134284632).

Source review and the described diagnostics are complete. The focused probe compiled exact head `SpriteEnv` and `Redaction` modules and exact broker-split/egress-refresh function bodies, with fake storage, broker-session and runtime boundaries. It was not a database-backed or sandbox/proxy integration test. No real credentials, deployment or provider resource was accessed. This is an independent Codex review, not a Review Loop App verdict.

BinaryBourbon (approved):
**APPROVE** — independent adversarial API/operator re-review by Codex of the whole PR at `8c366686762e2b812e6b611d6e2887a5ea55f637`, against `dd9c453763c18611a7832db61ee61ae8ef93772e`. My previous P1 finding is resolved; no current blocking defect found.

`Egress.refresh_before_turn/1` now registers refreshed broker credentials before either rewriting live rules or minting a replacement session. The additive registry preserves old values for later output within the current conversation server's registry lifetime, and environment rebuilds and the fallback handler use the same operation. I traced the affected callers through the existing log writer and history/SSE consumers. The changelog now correctly distinguishes transcript protection from the runtime/model receiving an upstream echo inside the sandbox.

Coverage: all eight changed paths — `conversation_server.ex`, `egress.ex`, `redaction.ex`, `sprite_env.ex`, their three changed test files and the changelog — including broker value types, initialization, refresh ordering, rebuilds, fallback/source binding, registry lifecycle and output compatibility. `SpriteEnv.build/4` retains its arity, optional inputs and returned environment ordering. Real broker credentials remain outside the sandbox environment; no endpoint or SDK wire schema change is needed.

My local compatibility diagnostic compared the exact base/head builder bodies in four cases covering omitted options, plain/secret values, inference-input filtering and CA/proxy precedence: returned environments were identical. With the actual new redaction module, broker values registered without entering the returned environment, repeated builds retained both old/new redaction, short-value additions preserved existing entries, and deletion cleared them. Runtime/metadata dependencies were deterministic stubs.

I also inspected the coordinated exact-source rotation diagnostic: tenant-secret and connection-token rotation both pass through live-rule rewrite and replacement-session paths, with registration checked before broker I/O, old/new echoes masked, rebuilt environments retaining both registrations, and separate conversations isolated. That diagnostic stubs storage and broker-write boundaries; it is not a live proxy/deployment test. [Exact-head CI passed](https://github.com/managoat/fountain/actions/runs/35134284632). No real credentials or live deployment resources were used.

The coordinating reviewer ran the full conversation suite on Elixir 1.19.2/OTP 28.3 with disposable PostgreSQL 18: **1,572 tests, zero failures, two existing Linux-only skips**. I inspected those results and both isolated mutation logs: removing refresh registration caused the two ordering tests to fail; replacing `add` with `put` caused the two late-old-value assertions to fail. After restoring the exact source, the three changed test files passed all 52 tests and the checkout was clean. Formatting and changelog checks also passed.

Retention limit: unchanged `terminate/2` clears the ETS registration when the actor stops. A new actor reconstructs it from currently available credentials, so this does not retain every historical value across server restarts, suspend/wake or deployment reattachment. That is an inherited limitation, not a new regression or the unresolved per-turn rotation finding.

PR #2132, opened 2026-09-13
fix(broker): withhold credential-bearing paths from egress records

jhgaylor:
Egress logging recorded URL path credentials in both Logger and `broker_requests`, including Telegram tokens in the first `/botTOKEN` segment and arbitrary signed segments. Replace the entire path with `/[REDACTED]` before logging and before buffering database writes. Method, host, outcome, matched binding, credential names, status, duration and terminal error remain available. The existing public `path` string field is preserved; its schema description and generated TypeScript comment now describe redaction.

Legacy rows are also redacted when returned by the egress API. This change does not rewrite or delete historical data: existing database paths remain until the normal `BROKER_LOG_RETENTION_HOURS` sweep removes them, and historical server logs follow their storage system's retention. The documentation records this limit. The pinned broker's source emits a transient pre-substitution path telemetry field but does not write that path to Logger; no upstream logging fix was needed.

Real proxy regressions send unbrokered Telegram first-segment credentials and arbitrary signed path/query values. They verify unchanged origin requests, safe Logger output and raw stored rows. Additional tests check redaction before buffering and read-time redaction of legacy rows without rewriting them.

Validation: 65 focused broker/API tests and 29 docs tests passed; compilation with warnings as errors, formatting, strict Credo, unchanged wire-contract verification, TypeScript regeneration, typecheck and SDK contract verification passed. The repository-wide prose checker reports only three existing warnings in untouched files (docs/api.md:128, docs/catalog/runtimes/codex.md:62 and docs/concepts/environment.md:73). Full precommit and CI remain for the draft PR.

SDK release: `sdk-no-release` is intentional: the only SDK change is a generated documentation comment; runtime code and TypeScript types are unchanged, so no package version or release is needed.

Closes #1712.

PR #2358, opened 2026-09-16
fix(deployed): let the secrets profile pass against a real deployment

jhgaylor:
The deployed suite's `secrets` profile had **never passed** against a real
deployment. Its first failure was a genuine disclosure — fixed in #2355 — and
that failure stopped the run before anything after it could execute. With the
leak gone, three later checks surfaced one at a time. Each asserted something
the deployment is correct not to provide.

After this change the profile passes against production for the first time.

## The three stale checks

**A refused tunnel is exit 7, not 56.** The script required curl exit 56 for
the blocked destination. Captured from production, the broker refused exactly
as intended:

```
rc=7 connect='403' stderr='curl: (7) CONNECT tunnel failed, response 403'
```

Curl reports a refused CONNECT as exit 7 where older curl reported 56. The
proxy's `403` is the signal; both exit codes are accepted, and nothing else is.

**The placeholder cannot appear unredacted.** Fountain registers every sandbox
environment value for redaction, and the binding's placeholder is one of them,
so durable output reads `fixture-placeholder=[REDACTED]`. Requiring the plain
form was unsatisfiable. Substitution is still proven twice: the script raises
before printing its completion marker if the variable held anything but the
placeholder, and the receiver records `placeholder_matches`.

**The broker no longer stores paths.** The allowed egress row was located by
`path === '/capture/<run>/allowed/<nonce>'`. Since #2132 every stored path is
`/[REDACTED]`, because a path can carry a token — so the match could never
succeed, and the check polled for 20s and aborted. It now matches the run's own
receiver hostname and method, within the run's own conversation; the nonce is
proven by the receiver's `nonce_matches`.

None of the replaced evidence was unproven elsewhere. The durable-output
assertion also names the missing marker now, instead of one combined message
that made the failing one findable only by searching the traces by hand.

## Verification

- `node --test deployed/test/*.test.mjs` — 293 pass.
- The embedded Python compiles (`py_compile` on the generated script).
- Against production on `sha-da3c7d68f`: **exit 0, 9 of 9 checks, zero
  resources remaining.** Identity, capabilities, setup, fixtures, provision,
  sandbox delivery and output, receiver and egress evidence, receiver cleanup,
  and non-disclosure.

One run in between failed at setup with `EHOSTUNREACH` on a Cloudflare quick
tunnel and was retried. That is the known best-effort tunnel behaviour, reported
as setup rather than as a verdict, and says nothing about the deployment.

## Found along the way, not fixed here

While establishing which marker was missing, the placeholder turned up
**unredacted** in one earlier run's durable history — inside a single streamed
model-output chunk whose text ended mid-value: `"\n- fixture-placeholder=__suite_"`.
Redaction runs per persisted event, so a registered value split across two
chunks is matched in neither. Tracked separately, starting with a reproduction
using a real synthetic secret rather than a placeholder.


BinaryBourbon (approved):
Adversarial correctness and test review: **APPROVED** at `9633fa7fed8ea92a87dbac5d59ba89079165bab0`.

Reviewed all three changed files: `deployed/profiles/secrets.mjs`, `deployed/test/secrets.test.mjs`, and `changelog.d/1614-secrets-profile-harness.md`, against merge-base `da3c7d68fb1e7c36332da330d2ebc1dbebc2dd79`. No actionable correctness defects found.

The receiver still independently requires the exact run nonce, placeholder and both vault values; one successful credential-bearing POST and denied CONNECT evidence remain required. Matching the conversation-scoped egress row without its path agrees with the broker's unconditional path redaction.

Validation: executed the generated Python script with mocked curl responses. Exit 7/403 and 56/403 pass; exit 0/403, 6/403, 7/000, 56/200 and malformed 403 output fail. An incorrect placeholder fails before any request. The coordinating reviewer also ran the complete deployed unit suite: 293 passed, zero failures or skips. No live deployment was exercised, and this manual review does not claim Review Loop service verification.

BinaryBourbon (approved):
Adversarial security review: APPROVED for `9633fa7fed8ea92a87dbac5d59ba89079165bab0`.

Reviewed all three PR files: `deployed/profiles/secrets.mjs`, `deployed/test/secrets.test.mjs`, and `changelog.d/1614-secrets-profile-harness.md`, plus the receiver/session, raw HTTP/SSE inspection, transcript checks, environment redaction, and tenant-scoped egress implementation. No introduced security blocker found.

The removed placeholder assertion has independent coverage: the generated script rejects an incorrect sandbox value, and the receiver checks the actual placeholder, nonce, and both vault-secret fingerprints. The relaxed egress lookup still requires a single matching POST in the new conversation, status 200, the run-specific binding key, independent receiver evidence, and a credential-free CONNECT denial. Raw secret-disclosure checks remain before harness redaction.

Validation: 11 offline generated-Python cases passed, accepting only curl exit 7/56 with CONNECT 403 and rejecting wrong codes/statuses, cleartext bound values, incorrect placeholders, missing plain values, and a wrong receiver nonce. The coordinating reviewer also ran the deployed suite: 293/293 passed. No production replay or full Review Loop service verification was performed.

Nonblocking documentation follow-up: `deployed/README.md:484` still describes curl exit 56 as mandatory; align it with the newly accepted 7/56 outcomes.

BinaryBourbon (approved):
Independent adversarial review: product/API compatibility.

**Verdict: APPROVED** for `9633fa7fed8ea92a87dbac5d59ba89079165bab0`. No blocking findings; source review complete.

Reviewed all three changed files: `deployed/profiles/secrets.mjs`, `deployed/test/secrets.test.mjs`, and `changelog.d/1614-secrets-profile-harness.md`. Traced the changed assertions through the generated Python, turn/history collection, receiver registration and receipts, fixture ownership, actual egress controller/query/logging, and sandbox environment redaction. Compared the behavior with issue #1614, merged #2132, and the accepted base's ADR 0019/API review guidance.

The relaxed assertions preserve the relevant proof: denial still requires a failed CONNECT with status 403; the script and independently recorded receiver observation both check the exact placeholder; and egress remains scoped to the new conversation, expected hostname/method, exactly one successful credential-bearing request, and independent nonce/receipt evidence. No server or SDK wire contract changes are introduced.

Local diagnostic: compiled the generated Python and executed eight synthetic cases without network access. `(7, 403)` and `(56, 403)` complete; success, unrelated exit/status combinations, and invalid environment delivery are rejected. The coordinator separately ran `node --test deployed/test/*.test.mjs`: 293 passed, zero failures/skips. These checks do not constitute a new deployed smoke run or completion of the Review Loop service's verification recipe.

Nonblocking documentation correction: `deployed/README.md:484` still says the blocked request must return curl exit 56. Update it to describe CONNECT 403 with exit 7 or 56 so operators do not diagnose an expected passing result as a failure.

BinaryBourbon (approved):
Independent adversarial review — simplicity and maintainability

Verdict: APPROVED for `9633fa7fed8ea92a87dbac5d59ba89079165bab0` (target base `6d206bdd1a409075430613255cccbde23d32e006`).

Reviewed all three changed files: `deployed/profiles/secrets.mjs`, `deployed/test/secrets.test.mjs`, and `changelog.d/1614-secrets-profile-harness.md`, plus the receiver, durable-turn/history flow, broker request log, sandbox environment redaction, and trusted ADR 0019/contributor requirements.

No blocking defect found. Accepting curl exits 7 and 56 still requires CONNECT 403. Removing the impossible plaintext-placeholder assertion preserves independent receiver proof of the exact placeholder, vault precedence and bound/unbound delivery. Selecting the egress request by this conversation's receiver hostname and POST remains constrained by exactly one successful row, the expected binding key, the independently checked nonce/receipt, and no blocked receiver arrival. The changes stay local without introducing a second source of truth or new abstraction.

Nonblocking documentation correction: `deployed/README.md:484` still says the blocked request must have curl exit 56; update it to CONNECT 403 with exit 7 or 56 to match the changed condition at `deployed/profiles/secrets.mjs:50`.

Validation: independent source review completed; the coordinating reviewer reports `node --test deployed/test/*.test.mjs` passed all 293 tests, with no failures or skips. The modified test pins generated script text; it is not a live deployment replay. No production resources were used and this is not a Review Loop service verification.

PR #2360, opened 2026-09-16
fix(deployed): let the webhooks profile pass against a real deployment

jhgaylor:
The deployed suite's `webhooks` profile had never passed against a real
deployment. Two checks were stale, and the first hid the second. After this
change it passes against production — 10 of 10 checks.

## The receiver refused every real delivery

The first live run: `signature_valid: true`, `payload_valid: false`, status 422.
Fountain signed correctly; the receiver rejected the envelope.

It checks the envelope's exact key set, and #1637 added `labels` to the data
Fountain signs. Every delivery was refused, so the controlled 503 and the retry
it exists to exercise never happened.

**The exact key set stays.** `Fountain.Webhooks.payload/3` promises *"ids, a
stage, a status, a duration, the labels. Nothing else, ever."* A receiver that
accepted any added field would let content ride in a webhook unnoticed, which is
the property worth verifying. So `labels` is expected by name and bounded as
`Fountain.Conversations.Labels` bounds it — a plain object, at most 32 entries,
string values, keys 1–64 bytes, values at most 256 — and a test pins that a
field the envelope does not promise is still refused.

The test fixture had no `labels` either. It was written to match the receiver
rather than the payload Fountain sends, which is how a receiver refusing every
real delivery still passed its tests. It now carries the envelope as `payload/3`
builds it; reverting the receiver's key set fails three tests.

## The payload was compared against a stream frame

With deliveries accepted, the strict comparison failed on `duration_ms`:
`undefined` expected, `null` delivered. The event came from
`GET /api/conversations/:id/stream`, and `StreamLogEvent` documents that stream
as **not sending `duration_ms`**. The comparison now uses the durable event from
history, which serves the whole event, after asserting it matches the streamed
one.

`?? null` would have passed this run and been wrong: a terminate event that ever
carried a real duration would then fail for the wrong reason.

The expected payload also now carries the conversation's own labels, so a
webhook carrying different labels is detected as drift.

## Verification

- `node --test deployed/test/*.test.mjs` — 302 pass, up from 293.
- Reverting the receiver to its pre-`labels` key set fails three tests,
  including the new envelope test.
- Against production on `sha-da3c7d68f`: **exit 0, 10 of 10 checks, zero
  resources remaining** — identity, capabilities, setup, conversation,
  registration, event, automatic retry, duplicate observation, receiver cleanup,
  non-disclosure. `duplicate-observation` had never run before.


BinaryBourbon (approved):
Adversarial correctness and tests review — APPROVED at `eebaecc0d924b4674d97aa09ea0faea6864bdd73` (base `7725e73124d02814dbcc2344f98ba4d8a918365e`). No concrete introduced blocking defects found.

Reviewed all changed paths: `changelog.d/1616-webhooks-profile-harness.md`, `deployed/profiles/webhooks.mjs`, `deployed/receivers/webhooks.mjs`, and `deployed/test/webhooks.test.mjs`. Traced them through Fountain's webhook payload builder, conversation label validation/rendering, event persistence, REST history pagination, SSE serialization, and the profile's zero-inference fixture lifecycle.

The durable lookup supplies the duration omitted by the conversation stream; event identity/timestamp checks and exact label comparison retain independent evidence. Five isolated probes invoking the full profile passed: null, zero, and positive durable durations with no stream duration; mismatched timestamps and absent durable events correctly failed. These probes stubbed all network access. The coordinating reviewer also ran the complete deployed suite: 302 tests passed, no failures or skips.

This is an independent source review with local fixture validation, not a deployed-environment result or a claim that the full Review Loop verification service ran.

BinaryBourbon (approved):
Adversarial security review: **APPROVED** at `eebaecc0d924b4674d97aa09ea0faea6864bdd73` (base `7725e73124d02814dbcc2344f98ba4d8a918365e`).

Reviewed all four changed files and traced labels through Fountain's validator/envelope, raw-byte HMAC verification, authenticated receiver evidence, exact conversation-label comparison, and tenant-scoped durable history/stream access. No introduced security blocker found. Labels remain inert JSON; invalid signatures and unexpected envelope fields do not retain their payloads. The added history request uses the existing conversation ownership check before reading log events.

Validation: 12 additional local handler assertions passed, including UTF-8 byte boundaries, wrong admin identity, unsigned-content non-retention, unexpected content fields, and `__proto__` as an inert label. The coordinating review also ran all deployed tests at this head: 302 passed. Server tenant-denial regression assertions were inspected, not executed; no production replay was performed.

Nonblocking contract nit: `deployed/receivers/webhooks.mjs:33-37` accepts a NUL in a label key/value, whereas `Fountain.Conversations.Labels.check_entry/2` rejects both. A signed local NUL fixture is accepted. Reject `\u0000` in both strings and add corresponding cases to match the documented validator exactly. The profile's comparison against the conversation's actual labels still detects such drift, so this is not an approval bypass.

BinaryBourbon (approved):
Adversarial product/API compatibility review of `eebaecc0d924b4674d97aa09ea0faea6864bdd73`: **APPROVED**.

Reviewed all four changed files: `deployed/profiles/webhooks.mjs`, `deployed/receivers/webhooks.mjs`, `deployed/test/webhooks.test.mjs`, and `changelog.d/1616-webhooks-profile-harness.md`. Traced the actual webhook envelope, label normalization and byte/count limits, conversation JSON, SSE rendering, durable history ordering, and this profile's fresh zero-turn lifecycle. Labels now agree with the published envelope, and the durable record supplies the duration omitted from conversation SSE. Existing API/SDK contracts are unchanged. No concrete introduced compatibility defect found.

Validation: the coordinated review run passed all 302 deployed tests at this head, plus changelog, conflict-marker, and whitespace checks. No deployed replay was performed. This is an independent manual review, not Review Loop service verification.

BinaryBourbon (approved):
Adversarial simplicity and maintainability review: **APPROVED** at `eebaecc0d924b4674d97aa09ea0faea6864bdd73` (base `7725e73124d02814dbcc2344f98ba4d8a918365e`).

Reviewed the complete triple-dot diff across all four files: `changelog.d/1616-webhooks-profile-harness.md`, `deployed/profiles/webhooks.mjs`, `deployed/receivers/webhooks.mjs`, and `deployed/test/webhooks.test.mjs`, using the base revision's contributor guidance, simplicity policy, and webhook ADR.

No blocking simplicity or maintainability findings. The change keeps the receiver's explicit envelope boundary, adds a small local label validator, and verifies labels against the public conversation. Reading the matched durable event directly supplies `duration_ms` while retaining the SSE identity, stage, state, and timestamp checks. I traced these choices through the actual webhook builder, label rules, event persistence, REST/SSE serializers, and profile callers. The tests exercise signed HTTP delivery acceptance/refusal and final evidence drift, without introducing a new abstraction or changing production behavior.

Validation: inspected the exact-head deployed suite log recorded by the coordinating reviewer: 302 passed, zero failed or skipped (`node --test deployed/test/*.test.mjs`). Source review is complete. No live deployment replay was performed; this manual review does not claim the full Review Loop service verification recipe ran.

PR #2363, opened 2026-09-16
fix(inference): skip an exhausted ChatGPT grant for the platform key (#2362)

### LINKED ISSUE #2362: An exhausted ChatGPT grant blocks Codex instead of falling back to the platform OpenAI key
Found on 2026-09-16 while running the deployed suite's provider/runtime matrix
against production (#1606).

## What happened

Production's platform ChatGPT grant (ADR 0047) exhausted its Codex usage quota:

```
codexErrorInfo: "usageLimitExceeded"
"You've hit your usage limit. Visit https://chatgpt.com/codex/settings/usage to
 purchase more credits or try again at Sep 20th, 2026 11:40 AM."
```

Every Codex turn that selected the platform credential failed that way.
`PLATFORM_OPENAI_API_KEY` was configured in production the whole time, but
nothing ever fell back to it.

Reconnecting the grant did not help either: the operator disconnected and
reconnected at 20:00:53–20:01:20 UTC, but it was the same ChatGPT account, and
usage limits belong to the account rather than the token — so the next turn
failed with the identical reset time. Only disconnecting the grant outright
moved Codex onto the API key.

## Why

`Fountain.PlatformInference.credential_for/2` (`platform_inference.ex`):

```elixir
def credential_for("openai" = provider, "codex") do
  if Fountain.Broker.configured?() do
    case Fountain.ChatGPTAccounts.platform_credential(refresh: false) do
      {:ok, token} -> {:ok, :codex_chatgpt_access_token, token}
      :none -> key_for(provider)
    end
```

The grant wins whenever one is **active**, and the fall-through to the platform
key happens only on `:none`. An exhausted grant is still active — its token is
valid; it is the account's quota that is spent — so selection never reaches the
key. This is ADR 0047 decision 6 working as written. The gap is that the rule
treats "has a credential" and "can serve a turn" as the same thing, and quota
exhaustion is the case where they differ.

## Why this is a design question, not a patch

- **It changes billing mid-flight.** The grant is subscription-priced; the
  platform key is metered per token, pass-through (#1388). Falling back silently
  moves an operator from one to the other. That might be exactly right, or it
  might be a surprise invoice — it wants to be a decision.
- **The signal arrives late.** `usageLimitExceeded` comes back from the provider
  inside a turn, after selection. A fallback needs either to remember a recent
  exhaustion and skip the grant for new turns, or to retry the failed turn on
  the key — the second touches turn-retry semantics.
- **It amends an ADR.** ADR 0047 decision 6 is the rule being changed.

## Options to weigh

- **Skip a recently exhausted grant.** Record `usageLimitExceeded` with its reset
  time; `credential_for/2` skips the grant for new turns until then and uses the
  key. The in-flight turn still fails, and nothing is retried.
- **Make the preference explicit.** An operator setting: grant-only, grant with
  key fallback, or key-only. Makes the billing trade-off a stated choice.
- **Surface it and stop there.** Keep "grant wins", but raise a visible admin
  alert on exhaustion with the reset time, so the fix is a deliberate manual
  switch rather than an outage someone has to trace.

Related: ADR 0047, #1388, #1606.

jhgaylor:
Closes #2362

Implements the maintainer's chosen option, **skip a recently exhausted grant**, and amends ADR 0047 decision 6 in place.

## What happens now

- A codex turn on the platform ChatGPT grant fails with `codexErrorInfo: "usageLimitExceeded"`. **That turn still fails, and nothing retries it.**
- Fountain writes the provider's reset time to the grant row (`usage_exhausted_at`, `usage_exhausted_until`). The value is in the database, so every node sees it.
- Until the reset time, `PlatformInference.credential_for("openai", "codex")` skips the grant and returns the platform OpenAI key. `gate_source/1` gates the source that was actually selected, so the daily ceiling and per-token billing apply.
- If no platform OpenAI key is set, the grant is still selected. The turn then fails with the provider's reset time instead of failing with no credential.
- Nothing has to clear the skip: once the reset time passes, the grant is selected again.

## Design details

- **Where the error arrives.** codex-acp 1.10.0 answers `session/prompt` with a JSON-RPC internal error. Its `data` holds `codexErrorInfo` and the provider's message. I checked this in the published bundle (`createTurnErrorData`). `Managoat.ACP.Peer` reports it as `{:failed, {:acp_error, :prompt, error}}`, and a new `TurnMachine` clause handles it.
- **Reset time.** There is no structured reset time on this path. The adapter keeps the rate-limit `resetsAt` in its own session state and never sends it over ACP. So `PlatformChatGPT.UsageLimit` parses the message. It reads "Sep 20th, 2026 11:40 AM", or a bare "11:40 AM" meaning today, or tomorrow if that time has passed. The time is read as UTC, on the assumption that sandboxes run in UTC. If the time is unreadable, already past, or more than 8 days away, the skip lasts 1 hour.
- **Recorded only for the platform grant.** The bound source must have scope `:platform` and kind `:codex_chatgpt_access_token`. The write is fenced on the grant id and generation from the source's identity, and runs under the platform source lock. If a tenant's own credential hits its limit, nothing is recorded.
- **Audit.** An `admin.platform_chatgpt.exhausted` event records the account id, kind, `until`, and whether the reset time came from the provider or the default. It is written after the update, outside the transaction, only when the row changed, and never includes a token. Turns that fail together leave one event.
- **Reconnect.** Reconnecting the same `account_id` keeps the exhaustion, because usage limits belong to the account (this is what happened in production). Connecting a different account clears it. Disconnecting deletes the row.
- **Conversations already bound to the grant** stay bound, per ADR 0053. Waking or provisioning one during the window returns `inference_source_changed`. A reconnect already gives the same answer. This is written down in the ADR and the docs.
- **Admin page.** `/admin/inference` shows a "usage limit" badge and "hit its Codex usage limit until … UTC", plus a line about the fallback.

## Tests

- New `test/fountain/platform_chatgpt_exhaustion_test.exs`:
  - parser cases
  - `credential_for` before and after the reset, with and without a key
  - `resolve` and `gate_source` agree
  - recording is fenced, audited once, and ignores non-grant sources
  - reconnect with the same account vs a different account
  - `TurnMachine` records only for a grant-bound turn, and the turn still fails
- New admin LiveView test for the exhausted state.
- **Revert checks:**
  - Removing the skip in `credential_for` made the selection test fail.
  - Loosening the scope/kind guard in `TurnMachine` and `platform_record_exhausted` made 2 tests fail.
  - Removing the reconnect clearing made the reconnect test fail.
  - Each file was restored from a copy afterwards.
- `mix precommit`: passed, all 9 stages. Related files passed (platform_chatgpt, the admin inference pages, audit_live, turn_machine, platform_inference, conversation_server_platform_inference, docs): 199 + 29 tests, 0 failures.
- `scripts/decisions-index.sh`, `okf validate decisions` (valid) and `python3 scripts/changelog.py check` all pass.


apps/fountain/lib/fountain/conversations/turn_machine.ex
```diff
@@ -554,6 +554,25 @@ defmodule Fountain.Conversations.TurnMachine do
...
+  # would have, and nothing is retried; what changes is the next selection,
+  # which skips the grant until the reset. Recorded only when this turn ran
+  # on the platform grant: a tenant's own credential hitting its own limit
+  # says nothing about the deployment's account, and the context refuses
+  # any other source.
+  def handle(%__MODULE__{} = turn, {:failed, {:acp_error, :prompt, error} = reason}, ctx) do
+    with %Source{scope: :platform, kind: :codex_chatgpt_access_token} = source <-
+           Map.get(ctx, :inference),
+         {:ok, until, reset} <- Fountain.PlatformChatGPT.UsageLimit.exhaustion(error) do
+      Fountain.ChatGPTAccounts.platform_record_exhausted(source, until, reset)
```
BinaryBourbon: **[P1] Authenticate provider exhaustion before changing the shared grant**

This promotes tenant-controlled sandbox output into deployment-wide credential state. A verified tenant can select a codex agent with no personal key, and use its environment `setup_script` (`Provisioning.run_setup_script/4`) to place a fake `/home/sprite/.local/bin/codex-acp`. The pinned runtime installer accepts that executable when `--version` prints `1.10.0`; it does not authenticate its code or responses. The fake can complete `initialize`/`session/new` and return a `session/prompt` error containing `data.codexErrorInfo = "usageLimitExceeded"` and any reset within eight days, without contacting OpenAI.

The real `Managoat.ACP.Peer` forwards that as `{:failed, {:acp_error, :prompt, error}}` (confirmed in a local diagnostic). This clause then supplies the legitimate server-bound `Source` to `platform_record_exhausted`, so its ID/generation fence does not prevent the forgery. With a fallback key configured, one tenant can disable the shared grant for all new codex conversations, force other tenants onto metered inference, and cause grant-bound wake/provision attempts to fail with `inference_source_changed`.

Require a trusted provider response observed by the server/broker before writing platform-wide exhaustion, or keep unverified ACP reports scoped to their originating tenant/conversation. Add a two-tenant regression showing that a forged adapter error cannot change the other tenant's selected platform source.

BinaryBourbon (changes requested):
Request changes: one P1 tenant-isolation defect. A tenant-controlled ACP response can mark the deployment-wide ChatGPT grant exhausted, changing credential selection and billing for other tenants. The inline finding identifies the complete path and remedy.

Reviewed the entire PR at `ce06adc6ae48fe2230f19ec3ce79483948d89b9e` against `3ccd36a4f2c3bcfc9dd43a1f74431e2a77b980f6`, including credential resolution/binding, setup and adapter launch, lifecycle locking, audit, migration, tests, admin UI, and documentation. No other blocking defect found.

Validation: the two focused regression files (`platform_chatgpt_exhaustion_test.exs` and `admin_inference_chatgpt_live_test.exs`) passed locally: **24 tests, 0 failures**, on the pinned Elixir/OTP toolchain and a separately created/migrated local database. An initial attempt against the existing test database failed because its schema lacked the PR's new columns; the shared schema was left unchanged. A separate diagnostic using the real pinned `Managoat.ACP.Peer` confirmed that a fake adapter, with no provider call, produces the exact failure tuple consumed by the new handler. Exact-diff whitespace check passed. GitHub reports CI required, all six test partitions, static analysis, coverage, release/contract, and other selected checks successful for this head.

This is a manual adversarial review; it does not claim that Review Loop's separate four-reviewer verification recipe ran.

<details><summary>Changed-file coverage (14 files)</summary>

- `apps/fountain/lib/fountain/audit/admin_event.ex`
- `apps/fountain/lib/fountain/platform_chatgpt/account.ex`
- `apps/fountain/lib/fountain/platform_chatgpt/usage_limit.ex`
- `apps/fountain/lib/fountain_web/live/admin_live/inference.ex`
- `apps/fountain/priv/repo/migrations/20260916180000_add_usage_exhaustion_to_platform_chatgpt_account.exs`
- `apps/fountain/test/fountain/platform_chatgpt_exhaustion_test.exs`
- `apps/fountain/test/fountain_web/live/admin_inference_chatgpt_live_test.exs`
- `changelog.d/2362-exhausted-chatgpt-grant-falls-back.md`
- `decisions/0047-codex-platform-chatgpt-account.md`
- `decisions/index.md`
- `docs/configuration.md`

</details>

apps/fountain/lib/fountain/conversations/turn_machine.ex
BinaryBourbon: Fixed in 848e6df49. A sandbox's ACP report no longer writes shared state.

**What changed**
- **Hint only.** The `TurnMachine` clause checks `UsageLimit.hint?/1` and, on a platform-grant turn, calls `ChatGPTAccounts.platform_check_exhaustion/1`. That only starts a background check under `Fountain.TaskSupervisor`. It writes nothing.
- **Server-side confirmation.** `platform_confirm_exhausted/2` calls `GET https://chatgpt.com/backend-api/wham/usage` with the grant's own access token and `ChatGPT-Account-Id`. This is the endpoint Codex's own client reads rate limits from (`codex-rs/backend-client`, `rate_limit_status_url`). The account counts as limited when `rate_limit.allowed` is false, or when `limit_reached` is true and there are no credits. The exhaustion is recorded with the backend's `reset_at` (or `reset_after_seconds`) for the spent window. That reset is capped at 8 days, and a confirmed limit that gives no reset time is skipped for 1 hour. If the backend says not limited, the call fails, or the body can't be read, nothing is recorded.
- **Prose parsing removed.** The reset time is no longer parsed from the adapter's message. That also removes the old UTC guess.
- **Throttle.** A single fenced `UPDATE` of a new `usage_checked_at` column claims each check (grant id, generation, `active`, no check in the last 5 minutes). That allows at most one check at a time and one per cooldown across nodes. No check runs while an exhaustion is already recorded. The HTTP call runs outside any lock or transaction, as the refresher's does. The column is added in the same unmerged migration.
- ADR 0047's amendment, `docs/configuration.md` and the changelog fragment now describe the confirmation step, and the prose-reset claims are gone.

**Tests** (`apps/fountain/test/fountain/platform_chatgpt_exhaustion_test.exs`). `/wham/usage` is stubbed through the existing `Req.Test` seam (`:platform_chatgpt_req_options`); no test contacts OpenAI.
- Two-tenant regression: an attacker's grant-bound turn goes through `TurnMachine.handle/3` with the forged adapter error while `/wham/usage` says not limited. The victim's `resolve/4` still returns the same grant source with the same token, and no row or event is written. When the backend confirms the limit, the victim moves to the platform key and `gate_source/1` passes.
- Also covered: the throttle (5 hints make one call; 8 concurrent hints make one call), not-limited / 401 / unreadable responses recording nothing, non-grant sources and a replaced grant being ignored, and the request's bearer token and account-id headers.
- Revert checks: making confirmation always return "limited" failed 6 tests, including the forged two-tenant test. Removing the claim check failed both throttle tests. Both files were restored by copy.
- `mix precommit` passes all 9 stages. The related test files pass: 240 tests, 0 failures.

docs/configuration.md
```diff
@@ -246,6 +246,28 @@ sandbox holds a placeholder, and the egress broker puts the real token into
...
+ChatGPT for the account's usage itself, with the account's own token. It
+asks at most once every five minutes.
+
+If ChatGPT confirms the limit, Fountain records the reset time that ChatGPT
+gives and shows it at `/admin/inference`. It also records an
+`admin.platform_chatgpt.exhausted` event. Until the reset time, new codex
+conversations use `PLATFORM_OPENAI_API_KEY` instead of the account. Those
+turns are billed per token, and the daily ceiling applies. If no OpenAI
+platform key is set, codex conversations continue to use the account.
+After the reset time, the account is used again. If ChatGPT does not confirm
```
BinaryBourbon, line 262: **[P2] Document the fresh-sandbox requirement at both quota transitions**

These unqualified promises also appear in ADR 0047's amendment, but a new conversation does not necessarily mean a new sandbox. `Launch.home_or_new/5` automatically reuses a persistent home, and `InferenceBinding.compatible_machine/2` retains that machine's Codex `kind/identity/revision` even after every old conversation fails or is deleted (ADR 0053).

I reproduced both boundaries using the real launch context and a stubbed backend confirmation:

1. A new persistent conversation on a grant-bound home succeeds before confirmation; after confirmation selects the API key, the same launch returns `{:error, :codex_inference_conflict}`.
2. A home bound to the fallback key accepts new conversations during the skip; after `usage_exhausted_until` passes, new launches return `:codex_inference_conflict`, and validation of an existing fallback-bound conversation returns `:inference_source_changed` (so wake/provision also fails).

Thus starting a new conversation or waiting for the advertised reset does not provide the stated recovery on an existing home. The new tests only call `resolve/4` for the fallback, so they do not expose either admission failure.

Preserve the machine binding. At minimum, qualify this section, the ADR and the corresponding user/operator messages to explain that switching sources requires a **new compatible sandbox**, describe the persistent-home/attach errors and the reverse transition for fallback-bound conversations, and give the supported fresh-sandbox recovery path. Add an integration regression for both transitions through launch/source validation. Do not silently replace a persistent home or clear its binding to make the claim true.

BinaryBourbon (changes requested):
Request changes: one P2 operational-documentation defect remains. The new fallback/recovery claims omit the existing machine-lifetime Codex source binding: starting a new conversation on a persistent home can fail both when exhaustion is confirmed and when the reset passes. The inline finding includes the reproduced behavior and the minimum remedy; preserving ADR 0053's isolation guard is required.

The prior P1 tenant-forgery finding is addressed in this head. I independently traced the sandbox hint through the supervised task, server-side `/wham/usage` request, fenced cross-node claim, generation-fenced write, audit, and selection. A forged ACP report with a backend response that is not limited leaves the other tenant's selected source unchanged. No additional security blocker found.

Reviewed all 14 changed files at `848e6df49faffe5bda1875ce2917d78f0a8a46c3`, the prior-review fix delta, relevant source-resolution/admission/lifecycle/billing callers, and ADRs 0047/0053. The recorded base branch is `3ccd36a4f2c3bcfc9dd43a1f74431e2a77b980f6`; the actual PR diff uses merge-base `9d3cd19f9f846fad8c150f8bbd9819cd09361c11`.

Validation: **187 existing focused tests passed**, covering exhaustion, admin UI, grant lifecycle, platform inference, TurnMachine, conversation-server platform inference and source binding. **Two independent temporary integration diagnostics passed**, reproducing the persistent-home failures through `Conversations.Launch.start_conversation/2` and source validation. Pinned Elixir 1.19.2 / OTP 28, copied build directories, newly created/migrated isolated local database, mocked provider responses; no production credentials or shared database mutations. One initial diagnostic invocation used a nonexistent facade function and was corrected to the actual Launch context before reproduction. A mistaken companion-test path was corrected and that file was then run. Exact-PR-diff whitespace check passed. GitHub reports all executed selected checks successful, including all six partitions, static analysis, release/contract and coverage, for this unchanged head.

This is a manual adversarial review, not a claim that Review Loop's separate four-reviewer/full-verification recipe ran.

<details><summary>Changed-file coverage (14 files)</summary>

- `apps/fountain/lib/fountain/audit/admin_event.ex`
- `apps/fountain/lib/fountain/platform_chatgpt/account.ex`
- `apps/fountain/lib/fountain/platform_chatgpt/usage_limit.ex`
- `apps/fountain/lib/fountain_web/live/admin_live/inference.ex`
- `apps/fountain/priv/repo/migrations/20260916180000_add_usage_exhaustion_to_platform_chatgpt_account.exs`
- `apps/fountain/test/fountain/platform_chatgpt_exhaustion_test.exs`
- `apps/fountain/test/fountain_web/live/admin_inference_chatgpt_live_test.exs`
- `changelog.d/2362-exhausted-chatgpt-grant-falls-back.md`
- `decisions/0047-codex-platform-chatgpt-account.md`
- `decisions/index.md`
- `docs/configuration.md`

</details>

docs/configuration.md
BinaryBourbon: Addressed in a51f12c18. I kept the machine binding: no home is replaced and no binding is cleared. The changes document and test what the binding means at both switch points.

**Regressions.** The new `apps/fountain/test/fountain/conversations/platform_chatgpt_exhaustion_launch_test.exs` goes through the real `Launch.start_conversation/2` and `InferenceResolution.revalidate/3`, with `/wham/usage` stubbed:
1. **Grant to key, once the limit is confirmed.** A persistent launch onto a home bound to the grant succeeds before the limit is confirmed and returns `{:error, :codex_inference_conflict}` after. The home's `codex_inference_source` is unchanged, and the conversation already on the grant gets `{:error, :inference_source_changed}` when it is revalidated. An `ephemeral` launch gets a new sandbox and runs on the key.
2. **Key to grant, once `usage_exhausted_until` passes.** A persistent launch onto a key-bound home returns `:codex_inference_conflict`. The home keeps the key binding, and the key-bound conversation gets `:inference_source_changed` when it is revalidated. An `ephemeral` launch runs on the grant.

**Recovery path.** Both routes are in the existing code:
- A launch with `sandbox_mode: "ephemeral"` never lands on the home (`home_or_new("ephemeral", ...)` returns `:new`).
- `DELETE /api/sandboxes/{id}` (`Conversations.reset_sandbox/2`) retires the home. `_unsafe_find_home/4` skips terminated or failed homes, so the next persistent launch builds a new one.

**Messages.** The error codes are unchanged. Only the detail text changed:
- The `codex_inference_conflict` and `inference_source_changed` messages in `FallbackController` now name the platform ChatGPT usage-limit case and both recovery routes. A new controller test checks this.
- The exhaustion notice on `/admin/inference` now says that existing persistent homes and their conversations stay on the source they were bound to, both during the limit and after the reset.

**Docs.** `docs/configuration.md`, the ADR 0047 amendment and the changelog fragment now say that a switch applies only to new selections on a sandbox not bound to the old source. They describe both switch points and the recovery path, and the unqualified "new conversations use the key" wording is gone.

**Checks.** `mix precommit` passes all 9 stages. The focused files pass: the launch regressions, the exhaustion tests, the admin LiveView, the fallback controller, docs, launch credential sets, source binding, admission isolation, platform_chatgpt and turn_machine, 214 tests with 0 failures.

BinaryBourbon (approved):
Fresh adversarial review of the entire PR at `a51f12c182f3b35c82b2ba68eb03ee3cf38f3eaa`, including all 17 changed files and the resolver, launch, machine-binding, egress, and sandbox-reset callers. No remaining confirmed blocking defect found.

Both prior findings are addressed. A sandbox error only triggers a throttled, generation-fenced server-side usage check; the two-tenant regression verifies that an unconfirmed forged hint cannot change another tenant's source. The latest launch regressions cover both grant-to-key and key-to-grant transitions, preserve existing machine bindings, and verify fresh ephemeral launches. The documentation and error messages now describe the persistent-home conflicts and recovery paths.

Independent validation: migrated a new disposable local test database with the pinned OTP 28.3 / Elixir 1.19.2 toolchain, then ran 223 focused tests with 0 failures (exhaustion and launch regressions, grant lifecycle, turn machine, source binding/admission, controller, admin LiveView, and docs). All currently executed GitHub checks are successful; head and base were rechecked before posting.

This is the requested manual review, not a completed four-reviewer Review Loop run. I did not run the full local precommit/20-command verifier or contact a live ChatGPT account; provider responses in the tests are stubbed.

Issue #331, opened 2026-08-02
cost: log_events has no per-conversation volume bound inside the retention window


jhgaylor:
Every stdout/stderr chunk is persisted verbatim as one row (`conversation_server.ex:1325-1339`, `conversations.ex:665-680`). There is no per-turn or per-conversation byte cap anywhere.

The only bounds are indirect: the 24h max sandbox lifetime times whatever the sprite can print, pruned at 90 days (`retention_pruner.ex:39-45`). The #217 retention work is wired and working (`config.exs:16`, daily 04:23) but it bounds **age**, not **rate**.

A hostile or merely buggy agent (`while true; do cat /dev/urandom | base64; done`) can write tens of GB into Postgres per sandbox-day, at 5 sandboxes per user. Postgres is the same volume the app depends on, so this is an availability risk, not just a storage bill — and `log_events` was already 114MB of a 155MB database when #217 landed.

**Fix:** a per-conversation (or per-turn) byte budget that stops persisting and records a single truncation marker once exceeded. The live SSE stream can keep flowing; it's the durable copy that needs the cap.

PR #2364, opened 2026-09-16
fix(conversations): redact a registered value split across output chunks (#2359)

### LINKED ISSUE #2359: Redaction misses a registered value split across streamed model-output chunks
Found on 2026-09-16 while making the deployed suite's `secrets` profile pass
against production (#2358). **Not yet shown with a real secret** — the first
work here is to establish whether it is exploitable, not to design a fix.

## What was observed

`Fountain.Conversations.Redaction` scrubs registered values when a sandbox's
output is persisted, in `Conversations.log!/1`. It runs **per persisted
event**.

A model's reply streams as many small `agent_message_chunk` events, each
persisted on its own. In one production run, the binding placeholder — a
registered value, redacted everywhere else in the same conversation — appeared
**unredacted**, inside one chunk whose text ended partway through the value:

```
"\n- fixture-placeholder=__suite_"
```

The rest of the value was in the next chunk. Neither chunk contained the whole
registered string, so neither matched, and both fragments were persisted and
streamed in plaintext. In the same conversation's history the placeholder
appeared redacted three times (whole, inside tool output) and unredacted twice
(split, inside model text). A later run redacted every occurrence, because the
model happened not to split it — so this depends on where the stream breaks.

The evidence, from that run's retained traces:

- `GET /api/conversations/:id/events`, `data[23]` — an `agent_message_chunk`
  whose `blocks[0].body` is the fragment above.
- `GET /api/conversations/:id/stream` — the same fragment in a live `output`
  frame.

## Why it might matter

A placeholder is not secret, so this observation is harmless in itself. The
mechanism is not specific to placeholders.

#2355 established that an agent **still sees** a brokered credential an
upstream echoes back: the echo arrives inside the sandbox, before Fountain
persists anything, and ADR 0019 §10 records that the broker does not rewrite
responses. If the model then repeats that credential **in its own reply**, the
reply streams in chunks, and a credential long enough to be split — which is
most of them — could reach `log_events`, history and the stream in fragments,
past the redaction #2355 added.

That would be the same disclosure #2355 closed, through a different door.

## What is not yet known

- Whether a model reply ever actually carries a credential it read from tool
  output, and how often, rather than a placeholder.
- Whether fragments of a real credential across chunks are recoverable in
  practice — adjacent chunks reassemble trivially in a client, so probably yes,
  but it should be shown.
- Whether other chunked event types (tool-call arguments streamed as deltas,
  for instance) have the same property.

## First step

A controlled reproduction with a **synthetic secret**, not a placeholder:
extend the `secrets` profile, or add a sibling, so the model is prompted to
restate an echoed synthetic credential in its reply. Then inspect whether
fragments reach durable history unredacted, and across how many runs. The suite
already plants synthetic values and inspects every public response, so the
harness is most of the way there.

Only once that establishes the leak is real is it worth choosing a design.

## Shapes a fix could take, for later

None is proposed yet; each has costs worth weighing against a measured risk.

- **Scan across a boundary.** Hold back the tail of each chunk — as many bytes
  as the longest registered value, less one — and prepend it to the next before
  matching. Catches splits; delays the last few bytes of every streamed chunk.
- **Redact the reassembled message.** Persist chunks as now, but redact the
  concatenated text when the message completes, and rewrite the stored chunks.
  No streaming latency; the live stream still leaks until the rewrite.
- **Do not persist chunks at all** where a message-level event exists. Removes
  the per-chunk surface; changes what history and replay contain.

Related: #2355, ADR 0019 §10, #1614.

jhgaylor:
Closes #2359

## Problem

The [findings on #2359](https://github.com/managoat/fountain/issues/2359#issuecomment-5704185476) were reproduced deterministically through `Output.log/4`, with no deployment involved.

- **Split values leaked in every sandbox stream.** `Conversations.log!/1` redacts each row on its own, so a registered value cut by a chunk boundary matched in neither row. Both fragments were persisted and broadcast. This happened for `agent_message_chunk` lines and for raw `stdout`/`stderr` chunks.
- **`turns.reply_text` stored the value whole.** It joins a turn's text blocks when the turn ends, and search returns it.
- **Escaped values were missed even whole.** An `acp` row is a JSON line that the peer encoded, so a value containing `"`, `\` or a newline (a PEM key, a JSON credential) is stored escaped. Redaction matched only the raw bytes, so it missed these values even when they were whole in one row.

## Fix

- **`Fountain.Conversations.RedactionCarry`** (new, pure) holds back output only when its end could be the start of a registered value. It holds at most longest − 1 bytes, plus any match that tail overlaps, because a longer value may still be arriving. Output that cannot start a value is written at once. A conversation with nothing registered holds nothing.
  - **Raw streams:** the tail bytes are held, per stream, and prepended to that stream's next chunk.
  - **`acp`:** holding bytes would cut a JSON frame, so the carry reads the decoded text of consecutive chunk lines and holds **whole lines**. Lines are written byte for byte unless a value really spans them. Only in that case are the lines joined into one row. This keeps reattach replay dedup, which matches rows by exact content, working in the common case. A line of any other kind releases what is held first, so order is preserved.
- **`Output`** owns the carry (`output_carry` on the server) and flushes it:
  - when output arrives for another turn;
  - in `finish_acp_turn`, before `TurnMachine.finish` derives `reply_text`;
  - on interrupt, command exit and transport failure.
  - The budget (#331) counts bytes as they are written.
- **`Redaction`** registers the JSON-escaped form of each value that escapes (`patterns/1`). `redact/2` matches both forms, and `lookup/1` still returns the values themselves. `log!/1` remains the single writer-side redaction. The moduledoc now states the boundary guarantee and its limits.
- `format_status` redacts `output_carry` (by construction it holds the start of a secret).

`conversation_server.ex` grows by 2 lines, to 2214 against the pin of 2220.

No ADR. This fix restores the existing redaction guarantee and adds no new architectural choice. The guarantee and its limits are written in the `Redaction` and `RedactionCarry` moduledocs.

## Tests

- `redaction_carry_test.exs` checks the equivalence property: for every single cut, for one chunk per byte or per codepoint, and for a StreamData property over up to 12 cuts, redacting what the carry releases row by row equals redacting the stream whole. The streams include a value that is a prefix of another, two overlapping values, a JSON-escaping value and multi-byte UTF-8. Both raw bytes and ACP lines are covered.
- `output_redaction_test.exs` covers the writer: split `agent_message_chunk` rows (stored rows, `reply_text`, PubSub frames), one chunk per byte, ordering against other lines, flush and turn change, verbatim release with no holding, `stdout`/`stderr` with a separate tail per stream, and escaped values whole and across lines.
- `conversation_server_acp_test.exs` covers the server end to end. The reply splits the value and ends with its first bytes. No row carries a fragment, `reply_text` is `the key is [REDACTED], and it ends with sk-syn`, and every output row precedes the `turn done` stage.
- **Revert checks:**
  - Disabling holding (`hold_from/2` returns the whole size) and dropping the escaped patterns fails 15 carry and writer tests, plus the server test.
  - Removing only the flush in `finish_acp_turn` fails the server test (`reply_text` is `nil`).
- `conversation_server_log_budget_test.exs` now uses a chunk of `.` characters. A chunk of `a` characters ends in the first byte of a registered test value, so the carry correctly holds that byte back.

Local runs:
- `mise exec -- mix precommit`: PASSED (9 stages).
- `apps/fountain` suite: 6171 tests, 1 property, 0 failures.

## Residual risks

- **Latency.** A chunk whose tail matches the first bytes of a registered value waits for the next chunk or the turn's end. For `acp` it is the whole line. For a one-byte match this is common but short.
- **Reattach replay.** A mid-turn reattach whose replay covers lines that were **joined** will not dedup them, so that text can repeat. Redacted rows already have this limitation. Lines held when a server stops were never persisted, and the replay writes them again.
- `terminate/2` does not flush. A server that stops mid-turn loses what it holds (nothing leaks).
- Tool-call input is not delta-streamed in the rows Fountain stores (`tool_call` carries the whole `rawInput`). In-progress `tool_call_update` content was not audited for every adapter.


apps/fountain/lib/fountain/conversations/redaction_carry.ex
```diff
@@ -0,0 +1,252 @@
...
+          case held do
+            %{kind: ^kind, lines: lines} -> {[], lines}
+            _ -> {release(held, values), []}
+          end
+
+        {out, rest} = settle(pending ++ [{line, map, text}], values)
+        {released ++ out, if(rest == [], do: nil, else: %{kind: kind, lines: rest})}
+
+      _other ->
+        {release(held, values) ++ [line], nil}
```
BinaryBourbon: [P1] Preserve the redaction boundary across intervening ACP updates

Flushing on another event kind is intentional, but it does not end the assistant text stream. Register `abcdefgh`, then feed `agent_message_chunk("abc")`, an in-progress `tool_call_update`, and `agent_message_chunk("defgh")`. This branch releases `abc`; neither emitted text row contains the whole value, so writer-side redaction changes neither. I reproduced `Blocks.assistant_text(redacted_events) == "abcdefgh"` with the exact PR source. `Conversations._unsafe_turn_reply_text/1` uses that same join, so the split credential still reaches stored reply text/search, while live events expose both fragments.

Keep the pending text boundary across non-text updates (while preserving event order), or safely redact the unresolved prefix before releasing it. Extend the existing "held text is flushed before a different line" test with the later suffix and assert both PubSub frames and derived reply text contain no recoverable secret.

apps/fountain/lib/fountain/conversations/redaction_carry.ex
```diff
@@ -0,0 +1,252 @@
...
+        {emit(ready, values), rest}
+    end
+  end
+
+  defp held_index(cut, starts, ends, matches) do
+    with index when is_integer(index) <- Enum.find_index(ends, &(&1 > cut)) do
+      start = Enum.at(starts, index)
+
+      case Enum.find(matches, &crosses?(&1, start)) do
+        {earlier, _length} -> held_index(earlier, starts, ends, matches)
```
BinaryBourbon: [P1] Bound the ACP carry before retaining another whole line

This recursion can walk back through every previous line, so ACP retention is not bounded by the longest secret. With `abcdefgh` registered, feed text chunks `"abc"`, then repeat `"defghabc"`. Each new suffix needs holding, and the completed match crossing the preceding line boundary pulls the cut backward all the way to the first line. I reproduced 11, 101, and 1,001 input lines yielding zero output rows and retaining every line (164,159 wire bytes at 1,001).

`Output.log/4` only applies its byte budget in `write/4`, so this sequence never reaches the cap while streaming. A sandbox can keep expanding server memory and the repeated scans of the entire pending transcript indefinitely; benign output in the same chain also remains invisible until turn end. Bound pending wire bytes before appending, or compact completed redactions so only a bounded unresolved tail remains. Add a repeated-boundary regression through `Output.log/4` with a small budget and verify retained state is bounded before flush.

BinaryBourbon (changes requested):
Adversarial review of all 13 changed paths at `6bafb02d88408e352deb14d6d7d2d2d2a8e8d682`, including callers, lifecycle flushing, tenant/credential boundaries, replay compatibility, budgets, and regression tests.

Requesting changes for two independently reproduced defects: intervening ACP updates still let a split secret reach reply text, and whole-line carry can grow without bound before the output budget runs. Details and minimal reproductions are inline.

Validation: loaded the exact PR source with the pinned Elixir 1.19.2 / OTP 28 toolchain and installed dependency BEAMs. The existing carry suite passed (6 tests, 1 property), and separate source-level diagnostics reproduced both findings. The initial standalone property invocation lacked loaded StreamData application configuration; explicitly loading that application resolved the harness error. `git diff --check` passed. GitHub's current `CI required` check and all executed checks are green. I did not run the database/server suite or the full Review Loop verification recipe locally.

apps/fountain/lib/fountain/conversations/redaction_carry.ex
BinaryBourbon: Fixed in d64a80981 by keeping the pending text tail across non-text lines (option (a)).

**What changed.** The carry now tracks one channel per ACP chunk kind, alongside the `stdout` and `stderr` channels.
- A tool call, `tool_call_update` or plan line is written at once and no longer releases a text tail.
- The tail carries forward to the next chunk of the same kind.
- `abc` | update | `defgh` is now written as: the update first, then one `abcdefgh` row, which the writer redacts.

**Why (a) rather than (b).** Redacting the unresolved prefix at a tool line would over-redact on every tool call whose preceding text happened to end in a value's first byte, which is common. Option (a) loses no text. Its cost is order: at most a tail's worth of text lands after a tool line it streamed before. Nothing written ahead of the tail can hold a fragment. A tool line carries no reply text, and the text written up to the cut contains no part of a value that crosses the cut.

**Tests.**
- `output_redaction_test.exs`, "a tool update between the halves does not release the first half": your reproduction through `Output.log/4`. It asserts that `Blocks.assistant_text` over the stored rows and over the PubSub frames, the join of every text block, and `_unsafe_turn_reply_text/1` contain neither `abcdefgh` nor `abc`.
- `redaction_carry_test.exs`: the same reproduction at the carry level. The property test now puts tool lines at random points between up to 12 cuts. It checks joined text equals whole-stream redaction, contains no registered value, and that every tool line is written exactly once. There are also exhaustive single-cut and per-codepoint cases with a tool line after every piece.

**Revert check.** I restored the old flush at a non-text line and restored the file by copy afterwards. It fails 6 tests: both reproductions, the property, and the exhaustive cases.

apps/fountain/lib/fountain/conversations/redaction_carry.ex
BinaryBourbon: Fixed in d64a80981. The carry no longer holds whole lines, and the amount it holds is now a hard bound.

**What changed.**
- **Resolved text is written at once.** Each channel writes a chunk's text up to the cut, and only the unresolved tail is kept. The ACP line is re-encoded with the shorter text, or skipped if that text is empty. A match that is already whole is released immediately and redacted by the writer. Your `abc` + `defghabc`×N sequence now writes one `[REDACTED]` row per chunk and keeps only `abc`.
- **The tail is bounded by construction.** It is at most twice the longest registered value, less one byte: a proper prefix, plus one straddling match that a longer value might still extend.
- **A hard cap of 8 KiB per channel (`RedactionCarry.max_hold/0`) fails safe.** If the tail would exceed the cap, which only a value longer than 4 KiB can cause, the placeholder is written in its place. The carry records the remainder of every value that tail could begin, and drops what continues one as it arrives. The trade-off is over-redaction, and it is written up in both the `RedactionCarry` and `Redaction` moduledocs.
- **Replay dedup still works in the common case.** A line is written byte for byte whenever nothing is held before or after it.

**Tests.**
- `output_redaction_test.exs`, "stays bounded when every chunk both completes and begins a value": your reproduction through `Output.log/4`, run 1,001 times starting 4,000 bytes under the budget. It asserts at every step, before any flush, that `held_bytes ≤ 2*8-1`. It also checks that rows were written while streaming until the cap was reached, and that the stored text contains no `abc`.
- "a value longer than the cap is replaced, never held or leaked": a value of about 10 KB split across three `stdout` chunks stores `cert: [REDACTED] end\n`, and the held bytes stay within the cap.
- `redaction_carry_test.exs`: the same 1,001-repeat reproduction yields 1,002 rows. The property and exhaustive tests assert the bound after every input.

**Revert checks** (each file restored by copy afterwards):
- Holding everything back from the first possible prefix, with the cap disabled: 9 failures, including both bound tests.
- The cap disabled alone: the oversized-value test fails.

apps/fountain/lib/fountain/conversations/redaction_carry.ex
```diff
@@ -0,0 +1,282 @@
...
+      {out, channel} = step(channel, values, text)
+
+      rows =
+        cond do
+          out == text -> [line]
+          out == "" -> []
+          true -> [encode(map, out)]
+        end
+
+      channel = Map.put(channel, :template, map)
```
BinaryBourbon: [P2] Bound the retained ACP template, not only `channel.tail`

`Map.put(channel, :template, map)` retains the entire decoded sandbox notification, including its original text and unrestricted `_meta`. With `abcdefgh` registered, send an `agent_message_chunk` whose text is `abc` and whose `_meta.padding` contains a large string. All text is held, so zero rows reach `Output.write/4`; its byte budget never runs. The tail counter says 3 bytes, but the conversation server retains the complete metadata indefinitely until that channel resolves or the turn ends.

Reproduced against this exact source: padding of 1,000 / 1,000,000 / 10,000,000 bytes produced zero rows and serialized carry sizes of 1,325 / 1,000,325 / 10,000,325 bytes, while `held_bytes/1` stayed 3. Through actual `Output.log/4` with a 1,000-byte budget, a 1,000,000-byte padding still returned `bytes: 0`, `capped: false` and a 1,000,423-byte carry. This is persistent uncharged state, beyond the decoder's pre-existing transient frame allocation; the advertised 8 KiB fail-safe does not bound it.

Keep a bounded template containing only what a later tail emission needs, without the original text, and bound or safely reject oversized retained metadata/frame content. Ensure retained sub-binaries cannot pin the original large frame. Add a regression that measures all retained carry state (including template backing binaries) after an oversized metadata-bearing prefix-only frame and verifies bounded state or safe truncation; checking `held_bytes/1` alone misses this defect.

BinaryBourbon (changes requested):
Independent adversarial re-review of the whole PR at `d64a809816d727580445219c71ac9ca75a98b7de`, with base `3ccd36a4f2c3bcfc9dd43a1f74431e2a77b980f6` (the PR diff starts at merge-base `9d3cd19f9f846fad8c150f8bbd9819cd09361c11`).

The prior interleaved-tool secret leak and repeated-boundary tail growth reproductions are addressed. Requesting changes for the remaining retained-template memory bypass detailed inline, and the current CI regression:

- **[P2] Update the session-isolation regression for the new chunk boundaries.** [CI partition 2, run 35155112330](https://github.com/managoat/fountain/actions/runs/35155112330/job/104992650498) fails `ConversationServerSessionIsolationTest` at `conversation_server_session_isolation_test.exs:174`: expected `["before restart", "during restart", "live"]`, got `["before restart", "during restart", "liv", "e"]` (seed `520099`; 1,309 tests, one failure). This revision now splits an otherwise harmless chunk whenever its suffix is a registered-value prefix. The unchanged assertion requires the original row boundaries, so the new behavior breaks this integration test and `CI required` is failed. Assert the assembled expected transcript while retaining every session/turn ownership and replay-duplication check, or preserve the original boundaries; then rerun the affected test and required CI. Do not skip the test or remove its ownership checks.

Coverage: all 13 changed paths—production `conversation_server.ex`, `output.ex`, `reattachment.ex`, `redaction.ex`, `redaction_carry.ex`; tests `conversation_server_acp_test.exs`, `conversation_server_log_budget_test.exs`, `conversation_server_redaction_test.exs`, `output_redaction_test.exs`, `output_test.exs`, `redaction_carry_test.exs`, `redaction_test.exs`; and `changelog.d/2359-redaction-across-chunks.md`. Followed the writer, turn completion/interruption/transport failure, ACP peer and block rendering, replay deduplication, redaction registration/status masking, and tenant-scoped transcript access. The documented shortened-line replay limitation remains explicit.

Validation: exact final source loaded with pinned Elixir 1.19.2/OTP 28 and existing dependency BEAMs; the carry suite passed **8 tests and 1 property**. An independent diagnostic reproduced the memory bypass through both `RedactionCarry.feed/4` and actual `Output.log/4` without a database. `git diff --check` passed. Reviewed the failing CI assertion and current checks. I did not run the database/server suite or the full Review Loop verification recipe locally. GitHub identity, open state, head, base, reviews and checks were verified immediately before submission.

apps/fountain/lib/fountain/conversations/redaction_carry.ex
BinaryBourbon: Fixed in 3030228cb. A channel no longer keeps any template.

**What each channel retains now, and nothing else:**
- **A held unit** of at most `max_unit/0` (4 KiB): the raw chunk, or the ACP line byte for byte, `:binary.copy`'d. A larger chunk is never held whole. It goes straight to the tail form.
- **A text tail** of at most `max_hold/0` (8 KiB), `:binary.copy`'d. The existing fail-safe applies above that cap: the tail is written as the placeholder, and whatever continues a value it could begin is dropped.
- **On `acp`, the session id**, copied, only if it is at most 256 bytes. Otherwise it is dropped.
- **Remainders of registered values** while the fail-safe is consuming. These are bounded by what was registered, not by what the sandbox wrote.

**How rows are rebuilt without a template.** A tail is written into the *next* chunk's own line, which is decoded transiently while it is processed. At the turn's end it goes into a minimal `session/update` built from the kind and the session id. So an oversized prefix-only frame's `_meta` is not carried anywhere. When that frame's text is held, its metadata is dropped rather than kept. That trade-off is documented under "What is retained" in the `RedactionCarry` moduledoc.

**Tests** (`output_redaction_test.exs`, "what the carry retains"):
- **"a prefix-only frame with huge metadata retains none of it":** your reproduction through `Output.log/4`, with 10 MB of `_meta.padding` and 1,000 bytes left under the budget. It runs two cases: `abc`, and a 300-byte prefix with a 200-byte session id, so the kept binaries are large enough to be real sub-binaries. For each it asserts zero rows, not capped, `:erlang.external_size(carry) < 2_000`, and `:binary.referenced_byte_size < 2_000` for every binary in the carry. It then checks that a held prefix still completes and is redacted.
- **"a tail cut from a huge chunk does not pin the chunk":** a 1 MB chunk ending in a 300-byte prefix, on `stdout` and on `acp`. It asserts no retained binary references the chunk.

**Revert checks** (each restored by file copy):
- Re-adding `Map.put(channel, :template, map)` fails the `external_size` assertion.
- Dropping the session-id copy fails the `referenced_byte_size` assertion.
- Dropping the tail copy fails the pin test. With a tail of 64 bytes or less it would not: the runtime copies a sub-binary that small anyway, which is why that test uses 300 bytes.

BinaryBourbon:
**CI fix: `ConversationServerSessionIsolationTest` (`["before restart", "during restart", "liv", "e"]`, run 35155112330)**, fixed in 3030228cb.

**Why `live` was split.** `SpriteEnv` registers every value in the sandbox env for redaction. That includes non-secret identifiers such as `FOUNTAIN_SANDBOX_ID`, a UUID. A probe on this test printed the registered value `eb55c928-3aea-…`, the shared sandbox id of both conversations, in the runs where it started with `e`. Any chunk ending in `e` then looked like the start of a value, so d64a80981 cut the row to `liv` and held `e`. The permission lines that followed did not release it, and the turn's end wrote `e` as its own row. With a few hex UUIDs registered, most chunks ending in a hex letter get cut this way. The test's own randomness decides which letters, not the seed. It failed within 60 local repeats.

**Decision: fix the behaviour, keep the test's assertion.** Splitting ordinary rows was costly. It changed row boundaries at the end of most text runs, and each rewritten line also defeats reattach replay dedup. So a chunk whose end could begin a value is now held **whole** (at most 4 KiB), as the raw chunk or the exact ACP line. If the next chunk shows no value crossing the boundary, the held unit is written byte for byte. Only a real crossing, or a chunk too large to hold, falls back to joining text into a tail. A held chunk also survives tool and permission lines unchanged, so the text is not cut there. The cost is latency and order. Such a chunk appears when the next chunk of its kind arrives or the turn ends, and at most one held chunk per kind lands after a tool line it streamed before.

**The isolation test.**
- Unchanged, including every ownership and replay-duplication check. It passed 150 consecutive local repeats, and seed 520099.
- Disabling whole-unit holding makes it fail again (`"liv", "e"`), and fails 4 carry and writer tests.
- A new carry test pins the shape directly: `alivE` with a value starting `E` is written as `alivE`, on `acp` and on `stdout`.

**Other tests that asserted a chunk was persisted before the turn ended.** Found by grepping for exact data and boundary assertions. They are hold-proofed by ending their chunks in a byte no registered value can start with:
- `conversation_server_test` ("hello from the sprite", which ends in hex `e`);
- `conversation_server_acp_test` ("the answer");
- `conversation_server_log_budget_test` (`,`, `;` and `hello\n` instead of `c`, `d` and `hello`).

The budget test's `c` and `d` chunks could otherwise be held and never reach the cap. Each test still asserts what it did.

**Local results:** `mix precommit` PASSED (9 stages); full `apps/fountain` suite 6 doctests, 1 property, 6179 tests, 0 failures.

apps/fountain/lib/fountain/conversations/redaction_carry.ex
```diff
@@ -0,0 +1,387 @@
...
+
+  # The remainders of every value `tail` could be the start of.
+  defp in_progress(patterns, tail) do
+    size = byte_size(tail)
+
+    for pattern <- patterns,
+        k <- min(byte_size(pattern) - 1, size)..1//-1,
+        binary_part(tail, size - k, k) == binary_part(pattern, 0, k),
+        uniq: true,
+        do: :binary.copy(binary_part(pattern, k, byte_size(pattern) - k))
```
BinaryBourbon: [P2] Keep fail-safe continuations compact instead of copying every suffix

When a registered value overlaps itself, this comprehension copies a different remainder for every matching prefix. Register `String.duplicate("a", 8194)` and feed `String.duplicate("a", 8193)` to `stdout`: crossing the 8 KiB hold cap emits `[REDACTED]`, then retains 8,193 independently copied suffixes totaling **33,566,721 bytes** while `held_bytes/1` reports **0**. With a 16,384-byte registered value, the same input retains **100,667,391 bytes** of suffix data. `uniq: true` cannot collapse suffixes of different lengths.

Reproduced through exact `Output.log/4` with a 1,000-byte budget and only persistence stubbed: `bytes: 10`, `capped: false`, one placeholder row, serialized carry **33,607,887 bytes**, and **33,567,009 process binary bytes after GC**. That state remains while the turn awaits more output. Environment/vault changesets impose no value-length limit, and `SpriteEnv` registers their values, so an ordinary-size tenant-controlled value can amplify a small output into tens of MB of retained server memory. This is additional quadratic allocation, beyond retaining the registered value itself.

Represent continuations as references/offsets into a bounded set of registered patterns (or an equivalent compact matcher), avoiding a full copy per possible offset. Add a self-overlapping long-value regression through `Output.log/4` that checks all retained state/backing binaries after the cap is crossed and still verifies subsequent secret bytes are consumed safely; the existing tail counters and prefix-only metadata tests do not exercise this state.

BinaryBourbon (changes requested):
Fresh adversarial review of the entire PR at `3030228cbf8f89c615e96b8eeb17147a4c937f3f`, against base `3ccd36a4f2c3bcfc9dd43a1f74431e2a77b980f6` (PR merge-base `9d3cd19f9f846fad8c150f8bbd9819cd09361c11`).

The earlier interleaved-event secret leak, repeated-boundary growth, oversized ACP template retention, and session-isolation CI regression are addressed. Requesting changes for one separately reproduced availability defect: the fail-safe materializes every overlapping suffix of a registered value, creating quadratic retained memory outside the carry and output-budget counters. Details and remedy are inline.

Coverage: all 14 changed paths: production `conversation_server.ex`, `output.ex`, `reattachment.ex`, `redaction.ex`, `redaction_carry.ex`; tests `conversation_server_acp_test.exs`, `conversation_server_log_budget_test.exs`, `conversation_server_redaction_test.exs`, `conversation_server_test.exs`, `output_redaction_test.exs`, `output_test.exs`, `redaction_carry_test.exs`, `redaction_test.exs`; and `changelog.d/2359-redaction-across-chunks.md`. Followed registration, per-conversation writer ownership, ACP/raw output, reply-text assembly, budget accounting, replay deduplication, lifecycle flushing and status masking. The documented latency, ordering, replay and metadata tradeoffs are acknowledged.

Validation: loaded exact PR source with pinned Elixir 1.19.2 / OTP 28 and existing dependency BEAMs. Carry suite: **9 tests and 1 property, 0 failures**. Independent diagnostics reproduced the new finding through both `RedactionCarry.feed/4` and actual `Output.log/4`; the latter stubbed only persistence in an isolated VM, without a database. `git diff --check` passed. Current `CI required`, all six test partitions and all other executed checks are green. I did not run the database/server suite or full Review Loop verification recipe locally. Verified BinaryBourbon identity, open state, unchanged head/base, reviews and checks before submission.

apps/fountain/lib/fountain/conversations/redaction_carry.ex
BinaryBourbon: Fixed in a3c0a4e7b. I kept the change minimal, since #2366 is expected to replace this carry.

**What changed.** Continuations are no longer binaries. The fail-safe now keeps:
- **`pairs`:** `{value index, offset}` integers, deduplicated and capped at 64 (`max_continuations/0`).
- **`fingerprint`:** the `:erlang.phash2/1` of the value list those indexes point into. The values stay in the Redaction registry and none are copied.
- **`skip`:** the most bytes any continuation could still need. That is the longest value's length minus its shortest matching prefix.

When enumerating would pass 64 pairs, it stops listing them and keeps only `skip`. For each value, the prefix scan runs from the shortest prefix up, so the first match already gives that value's longest remainder.

**Consuming.**
- **With pairs, and the registry fingerprint unchanged:** a continuation that the data completes ends there. One that the data only begins stays pending. The most any of them consumed is dropped.
- **Without pairs, or after the registry changed** (for example `Redaction.add` sorted a longer value in front, so an index would name the wrong value): exactly `skip` bytes are dropped.

**Why dropping `skip` bytes is safe** (the argument is also in the code comment). Every value that could have been cut off in the replaced tail needs at most `skip` more bytes, and `skip` only counts down by bytes actually dropped. So the fallback can drop output that was not a secret, but it can never write one.

**Retained state after the cap, measured with `:erlang.external_size` of the whole carry:**

| Registered value | Before | After |
|---|---|---|
| `"a"` × 8194 (fed 8193) | 33,607,961 bytes | 136 bytes |
| `"a"` × 16384 | 134,291,726 bytes | 136 bytes |

After the fix, the carry holds only `%{skip: 8193, pairs: nil, fingerprint: …}`.

**Tests.**
- **`output_redaction_test.exs`, "a self-overlapping #{size}-byte value past the cap retains a few integers", for 8,194 and 16,384.** This is your reproduction through `Output.log/4` with 1,000 bytes left under the budget. After the cap it asserts:
  - one placeholder row, and not capped;
  - `external_size(carry) < 2_000`;
  - `referenced_byte_size < 2_000` for every retained binary;
  - the value's last byte and the following would-be continuation bytes are dropped, never written;
  - once nothing can still continue the value, `visible\n` is written.
- **`redaction_carry_test.exs`, "a registry change while the fail-safe consumes drops the remainder by count".** An exact single-pair continuation, then `Redaction.add` of a longer value, which shifts the indexes. It asserts the remainder is dropped and ` end` is still written.

**Revert checks** (each restored by file copy):
- The previous fail-safe fails both size tests, with 33,607,961 and 134,291,726 bytes retained.
- Skipping the fingerprint comparison fails the registry-change test, because the remainder is written in plaintext.

**Other checks.** Conversations tests plus `sandbox_files_test`: 1,650 tests and 1 property, 0 failures. `mix precommit` passed all 9 stages.

apps/fountain/lib/fountain/conversations/redaction_carry.ex
```diff
@@ -0,0 +1,464 @@
...
+    if :erlang.phash2(patterns) == fingerprint do
+      consume_pairs(consuming, List.to_tuple(patterns), data)
+    else
+      consume(%{consuming | pairs: nil}, patterns, data)
+    end
+  end
+
+  defp consume(%{skip: skip}, _patterns, data) do
+    dropped = min(skip, byte_size(data))
+    rest = binary_part(data, dropped, byte_size(data) - dropped)
```
BinaryBourbon, line 384: [P2] Keep ACP text on a UTF-8 boundary when the count fallback ends

`consume/3` is shared by raw streams and decoded ACP text, but this slice can leave the ACP remainder starting inside a multibyte codepoint. Source-derived example: a registered value of 8,194 ASCII `a` bytes and an `agent_message_chunk` containing its first 8,193 bytes enter the new fallback with `pairs: nil, skip: 8193`. If the next valid text chunk contains 4,097 `é` characters (8,194 bytes), dropping 8,193 bytes leaves the single continuation byte `0xA9`. `tail_step/3` emits that remainder, and `render/4` calls `encode/2`'s `Jason.encode!`, which raises on invalid UTF-8. This exception propagates through `Output.log/4` in the ConversationServer. The registry-change fallback can reach the same problem.

For ACP, conservatively consume through the end of the codepoint when the byte count expires; retain byte-oriented handling for stdout/stderr. Add regression coverage for multibyte output following the count fallback and a registry change. The current self-overlap fallback tests use only ASCII stdout, so they do not exercise this path. This finding follows directly from the source and was not executed as a new diagnostic.

BinaryBourbon (changes requested):
Fresh independent review of the entire PR at `a3c0a4e7bcd97c89ddef204bd7174d624b929515`, with security and resource-boundary emphasis. Reviewed against base `3ccd36a4f2c3bcfc9dd43a1f74431e2a77b980f6`, using PR merge-base `9d3cd19f9f846fad8c150f8bbd9819cd09361c11`.

Requesting changes for one P2 correctness defect: the new byte-count fallback can split a UTF-8 codepoint in decoded ACP text, causing `Jason.encode!` to raise in the conversation server. The concrete source-derived condition and remedy are inline. A second independent reviewer confirmed the code path. The earlier copied-continuation memory growth is addressed by the bounded integer pairs/count representation; it is not being raised again.

Coverage: all 14 changed paths—production `conversation_server.ex`, `output.ex`, `reattachment.ex`, `redaction.ex`, `redaction_carry.ex`; tests `conversation_server_acp_test.exs`, `conversation_server_log_budget_test.exs`, `conversation_server_redaction_test.exs`, `conversation_server_test.exs`, `output_redaction_test.exs`, `output_test.exs`, `redaction_carry_test.exs`, `redaction_test.exs`; and `changelog.d/2359-redaction-across-chunks.md`. Followed registry ownership, ACP peer canonicalization/session filtering, raw and decoded output, carry retention, lifecycle flushing, reply-text persistence, budget accounting, replay limitations and status masking.

Validation: the coordinating reviewer ran the standard focused suite on this exact head with the pinned toolchain and an isolated database: **189 tests and 1 property, 0 failures**. GitHub `CI required` and all executed current checks are green. This reviewer completed source inspection; optional new diagnostics were stopped by an automated filter and were not retried. The Unicode condition is source-derived, not an executed reproduction. This is a manual review, not a claim that the full installed Review Loop verifier ran. Verified BinaryBourbon identity, OPEN state, unchanged head/base, latest reviews and checks immediately before submission.

BinaryBourbon (changes requested):
Independent adversarial correctness / lifecycle / API review of the whole PR at a3c0a4e7bcd97c89ddef204bd7174d624b929515, against base 3ccd36a4f2c3bcfc9dd43a1f74431e2a77b980f6 (triple-dot diff).

**Changes requested: one P2 correctness defect.** I independently identified and source-checked the UTF-8 boundary failure recorded in [the other independent review](https://github.com/managoat/fountain/pull/2364#pullrequestreview-5229630392). The new count-based fallback can return a suffix starting inside a multibyte codepoint from otherwise valid ACP text (`redaction_carry.ex:379-383`). That suffix reaches `Jason.encode!` through `tail_step` and `render`, which raises and terminates the conversation server instead of merely over-redacting. Keep ACP text on valid UTF-8 boundaries when consuming by count, preserve the bounded-memory guarantee, and cover both count-mode and registry-change fallback with multibyte-text regressions. This specific failure is established by source inspection, not an executed reproduction. No duplicate inline finding is added.

Reviewed all 14 changed paths: the five production modules (`conversation_server.ex`, `output.ex`, `reattachment.ex`, `redaction.ex`, `redaction_carry.ex`); the eight changed conversation tests (`conversation_server_acp_test.exs`, `conversation_server_log_budget_test.exs`, `conversation_server_redaction_test.exs`, `conversation_server_test.exs`, `output_redaction_test.exs`, `output_test.exs`, `redaction_carry_test.exs`, `redaction_test.exs`); and `changelog.d/2359-redaction-across-chunks.md`. Traced persistence, turn completion/interruption/transport failure, reattachment, registry rotation, peer normalization, Blocks/reply text, and CLI ACP forwarding. The prior copied-continuation memory defect is fixed; no additional blocker found in this coverage.

Validation: independently ran 3,000 randomized cases through both raw and ACP streams using overlapping ASCII values plus quote/backslash/newline escaping; all passed. Inspected the coordinating reviewer's exact-head focused-suite log: 189 tests and 1 property, 0 failures. Triple-dot `git diff --check` passed. GitHub CI required is successful. The optional Unicode reproduction was not run; the passing checks do not cover the source-derived defect above.

apps/fountain/lib/fountain/conversations/redaction_carry.ex
jhgaylor: Fixed in e87b40e60. The change is minimal, since #2366 is expected to replace this carry.

**Reproduced first.** I wrote the tests before the fix and ran them against a3c0a4e7b. Your example raises as you described: `(Jason.EncodeError) invalid byte 0xA9 in <<169, 32, 118, ...>>` from `encode/2`, through `feed/4`, out of `Output.log/4`. The registry-change fallback raises the same way.

**What changed.** A text channel on `acp` is now marked (`text?: true`, one boolean on the channel). In `tail_step/3`, after `consume/3`, a text channel drops any leading continuation bytes (`0b10xxxxxx`) from what is left, which is the rest of the codepoint the count ended in, at most three bytes. `stdout` and `stderr` are untouched and stay byte-oriented.

**Why it is safe.** It only drops more, never less, so the existing argument holds: the fallback can drop output that was not a secret, and cannot write one. The count is already zero when any bytes remain, so `skip` is unaffected.

**No state across chunks.** Each `acp` chunk's text comes from `Jason.decode/1`, which rejects invalid UTF-8 and lone surrogate escapes (checked). So every chunk is whole UTF-8 on its own, a count that uses up a whole chunk ends on a codepoint, and the leftover bytes can never arrive in the next chunk. Nothing new is retained except the boolean.

**Other slices of `acp` text.** I checked each one; none can land inside a codepoint while registered values are valid UTF-8:
- **`consume_pairs/3`:** a completed value is dropped to the value's end. The dropped bytes equal a suffix of the value that starts on a lead byte (the chunk's first byte), so it ends on a codepoint. A pending continuation drops the whole chunk. A miss drops nothing.
- **`hold_from/2` (the cut, including the `max_unit/0` and `max_hold/0` paths and a tail cut from a large chunk):** the cut is at a byte equal to some value's first byte, or at a match's start. Both are lead bytes, so both are codepoint starts in valid text.
- **The cap:** the tail is replaced by the placeholder, not sliced.

A registered value that is not valid UTF-8 (for example one that starts with a continuation byte) could still cut inside a codepoint. I did not change that. Values reach the registry from JSON and Postgres text, so I believe it is unreachable, and `Redaction.redact/2` on the row has the same property today.

**Tests.**
- **`output_redaction_test.exs`, "multibyte text after the fail-safe drops by count":**
  - "a count that ends inside a codepoint drops the rest of it" is your example through `Output.log/4` on `acp`: no raise, every row is valid UTF-8 and valid JSON, the stored text is the placeholder plus ` visible and after`.
  - "a registry change mid-value does the same": an exact pair, then `Redaction.add` of a longer value, then `é` text padded so the count ends on the second byte of an `é`. Same assertions, and no part of the value is stored.
- **`redaction_carry_test.exs`, "a count that ends inside a codepoint: acp drops the rest of it, stdout does not":** 2-, 3- and 4-byte codepoints at every alignment through `feed/4`. On `acp` exactly the codepoints the count reaches into are dropped. On `stdout` the same input is cut at the byte.

**Revert checks** (each restored by file copy):
- With the a3c0a4e7b file, all three new tests fail with `Jason.EncodeError ... invalid byte 0xA9`.
- With the alignment applied to every channel, the `stdout` assertion fails (`"é end"` instead of `<<169, 195, 169, ...>>`), and so do two existing byte-cut tests.

**Other checks.** Conversations tests plus `sandbox_files_test`: 1,653 tests and 1 property, 0 failures, 2 skipped. `mix precommit` passed all 9 stages.

apps/fountain/lib/fountain/conversations/redaction_carry.ex
```diff
@@ -0,0 +1,480 @@
...
+    end
+  end
+
+  defp advance(channel, patterns, text, _unit), do: tail_step(channel, patterns, text)
+
+  # Write as far as the cut, keep the rest. Over the cap, the tail is replaced
+  # rather than kept (see "What is retained, and the fail-safe").
+  defp tail_step(channel, patterns, data) do
+    {data, consuming} = consume(channel.consuming, patterns, data)
+    data = if channel.text?, do: to_codepoint(data), else: data
```
BinaryBourbon: [P2] Keep raw fail-safe output valid for PostgreSQL text

The ACP-only alignment leaves the same problem at the raw streams' persistence boundary. Register an 8,194-byte value consisting of `a`, then send two individually valid UTF-8 stderr chunks: 8,193 `a` bytes, followed by 4,097 `é` characters plus ` visible`. The first chunk activates count-only consumption with `skip = 8193`; the second leaves `<<0xA9>> <> " visible"`. The new pure carry test explicitly preserves this raw byte slicing (redaction_carry_test.exs:286–290).

The live stderr handler forwards chunks directly to Output.log/4. Output.persist/3 then calls Conversations.log!/1 without UTF-8 normalization; LogEvent.data is an Ecto :string backed by PostgreSQL text, and Ecto/Postgrex forward this binary unchanged. On the UTF8 database, insertion therefore raises out of the conversation server. This PR creates invalid database text from valid input chunks, independently of the older handling of already-invalid raw input.

Make the fail-safe's retained raw suffix safe for the existing text persistence boundary, and cover stdout/stderr through Output.log/4 and the database. This is a source-derived finding, not a newly executed reproduction.

BinaryBourbon (changes requested):
Adversarial review — resource bounds, serialization, and compatibility. Reviewed the entire 14-path PR at e87b40e602ea840871b13973f29d3523e56e8166 against the approved base, including the original changes and all follow-up fixes.

The ACP JSON encoding fix addresses the previously reported continuation-byte failure, and the bounded held-unit/tail/continuation state addresses the prior retention findings. One P2 remains: the count fallback can turn valid multibyte stderr (or raw stdout) chunks into invalid UTF-8, which then reaches PostgreSQL text persistence unchanged. See the inline finding. This conclusion is derived from source and existing regression assertions; I did not execute a new reproduction.

Coverage included the carry and registry, output budget and writer, terminal flush paths, replay deduplication, transcript/reply_text extraction, CLI/SDK consumers, all changed tests, and changelog. Shared validation on this exact head: 192 focused tests and 1 property passed with 0 failures using the pinned toolchain and an isolated database. I also verified git diff --check and current GitHub CI (all executed checks successful). This is a manual review, not a claim that the full Review Loop verifier was run locally.

BinaryBourbon (changes requested):
Independent adversarial correctness, lifecycle and API review of the whole PR at `e87b40e602ea840871b13973f29d3523e56e8166`, against base `3ccd36a4f2c3bcfc9dd43a1f74431e2a77b980f6` (PR merge-base `9d3cd19f9f846fad8c150f8bbd9819cd09361c11`).

**Changes requested: one remaining P2 correctness defect.** I independently traced and concur with [the raw-output UTF-8 finding](https://github.com/managoat/fountain/pull/2364#discussion_r4036917545). The ACP JSON-encoding failure is fixed, but the same count cutoff can turn valid `stderr` or raw `stdout` chunks into an invalid UTF-8 suffix. For example, register 8,194 ASCII `a` bytes, output its first 8,193 bytes, then output 4,097 `é` characters plus ` visible`. Count consumption leaves a leading `0xA9`. ACP aligns that suffix; raw output passes it through `Output.persist` and `Conversations.log!` to `LogEvent.data`, a PostgreSQL TEXT column. Ecto/Postgrex do not normalize those bytes, so the insert raises out of the conversation-server callback. The direct current-command stderr handler makes this a normal reachable path.

This finding concerns the carry corrupting otherwise valid complete chunks; it does not require solving the pre-existing handling of already-invalid transport chunks. Ensure count-mode/registry-change consumption cannot send such an invalid suffix into durable text persistence, keep retained state bounded, and add writer-level regressions for stdout and stderr. The existing raw assertion checks only the pure carry and explicitly permits the invalid suffix. The condition is established by source inspection, not an executed reproduction. No duplicate inline thread is added.

Coverage: all 14 changed paths: production `conversation_server.ex`, `output.ex`, `reattachment.ex`, `redaction.ex`, `redaction_carry.ex`; tests `conversation_server_acp_test.exs`, `conversation_server_log_budget_test.exs`, `conversation_server_redaction_test.exs`, `conversation_server_test.exs`, `output_redaction_test.exs`, `output_test.exs`, `redaction_carry_test.exs`, `redaction_test.exs`; and `changelog.d/2359-redaction-across-chunks.md`. Followed registry ownership/rotation, peer normalization/session filtering, persistence/budget accounting, completion/interruption/transport failure, reply-text assembly, replay deduplication, status masking and CLI ACP forwarding. Earlier interleaved-event leaks and retained-memory findings are fixed; their documented latency, ordering, replay and over-redaction tradeoffs are acknowledged.

Validation: inspected the coordinating reviewer's exact-head isolated-database log: **192 tests and 1 property, 0 failures**. Triple-dot `git diff --check` passed. Current GitHub CI required and all executed checks are successful. I ran no optional diagnostic for the new condition and did not run the full installed Review Loop verification recipe. Verified BinaryBourbon identity, OPEN state, unchanged head/base, latest reviews and checks immediately before submission.

apps/fountain/lib/fountain/conversations/redaction_carry.ex
jhgaylor: Fixed in e3cdf6ce9. This reverses the byte-oriented handling of `stdout`/`stderr` that round 4 asked to retain; the newer instruction is followed.

**Reproduced first.** I wrote the writer-level tests before the fix and ran them against e87b40e60, through `Output.log/4` to the test database. All four raise as you described: `(Postgrex.Error) ERROR 22021 (character_not_in_repertoire) invalid byte sequence for encoding "UTF8": 0xa9`, from the insert in `Conversations.log!/1`, out of `Output.persist/3` and `Output.log/4`. That is `stdout` and `stderr`, for both the count fallback and the registry-change fallback.

**What changed.** In `tail_step/3`, the remainder is passed through `to_codepoint/1` on every channel, but only when `consume/3` actually cut this chunk (it returned fewer bytes than it was given). That condition replaces round 4's `text?` flag, which is removed. On `acp` the two are the same, because an uncut text starts on a codepoint. No state is added; the channel is back to its round-3 shape.

**Why it is safe.** It only drops more (at most three continuation bytes), so the existing argument holds: the fallback can drop output that was not a secret, and cannot write one. The count is already zero when any bytes remain.

**What it deliberately does not touch.** A raw chunk that the transport cut inside a codepoint, and that the carry did not cut, is written as it arrived. That case is older than this PR: with nothing registered, `Output.log/4` given `"caf" <> <<0xC3>>` raises the same 22021 today, and `persist/3` has no scrub on `main` either. I left it alone as you scoped, and will raise it separately.

**No state across chunks.** A count that uses up a whole chunk ends at the chunk's end. If the chunk is valid UTF-8, that is a codepoint's end. If it is not, the chunk was already invalid on its own, which is the older case above.

**Other slices, rechecked for raw streams** (valid chunks, valid registered values; the JSON-escaped patterns of valid values are valid too). None can turn a valid chunk sequence into an invalid row:
- **`consume_pairs/3`:** for every `:done` pair, the dropped bytes equal a suffix of that value which begins with the chunk's first byte. That byte is a lead byte, so the suffix is whole codepoints and ends on one. This holds for each pair on its own, so it holds for the most-consumed one, whether or not the match was a coincidence. A pending pair drops the whole chunk. A miss drops nothing.
- **`hold_from/2`, including the `max_unit/0` and `max_hold/0` paths:** the cut is at a byte equal to some value's first byte, or at a match's start. Both are lead bytes. The text it cuts is a tail that began on such a cut, joined to whole chunks.
- **The cap:** the tail is replaced by the placeholder, not sliced.

**Tests.**
- **`output_redaction_test.exs`, "multibyte raw output after the fail-safe drops by count"**, for `stdout` and `stderr`, through `Output.log/4` and the database:
  - "a count that ends inside a codepoint of #{stream} drops the rest of it": your example. No raise, every stored row is valid UTF-8, the stored bytes are the placeholder plus ` visible and after`.
  - "a registry change mid-value on #{stream} does the same": an exact pair, `Redaction.add` of a longer value, then `é` text padded so the count ends on the second byte of an `é`. Same assertions, and no part of the value is stored.
- **`redaction_carry_test.exs`:**
  - The round-4 test that pinned raw byte slicing now expects the same result on `stdout` as on `acp` (renamed "... drops the rest of it, on acp and on stdout"). It pinned the defect.
  - New: "a raw chunk the fail-safe did not cut is written as it arrived". While the fail-safe is consuming, a chunk that starts with `0xA9` and misses is written byte for byte.
  - The two existing arbitrary-byte-cut tests ("any cuts, with tool lines anywhere between them", "one chunk per byte, ...") are unchanged and pass. They pin transport-split chunks, which is the behaviour left alone.

**Revert checks** (each restored by file copy):
- With the e87b40e60 file, the four writer tests fail with the 22021 error above, and the updated carry test fails on `<<169, 195, 169, ...>>` for `stdout`.
- With the alignment applied to every chunk, cut or not, the new "did not cut" test and the two existing byte-cut tests fail.

**Other checks.** Conversations tests plus `sandbox_files_test`: 1,658 tests and 1 property, 0 failures, 2 skipped. `mix precommit` passed all 9 stages.

BinaryBourbon (approved):
Adversarial review — secret boundaries, retained state, and persistence compatibility.

Reviewed the complete 14-file PR diff at e3cdf6ce962b307cd6c425236e29179c153155bb against merge base 9d3cd19f9f846fad8c150f8bbd9819cd09361c11, using the contributor/reviewer guidance from target base 3ccd36a4f2c3bcfc9dd43a1f74431e2a77b980f6. Traced the changed behavior through the conversation server, Output persistence/broadcast, log_events schema, Blocks/reply_text, registry updates, and replay/turn completion.

No remaining change-requesting defect found. The previous raw UTF-8 blocker is resolved: after the fail-safe consumes bytes, the remaining data is advanced past a cut codepoint before either JSON encoding or raw text persistence. Uncut chunks keep their existing behavior. The added Output/database regressions cover stdout and stderr, including registry changes. Per-channel retention limits, secret continuation handling, escaped values, budget enforcement, and crash-state redaction remain consistent with the intended behavior. The documented over-redaction, delayed rows, and replay limitations are accounted for.

Validation: the coordinating reviewer ran the nine focused regression files in an isolated database at this exact head: 197 tests and 1 property, 0 failures; I inspected the result log. This is an independent manually requested source review using shared test evidence, not a claim that Review Loop's full verification recipe ran.

BinaryBourbon (approved):
Adversarial review — correctness and lifecycle: APPROVED.

Reviewed the complete PR diff at `e3cdf6ce962b307cd6c425236e29179c153155bb` against merge base `9d3cd19f9f846fad8c150f8bbd9819cd09361c11`, with approved-base contributor and review guidance. Traced carry state through raw/ACP output, the ordered database writer and live broadcasts, replay handling, reply-text construction, turn completion, interruption, and transport failure. No remaining concrete blocker found.

The previous raw UTF-8 finding is addressed: after consumption cuts bytes, the remainder advances to a codepoint boundary before either raw text persistence or ACP JSON encoding. Unconsumed raw chunks retain their incoming bytes. The new database regressions cover both stdout and stderr for count overflow and registry changes; the existing tests cover multibyte ACP output, carry bounds, split-value redaction, and lifecycle behavior.

Validation: the coordinating worker's exact-head run of nine focused regression files completed with **197 tests and 1 property, 0 failures**; I inspected the result log. This is an independently requested manual source review with shared focused-test evidence, not a claim that the installed Review Loop's full verification recipe ran.

<details>
<summary>Full PR source coverage (14 files)</summary>

- `apps/fountain/lib/fountain/conversations/output.ex`
- `apps/fountain/lib/fountain/conversations/reattachment.ex`
- `apps/fountain/lib/fountain/conversations/redaction.ex`
- `apps/fountain/lib/fountain/conversations/redaction_carry.ex`
- `apps/fountain/test/fountain/conversations/conversation_server_log_budget_test.exs`
- `apps/fountain/test/fountain/conversations/conversation_server_redaction_test.exs`
- `apps/fountain/test/fountain/conversations/conversation_server_test.exs`
- `apps/fountain/test/fountain/conversations/output_redaction_test.exs`
- `apps/fountain/test/fountain/conversations/output_test.exs`
- `apps/fountain/test/fountain/conversations/redaction_carry_test.exs`
- `apps/fountain/test/fountain/conversations/redaction_test.exs`
- `changelog.d/2359-redaction-across-chunks.md`

</details>

Issue #2373, opened 2026-09-17
Conversations: a prompt that wakes a parked conversation loses its images


jhgaylor:
Found by reading the prompt path for #1406. Not reproduced against a deployment; the code path is short enough to follow by eye.

## What happens

`POST /api/conversations/{id}/prompts` accepts `images`. `ConversationServer.send_prompt/4` has two roads:

- A live server takes `{:send_prompt, prompt, images}` and the turn gets its images.
- With no server registered (a parked conversation, or the gap after a deploy), it calls `Wake.wake_conversation(conv_id, prompt)`. **`images` is not an argument.** `Wake` later delivers the prompt with `ConversationServer.queue_initial_prompt(pid, initial_prompt)`, whose `images` parameter defaults to `[]`, at all four of its handoff sites.

So a prompt with images sent to a conversation that has to be woken opens a turn with the text and no images. The request is answered `200 {"status":"queued"}`, and the `conversation.prompted` audit row records the real `image_count`, so the trail says images were sent and the turn says none arrived. Nothing reports the loss.

`Launch` (a fresh conversation) is not affected: it calls `queue_initial_prompt/3` with `attrs["images"]`.

## Suggested fix

#1406 adds `Fountain.Conversations.PromptDelivery`, which already carries what travels with a prompt's text from `send_prompt/4` through `Wake` to the cast (`for_wake/2`, `hand_over/2`). Images belong in the same envelope. The rollout caveat in that module's doc applies: a server on the previous release matches the three-element cast only.

## Acceptance

- [ ] A prompt with images sent to a conversation with no live server opens a turn that has them.
- [ ] A test that fails on `main`: stub `Wake`'s provider calls as `conversations_wake_test.exs` does, send a prompt with one image through `send_prompt/4`, and assert the turn's `images`.
- [ ] The image size and count checks still run once, in `FountainWeb.PromptImages`, before the wake.

PR #2374, opened 2026-09-17
feat(conversations): carry a prompt's client_request_id to the turn it opens (#1406)

jhgaylor:
Refs #1406 (step 1 of 3). Design and what is deliberately not built: https://github.com/managoat/fountain/issues/1406#issuecomment-5717338773. Tracked from #2247.

## The problem

`POST /api/conversations/:id/prompts` answers `{"status":"queued"}`. The turn gets its id later, so a client that owns a durable work item has to infer which turn is its own from turn order. Two clients submitting at a turn boundary make that inference wrong.

## Why the response does not carry `turn_id`

There are two roads from a prompt to a turn, and only one has a turn when it answers. A live `ConversationServer` opens the turn before replying. A conversation with no server is woken, and `Wake` hands the prompt over in a cast after provisioning, long after the request was answered. A field that is present on one road and null on the other would be relied on by the clients that only tested the first. So the caller names the request, and the turn carries that name whichever road opened it.

## What changes

- `PromptRequest.client_request_id`: optional, 1 to 200 characters, refused with 422 outside that. `PromptResponse.client_request_id` repeats it (null when none was sent; always rendered, not `required`, so a generated client still decodes an older server).
- `turns.client_request_id` (additive, nullable, no index: nothing looks a turn up by it). Rendered on `Turn`.
- The `turn` / `started` stage event carries it beside `turn_id`, on both publishers (fresh launch and a reused connection). An event for a prompt that carried nothing keeps its old shape.
- `Fountain.Conversations.PromptDelivery` is the one place that says how the value travels: `call/3`, `cast/3`, `for_wake/2`, `hand_over/2`. `Wake`'s four copies of "if there is a prompt, queue it" become one `hand_over/2`.
- The manual gets "Find the turn your prompt opened", including what is not promised.

## Rollout

Horde places a conversation's server on any node (`UniformDistribution`, `members: :auto`) and a rollout keeps old and new pods in one cluster (`maxSurge: 1`, `maxUnavailable: 0`). So any prompt, on either road, can be addressed to a server of the previous release. That server's catch-all clauses answer a four-element call `:unknown_call` (rendered 422), **drop a four-element cast after the caller was told `queued`**, and log the whole message, prompt text included.

`PromptDelivery` therefore changes a message's shape only when there is an id to carry **and the receiving node has this module** (`understands?/1`: local is yes; remote is one `:erpc` call to `Code.ensure_loaded?/1`, and any failure reads as no). Otherwise the prompt goes in the shape every release matches and the id is left behind, with a warning that names the node and not the prompt. A turn without its label is a smaller loss than a prompt that never ran. A prompt with no id never pays the check. With `MACHINE_OWNER_ENABLED` on, an owner of the previous release has the same window (its changeset does not cast the column). The manual names the window. The migration is additive and nullable.

## Not built, and documented as such

- **Idempotent retries.** A repeat of the same value opens a second turn. Doing better needs a unique index and a refusal through `Machine.admit_turn` that does not drop the connection, plus an answer for a repeat that arrives while the first is still waking. Separate issue to follow.
- **`turn_id` in the response**, for the reason above.
- Webhook deliveries do not carry the value (they carry `turn_id`). The manual says to read the turn.

## Tests

`prompt_correlation_test.exs`, against a real `ConversationServer`: the live call, the cast after a wake, the reapply continuation, the `started` event on both publishers, a refused prompt leaving its id on nothing, the message shapes, and **two clients submitting at once**, on both roads. On a live server the mailbox serialises them, so exactly one is accepted and the other is `busy`; on a waking server both were answered `queued` and the second cast is dropped silently. In both, the turn names the prompt that ran and the other id is on no turn. Also: the real `Wake` driven with a tuple prompt, the node capability check, the bound in `travelling/1`, and an explicit null. Controller tests cover the echo, the null, the three 422s and the turn list.

Revert probes, each run and restored: dropping the id from the turn insert fails 5 tests; un-correlating either `started` publisher fails the one test that covers it; dropping the id from the reapply continuation fails its test.

## Checks run locally

- `mix precommit toolchain conflict-markers compile deps format credo dialyzer sobelow`: all eight ok. The release-assemble stage was not run.
- `mix test` over `test/fountain/conversations`, `test/fountain_web`, `test/fountain/machines`: 3965 tests, 0 failures. Size pin, schema guardrail and direct-writes tests pass. `conversation_server.ex` grows 2010 → 2017 against a pin of 2025.
- `scripts/test-docs.sh` and `scripts/changelog.py check` pass.
- `generate.py --check`, `build.sh --check`, and the TypeScript, Python, Elixir and Swift contract verifiers pass. Generated diff: 14 lines in `contract.json`, 6 in `openapi.ts`, 2 in `ConversationWire.generated.swift`. No handwritten client edit was needed. `release:skip-sdk`: the SDK surface for this lands with step 3 and carries the version bump.

## Review

Round 1 was an independent read-only review; its findings and what changed are in commit `ec14aa972` and the comment below. No blockers; three should-fix (the rollout claim, which was wrong; the manual stating the live-server answer as the rule; no test through the real `Wake`) and three nits, all addressed. One suspicion left open and pre-existing: a NUL byte in the value would fail the turn insert, exactly as one in `prompt` does today.

## Collision note

ADR 0058 stage 8b (`stack/0058-8b-binding`) is in flight in `Wake` and the turn writes. This touches `Wake` only at the four handoff sites and adds one attribute to the turn insert. Whichever lands second rebases, and should re-check the size pin.

Found on the way and filed: #2373 (a prompt that wakes a parked conversation loses its images).


jhgaylor:
Idempotent retries, the part this PR declares unbuilt, are now #2375.

BinaryBourbon (changes requested):
Adversarial review of `ec14aa9722e9c338caf0c4cf443717ede4edf006` against `02883de9197c3c63436a6396011c1f3f8fe36409`: **changes requested**.

1. **[P2] The new column cannot store every ID the API accepts.** At `apps/fountain/priv/repo/migrations/20260917150000_add_client_request_id_to_turns.exs:21`, `:string` creates `varchar(255)`. The API, `PromptDelivery.travelling/1`, and `Turn.changeset/2` all count graphemes: `String.duplicate("e\u0301", 200)` passes all three but contains 400 PostgreSQL characters. A disposable PostgreSQL insert reproduces `value too long for type character varying(255)`. The insert in `Conversations._unsafe_create_turn_on_sandbox/4` does not rescue that exception; on the default inline admission path it propagates through `TurnMachine.open/7` and terminates the conversation server instead of refusing the label with 422. A wake may already have answered `queued`. Use a text column with the declared bound, or make the API, delivery, changeset, and column agree on codepoint limits; cover accepted Unicode IDs through persistence.

2. **[P2] Whitespace-only IDs are echoed but silently disappear from the turn.** At `apps/fountain/lib/fountain/conversations/turn.ex:152–158`, the ordinary Ecto cast treats an all-whitespace string as empty. Both `" "` and `"\t\n"` pass the exact API schema and travel with the prompt, yet produce a valid changeset with `client_request_id: nil`; `Turn.correlate/2` then omits the field from the started event. The response still echoes the supplied string, so the caller cannot find the turn by its accepted correlation. Preserve the literal ID during casting, or reject whitespace-only IDs consistently before accepting the prompt. Add a request-to-turn/event regression case.

3. **[P2] Absence of a correlated turn does not establish that retrying is safe.** `docs/api.md:409–410` says that after a timeout and an empty lookup, the prompt did not run and can be sent again. The immediately following rollout caveat permits a completed prompt to have a null ID; separately, an accepted initial prompt can still be queued behind provisioning when the lookup runs (`ConversationServer.queue_initial_prompt/4`). Either leaves no matching turn without proving nonexecution, and a retry can run the work twice. Describe this outcome as unknown and remove the safe-retry guarantee until an authoritative refusal/settlement mechanism or idempotency exists.

Coverage: all 20 changed files, including live/wake/reapply delivery, both user-turn started publishers, turn admission and persistence, controller authorization and responses, migration, tests/harness, manual/changelog, and generated contract/TypeScript/Swift models. Traced relevant unchanged admission, autonomous-turn, audit, redaction, and tenancy callers. Targeted probes loaded the exact-head Turn, PromptDelivery, and ClientRequestId schema with the pinned toolchain and matching locked dependencies; they reproduced the validation/correlation defects and generated column SQL. PostgreSQL rejection was independently reproduced in a disposable database. Exact-head GitHub CI is green, including all six test partitions, release/API contract, SDK checks, and CI required. I did not run the full application suite locally or reproduce the full GenServer crash at runtime; that consequence follows from the inspected uncaught exception path. No source changes were made.

BinaryBourbon:
Round two fixed in `d8fa06671`. All three, with the probes.

**1. The column could not hold every id the door takes.** Confirmed exactly as described: `String.duplicate("é", 200)` is 200 graphemes to the API, to `PromptDelivery.travelling/1` and to `Turn.changeset/2`, and 400 characters to `varchar(255)`. A disposable insert gave `ERROR 22001 (string_data_right_truncation)`. The migration now adds `:text`, so the bound stays in the three places that declare it in graphemes and the column stops holding a fourth, different opinion. Covered end to end (request → delivery → admission → row → started event) in `PromptCorrelationTest`, "an id of 200 combining graphemes fits the column"; with the column altered back to `varchar(255)` that test is the only one in the file that fails.

**2. Whitespace-only ids were echoed and then dropped.** Also confirmed: Ecto trims a string before deciding it is empty, so `" "` and `"\t\n"` cast to `nil` while the response repeated them. I took the "preserve the literal id" option — the id is an opaque label, and the API (`minLength: 1`) and `travelling/1` are the layers that decide what is acceptable. `Turn.changeset/2` now casts the field untrimmed, leaving a literal `""` as the one string that still reads as "the caller sent none", which both of those layers already refuse to carry. Regression cases at the door (controller: the response echo and the opts that go out) and at the turn (the row and the started event). Reverting the cast fails the spaces test and nothing else. The `""`-is-none semantics are pinned too, so the backstop that keeps a non-API caller from failing turn admission stays.

**3. The safe-retry guarantee is gone.** `docs/api.md` no longer says a missing turn means the prompt did not run. It now says a turn carrying your value shows the prompt ran, no such turn shows nothing — it may still be waiting for the machine that must start first, or the rollout window may have opened its turn without the value — and that sending it again can run the work twice. Idempotency stays not-built and is tracked in #2375.

Checks on `d8fa06671`: `mix precommit toolchain conflict-markers compile deps format credo` passed; the whole suite passed (6637 core tests + the five extension apps, 0 failures); `scripts/test-docs.sh` passed. I did not run the release-assemble stage locally. `conversation_server.ex` is untouched by this round, so it stays at 2017 lines against the 2025 pin.

apps/fountain/lib/fountain/conversations/turn.ex
```diff
@@ -130,9 +151,21 @@ defmodule Fountain.Conversations.Turn do
       :origin,
       :conversation_id
     ])
+    |> cast_client_request_id(attrs)
     |> validate_required([:turn_number, :prompt, :status, :conversation_id])
     |> validate_inclusion(:status, @statuses)
     |> validate_inclusion(:origin, @origins)
+    |> validate_length(:client_request_id, min: 1, max: @client_request_id_max)
```
BinaryBourbon, line 181: [P2] Reject NUL in client_request_id before turn admission

An ordinary prompt (`hello`) with `client_request_id` containing U+0000 passes `ClientRequestId.request()`, `PromptDelivery.travelling/1`, and this changeset. PostgreSQL `text` cannot contain U+0000: a parameterized insert with this ID raises `Postgrex.Error` / SQLSTATE 22021 (`invalid byte sequence for encoding "UTF8": 0x00`). The new field therefore makes an otherwise valid prompt fail.

`Conversations._unsafe_create_turn_on_sandbox/4` does not rescue this insert exception. With machine ownership disabled it escapes the conversation actor; with ownership enabled it crashes the machine owner and becomes `:sandbox_unavailable`, after which `kick_turn` drops the healthy connection. On a wake this can happen after the caller already received `queued`, with no turn created.

Reject NUL at the API with 422 and align the internal delivery guard and changeset so the ID cannot reach the database through either path. Add regression coverage with an ordinary prompt and NUL only in the ID for live and wake delivery. The older prompt field also accepting NUL does not make this inherited: this PR introduces the failing value through a new field on an otherwise valid request.

apps/fountain/lib/fountain_web/controllers/conversation_controller.ex
```diff
@@ -822,19 +822,23 @@ defmodule FountainWeb.ConversationController do
     user = conn.assigns.current_user
 
     with {:ok, images} <- decode_images(params["images"]) do
-      do_prompt(conn, id, prompt, user, images)
+      do_prompt(conn, id, prompt, user, images, params["client_request_id"])
```
BinaryBourbon: [P2] Read the client request ID from the validated request body

This controller uses `CastAndValidate` with `replace_params: false`. The request schema validates `conn.body_params`, while this `params` map also includes query parameters. With the ordinary JSON body `{"prompt":"hello"}`, a query-only `client_request_id` containing 201 characters passes validation, is echoed in the successful response, and is then silently discarded by `PromptDelivery.travelling/1`. The turn and its started event carry no ID, so the response names a correlation the caller can never find. `?client_request_id[]=x` also passes and makes the response render an array where `PromptResponse` declares a nullable string.

An isolated probe using this head's schemas and the locked Plug/OpenApiSpex dependencies confirms both cases. Read this optional field from the validated body (ignoring undeclared query values), or explicitly validate the merged value actually forwarded. Add a controller regression with the ID supplied only in the query string. Any NUL check added to the JSON schema must also avoid leaving this alternate input path open.

BinaryBourbon (changes requested):
Fresh adversarial review of d8fa0667189618930117ee96f318c2d31c69ac3f against 02883de9197c3c63436a6396011c1f3f8fe36409. Two independent reviewers inspected the whole 20-file change and relevant callers. Two P2 input-validation defects remain; see the inline comments.

The three findings from the previous review are fixed: combining Unicode IDs now fit the text column; whitespace-only IDs survive casting; and the manual now explains that an absent correlated turn does not make a retry safe.

Validation: inspected live, wake, reapply, admission, autonomous-turn, event, API, migration, test, and generated-contract paths. Exact-head source probes with the pinned Elixir/Ecto/OpenApiSpex dependencies confirm whitespace/Unicode preservation and NUL acceptance. A disposable PostgreSQL instance with the locked Postgrex dependency confirms whitespace/Unicode inserts succeed and the NUL-only ID fails with SQLSTATE 22021 while the prompt is ordinary text. A separate isolated Plug/OpenApiSpex probe confirms query-only oversized/array IDs pass validation but are consumed from merged params and dropped by prompt delivery. Actor failure paths were traced in source; the full application suite was not run for this review.

BinaryBourbon:
Round three fixed in `fe4154a62`. Both reproduced first, both covered, both probed by reverting.

**1. NUL in the id.** Confirmed end to end: a JSON body whose `client_request_id` carried U+0000 was answered `200` and echoed, and a `Repo.insert` of that turn raises `ERROR 22021 (character_not_in_repertoire)`. Three layers now hold one rule:

- `Schemas.ClientRequestId` declares a pattern that refuses U+0000, so the door answers 422 before anything is delivered — which is what matters on the wake path, where the caller would otherwise already hold `queued`;
- `Turn.has_nul?/1` is that same rule for a caller that is not the door;
- `PromptDelivery.travelling/1` and `Turn.changeset/2` both ask it, so an id that would raise at the insert never reaches turn admission — the changeset refuses it as an ordinary validation error instead.

A test pins the pattern and `has_nul?/1` to each other over the same ids, so the two statements of the rule cannot drift. Coverage: 422 at the door with an ordinary prompt and `reject(&send_prompt/4)` proving nothing was sent; the NUL id added to the "does not travel" list, which is the delivery guard both the live and wake roads share; and the changeset case. Reverting the pattern fails only the door test; reverting the two internal guards fails only their two.

**2. Query-only ids.** Also confirmed: with the body `{"prompt":"hello"}`, a 201-character `client_request_id` in the query string came back in the response, and the `[]=` form rendered a list where `PromptResponse` declares a nullable string. `prompt/2` now reads the field from the body the cast approved (`conn.private.open_api_spex.body_params`) instead of the merged `params`, so an undeclared query value is ignored. The regression covers both shapes: response `null`, and nothing in the delivery opts. Reverting the read fails it.

Scope note: `prompt` and `images` are still read from `params`, as they always were. A query `prompt` cannot win over the body's, and neither field is what this PR introduces, so I left them alone rather than widen the change.

The manual now says both: the value cannot contain a null character, and Fountain reads it from the request body.

Checks on `fe4154a62`: `mix precommit toolchain conflict-markers compile deps format credo` passed; the whole suite passed (6641 core tests plus the five extension apps, 0 failures); `scripts/test-docs.sh` passed; `scripts/sdk-contract/generate.py` produced no diff, because the distilled contract carries neither `pattern` nor the length bounds, and the TypeScript and Python contract verifiers pass. `sdk/contract/contract.json` is byte-identical to the previous commit, so the Swift and Elixir verifiers cannot have moved; I did not run those, or release-assemble, locally. `conversation_server.ex` is untouched by this round: 2017 lines against the 2025 pin.

docs/api.md
```diff
@@ -372,6 +372,59 @@ event cursor so a reconnect can resume after the last event processed.
...
+
+```json
+{"prompt": "Run the approved plan.", "client_request_id": "plan-7-step-3"}
+```
+
+The response repeats the value. Fountain stores it on the turn that the prompt
+opens, and `GET /api/conversations/{id}/turns` shows it as `client_request_id`.
+The `turn` stage event with the `started` state carries it next to `turn_id`.
+Read that event to bind your work item to the turn. Every later event of the
+turn carries the same `turn_id`.
```
BinaryBourbon: [P2] Confirm a stream match against the stored turn before binding

A started event can carry a different, colliding ID even after rollout. With an ordinary plain environment value `NODE_ENV=production`, `SpriteEnv.build/4` registers `production` for redaction. The valid ID `production-build-7` is echoed and stored literally on its turn, but both new started publishers send their JSON through `Conversations.log!/1`, whose redactor changes it to `[REDACTED]-build-7`. An exact-head probe reproduces that rewrite. Another client can legally have submitted the distinct ID `[REDACTED]-build-7`, so the instruction here can bind that client to the first client's turn; a missing-event timeout does not protect against this false match.

Preserve the existing single-writer redaction. A minimal safe correction is to make the persisted turn authoritative: require checking every candidate event's `turn_id` against the original `client_request_id` in the turns response before binding, update the schema wording consistently, and add a regression using a configured plain env value and the two colliding legal IDs. An implementation that preserves unambiguous correlation without weakening redaction would also resolve this.

BinaryBourbon (changes requested):
Fresh adversarial review of fe4154a62d546a84c69a74bd5bad026f024559d0 against 02883de9197c3c63436a6396011c1f3f8fe36409. Two independent reviewers covered the entire change and relevant callers. **Changes requested:** one P2 remains, described inline: transcript redaction can make distinct client IDs collide in the started event and cause a client to bind the wrong turn.

The previous five findings are fixed: Unicode IDs fit the text column, whitespace survives casting, retry uncertainty is documented, U+0000 is rejected before delivery, and query-only IDs are not consumed.

Validation: exact-head source probes with the pinned Elixir and locked dependencies passed the valid/invalid ID and validated-body cases. A separate exact-head Turn/Redaction probe reproduces the event-label rewrite; the real event persistence and broadcast path was traced in source. Swift generated-model freshness and the Python contract check pass. Current-head GitHub CI is green, including all six test partitions, static analysis, release/API contract, SDK checks and CI required. We did not run a full application/GenServer or database test locally this round. No source changes were made.

Coverage includes live/wake/reapply delivery, fresh/reused/autonomous turns, owner admission, persistence, billing/concurrency and tenant gates, event redaction/broadcast, API, migration, regression tests, docs and SDK consumers.

<details>
<summary>All 20 changed paths reviewed</summary>

```text
apps/fountain/lib/fountain/conversations/connection.ex
apps/fountain/lib/fountain/conversations/conversation_server.ex
apps/fountain/lib/fountain/conversations/prompt_delivery.ex
apps/fountain/lib/fountain/conversations/turn.ex
apps/fountain/lib/fountain/conversations/turn_launch.ex
apps/fountain/lib/fountain/conversations/turn_machine.ex
apps/fountain/lib/fountain/conversations/wake.ex
apps/fountain/lib/fountain_web/controllers/conversation_controller.ex
apps/fountain/lib/fountain_web/controllers/conversation_json.ex
apps/fountain/lib/fountain_web/schemas.ex
apps/fountain/priv/repo/migrations/20260917150000_add_client_request_id_to_turns.exs
apps/fountain/test/fountain/conversations/prompt_correlation_test.exs
apps/fountain/test/fountain/conversations/prompt_replay_test.exs
apps/fountain/test/fountain_web/controllers/conversation_controller_test.exs
apps/fountain/test/support/conversation_server_case.ex
changelog.d/1406-prompt-client-request-id.md
docs/api.md
sdk/contract/contract.json
sdk/swift/Sources/FountainKit/Models/ConversationWire.generated.swift
sdk/typescript/src/generated/openapi.ts
```

</details>

jhgaylor:
Round four fixed in `1448369a0`. Reproduced first: `NODE_ENV=production` registers `production` (ten bytes, over the redaction floor), the ordinary id `production-build-7` reaches its started event as `[REDACTED]-build-7`, and a second client sending exactly that string gets an event that is byte-identical to the first client's. Both turns keep their own literal value, because the row is written by the changeset and not by `log!/1`.

I kept the redaction where it is. It is a single writer over the whole event on purpose, and a correlation label is not a good enough reason to carve a hole in it — particularly as the hole would have to be a key-shaped exception inside a function that deliberately knows nothing about the shape of what it scrubs.

So what changed is what Fountain promises:

- `docs/api.md` now says the turn is the record. Use the started event to find a candidate turn, then read the turn and compare its `client_request_id` with the value you sent; do not bind on the event alone. It says why, in one paragraph: event data has the sandbox environment values removed, so an event can show `[REDACTED]` inside an id, and another client can legally send that same text.
- The request-side schema description says the same thing, and so does the `Turn.client_request_id` description on the response side — that one is the value the caller sent, byte for byte, and it is the one to compare against.
- `Turn.correlate/2` says it in the code that writes the event.

The regression is the collision itself, not a paraphrase of it: one conversation, a registered plain value, and the two colliding legal ids — equal in the two started events, distinct on the two turns, each event still pointing at its own `turn_id`.

Checks on `1448369a0`: `mix precommit toolchain conflict-markers compile deps format credo` passed; the whole suite passed (6642 core tests plus the five extension apps, 0 failures); `scripts/test-docs.sh` passed; `scripts/sdk-contract/generate.py` gives a comment-only diff in the generated TypeScript (two descriptions, no type moved), committed, and the TypeScript contract verifier passes. The first full-suite run had one unrelated failure — `DeletionFenceTest`, a 55P03 lock timeout in `InferenceCredentials.lock_platform_source/0`, nothing my diff touches; it passes alone and the re-run was clean. `conversation_server.ex` is untouched by this round: 2017 lines against the 2025 pin.

#2376 is rebased on this (`0d151fd9e`) and carries the same wording through the shared definition.

apps/fountain/lib/fountain/conversations/conversation_server.ex
```diff
@@ -105,15 +98,15 @@ defmodule Fountain.Conversations.ConversationServer do
     result =
       case whereis(conv_id) do
         nil ->
-          case Wake.wake_conversation(conv_id, prompt) do
+          case Wake.wake_conversation(conv_id, PromptDelivery.for_wake(prompt, opts)) do
```
BinaryBourbon: Withdrawn: this finding incorrectly attributed an incoming main-branch change to this PR.

The PR merge base is `0facb18a11c3503dfa5c77e320db07aab4eb70a0`; image preservation was added afterwards on main at `47326f1b0b45d6301c10c05d12b743df68ad3988`. Comparing the main tip directly with this head made the image plumbing, test, and changelog appear deleted. They are not deletions in the actual 20-file PR diff.

The isolated probe demonstrates the pre-existing image-loss behavior in the older head, not a newly introduced regression. Please disregard the original P2 request. The current merge conflicts still need integration and validation against main.

BinaryBourbon (dismissed):
Correction: the blocking finding in this review is withdrawn. I compared head `719f0beca8b2af3208b92ce9156cabced02d6401` directly with the current main tip `47326f1b0b45d6301c10c05d12b743df68ad3988`, instead of the PR merge base `0facb18a11c3503dfa5c77e320db07aab4eb70a0`.

The actual PR diff has 20 changed files. The image-preservation fix landed subsequently on main; this PR does not delete that fix, its test, or its changelog fragment. My isolated probe established the older head's behavior but did not establish a regression introduced by this PR. The contrary statements in my original review were incorrect.

All 20 actual changed files were included in the completed source review, and the six earlier correlation findings are addressed. Current main and this head have merge conflicts, so their integration needs resolution and fresh validation. The current head has an empty check rollup. This mistaken CHANGES_REQUESTED verdict is being dismissed.

BinaryBourbon (approved):
Source approval for the actual 20-file PR diff at `719f0beca8b2af3208b92ce9156cabced02d6401` against merge base `0facb18a11c3503dfa5c77e320db07aab4eb70a0`. Two fresh independent reviewers found no remaining blocker in that diff. All six earlier findings are addressed, including confirming an event candidate against the literal ID on the stored turn.

Passed: isolated exact-head ID validation/delivery probes, archived-head TypeScript/Python contract verification, 28-scenario conformance lint, Swift generator freshness, and diff checks. Full application/database tests, SDK builds/conformance runs, and `mix precommit` were not run.

The current main tip `47326f1b0b45d6301c10c05d12b743df68ad3988` conflicts with this head, and the head has no check rollup. Resolve the conflicts while preserving main's wake-image fix, then review and validate the resulting revision before landing. This approval covers the reviewed source revision; it does not claim CI success or Review Loop service acceptance.

My earlier image-loss finding was based on comparing the main tip rather than the PR merge base. Its text has been corrected and that mistaken review dismissed.

jhgaylor:
Rebased onto `main` (`02399178a`). The approval above is for `719f0beca`; this is what changed after it, since the review asked for the conflict resolution to be reviewed on its own.

`main` landed #2384 for #2373 — the images this branch's wake road also touches — so both changes thread a payload through the same four handoff sites. They merge rather than collide:

- `ConversationServer.send_prompt/4` wakes with `Wake.wake_conversation(conv_id, PromptDelivery.for_wake(prompt, opts), images)`: main's images as their own argument, the correlation inside the prompt in the shape `Wake` takes.
- `PromptDelivery.hand_over/3` replaces the four `if is_binary(initial_prompt) and initial_prompt != "" do ... queue_initial_prompt(pid, initial_prompt, images)` blocks — it already owned the "is there a prompt" question — and hands both payloads to `queue_initial_prompt/4`. `images` keeps **no default** there, which is the point #2373 made: the default was how they were lost.
- `Wake.wake_conversation/3` is main's arity, unchanged by this branch.

Two tests knew the old shapes and moved with it: the wake stub is arity three and asserts the images beside the correlation, and the hand-off test sends images down both the correlated and the bare road. `prompt_images_wake_test.exs` — main's own regression for #2373 — passes unchanged on this head.

One more commit before that: stage 8b (#2377) took `conversation_server.ex` to three lines under its pin, and this branch adds seven — the two compatibility clauses for a prompt from a node on the previous release, and the `meta` that rides to `TurnLaunch.open/7`. That is plumbing a message shape cannot avoid; the logic is in `PromptDelivery`. Rather than raise the pin, the alias block pays for it: `alias Fountain.{...}` was six lines for four modules and is now one, and the `Fountain.Conversations` aliases repack from seven lines to five. No alias added or removed. The file is 2025 against the pin of 2025 — zero headroom, said out loud rather than papered over.

Checks on `35964b7a3`: `mix precommit toolchain conflict-markers compile deps format credo` passed; the whole suite passed (6698 core tests plus the five extension apps, 0 failures) against a test database rebuilt from the root alias; `scripts/test-docs.sh` passed; `scripts/sdk-contract/generate.py` produced no diff. Release-assemble was not run locally. The head's CI is green: 25 successes, 4 skipped.

PR #2376, opened 2026-09-17
feat(conversations): take client_request_id on conversation create (#1406)

jhgaylor:
Refs #1406 (step 2 of 3) and #2248. **Stacked on #2374**; base is `feat/1406-prompt-correlation`, so review that first and do not arm `--auto` here until this is retargeted to `main`.

## What changes

`POST /api/conversations` takes `client_request_id` for its first prompt, with the bound and wording `PromptRequest` already has (one `Schemas.ClientRequestId` definition). Every door that delivers a first prompt carries it:

| Door | How |
|---|---|
| Fresh launch | `PromptDelivery.queue/4` hands it to the new server with the prompt. It is never in the Horde child spec, which a deploy replays. |
| Attach (`sandbox_id`) | Added to the opts `send_prompt/4` already takes, beside the attribution. |
| Queued start (`queue: true`) | `client_request_id` joins `@queued_attr_keys`, so it waits with the prompt and the drainer replays it through `Launch`. |
| `channel_id` resume | **Ignored, and documented as ignored.** A resume does not deliver the prompt; the client sends it on the prompts route. |

With nothing travelling, `PromptDelivery.queue/4` makes the three-argument `queue_initial_prompt` call it always was, so the suites that stub that arity are untouched.

## #2248's exercise: what a genuine conversation field actually cost

This is the first real `ConversationCreateRequest` field since the foundation (#2227) merged. Recorded as that issue asks.

**Server:** one schema property, one allowlist entry, two call sites in `Launch`, two functions in `PromptDelivery` (`from_request/1`, `queue/4`). The field does real work, so this is more than the "shape-only" case in CONTRIBUTING.

**Generated diffs, no hand editing:** `contract.json` +4, `openapi.ts` +2, `ConversationWire.generated.swift` +9 (`clientRequestID` on `ConversationCreateRequest`: property, init parameter, coding key, encode).

**Handwritten client edits needed to send the field on create through the API-shaped doors: none.** That is `runRequest` (TypeScript, Swift `Fountain`, FountainKit), `run_request` (Python, Elixir), FountainKit `create`, and `fountain conv create --file`. `check-propagation.py` passes for TypeScript, Python, Elixir, the CLI and Swift, and the four contract verifiers pass with no manifest change. The foundation does what it set out to do for those doors.

**Where the field cannot be sent at all**, because the body is built key by key, as CONTRIBUTING says those helpers are meant to be: `run(prompt, config)` in TypeScript, Python and Elixir, the CLI's `conv run`, and the ACP bridge (`acp.go` creates without a prompt and later posts `{"prompt","images"}`). The ACP bridge is a likely consumer of #1406, so that one is a real gap rather than a design choice.

**What the synthetic probe could not show, and this field did:** every client's `runRequest` has a handwritten `channel_id` resume branch that re-sends the prompt on the prompts route from a fixed list, `prompt` and `images`. A create field that belongs *with the prompt* is dropped there in **five** places: `client.ts` `startConversation`, `client.py` `_start_conversation`, `fountain.ex` `start_conversation/2`, `Fountain.swift` `startConversation`, and FountainKit's `Run.swift` `startConversation`. FountainKit is the sharpest case: after this PR `ConversationCreateRequest(clientRequestID:)` is a typed, compiling field, it is dropped on resume with no error, and `ConversationsResource.prompt(_:_:images:)` encodes a private two-field body, so FountainKit cannot send the value on the prompts route at all. The probe's field is inert, so it never meets these branches. The foundation removed the per-field tax from the create request; it did not remove it from the second request the clients build by hand.

Those edits, a conformance scenario for the resume branch (there is none today), and the version bumps are step 3. **Step 3 has to land before any SDK release**, or a typed field ships that is silently dropped on resume; `release:skip-sdk` holds that line until then.

Second observation for step 3: the same branches work out which turn is theirs with `nextTurnNumber` (highest `turn_number` plus one), which is the inference #1406 exists to replace.

## Deploy order

This must not deploy before #2374 has. #2374's node capability check keeps an id-carrying prompt from being dropped by a pod of the previous release, and this PR routes the most common door, a fresh launch, through it.

## Tests

- Fresh launch: the value reaches `queue_initial_prompt/4` and is **not** in the child spec (`prompt_replay_test.exs`); a launch without it still makes the three-argument call.
- Attach: the value is in `send_prompt/4`'s opts beside `actor`.
- Queue: the value is in the stored request's `attrs`.
- 422 for an empty, oversized or non-string value, before `Launch` is called.
- End to end (the cast carrying the value to turn 1 on a real server) is covered in #2374.

Revert probes, each run and restored: dropping the value at the fresh-launch door, at the attach door and from the queue allowlist each fail exactly the one test that covers it.

## Review

Round 1 was an independent read-only review: no blockers. Addressed here: this body's undercount (five resume sites, not four; the helpers that cannot send the field), the deploy order, the shallow not-in-the-spec assertion (now `inspect`-based, plus the `:initial_prompt` key check), and the `Launch` attrs docstring. Rebased onto #2374's review round. Found by that review and filed separately, both older than this PR: #2378 (a queued `channel_id` start that resumes on replay drops its prompt) and #2373 (the wake path drops images).

## Checks run locally

- `mix precommit` static stages (toolchain, conflict-markers, compile, deps, format, credo, dialyzer, sobelow): all ok. Release assemble not run.
- `mix test` over `test/fountain/conversations`, `test/fountain_web` and the sandbox queue: 3432 tests, 0 failures.
- `scripts/test-docs.sh`, `scripts/changelog.py check`, `generate.py --check`, `build.sh --check`, the four contract verifiers and `check-propagation.py`: pass.

`release:skip-sdk`: the generated types move, the SDKs' public behaviour does not until step 3.


BinaryBourbon (approved):
Reviewed incremental diff `ec14aa9722e9c338caf0c4cf443717ede4edf006..531a19c14eedf06c5a2a98735ccfe798bd861d9c` (stacked on #2374).

No new blocking defects found in the 12 changed files. Traced fresh launch, attach/home reuse, queue persistence/replay, API validation, generated TypeScript/Swift models, and SDK resume boundaries. Initial correlation travels with the prompt and stays out of the Horde child spec; attach attribution is preserved.

Independently passed the TypeScript and Python contract verifiers and `git diff --check`. Inspected this head's successful CI required, six server test partitions, release/API and SDK checks; did not rerun the full Elixir suite locally.

This approves only the incremental #2376 change. #2374's identified defects still need fixing; rebase and retest this PR on that corrected base before landing/deploying.

BinaryBourbon:
Rebased onto `d8fa06671`, the round-two fixes on #2374 (text column for `client_request_id`, untrimmed casting so an id of spaces is not silently dropped, and the manual no longer promising a safe retry). Clean rebase, no conflicts; this PR adds no column and reuses the one request-side definition, so it inherits all three.

Re-checked on `b8fe2f5c1`: `mix precommit toolchain conflict-markers compile deps format credo` passed, `scripts/test-docs.sh` passed, and the four suites this PR touches plus `PromptCorrelationTest` passed (148 tests, 0 failures). Still stacked on #2374 — retarget to `main` before arming `--auto`.

BinaryBourbon:
Rebased onto `fe4154a62` (#2374 round three: the door refuses a null character in the id, and the prompts route reads the field from the validated body instead of the merged `params`). One conflict, in `schemas.ex`: this PR moves `ClientRequestId` and gives `request/1` its lead sentence, while the parent added the pattern to it in the old place. Resolved by keeping this PR's definition and carrying the pattern into it, so both doors still declare one rule.

While rebasing I checked whether this PR's door had the second defect too. It did, without the echo: `create/2` hands the whole merged `params` map to the launch and the request schema validates the body alone, so `?client_request_id=from-the-query` on an ordinary create reached the turn as if the body had carried it — unchecked against the bound the field declares. Fixed in `1323155ad`, which reads the field from the body the cast approved, exactly as the prompts route now does. It is a separate commit so the diff since your approval is one file plus its regression. Reverting the one line fails that test and nothing else.

That fix is also what makes the manual true: #2374 now tells clients to send this field in the request body, and that sentence has to hold on both doors.

Re-checked on `1323155ad`: `mix precommit toolchain conflict-markers compile deps format credo` passed, the whole suite passed (6645 core tests plus the five extension apps, 0 failures), `scripts/test-docs.sh` passed, and `scripts/sdk-contract/generate.py` produced no diff.

Still stacked on #2374 — retarget to `main` before arming `--auto`.

jhgaylor:
Rebased onto `1448369a0` (#2374 round four: the manual and both schema descriptions now say to confirm a correlation against the turn, because event redaction can rewrite the event's copy of an id). Same conflict as last time in `schemas.ex` — this PR owns the shared `ClientRequestId` definition — resolved by carrying the new wording into `@shared`, so both doors say it. `0d151fd9e` is the regenerated TypeScript for that description: comment-only, no type moved.

Re-checked: precommit static stages passed, the whole suite passed (6646 core tests plus the five extension apps, 0 failures), `scripts/test-docs.sh` passed, and the TypeScript contract verifier passes. Still stacked on #2374 — retarget to `main` before arming `--auto`.

jhgaylor:
Restacked onto `35964b7a3` (#2374 rebased onto main, merging main's wake-image fix from #2384 with this stack's correlation on the same four handoff sites). No conflicts in this PR's own commits this time.

Re-checked on `ad8554363`: precommit static stages passed, the whole suite passed (6702 core tests plus the five extension apps, 0 failures), `scripts/test-docs.sh` passed, and the contract regenerates with no diff. CI on the head is green. Still stacked — retarget to `main` before arming `--auto`.

Issue #2366, opened 2026-09-16
Redaction registry holds non-secrets, so the cross-chunk carry delays ordinary output


BinaryBourbon:
Follow-up to #2359 / #2364. The cross-chunk redaction carry in #2364 fixes a real
leak, but it gets its safety by holding output back, and it holds far more than it needs
to, because the registry it matches against contains values that are not secrets.
Treat #2364 as a stopgap until this lands.

## The problem

`SpriteEnv` registers the **whole sprite env** with `Redaction.add/2`
(`apps/fountain/lib/fountain/conversations/sprite_env.ex`, the register step before
return), and `ConversationServer`'s Claude API-key fallback does the same with
`fallback_env`. So the registry holds identifiers like `FOUNTAIN_CONVERSATION_ID` and
`FOUNTAIN_SANDBOX_ID` (UUIDs), plus every other env value of 8 or more bytes,
whether it is a credential or not.

Under #2364, a chunk whose end could be the start of **any** registered value is held
whole (up to 4 KiB) until the next chunk of its kind arrives or the turn ends. With UUIDs
registered, every hex character (`0-9`, `a-f`) is a possible first byte, so a large share
of ordinary chunks, including model reply text, is delayed by one chunk. The same cause
split `"live"` into `"liv"` / `"e"` in `ConversationServerSessionIsolationTest` on an
earlier revision of #2364, whenever the random sandbox UUID started with `e`.

Costs today:
- **Streaming latency.** Reply text arrives a chunk late, often at turn end, and in a stutter.
- **Replay dedup.** A line the carry rewrites no longer deduplicates on reattach, so
  its text can appear twice (documented in #2364).
- **Over-redaction.** A conversation's own id is printed as `[REDACTED]` wherever it appears.

## What to decide

1. **Register secrets, not the env.** Build the registry from what is actually
   sensitive: environment and vault secret values, brokered credentials, inference
   credentials, the callback token, and connection tokens. Leave out Fountain-issued
   identifiers and config (conversation/sandbox ids, base URLs, model names, broker
   placeholders). Open question: tenant env vars that are plain config rather than
   secrets. Today they are indistinguishable, so they stay registered unless
   environments gain a secret/non-secret distinction.
2. **Raise the bar for holding.** Even with only secrets registered, a one-byte prefix
   match is common. Options: hold only when the tail matches a prefix of at least N
   bytes (N=1 today), and accept that a split in the first N-1 bytes leaks at most
[...]

PR #2382, opened 2026-09-17
fix(conversations): sanitize raw output before persistence (#2372)

### LINKED ISSUE #2372: Raw stdout/stderr that is not valid UTF-8 raises out of the conversation server
Found while fixing #2364 (review round 5) and scoped out of that PR by the reviewer. This is the older case the [#2364 thread](https://github.com/managoat/fountain/pull/2364#discussion_r4037837683) said would be raised separately. It does not involve redaction: it happens with nothing registered for the conversation.

## The problem

A `stdout` or `stderr` chunk that is not valid UTF-8 on its own raises inside the `ConversationServer` callback that handles it:

```
** (Postgrex.Error) ERROR 22021 (character_not_in_repertoire) invalid byte sequence for encoding "UTF8": 0xc3
```

The path, on `main` at `ede81281b`, has no UTF-8 check or rescue anywhere along it:

- `conversation_server.ex:1062` and `:1069`: the `{:stdout, …}` and `{:stderr, …}` handlers pass the transport's bytes unchanged to `log_output/3` (`:2001`).
- `log_output/3` calls `Output.log/4` (`output.ex:109`), which reaches `Output.persist/3` (`output.ex:197`).
- `persist/3` calls `Conversations.log!/1` (`conversations.ex:2512`) with the bytes as `data`.
- `LogEvent.data` is `field :data, :string` (`log_event.ex:22`), a PostgreSQL `text` column in a UTF8 database. The insert in `insert_ordered_log_event!/1` raises.

Three kinds of input reach it:

1. **A chunk boundary inside a codepoint.** The transport cuts by bytes. `"caf" <> <<0xC3>>` followed by `<<0xA9>> <> " au lait\n"` is valid output cut in the wrong place; each row is invalid alone.
2. **Output in another encoding**, for example latin-1 from a tool or a setup script.
3. **Binary output**, for example a command that prints a file.

Claude, Codex and OpenCode turns run over ACP, whose text arrives as decoded JSON and is always whole UTF-8. So the exposed surface is the adapter's `stderr`, and raw `stdout` when the peer is gone (the comment at `conversation_server.ex:1063` says that handler exists to keep adapter diagnostics).

## Reproduction

This test fails with the error above. Both cases raise; nothing is registered with `Redaction`.

```elixir
defmodule Fountain.Conversations.RawOutputInvalidUtf8Test do
  use Fountain.DataCase, async: true
  alias Fountain.Conversations.{Output, Redaction}

  setup do
    user = insert_verified_user()
    conv = insert_conversation(user_id: user.id)
    turn = insert_turn(conv, status: "running")
    {:ok, ctx: %{conversation_id: conv.id, turn_id: turn.id, user_id: user.id}}
  end

  test "a chunk boundary inside a codepoint", %{ctx: ctx} do
    assert Redaction.lookup(ctx.conversation_id) == []
    output = Output.log(%Output{bytes: 0}, ctx, "stdout", "caf" <> <<0xC3>>)
    Output.log(output, ctx, "stdout", <<0xA9>> <> " au lait\n")
  end

  test "a lone continuation byte on stderr", %{ctx: ctx} do
    Output.persist(ctx, "stderr", <<0xA9>> <> " visible")
  end
end
```

## What is and is not verified

- **Verified:** the test above raises `22021` on the #2364 branch at `e87b40e60`, through the real test database. The files on the path are the same on `main` by source.
- **Not verified:** the test has not been run on a `main` checkout.
- **Not verified:** whether the Sprites transport, or any other provider, ever delivers a chunk that ends inside a codepoint. If it never does, case 1 is theoretical and cases 2 and 3 remain.
- **Not verified:** what a live conversation does after the raise. By source the exception leaves the `handle_info` callback, so the server would exit and the turn would fail; nobody has watched that happen.

## Options

- **Make each row valid before the insert.** `sandbox_files.ex:811-820` already has both shapes: base64 for bytes that are not text, and a latin-1 recode for a diff hunk. A row could be scrubbed (`String.replace_invalid/2`) or recoded the same way. This is one line in `Output.persist/3`, but a codepoint split across two chunks becomes two replacement characters instead of the character.
- **Carry an incomplete trailing codepoint to the next chunk of the same stream**, at most 3 bytes, and scrub only what is still invalid after that. This keeps case 1 lossless. It needs a flush at the turn's end, which `RedactionCarry` already has.

Either one belongs before #2366 replaces the redaction carry, or inside it: #2364 aligns only the cuts the carry itself makes, and leaves a chunk the transport cut exactly as it arrived (`redaction_carry_test.exs`, "a raw chunk the fail-safe did not cut is written as it arrived").

jhgaylor:
Raw adapter stdout/stderr can contain incomplete UTF-8, non-UTF-8 text, or binary data that PostgreSQL rejects, raising out of the conversation server. The log writer now replaces invalid UTF-8 and NUL with `?` after redaction, so stored events and their broadcasts contain valid text and the turn keeps running. Redaction still matches registered binary secrets before sanitization.

This uses per-row replacement: a character split across rows is replaced on each side. The one-byte replacement cannot expand output beyond the existing byte budget. Valid Unicode and terminal control characters are preserved. The fix is independent of #2366 and introduces no carry state or migrations.

Validation:
- Reproduced PostgreSQL error `22021` with the new regression tests on unchanged main.
- Passed 61 focused tests and 2 property checks covering persistence, broadcasts, redaction, turn-end flush, byte limits, arbitrary binary data, and a running conversation server.
- Passed all 9 `mix precommit` stages, including Dialyzer, Sobelow, and production release assembly; changelog and diff checks also passed.

Closes #2372.


BinaryBourbon (approved):
Independent Codex correctness/reliability review: **APPROVE**.

Reviewed exact head `cfde4c13feccc7b634b6a78b23a3daab89c34335` against PR merge-base `2cd2d6e277949a6163096bc69e6bc2fcc724f8be`; trusted contributor guidance read from approved base `f43a536010fdee9f57019cfc0bcb363eb36c0714`. Verified every changed snapshot file byte-for-byte against the head commit. Review complete; no concrete introduced correctness or reliability defects found.

Complete changed-path coverage:

- `apps/fountain/lib/fountain/conversations.ex`: traced binary redaction before sanitization, strict output/stdout/stderr guard, invalid UTF-8 and NUL replacement, ordered insertion and bounded-turn write authorization. Raw setup/provisioning callers also reach this writer. ACP and stage payloads do not match the sanitizer.
- `apps/fountain/lib/fountain/conversations/output.ex`: checked documented per-row replacement against implementation; traced transport handling, raw redaction carry, turn ownership, end-of-turn flush, byte counting/cap, cumulative reload and persisted-event broadcasts. Sanitization adds no carry state and cannot expand the pre-sanitization payload.
- `apps/fountain/test/fountain/conversations/conversation_server_log_budget_test.exs`: reviewed the real-server regression and its harness. Messages are synchronized before asserting that the running turn survives and all three sanitized rows persist. Existing global budget setup remains isolated by the synchronous server harness.
- `apps/fountain/test/fountain/conversations/output_utf8_test.exs`: reviewed all examples and the arbitrary-byte property, including valid Unicode/control preservation, binary secrets, split secrets, held-output flush, exact budget threshold, and persisted/broadcast equality.
- `changelog.d/2372-raw-output-utf8.md`: behavior and deliberate replacement at row boundaries match the implementation.

Validation: inspected coordinator-produced focused test output from the exact clean head: **2 properties, 85 tests, 0 failures**, seed 2382, using the pinned mise toolchain and disposable PostgreSQL database. This reviewer performed independent source/caller/assertion inspection and exact-head blob verification; the coordinator executed the test suite. No source changes made. Full umbrella suite and precommit were not rerun by this reviewer; The coordinator verified successful [CI](https://github.com/managoat/fountain/actions/runs/35272431587) and [secret scanning](https://github.com/managoat/fountain/actions/runs/35272431591) for this exact head.

This is an independent Codex review, not a Review Loop App execution.

BinaryBourbon (approved):
Verdict: APPROVE
Reviewed head: `cfde4c13feccc7b634b6a78b23a3daab89c34335`
Review: independent Codex API, behavior, and test-quality review. Complete: true. Findings: none.

I reviewed the complete five-path PR diff against merge-base `2cd2d6e277949a6163096bc69e6bc2fcc724f8be`, using contributor/API policy and relevant ADR 0014 requirements from approved base `f43a536010fdee9f57019cfc0bcb363eb36c0714`.

Coverage:

- `apps/fountain/lib/fountain/conversations.ex`: common writer, redaction-before-sanitization, stdout/stderr guard, unchanged ordered insertion and execution guard. The sanitizer also covers direct provisioning/setup callers. It produces valid UTF-8 without NUL and cannot expand the redacted row.
- `apps/fountain/lib/fountain/conversations/output.ex`: persistence/broadcast call chain, redaction carry and turn-end flush, conservative byte accounting, and documentation of per-row replacement. Broadcasts carry the returned stored event, so live delivery and replay agree.
- `apps/fountain/test/fountain/conversations/conversation_server_log_budget_test.exs`: regression delivers real server mailbox messages and verifies the same turn remains running with the expected ordered persisted stdout/stderr rows.
- `apps/fountain/test/fountain/conversations/output_utf8_test.exs`: inspected all assertions covering split UTF-8, invalid encodings, NUL, valid Unicode/control preservation, binary-secret matching, split-secret carry, flush, byte limits, and arbitrary-byte persistence/broadcast agreement.
- `changelog.d/2372-raw-output-utf8.md`: accurately describes replacement and the deliberate loss of a codepoint split across stored rows.

I traced public REST event rendering and conversation/account/team SSE rendering: each uses stored `event.data`, with JSON encoding preserving valid controls without altering SSE framing. LogEvent/StreamLogEvent still promise string data, and no fields, enum values, cursors, filters, terminal states, or permissions change. ACP-only structured transcript/reply rendering remains intact. The live ACP peer reports invalid nonprotocol lines as stdout, which reaches the same writer. TypeScript, Python, Elixir, both Swift consumers, and the Go CLI decode the existing JSON/string contract; Buzz does not introduce a separate affected log decoder. No client/schema regeneration or conformance skip is warranted by this patch.

Validation evidence: I inspected the coordinator's exact-head focused run results: **2 properties, 85 tests, 0 failures** (seed 2382). The coordinator also verified successful exact-head [CI](https://github.com/managoat/fountain/actions/runs/35272431587) and [secret scanning](https://github.com/managoat/fountain/actions/runs/35272431591). I independently inspected source and regression assertions; I did not duplicate the suite or run a load benchmark. The per-row Unicode loss is explicit, tested, and improves previously failing input rather than breaking a successful caller contract.

BinaryBourbon (approved):
Independent Codex security review of **cfde4c13feccc7b634b6a78b23a3daab89c34335**: **APPROVE**. No introduced security findings.

Reviewed all five changed paths: the conversation writer, Output documentation, both changed test files, and the changelog. Traced raw output through secret carry/redaction, turn-end flush, persistence, broadcasts, and tenant-scoped event readers. Redaction precedes sanitization; broadcasts use the persisted event; no tenant or credential authority boundary changes.

Validation: 40 binary/overlapping-secret split cases with stream interleaving, bytewise secret delivery, separate tenant registries, and all 256 single-byte sanitizer inputs passed independent diagnostics. The cross-tenant SSE denial regression passed (1 test); the coordinated exact-head focused suite passed (85 tests, 2 properties). [Exact-head CI](https://github.com/managoat/fountain/actions/runs/35272431587) and [secret scanning](https://github.com/managoat/fountain/actions/runs/35272431591) passed.

No production resources were used. This is an independent review, not an official Review Loop App review.

PR #2383, opened 2026-09-17
fix(flags): say so when a built feature's flag is missing from PostHog (#2347)

### LINKED ISSUE #2347: Connections is dark in production: the `connections` flag was never created in PostHog
Found on 2026-09-16 while running the deployed suite's `secrets` profile
against production.

## What happens

`POST /api/secret-bindings` answers 404 `brokerage_not_enabled` for every
account on `https://managoat.com`, including a comped, brokered, verified one.
`GET /api/secret-bindings` answers 200, so an account can list and revoke
bindings it does not have and cannot create. `/api/auth/me` reports
`brokered: true` and `connections_enabled: false`.

## Why

`Fountain.Connections.enabled_for?/1` is
`Fountain.Broker.configured?() and Fountain.FeatureFlags.enabled?(:connections, user_id)`.

The first half is true: `BROKER_TENANTS=*` since 2026-09-04.

The second half asks PostHog for a flag keyed `connections`, and **no flag
with that key existed in the project**. Rule 2 in `Fountain.FeatureFlags`
applies wherever `POSTHOG_PROJECT_API_KEY` is set, so production got a real
answer of "off" rather than falling through to rule 4's `@on_without_posthog`
exception. `feature_flags.ex:44-56` describes that exception as being for
flags that gate "a **shipped** feature rather than an unfinished one", listing
`connections` among them — which reads as an intent that this feature is on
for a self-hoster and, presumably, here.

The result is that the feature is dark in production while the code treats it
as shipped, and nothing logs an error, because `FeatureFlags` fails closed: a
missing flag looks exactly like a flag deliberately turned off.

## What was done

The flag now exists and is scoped to the two deployed-suite verification
accounts only, so it changes nothing for any other account:
[890169](https://us.posthog.com/project/570897/feature_flags/890169), matching
`distinct_id` exactly against `48644c86-…` and `e7cd096f-…`, rollout 100,
`evaluation_runtime: all`. That was the minimum needed to unblock the
`secrets`, `mcp` and `webhooks` profiles; it is deliberately **not** a rollout
decision.

## What to decide

1. Is Connections meant to be available to production accounts? If yes, widen
   890169's release conditions. If no, the `@on_without_posthog` comment is
   misleading and should say so.
2. A shipped feature can be switched off in production by the absence of a
   flag in an analytics project, with no error anywhere. Whether that is worth
   a guard — a boot-time warning when a flag in `@on_without_posthog` is
   missing where PostHog is configured, for example — is a separate call.
3. `GET` succeeding while `POST` 404s is deliberate (#1693 keeps the doors
   that take a credential away open), but an account that never had brokerage
   reads it as an empty list rather than a closed door.

Related: ADR 0019, #1693, #1620.

BinaryBourbon:
Closes #2347.

Connections was dark on production because the `connections` flag had never
been created in PostHog. The flag exists now
([890169](https://us.posthog.com/project/570897/feature_flags/890169)), scoped
to the two deployed-suite verification accounts. This PR settles the three
things #2347 left to decide, none of which change who has the feature.

**1. The rollout stays per account.** `docs/reference/feature-status.md`
already lists Connections as Alpha, off by default, "ask us to turn it on for
your account", and 890169's release conditions match that. What was wrong was
the `@on_without_posthog` comment, which described the list as being for a
**shipped** feature and so read as though `connections` were on for everyone
wherever a PostHog is configured. It now says that membership is about a
deployment with no PostHog at all, and that the flag's release conditions are
the only thing deciding who has it here.

**2. A missing flag is no longer silent.** `Fountain.FeatureFlags` fails
closed, so "PostHog says no" and "PostHog has never heard of this key" were
the same answer at the call site, with nothing logged either way. For a flag
in `@on_without_posthog` the second one is a mistake in the analytics project,
not a rollout decision. PostHog omits a flag it does not have and includes one
it has and says no to, so the two are distinguishable in the answer we already
fetch: `warn_undefined/2` logs an error naming the flag, rate-limited to once
per `@fresh_ms` like the `$feature_flag_called` capture, and nothing else
changes. It still fails closed.

This also catches the `evaluation_runtime: server` trap — a project-token
evaluation never sees a server-only flag, so it arrives absent and was
likewise indistinguishable from one that was never created. That cost a day on
`openai_compat` in August.

Only a real answer can say a key is missing: `answered_flags/1` separates
PostHog's own answer (fresh, or cached, or stale-cached during an outage) from
an empty map produced by a failed lookup with nothing cached, and the check is
not reached for the latter. `flags_for/1` is unchanged from the outside.

**3. `GET` 200 / `POST` 404 stays as it is.** It is deliberate: #1693 keeps
the doors that take a credential away open, because an account whose flag goes
off keeps every credential already brokered into its sandboxes. The only
problem was that nothing said so, and an unenrolled account reads the empty
list as "nothing here" rather than "closed door". Documented in
`feature-status.md` and on `Connections.manageable_for?/0`, both pointing at
`/api/auth/me`'s `connections_enabled` as the direct answer.

## Tests

Seven tests in `feature_flags_test.exs`. Verified by reverting: with the
warning disabled the three that assert it fail; with the check widened to
every flag, or to the unanswered path, the two negative tests fail instead.


jhgaylor (changes requested):
REQUEST_CHANGES — independently reviewed all five changed paths at `c7a200fe7d96ebcb5216c15b0345cefce11f7869` for correctness, cache behavior, and concurrency.

**P2: Correct the documented connection creation operation** (`docs/reference/feature-status.md:27-28`). The new text says posting to `/api/connections` returns `404 brokerage_not_enabled` when the flag is off. There is no `POST /api/connections` route: `router.ex:376-383` defines GET and DELETE only. Connecting an account uses the browser OAuth flow `/connections/:provider/start`, so the documented POST remains unsupported after enrollment. Limit this POST/error example to `/api/secret-bindings` and describe the browser flow for connections. If showing provider creation, use `POST /api/connection-providers` with `connections_not_enabled`. The preceding “only … create” wording should also account for edits that retarget or re-enable credential access, which existing guards and rollout tests refuse when the flag is off.

I found no blocking regression in flag values, the public `flags_for/1` result, fresh/stale cache behavior, override precedence, or cache-key namespaces. A local diagnostic using the pinned source and Req/Analytics stubs passed eight cache/provenance/override cases; 30 runs of 100 concurrent readers each emitted one warning. This was a pure local diagnostic, not the application test suite or a live PostHog evaluation.

The coordinator ran the exact-head focused FeatureFlags, Analytics, rollout, and connection/controller suites: **86 tests, 0 failures**, seed 2383. [CI](https://github.com/managoat/fountain/actions/runs/35275808021) and [secret scanning](https://github.com/managoat/fountain/actions/runs/35275808023) passed on this same SHA.

jhgaylor (changes requested):
Independent Codex security and rollout review: **REQUEST_CHANGES** at `c7a200fe7d96ebcb5216c15b0345cefce11f7869`.

**P2 — Correct the omitted-flag diagnosis and global override advice** (`apps/fountain/lib/fountain/feature_flags.ex:188-191`). An intentionally disabled flag is omitted from PostHog's evaluation: its server filters inactive flags before producing results. The new error therefore treats an intentional shutdown as a missing definition or server-only configuration and recommends `FEATURE_FLAGS_ON=connections`. That override takes precedence over PostHog and enables every brokered account, bypassing the intended per-account rollout. Quota-limited or partial responses can also omit results. Describe the observed absence without asserting its cause, account for disabled/filtered/incomplete evaluation, and direct operators to inspect configuration while retaining rollout conditions. Remove the global override as the ordinary repair, or explicitly identify its deployment-wide effect. Add regressions for these response cases. [PostHog inactive-flag filtering](https://github.com/PostHog/posthog/blob/201b40179b9ea2ed30af348e459e484b131c41cd/rust/feature-flags/src/handler/flags.rs#L241-L247), [quota-limiting behavior](https://posthog.com/docs/feature-flags/cutting-costs#quota-limiting).

**P2 — Document real creation routes and edit gates** (`docs/reference/feature-status.md:23-28`). There is no `POST /api/connections`; connections are created through browser OAuth. `POST /api/connection-providers` returns `connections_not_enabled` when gated, while `POST /api/secret-bindings` returns `brokerage_not_enabled`. The new paragraph conflates them and would mislead enrolled callers into using an unsupported endpoint. Provider edits/discovery and binding edits or re-enablement are also gated, so “Only ... create” is inaccurate. Limit the POST/error example to secret bindings, describe the browser flow for connections, and explain that adding/changing credential access is gated while reads and removals remain available.

Reviewed all five changed paths, including both documentation files and the complete new test coverage. Traced tenant-scoped reads, OAuth ownership, management/creation gates, per-user fresh/stale caching, overrides, parsing and analytics. No newly introduced cross-tenant access or secret disclosure was found.

Four independent diagnostic tests passed using the exact FeatureFlags source with in-memory Req/Analytics stubs. They reproduce the incorrect omission warnings, verify two-tenant cache isolation and uncached outage denial, and confirm the override enables previously denied accounts. These were local diagnostics, not a live PostHog evaluation; no production credentials or project state were accessed.

The coordinator ran the exact-head focused FeatureFlags, Analytics, rollout, and connection/controller suites: **86 tests, 0 failures**, seed 2383. [CI](https://github.com/managoat/fountain/actions/runs/35275808021) and [secret scanning](https://github.com/managoat/fountain/actions/runs/35275808023) passed on this same SHA.

jhgaylor (changes requested):
# Independent Codex API/product review — PR #2383

Verdict: **REQUEST_CHANGES**. Source review complete at `c7a200fe7d96ebcb5216c15b0345cefce11f7869` (approved policy base `0facb18a11c3503dfa5c77e320db07aab4eb70a0`; PR merge base `2cd2d6e277949a6163096bc69e6bc2fcc724f8be`). This is an independent Codex review, not a Review Loop App attestation.

## Findings

1. **[P2] Do not diagnose an omitted evaluation result as a missing flag definition.** `apps/fountain/lib/fountain/feature_flags.ex:184-191` (`warn_undefined/2`, especially the exhaustive “Either” diagnosis).

   Turning off PostHog's **Enable feature flag** switch intentionally removes the flag from evaluations; this is distinct from an active flag evaluating to `false` because a person does not match its release conditions. PostHog documents disabled flags as returning undefined/null, and its server explicitly filters `!active` flags. Consequently, an operator deliberately disabling `connections` now receives a recurring ERROR saying the definition is missing or server-only, followed by advice to create/widen it or globally override the shutdown. That is a concrete regression in the diagnostic this PR introduces. [PostHog disabled-flag documentation](https://posthog.com/docs/feature-flags/creating-feature-flags#enable-feature-flag-optional), [server filtering](https://github.com/PostHog/posthog/blob/201b40179b9ea2ed30af348e459e484b131c41cd/rust/feature-flags/src/handler/flags.rs#L241-L257).

   HTTP 200 also does not imply complete evaluation: PostHog's response contract explicitly permits missing flags when `errorsWhileComputingFlags` is true; quota limiting returns an empty flags map with `quotaLimited`. Existing `fetch/1` and `parse/1` discard these markers, and the newly introduced `:answered` classification gives those incomplete maps the same configuration diagnosis, including after caching. The prior fail-closed evaluation is unchanged; the new false error is the introduced defect. [Response contract](https://github.com/PostHog/posthog/blob/201b40179b9ea2ed30af348e459e484b131c41cd/rust/feature-flags/src/api/types.rs#L201-L211), [quota response](https://github.com/PostHog/posthog/blob/201b40179b9ea2ed30af348e459e484b131c41cd/rust/feature-flags/src/handler/billing.rs#L15-L33).

   **Remedy:** Make absence diagnostics accurately describe an unavailable result, including the intentional disabled/filtering cases; preserve response error/quota metadata through cached diagnostic eligibility so incomplete answers do not diagnose missing configuration. Add disabled/partial/quota response regressions. Avoid claiming a per-person (possibly stale) answer establishes current state for every account or advising a global override as the ordinary repair. The project-token explanation in the comment and changelog also needs correction: runtime comes from an explicit request field or headers/SDK detection, not the token class; an unknown runtime excludes both client-only and server-only flags. [Runtime detection and filtering](https://github.com/PostHog/posthog/blob/201b40179b9ea2ed30af348e459e484b131c41cd/rust/feature-flags/src/handler/flags.rs#L73-L134).


2. **[P2] Document the actual creation endpoint and edit gates.** `docs/reference/feature-status.md:23-28` (especially “POST to either”).

   There is no `POST /api/connections`: `router.ex:371-383` explicitly says connecting requires the browser OAuth flow and exposes only GET/DELETE API operations. A caller following this paragraph receives a route-not-found response even after enrollment, rather than the promised `brokerage_not_enabled` capability signal. Actual provider creation is `POST /api/connection-providers`, whose disabled error is `connections_not_enabled`. Only `POST /api/secret-bindings` has the stated error. The same paragraph's “Only ... create” assertion also hides existing blocked PATCH operations: provider edits/discovery and binding edits or re-enablement require enrollment; only disabling a binding with `enabled: false` alone bypasses it (`connection_provider_controller.ex:233-249`, `secret_binding_controller.ex:161-196`; pinned by `connections_rollout_test.exs:114-157`).

   **Remedy:** Use `POST /api/secret-bindings` as the concrete POST example, describe connection creation as the browser OAuth flow, and describe the gate as adding/changing credential paths while reads and removals remain available. Keep `/api/auth/me.connections_enabled` as the supported availability check.


## Coverage and evidence

All five changed paths reviewed: `apps/fountain/lib/fountain/connections.ex` (doc maps correctly to `AuthMeController`); `apps/fountain/lib/fountain/feature_flags.ex` (all changed branches and public callers); `apps/fountain/test/fountain/feature_flags_test.exs` (all seven new tests, plus existing override/evaluation/cache/analytics assertions); `changelog.d/2347-missing-flag-warning.md` (same unsupported runtime/absence claim as finding 1); `docs/reference/feature-status.md` (finding 2).

Read trusted base `CLAUDE.md`, `CONTRIBUTING.md` API/validation guidance, relevant server conventions and ADRs 0019/0033. Traced `enabled?/2` through `Connections.enabled_for?/1`, management gates, AuthMe rendering/schema, and analytics cached flags. `flags_for/1` retains its binary input and map output; no in-repository production caller uses it directly. Reviewed corresponding committed contract, generated TypeScript/Swift models, four SDK connection/auth callers, Go CLI auth caller and Buzz/CLI search results. No changed wire schema, route, client method, omission policy, conformance behavior or SDK version requires regeneration here. Existing GET/read/revocation behavior and both AuthMe booleans are preserved.

No live PostHog project evaluation or credentials were accessed. External facts come from official docs and public source; public PostHog source revision recorded as `201b40179b9ea2ed30af348e459e484b131c41cd`. This reviewer did not duplicate the application suites; The coordinator ran the exact-head focused FeatureFlags, Analytics, rollout, and connection/controller suites: **86 tests, 0 failures**, seed 2383. [CI](https://github.com/managoat/fountain/actions/runs/35275808021) and [secret scanning](https://github.com/managoat/fountain/actions/runs/35275808023) passed on this same SHA. Tests in the PR do not model upstream disabled/partial/quota semantics.

BinaryBourbon:
Both findings were real. Fixed in `939e39754`, and the first one turned out to
have a third cause that none of the three reviews named.

## The diagnosis (P2, `feature_flags.ex`)

I checked the behaviour against our own PostHog project rather than reasoning
from the source links, using the public client token and a throwaway key
nothing reads (`zz_probe_2383_disabled`, created disabled, evaluated once,
deleted):

| Case | `/flags?v=2` answer | Old warning |
|---|---|---|
| Enrolled account, matches 890169 | present, `enabled: true` | silent |
| Any other account, no condition matches | present, `enabled: false`, `reason: no_condition_match` | silent |
| Flag switched **off** in PostHog | **absent** | fired, wrongly |
| Flag never created | absent | fired, correctly |

So the review is right about the consequence, by the route row 3 describes:
switching the flag off is the ordinary way an operator turns the feature off,
and it produced a recurring error insisting the key was missing or
server-only, then recommended `FEATURE_FLAGS_ON` — deployment-wide, which
would have turned Connections on for every brokered account, the opposite of
the intent.

The log now reports the observation and stops there: PostHog is not evaluating
this flag; it may not exist, may be switched off, or its evaluation runtime
may exclude the call; check the flag in PostHog. `FEATURE_FLAGS_ON` is gone —
it is deployment-wide, and this only fires where a PostHog is configured to be
asked. The "project token means a client-side evaluation" claim is dropped
from the comment and the changelog; the observation that a `server` flag never
arrives is kept without the mechanism.

Row 2 is worth recording, because it is the failure mode this guard could have
had and does not: a per-account rollout does **not** make the warning fire for
everyone outside it. A non-matching flag comes back present and `false`. The
docs sentence about a boolean flag returning `null`/`undefined` when no
condition matches does not describe this endpoint's map.

**Incomplete answers.** Correct, and now handled. `errorsWhileComputingFlags`
and `quotaLimited` were both discarded by `parse/1`. Completeness now travels
with the flags into the cache, and `answered_flags/1` grades an answer
`:complete`, `:partial`, `:stale` or `:unavailable`. Only `:complete` can
support a conclusion about what the answer does not contain. That also drops
the stale-answer diagnosis — an answer kept from before an outage describes
the project as it was, which was the over-claim in "for every account".

Four regressions added for exactly these: a partial answer, a quota-limited
empty answer, a cached incomplete answer read again inside the cache window,
and a stale answer served through an outage. Each verified by reverting: with
completeness ignored and `:stale` diagnosable, all four fail.

## The documentation (P2, `feature-status.md`)

Confirmed against `router.ex:371-383` and both controllers. There is no
`POST /api/connections`; the error atoms are `connections_not_enabled` for
providers and `brokerage_not_enabled` for bindings; and "only create" missed
the gated `PATCH` and `discover` paths. Replaced with a table of what is gated
and what stays open, matching what `connections_rollout_test.exs:114-157`
pins, plus a line that a connection is made through the browser flow and
`GET /api/connections/providers` says where to send the owner.

## Checks

`mix precommit` PASSED (9 stages). `scripts/test-docs.sh` exit 0. FeatureFlags,
Analytics, connections rollout and connections context: **103 tests, 0
failures**. The temporary PostHog flag was deleted; 890169 is untouched and
still scoped to the two verification accounts.

jhgaylor (approved):
Independent Codex correctness, cache, concurrency, and recovery review: **APPROVED**, with one nonblocking low-severity finding.

Source review complete (`complete: true`) at **`939e3975480f35fcb90d67981358bee6bbb91f55`**. Reviewed the entire five-path triple-dot PR diff from merge base `2cd2d6e277949a6163096bc69e6bc2fcc724f8be`, using trusted review policy and contributor guidance from approved base tip `0facb18a11c3503dfa5c77e320db07aab4eb70a0`. This is an independent Codex review, not a Review Loop App attestation. No repository source was edited and no GitHub review was posted by this worker.

**Nonblocking finding**

- **P3 / low — Claim the warning window atomically.** `kind: defect`; `needsHuman: false`; stable rule `feature-flag-warning-atomic-window`; semantic anchor `Fountain.FeatureFlags.warn_once/2`. At [feature_flags.ex:208–211](https://github.com/managoat/fountain/blob/939e3975480f35fcb90d67981358bee6bbb91f55/apps/fountain/lib/fountain/feature_flags.ex#L208-L211), concurrent requests can each observe an absent or expired marker in `stale?/2` before any request inserts the replacement. Each then emits the same ERROR, so the promised once-per-minute throttle is best-effort. A diagnostic using the exact source, 100 synchronized cached readers, and 100 empty-warning windows produced 1 warning in 64 windows, 2 in 23, 3 in 5, 4 in 2, 5 in 3, and 6 in 3. This affects duplicate log volume only; flag values, tenant separation, and access gates remain correct. Use an atomic ETS claim (`insert_new` for an absent marker, compare-and-replace for an expired marker), or serialize the warning claim; log only for the winner. A concurrency regression should cover both initial and expired markers. This is below the policy's medium blocking threshold.

**Earlier blockers independently checked**

The omission warning now names missing, disabled, and runtime-filtered possibilities without diagnosing one cause or recommending `FEATURE_FLAGS_ON`. Present `true` and `false` results remain silent. The new completeness bit survives both public cache readers and fresh cache hits; `errorsWhileComputingFlags` or a nonempty `quotaLimited` prevents omission diagnosis. Stale and unavailable answers also suppress that diagnosis. Recovery to a fresh complete answer resumes it. This matches the relevant camelCase markers in PostHog's [public response contract](https://github.com/PostHog/posthog/blob/master/rust/feature-flags/src/api/types.rs#L183-L210). No production PostHog request or state change was made.

The manual now identifies actual provider and secret-binding endpoints, their distinct error codes, provider edits/discovery, binding retargeting/re-enablement, and the disabling-only exception. It correctly describes connection creation as the browser OAuth flow and `/api/auth/me.connections_enabled` as the creation capability. Verified against the router, three API controllers, browser OAuth controller, AuthMe controller, and existing rollout regression assertions. The previously documented unsupported `POST /api/connections` operation is gone.

**Coverage**

| Changed path | Review performed |
|---|---|
| `apps/fountain/lib/fountain/connections.ex` | Entire documentation change; traced broker-plus-rollout creation capability and broker-only management capability through controllers and AuthMe. |
| `apps/fountain/lib/fountain/feature_flags.ex` | Entire diff and surrounding module: override order, no-PostHog defaults, nil users, response parsing, cache tuple propagation, freshness, partial responses, outage fallback/recovery, analytics properties, namespaced warning/event keys, ETS ownership, and concurrent reads. |
| `apps/fountain/test/fountain/feature_flags_test.exs` | All new cases plus existing evaluation, override, stale-cache, old-response-shape, analytics, and public-cache assertions; global configuration cleanup remains intact. Sequential throttle coverage does not catch the low-severity race above. |
| `changelog.d/2347-missing-flag-warning.md` | Entire fragment; observed diagnostic and unchanged access semantics agree with implementation, subject to the throttle race above. |
| `docs/reference/feature-status.md` | Entire change and surrounding page; all added route/error/gate claims traced to implementation. |

Also searched core, ee, and all sibling extensions for changed API/cache consumers. The public `flags_for/1` and `cached_flags/1` results remain maps; no production caller accesses the changed ETS tuple directly. Core gained no extension-module references, asynchronous work, persistence changes, or sandbox/auth/rate-limit modifications. Read trusted correctness instructions, CLAUDE, CONTRIBUTING, server conventions, documentation guidance, Review Loop skill/policy, and verification instructions from the approved base.

**Actual validation and limits**

- `mise exec -- elixir /private/tmp/pr2383-r2-qa-diagnostic.exs` exited 0. It loaded the exact source with in-memory Req/Analytics substitutes and verified partial/quota metadata through fresh cache hits, recovery, stale warning suppression, both public map APIs, two-user isolation, and static override precedence. Its concurrent-reader probe reproduced the nonblocking finding above. This was a standalone diagnostic, not the application suite or a live PostHog evaluation.
- `git diff --check 2cd2d6e277949a6163096bc69e6bc2fcc724f8be...939e3975480f35fcb90d67981358bee6bbb91f55` exited 0. Final local HEAD matched the supplied SHA and source worktree status was clean.
- Coordinator-reported exact-head focused application runs: **137 tests, 0 failures, seed 2383** (122 across FeatureFlags, Analytics, connections, rollout, and the three connection controllers, plus 15 direct connections-context tests). Evidence: `/private/tmp/pr2383-r2-focused-tests.log`.
- This worker did not rerun Mix against the shared build while coordinator testing was active. `rl-env` was absent locally, so the standalone diagnostic used the installed pinned mise toolchain. Direct `gh` reads failed due to sandbox network access; prior reviews/comments were read from coordinator-supplied JSON, and primary upstream response types were inspected through public web access. Coordinator reports exact-head [CI](https://github.com/managoat/fountain/actions/runs/35278859724) and [secret scanning](https://github.com/managoat/fountain/actions/runs/35278859720) completed successfully. Coordinator also completed 98 core/extension docs tests, bringing local verification to **235 tests, 0 failures**. Full coordinator evidence is `/private/tmp/pr2383-r2-verification.md`; the full twenty-command local Review Loop recipe was not rerun.

No blocking correctness or recovery defect remains in the reviewed revision. The approval accepts the isolated log-throttling race as a nonblocking issue under the supplied policy.

jhgaylor (approved):
# Independent Codex security and rollout review — PR #2383

Verdict: **APPROVED**. `complete: true`; `findings: []`.

Exact reviewed head: `939e3975480f35fcb90d67981358bee6bbb91f55`.
Trusted policy base: `0facb18a11c3503dfa5c77e320db07aab4eb70a0`.
PR diff: `2cd2d6e277949a6163096bc69e6bc2fcc724f8be...939e3975480f35fcb90d67981358bee6bbb91f55`.

This is an independent Codex review, not a Review Loop App attestation. The source worktree was read-only and clean at completion. No newly introduced defect requiring changes was found.

## Repairs verified

- The omitted-flag diagnostic now explicitly includes intentional disabling, removes the global `FEATURE_FLAGS_ON` repair advice, and directs the operator to inspect the flag. The code still returns the same per-account boolean; it does not broaden enrollment or change credential access.
- Response completeness is preserved through the fresh cache. `errorsWhileComputingFlags: true`, nonempty `quotaLimited`, failed lookups, and stale fallback do not produce the missing-evaluation error. An explicit `connections: false` remains silent. PostHog's public response source confirms the camelCase completeness/quota fields and permits missing results during evaluation errors. [Response contract](https://raw.githubusercontent.com/PostHog/posthog/master/rust/feature-flags/src/api/types.rs)
- Disabled and runtime-filtered flags are excluded by PostHog's public server implementation, consistent with the repaired text. [Server filtering](https://raw.githubusercontent.com/PostHog/posthog/master/rust/feature-flags/src/handler/flags.rs), [disabled flag documentation](https://posthog.com/docs/feature-flags/creating-feature-flags#enable-feature-flag-optional)
- The manual now documents the real provider/binding API methods and error atoms, the browser OAuth connection flow, and the disabling-only binding PATCH exception. These match `router.ex`, `ConnectionProviderController`, `SecretBindingController`, `ConnectionController`, `ConnectionsController`, and existing rollout regression assertions. `GET /api/auth/me.connections_enabled` is computed from the authenticated user's ID through `Connections.enabled_for?/1`.

## Entire-PR coverage

| Changed path | Review result |
|---|---|
| `apps/fountain/lib/fountain/connections.ex` | Doc-only change; availability signal traced to AuthMe and the existing broker/rollout conjunction. Tenant-scoped provider/connection reads and owner attribution remain unchanged. |
| `apps/fountain/lib/fountain/feature_flags.ex` | Reviewed complete module: overrides, distinct ID extraction, remote request, parser, response provenance, ETS cache keys/values, fresh/stale fallback, warning limiter, and analytics cache consumer. Cache data remains keyed by the individual user's ID; shared warning keys contain only a known flag name and no tenant data. New logging contains fixed text plus the allowlisted flag key, with no ID, secret, credential, or remote payload. No new SQL, shell, path, callback, rendered-HTML, SSRF, or credential-forwarding boundary. |
| `apps/fountain/test/fountain/feature_flags_test.exs` | Reviewed all new assertions and existing evaluation/override/cache/analytics tests. Partial, quota, cached partial, stale, explicit on/off, and omission paths are covered; global environment mutation remains serialized and restored. |
| `changelog.d/2347-missing-flag-warning.md` | Accurately describes the new diagnostic, incompleteness/outage exceptions, and unchanged access decisions. |
| `docs/reference/feature-status.md` | Checked all introduced endpoint/gating statements against routes and controllers, including provider edits/discovery, binding retarget/re-enable versus disable-only, existing reads/removals, and OAuth. |

Relevant callers inspected include API gates, AuthMe, Connections/SecretBindings LiveViews, OAuth start/callback (signed session state and scoped provider lookup), and `Analytics.feature_flag_properties/1`. The changed boolean/cache behavior introduces no new cross-tenant resource lookup, no expanded authority, and no change to existing secret encryption/audit flows. Read trusted-base `CLAUDE.md`, `CONTRIBUTING.md`, `.github/review/tenant-isolation.md`, `contributing/server.md`, and `contributing/docs.md`.

## Actual diagnostic evidence and limits

`mise exec -- elixir /private/tmp/pr2383-r2-security-diagnostic.exs` completed successfully: **7 tests, 0 failures**, seed **2383**. It compiled the exact reviewed FeatureFlags source with local in-memory Req/Analytics stubs and exercised:

1. Disabled/omitted flag text without global override advice or ID/key disclosure.
2. Partial/quota markers in both parsed response shapes through a fresh cache, then expiry/refetch.
3. Opposite answers for two tenants, isolated analytics cache values, stale fallback, and a third uncached tenant denied during outage.
4. Stale omission without a current-configuration diagnosis.
5. Explicit on/off rollout answers without errors and nil identity without a remote call.
6. Shared warning window across two users without sharing their cached answers.
7. Unchanged static override precedence without a remote call.

`git diff --check` passed for the entire triple-dot diff. These diagnostics do not exercise real HTTP, PostHog project configuration, the database, or application startup. No production credentials, PostHog queries, or provider resources were used. Official public sources were read; their master URLs are mutable. The public cost guide shows a different snake_case quota example, but the inspected current response source uses camelCase, and no supported-endpoint evidence established a further regression; this discrepancy is not a blocking finding. Likewise, the server's evaluation-context filtering returns no exclusions when the request supplies no contexts, as Fountain currently does.

Coordinator verification at this exact head: **235 local tests, 0 failures** (137 focused application tests plus 98 core/extension documentation tests); changelog and diff checks passed. [CI run 35278859724](https://github.com/managoat/fountain/actions/runs/35278859724) and [secret scan 35278859720](https://github.com/managoat/fountain/actions/runs/35278859720) succeeded. Evidence: `/private/tmp/pr2383-r2-verification.md`. The full 20-command local Review Loop recipe was not run; this review does not claim otherwise.

jhgaylor (approved):
# Independent Codex API/product re-review — PR #2383

**APPROVED** — source review complete (`complete: true`, `findings: []`) at **939e3975480f35fcb90d67981358bee6bbb91f55**. Trusted policy/base tip: `0facb18a11c3503dfa5c77e320db07aab4eb70a0`; PR merge-base: `2cd2d6e277949a6163096bc69e6bc2fcc724f8be`. This is an independent Codex review, not a Review Loop App attestation. The coordinator owns publication and final integration-check evidence.

No introduced blocking defect remains in the complete five-path diff. I independently checked the original changes and the repairs, rather than treating the prior findings as the whole review.

## Prior findings verified repaired

- **Omitted-result diagnosis and rollout advice:** `feature_flags.ex:188-196` now describes the observed omission, explicitly includes a deliberately disabled flag among possible causes, and directs operators to inspect PostHog. It no longer recommends `FEATURE_FLAGS_ON` as the ordinary repair or claims the project-token class determines the evaluation runtime. The static override remains deployment-wide and takes precedence exactly as before, but the error no longer recommends bypassing per-account rollout.
- **Partial/quota/stale answers:** `complete?/1` checks `errorsWhileComputingFlags` and `quotaLimited`. The completeness bit survives per-user caching and is used by both fresh-fetch and fresh-cache paths; partial, stale fallback, and failed uncached results cannot reach the omission warning. The new tests cover the partial and quota cases, cached partial results, and stale outage suppression. The disabled and missing-definition cases have the same wire shape and now receive an accurate shared diagnostic.
- **Creation routes and edit gates:** `feature-status.md:26-43` accurately names `POST /api/connection-providers`, provider PATCH/discovery, and `POST /api/secret-bindings`; it uses the controllers' distinct `connections_not_enabled` and `brokerage_not_enabled` errors. It documents the browser OAuth start flow instead of an invented `POST /api/connections`. Provider modifications and binding retarget/re-enable are gated, while reads, removal, and a binding-only disable remain available when brokerage is configured. The existing rollout tests corroborate these distinctions.

## Whole-PR coverage

| Changed path | Review result |
|---|---|
| `apps/fountain/lib/fountain/connections.ex` | The new doc correctly points clients to `GET /api/auth/me.connections_enabled`; traced to the actual AuthMe renderer, schema, and `enabled_for?/1`. Management still depends only on configured brokerage. |
| `apps/fountain/lib/fountain/feature_flags.ex` | Reviewed all changed branches, warning eligibility/rate-limit namespace, response parsing, override/no-PostHog/nil-user behavior, per-user fresh/stale cache, and analytics enrichment. Public `flags_for/1` and `cached_flags/1` still return maps of flags, while only the internal cache payload gains metadata. Flag values and rollout gates are preserved. |
| `apps/fountain/test/fountain/feature_flags_test.exs` | Reviewed the full suite, including the new omission, intentional false, override exclusion, suppression, and cached partial-response assertions. Existing tests continue to exercise stale on/off values, legacy response shape, cache-only analytics, and capture suppression. |
| `changelog.d/2347-missing-flag-warning.md` | Describes the revised error and incomplete/stale suppression without the previous unsafe override or token-class advice. |
| `docs/reference/feature-status.md` | Compared the complete page with router entries, API controllers, browser OAuth controller, AuthMe, and the Google extension's broker-only runtime gate. New instructions match implemented operations. |

Trusted instructions read from the supplied base include `CLAUDE.md`, `CONTRIBUTING.md` (including API and validation sections), `.github/review/api-compatibility.md`, `.github/review-loop.yml`, the Review Loop skill, server/docs conventions, and relevant ADR 0019/0025/0028/0033 content.

Traced callers through Connections, AuthMe, console capability assignments, provider/binding guards, and `Analytics.feature_flag_properties/1`. Compared the relevant committed wire contract and generated TypeScript/Swift AuthMe fields, four SDK auth/connection/provider callers, and the Go CLI auth reader; searched Buzz CLI and integrations for affected callers. No wire schema, endpoint, public SDK signature, conformance behavior, omission manifest, or version changes are required by this internal diagnostics patch. Existing callers receive the same booleans and response shapes.

## Primary upstream evidence

PostHog's [flags API documentation](https://posthog.com/docs/api/flags) explicitly describes an `enabled: false` result with `no_condition_match`, the incomplete-computation marker, and the v2 empty-map quota response with `quotaLimited: ["feature_flags"]`. These match the new classifier.

I independently fetched [PostHog's pinned public flag handler](https://github.com/PostHog/posthog/blob/201b40179b9ea2ed30af348e459e484b131c41cd/rust/feature-flags/src/handler/flags.rs): it excludes inactive/deleted flags before evaluation, detects runtime from the explicit request field or headers, and excludes runtime-specific flags when detection is unknown. It also returns no context-filter exclusions when no evaluation contexts are supplied, as in Fountain's request. This supports the warning's repaired causal wording. The [runtime documentation](https://posthog.com/docs/feature-flags/creating-feature-flags#evaluation-runtime-optional) separately confirms header/SDK-based detection. The downloaded handler is 38,458 bytes, SHA-256 `7cb62ae286b7507e90d398efc0aeb7b26173f4661a252e685d269b2d4a84d713`, and matches the earlier cached public-source copy.

## Diagnostics and limits

- This reviewer ran `git diff --check` across the exact triple-dot diff: passed. Final source worktree `git status --short` was empty.
- I inspected the coordinator's exact-head focused-test log: **122 tests, 0 failures**, covering FeatureFlags, Analytics, connections submodules, rollout, and controllers. The coordinator additionally reports **15 direct Connections tests and 98 core/extension documentation tests, all passing**, for **235 tests, 0 failures**, plus successful exact-head [CI](https://github.com/managoat/fountain/actions/runs/35278859724) and [secret scanning](https://github.com/managoat/fountain/actions/runs/35278859720). Its detailed evidence is `/private/tmp/pr2383-r2-verification.md`; the entire 20-command local Review Loop recipe was not duplicated.
- I did not run a duplicate Mix build against the shared worktree; the coordinator was already testing it, and `rl-env` is absent in this worker. No test pass is inferred from setup. A sandboxed public-source fetch initially failed DNS; the authorized read-only network retry succeeded.
- No production PostHog evaluation, credentials, project queries, or project edits were used. No repository source was changed and no review was posted by this worker.

jhgaylor (approved):
**Independent adversarial simplicity and maintainability review — APPROVED**

Source review complete for PR #2383 at `939e3975480f35fcb90d67981358bee6bbb91f55`, covering the entire five-path diff from merge-base `2cd2d6e277949a6163096bc69e6bc2fcc724f8be`. Contributor and review policy came from trusted base `0facb18a11c3503dfa5c77e320db07aab4eb70a0`.

No blocking findings. I checked the original change and the repairs to the previous changes-requested verdicts:

- The log now reports the absent evaluation result and includes a disabled flag or runtime filtering among possible causes. It no longer recommends the deployment-wide `FEATURE_FLAGS_ON` override. An explicit `false` remains silent and denies access.
- The small cache-shape change preserves whether an answer was complete. Partial/quota-limited responses remain ineligible for the absence diagnostic after caching; stale answers during outages remain usable for flag values but cannot produce a fresh configuration diagnosis. Public `flags_for/1` and `cached_flags/1` still return maps, and the analytics caller still reads cached values without making a flag lookup.
- The response markers match the raw `/flags?v=2` contract: `errorsWhileComputingFlags` and `quotaLimited`. I checked the supplied pinned PostHog response/billing source and the [endpoint documentation](https://posthog.com/docs/api/flags#quota-limiting). The broader cost page uses a different spelling; the endpoint-specific contract and serializer support the implementation here.
- The manual now names the actual provider and binding POST/error pairs, provider edits/discovery, binding retargeting/re-enablement, the disable-only exception, and browser OAuth for creating a connection. These agree with the router, controller guards, rollout tests, and `/api/auth/me.connections_enabled`.

The new tests assert observable warning/denial behavior, cached incompleteness, stale outage behavior, and suppression within the warning window. Existing override precedence, legacy response parsing, flag cache, outage fallback, and analytics assertions remain intact. The implementation reuses the existing cache and timestamp helper without adding another process, configuration surface, or source of feature enrollment truth.

All changed paths reviewed:

- `apps/fountain/lib/fountain/connections.ex`
- `apps/fountain/lib/fountain/feature_flags.ex`
- `apps/fountain/test/fountain/feature_flags_test.exs`
- `changelog.d/2347-missing-flag-warning.md`
- `docs/reference/feature-status.md`

Validation: my `git diff --check` passed for the reviewed diff. I inspected the coordinator's exact-head verification record: **235 tests passed, 0 failures** (137 focused tests plus 98 documentation tests), with changelog validation passing. Exact-head [CI 35278859724](https://github.com/managoat/fountain/actions/runs/35278859724) and [secret scanning 35278859720](https://github.com/managoat/fountain/actions/runs/35278859720) succeeded. I did not duplicate those application suites or execute the full local 20-command Review Loop verifier. No production PostHog evaluation or project-state mutation was performed by this reviewer. This is an independent Codex source review, not a Review Loop App attestation.

PR #2385, opened 2026-09-17
chore: ignore the scratch directories that sit next to the tree

BinaryBourbon:
Found while sweeping for untracked work after #2383.

`work/` (per-session briefs, probes, throwaway test files) and
`scripts/destink/node_modules` (5 MB of npm state for a tool that is not on
main) were both untracked and unignored, which is one `git add -A` away from
the history — the same shape as the scratchpad symlink that reached main in
#2035.

Neither rule carries a trailing slash, for the reason the `/deps` comment in
the same file gives: a `dir/` pattern matches a directory only, so a symlink
named `work` would slip past it.

The stray `scripts/destink/` in my checkout is deleted. It held nothing but
`node_modules` — the tool itself lives on an unmerged branch, so the rule here
is prospective and does not hide the tool when that branch lands.

Verified with `git check-ignore work/x` rather than `work`: a directory
pattern does not match a path that does not exist yet, so testing the bare
name reports a false negative.


jhgaylor (approved):
Independent adversarial correctness/QA review of a64704ca51bc7a3207a8ea3199b4d33abcb6197f (base 02399178af5341117e4d03068e87c4b2c30c672e): APPROVE.

Reviewed the complete .gitignore diff and relevant repository guidance. A disposable Git fixture passed 96 assertions covering files, directories, symlinks, root anchoring, nested paths, source/manifests remaining visible, existing tracked files and actual git add -A behavior. No correctness defects found.

This is a manual Codex source review; it does not claim Review Loop service verification.

jhgaylor (approved):
Independent adversarial security review of `a64704ca51bc7a3207a8ea3199b4d33abcb6197f`: APPROVE.

Reviewed the complete `.gitignore` diff against base `02399178af5341117e4d03068e87c4b2c30c672e`, repository security guidance, and relevant contributor instructions. No actionable security defects found.

An isolated Git fixture passed 16 diagnostic cases. Both new rules ignore files, directories, and live or dangling symlinks at their intended root paths. Nested and similarly named paths, tool source, and package manifests remain visible. Previously tracked paths remain visible when modified. Neither ignored location contains tracked files at this head.

Manual Codex review; this does not represent Review Loop service execution or full application verification.

jhgaylor (approved):
Independent adversarial compatibility review of `a64704ca51bc7a3207a8ea3199b4d33abcb6197f`: APPROVE.

Reviewed the entire `.gitignore` change against repository conventions and tool/source paths. No actionable compatibility defects found. A disposable Git fixture passed 18 assertions covering `git add -A`, tracked-file updates, source/manifest/lockfile visibility, path boundaries, and file/directory/symlink exclusions. Prospective `scripts/destink` source remains addable; the new npm rule excludes its dependency path.

Manual Codex review; no application suites run for this ignore-only change.

jhgaylor (approved):
Independent adversarial maintainability review of `a64704ca51bc7a3207a8ea3199b4d33abcb6197f`.

Reviewed the entire `.gitignore` change against the base contributor and simplicity guidance. No maintainability defects found. Both rules remain narrowly scoped; the prospective destink rule leaves its manifest, source, and tests visible. Issue #2035 supports the symlink rationale.

Validation: a disposable Git fixture confirmed root `work` is ignored as a file, directory, and dangling symlink; dependency contents are ignored while neighboring source paths remain visible. `git diff --check` passed. This is an independent Codex source review with focused diagnostics; no application suite or Review Loop verification was run.

Approved.

PR #2386, opened 2026-09-17
feat(machines): make destroying durable and drive the abandoned ones (ADR 0058, stage 9a)

jhgaylor:
Stage 9a of [ADR 0058](decisions/0058-the-machine-has-one-owner.md), off `main` after 8b (#2377). Tracker #2344.

Round 1 was seven reviews and eight blockers. **Six of the eight were a claim rather than code** — a comment, a changelog line or a docstring asserting something the code did not do, and in two cases that wrong sentence was the stated justification for a behaviour change. Every one is corrected below and the design question each opened is answered rather than papered over.

## The rule

`reset_requested_at` and `teardown_requested_at` are **durable intent**: refused whatever the lease says, surviving a crashed owner indefinitely. A `transition` stamp is **abandonable** — three doors clear a lease-less one on sight, which is what stops an owner that died mid-operation withholding a machine for up to 75 minutes (stage 6a round 1). The ADR says the columns "become `transition: destroying`", and they cannot until the stamp carries what they carry.

So **`destroying` becomes the one durable transition**: every reader refuses it regardless of lease, none clears it, `Lease.cas_update/4` keeps it through any write that does not retire the row, both fence writers stamp it beside the columns, `Quotas` counts it until terminal, and `SandboxReaper.sweep_fenced_teardowns/0` stops writing the row and becomes the driver that asks the owner to finish it. Nothing here is gated: `Destroy.run/2` takes a lease with the gate off too, so these rows exist either way.

## Behaviour changes, each with its symmetric case

1. **`Lease.cas_update/4` keeps a `destroying` stamp** on a live row through a write naming `transition` as anything else. Symmetric: a write that retires the row clears it (that is the finalize), a write to an already-terminal row clears it (that is `clear_stale_transition/2`'s tidying), a re-stamp passes with the new reason, and every other transition is still cleared. Keyed on the **row's** status, not the write's — keying it on the write would have stopped stale stamps ever clearing off terminal rows. `refuse_fenced: true` refuses the stamp as it refuses the columns.

[...]

PR #2377, opened 2026-09-17
feat(machines): bind, detach and retarget through the owner (ADR 0058, stage 8b)

jhgaylor:
Stage 8b of ADR 0058, *The machine has one owner* (#2344): **the binding is the owner's.**

## What

`Fountain.Machines.Binding` — the sixth protocol. `Machine.attach/3` is the only way a conversation row is bound to an existing machine (the `sandbox_id` attach door and, new, a teammate's successor), `Machine.detach/2` is the last-detach decision every terminate and release goes through, `Machine.retarget/3` is the one write of the machine's binding identity and skills record, and `Machine.bind_inference/2` is the machine's Codex auth binding. Alongside it: the owner **ends the turns it operates over** (a destroy ends every running turn on the machine; a ceiling park ends the one it cuts), `{:machine_gone, …}` is **sent by the owner only** (a lexical pin keeps `MachineEvents.tell_cotenants/5` callable from `lib/fountain/machines/` alone), a lease whose holder's **node is gone** is taken over before it expires, `ensure_up` carries the caller's **deadline**, and `release_stuck_sandboxes/0` no longer stops fleet-wide on one refused row. `SandboxIdentity` (no caller in any app) and `claim_sandbox/2` (no caller since 7b) are deleted.

## The protocol as shipped

```
attach   — one transaction, source lock then 4316 (lock_order_test), as main's create_attached_conversation/3:
             tenant, agent FOR SHARE, rotation unbind, machine FOR NO KEY UPDATE + statement_timestamp()
             attachable/5 under the lock: reset fence (409) → status → identity ×2 → runtime → live lease (503, no wait)
             deadline vs the clock the row was read with → :attach_expired
             limits, insert, inference reservation (skipped when the row resolved none), allowance
           after commit: after_conversation_created/1, the allowance event
           in-owner under the gate (@attach_timeout 20 s, deadline in the message); inline off
detach   — policy :mode: the teardown fence with terminating_conversation_id, plus two things it never did:
             refuse_busy (a live lease → :machine_busy, waited out 5 s → :sandbox_unavailable) and :deadline
             {:ok, :kept} | {:ok, :detached} (fenced, caller finishes) | destroy: true → Destroy.run → {:ok, :destroyed}
           policy :keep: a release — the conversation's row only, inline whichever way the gate is set
retarget — inline, inside the caller's transaction (Reapply holds 4316 across both writes); refuses a
           co-tenant on another identity and a build_fingerprint that is not the one expected
[...]

PR #2387, opened 2026-09-17
fix(conversations): give every GenServer callback a catch-all (#2380)

### LINKED ISSUE #2380: Two GenServers crash on any unmatched message; add a guardrail for the class
`use GenServer` supplies a `handle_info/2` that logs and ignores an unmatched message. **Defining any `handle_info` clause of your own silently removes it**, so every message the module does not match raises `FunctionClauseError` and takes the process down. The same is true of `handle_cast/2`: with none defined, the generated one stops the process with `{:bad_cast, …}`. Neither is visible at a call site or in review.

Found while reviewing #2377 (ADR 0058 stage 8b), where `Fountain.Machines.Machine` had exactly this hole — fixed there. A read-only sweep of `apps/fountain/lib` found two more.

### `Fountain.Conversations.ExecutionDeadlineWorker`

`execution_deadline_worker.ex:86,109,126` — clauses for `:tick`, `{ref, result}` when `is_reference(ref)`, and `{:DOWN, ref, :process, _, _}`; no catch-all. An `:EXIT`, a `:DOWN` that is not a process down, or any telemetry/PubSub message raises. It is an application-supervisor child so it restarts; the cost is the in-flight jobs map (its own comment says recovery covers that) plus a share of the supervisor's restart budget.

### `Fountain.Broker.Native.RequestLog`

`request_log.ex:165` — a single `handle_info(:flush, …)`. A crash loses the buffered broker rows and restarts: observability rather than correctness.

Checked and safe: `ConversationServer` (one clause delegating to `handle_execution_info/2`, whose last clause is `(_msg, state)`), `ExecutionTransport` (nine clauses, catch-all last), and every other module under `Fountain.Machines` (only `machine.ex` defines callbacks).

### What to do

1. Give both a catch-all that logs and continues, with a test each. One PR, not two.
2. Add a guardrail in the shape of the existing ones (`audit_guardrail_test.exs`, `extension_guard_test.exs`): walk `apps/fountain/lib`, and for every module defining at least one `handle_info/2` or `handle_cast/2` clause, require a final clause with an unbound first argument. Exempt LiveViews, where a crash is a reconnect, and take an explicit exemption list with a reason per entry so the next hole has to be argued for rather than merely added.

The guardrail is the point: the two fixes are one line each, and the class will recur without a standing check.

BinaryBourbon:
Closes #2380.

`use GenServer` supplies a `handle_info/2` that logs an unmatched message and
continues, and a `handle_cast/2` that stops with `{:bad_cast, message}`.
**Defining a clause of your own removes the generated one**, so from that
commit on every message the module does not match raises `FunctionClauseError`
and takes the process down. It is not a compiler warning, it is not visible at
a call site, and it does not show in a diff that adds a second clause to a
module whose first clause was fine — #2377 shipped exactly this hole in
`Fountain.Machines.Machine`.

## The holes

| Module | Callback | What reaches it |
|---|---|---|
| `Conversations.ExecutionDeadlineWorker` | `handle_info/2` | an `:EXIT` from a linked process, a `:DOWN` that is not a process down, any telemetry or PubSub message. `terminate/2` brutal-kills the in-flight tasks after its one-second grace, so a termination already in flight is lost and its `submitted` row waits out the 60-second recovery window before anything can touch it — plus a share of the application supervisor's restart budget |
| `Broker.Native.RequestLog` | `handle_info/2`, `handle_cast/2` | cast to from the proxy's connection processes; a crash loses the buffered rows, a gap in the egress log of every tenant proxying through the node |
| `Analytics.Sink` | `handle_cast/2` | a crash drops every event queued on the node |

The first two are the ones #2380 names. `Analytics.Sink`'s `handle_cast/2` is a
third, found by the guardrail below: the issue's sweep was `handle_info`-shaped
and the sink's `handle_info` already ends in a catch-all.

Each catch-all logs the message's **shape** — the tuple tag and arity, never
the payload. That is not decoration: with the fallback removed, the crash
report itself prints the whole term, and the revert probe on the sink shows a
PostHog event's `distinct_id` and properties in the log. The broker's request
log redacts paths on the way in, so a fallback that inspected the term would
have been the one place a raw one reached the node's logs.

`shape/1` is a private five-line helper in each of the three, rather than one
shared module. `Fountain.Machines.Machine` already carries a richer private
copy and would be the fourth caller, but #2386 (ADR 0058 stage 9a) is open
against that file; folding all four into one helper is the follow-up once it
lands, not a reason to conflict with it here.

## The guardrail

`apps/fountain/test/fountain/genserver_catch_all_guardrail_test.exs`, in the
shape of `audit_guardrail_test.exs` and `extension_guard_test.exs`: it parses
every `.ex` under `apps/fountain/lib` and, for each module defining at least
one `handle_info/2` or `handle_cast/2` clause, requires the last clause to have
an unbound first argument and no guard.

LiveViews are exempt by their `use FountainWeb, :live_view` — a LiveView
crashing is a reconnect, where a supervised singleton crashing drops what it
was holding. Anything else goes in `@exemptions` with a reason; the list is
empty today, and a stale entry (the file grew a catch-all, or no longer exists)
fails its own test so the list cannot quietly stop meaning anything.

The walk is scoped to `apps/fountain/lib`, as #2380 asks. `apps/fountain_*`
and `ee/lib` were checked by hand and are clean under the same rule —
`FountainBuzz.Harness` ends in `handle_info(_msg, state)`, and the one `ee/`
module with a bare clause is a LiveView. Whether to widen the walk to them is
a separate call.

## Verified by reverting

Every row was driven: the fix backed out, the named test run, the failure
observed, the fix restored.

| Backed out | What failed |
|---|---|
| `ExecutionDeadlineWorker` catch-all | `a stray message neither stops the worker nor loses its in-flight jobs` — `(exit) exited in: :sys.get_state/1 … FunctionClauseError`; guardrail flags `execution_deadline_worker.ex:128` |
| `RequestLog` both catch-alls | both `messages the writer does not match` tests — `(exit) GenServer.call(…, :flush)` raised in `handle_cast/2` and in `handle_info/2`; guardrail flags both lines |
| `Sink` catch-all | `an unknown cast neither stops it nor loses the queue` — the sink terminated, and the captured crash report carried the full payload; guardrail flags `sink.ex:98` |
| the LiveView exemption (string broken) | the guardrail flags 9 LiveViews, so the exemption is load-bearing rather than vacuous |
| a catch-all's unbound arg (`when is_tuple(message)`, `{_tag, _payload} = message`) | the guardrail flags both, so it discriminates guards and destructuring rather than matching on the name |
| — | an `@exemptions` entry silences a real hole, and the same entry against a module that has a catch-all fails `no exemption is stale` |

## Checks

`mix precommit` and the four touched test files, against a private database.


jhgaylor (approved):
Independent QA review: **APPROVED** for `2a992a949b43fd1f5ff3efdc8537b76198bf1b43`. No medium-or-higher correctness or test-coverage defect found.

The new fallbacks preserve the worker's jobs, buffered requests and queued analytics events without changing normal callback dispatch or arming extra timers. I traced the analytics and broker callers, deadline task completion/shutdown and journal recovery, and checked the guardrail against the existing callback definitions. The new behavioral tests check process survival and eventual delivery/completion.

Coverage includes all eight changed paths:

- `apps/fountain/lib/fountain/analytics/sink.ex`
- `apps/fountain/lib/fountain/broker/native/request_log.ex`
- `apps/fountain/lib/fountain/conversations/execution_deadline_worker.ex`
- `apps/fountain/test/fountain/analytics/sink_test.exs`
- `apps/fountain/test/fountain/broker/native_request_log_test.exs`
- `apps/fountain/test/fountain/conversations/execution_deadline_worker_test.exs`
- `apps/fountain/test/fountain/genserver_catch_all_guardrail_test.exs`
- `changelog.d/2380-genserver-catch-all.md`

Nonblocking robustness note: the three `shape/1` helpers inspect the first element of any tuple. Restricting displayed tags to atoms would also avoid logging payloads from malformed tuples with a map or binary first element. Current broker/analytics constructors use fixed atom tags, and deadline task replies use references; I found no current caller exposing sensitive data through this case.

Validation: independently read the full diff and affected callers and verified all eight snapshot files against the exact Git head. The coordinating reviewer ran the four changed test files in an isolated checkout: **36 tests, 0 failures**, seed `370241`, 14.0s; I inspected that result log. The coordinator also reports `git diff --check` clean. I did not run a duplicate Mix suite or the Review Loop service verification recipe.

jhgaylor (approved):
Security and tenant-isolation review of `2a992a949b43fd1f5ff3efdc8537b76198bf1b43`: **APPROVED**. No blocking security findings across all eight changed paths.

Traced broker telemetry through path redaction and the fixed `:record` message, analytics capture through the fixed `:enqueue` message, and deadline task replies through recorded job references. The new fallbacks preserve existing state without granting provider authority or changing tenant-scoped egress access. Existing execution claims still recheck the tenant, sandbox and provider binding before termination.

Reviewed the new regression assertions and existing cross-tenant/scope denial tests. This is a completed source review; test execution is coordinated separately by the lead reviewer.

jhgaylor (approved):
APPROVED — product/API compatibility and operability review of `2a992a949b43fd1f5ff3efdc8537b76198bf1b43`.

Completed an independent source review of all eight changed paths. No medium-or-higher actionable defects found.

The supported message protocols remain unchanged: analytics uses `{:enqueue, payload}`, the broker writer uses `{:record, row}`, and the deadline coordinator retains its existing tick, monitored-task result and process-DOWN handlers. Each new fallback preserves the same state without rearming timers, flushing buffered data, deleting jobs or granting a provider retry. Existing shutdown handling remains outside these ordinary unmatched-message clauses.

Followed the broker producer through `Native.handle_request/4`, write-time path redaction, buffering, `page/2`, tenant-scoped egress controller/rendering and the retention reaper. The wire fields, nullable values, cursor ordering and retention behavior do not change. Checked the analytics producer and delivery configuration, deadline task lifecycle and application supervision against the relevant base ADRs. No SDK, CLI, SSE, permission or API-contract update is required for this diff.

Reviewed the new tests for preservation of queued events, buffered rows and an in-flight termination, plus the guardrail's core-source scope, explicit LiveView exclusion and final-clause AST checks. Reviewed the changelog against the implemented callbacks.

Coverage:

- `apps/fountain/lib/fountain/analytics/sink.ex`
- `apps/fountain/lib/fountain/broker/native/request_log.ex`
- `apps/fountain/lib/fountain/conversations/execution_deadline_worker.ex`
- `apps/fountain/test/fountain/analytics/sink_test.exs`
- `apps/fountain/test/fountain/broker/native_request_log_test.exs`
- `apps/fountain/test/fountain/conversations/execution_deadline_worker_test.exs`
- `apps/fountain/test/fountain/genserver_catch_all_guardrail_test.exs`
- `changelog.d/2380-genserver-catch-all.md`

Validation: this reviewer inspected source and regression assertions without running concurrent Mix commands. The coordinating reviewer reports that the four affected test files passed against the isolated head/database: **36 tests, 0 failures**, 14.0 seconds, seed 370241. This is a completed source review, not Review Loop service verification.

jhgaylor (approved):
Adversarial maintainability and recovery review of `2a992a949b43fd1f5ff3efdc8537b76198bf1b43`: **APPROVED**.

Reviewed all eight changed paths and the relevant callback and caller code. The new fallback clauses preserve existing state and timer ownership; the regression tests exercise worker task survival, buffered request persistence and analytics queue delivery. The guardrail detects missing final clauses and rejects guards/destructuring for the current core definitions, with explicit exemptions and stale-exemption checks. No blocking correctness or maintainability defect found.

Validation: ran all four affected test files using the pinned toolchain, an isolated source/build copy and a disposable database: 36 tests, 0 failures (seed 370241). `git diff --check` passed. Applicable GitHub CI checks passed. This review covers this commit and does not merge the PR.

PR #2396, opened 2026-09-18
fix(conversations): register the secrets, not the whole sprite env (#2366)

### LINKED ISSUE #2366: Redaction registry holds non-secrets, so the cross-chunk carry delays ordinary output

BinaryBourbon:
Closes #2366. Follows #2364, which this makes cheap enough to keep.

## The problem

`SpriteEnv.build/4` registered the **whole assembled sprite env** with
`Fountain.Conversations.Redaction`, so the registry held the conversation and
sandbox ids, `FOUNTAIN_BASE_URL`, `SANDBOX_URL`, the trace context, the git
author and the broker's CA paths beside the credentials.

Free while the registry only scrubbed a row. Not free since #2364:
`RedactionCarry` holds a chunk back when its end could begin a registered
value, and a registered UUID makes every hex character such an end. A share of
every reply arrived a chunk late or at turn end, and a conversation that
printed its own id read it back as `[REDACTED]`.

## What this does

Decision 1 of the issue, at the source rather than by filtering the assembled
list. `secret_values/6` builds the registry from:

- the inference credentials the runtime **exported** (`exported_credentials/2`:
  Claude takes the OAuth token or the API key, never both, and the kind it
  passed over is not in the sandbox — `ConversationServerAcpTest` already
  asserted that of a refused subscription, and the API-key fallback in
  `ConversationServer` now registers through the same function);
- the callback token;
- the decrypted environment and vault values;
- the brokered credentials `Broker.split/2` withheld from the sandbox;
- the broker session token, via `Broker.proxy_secrets/1` — the token out of the
  proxy URL's userinfo, not the URL, so a chunk ending in `h` is not held back
  against `http://`.

The tenant's plain `env_vars` stay registered: they are config by intention,
but nothing stops a token being pasted into one, and a tenant's own variables
are few, unlike the identifiers that were on every conversation. Dropping them
needs the secret/non-secret distinction the issue's open question asks for.

Nothing secret leaves the registry. Every value the old code registered that is
a credential is still registered, by the source it came from.

## What this does not do

- **Decision 2, a minimum prefix length.** It buys a smaller hold at the price
  of leaking N-1 bytes of a secret at a chunk boundary, and with only secrets
  registered the hold is now rare. Not worth the weakened guarantee; say the
  word and it is a separate PR.
- **Decision 3, moving the boundary** (redact at message completion, or stop
  persisting chunks). That is a redesign of how a turn's text is stored, with
  the replay-dedup residual of #2364 attached to it. ADR-sized, not this PR.

## Tests

In `SpriteEnvTest`, all four fail on `main`'s registration except where noted:

- the secrets are registered and none of the identifiers are, and a
  conversation that prints its own id reads it back;
- the credential the runtime exported is registered and the one it passed over
  is not (this one guards the new code: it fails when the filter is removed,
  not on `main`);
- the broker session token is registered and the proxy URL is not, and an
  agent printing `HTTPS_PROXY` still hands over nothing;
- the issue's latency assertion: a chunk ending in **any** of the sixteen hex
  characters is written at once, while a chunk ending in a secret's first byte
  is still held.

Plus `Broker.proxy_secrets/1` unit coverage in `EgressTest`, and the #2364
property suite is unchanged and green.

`mix precommit` passed (9 stages); `apps/fountain/test/fountain/conversations`,
`sandbox_files_test.exs`, the broker tests and `scripts/test-docs.sh` are green
in a worktree with its own database.


jhgaylor (approved):
Adversarial security review — APPROVED

Reviewed all nine changed files at 019efce860d6d2ec39722b546556fef09dc5d3ba, independently tracing tenant-scoped credential resolution, environment/vault precedence, every shipped runtime's credential exports, the Codex ChatGPT grant, callback keys, broker placeholders and proxy-token extraction, registry retention, and streaming redaction consumers.

No actionable medium-or-higher security regression found in this diff. Exported credentials and broker-held originals remain registered before provisioning output; removing Fountain's public identifiers does not change credential ownership or access boundaries.

Validation: source and regression-assertion inspection using gh CLI and local git; no test execution by this reviewer. This is an independent reviewer verdict, not a claim that the full Review Loop verification policy ran.

This approval covers my security assessment; separate adversarial reviewers are reporting a confirmed P2 placeholder-classification issue that still requires correction.

jhgaylor (changes requested):
Adversarial correctness review — REQUEST CHANGES for `019efce860d6d2ec39722b546556fef09dc5d3ba`.

**[P2] Exclude generated broker placeholders from the secrets registry.** In `apps/fountain/lib/fountain/conversations/sprite_env.ex:171–175`, `Map.values(secrets)` receives values *after* `Broker.split/2`: a brokered `GITHUB_TOKEN` contributes `__github_token__`, while `broker_credentials` already supplies its actual secret. `exported_credentials/2` at lines 195–198 likewise retains generated inference placeholders, including `__codex_chatgpt_access_token__`.

[Issue #2366, Decision 1](https://github.com/managoat/fountain/issues/2366) explicitly requires leaving broker placeholders out. The current classification still delays ordinary output ending in `_` even when no actual secret starts with it, and printing these non-secret values becomes `[REDACTED]`. Exclude generated replacements using their broker source, retain real brokered credentials and tenant-supplied values, and add regressions through the actual `Broker.split/2` and `split_inference/2` → `SpriteEnv.build/4` → `RedactionCarry` paths. Avoid a blanket string-prefix filter that could discard a real tenant secret.

I reproduced both paths against the reviewed head using the pinned Elixir toolchain, recompiling changed source in memory without touching the database:
- GitHub token path registered `["ghp_actual_fixture_secret", "__github_token__"]`; feeding `ordinary_code_` emitted zero rows and held the whole chunk; `GITHUB_TOKEN=__github_token__` became `GITHUB_TOKEN=[REDACTED]`.
- Codex ChatGPT path registered `["__codex_chatgpt_access_token__", "eyJ_actual_fixture_grant"]`; feeding `ordinary_code_` again emitted zero rows.

Reviewed the full nine-file diff and related provisioning, runtime, broker, credential-binding, and streaming callers. Coverage: `broker.ex`, `conversation_server.ex`, `redaction.ex`, `redaction_carry.ex`, `sprite_env.ex`, `egress_test.exs`, `sprite_env_test.exs`, `changelog.d/2366-registry-holds-secrets-only.md`, and `docs/concepts/secrets.md`. `git diff --check` passed. The coordinating reviewer also completed ten focused test files at this exact head in an isolated worktree/database using Elixir 1.19.2 and OTP 28.3: 1 property, 232 tests, 0 failures (76.2 seconds). Those existing tests pass while the targeted diagnostics above expose the missing placeholder case. This is an independent reviewer verdict, not a claim of full Review Loop service verification.

apps/fountain/lib/fountain/conversations/sprite_env.ex
```diff
@@ -127,12 +128,76 @@ defmodule Fountain.Conversations.SpriteEnv do
...
+    credentials = exported_credentials(sprite_env, Keyword.fetch!(opts, :env_credentials))
+
+    # The proxy variables carry the broker session token inside a URL. The
+    # token is the secret; the URL around it is public, and registering it
+    # whole would hold back every chunk of output that ends in `h`.
+    proxy_token = Fountain.Broker.proxy_secrets(proxy)
+
+    credentials ++
+      [Keyword.fetch!(opts, :callback_token)] ++
+      Map.values(secrets) ++
```
jhgaylor: [P2] Exclude broker-generated placeholders from this registry

`secrets` is already the sandbox half of `Broker.split/2`, so it contains `__github_token__` for a brokered GitHub token. `Map.values(secrets)` registers that non-secret again; the `credentials` term also admits inference placeholders from `Broker.split_inference/2` through `exported_credentials/2`. Issue #2366 explicitly requires leaving broker placeholders out.

Reproduced with the actual head modules: splitting `%{"GITHUB_TOKEN" => "ghp_example_secret"}` and building the environment registers both `ghp_example_secret` and `__github_token__`. Feeding `model reply_` to `RedactionCarry` emits no row and holds the whole chunk even though no real secret starts with `_`; printing `GITHUB_TOKEN=__github_token__` still becomes `GITHUB_TOKEN=[REDACTED]`. This leaves the reported streaming latency and over-redaction on ordinary brokered runs.

Exclude synthetic placeholders using their broker provenance when collecting sandbox and inference values, while retaining the real values from `broker_credentials` and tenant-supplied values. Add brokered-environment and brokered-inference regressions asserting that a chunk ending in `_` is emitted immediately when no actual secret begins with it.

jhgaylor (changes requested):
Adversarial API/product review of 019efce860d6d2ec39722b546556fef09dc5d3ba.

Request changes for one P2 issue: broker placeholders still enter the purported secrets-only registry, retaining avoidable streaming latency and over-redaction on supported brokered conversations. See the inline finding.

I inspected all nine changed files, issue #2366, the trusted API review guidance, relevant ADRs, and the runtime credential, broker, refresh/fallback and stream-writing callers. There is no wire-schema change requiring SDK regeneration.

Validation: an isolated BEAM diagnostic compiled the actual head modules and reproduced the retained placeholder, delayed ordinary chunk, and redacted placeholder while confirming that the real credential remains scrubbed. This is a source review with a focused diagnostic, not a claim that I ran the full repository verification policy.

Shared validation reported by the coordinating reviewer: ten focused test files at this exact head passed (1 property, 232 tests, 0 failures) under the pinned Elixir 1.19.2/OTP 28.3 toolchain with an isolated PostgreSQL 15.14 database. The full repository verification policy was not run.

BinaryBourbon:
Fixed in 3828eaf3e — the finding was right, and it was the same mistake in two places.

`secrets` reaches `build/4` as the sandbox half of `Broker.split/2`, and
`split_inference/2` does the same to the credential a runtime exports, so both
`Map.values(secrets)` and `exported_credentials/2` were registering a generated
string that stands in for a secret the sandbox never held.

The exclusion is keyed, not pattern-matched: `Broker.placeholder?/2` compares a
pair's value against `placeholder/1` for that pair's own key, so a tenant secret
that happens to look like a placeholder is still a secret. The credential each
placeholder stands for keeps arriving through `broker_credentials` and is still
registered — an upstream that echoes it back is still scrubbed.

Two regressions, through the real paths rather than a hand-written placeholder:

- `Broker.split/2` on `%{"GITHUB_TOKEN" => ...}` → `build/4`: the credential is
  registered, `__github_token__` is not, `GITHUB_TOKEN=__github_token__` survives
  redaction unchanged, and `ordinary_code_` is emitted by `RedactionCarry` at once
  with nothing held.
- `Broker.split_inference/2` on the ChatGPT grant → `build/4` with
  `Managoat.Runtimes.Codex`: same three assertions for
  `__codex_chatgpt_access_token__`.

Both fail when the filter is removed (checked by reverting it: 2 failures, and
the other four #2366 tests stay green).

`mix precommit` passed (9 stages). `apps/fountain/test/fountain/conversations`,
`sandbox_files_test.exs` and the broker suites: 2 properties, 1748 tests, 0
failures, 2 skipped, in a worktree with its own database.

apps/fountain/lib/fountain/conversations/sprite_env.ex
```diff
@@ -127,12 +128,92 @@ defmodule Fountain.Conversations.SpriteEnv do
...
+  Every credential reaches a sprite through this list. One that ever reaches
+  it another way — as a brokered value does, which `build/4` registers
+  explicitly — has to be registered where that happens.
+  """
+  @spec exported_credentials([{String.t(), String.t()}], map()) :: [String.t()]
+  def exported_credentials(sprite_env, env_credentials) do
+    exported =
+      for {key, value} <- sprite_env,
+          is_binary(value),
+          not Broker.placeholder?(key, value),
```
jhgaylor: [P2] Resolve inference placeholders through credential identity and runtime aliases

For a supported OpenCode `google/*` agent, `Broker.split_inference/2` generates `AIza__gemini_api_key__` for `:gemini_api_key`, but `Managoat.Runtimes.OpenCode.default_env/2` exports it under `GOOGLE_GENERATIVE_AI_API_KEY` (`deps/managoat_runtimes/lib/managoat/runtimes/open_code.ex:69`). Checking `placeholder?/2` against that alias returns false, so the generated replacement still enters the registry.

At this head, a private BEAM probe through split → actual OpenCode runtime → build produced a registry containing both `AIzaSy_actual_gemini_fixture_secret` and `AIza__gemini_api_key__`. Printing the latter became `GOOGLE_GENERATIVE_AI_API_KEY=[REDACTED]`; feeding `public placeholder AIza__` emitted no rows and held the whole chunk, although that suffix cannot start the real credential. This leaves issue #2366's placeholder over-redaction and avoidable hold on this runtime/provider pair.

Classify replacements using the broker's credential identity/provenance, accounting for runtime aliases, while retaining actual secrets and broker-held originals. Add a regression through `Broker.split_inference/2`, the real OpenCode Google export, `SpriteEnv.build/4`, and `RedactionCarry`.

jhgaylor (changes requested):
Adversarial correctness review — REQUEST CHANGES for `3828eaf3e455965166e6cd7a8152a08f13afd9f3`.

The earlier GitHub and Codex placeholder cases are addressed, but two P2 classification defects remain:

- An actual unbrokered tenant secret equal to its key's generated placeholder is dropped. `Broker.split(%{"PASSWORD" => "__password__"})` returns the unchanged secret and an empty broker map; `SpriteEnv.build/4` then registers nothing, and `Redaction.redact/2` returns `PASSWORD=__password__` verbatim. Equality does not establish that the broker replaced a value. Preserve unbrokered secrets and use broker provenance when excluding replacements. The security reviewer is posting the canonical inline finding.
- OpenCode's Google credential uses the environment alias `GOOGLE_GENERATIVE_AI_API_KEY`, so the new keyed predicate fails to recognize the broker's `AIza__gemini_api_key__`. The placeholder is still over-redacted and can hold output that no actual secret could begin. See the inline finding.

I reproduced both through the actual head modules in a private BEAM VM under the pinned toolchain, compiling changed source in memory without starting the application, touching the database, or changing repository files. The Google probe used an `AIzaSy_...` credential: the registry retained its real value and the placeholder, printing the placeholder became `[REDACTED]`, and feeding `public placeholder AIza__` held the whole chunk. The ordinary one-byte `A` hold also comes from a real Google key and is not claimed as a new defect.

Reviewed the full base `ae736fe0eff2aed7ac07675041b09185002c5425` to head diff, approved-base contributor/reviewer guidance, issue #2366, prior reviews, provisioning and refresh/fallback callers, shipped runtime exports, registry retention, and streaming/log consumers. Coverage: `apps/fountain/lib/fountain/broker.ex`, `apps/fountain/lib/fountain/conversations/conversation_server.ex`, `apps/fountain/lib/fountain/conversations/redaction.ex`, `apps/fountain/lib/fountain/conversations/redaction_carry.ex`, `apps/fountain/lib/fountain/conversations/sprite_env.ex`, `apps/fountain/test/fountain/conversations/egress_test.exs`, `apps/fountain/test/fountain/conversations/sprite_env_test.exs`, `changelog.d/2366-registry-holds-secrets-only.md`, and `docs/concepts/secrets.md`.

Validation: my `git diff --check` passed. The coordinating reviewer reports ten focused test files at this exact head passed: 1 property, 234 tests, 0 failures in 26.2 seconds, with Elixir 1.19.2/OTP 28.3 and a private PostgreSQL 15.14 database. Those tests do not cover the two reproduced cases. This is an independent source review with focused diagnostics, not a claim of full Review Loop service verification.

apps/fountain/lib/fountain/conversations/sprite_env.ex
```diff
@@ -127,12 +128,92 @@ defmodule Fountain.Conversations.SpriteEnv do
...
+  # conversation. Environments would need a secret/non-secret distinction of
+  # their own before this could drop them (#2366).
+  defp secret_values(sprite_env, secrets, plain, proxy, broker_credentials, opts) do
+    credentials = exported_credentials(sprite_env, Keyword.fetch!(opts, :env_credentials))
+
+    # A brokered key's value here is the placeholder `Broker.split/2` left in
+    # its place, and the credential it stands for arrives in
+    # `broker_credentials` below. The placeholder is generated from the key
+    # and is not a secret (`Broker.placeholder?/2`).
+    tenant_secrets = for {key, value} <- secrets, not Broker.placeholder?(key, value), do: value
```
jhgaylor: **[P2] Preserve real tenant secrets that equal their key’s placeholder.**

`Broker.placeholder?/2` proves only byte equality, but this filter applies to every secret, including keys the broker never replaced. A valid encrypted environment or vault secret `PASSWORD=__password__` has no catalog entry or binding, so `Broker.split/2` returns it unchanged with empty `broker_credentials`; broker-disabled `Egress.split_brokered/2` does the same. `build/4` exports the real value but registers nothing, so an `env`, setup script, or agent printing it passes plaintext through `Redaction.redact/2` into `Conversations.log!/1` and the stream.

An isolated diagnostic against this head confirmed the environment secret changeset is valid, `registered: []`, and `output: "PASSWORD=__password__"`. Exclude a placeholder only when broker provenance confirms that this key was replaced, retaining arbitrary unbrokered tenant values. Add regressions for this same-key collision both with brokerage disabled and with an unbound key on a brokered deployment.

jhgaylor (changes requested):
Adversarial security review — REQUEST CHANGES for 3828eaf3e455965166e6cd7a8152a08f13afd9f3.

One P2 security regression: an encrypted tenant secret matching its own generated placeholder is omitted from redaction even when it was never brokered. Printing the value then exposes it in persisted logs and streamed output; see the inline finding.

I independently inspected all nine changed files, the approved-base security and contributor guidance, issue #2366 and prior reviews, tenant-scoped credential resolution and encryption, every shipped runtime’s exports, brokered and unbrokered secrets, custom bindings and connection tokens, proxy-token extraction, refresh and fallback paths, registry retention, and the streaming log writer. I also independently reproduced the OpenCode Google alias defect reported by the correctness reviewer, including a retained placeholder and a held `AIza__` suffix while the actual fixture credential had a different suffix.

Validation: isolated BEAM diagnostics recompiled the exact-head modules in memory, without starting the application or touching the database; the secret changeset accepted the reported value, and the redactor returned it unchanged. `git diff --check` passed. Shared validation reported by the coordinator: ten focused test files at this exact head passed (1 property, 234 tests, 0 failures) under Elixir 1.19.2 / OTP 28.3 with an isolated PostgreSQL 15.14 database. Those tests do not cover the failing case. This is an independent reviewer verdict, not a claim of full Review Loop service verification.

jhgaylor (changes requested):
Adversarial API/product review of `3828eaf3e455965166e6cd7a8152a08f13afd9f3`.

Request changes for two independently reproduced P2 defects in the new placeholder classification:

- [`sprite_env.ex:170`](https://github.com/managoat/fountain/pull/2396#discussion_r4042978285) drops a permitted unbrokered tenant secret such as `API_PASSWORD=__api_password__`. `Broker.split/2` leaves it unchanged with an empty broker map; `SpriteEnv.build/4` exports it but registers no value, so output retains the plaintext. The real secret changeset accepts this pair. Exclude a replacement only when broker provenance establishes that it is synthetic.
- [`sprite_env.ex:210`](https://github.com/managoat/fountain/pull/2396#discussion_r4042974160) misses OpenCode's Google alias. The broker creates `AIza__gemini_api_key__`, while the runtime exports it as `GOOGLE_GENERATIVE_AI_API_KEY`; comparing against the exported key's placeholder leaves the synthetic value registered. Its printed value becomes `[REDACTED]`, and `public placeholder AIza__` is held despite not matching the prefix of the registered real-shaped fixture `AIzaSy_actual_gemini_fixture_secret`. Resolve canonical credential aliases/provenance before excluding placeholders.

I inspected all nine changed files, issue #2366 and prior reviews, trusted API review guidance and relevant ADRs, and the provisioning, runtime, fallback/refresh and stream-writing callers. No wire-schema change requires SDK regeneration. The GitHub synthetic-placeholder path is now fixed; these remaining cases need regression coverage.

Validation: independent isolated BEAM diagnostic compiling the exact-head changed modules in memory, with no database writes; `git diff --check` passed. The coordinating reviewer reports ten focused test files at this head passed: 1 property, 234 tests, 0 failures (26.2 seconds, Elixir 1.19.2/OTP 28.3, isolated PostgreSQL 15.14). This is an independent source review with focused diagnostics, not full Review Loop service verification.

BinaryBourbon:
Both fixed in f8ee19e5e. Both were mine from the previous round, and they share a cause: I took "is a placeholder" from the value's shape instead of from the broker.

**The unbrokered secret.** Right, and it is the worse of the two — equality with `placeholder/1` says nothing about whether the broker replaced anything, and `Broker.split/2` leaves an unbrokered key alone, so `PASSWORD=__password__` was a real secret dropped out of the registry and into `log_events` in plaintext. The exclusion now requires the key to be present in the brokered map the split produced (`broker_credentials`, which is `state.brokered`); `Broker.placeholder?/2` keeps answering only "is this that key's placeholder", with its doc saying so explicitly.

**The opencode Google alias.** Also right. The broker keys its map by its own name (`GEMINI_API_KEY`), the runtime exports the credential under another (`GOOGLE_GENERATIVE_AI_API_KEY`), and I matched on the exported name. `exported_credentials/3` now takes the brokered map and matches each credential against the placeholder for the **broker's** key, resolved through `Broker.inference_keys/0` — by credential atom, so every alias in `InferenceCredentials.env_aliases/0` is covered by construction rather than by a list I would have to keep. `ConversationServer`'s API-key fallback passes `state.brokered` through the same function.

Regressions, both through the real splits into `build/4`:

- `Broker.split(%{"PASSWORD" => "__password__"})` → empty broker map, value unchanged; the value is registered and `PASSWORD=__password__` redacts to `PASSWORD=[REDACTED]`.
- `Broker.split_inference(%{gemini_api_key: "AIzaSy_..."})` → `build/4` with `Managoat.Runtimes.OpenCode` and `google/gemini-2.5-pro`: the env exports `GOOGLE_GENERATIVE_AI_API_KEY` carrying `AIza__gemini_api_key__`, the real key is registered, the placeholder is not and survives redaction, `public placeholder AIza__` is emitted by `RedactionCarry` at once with nothing held, and a chunk ending in `A` — where the real key begins — is still held.

Both fail against the previous head's logic (checked by restoring it: 2 failures, the other five #2366 tests green).

`mix precommit` passed (9 stages). Conversations, `sandbox_files_test.exs`, the broker suites and `inference_credentials`: 2 properties, 1793 tests, 0 failures, 2 skipped, in a worktree with its own database.

jhgaylor (approved):
Adversarial API/product review of `f8ee19e5e3f810702f7f16289c7b709f630820c6` — APPROVE.

Reviewed all nine changed files against trusted base `ae736fe0eff2aed7ac07675041b09185002c5425`, issue #2366, previous findings, trusted API/contributor guidance and relevant ADRs. Traced runtime exports, broker split and connection credentials, launch/reattach and refresh/fallback, registry retention, and output/log consumers.

Broker-map provenance now preserves real unbrokered secrets resembling placeholders; credential identity correctly recognizes OpenCode’s Google alias. Actual secrets remain scrubbed while public identifiers and generated placeholders remain readable and do not create avoidable carry holds. No actionable medium-or-higher API/product defect found; no wire-schema change requires SDK regeneration.

Validation: an independent private BEAM diagnostic compiled the exact-head Broker, SpriteEnv, Redaction and RedactionCarry modules in memory without application startup or database access. It passed same-key tenant-secret cases with brokerage off and on, actual Claude/Gemini/OpenCode/Codex credential exports, real-secret scrubbing, unchanged generated placeholders with immediate output, proxy-token scrubbing and readable public identifiers. `git diff --check` passed. The coordinating reviewer reports ten focused test files passed at this head: 1 property, 236 tests, 0 failures (23.4 seconds, Elixir 1.19.2/OTP 28.3, private PostgreSQL 15.14). This is an independent source review with focused diagnostics, not full Review Loop service verification.

jhgaylor (approved):
Adversarial correctness review — APPROVE for `f8ee19e5e3f810702f7f16289c7b709f630820c6`.

I reviewed the entire diff from base `ae736fe0eff2aed7ac07675041b09185002c5425`, the approved-base correctness/contributor guidance, issue #2366 and earlier reviews. No actionable medium-or-higher correctness defect remains in the reviewed change. The prior unbrokered `PASSWORD=__password__` disclosure and OpenCode Google alias over-redaction are addressed using broker custody and canonical credential identity; their regression assertions exercise the actual split/build/redaction paths.

I traced fresh provisioning and reattach, credential selection and runtime exports, broker and connection-token custody, callback/proxy registration, refresh and subscription-refusal handling, registry retention, and the output carry/log writer. The narrowed registry retains actual credential values and excludes generated broker placeholders and Fountain identifiers on the supported paths.

Validation: my private BEAM diagnostic compiled the exact-head changed modules in memory without starting the application or touching the database. It passed 16 brokered/unbrokered runtime-export cases across Claude, Codex, Gemini and OpenCode, plus eight actual placeholder-shaped unbrokered inference credentials. Real values remained scrubbed; synthetic placeholders were excluded; ordinary underscore-ending output emitted immediately. My `git diff --check` passed. The coordinating reviewer reports ten focused test files at this exact head passed: 1 property, 236 tests, 0 failures in 23.4 seconds, under Elixir 1.19.2/OTP 28.3 and an isolated PostgreSQL 15.14 database. This is an independent source review with focused diagnostics, not a claim of full Review Loop service verification.

Coverage: `apps/fountain/lib/fountain/broker.ex`, `apps/fountain/lib/fountain/conversations/conversation_server.ex`, `apps/fountain/lib/fountain/conversations/redaction.ex`, `apps/fountain/lib/fountain/conversations/redaction_carry.ex`, `apps/fountain/lib/fountain/conversations/sprite_env.ex`, `apps/fountain/test/fountain/conversations/egress_test.exs`, `apps/fountain/test/fountain/conversations/sprite_env_test.exs`, `changelog.d/2366-registry-holds-secrets-only.md`, and `docs/concepts/secrets.md`.

jhgaylor (approved):
Adversarial security review — APPROVED for `f8ee19e5e3f810702f7f16289c7b709f630820c6`.

Independently reviewed the complete nine-file base-to-head diff, trusted-base security and contributor guidance, issue #2366 and prior findings, tenant-scoped source resolution and encryption, every shipped runtime’s exports, custom bindings and connection tokens, proxy-token extraction, refresh and fallback paths, registry retention, and streaming, file and log consumers.

The previously reported unbrokered placeholder-shaped secret leak and OpenCode Google alias defect are fixed. Broker originals remain registered, and exclusion now requires broker provenance. No new actionable security regression found.

Validation: a private BEAM VM compiled exact-head modules in memory and passed 16 runtime/broker combinations, secret-source cases, conversation-registry isolation and all proxy-token split positions, with no application startup or database access. `git diff --check` passed. Shared validation reported by the coordinator: ten focused files passed, 1 property, 236 tests, 0 failures (23.4 seconds; Elixir 1.19.2/OTP 28.3; isolated PostgreSQL 15.14). This is an independent reviewer verdict, not full Review Loop service verification.

Issue #2402, opened 2026-09-18
Investigate Claude runtime exit 139 during concurrent turns on a shared Sprites sandbox


BinaryBourbon:
## Observed failure

During production verification of #1671 / #2400 on **2026-09-18 at 01:59–02:00 UTC** (September 17 in New York), one of six concurrent Claude turns on one fresh shared Sprites sandbox failed with **exit code 139**. Five turns completed. The failed conversation had already completed broker setup and reattachment successfully.

This is one observed wave, not an established failure rate or a confirmed concurrency defect. Exit 139 is consistent with a signal-related process crash, but the captured evidence does not identify the crashing executable or establish SIGSEGV as the cause.

## Environment and trigger

- Fountain deployment: `1103fe84d3ebf2c8a0ee30a20083088a3d9472e7` (#2400); both production replicas were healthy.
- Provider: Sprites; mode: ephemeral; runtime: `claude`; requested model: `anthropic/claude-haiku-4-5`.
- Fresh empty environment: no packages, repositories, setup script, or environment-variable overrides; normal `sandbox_api_access=owner`.
- Account's existing inference credentials were resolved by Fountain. The broker setup identified `CLAUDE_CODE_OAUTH_TOKEN`; this was not a test with a newly supplied API key.
- Create one bootstrap conversation without a prompt, wait for `provision/done`, then submit six `POST /api/conversations` requests together, all naming the same agent, environment and `sandbox_id`, with distinct channels.
- Each request supplied: `Reply with exactly BROKER_SETUP_OK. Do not use tools, access files, or perform any other task.`
- All six requests returned 201. No prompt retry was submitted.

## Durable evidence

Failed conversation: `ae506206-5eaf-4142-84e9-25dc84ef64a9`  
Failed turn: `dabdcfd0-d481-4d7f-bb42-fa9fb5d560d8`  
Shared sandbox: `fe7c98fa-1d04-4d84-86cc-a59069156a53`

| Event, September 18 UTC | Time |
| --- | --- |
| Broker started | 01:59:18.869038 |
| Broker completed | 01:59:21.520146 |
| Reattachment completed (`outcome=no_running_turn`, `reaped=0`) | 01:59:56.314874 |
| Turn started | 01:59:56.424054 |
| Turn terminal event, exit 139 | 02:00:00.287453 |

The turns API returned:

```json
[...]

PR #2409, opened 2026-09-18
chore(acp): preserve Node startup crash reproduction and upstream report

jhgaylor:
A shared Sprites sandbox produced exit 139 during Claude turns. Follow-up probes established native Node/npm crashes during image-launcher startup, including helper crashes hidden behind an outer exit of zero. This adds the bounded, credential-free diagnostic and a sanitized upstream report so the failure can be reproduced and candidate runtime/image fixes compared.

The diagnostic isolates home/cwd/temp and environment, preserves native signals and streams, detects helper cores, and cleans up timed-out process groups. The report distinguishes the symbolized core from later generated-code crashes and records missing platform provenance. Private memory dumps stay outside the repository.

Refs #2402. This does not identify the executable in the original incident, establish a fixed runtime version, or change production runtime settings. The issue remains open. No upstream report has been submitted.

Validation: four diagnostic unit tests passed; Linux CLI smoke passed in an offline local Node 24.20.0 container. That smoke uses a different architecture and no Sprites launcher, so it is not remedy validation. Conflict-marker, changelog, and whitespace checks passed. Original affected-platform evidence and its limitations are documented in the README and report.


BinaryBourbon (approved):
**APPROVE — independent adversarial correctness review**

Reviewed all changes in `068b785914f7aabe12f763d0b86e361891e6a463` against `361c5c6bd9f3abfed07809f626b6b2b365bbb670`. No actionable medium-or-higher correctness defect found. Source review is complete.

Coverage:

- `scripts/runtime-startup/probe.py`: startup scheduling, isolated environment/directories, native exit and stream preservation, timeout cleanup, helper-core detection, evidence output and early stopping.
- `scripts/runtime-startup/test_probe.py`: all four regression tests and their assertions.
- `scripts/runtime-startup/README.md`: usage, resource constraints, diagnostic interpretation and issue #2402 evidence.
- `scripts/runtime-startup/upstream-report.md`: reproduction, distinction between separate cores/reporting paths, unresolved attribution and candidate validation.
- `changelog.d/2402-runtime-startup-probe.md`: scope and format.

Validation: all four unit tests passed on macOS and again in the cached Linux/arm64 Node container with networking disabled. Additional Linux CLI checks passed: ordinary launches produced 10 successful records; an adapter fixture received `--version` and produced 10 successful records; a successful wrapper leaving a synthetic helper-core file returned exit 1 and stopped after its first shim control. Changelog validation and the exact-head whitespace/conflict-marker checks passed. Verified all five reviewed snapshot files match the pinned Git objects, and checked the factual investigation summary against issue #2402.

Limits: these checks validate the diagnostic, not the affected Sprites image, real ACP adapter, native crash attribution or a runtime remedy. `rl-env` is unavailable; no server suite was needed for these standalone Python/documentation changes. The standard conflict-marker script requires a Git checkout, so its exact expression was run against the pinned Git tree instead.

Optional low-priority hardening: `probe.py:48` uses an unbounded second `communicate()` after killing the process group. A custom launcher that detaches a pipe-inheriting helper can exceed the advertised timeout. No evidence shows that the documented Bash/NVM launcher or ACP version path does this, so this is not a blocking finding for the present reproducer. If that scope expands, bound the final drain, close inherited pipes on expiry and cover a detached helper in a regression test.

BinaryBourbon (approved):
Adversarial security and subprocess review of head `068b785914f7aabe12f763d0b86e361891e6a463`: **APPROVE**. No actionable medium-or-higher defect found for the documented disposable-sandbox diagnostic.

Coverage: inspected all five changed paths (`scripts/runtime-startup/probe.py`, `test_probe.py`, `README.md`, `upstream-report.md`, and `changelog.d/2402-runtime-startup-probe.md`) against the trusted base ownership, secrets, audit, and review policy. No server/API/tenant authorization paths change. The explicit child environment, fresh HOME/cwd, stdin closure, argv-based execution, core-capture opt-in and private-artifact guidance are appropriate for the documented use. I found no actionable credential-exposure defect.

Optional hardening at [`probe.py:48`](https://github.com/managoat/fountain/blob/068b785914f7aabe12f763d0b86e361891e6a463/scripts/runtime-startup/probe.py#L48): the final `communicate()` has no timeout. A helper that creates a separate session and retains stdout/stderr survives the launcher's process-group kill and extends the wait. A synthetic Python launcher with a detached 1.2-second sleeper reproduced a 0.1-second timeout returning after 1.276 seconds. A finite post-kill drain deadline and a detached-helper regression test would strengthen the general boundedness claim. However, this trigger requires behavior not established for either the affected Bash/NVM launcher or the optional ACP `--version` path. The current process-group cleanup covers the supplied intended-use evidence. I therefore classify this as low-priority hardening, not a merge-blocking defect or a demonstrated credential/security boundary failure.

Validation: all four checked-in Python tests passed. Additional synthetic credential/options sentinels were excluded, HOME matched cwd, and stdin returned EOF. The detached-helper reproduction above completed as described. These diagnostics ran on macOS, whose Python process added `__CF_USER_TEXT_ENCODING` beyond the explicit four-key environment; they do not establish Linux core behavior. `rl-env` is unavailable, Docker socket access was denied, and the conflict-marker gate could not run against the archive snapshot because it has no `.git`. No Linux/Sprites reproduction or remediation validation is claimed.

BinaryBourbon (approved):
Adversarial product/API review — **APPROVE**

Reviewed the complete PR at `068b785914f7aabe12f763d0b86e361891e6a463` against `361c5c6bd9f3abfed07809f626b6b2b365bbb670`. No actionable medium-or-higher defect found. Source review complete.

Coverage includes all five changed paths: the changelog fragment, diagnostic README, `probe.py`, `test_probe.py`, and sanitized upstream report. Traced CLI defaults, argument bounds, early stopping, native return codes, core detection, environment isolation, timeout handling and artifact output. No server, SDK, main CLI, Buzz CLI or plugin contract changes require propagation.

The report's affected-guest counts, sanitized result, core identity/checksum and distinction between the two native stacks agree with the retained investigation artifacts. The preparation-failure versus running-command exit-code distinction agrees with the runtime installation and pre-start failure paths. The linked Node issues support the cautious upstream comparison; the report does not claim an established root cause or verified remedy.

Validation: all four focused Python unit tests passed. An independent offline Linux smoke test using the cached image digest documented in the report passed all ten actual Node launches (three serial and two concurrent for each path, both using `/usr/local/bin/node`), with networking disabled and a read-only container filesystem. This validates the diagnostic CLI, not reproduction or remediation on the affected Sprites guest. `rl-env` was unavailable; these checks used the local Python and cached Linux container directly.

BinaryBourbon (approved):

Verdict: **APPROVED** for `068b785914f7aabe12f763d0b86e361891e6a463` (base `361c5c6bd9f3abfed07809f626b6b2b365bbb670`).

Reviewed all five changed files: `probe.py`, `test_probe.py`, the runtime-startup README and upstream report, and `changelog.d/2402-runtime-startup-probe.md`. The diagnostic keeps the control/concurrent schedules explicit, records native exits separately from discovered helper cores, and stops after a failing batch. The report separates observed crashes from unresolved original-failure attribution and candidate remedies. I found no blocking defect within the documented Node/Bash/NVM startup use case.

Validation: all four focused Python tests passed locally on macOS; the pinned diff passed `git diff --check`. This is an independent source review, not a claim that the Review Loop service's separate verification recipe ran here or that the affected Sprites crash was reproduced.

Nonblocking hardening: `scripts/runtime-startup/probe.py:48` drains pipes without a second deadline after killing the process group. A synthetic wrapper that starts a helper in a new session while retaining stdout can outlive that kill: with a 0.1-second timeout, a helper sleeping 1.5 seconds held the call open for 1.589 seconds. A bounded final drain would make this robust to daemonizing wrappers. I found no evidence that the documented Bash/NVM launcher or ACP version payload detaches this way, so this is optional hardening, not a requested change.

PR #2414, opened 2026-09-18
feat(conversations): serve a turn's prompt on the events feed

alecraso:
## The gap

`GET /api/conversations/{id}/events?blocks=true` is the read-model every SDK's
history is built on, and it structurally cannot carry the human's half of the
conversation. `Fountain.Conversations.Blocks` has a closed kind enum with no
member for the person, and `Blocks.for_event/1` only parses the `acp` stream —
what the *runtime* wrote. The prompt that caused a turn is on the turn row and
nowhere in the feed.

So a client replaying a conversation transcript from `/events` alone renders it
as a monologue in the agent's voice. The workaround is a second call to
`/turns` and a client-side weave, and every client has to write it.

## The change

An opt-in `?prompts=true`, off by default.

Every turn already emits exactly one `kind: "stage", stage: "turn",
state: "started"` log event. It carries that turn's `turn_id`, it is already
ordered immediately before the turn's output, and its `blocks` array is empty
today. With `blocks=true&prompts=true` that array holds one `prompt` block
carrying the turn's text.

**Anchoring on the existing event rather than interleaving a synthetic one is
the substance of the proposal.** `meta.next_cursor` is `List.last()`'s id, so a
fabricated event would hand a client a cursor that matches no row and corrupt
its pagination resume. Here nothing is added, removed or reordered, and
`next_cursor`, `has_more` and the page size are untouched. There is a test that
asserts the whole `meta` map and the page's id list are byte-identical with and
without the parameter, and that resuming from the cursor lands on the same page.

`prompt` is the name this codebase already uses for the human's text:
`SearchHit.kind` is `title | prompt | reply`.

### Files

| File | Change |
|---|---|
| `apps/fountain/lib/fountain/conversations/blocks.ex` | `prompt` added to `@kinds`, its row added to the block-shape table, and a moduledoc note that it is the one kind `for_event/1` never returns |
| `apps/fountain/lib/fountain_web/schemas.ex` | `Block`'s description gains a `prompt` sentence. The `kind` enum needed no edit — it already reads `Blocks.kinds()` |
| `apps/fountain/lib/fountain_web/controllers/conversation_controller.ex` | `operation(:events, ...)` gains the `prompts` boolean query parameter; the action parses it with the existing `parse_bool_param/2` and builds a `turn_id => prompt` map |
| `apps/fountain/lib/fountain_web/controllers/conversation_json.ex` | `put_blocks/4`, with `put_blocks/3` delegating so the SSE route's caller is unchanged; the new clause fills a `turn`/`started` stage event |
| `apps/fountain/test/fountain/conversations/blocks_test.exs` | `prompt` is on the enum and in the schema, and no event parse produces one |
| `apps/fountain/test/fountain_web/controllers/conversation_events_controller_test.exs` | 8 tests: off by default, on when asked, per-turn placement, other stage events stay empty, autonomous omitted, no-op without `blocks=true`, cursor/page accounting unmoved, and the `streams=acp` interaction |
| `sdk/contract/contract.json`, `sdk/typescript/src/generated/openapi.ts` | regenerated, not hand-edited, with `mise exec -- python3 scripts/sdk-contract/generate.py`. `ConversationWire.generated.swift` regenerated to a zero diff |
| `sdk/contract/manifests/typescript.json` | see below |

The server change is about 40 lines. I have deliberately kept it to the events
feed and left the three decisions below to you rather than choosing them
silently.

### The TypeScript manifest

`sdk/contract/manifests/typescript.json` declares `Block.kind` as
`{"values": [...], "exhaustive": true}`, so widening the enum fails
`npm --prefix sdk/typescript run verify-contract` until `prompt` is listed
there. That is the gate working as designed, and it is handled by adding the
value, not by relaxing the claim. I confirmed it is load-bearing by removing
the value again and watching the verifier name it:

```
enum Block.kind
    this client claims to handle every value but the API added: prompt
```

The swift, python and elixir manifests declare `Block.kind` as a bare list — a
subset check — and need no change.

I did **not** add a `prompt` case to the handwritten `sdk/typescript/src/turn.ts`.
Its `applyBlock` is an `if` chain that returns no chunk for an unrecognised
kind, and the raw block still reaches the caller as `{ type: "block" }`, so an
unhandled `prompt` degrades rather than throws. Say the word if you would
rather it produced its own `RunEvent`.

## Checks run

Run against a clean clone of `main` (`361c5c6b`) on the pinned toolchain
(Erlang 28.3, Elixir 1.19.2 from `.tool-versions`):

- `mix precommit` — **PASSED**, all 9 stages (toolchain, conflict-markers,
  compile with `--warnings-as-errors`, deps, format, credo `--strict`,
  dialyzer, sobelow, prod release assemble)
- `mix precommit test` — **PASSED**: 6 doctests, 2 properties, 6743 tests,
  0 failures, 2 skipped, plus every sibling app (136 / 30 / 33 / 60 / 29, all
  green)
- `python3 scripts/conflict-markers.py`, `python3 scripts/changelog.py check`,
  `git diff --check` — clean
- `mise exec -- python3 scripts/sdk-contract/generate.py` — regenerated;
  the diff is the new query parameter, the new enum value and the updated
  `Block` description, and nothing else
- `npm --prefix sdk/typescript run verify-contract` — ok (73 operations,
  46 schemas, 5 enums)
- `python3 sdk/python/scripts/verify_contract.py` — ok (67 operations,
  40 schemas, 5 enums)
- `mix test test/contract_test.exs` in `sdk/elixir` — 1 test, 0 failures

**One check I could not run: `swift test --filter ContractTests`.** This
machine has only the Command Line Tools and no full Xcode, so the `Testing`
module is unavailable and the Swift test target does not build. That is an
environment gap here, not a result — please treat the Swift client as
unverified by me. What I can say concretely is that the Swift manifest
declares `Block.kind` as a bare list, which is a subset check that a new value
cannot break, and that regenerating
`sdk/swift/Sources/FountainKit/Models/ConversationWire.generated.swift`
produced no diff at all.

## Open questions

I would rather you decided these than have me pick and bury them.

**1. Should `prompts=true` extend to the SSE stream as well as the events
feed?** I have only done the events feed. The argument for stopping there is
that a live client already knows the prompt it just sent, so history is the
whole of the need. The argument for symmetry is symmetry. `write_event/3` on
the stream path would need the same map threaded through it.

**2. What should happen for `autonomous`-origin turns (#817)?** I chose to
omit them. Their prompt is the literal string the server writes at
`Connection.open_autonomous_turn/5` — `"(background task follow-up)"` — for a
cycle nobody asked for, so rendering it in the human's voice seemed like
putting words in his mouth. The alternatives are emitting it with an origin
marker on the block, or emitting it plainly. Easy to change; there is a test
pinning the current choice either way.

**3. `streams=acp` filters out stage events, so a transcript client narrowing
to `acp` loses the prompts again.** `Conversations.event_in_streams?/2` drops
every `kind: "stage"` row unless `"stage"` is named, and the filter runs in the
query. I documented the interaction in the parameter description and pinned it
with a test rather than special-casing it, because changing the stream filter's
meaning felt like your call and not mine. The alternative is to keep a
prompt-bearing stage event when `prompts=true`, which makes the parameter
easier to use and the filter less predictable.

## Notes

- Changelog fragment added under `changelog.d/`, named after this PR, per
  `changelog.d/README.md`.
- The commit is signed off (DCO). Contributing under Apache-2.0 per
  `CONTRIBUTING.md`; I have read the relicensing asymmetry it describes and
  have no objection to it.
- Happy to split this, rename the kind, or rework the anchor if you would
  rather it were shaped differently — the design is the part worth arguing
  about, and the code is small enough to redo.


BinaryBourbon (changes requested):
**Correctness/testing review — REQUEST_CHANGES**

Reviewed head `8133fcb420e7dff042e0279a4b677554bede426b`, covering the complete 10-file PR patch against its merge-base `361c5c6bd9f3abfed07809f626b6b2b365bbb670`, using the trusted policy and guides from the supplied current base `e2241a146379c54a67d7b537d6ea6a7202ff0c42`. Source review is complete.

**[P2] Bound the prompt lookup to the event page and avoid loading image bytes.** At [`conversation_controller.ex:488–492`](https://github.com/managoat/fountain/blob/8133fcb420e7dff042e0279a4b677554bede426b/apps/fountain/lib/fountain_web/controllers/conversation_controller.ex#L488), `turn_prompts/1` calls `_unsafe_list_turns/1`. That context query fetches every turn in the conversation and preloads every associated `TurnImage`, including its binary `data`, before this helper filters or projects anything. Thus `?blocks=true&prompts=true&limit=1` on a long, image-rich conversation reads the entire turn/image history for each small page. Individual prompt images may be up to 10 MiB; unrelated historical images can dominate request memory and database traffic and cause timeouts or memory exhaustion. The same work happens when `blocks=false`, after the cursor reaches the end, or with `streams=acp`, where no prompt can be rendered, because line 475 checks only `prompts?`.

Please invoke a dedicated lookup only when both flags are enabled, collect the turn IDs from this page's `turn/started` anchors, and select only the matching user turns' IDs/prompts within the authorized conversation, without image preloads. Skip the query when no anchor IDs are present. Add a regression with unrelated image-bearing turns and a small page, plus the no-op flag/empty/stageless cases, asserting that historical image payloads are not fetched. This finding follows the actual query, preload and schema data path; I did not measure a production memory failure.

The rendering behavior otherwise checks out: real fresh-launch and accepted warm-reuse stage events persist the relational `turn_id`; rejected reuse leaves the fresh launch to emit the start; session recovery does not create another start anchor. Tenant ownership is checked before the unsafe query. Prompt attribution, autonomous-turn exclusion, original event IDs and pagination, stream filtering, the default-off behavior and the unchanged SSE serializer were traced through their callers and assertions. Generated TypeScript declarations expose the new parameter/kind; raw history LogEvents can carry prompt blocks, while TurnFollower continues to treat stage events as lifecycle signals.

Validation: I inspected the actual [CI run 35309174477](https://github.com/managoat/fountain/actions/runs/35309174477) and downloaded its test timing artifacts. All six partitions report **6,815 tests, zero failures**; partition 6 explicitly lists `BlocksTest`, `ConversationEventsControllerTest` and `EventsStreamTest`. CI required, coverage, API contract and SDK jobs succeeded. The recorded tested tree is `9b0224be0a17114f0c0e05ad5554b1d271e02681`, distinct from the head checkout tree `6ed9b12e16e3f1be5853f3d25e95c73f7c3805e1`; these are CI evidence for the PR's tested integration tree, not a claim that I reran the suite locally. The isolated checkout has no prepared dependencies/build, and I did not mutate shared build caches. Existing tests verify response shapes but do not bound this new lookup's database or image-loading work.

The separate [SDK release gate](https://github.com/managoat/fountain/actions/runs/35309174455) remains failed: generated shipped TypeScript source changed while package version `6.1.0` is unchanged and `release:skip-sdk` is absent. That release-policy gate also needs resolving; it is separate from the concrete source defect above.

Coverage (all changed paths):

- `apps/fountain/lib/fountain/conversations/blocks.ex`
- `apps/fountain/lib/fountain_web/controllers/conversation_json.ex`
- `apps/fountain/test/fountain/conversations/blocks_test.exs`
- `apps/fountain/test/fountain_web/controllers/conversation_events_controller_test.exs`
- `changelog.d/2414-events-prompts.md`
- `sdk/contract/manifests/typescript.json`

BinaryBourbon (changes requested):
REQUEST_CHANGES — independent adversarial security and tenant-isolation review of `8133fcb420e7dff042e0279a4b677554bede426b`.

**[P2] Bound prompt hydration to the returned event page and avoid image preloads.** At [`conversation_controller.ex:488–492`](https://github.com/managoat/fountain/blob/8133fcb420e7dff042e0279a4b677554bede426b/apps/fountain/lib/fountain_web/controllers/conversation_controller.ex#L488), `turn_prompts/1` calls `_unsafe_list_turns/1`, which loads every turn in the conversation **and preloads every attached image, including its binary `data`**. This runs whenever `prompts=true`, even with `limit=1`, `blocks=false`, an ACP-only stream filter, or an empty event page. Consequently each bounded event-page request now reads the entire turn/image history, including rows that cannot contribute a prompt block. Valid prompt images can each be 10 MiB; ten such attachments imply 100 MiB of irrelevant image payload read for each request, before other data and allocation overhead. That is a calculation from the accepted image limit and query, not a measured memory-exhaustion result. Repeated pagination multiplies this unnecessary database traffic and memory pressure as conversations grow.

Only hydrate when both `blocks` and `prompts` are enabled. Collect the current page's `stage=turn`, `state=started` turn IDs, skip an empty set, and query those IDs within the already-authorized conversation, projecting only `id` and `prompt` while applying the origin/nonempty checks. Do not use the image-preloading full-history helper. Add a regression demonstrating that off-page image-bearing turns are not fetched and that requests without eligible anchors do not hydrate prompts.

The tenant and disclosure boundaries otherwise hold: `events` establishes ownership with `get_conversation(id, current_user.id)` before the unsafe reads; the prompt lookup remains conversation-scoped; a foreign turn ID cannot hydrate a foreign prompt. The existing authenticated `/turns` response already exposes the same raw prompt to this caller, so I found no new authorization or redaction bypass. Autonomous turns are excluded, prompt blocks require the stage turn-start anchor, SSE continues through the existing three-argument serializer without hydration, and the response remains JSON. No mutations or new audit obligations are introduced.

Coverage: inspected all 10 changed paths—the block vocabulary/serialization, conversation controller, conversation JSON, OpenAPI schemas, both test files, changelog, contract JSON, TypeScript capability manifest, and generated TypeScript API—plus the router/auth scope, scoped conversation lookup, turn/image schemas, image size validation, full-history query, and existing turn/SSE output paths. The added tests cover opt-in behavior, exclusions, and response pagination, but do not bound the hydration query or its image preload.

Validation: source and call-path review of the pinned head and complete PR diff. I did not execute ExUnit/PostgreSQL tests or a database memory benchmark locally; the repository's `rl-env` runner is unavailable in this environment. The finding follows directly from the newly reachable query and accepted input limits, without assuming an observed outage.

BinaryBourbon (changes requested):
Adversarial product/API review — **REQUEST_CHANGES**

Reviewed head `8133fcb420e7dff042e0279a4b677554bede426b` against base `e2241a146379c54a67d7b537d6ea6a7202ff0c42`. Source review complete: all 10 changed paths, trusted API policy, CONTRIBUTING API guidance, and relevant ACP/request/release ADRs.

**P2 — Keep prompt enrichment bounded to the requested event page.** At [`conversation_controller.ex:488–492`](https://github.com/managoat/fountain/blob/8133fcb420e7dff042e0279a4b677554bede426b/apps/fountain/lib/fountain_web/controllers/conversation_controller.ex#L488), `turn_prompts/1` calls `_unsafe_list_turns/1`, which selects every turn in the conversation and preloads every associated image, including its binary `data` (`conversations.ex:1484–1490`, `turn_image.ex:20–24`). Therefore `GET /api/conversations/:id/events?blocks=true&prompts=true&limit=1` materializes the entire conversation's turn/image history for one event. A client draining multiple pages repeats this work on every page. Even an empty page, `streams=acp`, or `prompts=true` without `blocks=true` pays that cost, although none can return a prompt in those cases. This makes a bounded transcript read consume database bandwidth and application memory proportional to all historical attachments; individual accepted prompt images can be 10 MiB. The new response-only assertions do not catch that work.

Collect the non-null turn IDs of the current page's `turn`/`started` events, and query only the required prompt fields for those IDs within the already authorized conversation, without image preloads. Skip that query unless both flags are enabled and eligible anchors exist. Add a regression asserting unrelated turns/images are not loaded and no enrichment query runs for the ignored/empty cases. `kind: defect; needsHuman: false`.

The remaining public behavior is coherent with the stated scope. The renderer enriches existing rows, preserving IDs, ordering, cursors and page accounting. Fresh launch and accepted connection reuse publish the matching turn-ID anchor; autonomous turns are excluded. Default responses and the SSE three-argument renderer retain their existing behavior. I do not require SSE support or a different `streams=acp` policy for this explicitly history-only option.

Coverage includes `blocks.ex`, the controller and JSON renderer, `schemas.ex`, both changed test files, the changelog, committed wire contract, TypeScript manifest and generated OpenAPI types. I also traced TypeScript/Python/Elixir/Swift history and raw API access, Swift's extensible block-kind decoding, main CLI/Buzz stream consumers, and existing shared pagination expectations. Convenience history methods do not expose `prompts`; callers can use public raw HTTP access. The TypeScript follower ignores blocks on stage events, so this change does not add a semantic prompt event to run following; the raw history response carries the new block.

Verification: TypeScript contract verifier passed (73 operations, 46 schemas, 5 enums); Python verifier passed (67/40/5). The focused TypeScript plan/follower test passed. A broader TypeScript run test attempt was interrupted after remaining pending, so it is not claimed as passing. Focused BEAM tests could not run because `rl-env` is unavailable in this worker; no database performance measurement is claimed. The reviewed source directly establishes the unbounded preload. The PR's CI server/SDK checks subsequently passed on its tested integration tree (not a local rerun of the head checkout), but the separate SDK release gate remains failed and must be resolved under the repository's independent-release policy.

BinaryBourbon (dismissed):

Verdict: **REQUEST_CHANGES** on `8133fcb420e7dff042e0279a4b677554bede426b`.

**[P2] Bound prompt hydration to the events being returned.** In [`conversation_controller.ex:475,488–492`](https://github.com/managoat/fountain/blob/8133fcb420e7dff042e0279a4b677554bede426b/apps/fountain/lib/fountain_web/controllers/conversation_controller.ex#L475), every `prompts=true` request calls `_unsafe_list_turns/1`. That helper retrieves the entire conversation's turns and preloads every `TurnImage`, including its binary `data` (`conversations.ex:1484–1490`, `turn_image.ex:23`). The event `limit` and `after` cursor do not bound this work. Replaying successive small pages therefore repeatedly loads the whole history and all attachments just to return a few text prompts. The same work runs for an empty page, for `streams=acp` (which cannot carry these blocks), and even without `blocks=true`, where the option is documented as ignored. Accepted images can be up to 10 MiB each, so unrelated attachments can dominate the database transfer and request memory.

Please collect the turn IDs of the current page's `turn`/`started` anchors, query only their IDs and nonempty user prompts within the already-authorized conversation, and omit image preloads. Skip hydration unless both options are enabled and there are matching anchors. Add a focused regression that checks the query boundary with unrelated image-bearing turns and verifies no prompt/image lookup for a no-op request. Existing response-equality and cursor tests cannot detect this cost.

The opt-in behavior, reuse of persisted event IDs, and separate prompt block kind otherwise keep the change localized and preserve default responses. Reviewed all 10 changed paths and their relevant readers against the approved-base policy.

Validation performed here: TypeScript and Python SDK contract verification passed; shared conformance lint passed (29 scenarios); changelog validation passed. I did not run the full Elixir or independent Review Loop verification recipe locally. The separate [SDK release gate](https://github.com/managoat/fountain/actions/runs/35309174455/job/105487450972) failed because shipped TypeScript declarations changed without a version bump or an explicit release-deferral label; that release choice remains outstanding.

alecraso:
Thanks — the finding is correct and I've taken it in full. Pushed as `fe34191e`, rebased on `main` at `57f60f99`.

## The fix

`turn_prompts/1` is gone. `turn_prompts/3` hydrates only when `blocks` and `prompts` are both on, and only from the turn ids carried by the current page's `turn`/`started` anchors — the only events `put_blocks/4` renders a prompt onto. An empty anchor set returns `%{}` without touching PostgreSQL.

The query is a new `Conversations._unsafe_list_turn_prompts/2` rather than a filter layered over `_unsafe_list_turns/1`. It selects `{t.id, t.prompt}` for those ids inside the already-authorized `conversation_id`, applies the origin and non-empty checks in SQL so excluded rows never leave the database, and preloads nothing. `_unsafe_list_turns/1` itself is untouched — `/turns` legitimately serves those images.

Semantics are preserved exactly, including NULL `origin` meaning a user turn: `(&1.origin || "user") == "user"` became `is_nil(t.origin) or t.origin == "user"`.

## Proving it is bounded, not merely unchanged

You were right that response-shape assertions cannot close this, so I did not add one. The two new tests count `turns` and `turn_images` reads off `[:fountain, :repo, :query]` — the idiom `webhooks_test.exs` already uses for "what does every tenant pay" — and assert on `num_rows`, so a full-history query fails even though it is still *one* query.

I then broke each guard to confirm the tests are load-bearing rather than incidental:

| mutation | result |
| --- | --- |
| restore the `_unsafe_list_turns/1` lookup | both new tests fail; the one-event page reads `[{"turns", 4}, {"turn_images", 4}]` against the expected `[{"turns", 1}]` |
| gate on `prompts?` alone, dropping `blocks?` | the no-op test fails with `[{"turns", 1}]` against `[]` |
| drift the anchor predicate away from `put_blocks/4` | 4 tests fail, 3 of them pre-existing |

The first mutation also confirms each no-op case you named by measurement rather than by reading: with the old lookup restored and the flag gate still in place, `streams=acp`, a drained cursor and a stageless page produced three `turns` + `turn_images` round-trips between them; adding the second mutation produces the fourth, `prompts=true` without `blocks=true`. All four did hydrate.

The third mutation is the one nobody asked for. The controller's anchor predicate and `put_blocks/4`'s render predicate have to agree, and if they drift the failure mode is a silently missing prompt rather than an error. It turns out your existing assertions already guard that, so I left the predicate local with a comment pointing at `put_blocks/4` rather than extracting a shared helper — but it is a real coupling and worth naming.

I have not measured a production memory failure either and am not claiming one. The 10 MiB figure does check out: `Fountain.Images.max_prompt_image_bytes/0` is `10 * 1024 * 1024`.

## Verification

- `mix precommit` — PASSED, 9 stages
- `mix precommit test` — PASSED, 6,847 tests, 0 failures, plus the umbrella app suites (136 / 60 / 33 / 30 / 29, all 0 failures)
- TypeScript contract verifier — ok, 73 operations / 46 schemas / 5 enums
- Python contract verifier — ok, 67 / 40 / 5
- `sdk/elixir`, `mix test test/contract_test.exs` — 1 test, 0 failures
- `scripts/changelog.py check` — 52 fragments, 104 entries
- `python3 scripts/sdk-contract/generate.py` — regenerated, zero diff. This change is internal; the contract does not move.

`swift test --filter ContractTests` I still cannot run: this machine has Command Line Tools only and no `Xcode.app`, so the `Testing` module is unavailable and the test target does not compile. I am not claiming Swift coverage. Postgres 15 in a throwaway container, as the local install is 14.

## The SDK release gate

This part I can't resolve myself, so I would rather ask than guess.

`sdk/typescript/src/generated/openapi.ts` moved in the feature commit because the server contract grew `?prompts=true` and the `prompt` block kind. The TypeScript package's own public surface did not change — no hand-written source, no dependency, no packaging change — and that file is the contract's shadow.

CONTRIBUTING.md is explicit about exactly this case:

> Do not bump an SDK's version because the contract moved. Merging a version bump publishes that SDK, so a version moves when its own public surface changes. Label a PR `release:skip-sdk` where the distinction needs saying out loud.

`contributing/sdk-maintenance.md` agrees — "changing the contract or another SDK does not demand a no-op release" — and adds that the label wants a sentence in the PR. This is that sentence, if you want it: **the shipped TypeScript declarations changed only because the server contract did, so `6.1.0` should not be republished for it.**

So by the policy the resolution is the label rather than a bump, and `sdk-release.mjs`'s own failure message offers the same two options. I have `pull` permission only, so I cannot apply it.

**Could a maintainer add `release:skip-sdk`?** The gate listens on `labeled`, so it should re-evaluate without another push.

I would rather not bump `6.1.0` to make the gate green, since sdk-maintenance.md says merging a bump publishes and warns against bumping a version you intend to hold. If you would actually prefer a release here, say so and I will do the full bump properly — `package.json`, lockfile, `USER_AGENT`, the SDK changelog and `scripts/sdk-catalog.py --write`.

jhgaylor (approved):
Follow-up review of `fe34191ef7d5f17df527f28914876ba214c7d243` against `57f60f99ebe6a9bc6f1ed96d0102f6fe0fa53945`: **APPROVE**. Reviewed the complete 11-file diff, including the prompt lookup repair. No new blocking findings.

The earlier P2 is addressed. Hydration now requires both flags and collects only the returned page's non-null `turn`/`started` turn IDs, after removing the pagination lookahead row. An empty set performs no query. `_unsafe_list_turn_prompts/2` stays scoped to the authorized conversation, selects only `{id, prompt}`, applies the user-origin/nonempty predicates in SQL, and never preloads images. The renderer and anchor-selection predicates agree; the NULL-origin compatibility behavior is preserved.

The new regression checks the database work: a one-event page among four image-bearing turns returns one turn row and reads no image rows. The companion test verifies no turn/image reads without `blocks=true`, with an ACP-only filter, after a drained cursor, or on a page without a turn-start anchor. Existing assertions preserve prompt attribution, autonomous exclusion, event IDs, pagination and default responses. SSE continues through the unchanged three-argument call, and prompt blocks remain opt-in history data rather than agent-answer text.

Validation: [CI run 35313290811](https://github.com/managoat/fountain/actions/runs/35313290811) passed. Its downloaded `tested-tree` artifact is `c8c969700ff9501cf213a6e090db228627247f03`, exactly matching this head's tree. The six partition logs report **6,847 tests, zero failures**, and explicitly include `ConversationEventsControllerTest`, `BlocksTest` and `EventsStreamTest`. Static analysis, coverage, API contract and SDK jobs—including Swift on Linux and macOS—passed. Locally, TypeScript and Python contract verification, conformance lint, changelog validation, conflict-marker checking and `git diff --check` passed. I did not rerun ExUnit locally.

The SDK release decision is now recorded by the author's explanation and the `release:skip-sdk` label. The newer [label-triggered gate run](https://github.com/managoat/fountain/actions/runs/35314998782) completed as **skipped**, as the approved workflow specifies. The older failed check still appears in the PR rollup.

PR #2416, opened 2026-09-18
fix(api): describe prompts refused during sandbox deletion

### LINKED ISSUE #2406: OpenAPI: 409 on POST /api/conversations/{id}/prompts says "being reset" but also means "being deleted"
The OpenAPI response description for the `409` on `POST /api/conversations/{id}/prompts` reads **"Sandbox is being reset"** (`conversation_controller.ex:814`). Since ADR 0058 stage 9a (#2386) the same `409 sandbox_reset_pending` also answers a prompt to a conversation whose computer is being **deleted**. Stage 9a made `transition: "destroying"` durable, and every door now refuses a destroying machine with this 409. So the description is only half right for SDK users.

In the same PR, the internal user-facing strings were already corrected: `Team.Schedules.describe_error(:sandbox_reset_pending)` now says "being reset or deleted", and `changelog.d/2344-durable-destroying.md` says the same. Only the contract description lags.

**Why this wasn't fixed in #2386:** changing the description regenerates `sdk/typescript/src/generated/openapi.ts` and alters the SDK surface, so it is a contract change with its own `release:skip-sdk` judgement. That belongs in its own PR rather than inside a stage PR about the destroy protocol.

**Fix:** change the description to "Sandbox is being reset or deleted" (or split out the two cases if the codes should differ), regenerate with `npm ci` then `scripts/sdk-contract/generate.py`, and check `scripts/sdk-contract/build.sh --check`. Grep for the same description on the other routes that return `sandbox_reset_pending` (the attach door, the team route) and fix them together.

Found by the surfaces review of #2386 (round 3), marked non-blocking there.

jhgaylor:
The prompts API described every `409 sandbox_reset_pending` response as a reset, although the same response also covers deletion. Correct the OpenAPI description to “Sandbox is being reset or deleted” and regenerate the TypeScript declaration comment. Related team/create/attach descriptions were checked and require no change.

Closes #2406.

Validation: contract generation and freshness checks passed, including `scripts/sdk-contract/build.sh --check`; TypeScript, Python, Elixir, and Swift contract verifiers passed. The contract projection and Swift wire models are unchanged because this only corrects documentation. All nine stages of server `mix precommit`, plus conflict-marker, changelog, and whitespace checks passed.

SDK release: documentation only; status codes, response shapes, SDK behavior, and package versions are unchanged. `release:skip-sdk`.


BinaryBourbon (approved):
Adversarial review of 3abdd0914cb848bd363c3e7409bc4f0ed0aef124: approved; no actionable findings across all three changed files.

Traced the prompt path through the tenant-scoped conversation lookup, Wake's reset/destroying refusals, and the fallback's existing 409 sandbox_reset_pending response. The revised description matches that behavior and the existing teardown/durable-destroying regression assertions. The generated TypeScript comment agrees with the controller; the contract projection intentionally drops descriptions, so the unchanged wire contract and Swift models are appropriate. The changelog accurately addresses #2406.

Validation: isolated exact-head conflict-marker, changelog, Python contract and whitespace checks passed. GitHub's API-contract, generated TypeScript/SDK and Swift checks passed. I inspected the relevant Elixir regression tests but did not rerun the server suite locally; remaining CI checks retain their own merge-gate authority.

PR #2417, opened 2026-09-18
test: prove execution lifetime gaps blocking #2394

jhgaylor:
File reads currently cannot establish that remote execution has ended before park/delete. The opt-in probes reproduce timeout and termination gaps using Fountain's locked adapters and real runner process backend, and record the agreed full-guarantee, fail-fast, preserve-provider-coverage requirements.

Live Sprites evidence is decisive: continuous output extended a 500 ms timeout to about 2.2 seconds; explicit termination returned `:ok` while a child ignoring `SIGTERM` continued writing. The disposable private sandbox was deleted and its absence verified. The handoff describes the execution identity, uncertainty, deadline, descendant cleanup and lifecycle-priority work this requires. E2B, Daytona and Firecracker live validation remain outstanding.

Validation:
- 8 offline Elixir characterization cases passed, meaning the existing gaps were reproduced.
- Both runner subcases reproduced on macOS and isolated Linux arm64.
- Live Sprites stdout, stderr, silent timeout, ordinary-child and TERM-resistant-child cases completed, with cleanup verified.
- Elixir formatting, Go formatting, Python syntax, conflict-marker/changelog/diff checks and redacted secret scan passed.

These diagnostics are outside normal test discovery. The live command explicitly requires provider credentials and creates a real resource. No production behavior changes; #2394 and its parent remain open.

Related: #2394.

Issue #2394, opened 2026-09-17
Sandbox files: coordinate read admission with machine park and destroy


BinaryBourbon:
## Outcome

Make the sandbox files API participate in the existing machine ownership protocol so a read cannot race a Fountain-controlled park or destroy. This is the executable lifecycle follow-up to #1715, estimated as one medium-sized PR once the admission contract is explicit.

## Current evidence

Static inspection of main at `ae736fe0eff2aed7ac07675041b09185002c5425` on 2026-09-17 found that `Fountain.SandboxFiles.ready?/1` checks only the supplied sandbox struct, then `run/3` calls `Managoat.Sandbox.exec/4` independently of `Fountain.Machines`. This applies to listing, file reading, git diff and git status. No new race reproduction was run during this assessment; reproduce it deterministically before implementing the fix.

The separate symlink-confinement defect already landed in #2139. Do not redo it or expand this task into descriptor-relative filesystem confinement.

## Sequencing

- Implementation base: current `main`. #2386 (ADR 0058 stage 9a) merged as `e2241a146` on 2026-09-18 and makes `destroying` durable intent that readers must refuse even after its lease expires. #2423 (stage 9b-i, `9816560d4`) deleted the `MACHINE_OWNER_ENABLED` gate, so every lifecycle verb now runs in the machine's owner; there is no inline path to coordinate with.
- #2425 (stage 9b-ii, dropping the fence columns) was still open on 2026-09-18 and edits `machines/machine.ex` and `machines/lease.ex`. If it lands first, rebase onto it before landing; it is not a hard dependency.
- Do not wait for all of #2344 or its later stage 9b cleanup. Reuse its machine-owner/lease architecture rather than introducing another lifecycle coordinator or baking the old fence columns into the public read contract.
- This issue establishes coordination with Fountain lifecycle operations. It does not establish the separate provider-level promise that exec never wakes an autonomously sleeping VM; #2395 owns that decision. Do not silently change ADR 0039's public promise or close #1715 on the strength of this fix alone.

## Implementation boundary

Route all four operations through one bounded read-admission protocol in `Fountain.Machines`. Decide eligibility from fresh state under the existing coordination mechanism and maintain exclusion across provider execution. A second row check or a GenServer mailbox alone is insufficient across nodes and the gate's inline mode. Keep provider I/O outside database transactions and advisory locks, as ADR 0058 requires.

[...]

PR #2419, opened 2026-09-18
fix(runner): bound exec child lifetime and output draining

jhgaylor:
Runner `exec` previously killed only the parent on timeout: children could continue running, and inherited output pipes could keep the call waiting indefinitely. Run each one-shot command in a process group, clean up that group on timeout or parent exit, and bound subsequent output draining to 250 ms. Long-lived background work must use a streaming session.

Keep the leader unreaped with `waitid(WNOWAIT)` until all group signaling finishes, and fence late cancellation callbacks before `cmd.Wait` releases its identity. This prevents cleanup from signaling a reused PID/PGID. Linux and Darwin have explicit implementations; Darwin handles nonterminal wait events and zombie-only groups. Existing timeout exit code 137 remains. An otherwise-successful command with an escaped child retaining its output pipe returns 124 with an output-drain diagnostic.

Validation:
- Full CLI `go test -mod=readonly -count=1 ./...` and `go vet -mod=readonly ./...` passed.
- Real-process lifetime tests cover timeout/normal parent exit, inherited/redirected output, stopped leaders and escaped children. Ordering tests verify unreaped identity and cancellation fencing; targeted race-detector run passed.
- Focused tests passed in isolated Linux arm64; Darwin amd64 cross-compilation passed.
- Formatting, conflict-marker/changelog/diff checks passed. Independent review's PID-reuse finding was repaired and re-reviewed with no remaining blockers.

This is a prerequisite for #2394, not the full read-admission fix. Process-group escape, daemon loss, durable execution identity, Firecracker live validation and shared lifecycle coordination remain outstanding. No exclusion guarantee is inferred from a local timeout or a sent signal. Related evidence: #2417.


BinaryBourbon (approved):
Approved after an independent adversarial review of all five changed files at da9a7d9c1bf8f04940554443b3cf981352dea591 against base 57f60f99ebe6a9bc6f1ed96d0102f6fe0fa53945. No actionable blocking defect found.

Traced Exec through daemon dispatch and Firecracker guest forwarding, and checked timeout/normal-exit cleanup, inherited output, stopped leaders, unreaped PID/PGID ownership, cancellation fencing, and exit-code compatibility.

Validation:
- Darwin/arm64: full CLI `go test -mod=readonly -count=1 ./...` and `go vet -mod=readonly ./...` passed.
- Darwin/arm64: `GORACE=atexit_sleep_ms=0 go test -mod=readonly -race -count=1 ./internal/runner -run TestExec` passed. With the default race-runtime exit delay, the two success fixtures exceed their 500ms deadline and return 137; disabling that instrumentation delay passes without source changes.
- Linux/arm64: focused runner Exec tests passed in an isolated container.
- Changed Go files are gofmt-clean; diff whitespace check and current-head GitHub CI/secrets checks passed.

This approval covers the bounded local exec improvement. Process-group escape, daemon-loss recovery, live Firecracker validation, and read/lifecycle exclusion remain the explicitly documented follow-up work.

PR #2420, opened 2026-09-18
fix(conversations): relaunch an adapter a native crash killed while starting (#2402)

jhgaylor:
Refs #2402. This does not close it: the upstream Node report and the `managoat_runtimes` install-probe hazard remain open there.

## The failure

One of six concurrent Claude turns on a shared Sprites sandbox failed with `turn/done` `exit_code=139` and an empty transcript. The [attribution comment](https://github.com/managoat/fountain/issues/2402#issuecomment-5725851731) places that exit on the adapter's own Node process:

- The wave set no execution limits, so the turn was unbounded. On that path, `Connection.spawn_command/5` runs the install check as its own `Sandbox.exec` first; a failure there fails the turn before start. The 139 is therefore the `:exit` of the spawned `claude-agent-acp`.
- On a live guest, Sprites reports a signal death as 128 + signal (`sh -c 'kill -SEGV $$'` → 139).
- The `/.sprite/bin/node` launcher's npm helpers run under `>/dev/null 2>&1 || true`, so they cannot produce the exit. The launcher ends in `exec` of native Node.
- Successful turns logged adapter stderr (`[session/create] …`) from +3.9 s. The failed turn ended at +3.86 s with nothing on either stream.

Earlier standalone probes on the issue caught native Node 24 SIGSEGVs during concurrent startup, including `claude-agent-acp --version`.

## The fix

`TurnLaunch.relaunch_crashed/3` launches an unbounded turn's adapter **once more** when all of these hold:

| Condition | Why |
| --- | --- |
| Exit is 132, 133, 134, 135, 136 or 139 (SIGILL, SIGTRAP, SIGABRT, SIGBUS, SIGFPE, SIGSEGV) | A native crash. SIGKILL/SIGTERM came from outside and would come again. |
| `turn_metrics.first_output?` is false | Stdout and the exit reach the actor from the same `Sprites.Command` process, in order. A silent adapter never answered `initialize`; the peer sends `session/new` only from `handle_response(:initialize, …)`. So no prompt existed, and the retry cannot run one twice. |
| `turn_metrics.launch` is present and not yet relaunched | Set only by a fresh launch. An idle peer carried across turns prompts at once, so a silent adapter proves nothing there. One retry per turn. |
| `turn_execution` is nil | A bounded turn's command belongs to its execution journal and deadline. |
| The conversation is still bound to this actor's sandbox | Checked after the dead command's peer has stopped (the stop waits, and Wake can rebind meanwhile), and before any announcement or spawn. The plan write carries the actor's `sandbox_id` (`ExecutionGuard._unsafe_write_parent/4`, the same `rebound?/2` as `_unsafe_recover_turn/3`). An unbounded turn has no journal to name its machine, so without it a stale actor whose conversation Wake moved to a replacement would write its session id onto the moved row and spawn on the old handle. |

The relaunch continues the same turn:
- It publishes no second `turn/started`.
- It keeps the turn's span and start time.
- It reuses the first launch's `{mode, id}` through the new `TurnMachine.session_plan/4`, which runs the fence check again, now with the binding. Recomputing the plan picked `session/resume` for a session that was never created.
- It records `session`/`done` with `event: "restarted"`, `reason: "adapter_crashed"`, `detail: "SIGSEGV"`. This is the shape #1667's restart already uses mid-turn.

`relaunch_crashed/3` takes the server's exit path (`finish_exited_turn/2`) as a callback, and every outcome other than a relaunch ends through it:
- an ineligible exit;
- a refused fence, including a binding moved during peer shutdown;
- a relaunch that cannot start.

A failed retry therefore ends exactly as the crash would have. That means guarded `Machine.end_turn/3` with the crash's exit code, `turn/done`, the span, and one failed `turn.completed` measured from the first launch, which keeps `FountainTurnFailureRate`'s counts whole. A stale actor emits nothing.

**Size pin.** The server's `:exit` clause is one line delegating to `TurnLaunch`, and its body moves to `finish_exited_turn/2`. `restart_acp_peer/1` moves to `Connection.restart_peer_session/1` to pay for the lines. `conversation_server.ex` is 2019 lines against a pin of 2025, and the pin is unchanged.

**Redaction.** The launch context sits on `turn_metrics` and carries the prompt, images, agent and conversation. `Redaction.server_state/1` now reduces it to `relaunched?`. `turn_metrics` moves from `@plaintext_fields` to a sentinel shape in the field guard.

**No cross-release concern.** No message shape changes. The new key lives in one actor's own state.

## Verified by breaking it

Each probe restored the file from a `cp` snapshot taken before the edit. With relaunch disabled entirely (the base behavior), the replacement-sandbox regression passes and only the three relaunch tests fail.

| Probe | Test that failed |
| --- | --- |
| Never relaunch | SIGSEGV relaunches; second crash fails; relaunch that cannot start |
| Drop `first_output?: false` | crash after the adapter answered is not relaunched |
| Drop `relaunched?: false` | a second crash fails the turn with its exit code |
| Any exit code qualifies | exit 1 / exit 137 fail as before |
| Relaunch a carried peer from the turn row | a crash on the peer carried over from the last turn is not relaunched |
| Binding dropped from the plan write | a conversation Wake moved to a replacement sandbox is not relaunched |
| Fence moved back before the peer stop | a crash retry whose conversation moves during peer shutdown is not relaunched (`command_exit_binding_test`) |
| Metrics dropped when the relaunch cannot start | a relaunch that cannot start ends the turn as the crash would have |

Local checks, on a private test database:
- `test/fountain/conversations`, `test/fountain/conversation*` and `test/fountain/machines`: 2499 tests, 0 failures.
- `mix precommit compile format credo dialyzer sobelow` passed. The release stage was left to CI.
- `scripts/changelog.py check` and `scripts/conflict-markers.py` passed.


apps/fountain/lib/fountain/conversations/turn_launch.ex
```diff
@@ -162,6 +163,128 @@ defmodule Fountain.Conversations.TurnLaunch do
...
+      | current_command: nil,
+        current_command_ref: nil,
+        acp_peer: nil,
+        acp_peer_mon: nil
+    }
+
+    {mode, id} = launch.plan
+
+    state =
+      case TurnMachine.session_plan(turn, mode, id) do
```
BinaryBourbon: **[P2] Check the actor's sandbox binding before retrying the launch**

`session_plan/3` is not a sandbox-binding fence for this path: `_unsafe_set_turn_session` reaches `ExecutionGuard.parent_write_allowed?/4`, where `execution == nil` makes `bound?` true. Every turn eligible for this retry is unbounded.

A reachable trigger is replacement of a shared **ephemeral** sandbox when the sibling's provider probe reports it gone and retirement of the old machine returns `:sandbox_unavailable` (for example, a lease/lock refusal after that probe). `Wake.create_fresh_sandbox_and_start/5` ignores that retirement result and `follow_cotenants/2` changes the cotenant's `sandbox_id` and clears its `runtime_session_id`, while its old turn remains running. If that old actor then handles a silent exit 139, this call writes the old launch UUID back onto the replacement conversation and `launch/5` spawns with the old `state.handle`.

I reproduced this through `Wake.wake_conversation/1` with the documented provider/retirement outcomes: after Wake moved the conversation, `runtime_session_id` was nil; after exit 139 it became the old UUID and a second spawn occurred. The same regression passes against the approved base: the former exit path calls `Machine.end_turn(..., sandbox_id: state.sandbox_id)`, which refuses the stale binding and leaves the replacement untouched.

Carry the captured sandbox binding into the locked session-plan check before writing session state or relaunching, and abandon the stale local turn when it no longer matches. Add a regression using an actual replacement binding; the existing test that stubs `_unsafe_set_turn_session` to return `applied: false` cannot detect this gap.

BinaryBourbon (changes requested):
Request changes for one reproduced P2 regression: the crash retry can act on a conversation that has moved to a replacement sandbox. See the inline finding for the production failure path and remedy.

Reviewed all nine changed files, the ACP handshake and Sprites stdout/exit ordering, fresh versus reused peers, bounded-turn exclusion, terminal cleanup, session planning, and redaction.

Validation: the existing ACP and two redaction suites passed (144 tests, 0 failures) on this head using an isolated PostgreSQL database. An additional adversarial test through the real Wake lifecycle fails on this head (1 test, 1 failure) and passes with approved-base source 6fa0f106966423d7d1738d131257034f1855bb12 (1 test, 0 failures). The probe controls provider lookup and the documented machine-retirement refusal; it does not mutate the conversation binding directly. Current GitHub CI, including all six test partitions, static analysis, release boot, coverage, and secrets scanning, is green. I did not run the complete local precommit suite or live provider crash experiments.

apps/fountain/lib/fountain/conversations/turn_launch.ex
BinaryBourbon: Fixed in a40279fdf. Confirmed the gap: for an unbounded turn, `execution == nil` makes `bound?` true, so the plan write did not compare the actor's sandbox.

- `ExecutionGuard._unsafe_write_parent/4` takes `:sandbox_id` and rolls back `:ownership_changed` through the existing `rebound?/2`, the same fence `_unsafe_recover_turn/3` uses. `_unsafe_set_turn_session/3` and the new `TurnMachine.session_plan/4` carry `state.sandbox_id`. Callers that pass no option are unchanged.
- `relaunch_crashed/3` now fences **first**, before it stops the peer, publishes the restart stage or spawns. On any refusal it returns `{:finish, state}`, so the exit ends the turn through `Machine.end_turn/3` with the actor's binding, exactly as on the base. That replaces the earlier `fail_before_start` branch for a fenced relaunch.
- Regression: "a conversation Wake moved to a replacement sandbox is not relaunched". It applies `follow_cotenants/2`'s own `update_all` to a real replacement sandbox row, then sends exit 139. It asserts no second spawn, `runtime_session_id` still nil, the binding unchanged, and no restart stage.
  - It **fails** with the binding dropped from the plan write (`refute_receive {:spawned_ref, _}`).
  - It **passes** with relaunch disabled, i.e. base behavior.

The stubbed `applied: false` test is kept, renamed and re-asserted for the new fallback: `turn/done` with exit 139.

Local checks: conversations, conversation* and machines suites, 2498 tests, 0 failures. `mix precommit compile format credo dialyzer sobelow` passed.

apps/fountain/lib/fountain/conversations/turn_launch.ex
```diff
@@ -162,6 +163,136 @@ defmodule Fountain.Conversations.TurnLaunch do
...
+
+  def relaunch_crashed(state, _code, _fail_before_start), do: {:finish, state}
+
+  defp relaunch(state, turn, spec, plan, signal, fail_before_start) do
+    Logger.warning(
+      "conv #{state.conversation_id}: adapter died on #{signal} before any output; " <>
+        "launching it once more (#2402)"
+    )
+
+    Connection.stop_peer(Connection.from_state(state))
```
BinaryBourbon: **[P2] Revalidate the binding after the blocking peer shutdown**

The new binding-aware plan check has already committed before this call. `Connection.stop_peer/1` synchronously waits in `GenServer.stop(peer, :normal, 1_000)`, so another process can move the conversation while this actor waits. Once shutdown returns, the code publishes `adapter_crashed` and calls `launch/5` on the old `state.handle` without checking the binding again.

This is reachable through the previously identified shared-ephemeral replacement path: a sibling's provider lookup reports the old sandbox gone, retirement returns `:sandbox_unavailable`, and Wake continues to `follow_cotenants/2` while this turn remains running. Scheduling that real Wake operation during peer shutdown reproduces an unexpected second spawn after the conversation has moved. The old session ID is no longer restored, but the retry still uses the stale sandbox and announces a restart on the moved conversation.

The identical scheduling probe passes on the approved base: its exit path stops the peer before `Machine.end_turn(..., sandbox_id: state.sandbox_id)` performs the final binding check. Revalidate the captured binding and turn after peer cleanup, before the restart event and new spawn (retaining the early check if desired), without holding database locks across I/O. Cover a replacement during cleanup, since the new test only replaces the binding before retry begins.

BinaryBourbon (changes requested):
Re-review of a40279f: the original before-rebind failure is fixed, but a P2 ownership race remains when the binding changes during peer shutdown. See the inline finding.

The unchanged original real-Wake reproduction now passes. The re-review's focused ACP, redaction, execution-guard, parent-write, and exit-binding checks passed (224 tests). A new deterministic test holds the real ACP peer's scheduling without changing its state, observes the actor inside Connection.stop_peer, and invokes the real Wake replacement path. It fails on this head (1 test, 1 failure: an unexpected second spawn) and passes against approved-base source (1 test, 0 failures). The probe controls documented provider/retirement outcomes; it does not directly mutate the conversation binding or stub the session-plan check.

Reviewed all 11 changed files and the fix delta. Current-head required CI is green. No full local precommit or live-provider crash test was rerun.

BinaryBourbon (changes requested):
Request changes for the remaining P2 ownership race in the crash-retry path: `session_plan/4` validates the actor's sandbox before the synchronous `Connection.stop_peer/1`, but the restart event and fresh spawn then proceed without another check. If a cotenant's Wake moves this conversation to a replacement while peer cleanup waits, the actor resumes the retry on its old handle. The former exit path intentionally stopped the peer before its final binding check. Revalidate the binding and running turn after cleanup, before announcing or spawning the retry, and cover that Wake interleaving.

I independently inspected all 11 changed files, the actual peer/transport ordering, session planning, bounded and reused-peer exclusions, failure cleanup, and retained-prompt redaction. I also inspected the adversarial reproduction's unchanged-state scheduling barrier and real Wake path; this is a scheduling window around a real blocking call.

Independent validation on a private PostgreSQL 18 database with the pinned toolchain: ACP plus both redaction suites, 145 tests, 0 failures; two additional private probes, 2 tests, 0 failures, covering stale old-command events, stderr-only retry, original timing/span continuity, and resume-mode session/prompt/image preservation. Current-head GitHub CI is green. I did not run the entire local precommit suite or live provider crash experiments.

The primary review records the head-versus-base reproduction and inline finding: https://github.com/managoat/fountain/pull/2420#pullrequestreview-5245101607.

BinaryBourbon (changes requested):
Manual independent security/ownership review of a40279fdf2157e190d82558ca6910186f0c60d51. I inspected all 11 changed paths and traced the affected launch, Wake, execution-fencing, redaction and telemetry callers.

Request changes for the remaining P2 ownership race at `TurnLaunch.relaunch/6`: the binding-aware session-plan transaction commits before `Connection.stop_peer/1` waits for peer shutdown. Wake can rebind the conversation while this actor waits, but the actor then publishes the restart event and launches the retained prompt/environment on its old sandbox handle without another check. Wake's asynchronous machine notification cannot preempt this callback. Revalidate the captured sandbox binding and current running turn after shutdown, before the restart event and spawn, and cover replacement during cleanup. This independently confirms the existing current-head finding: https://github.com/managoat/fountain/pull/2420#discussion_r4044520985.

I found no additional secret-disclosure defect: the retained launch map is replaced with its retry flag by status/crash-report redaction, and telemetry exports explicit timing/identity fields. This is a manual source review, not a Review Loop App run. I did not independently execute tests or access live provider resources; the coordinated review reports local diagnostic results separately.

BinaryBourbon (changes requested):
Request changes for the remaining P2 ownership race at `TurnLaunch.relaunch/6` (line 257): `session_plan/4` commits its binding check before the synchronous peer shutdown. If the conversation moves to a replacement during that wait, the actor still publishes the restart and spawns on its old handle. Revalidate the captured sandbox binding and current running turn after cleanup, before publishing or spawning. The existing inline finding describes the real Wake trigger: https://github.com/managoat/fountain/pull/2420#pullrequestreview-5245101607.

I independently reviewed all 11 changed paths plus the ACP/Sprites ordering, session planning, bounded and reused-peer exclusions, Wake/reassignment, completion, recovery, billing callers, and retained-prompt redaction. I found the shutdown race before reading prior reviews; the earlier before-retry binding defect is fixed on this head.

Validation on exact head a40279fdf2157e190d82558ca6910186f0c60d51: I added a private diagnostic using a real ConversationServer and database, placing Wake's exact binding write inside a controlled peer shutdown as the existing exit-binding tests do. The coordinated local run failed as expected (1 test, 1 failure: an unexpected spawn on the old handle). I inspected the shared baseline log: ACP, redaction and exit-binding checks passed (155 tests, 0 failures). No complete local precommit or live-provider crash test was run for this review.

apps/fountain/lib/fountain/conversations/turn_launch.ex
```diff
@@ -162,6 +163,136 @@ defmodule Fountain.Conversations.TurnLaunch do
...
+    # which leaves the first launch's measurements behind.
+    if state.current_command_ref,
+      do: {:relaunched, state},
+      else:
+        {:relaunched,
+         %{
+           state
+           | current_turn: nil,
+             current_turn_span: nil,
+             turn_metrics: nil,
```
BinaryBourbon: **[P2] Keep failed-turn telemetry when the retry cannot spawn**

If the original command exits 139 and the retry's spawn returns an error, launch/5 invokes fail_before_start. That path marks the admitted turn failed and closes its span, but never calls TurnMachine.emit_completed/2. This cleanup then drops the original turn_metrics. Unlike an initial spawn failure, this turn already ran; the previous exit path emitted its failed completion metric.

The missing event affects an actual operator alert: deploy/k8s/prometheusrule.yaml:222–243 computes FountainTurnFailureRate from fountain_turn_completed_duration_ms_count, including status="failed" and a minimum of ten completed samples. These failed turns disappear from both counts, so retry-spawn failures can suppress the provider failure alert.

A private extension of the existing “a relaunch that cannot start fails the turn and clears it” test confirms the row is failed but receives no [:fountain, :turn, :completed] event (1 test, 1 failure). The same observer with non-retried exit 17 receives exactly one failed completion event (1 test, 0 failures).

Emit one failed completion using the original start time when the retry's guarded terminal write succeeds, before clearing the metrics. Preserve the existing omission for a command that never initially started, and avoid emitting for a fenced stale actor. Add the telemetry assertion to the retry-spawn-failure regression.

BinaryBourbon (changes requested):
Request changes on a40279f: I found a P2 observability regression in addition to the ownership race. When the first command crashes and the retry cannot spawn, the turn becomes failed but emits no completion metric. The inline finding traces this to the deployed FountainTurnFailureRate alert, which loses these failures from both its failure count and minimum-volume threshold.

I also independently traced the ownership gap around Connection.stop_peer/1: ownership is checked before the blocking shutdown, then the restart event and old-handle spawn proceed without revalidation. The existing current-head finding documents the production Wake interleaving: https://github.com/managoat/fountain/pull/2420#discussion_r4044520985.

Reviewed all 11 changed paths, relevant lifecycle/ownership decisions, session planning, launch-data redaction, and TypeScript, Python, Elixir, Swift/FountainKit and Go event consumers. The new session/done metadata fits the existing wire contract and remains nonterminal for those callers; I found no separate compatibility or unnecessary-abstraction blocker.

Coordinated local validation used the pinned toolchain and an isolated PostgreSQL database: ACP, both redaction suites and exit-binding checks passed (155 tests). The coordinating reviewer ran my private telemetry probe: 1 test, 1 failure, confirming a failed turn without the completion event. A control with the same observer and a non-retried exit passed (1 test), asserting exactly one failed completion event. I inspected the probe source and outputs. These are local mocked-provider diagnostics; I did not run the full precommit suite or a live provider crash experiment.

apps/fountain/lib/fountain/conversations/turn_launch.ex
BinaryBourbon: Fixed in 85323aad8. `relaunch_crashed/3` now follows the exit path's own order: it stops the peer first, then runs the fenced plan write (`session_plan/4` with `state.sandbox_id`, which also requires the turn to still be running). Only after that write passes does it publish `adapter_crashed` or call `launch/5`. The fence is the last check before any side effect, with no database lock held across I/O. On refusal, the exit ends through `finish_exited_turn/2`, the base path.

Regression: `command_exit_binding_test.exs` "a crash retry whose conversation moves during peer shutdown is not relaunched". It uses the file's stand-in peer, whose `terminate/2` performs the rebinding inside `stop_peer`, on a relaunch-eligible exit 139. It asserts:
- no `turn` or `session` stage;
- `runtime_session_id` still nil and the binding unchanged;
- the stale turn still running;
- no completion metric.

With the fence moved back before the stop (the a40279fdf order), it fails.

apps/fountain/lib/fountain/conversations/turn_launch.ex
BinaryBourbon: Fixed in 85323aad8. `relaunch_crashed/3` now takes the server's exit path (`finish_exited_turn/2`) as its callback, and the relaunch uses it as the launch's cannot-start callback in place of `fail_before_start`. A retry that cannot spawn, or whose adapter install fails, ends as the crash would have:
- guarded `Machine.end_turn/3` with the crash's exit code (139);
- `turn/done` and the span ended;
- exactly one failed `[:fountain, :turn, :completed]`, measured from the first launch's `started_mono`.

A fenced stale actor still emits nothing, because `Machine.end_turn/3` returns `:noop` there. The binding test above asserts `refute_receive` for the metric. A command that never started initially still takes the unchanged `fail_turn_before_start/3`.

The retry-spawn-failure test ("a relaunch that cannot start ends the turn as the crash would have") now asserts one failed completion with an integer `duration_ms`, and no second event. It fails with the metrics dropped on that path.

BinaryBourbon (approved):
Approve after a fresh independent correctness/concurrency review of the complete PR at 85323aad8fb72a6fe85a9a0ed44c8febe2b650a7 against approved base 6fa0f106966423d7d1738d131257034f1855bb12 (three-dot diff). I found no remaining concrete correctness defect.

Both previously reported regressions are resolved in the code: peer shutdown now finishes before the sandbox-binding/current-turn session-plan fence; a retry that cannot start now uses the guarded original exit path, retaining its crash code, original timing/span and exactly one failed completion metric. I independently traced stdout/exit ordering through Sprites and the ACP initialize barrier, fresh versus reused connections, bounded-turn exclusion, stale command reports, retry exhaustion, session identity, terminal cleanup, output redaction and the single-turn billing identity. I read the historical findings after the source pass and checked their claimed fixes against the implementation and regression assertions.

Coordinated validation on this exact head used the pinned toolchain and isolated PostgreSQL: 235 focused ACP, redaction, exit-binding, execution-guard, parent-write and turn-metrics tests passed. The previous binding probe, real-Wake replacement-during-peer-shutdown probe and retry-failure telemetry probe each passed (1 test each); tenant-denial checks passed (2 tests). I inspected the shared logs and the Wake/telemetry probe source; the coordinating reviewer executed these commands. This is a manual source review, not a Review Loop App run. I did not independently execute the complete local precommit suite or a live-provider crash experiment.

Coverage (all 12 changed paths):
- `apps/fountain/lib/fountain/conversations/connection.ex`
- `apps/fountain/lib/fountain/conversations/execution_guard.ex`
- `apps/fountain/lib/fountain/conversations/redaction.ex`
- `apps/fountain/lib/fountain/conversations/turn_launch.ex`
- `apps/fountain/test/fountain/conversations/command_exit_binding_test.exs`
- `apps/fountain/test/fountain/conversations/conversation_server_redaction_test.exs`
- `apps/fountain/test/fountain/conversations/redaction_test.exs`
- `changelog.d/2402-adapter-crash-relaunch.md`

BinaryBourbon (approved):
Manual independent security/ownership review of 85323aad8fb72a6fe85a9a0ed44c8febe2b650a7: no blocking findings. I inspected all 12 changed paths against the trusted base's three-dot diff and traced the affected API/launch entrypoints, established-owner actor, Wake/binding fence, ACP handshake and command ordering, failure cleanup, redaction, and telemetry callers.

The previously reported shutdown race is fixed: peer cleanup completes before the binding-aware session-plan transaction checks the actor's captured sandbox and current running turn, and before the restart event or spawn. A refused retry returns through the guarded exit path. A retry that cannot start preserves the original measurements and emits the failed completion through that same path. The retained launch context is stripped to its retry flag in crash/status output; telemetry exports explicit identity/timing fields. Scoped conversation and agent lookups remain ahead of user-facing launch/prompt operations.

Validation: I inspected the coordinated exact-head test outputs: 235 focused tests passed; the unchanged previous controlled-peer ownership probe, real-Wake shutdown-interleaving probe, and failed-retry telemetry probe each passed; two foreign-tenant API denial tests passed. I inspected the regression assertions and relevant probe source but did not independently execute these tests. My diff whitespace check passed, and I verified the current head, BinaryBourbon identity, and passing required CI checks immediately before submission. No full local precommit or live-provider experiment was rerun. This is a manual review, not a Review Loop App run.

BinaryBourbon (approved):
Independent adversarial review — API/product compatibility, maintainability, and observability.

Reviewed the full 12-file PR at `85323aad8fb72a6fe85a9a0ed44c8febe2b650a7`, using the supplied base's API/simplicity review policy, contributor rules, and relevant ACP, bounded-execution, and machine-ownership ADRs. No blocking findings remain.

The two previously reported defects are repaired in the actual control flow:

- Peer shutdown precedes the sandbox-bound session-plan write. A conversation rebound during shutdown refuses the relaunch before its announcement or spawn; the ordinary completion path independently checks the binding.
- A retry that cannot spawn reaches the ordinary exit finalizer with the original crash code, span, and monotonic start. An applied failure emits one `turn.completed` sample, preserving the counts used by `FountainTurnFailureRate`.

I traced stdout/exit ordering through the command and ACP peer, the fresh-launch eligibility guard, one-retry cap, bounded/reused-peer exclusions, same-turn/session continuity, stale-reference handling, and redaction of the retained launch context. The new `session/done` event uses the existing vocabulary and remains nonterminal in the TypeScript, Python, Elixir, both Swift clients, main CLI/ACP gateway, and Hermes integration. The existing OpenAPI/committed log-event shape accommodates its metadata; the Buzz CLI adds no independent turn-stream interpretation. Public internal-call arities remain compatible.

Coverage: `conversations.ex`; `connection.ex`; `conversation_server.ex`; `execution_guard.ex`; `redaction.ex`; `turn_launch.ex`; `turn_machine.ex`; `command_exit_binding_test.exs`; `conversation_server_acp_test.exs`; `conversation_server_redaction_test.exs`; `redaction_test.exs`; and `changelog.d/2402-adapter-crash-relaunch.md`, plus affected callers/consumers.

Validation: I independently ran a Python SDK smoke against this exact-head archive: the restart event leaves the follower and run classifier waiting, and the terminal turn event retains exit code 139. I also inspected the coordinated local results: 235 focused tests passed; the original binding probe, real-Wake probe, and original completion-metric observer each passed; two tenant-denial tests passed. The metric observer's terminal-stage expectation was updated to the intentionally preserved `turn/done` path. This review did not rerun full precommit or live provider/inference exercises. This is a manual review, not a Review Loop App execution.

PR #2421, opened 2026-09-18
fix(sandboxes): adopt absolute exec collection deadlines

jhgaylor:
Finite Sprites and E2B `exec` calls previously refreshed their local collection timeout on every output frame. Adopt published Sandbox 0.5.0 so continuous stdout or ignored stderr cannot replenish that deadline, together with Runner 0.2.4 and Runtimes 0.4.4 for compatible dependency resolution. The lock changes only these three packages; existing Sprites, Gun and Cowlib pins are preserved.

This is a prerequisite for #2394. Synchronous startup bounds, remote command termination, durable read admission and lifecycle priority remain open. The optional explicit Sprites force-stop API is available in the dependency; this PR does not switch Fountain's existing termination calls to force mode or claim the full read/park/delete exclusion guarantee.

The historical counterexample assertions from #2417 remain intact. Their documentation now reproduces the old pinned behavior in a detached checkout at `6fa0f1069` and points to the library's current correctness regressions; these opt-in scripts are outside required CI.

Reviewed prerequisite releases:
- https://github.com/managoat/managoat_sandbox/pull/17
- https://github.com/managoat/managoat_runner/pull/9
- https://github.com/managoat/managoat_runtimes/pull/22

Validation against the real published artifacts:
- Affected sandbox-file, controller, deadline worker, provider, runner and runtime integration tests: 146 tests, 0 failures, using the dedicated `fountain_2394_test` database.
- `mix precommit`: all nine server stages passed (379 seconds), including Dialyzer, Sobelow and production release assembly.
- Conflict-marker, changelog and whitespace checks passed on the staged files.


BinaryBourbon (approved):
Adversarial review of 3207194e1e34dfcbde46b9f258f39de2e8883671: approved. No actionable regression found across all five changed files, the adopted library source, and Fountain's exec/termination callers.

Verified the three published Hex artifacts' inner and outer checksums against mix.lock; Sandbox 0.5.0 library bytes match release commit 09303b4. Runner executable code is unchanged, and Runtimes changes only dependency compatibility and documentation. Sprites/Gun/Cowlib pins remain unchanged.

Validation: 55 isolated deadline and termination regression tests passed against Fountain's locked runtime dependencies, using Elixir 1.19.2/OTP 28.3. Cases include continuous stdout and ignored stderr, queued frames, exhausted startup/lookup budgets, terminal success after expiry, zero timeout, infinity, and graceful/force termination. Changelog and exact-head whitespace/conflict-marker checks passed. Required CI and Detect secrets pass for this revision.

The docs correctly separate historical counterexamples from the corrected collector behavior. This review does not establish live-provider termination guarantees: synchronous startup bounds, remote descendants, and read/park/delete coordination remain the explicitly scoped follow-up work in #2394. No live provider resources were used.

PR #2370, opened 2026-09-17
feat(machines): admit and end turns through the owner (ADR 0058, stage 8a)

jhgaylor:
Stage 8a of ADR 0058, *The machine has one owner* (#2344): **turn admission through the owner.**

## What

`Fountain.Machines.Admission` — the fifth protocol. `Machine.admit_turn/3` is the only way a turn row is created, `Machine.end_turn/3` the only way an actor ends one, and `Machine.at_capacity?/3` the unlocked read the two doors render a refusal from. The locked insert stays in the context and stays one transaction under the per-sandbox advisory lock; what changes is what it decides **under that lock**: a live lease refuses the turn, either fence refuses it, and capacity is the runtime's, counted against that runtime's turns. A lexical pin in `direct_writes_test.exs` — the turn-row half of the ratchet — asserts that the context's four turn writes are called from `lib/fountain/machines/` only.

## Why

The turn insert was the last writer that never asked whether an owner was operating on the machine. A prompt landing during a cotenant's park, resume or destroy started a turn against a machine about to change under it. And #1089 blocker 4 was live: `_unsafe_running_turns_elsewhere/2` counted every runtime's turns while the caller computed capacity from its own runtime, so a `claude` turn on a shared home consumed the one `opencode` slot.

## The protocol as shipped

```
refuse an enclosing transaction
── one transaction, 4316 held throughout, as on main ──
  inference-source lock; parent FOR UPDATE; revision
  the machine's row FOR SHARE, with statement_timestamp()
    bound to this conversation and tenant   — else :sandbox_unavailable
    not terminal                            — else :sandbox_unavailable
    neither fence                           — else :sandbox_unavailable  (teardown is new)
    no live lease                           — else :machine_busy         (new)
  terminal parent; saved allowance
  capacity = concurrency(conv.runtime), against that runtime's running turns   (new)
  turn insert; parent → running; bounded journal
── commit ──
:machine_busy is waited out for Admission.busy_wait_ms (5 s), polling the write; every other refusal is final
```

[...]

PR #2423, opened 2026-09-18
feat(machines): delete what the owner replaced (ADR 0058, stage 9b-i)

### LINKED ISSUE #2021: Tracker: remaining sandbox-binding lifecycle gaps (#1767, ADR 0058)

jhgaylor:
Stage 9b-i of [ADR 0058](decisions/0058-the-machine-has-one-owner.md), off `main` after 9a (#2386). Tracker #2344.

**Do not queue this PR.** It deletes `MACHINE_OWNER_ENABLED`, and with it the one-line rollback. The lead queues it only after production has shown at least one park driven by the owner and several clean hourly passes of 9a's driver. 9b-ii, which drops the two columns and moves the ADR to Accepted, is a separate PR one release later.

## What this deletes

1. **The gate.** `MACHINE_OWNER_ENABLED`, its `config/runtime.exs` line, the `docs/configuration.md` row, `Machines.enabled?/0`, and the six gate splits in `Machine`. `destroy`, `park`, `ensure_up`, `admit_turn`, `attach` and `detach` always run in the owner. `provision`, `confirm_up`, `fail_provision`, `end_turn`, `retarget`, `bind_inference` and a `detach` with `policy: :keep` still run on the caller, by design (7b, 8a, 8b). The gate has been on in production since 2026-09-17.
2. **`SandboxResetReconciler`'s work.** It moved into `SandboxReaper.sweep_fenced_teardowns/0`, the one driver for abandoned destroys, which runs on its own five-minute cron entry (see "The driver" below). The module stays for one release as a no-op with no cron entry (round 1, S2). A job the previous release enqueues during the rolling deploy then completes, where it would otherwise fail as an unknown worker, be discarded after ten attempts and fire the critical `FountainObanJobsDiscarded`. 9b-ii deletes it.
3. **Every read and write of `reset_requested_at` and `teardown_requested_at`.** Both are no longer fields on `Sandbox`. The columns stay, nullable and unread, until 9b-ii.
4. **`release_stuck_sandboxes/0`'s direct write.** It asks the owner now (`Machine.fail_provision/2`).
5. **`Conversations.update_sandbox/2`**, whose last caller in `lib/` was item 4, with `sandbox_retired?/1` and the revival guard. It lives on as a test fixture, `Fountain.Factory.update_sandbox/2`.

The ratchet reads **0 row writes and 2 provider calls**. The brief asked for 0/0. The two calls that remain are explained under "Numbers".

## Rule 16: the path that now handles each deleted arm

| Deleted | Its case is now handled by |
|---|---|
| Each inline branch in `Machine` | The in-owner branch beside it, which production has run since 2026-09-17 |
| `SandboxResetReconciler`'s sweep and per-sandbox job | `sweep_fenced_teardowns/0`, which retries a row with reason `"reset"` through `retry_pending_sandbox_reset/2`, the door the reconciler called |
| The reaper's hourly driver call, `@driver_floor` and the budget split | The five-minute teardown run, which is not charged to the hourly budget |
| Every column fence read (Wake, `attachable/5`, Park, Provision ×2, Resume, Admission's locked insert, `Lease` refuse_fenced and its zero-row diagnosis, Quotas ×3, HomeCheckpoint, the reset door, its retry, the reaper's two liveness passes, the admin Retry reset button) | The `destroying` stamp read beside it since 9a, and now alone. Both fences write it. |
| The two fence writes of the columns | `Machines.Destroy.stamp_intent!/2`, the stamp alone, still inside each fence's own locked transaction |
| Destroy's column-only `caller_fenced_destroy/3` clause | Nothing: only a replica from before 9a could write a column-only fence. A row that reaches the clause unstamped is refused `:not_fenced`. |
| `release_stuck_sandboxes/0`'s `update_sandbox/2` write | `Machine.fail_provision/2`, which takes `pending`/`starting`, the pass's whole population |
| `update_sandbox/2`'s revival and fence guards | `Lease.cas_update/4`'s `refuse_revival` and `refuse_fenced`, on the owner's write, which is the only status writer left |
| `MachineEvents.reset/6`'s `reset_requested_at` condition | The cast itself (below) |

## Behaviour changes, each with its symmetric case

1. **The reset notice trusts its cast** (Jake, 2026-09-18). `MachineEvents.reset/6` asks only that the machine is terminated. Only the reset's completion sends `{:sandbox_reset, ...}`, so the message is the evidence. The dropped column check never told a reset from a teardown, because the teardown fence set it too. **Accepted residual:** a late reset notice that arrives after a *later* teardown of the same machine could bring a conversation back to `idle`. The dropped check did not prevent that either. Pinned by a new test: the cast for a terminated machine with no stamp moves a `running` conversation to `idle`. Symmetric case: a notice for a machine that is not terminated still changes nothing, which the existing "cannot retire an unfenced live actor" test covers.
2. **One five-minute pass finishes every abandoned destroy.** A reset is retried on the next run; a teardown is finished on the first run after its stamp is fifteen minutes old (round 1, S1: the first push said "within five minutes" for both, which was wrong for teardowns). Before this PR, an abandoned teardown of an **ephemeral** machine waited for the hourly run, behind a 5-destroy floor. An abandoned reset, or a forced teardown of a persistent home, went through the reconciler every five minutes with no cap. Now one pass takes both, every five minutes. Teardowns keep the 15-minute grace window and the live-server check. Resets keep the reconciler's shape: no grace window, no server check, and skipped on a provider without credentials. Symmetric case: a machine nobody asked to destroy is left alone, as is one whose owner holds a live lease. Each has a test.
3. **An attach to a home that was reset and then terminated** answers `{:sandbox_not_attachable, "terminated"}` where it answered `:sandbox_reset_pending`. The column stayed on a row after its reset completed, and `attachable/5` refused on it at any status. Both are 409. The new answer is the one every other terminated machine gets.
4. **A stuck `pending`/`starting` row whose lease is live is left alone.** The old write could not tell such a row from an abandoned one. `fail_provision/2` claims the lease and answers `{:ok, :claimed_elsewhere}`. New test. Symmetric case: an unleased stuck row is still failed, now with `sandbox.provision_failed` beside `sandbox.released_stuck` (test).
5. **A teardown fence on a row already stamped for a teardown is a repeat** and writes nothing, not even a new reason. The column told "repeat" from "new intent" until now. On a reset's stamp the fence is an escalation: it rewrites the reason and records `sandbox.teardown_requested`, as before. Tests cover both.
6. **Retired rows no longer carry fence evidence.** The columns stayed on a terminal row. The stamp is cleared at the finalize. The audit trail (`sandbox.teardown_requested`, `sandbox.reset_requested`, `sandbox.destroyed`, `sandbox.reset`) is the evidence now.

## The driver: why one five-minute run and not the hourly one

The brief said to fold the reconciler into the reaper's `destroying` sweep. I did, but not into the hourly run. 9a's reviewers established two properties of the reconciler: it matches forced teardowns, and it has no budget. They also established that `@driver_floor` was sized on the assumption that it always would. If the reconciler's rows had gone into the hourly budgeted pass, an abandoned reset or a forced teardown of a persistent home would wait up to 75 minutes instead of 5. Those homes would also sit under a floor that was never sized for them. So the pass took the reconciler's clock (`{"*/5 * * * *", SandboxReaper, args: %{"pass" => "teardowns"}}`) and dropped out of the hourly budget. The hourly run's saturated maximum goes from 35 provider destroys to 30. The teardown run is capped by `@owner_attempts_per_sweep` (100). The `:maintenance` queue runs one job at a time, so a backlog drains one machine at a time. Oban keys `unique:` on args, so the two runs are unique apart. Lifeline rescues either.

**Telemetry moved.** `reconciled` left `[:fountain, :reaper, :run]` for `[:fountain, :reaper, :teardowns]`, which also carries that run's `refused`. Prometheus: `fountain_reaper_run_reconciled` is now `fountain_reaper_teardowns_reconciled`, and `fountain_reaper_teardowns_refused` is new. `fountain_reaper_run_refused` counts only expiries and idle parks now.

## Tests: what deleting the gate cost

With the gate off, the suite ran every verb inline on the test process. Deleting it moves every verb into an owner process, and the first measurement was **300 failures**, all of two shapes:

- **The owner could not reach the test's connection or stubs.** In the test suite, the owner adopts its **caller's** `$callers` chain on every call, read with `process_info(pid, {:dictionary, :"$callers"})` for a local caller. It is per call, not per start. A chain recorded at start ran a later caller's work on the first caller's connection, and inside `unboxed_run` races that deadlocked. **This is test-only** (round 1, P1, decided by Jake). The owner adopts only when `:owner_adopts_callers` is set, which `config/test.exs` alone does. With it off, the owner never puts the key, which is `main`'s owner exactly. The first push said nothing in production reads `$callers`, and that was wrong. `opentelemetry_ecto` walks it on every query a process with no trace context makes. Adopting in production would copy the caller's whole process dictionary per owner query and re-parent the owner's database spans. `machine_test.exs` switches the flag off and asserts the owner puts nothing; it fails with the guard removed.
- **48 files stub `Horde.DynamicSupervisor.start_child/2`** to fake a conversation server, and the stub swallowed the owner too. They now call `Fountain.ServerStart.stub_server_start/1` (and `expect_`/`reject_`), which lets `Fountain.MachineSupervisor` through.

Race tests across real connections in `async: false` files (`setup :manual_pool`) switch the pool to manual mode. In shared mode, DBConnection resolves only the asking process, so it hands the owner the shared sandbox transaction. Its row locks then outlive the race and block the cleanup. In manual mode the pool follows the adopted chain to the caller's own connection, which is where the verb ran inline.

**The winner checks are exact** (round 1, B1). The first push loosened two of them to accept either the task or the owner, and a lockless reset fence then passed the reset-first arm. Now each names the process that takes the lock. In the reset/admission race that is the task for both sides: the reset fence and the locked turn insert both run on the caller, and the pause extension I had added there is reverted. In the termination/attach race it is the task for a termination and the owner for an attachment, which runs its insert there. Planting the fence's advisory lock on a different key fails both reset/admission arms and the machine-lock arm of termination/attach. The two row-lock arms of termination/attach serialize on the row lock, which that plant leaves in place, so they pass it by design.

The #1969 race (`retirement_admission_order_test.exs`) raced admission against `update_sandbox/2`'s locked write, and that writer is gone. The race now runs against the owner's compare-and-set, `Lease.cas_update/4`, which takes no advisory lock, so it is exactly the writer admission's `FOR SHARE` hold exists for. **One difference from production, stated in the test** (round 1, B3): production's finalize is a single statement outside any transaction, and the test runs it in a transaction so the retirement-first arm has a row lock to hold across the pause. The admission-first arm matches production exactly.

The column-blind work matched 9a's map, one file at a time. The two artifacts the brief named: the mixed-version reaper test was deleted with the column, and `destroy_test.exs`'s fixture back-dates `updated_at`.

## Before deploy: the query that has to read zero

9b-i stops reading the columns. A live row whose intent lived only in a column would lose that intent. Run this before the deploy:

```sql
select count(*) from sandboxes
 where status not in ('terminated','failed')
   and (reset_requested_at is not null or teardown_requested_at is not null)
   and (transition is null or transition <> 'destroying');
```

At 2026-09-18 07:10 UTC it read **0**. There were 404 live rows, none with a column set and none stamped. No backfill migration is needed while it reads 0.

## Mixed-version rollout

- **An old (9a) replica writes the columns and the stamp.** This release ignores the columns and reads the stamp, which the old replica also writes. The columns stay nullable until 9b-ii.
- **An old replica runs `SandboxResetReconciler`'s cron.** If it is the Oban leader, it inserts reconciler jobs. A new node runs them on the one-release no-op shim, so they complete on their first attempt instead of failing as an unknown worker and being discarded — verified by a reviewer draining both old job shapes through the real Oban executor with no exception event and no discard. The five-minute teardown run covers the same rows.
- **A new node leads cron.** Its `%{"pass" => "teardowns"}` jobs reach an old node's `perform/1`, which ignores args and runs a whole hourly pass. For the minutes of the rollout, an old node can run the hourly passes every five minutes. Every one of them is guarded and idempotent.
- **Messages to the owner are unchanged.** No message shape changed.

## Numbers

| | before | after |
|---|---|---|
| `@row_writes` | 6 | **0** |
| `@provider_mutations` | 3 | **2** |
| `conversation_server.ex` | 2024 lines on `main`, pin 2025 | untouched. The pin stays, because this PR removes no line from the file. |
| core suite | 6708 at 9a | **6820** at this head, 0 failures. The count moved with `main` as well as with this PR. |

The row writes went two at a time:

- **6 → 4:** the two fences became `Destroy.stamp_intent!/2`.
- **4 → 3:** the stuck pass went through the owner.
- **3 → 0:** `update_sandbox/2` was deleted, `create_sandbox/1` became `Machines.Provision.reserve/1`, and the wake marker became `Machines.Binding.mark_woken/1`.

`HomeCheckpoint` moved whole to `Fountain.Machines`: it runs under the park's lease, and "the checkpoint happens inside the transition".

**Why provider mutations stop at 2 and not 0.** The environment warm-start checkpoint runs in a detached task, off the lease. 7b recorded why holding the lease across it would answer 503 to prompts. It is also off by default and cannot restore on Sprites (#654). The reaper's destroy of a machine whose row is already terminal stays in the reaper, as the ADR's Decision says, because no owner will ever claim it. Moving either line into `lib/fountain/machines/` would satisfy the scan without changing who calls the provider.

## Verified by reverting

32 behaviours were each planted back one at a time, found by distinctive string. Each was restored with shell `cp` and `touch`, so mix recompiled, and checked with `cmp`. Every plant fails at least one test named for what it breaks. The harness is the one 9a's record asked for, not `shutil.copy2`.

| Plant | Fails |
|---|---|
| a direct `update_all` on `sandboxes`, and a direct `Managoat.Sandbox.suspend/1`, outside `machines/` | the 0/2 ratchet |
| the owner stops adopting its caller's `$callers` | 30 tests |
| the stub helper stops letting `MachineSupervisor` through | 24 tests |
| `MachineEvents.reset/6` asks for a stamp as well | the new pin, and the existing reset-notice test |
| the driver skips resets; puts resets behind the grace window; finishes resets as teardowns; stops checking the lease; loses its cron entry; runs from the hourly run; stops reporting `refused` | one to two named reaper and destroy-reset tests each |
| the stuck pass counts `:claimed_elsewhere` as released | the new leased-row test |
| the teardown fence treats a reset's stamp as a repeat, or treats nothing as a repeat | the escalation and repeat tests |
| each stamp reader made blind in turn: Quotas (both arms and the fleet exclusion), Park's foreign-stamp refusal, `attachable/5`, Provision's two, `Lease`'s `refuse_fenced`, Admission's locked insert, Wake, the reset door, its retry, `register_server/2`, HomeCheckpoint, the admin Retry reset button, the reaper's two liveness passes | 1 to 11 tests each |
| `caller_fenced_destroy/3` destroys an unfenced row | the `:not_fenced` test |

**Round 1 added eight plants**, each failing its named test: the `$callers` flag guard removed; the reset fence's advisory lock on another key; the teardown fence's likewise; resets given the live-server check; contention counted as refused, and its symmetric case; the shim failing; and the teardown run skipping the lease check under the reporting stubs that replaced a rescued `reject` (B6).

**The sweep found one unbreakable guard.** `Park.admissible/2`'s new `destroying` clause failed nothing, because `under_lease/3` refuses every foreign stamp first. With the columns unread it could not be reached, so it is deleted and says so in place (9a's rule for `Resume.admissible/2`). The plant that removes `under_lease/3`'s refusal instead fails 5 park tests.

## 9b-ii checklist

- Drop `reset_requested_at` and `teardown_requested_at` (migration, `okf`, root `ecto.migrate`).
- **Delete the `SandboxResetReconciler` shim** and its test.
- ADR 0058 to Accepted with its Outcome, including the Rollout section's gate text and the `MachineEvents.reset/6` plan that this PR carried out. Amend 0023's Outcome and add 0017's missing amendment note.
- Close #2021's remaining items and #2344.
- `docs/concepts/sandboxes.md` and `docs/configuration.md`: already free of the gate after this PR; re-check.

## Residuals

- 9b-ii: drop the columns, move ADR 0058 to Accepted, amend 0023 and 0017, close #2021's items and the tracker.
- Events that no longer fire, for the dashboards PR: none of the four audit actions stopped. `sandbox.teardown_requested`, `sandbox.reset_requested`, `sandbox.teardown_reconciled` and `sandbox.released_stuck` are all still recorded. `sandbox.released_stuck` now has `sandbox.provision_failed` beside it. The metric `fountain_reaper_run_reconciled` is renamed (above). The actor `system:sandbox_reset_reconciler` no longer appears, and resets finished by the driver carry `system:sandbox_reaper`.
- **The five-minute teardown run has no canary alert.** `FountainReaperSilent` watches only the hourly run's `released`. This is a home-cloud change for the lead's dashboards PR: alert on staleness of `fountain_reaper_teardowns_*`.
- A teardown-run job orphaned in `executing` makes the next cron insert dedupe onto it, so the run waits up to 30 minutes for Lifeline (round 1, P3). The deleted reconciler had the same shape, so this is not a regression.
- The ratchet cannot see raw SQL writes (`Repo.query!` with an `UPDATE`, an `Ecto.Multi`, a raw `execute`). None exist in `lib/` today, and its moduledoc says so (round 1, B7).
- A `refused` from the teardown run excludes a machine another run or operation holds (round 1, P4): on `sandbox_unavailable` the pass re-reads the row, and a live lease or a row gone terminal counts on neither gauge.
- The owner's two pre-8b compatibility clauses (`{:ensure_up, opts}` and the deadline-less `{:admit_turn, ...}`) are left in place. No current caller sends them.

## Gates

`mix precommit --full` **PASSED** (10 stages): compile with warnings as errors, `deps.unlock --check-unused`, format, `credo --strict`, dialyzer, sobelow, the release assemble, and the whole suite. The core suite ran 6812 tests with 0 failures and 2 skipped, and the five sibling apps passed. `scripts/test-docs.sh` and `scripts/changelog.py check` are green. Format was checked from the root and from `apps/fountain`. The contract is unchanged.

## Carried into 9b-ii (from round 2, all non-blocking)

- Pin the **reset** branch of the contention fix: a reset twin of the "counts on neither gauge" test (deleting that branch left this PR's tests green; only a reviewer's probe caught it).
- Three comments still describe deleted functions in the present tense: `lease.ex:680`, `lease.ex:691`, `destroy.ex:767`.
- `machine.ex:1214` says the flag "must stay off by accident" — inverted; it means must not be turned *on* by accident.
- `docs/guides/operate/sandbox-lifetime.md`: a teardown whose provider call fails counts as *reconciled* (the row retires, and the hourly terminal-row pass deletes the machine), not *refused* — add the clause.
- Delete the `SandboxResetReconciler` no-op shim.


BinaryBourbon (changes requested):
## Surfaces review: what a user, an operator, a client and the docs see

Reviewed head `eb8fd083208b950bf7a77a027055b5c070647573` in a private worktree and test database. Every result below was produced with a probe.

**Verdict: request changes.** The code behaves as claimed on every surface I drove. Two public manual pages and the changelog still describe the reconciler this PR deletes, and rule 15 needs them to describe the code. The fixes are prose only.

### Blocking

**1. Two public docs pages describe the deleted `SandboxResetReconciler` and the hourly driver.**

- `docs/concepts/sandboxes.md:99-104` says: "Every five minutes, Fountain finds pending resets and queues a separate retry for each machine. Failed deletes retry with backoff … A long outage remains eligible for later retries even after a retry job exhausts its attempts." None of that exists after this PR. No per-machine job is queued, and there is no backoff and no per-job attempt budget. `sweep_fenced_teardowns/0` calls the reset door inline on each five-minute run, capped at `@owner_attempts_per_sweep` (100) per run. It stays eligible on every later run with no exhaustion.
- `docs/guides/operate/sandbox-lifetime.md` has four stale statements:
  - Line 68 says there is one summary line per run, "each hour at :07". A second line now appears every five minutes: `reaper: teardowns reconciled=N refused=N` (`sandbox_reaper.ex:187`).
  - Line 72's example hourly line contains `reconciled=0`. The hourly line no longer prints `reconciled` (`sandbox_reaper.ex:217-220`). An operator grepping for it will find nothing.
  - Lines 91-98 say "the reaper's next **hourly** run finishes it". The five-minute run finishes it now. The text also says the trail records `sandbox.teardown_reconciled`, but `reconciled` now counts resets too, and a finished reset records `sandbox.reset` with no `teardown_reconciled` (probe below). Line 78's `refused` also still includes "finish deleting", which moved to `fountain_reaper_teardowns_refused`.
  - Lines 100-104 say abandoned deletions "are guaranteed a small share of their own on top of that". That describes `@driver_floor`, which this PR deletes. They now have their own run and cap.
- `changelog.d/2344-machine-owner-always-on.md:7` says an abandoned deletion "is finished within five minutes". That holds for a reset. A teardown still waits out `@fenced_teardown_grace_minutes 15` (`sandbox_reaper.ex:814`, filter at `s.transition_reason == ^@reset_reason or s.updated_at < ^cutoff`). So a teardown is finished 15 to 20 minutes after its stamp. The rest of that entry is accurate.

`scripts/test-docs.sh` passes because it checks structure and links, not content.

### Non-blocking

1. **A leftover `SandboxResetReconciler` Oban job fails on the new release, and after 10 attempts it is discarded.** Main's reconciler inserts per-sandbox jobs with `max_attempts: 10` and snoozes them for 60 or 300 seconds. During a rolling deploy, an old Oban leader can also insert the `*/5` sweep job, and a new pod can pick it up. I inserted such a row and drained the maintenance queue:
   ```
   PROBE drain: %{discard: 0, cancelled: 0, success: 0, failure: 1, snoozed: 0}
   PROBE job: %{state: "retryable", attempt: 1,
     errors: [%{"error" => "** (RuntimeError) unknown worker: Fountain.Workers.SandboxResetReconciler"}]}
   PROBE exception event: {"Fountain.Workers.SandboxResetReconciler", :failure, "unknown worker: ..."}
   ```
   That fires the home-cloud alert `FountainObanJobsRaising` straight away. On the tenth attempt, hours later, the job is discarded and `FountainObanJobsDiscarded`, which is **critical**, fires. Either keep a one-release shim module that returns `{:cancel, :retired}`, or add this to the queue checklist: after the deploy, run `update oban_jobs set state='cancelled', cancelled_at=now() where worker='Fountain.Workers.SandboxResetReconciler' and state in ('available','scheduled','retryable')`.
2. **The new five-minute run has no canary.** `FountainReaperSilent` keys on `absent(fountain_reaper_run_released)`, which only the hourly run emits, so the hourly run is still covered. If the `*/5` cron entry stops running, abandoned resets hold quota for good and nothing alerts. Two alerts are worth adding to the home-cloud overlay: `absent(fountain_reaper_teardowns_reconciled)` for about 30 minutes, and a sustained `fountain_reaper_teardowns_refused > 0`. The telemetry comment already says a non-zero `reconciled` is "worth an operator's attention".
3. ADR 0058's Rollout section still says "Behind `MACHINE_OWNER_ENABLED` (runtime configuration, default off)". Stage 9's row plans for `MachineEvents.reset/6` to read "most likely the reset's own `sandbox.reset` event", but the implementation trusts the cast, by Jake's decision. The PR body defers the ADR to 9b-ii, and that is fine. 9b-ii should record both changes.
4. `docs/troubleshooting/conversation-stuck-or-failed.md:96` shows an older hourly line without `refused` and `skipped`. That drift predates this PR, but it is worth fixing in the same doc pass.

### Held up under attack

- **The API change, through the real routes.** I ran a real reset on a persistent home, set `reset_requested_at` on the terminated row to stand in for a 9a replica, then drove each route:
  ```
  PROBE attach: {409, {"error":"sandbox_not_attachable","message":"the sandbox is terminated; a conversation attaches only to a ready or suspended one","status":"terminated"}}
  PROBE prompt: {200, {"status":"queued","client_request_id":null}}
  PROBE team rotate (before prompt): {201, ... new ephemeral sandbox, "status":"pending"}
  ```
  Only `POST /api/conversations` with `sandbox_id` changes. The team route calls `door_open?` only when `reusable_sandbox?` holds, which means ready or suspended, so a terminated home provisions fresh on both main and this PR. Wake's removed clause only matched non-terminal rows. As a client, I find `sandbox_not_attachable` with `status: "terminated"` the right answer. The old `sandbox_reset_pending` message says "Fountain retries automatically", so a client loop keyed on it would have retried forever against a finished reset.
- **Contract and SDKs.** 409 with the shared `Error` schema is declared on `POST /api/conversations`, `POST /api/conversations/{id}/prompts` and `POST /api/team/{agent_id}/conversations`. `scripts/sdk-contract/build.sh --check` reports `SDK wire contract: ok (173 operations, 195 schemas)`, so `release:skip-sdk` is right, as in #2368, #2369, #2370 and #2386. None of the four SDKs lists either code as retryable. TypeScript's `FountainErrorCode` is open (`string & {}`) and already names `sandbox_not_attachable`. Every SDK surfaces the code as a plain API error, and none crashes or loops.
- **Telemetry rename.** Nothing in this repo or in `jhgaylor/home-cloud` (read only) refers to `fountain_reaper_run_reconciled` or to the retired actor. Home-cloud's only reaper rules are `fountain_reaper_untracked_count` and `absent(fountain_reaper_run_released)`, and the hourly run still emits `released` (`sandbox_reaper.ex:264`). The rename silences no existing alert or dashboard.
- **Audit trail.** I compared an owner reset with an abandoned reset that `SandboxReaper.perform(%{"pass" => "teardowns"})` finished, with sprites credentials enabled:
  ```
  owner:     sandbox.reset_requested(api) · sandbox.reset(api, reason home_reset)                     · usage sandbox_terminated{final_status terminated}
  abandoned: sandbox.reset_requested(api) · sandbox.reset(system:sandbox_reaper, reason reset_reconciled) · usage sandbox_terminated{final_status terminated}
  ```
  The metadata and usage rows are identical apart from the actor and the `reset_reconciled` reason. Main's reconciler recorded the same reason through the same door. Neither path records `sandbox.destroyed`, because a reset suppresses it and records `sandbox.reset` instead, and main does the same. Nothing reads `system:sandbox_reset_reconciler`.
- **Budget.** No doc states the 35 or 30 figure, so no number in the docs is wrong. Only the prose described in blocking item 1 is.
- **`MACHINE_OWNER_ENABLED`.** It is gone from `config/runtime.exs`, `docs/configuration.md` and `.env.example`. Nothing in `lib/` or `config/` reads the variable or `:machine_owner_enabled`. A self-hoster who still sets it gets no error and no effect, and the changelog's upgrade note says so.
- **Checks.** `okf validate decisions` returns valid with 0 errors. `scripts/decisions-index.sh` leaves no diff. `scripts/changelog.py check` passes (56 fragments). `scripts/test-docs.sh` reports 13 tests and 0 failures. CI on the head SHA shows 25 success, 3 skipped, and nothing pending or failed.

BinaryBourbon (approved):
## Protocol review: the owner process, `$callers`, the deleted gate, and the teardown run

Reviewed at `eb8fd083208b950bf7a77a027055b5c070647573`, in my own worktree and database. Every claim below was driven, not read.

### Blocking

None.

### Non-blocking

1. **Adopting `$callers` changes production, and the comment says it does not.** `machine.ex:1207-1215` says that only test tooling reads the chain. The PR body says that nothing in `lib/` reads it. The first claim is wrong, and the second is true but misleading. `opentelemetry_ecto` is a `:prod` dependency (`apps/fountain/mix.exs:142`), and `application.ex:19` attaches it. For every query it runs `fetch_ctx(self())`, and if that is undefined it calls `fetch_parent_ctx(1, :"$callers")` (`deps/opentelemetry_ecto/lib/opentelemetry_ecto.ex:116-119`). The owner never sets its own context, so it always takes the fallback:
   - **Before this PR** (gate on, chain empty), every Ecto span from the owner was a root span.
   - **After it**, each one is a child of whatever span the caller has open. To find that span, `process_info(caller, :dictionary)` copies the caller's whole process dictionary on every query.

   This is probably better tracing. It is still a production change that nothing guards to tests. Please correct the comment and the PR body, or guard the adoption with a compile-time test flag. I could not drive this, because the OpenTelemetry dependencies are not compiled in `test`.

   Sentry's handler also reads the chain (`deps/sentry/lib/sentry/logger_utils.ex:61`), but only for `:sentry` logger metadata. Nothing in `lib/` sets that, so today it changes nothing.
2. **One doc is stale.** `conversations.ex:3065` still says that a write retiring a fenced reset "still goes through (`update_sandbox/2`)". This PR deleted that function from `lib/`.
3. **"Finished within five minutes" does not hold after a pod dies mid-run.** An orphaned `executing` teardown job takes the next cron insert as a duplicate. Probe output:
   ```
   PROBE oban orphan: t3=56 conflict?=true (same as t1? true)
   ```
   The five-minute run therefore stops until Lifeline rescues the job after 30 minutes (`config/config.exs:52`), about 35 minutes in the worst case. It does not wedge for good. `SandboxResetReconciler` had the same shape, so this is not a regression, but the PR body's five-minute claim should say so.
4. **`refused` counts ordinary contention.** Two teardown runs driven at once returned `{0, 1}` and `{17, 0}`. The one refusal was the reset door answering `:sandbox_unavailable` after the other run claimed the lease between scan and call. Oban uniqueness keeps two runs from overlapping in normal operation. A user's own reset or retry racing the sweep will still tick `fountain_reaper_teardowns_refused`, a new gauge that is supposed to mean "machines still standing".

### Held up under attack

- **No leak across callers.** Two callers ran back to back through one owner. The second caller's chain replaced the first's completely:
  ```
  PROBE owner=#PID<0.520.0> after A (#PID<0.519.0>): [#PID<0.519.0>, #PID<0.518.0>, #PID<0.511.0>]
  PROBE owner=#PID<0.520.0> after B (#PID<0.521.0>): [#PID<0.521.0>, #PID<0.511.0>]
  ```
  `Process.put` overwrites on every call, and a caller on another node gets `Process.delete`. The stale chain the owner keeps between calls is read only by `handle_cast/2` and `handle_info/2`, and they only log. Removing `adopt_callers(caller)` fails 27 tests across the machines and reaper suites.
- **The gate is gone.** `git grep` finds no `Machines.enabled?` and no config read, only historical comments. Each deleted inline arm called the same protocol function the owner now calls: `Destroy.run`, `Park.run`, `Resume.run`, `Admission.run`, `Binding.attach` and `Binding.detach`.
  - `provision`, `confirm_up`, `fail_provision`, `end_turn`, `retarget`, `bind_inference` and a `detach` with `policy: :keep` never depended on the gate.
  - Suites that drive every verb through real call sites passed on a fresh database:

    | Suites | Tests | Failures |
    |---|---|---|
    | `machines/` plus the reaper and reset suites | 669 | 0 |
    | `conversations/` plus `workers/` | 1821 | 0 |
- **A reset landing in the middle of a park keeps its fence.** I expected a park's stamp or finalize to erase a reset's `destroying` stamp now that the columns are unread. I paused a park inside `suspend` through the owner, then committed a real `reset_sandbox/2`:
  ```
  PROBE mid-park row: status=ready transition="parking"
  PROBE after reset fence: transition="destroying" reason="reset"
  PROBE park={:ok, :parked} reset={:ok, %Sandbox{status: "terminated", ...}}
  PROBE audit: ["sandbox.reset", "sandbox.suspended", "sandbox.reset_requested"]
  ```
  `Lease.cas_update/4`'s `preserve_destroying` (`lease.ex:523`, `:750`) keeps the stamp through the park's writes.
- **`Park.admissible/2` never sees a `destroying` row.** It has exactly two callers:
  - `under_lease/3`'s last clause, where `transition` is nil, because `park.ex:364` takes every other non-nil stamp first.
  - `take_over/3`, where `transition` is `"parking"`.

  `transition` is one column, so neither caller can hand it a `destroying` row. A stamp that lands after the read survives, as the previous probe shows. Deleting the clause is correct.
- **Two teardown runs at once, each on its own real connection** (`unboxed_run`, manual pool). They made 17 provider destroys, all distinct, and every row ended `terminated` with no stamp. The events came once per row: `sandbox.destroyed` 2, `sandbox.teardown_reconciled` 2, `sandbox.reset` 1.
- **The hourly run and the five-minute run at once**, covering an abandoned ephemeral teardown, an abandoned forced teardown of a persistent home, and an abandoned reset. Both runs returned `:ok`. There were 3 destroys, all distinct, and all three rows ended terminal. The two runs cannot contend for one row at the scan either: `release_stuck_sandboxes/0` and the idle sweep both filter out `destroying` rows.
- **Oban uniqueness works.** The hourly and teardown jobs are unique apart (`h1=55 t1=56`). A repeat of either dedupes onto the one already queued (`t2=56 conflict?=true`, `h2=55 conflict?=true`). Lifeline is configured, and `max_attempts: 3` means a rescued orphan is discarded rather than retried for ever.
- **Re-plants, each placed by distinctive string and restored with `cp` and `touch`, with recompilation confirmed:**

  | Plant | Fails |
  |---|---|
  | The hourly run also calls `sweep_fenced_teardowns()` | "the hourly run drives no teardowns; the five-minute run does" |
  | Resets put behind the grace window | the reaper's reset-retry test, and `DestroyResetTest`'s "the same row is retried once the lease has expired" |
  | The `Lease.live?` check dropped from the scan | both "…owner still holds…" tests |
- **The hourly maximum is 30.** `pass_two_budget/1` is `max(5, 25 - spent)`, and `Provision.fail/2` makes no provider call, so the stuck-row pass adds nothing.
- **Mechanical checks pass.** The diff adds no advisory-lock acquisition, so lock order is unchanged. No sweep write gained a `{:ok, _} =` match. `HomeCheckpoint` runs under the park's lease, not inside a transaction.

**Verdict: approve.** Please fix the `$callers` comment and PR body claim in item 1 before this merges, because the rollback is gone once it does.

BinaryBourbon (approved):
## Review of `eb8fd083`: tests, ratchet, deletions and verification evidence

**Verdict: APPROVE.** I found no blocking defects. I planted 24 production changes of my own choosing, 13 of them rows from the PR's own table. Every plant of a production behaviour fails at least one test named for what it breaks, with one exception: the reset carve-out from the live-server check (non-blocking item 4). I restored each plant with shell `cp` then `touch` and checked it with `cmp`, and every one recompiled (`Compiling N files`). Afterwards the head passed, and the working tree was clean:

```
test/fountain/machines test/fountain/workers + the 6 reset/race files: 836 tests, 0 failures
```

### Blocking

None.

### Non-blocking

1. **Two race tests were loosened more than they needed to be, and one arm stopped catching a real defect.** `reset_admission_order_test.exs:60` and `termination_attach_order_test.exs:60` now accept `winner_pid in [winner.pid, Machine.whereis(id)]`. I planted a reset fence with no advisory lock and no `FOR UPDATE` (`do_reset_sandbox/2`). With that plant the **reset-first arm still passes**, because its pause lands in the owner's `Lease.claim` instead of the fence, and the loosened assertion accepts that. On `main` this arm would have failed. Only the admission-first arm caught the plant:
   ```
   ##### PROBE: reset fence takes no advisory or row lock
     1) test admission wins against admission on independent connections
   2 tests, 1 failure
   ```
   The loosening is not needed on this head. The tightened `assert winner_pid == winner.pid` passes on the unplanted head (`2 tests, 0 failures`). With the same plant, both arms then fail (`2 tests, 2 failures`). In termination_attach_order, a per-arm assertion also passes on the head (`3 tests, 0 failures`): `winner_pid == if(first == :termination, do: winner.pid, else: Machine.whereis(sandbox.id))`. I suggest restoring the exact pin in both files.

2. **`sandbox_reset_test.exs:312`, "a server that gives up can still mark the machine failed", now drives only the Factory fixture.** The fixture deliberately has no guards, so the write cannot be refused, and no production path matches the scenario in the title. Provision's `failed` writes (`write_failed/5`, `retire/4`) only take `pending` and `starting` rows. Nothing writes `failed` over a fenced `ready` row. What is left is a Quotas assertion under a misleading name. I suggest deleting it or pointing it at a real retiring path. The two tests beside it already cover the reap and agent deletion.

3. **The #1969 retirement race runs against a shape production does not run.** `retirement_admission_order_test.exs:49` wraps `Lease.cas_update/4` in `Repo.transaction` with `nest: true`, so its row lock is held across the pause. Production's finalize (`destroy.ex:755`) calls `cas_update` with `nest: false`, as a single autocommit statement. The test is still a real test of admission's `FOR SHARE`. Removing that lock fails both arms (`2 tests, 2 failures`). The PR body's sentence "the race now runs against the owner's finalize" overstates it, though. It races a hand-built transaction around the finalize's write.

4. **An unpinned claim.** The PR body says resets keep "no server check". I made resets wait on `Lifecycle.any_server_alive?/1` too, by dropping `not reset?(sandbox) and` at `sandbox_reaper.ex:884`, and nothing failed:
   ```
   ##### PROBE: resets wait on the live-server check too
   102 tests, 0 failures
   ```
   With that change, an abandoned reset on a home whose servers are still registered would not be retried until they die. A one-line test would pin it.

5. **Stale references to deleted functions.**
   - `conversations.ex:3065`: `reset_sandbox/2`'s `@doc` still says "A write that retires the row still goes through (`update_sandbox/2`)". That function is gone. The operator's way out is `Termination.reap_sandbox/2` through the owner.
   - `conversations.ex:3341`: a comment says "the function is `do_update_sandbox/2` now", but no such function exists.

6. **A vacuous `reject` whose raise is rescued, but other layers cover it.** In `destroy_reset_test.exs:430`, "a row whose owner holds a live lease is not retried", the `reject(Sprites, :destroy, 1)` raise would be rescued into "provider destroy failed", and the row would stay `ready`. So the first half, "the run: nothing is attempted", does not check what it says. My plant of a driver that ignores the lease passed this test. The door's `Lease.live?` and the lease claim refuse first, and the reaper suite's lease tests catch the driver plant. So this is test hygiene, not a hole.

7. **A limit of the ratchet, not a regression.** The scan does not see raw SQL (`Repo.query!("UPDATE sandboxes ...")`). I grepped `apps/*/lib` and `ee/lib` for `update|insert into|delete from sandboxes` and found no hits.

### Held up under attack

**Ratchet 6/3 → 0/2.** I measured it with my own grep on both trees, independent of the test's scanner.
- On `main`, the three provider calls were `home_checkpoint.ex:101`, `provisioning.ex:129` and `sandbox_reaper.ex:1137`.
- On the head, the only two are `provisioning.ex:129`, the env warm-start checkpoint, and `sandbox_reaper.ex:1061`, the pass-2 destroy of terminal rows. There is no third.
- Every head `update_all` outside `machines/` targets `Conversation` or `Request`.
- No `update_sandbox(` call or string-dispatched call (`apply`, `:update_sandbox`) remains in `lib/`.

I planted three writes in `termination.ex` and each one fails the ratchet:

| Plant | Result |
|---|---|
| `Repo.update_all(from(s in Sandbox ...))` | both scan tests fail |
| `Sandbox.changeset \|> Repo.update()` | both scan tests fail |
| `Managoat.Sandbox.suspend(handle)` | "only go down" fails |

**Reset notice (trust the cast).**
- Putting back `reset_requested_at` fails 3 tests, including the new pin: "the reset cast alone brings the conversation back to idle on a terminated machine", "a matching reset tells the old server…" and "a rebound holder wins…". I planted it as a SQL `EXISTS` on the column, because the field left the schema.
- Dropping `s.status == "terminated"` fails "a reset notification cannot retire an unfenced live actor".

**The reconciler's deleted tests (57 assertions), under rule 16.** Each one maps to a reaper test, and I drove the mapping with plants:

| Plant | Fails |
|---|---|
| no credentials check | "a reset on a provider with no credentials waits" |
| resets behind the grace window | "a reset the provider never confirmed is retried at once" and destroy_reset "retried once the lease has expired" |
| resets finished as teardowns | 2 reaper tests, via the audit actor and the credential skip |
| the driver ignores a live lease | 2 reaper lease tests |
| cron entry deleted | "the teardown run has a five-minute cron entry of its own" |

The per-sandbox job's uniqueness test has no equivalent, because that job no longer exists.

**The deleted `update_sandbox/2` guards, now in `Lease.cas_update/4`.**
- Removing `refuse_revival` fails 4 tests, including "cas_update does not revive a retired machine".
- Removing `stamp_terminated_at` fails 9 tests across provision, destroy, the reaper and the lease.

**The 300 repaired tests.** I reverted production code under six repaired race tests. None had been rewritten until it went green:

| Plant | Fails |
|---|---|
| admission without `FOR SHARE` | retirement race, both arms |
| attach without `FOR NO KEY UPDATE` | termination/attach "attachment wins at row lock" |
| reset retry ignores a live lease | reset concurrency, 2 tests |
| a moved channel binding no longer refuses a rotation | execution_allowance rotation race, both arms, including the attach that runs in the owner |
| reset fence lockless | see item 1 |
| the stuck pass counts `:claimed_elsewhere` as released | "a row whose owner still holds its lease is left alone" |

The race tests still race. Each one asserts `pg_blocking_pids` between the two backends, and a pure mailbox queue could not satisfy that.

**Other stamp readers and the owner seam.**

| Plant | Fails |
|---|---|
| Quotas blind to the stamp | 3 tests |
| `attachable/5` blind to the stamp | 4 tests |
| `Lease` `refuse_fenced` blind | 2 tests |
| teardown fence treats a reset stamp as a repeat | 2 escalation tests |
| the owner stops adopting `$callers` | at least 7 tests in sandbox_reset_test alone |

**The 48 files moved to the start stub, and a sample of 10+ diffs.** I read every changed line in all 48 files, not only a sample. Outside the helper swap, the changes are either column-to-stamp rewrites of the same assertion, or `Conversations.update_sandbox` becoming the fixture. `reject_server_start` users also assert unchanged row counts, so a rescued raise could not hide a start. In machine_test the lost assertions are all gate cases. In conversations_context_test and sandbox_test they are the deleted function's own tests, now covered by the lease plants above.

**The Factory fixture's 144 call sites and impossible worlds.** I instrumented the fixture to log any write production would refuse: a revival of a terminal row, or a live-status write over a `destroying` row. I ran all 34 core files that call it, plus ee's usage metering and termination billing. None logged a line. A staged revival confirmed the probe fires:

```
REVIVAL terminated->ready at test/zz_probe_sanity_test.exs:6
```

The one fixture use that asserts behaviour production cannot produce is item 2.

**`Park.admissible/2`'s deleted clause is really unreachable.** I drove both real fences, not a forged stamp, over an abandoned `parking` row:

```
RESET_RESULT {:error, :sandbox_reset_pending} transition="destroying" reason="reset"
PARK {:error, :fenced} status=ready
AFTER_TEARDOWN_FENCE transition="destroying" reason="admin_reap"
PARK {:error, :fenced} status=ready
```

`stamp_intent!/2` overwrites whatever stamp is there, so a fenced row never reaches `take_over/3`'s `admissible/2` wearing `parking`. The ordinary path reaches `admissible/2` only when no stamp is present. The window between a park's read and its `parking` stamp predates this PR, and the deleted clause read the same snapshot, so it would not have closed that window either.


### Addendum: the stubs that send to `self()` (the implementer's own trap), hunted across the whole suite

**No vacuous instance survives on this head. The verdict stands.**

**The trap.** Every verb now runs in the owner, so a stub written as `fn … -> send(self(), msg) end` delivers `msg` to the owner instead of the test. Any `refute_received msg` then passes whether the code is right or not. I confirmed the shape with a scratch test on this head: the provider really is called, the row ends `terminated`, and the refute still passes. The same call with a captured `test = self()` fails as it should:

```
VACUOUS_REFUTE_PASSED status=terminated
  1) test CAPTURED SHAPE: same call, refute fails
     Unexpectedly received message :provider_destroy_called
OWNER_GOT :provider_destroy_called
```

**The static sweep.**
- `grep -rn 'send(self()'` over every test tree finds 11 sites. None is a stub reached through a machine verb with a refute on its message.
  - The two in `deletion_fence_test.exs` and the one in `conversation_controller_test.exs` are `assert_received`, which would fail loudly, and they pass.
  - The ones in `runner_recovery_test.exs` and `provisioning_test.exs` feed the calling process's own mailbox on purpose.
  - The rest are not stubs.
- This PR converted exactly one site, the reaper suite's `{:destroyed, name}`.
- No stub body binds `self()` to a name and then sends to it.
- The `whereis → self()` stubs are liveness answers, and an owner pid is alive too.
- The `Process.get` telemetry pauses in binding and admission tests run on the caller, and a missed pause fails loudly on `assert_receive :locked`.

**The dynamic sweep.** Grep cannot see a refute written as a `receive … after 0` helper, for example `destroyed_names() == []`. So I made both catch-alls in `Machine`, `handle_info/2` and `handle_cast/2`, record every message they receive to a file. Then I ran the whole core suite, which includes ee:

```
6 doctests, 2 properties, 6812 tests, 1 failure, 2 skipped
```

The one failure is `ConfigReferenceTest`, which my instrumentation caused by reading a new env var in `lib/`. It passes once restored.

Across all 6812 tests, only three things reached an owner:
- `binding_test.exs`'s deliberate secret-not-logged message.
- The cast twin of that message.
- One `:DOWN` for a caller that had exited normally, with no stub payload.

No stub message reached an owner anywhere in the suite. `Renewal.around/5` runs the provider call in the owner itself, so there is no intermediate process where a send could vanish unrecorded. None of the five extension apps' refuting tests touch a machine verb.

**Refutes whose sender runs in the owner fail when their behaviour is planted.** Each plant was restored with `cp`, `touch` and `cmp`:

| Plant | Fails |
|---|---|
| the teardown driver ignores a live server | "a fenced row a server still holds is left alone" (`destroyed_names() == []`) |
| the teardown driver takes unstamped rows | "a machine nobody has asked to destroy is left alone" |
| cotenants told twice | 2 wake_cotenants tests and binding "each group its own notice" (`refute_receive … once each`) |
| cotenants told with no notice | "with no notice, nobody is told" |

jhgaylor:
Reply to the [surfaces review](https://github.com/managoat/fountain/pull/2423#pullrequestreview-5245444019), all addressed at `e3b2f806c`.

**S1: the docs.** `docs/concepts/sandboxes.md` and the sandbox-lifetime guide now describe the one five-minute driver. The per-machine jobs, backoff, the hourly finish, `reconciled=` on the hourly line and the driver floor's share are gone. The guide shows both log lines and says what happens to each kind of row:
- A reset is retried on every five-minute run until the provider confirms, and records `sandbox.reset`.
- A deletion is finished on the first run after it is fifteen minutes old, with no server holding the machine.

The changelog fragment and the PR body carried the same "within five minutes" error, and both are fixed.

**S2: the shim.** `SandboxResetReconciler` stays as a no-op `Oban.Worker` that answers `:ok`, with `max_attempts: 1` and no cron entry. Its moduledoc says why, and that 9b-ii deletes it. `sandbox_reset_reconciler_test.exs` checks three things:
- Both job shapes the previous release enqueues complete and touch nothing. The per-sandbox shape runs on a stamped reset home.
- The crontab never names it.
- A shim that fails, planted, fails both job tests.

The shim is on the body's new 9b-ii checklist.

**Non-blocking.** The five-minute run's canary is named as a residual. It is a home-cloud alert on staleness of `fountain_reaper_teardowns_*`, for the lead's dashboards PR. The ADR Rollout text and the `MachineEvents.reset/6` plan are on the 9b-ii checklist.

jhgaylor:
Reply to the [protocol review](https://github.com/managoat/fountain/pull/2423#pullrequestreview-5245483603), all addressed at `e3b2f806c`.

**P1: `$callers` is test-only now (Jake's decision).** `Machine.handle_call/3` adopts only when `Application.get_env(:fountain, :owner_adopts_callers, false)` is true, and only `config/test.exs` sets it. With it off, the owner never puts the key at all, which is `main`'s owner. The comment at the adoption now says three things:
- It exists so a test's SQL sandbox connection and Mimic stubs reach the owner.
- It is test-only because `opentelemetry_ecto` walks `$callers` on every query and would copy the caller's whole dictionary and re-parent the owner's spans.
- Better owner tracing would be a separate, deliberate change.

Two new tests in `machine_test.exs` cover it. With the flag off, the owner has no `:"$callers"` after a call. With the flag on, the first caller is the test. The flag-off test fails with the guard removed, planted and restored with `cp`, `touch` and `cmp`. All tests pass with the flag on: 6820 core, 0 failures. The body's claim that nothing in production reads `$callers` is corrected.

**P2.** The `reset_sandbox` doc now says a destroy through the owner still retires the row, continuing from the stamp. Other present-tense mentions of the deleted door are corrected too: in the drainer, `Fountain.Billing`, `SandboxUsage`, `Lease` and the reset comment. Historical "would have run" references stay.

**P3.** Named as a residual in the body. The deleted reconciler had the same shape.

**P4: contention.** On `sandbox_unavailable`, the driver re-reads the row. A live lease, or a row that has gone terminal, is another run's or operation's reclamation and counts on neither gauge. Anything else is still `refused`. There are two tests: a lease lost to another run gives `{0, 0}`, and `sandbox_unavailable` with nobody holding the machine gives `{0, 1}`. Each fails its plant.

jhgaylor:
Reply to the [behaviour review](https://github.com/managoat/fountain/pull/2423#pullrequestreview-5245520828), all addressed at `e3b2f806c`. Thank you for the dynamic `refute_received` hunt.

**B1: exact winner checks.** You were right that I weakened these. In `reset_admission_order_test.exs`, the check is `winner_pid == winner.pid` again, exactly as on `main`. Both the reset fence and the locked turn insert run on the calling process, so loosening it was never needed. I also reverted the pause extension I had added there. In `termination_attach_order_test.exs`, the check names the exact process per role: the task for a termination, and the owner for an attachment, which runs its insert there.

For verification I planted the fence's advisory lock on a different key. Keeping the statement keeps the pause, so the plant tests the mutual exclusion rather than a missing pause:
- The reset plant fails both reset/admission arms.
- The teardown plant fails the machine-lock arm of termination/attach.
- The two row-lock arms of termination/attach still serialize on the row lock, which that plant leaves in place.

**B2.** The test drives the production path now, a destroy through the owner that continues from the reset's stamp. It is retitled "can still retire the machine through its owner".

**B3.** The test's comment and the body now say the difference: production's finalize is one statement outside any transaction, and the test holds it in a transaction so the retirement-first arm has a lock to hold. The admission-first arm is production's shape.

**B4.** Pinned: a reset with a live server on the home is still retried. Giving resets the live-server check fails it.

**B5.** Fixed with P2, including the `do_update_sandbox/2` comment.

**B6.** The rescued `reject` became two reporting stubs: the reset's door and the provider. The test refutes both. The driver-ignores-the-lease plant now fails it.

**B7.** Stated in the body's residuals. The ratchet's moduledoc already names `Repo.query!`, `Ecto.Multi` and raw `execute`.

BinaryBourbon (approved):
## Protocol review, round 2 (`e3b2f806c`)

I re-checked my round 1 findings on the delta from `eb8fd0832`, in a fresh worktree and database.

### Blocking

None.

### Non-blocking

1. **No PR test pins the reset branch of the new contention handling.** The only `refused` I saw in round 1 came from the reset branch, where the reset door answers `:sandbox_unavailable`. The three new tests stub `Termination._unsafe_destroy_machine`, so they reach only the teardown branch. When I deleted the reset branch's `{:error, :sandbox_unavailable} = refusal -> contended_or_refused(...)` clause, `sandbox_reaper_test.exs` and `destroy_reset_test.exs` stayed green (76 tests, 0 failures from the PR's own tests). Only my probe caught it:
   ```
   PROBE round 1: runs={3, 0} {5, 1} destroys=8 unique=8 ...
   ```
   Please add the reset twin of "a destroy another run is finishing counts on neither gauge": stub `retry_pending_sandbox_reset/2` to claim the lease and answer `{:error, :sandbox_unavailable}`, then assert `{0, 0}`.
2. **Three comments still describe deleted functions in the present tense.** The one I raised at `conversations.ex:3065` is fixed. These remain:
   - `lease.ex:680` says the code "Mirrors `Fountain.Conversations.prevent_sandbox_revival/1`", but that function no longer exists.
   - `lease.ex:691-692` says it "Mirrors `Fountain.Conversations.stamp_terminated_at/1`, the second of `Conversations.update_sandbox/2`'s two guards". Neither function exists.
   - `destroy.ex:767` says "the same rule `Conversations.update_sandbox/2` follows".
3. **One phrase says the opposite of what it means.** `machine.ex:1214` reads "Off in production, and it must stay off by accident." It presumably means "it must not be turned on by accident."

### Held up under attack

- **Flag off, the owner never puts `$callers`.** I drove a real park through the owner from a `Task` whose own chain was set, then made a second call:
  ```
  PROBE task's own $callers: [#PID<0.510.0>]
  PROBE park={:ok, :parked} owner=#PID<0.518.0> $callers=:error
  PROBE after second call, same owner? true $callers=:error
  ```
  With the guard removed, the same probe shows `$callers={:ok, [#PID<0.1848.0>, #PID<0.1841.0>]}`. The PR's "with the flag off, as in production" test fails too. That test is not vacuous: if the owner crashed, `whereis` would return nil and fail the `is_pid` assertion.
- **Production does not set the flag.** `git grep owner_adopts_callers` finds only `config/test.exs:46`, the reader and the test. It is absent from `runtime.exs` and `prod.exs`, and the default is `false`.
- **The rewritten comment at `machine.ex:1207` is accurate.** It names `opentelemetry_ecto`, the per-query copy of the caller's process dictionary and the span re-parenting. It explains why the check is a runtime read.
- **The suites that needed adoption still pass with the flag on:** `machines/`, `conversations/` and `workers/` ran 2434 tests with 0 failures on a fresh DB.
- **Healthy concurrent runs no longer count as `refused`.** I repeated round 1's probe five times: two `sweep_fenced_teardowns/0` runs at once on separate real connections, each round with 4 abandoned teardowns and 4 abandoned resets.

  | Round | Runs | Destroys, all distinct | Rows |
  |---|---|---|---|
  | 1 | `{2, 0}` `{6, 0}` | 8 | 8 `terminated`, stamp cleared |
  | 2 | `{6, 0}` `{2, 0}` | 8 | same |
  | 3 | `{2, 0}` `{6, 0}` | 8 | same |
  | 4 | `{6, 0}` `{2, 0}` | 8 | same |
  | 5 | `{2, 0}` `{6, 0}` | 8 | same |

  Round 1 of this review, before the fix, gave `{0, 1}` and `{17, 0}`.
- **The teardown branch is pinned.** Removing its `contended_or_refused` clause fails "a destroy another run is finishing counts on neither gauge". The symmetric test, "`sandbox_unavailable` with nobody holding the machine is still a refusal", keeps real refusals counted.
- **The no-op `SandboxResetReconciler` is safe.** It has `max_attempts: 1`, answers `:ok`, and has no cron entry. A job the previous release enqueues during the rolling deploy completes instead of being discarded and paging.

**Verdict: approve.** Item 1 is a missing test for code I have driven and seen correct, so it need not block the merge.

BinaryBourbon (approved):
## Surfaces review, round 2 (`e3b2f806c`)

I reviewed the delta over `eb8fd0832` in a fresh worktree and test database. Every result below comes from a probe.

**Verdict: approve.** Both blockers from round 1 are fixed, and the fixes hold when driven.

### Blocking

None.

### Non-blocking

1. **One stale line in the PR body.** Under the rolling-deploy notes it still says old-release reconciler jobs "fail and are discarded". With the shim they complete, as shown below. Worth fixing before the squash message is written.
2. **What `refused` means for a teardown.** `refused` on the five-minute line counts a teardown the owner would not take. It does not count a provider failure. `Destroy.destroy_at_provider/2` retires the row on a provider error, and the hourly pass for terminal rows deletes the leftover machine. I drove it: a teardown whose provider call failed reported `{reconciled 1, refused 0}` and left the row `terminated`. This predates the PR, and the guide's "The machine is deleted at the provider on that run" was already there. A clause such as "or, if the provider does not answer, by the hourly run" would make that sentence exact.

### S1, the docs: every sentence checked against the code

- **`docs/concepts/sandboxes.md:99-105`**
  - A reset is tried again every run until confirmed. True: resets skip the grace window and the server check.
  - A provider without credentials waits. True: `finish_reset` returns `:noop` when the provider has no credentials.
  - A deletion is finished at 15 minutes with no server holding the machine. True: `@fenced_teardown_grace_minutes 15` and `any_server_alive?`.
  - A machine another operation holds is left alone. True: the live-lease filter and the new re-read.
  - The fence and reservation stay until confirmed. True.
- **`sandbox-lifetime.md`**
  - Both example log lines match the code. The hourly line has no `reconciled` (`sandbox_reaper.ex:217-220`). The five-minute line is `reaper: teardowns reconciled=N refused=N` (`:187`).
  - A finished reset records `sandbox.reset`. A finished teardown records `sandbox.destroyed` plus `sandbox.teardown_reconciled`, which I drove: `teardown_requested(self) · destroyed(system:sandbox_reaper) · teardown_reconciled(system:sandbox_reaper)`.
  - Each run asks about a fixed number of machines and deletes them one at a time. True: `@owner_attempts_per_sweep 100` and a sequential `Enum.reduce`.
- **Changelog.** "on the first five-minute run after it is fifteen minutes old" is now correct.
- **Telemetry description.** `fountain.reaper.teardowns.refused` says "could not finish this run; the machines are still up". That still matches, given the new rule for contended machines.

### S2, the no-op shim, driven through the real executor

I inserted both job shapes exactly as main's module did, with `max_attempts: 10` and `unique: [period: :infinity, states: :incomplete]`, and ran `Oban.drain_queue` (not `perform_job`). I also attached a handler to `[:oban, :job, :exception]` and `:stop`:
```
PROBE drain: %{discard: 0, cancelled: 0, success: 2, failure: 0, snoozed: 0}
PROBE rows: [{%{}, "completed", 1, 10, []},
             {%{"sandbox_id" => "d1df…"}, "completed", 1, 10, []}]
PROBE home untouched: %{status: "ready", transition: "destroying"}
PROBE event: {:stop, %{}, :success}
PROBE event: {:stop, %{"sandbox_id" => …}, :success}
PROBE no extra event          # no :exception
PROBE crons: [{"7 * * * *", SandboxReaper}, {"*/5 * * * *", SandboxReaper, [args: %{"pass" => "teardowns"}]}]
```
Both jobs complete on the first attempt with no errors. There is no exception event, so `FountainObanJobsRaising` does not fire, and nothing is discarded, so `FountainObanJobsDiscarded` does not fire. The old row keeps its `max_attempts: 10`, but that no longer matters because `perform/1` returns `:ok`. The new release's crontab has no entry for the worker, and nothing in `lib/` or `config/` inserts it.

### P4, contention no longer counts as `refused`

- **Two real concurrent five-minute runs.** I started two `SandboxReaper.perform(%{"pass" => "teardowns"})` runs 500 ms apart, with a provider destroy that takes 8 s. The gauges read `%{refused: 0, reconciled: 0}` and `%{refused: 0, reconciled: 1}`. There was one provider destroy, and the row ended `terminated`. The second run queued in the owner's mailbox and found the row terminal.
- **The cross-node race, on the real path with no stubbed destroy.** I claimed a foreign lease (`other@node`) between the scan and the owner call. The result was `{{0, 0}, 4819 ms}`: the owner waited about 5 s, answered `sandbox_unavailable`, and `contended_or_refused/2` re-read the row, found a live lease and counted nothing. The row was left `destroying` for `other@node`.
- **Revert probe.** I planted `:refused` in place of `:noop` on the live-lease branch. The real-path probe then read `{0, 1}`, and `a destroy another run is finishing counts on neither gauge` failed, with 62 tests and 1 failure. I restored the file with `cp` and `touch`, saw "Compiling 1 file", and the suite passed again with 60 tests and 0 failures.

### Checks

- `sandbox_reset_reconciler_test.exs` with `sandbox_reaper_test.exs`: 63 tests, 0 failures.
- `scripts/changelog.py check`: ok.
- `scripts/test-docs.sh`: 0 failures.
- `okf validate decisions`: 0 errors. `decisions-index.sh` leaves no diff.
- `build.sh --check`: `ok (173 operations, 195 schemas)`.
- CI on `e3b2f806c`: 25 success, 3 skipped, none pending or failed.

BinaryBourbon (approved):
## Round 2 at `e3b2f806c`: behaviour and tests

**Verdict: APPROVE.** B1 is fixed, and its argument holds, with one refinement described below. B2, B3, B4 and B6 are taken. I planted 14 changes by distinctive string, including all 8 of round 1's new plants. I restored each plant with `cp`, `touch` and `cmp`, and truncated the test DB before every run. The race tests commit through `unboxed_run`, and a failing plant can leave rows behind that make the next run fail on `users_email_index`. The first attach-side run I made hit exactly that, so I discarded it and re-ran everything from a clean DB. The round's eight changed test files pass on a clean tree afterwards:

```
132 tests, 0 failures
```

### Blocking

None.

### B1: the race tests are exact again, and the row-lock argument holds, with a refinement

- `reset_admission_order_test.exs` is back to `winner_pid == winner.pid`, and the `$callers` extension to its pause is reverted.
- `termination_attach_order_test.exs` names the task for a termination and the owner for an attachment.
- The implementer's plant holds: moving the reset fence's advisory lock to another key fails both reset/admission arms:
  ```
  1) admission wins …   left: {:ok, turn}  right: {:error, :sandbox_unavailable}
  2) reset wins …       no wait on the winning PostgreSQL connection observed
  2 tests, 2 failures
  ```

**Is "the row-lock arms serialize on the row lock" true?** Only once the advisory lock is gone. Both sides take the advisory lock before the row lock:
- the teardown fence (`Lifecycle.do_fence_sandbox_for_teardown/2`) takes the advisory lock, then the conversation, then `FOR UPDATE`;
- the attach (`Binding.locked_attach/3`) takes the source lock, then the advisory lock, then `FOR NO KEY UPDATE`.

So with both locks intact, the second side blocks on the advisory lock first, and the row lock never decides anything. I ran six plants on termination/attach, three per side. On the attach side I replaced the lock with `lock: "/* FOR NO KEY UPDATE */"`, so the test's pause still matches and only the lock disappears:

| Plant | machine-lock arm | termination-first row arm | attachment-first row arm |
|---|---|---|---|
| fence advisory lock on another key (the implementer's plant) | fails | passes | passes |
| fence row read without a lock | passes | passes | passes |
| fence: both | fails | fails | fails |
| attach advisory lock on another key | fails | passes | fails |
| attach row read without a lock | passes | passes | passes |
| attach: both | fails | fails | fails |

What this shows:
- The implementer's statement is right for the plant it names. With the fence's advisory lock moved, the row arms do serialize on the row lock.
- A lockless row read does fail both row arms, but only together with the advisory plant.
- On its own, a lockless row read on either side passes this file.

I then ran each lockless plant against 26 files, all of `machines/` plus every order, fence, isolation, race and concurrency file under `conversations/`:

```
F2 wide: fence row read lockless    751 tests, 0 failures
A2 wide: attach row read lockless   751 tests, 0 failures
```

**Non-blocking, and older than this PR.** No test pins the fence's `FOR UPDATE` or the attach's `FOR NO KEY UPDATE` on its own. What those locks guard is a writer that takes no advisory lock, the owner's `Lease.cas_update/4`. `main` has the same lock order and this PR does not touch those lines, so this does not block. A race of the fence and of the attach against a `cas_update` would pin them, in the shape the retirement test already uses.

### The other items

| Item | Plant | Result |
|---|---|---|
| B4 | resets given the live-server check | "a reset is retried with a server live on the home…" fails (`{1, 0}` vs `{0, 0}`) |
| B6 | the teardown run skips its lease check | "a row whose owner holds a live lease is not retried" now fails. At `eb8fd083` the same plant passed |
| B2 | the owner's destroy refuses a reset stamp | "a server that gives up can still retire the machine through its owner" fails. The plant is broad (16 failures), since the reset uses the same path. The test now drives production's path instead of the fixture |
| B3 | none (a comment change) | The comment states production's one-statement finalize and says the admission-first arm matches production |

### Round 1's new plants, re-planted

| Plant | Fails |
|---|---|
| `:owner_adopts_callers` guard removed | "with the flag off, as in production, the owner never puts $callers" |
| reset fence's advisory lock on another key | both reset/admission arms |
| teardown fence's advisory lock on another key | machine-lock arm (first table above) |
| resets given the live-server check | B4 test |
| contention counted as refused | "a destroy another run is finishing counts on neither gauge" (`{0, 0}` vs `{0, 1}`) |
| the symmetric case: a real `:sandbox_unavailable` counted as noop | "sandbox_unavailable with nobody holding the machine is still a refusal" |
| the reconciler shim returns an error | both "…from the previous release completes and does nothing" tests |
| the teardown run skips its lease check | B6 test |

### Non-blocking

- **A stale sentence in the PR body.** The "Mixed-version rollout" section still says "Those jobs fail and are discarded" of an old replica's `SandboxResetReconciler` jobs. The shim added this round (S2) makes them complete. Item 2 under "What this deletes" is right, so only the rollout bullet needs the edit.

PR #2425, opened 2026-09-18
feat(machines): drop the fence columns; ADR 0058 accepted (stage 9b-ii)

jhgaylor:
Stage 9b-ii of [ADR 0058](decisions/0058-the-machine-has-one-owner.md), the last stage. It is off `main` after 9b-i (#2423), which is deployed and has been watched, and after v0.20.1's backfill (#2427). Tracker #2344.

## Release ordering (decided: v0.20.0, then v0.20.1, then this)

This PR drops two columns. Two releases had to come first:

- **v0.20.0 (#2426) stops reading them.** A replica that still selects the columns fails every sandbox query once they are gone. v0.20.0 contains 9a and 9b-i but not this PR, so a rolling upgrade can pass through it.
- **v0.20.1 (#2427) stamps what v0.19.0 left in them.** 9a first shipped in v0.20.0, so v0.19.0 wrote its fences to the columns only. The migration review found that this drop would erase the last record of such a fence. #2427 adds migration `20260918190000_backfill_destroying_from_fence_columns`, which stamps `destroying` on those rows so the reaper finishes them, except a reset on a machine that has run a turn since the request, which it skips on purpose and logs by id.

v0.20.1 is tagged and released (#2433, `ffc186ac9`), and hosted production has run the backfill. This branch sits on that release commit.

**The backfill alone does not cover a rolling upgrade, so this drop runs it again first** (round 2, K1). Pods migrate at boot under a `RollingUpdate`, so the backfill runs when the first v0.20.1 pod boots. v0.19.0 replicas still serving after that can write a column-only fence, and if its writer dies, v0.20.x cannot see it. Jake decided that the drop re-runs the backfill immediately before it drops the columns. So every column-only fence is stamped before the columns go, however the upgrade is rolled, except a reset on a machine used since the request, which is neither stamped nor kept, and is logged by id by both runs.

- **A rolling upgrade** needs every replica on v0.20.0 or later before this release migrates, because the drop breaks an older replica's queries.
- **A stop-the-world upgrade** can come from any version. The backfill guarantee holds either way.

## What this does

1. **Migration `20260918200000_drop_sandbox_fence_columns`.** In one transaction, it does three things in order:
[...]

PR #2426, opened 2026-09-18
chore(release): v0.20.0

jhgaylor:
Version bump for v0.20.0: both `mix.exs` files, the six self-host image pins, and the changelog roll (`[Unreleased]` plus every `changelog.d/` fragment into `## [0.20.0]`).

Merging this PR tags the squash commit `v0.20.0` and dispatches `release.yml`, which builds the CLI binaries, publishes `ghcr.io/managoat/fountain:v0.20.0`, boots it through the documented quick start and creates the GitHub Release.


BinaryBourbon (approved):
Release bump for v0.20.0, checked before approving: 0.19.0 → 0.20.0 in every pinned location (both `mix.exs`, compose, `.env.compose.example`, render, fly, `deploy/k8s`), and the 58 `changelog.d/` fragments compiled into `CHANGELOG.md`. Nothing else in the diff. Minor bump because the fragments include 7 `Added` sections, matching how v0.18.0 and v0.19.0 were numbered.

This release is cut deliberately **before** #2425 (ADR 0058 stage 9b-ii, the fence-column drop) merges. v0.20.0 carries 9b-i (#2423), which stops reading the columns while they still exist, so a self-hoster upgrading from v0.19.0 can roll the upgrade safely. The drop ships in the release after this. CI is green: 25 success, 3 skipped.

PR #2427, opened 2026-09-18
fix(machines): stamp the fences v0.19.0 left in the columns (ADR 0058)

jhgaylor:
For v0.20.1. Must land before #2425, which drops the two fence columns.

## Problem

v0.19.0 fences only in the columns. A reset sets `reset_requested_at`. A forced teardown sets it and `teardown_requested_at`. v0.20.0 (ADR 0058 stage 9b-i, #2423) reads neither: its `Sandbox` schema does not declare them, and every reader refuses on the `destroying` stamp that 9a (#2386) introduced. A self-hoster who upgraded from v0.19.0 with a fence abandoned or in flight has a machine v0.20.0 treats as live. It keeps billing and holds a quota slot. #2425 would then erase the only record.

Hosted production is unaffected: it ran the dual-writing 9a before 9b-i, and a production query found no such rows.

## Change

One data migration, `20260918190000`, after the newest on `main` (`20260918050000`) and before #2425's `20260918200000`. No other open PR uses a version between those. Raw SQL, because the schema no longer declares the columns.

```sql
-- first, in its own statement, inside the migration's transaction
SELECT id FROM sandboxes
 WHERE (reset_requested_at IS NOT NULL OR teardown_requested_at IS NOT NULL)
   AND status NOT IN ('terminated', 'failed')
   AND transition IS DISTINCT FROM 'destroying'
 ORDER BY id
   FOR UPDATE;

UPDATE sandboxes
   SET transition = 'destroying',
       transition_reason = CASE
         WHEN teardown_requested_at IS NOT NULL THEN 'teardown'
         WHEN mode = 'persistent' AND status IN ('ready', 'suspended') THEN 'reset'
         ELSE 'teardown'
       END
 WHERE (reset_requested_at IS NOT NULL OR teardown_requested_at IS NOT NULL)
   AND status NOT IN ('terminated', 'failed')
   AND transition IS DISTINCT FROM 'destroying'
   AND NOT (
     teardown_requested_at IS NULL
     AND EXISTS (
       SELECT 1 FROM turns t JOIN conversations c ON c.id = t.conversation_id
        WHERE c.sandbox_id = sandboxes.id
          AND t.inserted_at >= date_trunc('second', sandboxes.reset_requested_at)))
```

- **A reset on a machine used since the request is skipped** (review round 1). v0.20.0 could not see the fence, so the user could go back to the machine and run turns on it. The reset door checks only the lease, so finishing the old reset would wipe that work and interrupt a running turn. The row is left unstamped, as the user last used it, and the migration logs its id, user, provider, machine name and request time at warning level. Only the machine's owner can reset it again, with `DELETE /api/sandboxes/:id`. The console has no reset. The admin retry acts only on a row stamped `destroying`, so it skips these rows. Operators have no path of their own, so they should tell the owner.
  - "Used" means a turn inserted at or after `reset_requested_at` on a conversation whose `sandbox_id` is the row. That is the binding both v0.19.0 and v0.20.0 write. Neither version clears it when a conversation ends. The only write that moves it, the wake's co-tenant move, moves it onto a replacement machine.
  - Any turn row counts, whatever its status, so a `pending` first turn is use. A conversation bound since the request with no turn does not count. Attaching runs nothing on the machine. Once the stamp lands, turn admission refuses a `destroying` machine.
  - The skip is decided by intent (`teardown_requested_at IS NULL`), not by label. It sits in the `WHERE`, so the odd-shape branch cannot turn a used reset into a destroy.
- **The fenced rows are locked first, in a statement of their own** (review round 2). In a rolling upgrade, a v0.20.0 replica can admit a turn on one of these machines while the migration runs. Admission takes the sandbox row `FOR SHARE`, then inserts the turn, then commits. On its own, the `UPDATE` waits for that share lock, but PostgreSQL then re-checks only the updated row. The `EXISTS` keeps the statement's first snapshot, misses the turn, and stamps the reset over it. The `SELECT ... FOR UPDATE` waits out every admission already holding a row. Under READ COMMITTED, the `UPDATE` after it then takes a fresh snapshot and sees the new turns. An admission that arrives after the lock waits for the migration to commit, then reads the stamp and is refused. The lock touches only `sandboxes`, which admission also locks first, and the migration reads `turns` without locking it, so no lock cycle through `turns` exists. Nothing in the app locks several sandbox rows in one transaction, and `ORDER BY id` fixes the order anyway. The migration runs in its own transaction, because `@disable_ddl_transaction` is not set.
- **Teardowns are never skipped.** They come from account deletion, termination and home destroys. The teardown run already waits for a live server to go.
- **Teardown is tested first**, because a v0.19.0 teardown sets both columns.
- **A reset the reset door would refuse is labelled a teardown.** `retry_pending_sandbox_reset/2` acts only on a persistent `ready` or `suspended` row and answers `:skipped` otherwise, on every run, so a `reset` label there would stand for ever. Nothing v0.19.0 wrote reaches this branch. A row v0.20.0 touched can.
- **Terminal rows are left alone.** v0.19.0 never clears either column, and a completed reset writes `terminated` and keeps its column.
- **Rows already `destroying` are left alone.** A v0.20.x fence carries its own stamp and reason.
- **Another live transition is overwritten, whatever the lease says** (item 4). `transition` and the lease columns arrive in `20260916120000`, which v0.20.0 ships and v0.19.0 does not. A v0.19.0-origin row therefore reaches this with both null. A row with `parking`, `resuming` or `provisioning` was touched on v0.20.x, which could not see the fence. Stamping over it is what a v0.20.x fence does: `Destroy.stamp_intent!/2` writes `destroying` over any transition and takes no lease. `Lease.cas_update/4` keeps the stamp through a live owner's non-terminal write, and a provision's finalize refuses a stamped row. Restricting to `transition IS NULL` would leave an abandoned park holding a fence that a later reader clears. A used reset is still skipped, whatever its transition.
- **No lease and no `updated_at` are written** (item 5). The teardown run picks up a stamp with no live lease, and skips a live lease until it lapses. A reset is retried on the first run, with no grace. A teardown's 15-minute grace reads `updated_at`, which stays at the v0.19.0 fence's write. A teardown older than 15 minutes is finished on the first run, and a newer one waits out the rest of its window. A row is never skipped for ever.
- **`down` is a no-op.** The columns still exist at this version, and clearing the stamps would make those machines look live again.

A finished reset records `sandbox.reset` from `system:sandbox_reaper`, and no `sandbox.destroyed`, because the reset door silences the destroy's own event. A finished teardown records `sandbox.destroyed` with reason `teardown`, plus `sandbox.teardown_reconciled`.

## Tests

`test/fountain/sandbox_fence_backfill_migration_test.exs` seeds v0.19.0-shaped rows with raw SQL. Its `setup` re-adds the two columns with `IF NOT EXISTS`, so the file survives #2425. Seven tests:

- **Labelling.** Covers reset-only, teardown, a reset escalated to a teardown, terminated, failed, unfenced, already stamped, `parking` with a reset column, and an odd-shape reset. No other column moves, and `down` changes nothing.
- **Used since the request.** A reset with a turn since, and an odd-shape reset with a turn since, are untouched and logged. A reset with only an older turn, and a reset bound since with no turn, are stamped `reset`. A teardown with a turn since is stamped `teardown`.
- **Running turn, end to end.** A reset with a running turn since goes through the migration and `perform(pass: "teardowns")`. The machine stays `ready`, the turn stays `running`, and nothing is destroyed.
- **A turn admitted while the backfill runs.** This test uses two independent connections and committed rows in a private schema, which `on_exit` drops even if a task crashes. One connection plays admission: it takes the row `FOR SHARE`, inserts a running turn and holds. The other runs the migration. The admission commits only once `pg_blocking_pids` shows the migration waiting on it, so the order is forced, not timed. The row must end unstamped and logged. With the lock it passed 20 of 20 runs. With the lock removed it failed 10 of 10, stamping `destroying`/`reset`.
- **End to end.** The reset completes through the reset door. The teardown completes, and `sandbox.teardown_reconciled` records the stored label `teardown`.
- **A reset has no grace.** A reset fenced a minute before the upgrade is retried on the first run.
- **A teardown has grace.** A teardown fenced five minutes before is skipped, then finished once it is 20 minutes old.

Revert probes. Each defect was planted, then restored with `cp` and `touch`, and `cmp` confirmed the restore:

| Planted defect | Tests failing (of 7) |
|---|---|
| `FOR UPDATE` lock statement removed | 1: the admission race |
| teardown label written as NULL | 3, including the end-to-end test |
| teardown run holds resets to the grace window (in `SandboxReaper`) | 1: the one-minute reset |
| skip predicate removed | 3, including the running-turn test and the race |
| skip applied only to rows labelled `reset` | 1: the odd-shape assertion |
| turns before the request count as use | 3 |
| teardowns skipped too | 1 |
| reset tested before teardown | 4 |
| terminal filter removed | 1 |
| `destroying` filter removed | 2 |
| predicate narrowed to `transition IS NULL` | 1 |
| odd-shape branch answers `reset` | 1 |
| `updated_at` set to now | 2 |

## Known limits

- **A deleted conversation hides its turns.** Deleting a conversation deletes its turns. If the user worked on the machine only in a conversation they have since deleted, the machine reads as unused, and the reset is finished.
- **The reset path checks only the lease.** `retry_pending_sandbox_reset/2` makes no running-turn or execution-journal check of its own, as on v0.20.0. Here the skip is what keeps a used machine away from it.

Refs #2344 (ADR 0058 tracker), #2386, #2423, #2425.


BinaryBourbon (changes requested):
Data-correctness review at `22418398`. I ran this on a private DB, on a fresh schema and on a real v0.19.0 schema. The SQL labels every v0.19.0 shape correctly, and the ordering, idempotency and `down` all hold. One thing blocks. The residual flagged in the PR body is reachable on documented upgrade paths, and the harm is irreversible. A persistent home the user has gone back to using is wiped, mid-turn if a turn is running. The fix is one predicate, and I tested it.

## Blocking

**A backfilled reset wipes a home the user resumed on v0.20.0, including one with a turn running.** See `apps/fountain/priv/repo/migrations/20260918190000_backfill_destroying_from_fence_columns.exs:75-84`.

On v0.20.0 the fenced home looked live, so a user could attach, run turns and build new state on it. This backfill stamps it `reset`. The teardown run's reset branch then goes through `retry_pending_sandbox_reset/2` (`apps/fountain/lib/fountain/conversations.ex:3218-3248`). That branch checks only `Lease.live?`. It makes no running-turn check and no `ExecutionGuard` check, and it has no grace window. The reaper comment at `apps/fountain/lib/fountain/workers/sandbox_reaper.ex:808-809` justifies that by saying "`reset_sandbox/2` refuses a machine with a turn running". That holds for a stamp written on v0.20.x, because the stamp blocks admission from that moment. It does not hold for a backfilled stamp, because v0.20.0 admitted work after the fence. A live server is not told either: `MachineEvents.reset/6` only acts when `current_turn: nil` (`apps/fountain/lib/fountain/conversations/machine_events.ex:24`).

Probe: a scratch test in my worktree, since deleted. It seeds a v0.19.0 reset-only persistent `ready` home, a conversation on it, and a `running` turn. It then runs the migration and `perform(pass: "teardowns")`.

```
# control: the same home, reset on v0.20.x
{:error, :sandbox_mid_turn} = Conversations.reset_sandbox(sandbox)
# after the backfill and one teardown run
PROBE sandbox=terminated destroys=["test-sprite-3"] turn=interrupted conv=idle
```

**This is reachable on supported upgrade paths.** `docs/guides/operate/upgrade.md:92-101` documents boot migrations with rolling replicas and a migrate-in-a-Job path. The same holds for Kubernetes surge, even with one replica, and for Fly `rolling`. On the Job path, the stamp lands while only v0.20.0 pods serve. v0.20.0 ships the same `sweep_fenced_teardowns/0` reset branch, so a v0.20.0 pod can reset the home under its own running turn. The PR body says the risk applies only to a rolling upgrade. But rolling and Job upgrades are the documented default for multi-replica deployments, so the risk is not an edge case. The larger harm needs no running turn at all. Everything the user did on that home since the v0.20.0 upgrade is destroyed by a reset they asked for before it.

**My recommendation is to not finish a reset whose home was used after the fence.** "Finish the reset the user asked for" was right until the user went back to work on the machine. Their later actions are the newer statement of intent. A reset left unfinished costs one machine that is still billing, and the user can ask again. A reset wrongly finished loses files, and nothing brings them back. Leave teardowns as they are. They come from account deletion, termination and home destroys. They must finish, and the teardown path already waits for `any_server_alive?`.

Suggested predicate, added after line 84:

```sql
AND NOT (
      teardown_requested_at IS NULL
      AND EXISTS (
        SELECT 1 FROM turns t
          JOIN conversations c ON c.id = t.conversation_id
         WHERE c.sandbox_id = sandboxes.id
           AND t.inserted_at >= date_trunc('second', sandboxes.reset_requested_at)))
```

A v0.19.0 reset refuses a running turn, so any turn inserted at or after the fence came from v0.20.0. The `date_trunc` is there because `turns.inserted_at` is stored to the second. On a tie it skips the reset, which is the safe side. Log the skipped sandbox ids at warning level so an operator can tell the owners, and note that #2425 then drops the stale request. I planted this in the migration and re-ran everything:

```
PROBE sandbox=ready destroys=[] turn=running conv=running
4 tests, 0 failures      # the probe plus the PR's three tests
```

(Restored with `cp` then `touch`.) Please add the probe as a regression test. I would also add the running-turn and `ExecutionGuard` checks to `retry_pending_sandbox_reset/2` as defense in depth, but that is not blocking.

**The changelog must say what happens to these homes** (`changelog.d/2427-backfill-v019-fences.md:7-8`). Today it says "the reaper finishes the deletions and resets within minutes. No operator action is needed." With the fix, it should say that a reset is not finished on a home that has been used since v0.20.0, and that the migration logs those machines. If you keep the current behaviour instead, the note must warn self-hosters plainly. A home reset requested on v0.19.0 is wiped on v0.20.1 even if it has been used since, and in-flight turns on it are interrupted, so upgrade with no turns running.

## Non-blocking

- The comment at migration lines 35-36 says nothing moves a fenced row out of `ready` or `suspended` except to terminal. That is true on v0.19.0. On v0.20.0, which could not see the fence, the row can be parked, woken or be mid-wake. So the odd-shape branch can match rows touched on v0.20.0, such as one caught in `starting` during a rolling upgrade. Labelling those `teardown` is still the right answer. But the destroy then records `sandbox.destroyed`, and the conversations get no reset event. The claim that the branch "matches nothing in practice" should be reworded to "matches nothing v0.19.0 wrote".

## Held up

- **Labels, from raw SQL on a real v0.19.0 schema.** I rolled back to `20260915210000`, seeded with `psql`, then ran `ecto.migrate` forward:

| Row (v0.19.0 shape) | Result |
|---|---|
| reset-only, persistent `ready` or `suspended` | `destroying` / `reset` |
| teardown with both columns set, ephemeral | `destroying` / `teardown` |
| reset escalated to teardown, persistent `suspended` | `destroying` / `teardown` |
| teardown column only | `destroying` / `teardown` |
| `terminated` or `failed` with fences | untouched |
| no fence | untouched |
| odd shapes: persistent `starting`, ephemeral reset-only | `destroying` / `teardown` |

  The migration logged "stamped 8 sandbox(es)". A CSV diff of `status`, `mode`, both fence columns and `updated_at` showed no change. No lease was written, and `lease_epoch` stayed 0.
- **Teardown is tested first.** Both labels are in `Machines.Destroy`'s `@reasons`. `reason_from_string/1` maps `"reset"` and `"teardown"` exactly, and even an unknown string falls back to `:teardown`.
- **The implementer's reset-door branch.** `retry_pending_sandbox_reset/2` matches only a persistent `ready` or `suspended` row stamped `destroying`, and returns `:skipped` for anything else. A `reset` label on any other shape would therefore stand for ever, so labelling it `teardown` is correct. v0.19.0's `reset_sandbox` refuses non-persistent rows and any status other than `ready` or `suspended`, so v0.19.0 cannot produce such a row. The one exception is described in the non-blocking note above.
- **Ordering.** The last migration at v0.19.0 is `20260915210000`. The lease and `transition` columns arrive in `20260916120000`, after v0.19.0, so v0.19.0 rows reach the backfill with both null. The last migration at v0.20.0 is `20260918050000`, which matches the comment. Nothing on `main`, in the extension apps, or in any open PR sits between that and `20260918200000`. The only open PRs with migrations are #2427 and #2425. A fresh DB ran `…050000` then `…190000`. An upgraded DB did the same. I then applied #2425's drop on top: the columns went, and all 8 stamps survived.
- **Idempotency and `down`.** `ecto.rollback --step 1` followed by `ecto.migrate` gave byte-identical snapshots across all 11 columns I compared. The second `up` stamped nothing and logged nothing. A no-op `down` is correct. The columns still exist, and clearing the stamps would make the machines look live again.
- **Grace window.** `updated_at` is left at the v0.19.0 fence's write, so old teardowns finish on the first run and new ones wait out their window. The PR's three tests pass on my DB.

BinaryBourbon (approved):
Review of the tests and the end-to-end path, at `2241839`. **Approve.**

## Blocking

None.

## Non-blocking

- **The end-to-end reason assertion does not pin the label on its own.** `sandbox_fence_backfill_migration_test.exs:132` asserts `sandbox.destroyed` has reason `"teardown"`. `Destroy.reason_from_string/1` also answers `:teardown` for a null or unknown `transition_reason`, so a second producer gives the same value. I planted the teardown branch as `THEN NULL`. Only the labelling test failed, and the end-to-end test stayed green:
  ```
  == teardown-label-null
  >                WHEN teardown_requested_at IS NOT NULL THEN NULL
    1) test the backfill stamps each v0.19.0 fence with its own reason and leaves everything else
       code:  assert transition(teardown) == {"destroying", "teardown"}
  3 tests, 1 failure
  ```
  The file as a whole catches it, so this is not blocking. The comment at `:122-125` says the end-to-end test alone would catch a mislabel, and it would not. An assertion on the `sandbox.teardown_reconciled` metadata `"transition_reason" == "teardown"` would read the raw column and close that gap.
- **The end-to-end reset is seeded past the grace window.** `:106` seeds the reset 60 minutes old, so the test cannot show the PR's claim that "a reset has no window". I re-ran it with `reset: 1` and it passed, so the behaviour holds. Seeding it at a minute or two would make the test pin that claim.

## Held up

- **All six revert probes reproduce the PR's table exactly.** I planted each by distinctive string, then restored with `cp` and `touch` and confirmed each restore with `cmp`. The migration is loaded with `Code.require_file` on every run, so a stale build is not possible.

  | Planted | Failing test and line |
  |---|---|
  | reset tested first | all 3: `:79`, `:118`, `:156` |
  | terminal filter removed | `:84` |
  | `destroying` filter removed | `:87` |
  | `transition IS NULL` | `:81` (parking) |
  | odd branch answers `reset` | `:82` |
  | `updated_at = now` | `:92` snapshot, `:118` |
- **The end-to-end path is sound against the source.** `sweep_fenced_teardowns/0` selects `transition_reason == "reset" or updated_at < now - 15m`. A reset goes through `retry_pending_sandbox_reset/2`, whose guard is the migration's CASE: persistent, `ready`/`suspended`, `destroying`. The migration's terminal list matches the reaper's `@terminal_statuses ~w(terminated failed)`. `stamp_intent!/2` writes `destroying` with no transition guard, so pinning the `parking` overwrite as correct matches what a v0.20.x fence does.
- **No vacuous shapes.** No assertion is rescued, and none runs inside a stub. The destroy stub accumulates into an Agent and is read after the run. Each reset/teardown split is asserted both ways, with a `sandbox.reset` present and no `sandbox.destroyed`, and the reverse.
- **The file survives #2425, and the re-add hides nothing.** I merged #2425's head `5d01fca` onto this locally and migrated, and the columns were gone. The file passed 3 of 3, and the re-added columns rolled back with the test transaction. The one thing the re-add cannot see is migration order, so I checked it on the real path. I rolled back to `20260918050000`, seeded v0.19.0-shaped rows with SQL, and ran `mix ecto.migrate` through both migrations:
  ```
  == Running 20260918190000 ...BackfillDestroyingFromFenceColumns.up/0
  [warning] migration: stamped 2 sandbox(es) fenced by v0.19.0 as destroying; ...
  == Running 20260918200000 ...DropSandboxFenceColumns.up/0
  rv-done|terminated|||18:41
  rv-live|ready|||18:41
  rv-reset|ready|destroying|reset|18:41
  rv-teardown|ready|destroying|teardown|18:41
  ```
  The `updated_at` values stayed at the seeded hour-old times.
- **Gates.** `mix precommit --full` passed all 10 stages in 829s. The core suite had 6833 tests with 0 failures, and every extension suite had 0 failures. `scripts/changelog.py check` passed with 1 fragment and 2 entries.

jhgaylor:
Reply to the [data-correctness review](https://github.com/managoat/fountain/pull/2427#pullrequestreview-5251939043). Fixed in a5a0069a4. The head is now 1eda80b01.

**Blocking: a reset on a machine used since the request is skipped.** Jake decided the rule: a row whose intent is a reset is left unstamped and live when a conversation bound to it has a turn inserted at or after `reset_requested_at`. The predicate is yours, with the `date_trunc` tie going to the safe side.
- **By intent, not by label.** It tests `teardown_requested_at IS NULL` and sits in the `WHERE`, ahead of the CASE. The odd-shape branch therefore cannot turn a used reset into a destroy. A probe that applies the skip only to rows labelled `reset` fails the odd-shape assertion.
- **Binding.** I checked it against both tags. A turn has only `conversation_id`, and `conversations.sandbox_id` is the binding. Ending or releasing a conversation keeps it at both v0.19.0 and v0.20.0. The wake's co-tenant move is the only write that moves it, and it moves it onto a replacement machine. `turn_executions.sandbox_id` exists, but only bounded turns get a row.
- **Bound with no turn does not count.** Nothing has run on the machine, and turn admission refuses a `destroying` machine once the stamp lands. The migration comment says so, and a test pins it.
- **Logging.** After the update, every fenced row still unstamped is logged at warning level with id, user, provider, machine name and request time. The log names exactly the rows left.
- **Tests.** Your reproduction is a regression test: a running turn since the request, then the migration, then `perform(pass: "teardowns")`. The machine stays `ready` and the turn stays `running`. Removing the predicate fails it.

**Changelog.** It now says that a pending reset is not finished if the machine has run a turn since, that the machine stays live and is logged by id, and that an operator can reset it again from the console or the API. Pending deletions still finish within minutes.

**Non-blocking.** The odd-shape comment now says nothing v0.19.0 wrote reaches that branch, and that a row v0.20.0 touched can. I did not name a mid-wake `starting` row, because I found no v0.20.0 path that moves an existing `ready` or `suspended` row to `starting`.

I did not add the defense-in-depth checks to `retry_pending_sandbox_reset/2` here. The PR body's "Known limits" section lists that, along with two gaps in the skip: a turn admitted during the `UPDATE` in a rolling upgrade, and a deleted conversation's turns.

jhgaylor:
Reply to the [tests review](https://github.com/managoat/fountain/pull/2427#pullrequestreview-5252035495). Both points are in 1eda80b01.

1. **The stored teardown label.** The end-to-end test now asserts `transition_reason == "teardown"` in the `sandbox.teardown_reconciled` metadata. A planted NULL label now fails three tests, including the end-to-end test, where before it failed only the labelling test.
2. **A reset has no grace.** A new test fences a reset one minute before the upgrade and asserts that the first `perform(pass: "teardowns")` finishes it with a `sandbox.reset`. Planting the reset path onto the grace window in `SandboxReaper` fails that test, and only that test.

The full probe table for the current head is in the PR body.

BinaryBourbon (changes requested):
Round 2 data review at `1eda80b0`. The skip works: my round-1 reproduction now keeps the machine and its turn. The predicate, the binding and the log all hold. Two things block. Both are small, and I tested the fix for the second.

## Blocking

**1. No console can reset the machine, and an operator cannot reset it at all.** The same claim appears in three places:

- `changelog.d/2427-backfill-v019-fences.md:13`: "reset the machine again from the console or the API".
- The warning the migration logs, at `apps/fountain/priv/repo/migrations/20260918190000_backfill_destroying_from_fence_columns.exs:165`: "Reset it again from the console or the API".
- The migration comment at line 31.

The only reset is `DELETE /api/sandboxes/:id` (`sandbox_controller.ex`). It is scoped to the owner through `get_sandbox(id, user.id)`. The console LiveViews have no reset. The admin page's `retry_reset` calls `retry_pending_sandbox_reset/2`, which acts only on a row stamped `destroying`, so it skips these unstamped rows. A code search across `managoat` finds reset callers only in the SDKs and the Salon demo. The changelog addresses the operator, who has no path. Please say that the machine's owner can reset it again with `DELETE /api/sandboxes/:id`, and that the operator should tell them. Use the same wording in the log line and the comment.

**2. The rolling-upgrade gap in "Known limits" is real, but the stated mechanism is wrong, and the gap can be closed cheaply.** The limit says the update "does not wait for that turn", and that closing the gap means locking `turns`, which could deadlock. I raced it on a private DB. Session A mimicked turn admission (`conversations.ex:1644`): it took `FOR SHARE` on the sandbox row, inserted a `running` turn, held for 4 s and committed. Session B ran the migration's UPDATE verbatim.

```
# as shipped: the UPDATE waits on A's share lock (~3 s), then its EXISTS still
# reads the statement's original snapshot
UPDATE 1          -> transition=destroying, running turns=1

# lock the fenced rows first, then UPDATE as a second statement
SELECT id FROM sandboxes WHERE <@fenced> FOR UPDATE;   -- waits for A to commit
UPDATE ... (unchanged)                                 -- fresh snapshot, sees A's turn
UPDATE 0          -> transition=<null>, running turns=1
```

The UPDATE does wait. It misses the turn because PostgreSQL re-checks only the updated row after the wait, not the `EXISTS` subquery. With READ COMMITTED, a separate locking statement fixes that, because the UPDATE takes its snapshot after every admission already holding the row has committed. An admission that arrives after the lock waits for the migration to commit, then reads the stamp and is refused. This locks only `sandboxes`, the table admission locks first, so it does not bring in the turns-lock deadlock the PR describes. Add `ORDER BY id` to the lock query so two locks always come in the same order.

The fix is one extra `repo().query!` before line 129. The limit can then be removed. If you would rather keep it, correct its mechanism. Either way, "Known limits" should not justify the gap with a deadlock that the fix does not cause.

## Non-blocking

- The bound-with-no-turn test uses `insert_conversation(..., status: "idle")`. A first turn that is `pending` also counts, because the predicate does not filter on status, so that case is covered too. A comment in the migration could say so.

## Held up

- **Round-1 probe, same seed as before.** It covers a real `insert_conversation` and `insert_turn` after the fence, then the migration, then `perform(pass: "teardowns")`:
  ```
  PROBE1 sandbox=ready transition=nil turn=running conv=running destroys=[] resets=0 logged=true
  ```
- **The same-second boundary.** I put the fence at `:00.9` and the turn at `:00`, in the same second, and the reset was skipped. A turn one second earlier was not skipped, and the row was stamped `destroying`/`reset`. `turns.inserted_at` is `timestamp(0)`, and `Turn` uses `timestamps(type: :utc_datetime)`, which truncates. So a stored turn time is never after its real time, and a same-second turn counts as used.
- **The binding.** At both v0.19.0 and v0.20.0, the only writes that move `conversations.sandbox_id` are the wake paths: the waking conversation and the co-tenant move in `wake.ex`. Both move it onto a replacement after the old machine is gone. The `turn_machine.ex` hit is a struct default. Ending and releasing a conversation leave `sandbox_id` as it is. `turns.conversation_id` cascades on delete, so the deleted-conversation limit is stated accurately.
- **Bound with no turn.** Nothing has run on the machine, and any turn row, `pending` included, counts. Admission takes `FOR SHARE` and checks the stamp, and the teardown run respects a live lease. So a machine with no turn has no work in flight to interrupt.
- **The log names exactly the skipped rows.** I seeded a terminal row, an already-stamped row and an unfenced row, each with a turn since the fence, plus one used reset. Only the used reset was named, and the other three were unchanged. The log's `SELECT` reuses `@fenced` after the write, so a stamped or terminal row cannot appear.
- **The skip is in the `WHERE`, keyed on `teardown_requested_at IS NULL`.** So an odd-shape reset with a turn since stays live and is not destroyed. The PR's test covers this, and so did my run. Teardowns are unchanged.
- **The PR's six tests pass on my DB.** The round-1 non-blocking comment is reworded accurately: nothing v0.19.0 wrote reaches the odd-shape branch, but a row v0.20.0 touched can.

jhgaylor:
Reply to the [round 2 data review](https://github.com/managoat/fountain/pull/2427#pullrequestreview-5252131358). Both blockers are fixed in cd32f11aa. The full `mix precommit --full` passed at that commit.

**1. The reset wording.** I confirmed each point in the code. `DELETE /api/sandboxes/:id` scopes through `get_sandbox(id, user.id)`. No console LiveView calls `reset_sandbox/2`. The admin `retry_reset` calls `retry_pending_sandbox_reset/2`, which matches only a row stamped `destroying`. The migration comment, the logged warning and the changelog now say the same thing. The machine is left as its user last used it. Operators have no reset of their own. If a reset is still wanted, only the owner can request one, with `DELETE /api/sandboxes/:id`, so the operator should tell them.

**2. The rolling-upgrade race is closed.** `SELECT id FROM sandboxes WHERE <@fenced> ORDER BY id FOR UPDATE` now runs as its own statement before the unchanged `UPDATE`. Your mechanism was right and my "Known limits" entry was wrong. I checked the conditions:
- Admission (`_unsafe_create_turn_on_sandbox/4`) takes the sandbox row `FOR SHARE`, then inserts the turn.
- The migration reads `turns` without locking it, so no lock cycle through `turns` can form.
- Nothing in the app locks several sandbox rows in one transaction. `ORDER BY id` fixes the order anyway.
- `@disable_ddl_transaction` is not set, so the lock holds until the update commits.

A new test forces the race deterministically. It uses two independent unboxed connections and committed rows in a private schema, which `on_exit` drops even when a task crashes. One connection takes `FOR SHARE`, inserts a running turn and holds. The migration starts on the other. The admission commits only once `pg_blocking_pids` shows the migration waiting on it. The row must end unstamped and logged. With the lock, the test passed 20 of 20 runs. With the lock removed, it failed 10 of 10:

```
left:  [["destroying", "reset"]]
right: [[nil, nil]]
```

The rolling-upgrade entry is gone from "Known limits". The deleted-conversation limit and the note that the reset path checks only the lease remain. Your non-blocking point is in the comment too: any turn row counts, `pending` included. The PR body has the updated probe table.

BinaryBourbon (approved):
Round 3 data review at `cd32f11a`. **Approve.** Both round-2 blockers are fixed. The new race test is deterministic, bounded and self-cleaning. It fails without the lock.

## Blocking

None.

## Non-blocking

- **The admin page does have a destructive action.** `AdminLive.Sandboxes`'s `reap_sandbox` calls `Termination.reap_sandbox/2`. On a persistent home that destroys the machine, disk included, which is close to what a reset does. "Operators have no reset of their own" is literally true. An operator reading the upgrade note may still reach for reap instead of telling the owner. One clause would settle it: "the admin reap destroys the machine; it is not a substitute for the owner's reset".
- **A killed test VM leaks one schema.** `on_exit` handles a pass, a failure and a task crash. After a `kill -9` of the BEAM while the admission held its lock, one `backfill_race_<n>` schema was left behind: three empty tables, and no open backends. A later run of the file still passed beside it. CI is unaffected, because each partition job has its own Postgres. Locally the names come from `System.unique_integer/1`, which can repeat across VMs. A leaked name that repeats would fail `CREATE SCHEMA` before `on_exit` is registered. A random suffix, or a `DROP SCHEMA IF EXISTS` before the create, would remove that edge.

## Held up

**Wording.** The migration comment (lines 30-35), the logged warning (lines 191-193) and the changelog now agree. The machine is left as its user last used it, and the operator has no reset. Only the owner can request one, with `DELETE /api/sandboxes/:id`. All three are true: the endpoint scopes through `get_sandbox(id, user.id)`, no console LiveView resets, and the admin `retry_reset` matches only a row stamped `destroying`.

**The lock.**
- Line 149 locks with `WHERE #{@fenced} ORDER BY id FOR UPDATE`.
- The UPDATE at line 163 is `WHERE #{@fenced} AND NOT (#{@used_since_reset})`. That is the same module attribute, narrowed further, so every updated row was locked first.
- A real `ecto.migrate --log-migrations-sql` on my private DB ran the lock, then the UPDATE.
- The whole `up` runs in one transaction. The module sets no `@disable_ddl_transaction`, and Ecto's `run_maybe_in_transaction/5` wraps `up` and the `schema_migrations` insert in `repo.transaction(..., timeout: :infinity)`. `repo().query!` runs in that same process.

**The race test.**
- **Isolation.** The file is `async: false`. Its committed rows live only in a private schema. The unboxed connections touch no shared table.
- **Real SQL.** It runs the real migration module through `Ecto.Migration.Runner.run/8`, with `search_path` pointed at the private schema, not a copy of the SQL.
- **Bounded waits.** Every wait has a timeout. Each `assert_receive` gives up after 5 s. The admission's `receive` raises after 5 s. `await_blocked/2` has a 5 s deadline. Each `Task.await` uses its 5 s default. The `after` brutal-kills the migration task.
- **Stability.** It passed 20 of 20 runs, and the whole file passed 7 of 7.
- **Revert probe.** With the lock removed, it failed 5 of 5:
  ```
  left:  [["destroying", "reset"]]
  right: [[nil, nil]]
  ```
- **Cleanup.** Leftover `backfill_race_*` schemas: 0 after 20 passes, 0 after the 5 lock-removed failures, 0 after a planted `raise` in the admission task, and 1 after a `kill -9` of the VM (see above).

**From round 2, still standing.** Nothing under `lib/` changed since `1eda80b0`.
- The skip keys on intent, in the WHERE clause, so odd-shape resets are skipped too.
- A turn in the fence's own second counts as use. A turn one second earlier does not.
- `conversations.sandbox_id` moves only onto a replacement machine.
- A bound conversation with no turn has nothing in flight to interrupt.
- The log names exactly the skipped rows.
- Teardowns are unchanged.
- The regression test for my round-1 reproduction passes.

PR #2433, opened 2026-09-18
chore(release): v0.20.1

jhgaylor:
Version bump for v0.20.1: both `mix.exs` files, the six self-host image pins, and the changelog roll (`[Unreleased]` plus every `changelog.d/` fragment into `## [0.20.1]`).

Merging this PR tags the squash commit `v0.20.1` and dispatches `release.yml`, which builds the CLI binaries, publishes `ghcr.io/managoat/fountain:v0.20.1`, boots it through the documented quick start and creates the GitHub Release.


BinaryBourbon (approved):
Checked the diff. The version moves from 0.20.0 to 0.20.1 in both mix.exs files and in the compose, k8s, fly and render image tags. The CHANGELOG entry is exactly the three fragments (#2427, and two from #2402), and the fragments are deleted. All three are fixes, so a patch release fits.

Issue #2395, opened 2026-09-17
Decide: sandbox files no-wake guarantee under autonomous provider sleep


BinaryBourbon:
## Decision needed

Define what ADR 0039's sandbox-files promise of "never a wake" means when a provider sleeps independently of Fountain. This is the provider-contract follow-up to #1715, separate from coordinating reads with Fountain-controlled park/destroy.

## Evidence as of 2026-09-17

- The original files API (#1397, ADR 0039) refuses non-ready sandbox rows and promises that reads do not resume parked compute.
- The pinned `managoat_sandbox` 0.3.0 Sprites adapter has a no-op `suspend/1`; its exec path passes environment/directory options to `Sprites.spawn` and has no conditional no-wake option.
- Sprites documents automatic wake on commands: https://docs.sprites.dev/quickstart/ . Its documented exec parameters expose no "execute only if already awake" condition: https://sprites.dev/api/sprites/exec . This is evidence about the current adapter and published API, not proof that the vendor cannot offer another capability. Recheck both when picking this up.
- #2139 already called out this limitation while fixing symlink confinement. A fresh status probe followed by exec still has a race, and a Fountain machine lease cannot prevent autonomous provider sleep.

## Questions to resolve

1. Must reads avoid every provider wake, including autonomous sleep while Fountain still records the sandbox as ready? If so, identify an atomic provider capability or another read mechanism that actually satisfies the promise across the supported operations/providers. Unsupported providers need explicit behavior rather than a silent fallback to waking exec.
2. Alternatively, is the intended guarantee that reads never resume a Fountain-parked sandbox or race its park/destroy, while an otherwise-admitted read may trigger the provider's own wake behavior? That changes the stated product/cost contract and needs an explicit maintainer decision and an ADR/docs amendment.

Neither alternative is selected by this issue. Do not assume a new adapter option is sufficient unless the provider enforces it atomically, or label this `needs:upstream` before establishing that dependency.

## Scope and sequencing

This investigation can start independently of #2386 and the rest of #2344. The companion lifecycle fix #2394 can proceed on its narrower guarantee, but it must not claim to close this decision or silently relax the public promise. No provider redesign or cross-repository implementation is bundled into that PR.

## Done when

- [ ] The maintainer's chosen guarantee is recorded in ADR 0039 and the public docs with its cost implications.
- [ ] Current provider/adapter support is checked and documented for list, read, diff and status; a preflight probe is not presented as atomic no-wake enforcement.
[...]

PR #2435, opened 2026-09-18
fix(sandbox-files): admit reads against machine park and destroy (#2394)

### LINKED ISSUE #2394: Sandbox files: coordinate read admission with machine park and destroy

jhgaylor:
Closes #2394. Part of #1715 (which stays open for #2395).

## The race

`SandboxFiles` checked `status: "ready"` on the struct the controller had fetched, then called `Managoat.Sandbox.exec/4` with nothing in between. A park or destroy that committed in that gap went unseen, so the read ran against a machine that was being suspended or deleted. The first test in `reads_test.exs` shows this deterministically: a struct that still says `ready` over a row that is now `suspended`. With the admission removed, that test and 14 others fail.

## The fix

The new module `Fountain.Machines.Reads` routes all four operations (list, read, diff, status) through one admission path. It is the pair of refusals that `Machines.Admission` already has for turns, seen from the read's side:

- **Admission.** One short transaction under advisory lock 4316, the same lock `Lease.claim/4` takes. The decision uses the row read under that lock, not the caller's struct:
  - a row that is not `ready` gets `409 sandbox_not_ready`, as before;
  - a row with any `transition` stamp gets `503 sandbox_unavailable`. That covers durable `destroying` with a lapsed lease, and an abandoned `parking` stamp, which can sit over a machine that is already suspended;
  - a row with a live lease also gets `503 sandbox_unavailable`.
- **Exclusion.** An admitted read commits a `sandbox_reads` row with a 40 s window before any provider I/O happens. `Park` and `Destroy` call `Reads.drain/2` inside their `Renewal.around`, before the checkpoint, suspend or destroy call, and wait until no unexpired read is left. No read can be admitted while their lease is live, so the wait is bounded by one window, and a polling client cannot keep a machine awake.
- **The window is enforced locally.** `exec`'s `:timeout` bounds only output collection (`ExecDeadline`), so the exec runs in a `Fountain.TaskSupervisor` task with a hard cutoff 5 s before the window ends. The cutoff is an absolute monotonic instant taken before admission. The task enforces it itself: before it calls the provider it starts a watchdog with an absolute timer for that instant, then refuses if the instant has passed. So a task the supervisor starts late never execs, and the cutoff fires whether or not the caller is still alive, even if the caller dies as the task starts. A read that hits it answers `503 sandbox_unreachable`.
- **Release** runs in an `after` on return, error and raise. A caller that is killed leaves a row that expires on its own; the next admission on that machine deletes it.
- **Queued work.** None exists. Admission runs inline in the caller, as `end_turn/3` does, and a read whose budget is already spent answers without calling the provider.

Provider I/O stays outside every transaction and advisory lock (ADR 0058).

## The guarantee this PR proves

Fountain does not start a park's or a destroy's provider work while an admitted read's window is open, and it does not start a read on a machine that an operation holds or has stamped. It does **not** decide whether a provider that suspends a machine by itself wakes it for an exec. That is #2395's decision under #1715, and ADR 0039's no-wake promise is not restated on the strength of this fix. The ADR 0039 amendment says so.

## Behavior and contract changes

- A sandbox that is mid-park or mid-destroy now answers `503 sandbox_unavailable` with `Retry-After`. Before, the read raced the operation. 503 was already declared on all four routes; only its description widened. `contract.json` is unchanged, and `openapi.ts` changed in doc comments only, so this PR carries `release:skip-sdk`.
- New table `sandbox_reads` (migration `20260918220000`). It cascades on sandbox delete. It does not touch `sandboxes`, and it sorts after #2425's `20260918200000` column drop, which has landed; this branch is rebased on it.
- Changelog fragment `changelog.d/2394-sandbox-files-read-admission.md`, plus an amendment to ADR 0039.

## Known cost, pinned rather than hidden

`machine_bounds_test.exs` pins that the read window (40 s) is **longer** than `Machine.destroy_timeout_ms/0` (20 s). If a read is stuck at the provider until its cutoff, a destroy's caller can time out during the drain. That caller sees `sandbox_unavailable`, and the destroy still finishes in the owner: `destroying` is durable and a late destroy message is idempotent. An ordinary read is gone in well under a second. The same test pins that the window holds the files API's 30 s exec timeout and sits under both lease TTLs.

## Tests

`test/fountain/machines/reads_test.exs` has 27 cases:

- the stale-ready race, deterministic;
- suspended, terminated, failed and provisioning rows, and a deleted row;
- a live lease; durable `destroying` with a lapsed lease; an abandoned `parking` stamp; a lapsed lease with no stamp, which is admitted;
- release on return, on provider error and on raise; the cutoff; a spent budget, which never execs; a task the supervisor starts after the window, which never execs; a caller killed as `async_nolink` returns, whose read is still killed at the cutoff;
- a killed caller: its row outlives it, the exec is killed at the cutoff, the drain ends within the window, and the next admission sweeps the row;
- **through the owner, with competing processes:** read-first then park, and read-first then destroy (the provider call waits, and a second read is refused meanwhile); park-first and destroy-first (the read is refused and never execs); a dead caller's read delays a destroy by at most its window;
- **on real connections (`unboxed_run`):** a claim holding 4316 blocks the admission (observed with `pg_blocking_pids`), then the committed lease refuses it; a read admitted on one connection is counted by a drain on another.

**Revert probes.** Each guard was removed in turn, restored from a `cp` snapshot, and the matching tests failed:

| Guard removed | Failing tests |
|---|---|
| admission in `SandboxFiles` | 15 |
| park drain | 1 |
| destroy drain | 2 |
| transition check | 2 |
| lease check | 2 (including the real-connection lock case) |
| the task's watchdog | 2 (including the caller killed as its task starts) |
| the task's own clock check | 1 (the late-started task) |
| release | 9 |
| budget check | 1 |
| expired-row sweep | 1 |

The controller test's `exec` stub now sends to a captured test pid, because the exec runs in the read's task (the same `$callers` trap as the owner).

## Size

This is one PR of about 500 lines of code and docs plus about 650 lines of tests, which is above the usual size. It isn't a stack because a failing repro can't merge on its own, and the admission path, the two drains and their race tests don't separate into PRs that could each be reviewed as correct.


BinaryBourbon (changes requested):
Adversarial correctness review — **CHANGES REQUESTED**.

Reviewed the complete 12-file diff from `6a73f4334031b40b7a8240511f8fae5bc0eb3682` to `f7ac4e7ed085a182f1f507139b9f4586d64363e4`, with emphasis on admission ordering, task cancellation, lease renewal, recovery, and regression coverage.

**[P2] Preserve the absolute cutoff through task startup.** [`apps/fountain/lib/fountain/machines/reads.ex:281–283`](https://github.com/managoat/fountain/blob/f7ac4e7ed085a182f1f507139b9f4586d64363e4/apps/fountain/lib/fountain/machines/reads.ex#L281-L283)

`run/3` computes the remaining budget before calling `bounded/3`, but `Task.Supervisor.async_nolink/2` synchronously waits for the shared supervisor (`GenServer.call(..., :infinity)` in the pinned Elixir implementation). The task invokes `fun` without checking the absolute cutoff, and the parent then arms a fresh relative timer. If startup is delayed past the admission window, a park/destroy can finish draining the expired row and call its provider before this read subsequently starts. That restores the same exec-versus-suspend/destroy race this PR intends to exclude.

I reproduced this with the exact `bounded/3`, `await/3`, and `cut_off/2` function ASTs from this head under Elixir 1.19.2 / OTP 28: suspend `Fountain.TaskSupervisor`, start with a 400 ms remaining budget for a 500 ms window, then resume after 700 ms. The provider callback started at **708 ms** and its watchdog killed it only at **1111 ms**. This was an isolated local process probe; it made no database or remote provider calls.

Carry the absolute monotonic cutoff into the task, refuse expired work before invoking `fun`, and establish its watchdog before provider work begins using the remaining time. Add a regression that stalls task startup until after admission expiry and verifies that the provider is never called. The existing zero-window test only exercises expiry before `bounded/3`, so it cannot detect this gap. Stable finding anchor: `read-admission-cutoff-covers-task-startup`.

Validation: source review and the focused process probe completed. I did not run the database-backed suite locally; local PostgreSQL was unavailable. The coordinating reviewer reported successful exact-head server partitions, static analysis, contract/release checks, SDK checks, and coverage. Those existing checks do not cover the reproduced startup delay. This review does not claim the installed Review Loop service or its verifier ran.

Coverage: `apps/fountain/lib/fountain/machines/destroy.ex`; `apps/fountain/lib/fountain/machines/park.ex`; `apps/fountain/lib/fountain/machines/reads.ex`; `apps/fountain/lib/fountain/sandbox_files.ex`; `apps/fountain/lib/fountain_web/controllers/sandbox_files_controller.ex`; `apps/fountain/priv/repo/migrations/20260918220000_create_sandbox_reads.exs`; `apps/fountain/test/fountain/machines/machine_bounds_test.exs`; `apps/fountain/test/fountain/machines/reads_test.exs`; `apps/fountain/test/fountain_web/controllers/sandbox_files_controller_test.exs`; `changelog.d/2394-sandbox-files-read-admission.md`; `decisions/0039-sandbox-files-over-the-api.md`; `sdk/typescript/src/generated/openapi.ts`. Also traced the sandbox lock, lease checks, renewal bracket, controller/fallback behavior, and task-supervisor implementation.

BinaryBourbon (changes requested):
Adversarial review — product and API compatibility.

Reviewed head `f7ac4e7ed085a182f1f507139b9f4586d64363e4` against base `6a73f4334031b40b7a8240511f8fae5bc0eb3682`. Source review complete. **Changes requested:** the delayed task-start case still violates the promised read-versus-park/destroy exclusion. No separate wire-shape or SDK error-mapping defect found.


**[P2] Preserve the absolute read deadline across task startup** — `apps/fountain/lib/fountain/machines/reads.ex:281–283`.

`run/3` computes a relative budget before `Task.Supervisor.async_nolink/2`, but starting that task may block behind the supervisor. The callback and `kill_after` both subsequently receive the original budget. If startup is delayed past the admission's expiry, a drain can already have permitted a checkpoint, suspend or destroy when this callback finally begins its provider exec. An ordinary files request can therefore reach a machine after it was parked, despite the new exclusion guarantee.

I independently reran the exact-function reproducer from the correctness review with the pinned toolchain: with a 500 ms window and a 400 ms budget, suspending the task supervisor and resuming it after 700 ms started the callback at 706 ms and killed it only at 1,113 ms. The callback should never have run after the window expired. This probe extracts the unchanged `bounded/3`, `await/3` and `cut_off/2` functions; it uses no database or provider.

Carry the original absolute cutoff into the task, reject expired work immediately before invoking the callback, and arm its independent cutoff before provider work begins. Add a regression that delays supervisor startup beyond the read window and proves no provider callback occurs. A timely 503 and existing SDK retry handling do not repair a provider call made after admission expired.

The four sandbox file routes retain their request and success-response shapes. The new admission refusal reaches the existing `503 sandbox_unavailable` fallback with `Retry-After: 30`; the hard cutoff uses the existing `503 sandbox_unreachable` fallback with `Retry-After: 10`. TypeScript, Python, Elixir and both Swift error layers already preserve this retry behavior. The controller declarations and committed wire contract accept these responses; the generated TypeScript changes are descriptions only, which the contract projection intentionally excludes. No endpoint, schema-guard exemption, SDK version or conformance skip was added.

Traced the scoped controller fetch through all four SandboxFiles entry points, the new admission and release, and the park/destroy drains. Reviewed the additive migration and its cascading sandbox reference, stale-state and concurrency test assertions, the accepted ADR 0039/0058 constraints, and existing SDK/CLI/plugin callers. The documented possibility of a slow read delaying a destroy beyond its caller's timeout is an explicit tradeoff; the durable destroy continues in the owner. Provider-autonomous suspend behavior remains #2395's separate scope.

Changed-path coverage: `apps/fountain/lib/fountain/machines/{destroy,park,reads}.ex`; `apps/fountain/lib/fountain/sandbox_files.ex`; `apps/fountain/lib/fountain_web/controllers/sandbox_files_controller.ex`; `apps/fountain/priv/repo/migrations/20260918220000_create_sandbox_reads.exs`; `apps/fountain/test/fountain/machines/{machine_bounds_test,reads_test}.exs`; `apps/fountain/test/fountain_web/controllers/sandbox_files_controller_test.exs`; `changelog.d/2394-sandbox-files-read-admission.md`; `decisions/0039-sandbox-files-over-the-api.md`; `sdk/typescript/src/generated/openapi.ts`.

Validation: the pinned-toolchain delayed-start probe reproduced the defect above. Python contract verification passed (67 operations, 40 schemas, 5 enums); shared conformance lint passed (29 scenarios). This reviewer did not execute the server suite, the four conformance adapters or a fresh OpenAPI export; inspected assertions and source are not claimed as executed tests. This is a user-requested review, not a Review Loop verifier result.

BinaryBourbon (changes requested):
Adversarial review — tenant isolation, credential boundaries, and resource lifetime.

Reviewed head `f7ac4e7ed085a182f1f507139b9f4586d64363e4` against `6a73f4334031b40b7a8240511f8fae5bc0eb3682`. Verdict: **CHANGES REQUESTED**.

**[P2] Arm the lifetime guard before the unlinked read can start.** In [`Reads.bounded/3`, lines 281–283](https://github.com/managoat/fountain/blob/f7ac4e7ed085a182f1f507139b9f4586d64363e4/apps/fountain/lib/fountain/machines/reads.ex#L281-L283), `async_nolink` releases the callback before the requesting process arms `:timer.kill_after`. If the request process dies in that interval, neither the timer nor `Task.yield` ever runs. A blocked provider startup then outlives its `sandbox_reads` expiry; park/destroy can proceed while that local read task remains capable of reaching the provider. This violates the dead-caller guarantee and can leak tasks or wake a machine after reclamation. It is distinct from a remote script continuing after a locally closed connection.

I reproduced this using the exact `bounded`, `await`, and `cut_off` function bodies extracted from this head, with only `bounded` made public, under the pinned Elixir 1.19.2 / OTP 28 toolchain. For 5,000 reads with an 80ms budget, the callback immediately killed its caller and then blocked: all 5,000 tasks were still alive 500ms after every callback started. As a control, delaying caller death by 20ms left 0 of 100 tasks alive. Probe tasks were explicitly cleaned up; no database or provider was contacted. The existing caller-death test waits for a callback message before the test process kills the caller, so it does not force the pre-arming interval.

Arm an independent watchdog from inside the task before invoking the provider callback, or use a handshake that cannot release provider work without a lifetime guard. Carry the absolute admission deadline into that guard and reject an already-expired task. Add a deterministic regression for caller death before the current caller-side arming step.

Scope: inspected all 12 changed paths and relevant auth, scoped sandbox lookup, redaction/path handling, lease, renewal, and provider-exec callers. The four API routes retain full-key scope and the initial `user_id`-scoped fetch; existing controller assertions cover foreign-tenant 404 and sprite-key 403 without provider exec. New SQL uses bound parameters. I found no additional concrete tenant or secret-exposure regression.

Validation limitation: this is a completed source review plus isolated BEAM fault-injection probes, not an integrated application/database test run. I did not run `mix precommit` or claim that the installed Review Loop service verified this review.

jhgaylor:
All three reviews found one defect, `read-admission-cutoff-covers-task-startup`. It is fixed in 72819e962.

**What changed in `Reads`.** The task now carries the absolute monotonic cutoff instead of a relative budget. Inside the task, before it calls the read function, it starts an unlinked watchdog. The watchdog sets `:erlang.start_timer(cutoff, self(), :cutoff, abs: true)` and kills the task when the timer fires, or exits when the task goes down. The task then checks the clock and returns `:spent` without calling the provider if the cutoff has passed. The caller's `yield` waits only until the same instant. Nothing on the caller's side of `async_nolink` enforces the cutoff any more.

- **Late task start (reviews 1 and 2).** A task that starts after the cutoff never calls the provider. One that starts in time already has its watchdog, set to the original instant.
- **Caller killed before arming (review 3).** On Elixir 1.19.2 the task waits for its caller's handshake and exits if the caller dies first. The open window was therefore from the handshake send to `:timer.kill_after`'s call to `timer_server`. Once that call is sent, the server arms the timer even if the caller is dead. The watchdog is now set up in the task before the callback runs, so that window no longer matters.

**Regressions.** Both are deterministic, and both failed 3 runs out of 3 against the previous `bounded/3`. That probe used the old code with only the new test seam added.

- *a task the supervisor starts after the window never calls the provider*: a private `Task.Supervisor` (new `:task_supervisor` test seam) is suspended for 700 ms against a 500 ms window.
- *the cutoff holds for a caller killed the moment its task has started*: `Task.Supervisor.async_nolink` is stubbed to call the original and then kill its caller before returning. This adds a new `Mimic.copy(Task.Supervisor)` to `test_helper.exs`.

Each half of the fix was also planted on its own: without the watchdog, 2 tests fail; without the task's clock check, 1 fails.

**Validation.** `reads_test.exs` passes 27 of 27, three runs. `test/fountain/machines/`, `conversations/`, the sandbox-files controller and the reaper suite: 2375 tests, 0 failures. `mix precommit compile format credo dialyzer` passes. The PR body's window paragraph and revert table are updated.

BinaryBourbon (approved):
APPROVE — independent adversarial correctness review of the complete PR at `72819e9625724c1bd2a10702fc67c5d9f77ade66`, against base `6a73f4334031b40b7a8240511f8fae5bc0eb3682`.

The two previously reported timeout defects are resolved. [The task now receives the original absolute cutoff, checks it before invoking the provider, and owns its watchdog](https://github.com/managoat/fountain/blob/72819e9625724c1bd2a10702fc67c5d9f77ade66/apps/fountain/lib/fountain/machines/reads.ex#L294-L361). Delayed supervisor startup cannot grant a fresh execution window, and caller death immediately after task creation cannot remove the task's cutoff. I found no remaining blocking correctness defect after tracing admission against lease claims, both read/park and read/destroy orderings, transaction visibility, release/error paths, and abandoned-read expiry.

Validation: four isolated ExUnit probes passed on the pinned OTP 28 / Elixir 1.19.2, using the exact source AST for the deadline functions: ordinary result propagation, supervisor startup delayed past the deadline with no provider call, caller killed immediately after `async_nolink`, and stalled-provider cutoff with the execution task confirmed dead. Reviewed the PostgreSQL-backed regression assertions, but did not execute the integration suite locally because a local test database was unavailable. This is an independent source review, not a claim that Review Loop's separate required verification ran.

Coverage included all 13 changed paths and relevant callers/lease/renewal code:

- `apps/fountain/lib/fountain/machines/destroy.ex`
- `apps/fountain/lib/fountain/machines/park.ex`
- `apps/fountain/lib/fountain/machines/reads.ex`
- `apps/fountain/lib/fountain/sandbox_files.ex`
- `apps/fountain/lib/fountain_web/controllers/sandbox_files_controller.ex`
- `apps/fountain/priv/repo/migrations/20260918220000_create_sandbox_reads.exs`
- `apps/fountain/test/fountain/machines/machine_bounds_test.exs`
- `apps/fountain/test/fountain/machines/reads_test.exs`
- `apps/fountain/test/fountain_web/controllers/sandbox_files_controller_test.exs`
- `apps/fountain/test/test_helper.exs`
- `changelog.d/2394-sandbox-files-read-admission.md`
- `decisions/0039-sandbox-files-over-the-api.md`

BinaryBourbon (approved):
**APPROVED — independent adversarial API/product and maintainability review**

Reviewed the complete PR at `72819e9625724c1bd2a10702fc67c5d9f77ade66` against base `6a73f4334031b40b7a8240511f8fae5bc0eb3682`, including the previous deadline fixes. I found no actionable blocking defect.

- The delayed-supervisor case carries the original absolute cutoff into the task and rechecks it before calling the provider. The task now creates its own watchdog, so killing the caller immediately after task startup no longer removes its cutoff. The two added regressions target those specific prior failures.
- Admission, drain and release fit the existing ownership protocol: admission commits under the sandbox lock, provider I/O runs outside the transaction, and park/checkpoint/destroy drain within lease renewal. Existing source tests cover both operation orders, real separate database connections, stale rows, release, expiry and caller death.
- All four routes retain their response shapes. Their new refusal uses the existing `503 sandbox_unavailable` fallback with `Retry-After: 30`; TypeScript, Python, Elixir and Swift already classify that code as retryable/not-ready. The generated OpenAPI descriptions reflect the change. The canonical wire projection deliberately omits those descriptions, so this change does not require a different `sdk/contract/contract.json`. I checked the actual controller/fallback, SDK transports/mappings, existing conformance scenario and CLI surface against trusted ADRs 0039/0058.

Coverage: all 13 changed paths — `machines/destroy.ex`, `machines/park.ex`, `machines/reads.ex`, `sandbox_files.ex`, `sandbox_files_controller.ex`, migration `20260918220000_create_sandbox_reads.exs`, `machine_bounds_test.exs`, `reads_test.exs`, `sandbox_files_controller_test.exs`, `test_helper.exs`, `changelog.d/2394-sandbox-files-read-admission.md`, `decisions/0039-sandbox-files-over-the-api.md`, and `sdk/typescript/src/generated/openapi.ts`.

Validation I actually ran on the reviewed snapshot: Python SDK contract verification passed (67 operations, 40 schemas, 5 enums); conformance scenario lint passed (29 scenarios). I did not run the Elixir/database integration suite in this reviewer. This approval is a completed source review, not a claim that those integration tests ran locally or that local task cancellation stops remote provider work.

BinaryBourbon (approved):
Adversarial security / tenant-isolation / resource-lifetime review: **APPROVE**.

Reviewed head `72819e9625724c1bd2a10702fc67c5d9f77ade66` against base `6a73f4334031b40b7a8240511f8fae5bc0eb3682`, including all 13 changed paths and the relevant authentication, ownership, redaction, lease, renewal, and machine-owner callers. No blocking security defect found.

- The four routes still require a full-scope authenticated key and establish sandbox ownership through `get_sandbox(id, current_user.id)` before read admission. The refreshed row does not create a new user-facing lookup path. Existing second-tenant and sprite-key rejection tests remain intact. [Ownership boundary](https://github.com/managoat/fountain/blob/72819e9625724c1bd2a10702fc67c5d9f77ade66/apps/fountain/lib/fountain_web/controllers/sandbox_files_controller.ex#L217).
- Provider execution uses the admitted machine row; fixed scripts, positional arguments, path confinement, tenant DEK selection, and redaction remain in place. Read records contain IDs and expiration timestamps, with a sandbox foreign key and cascading deletion. [Execution boundary](https://github.com/managoat/fountain/blob/72819e9625724c1bd2a10702fc67c5d9f77ade66/apps/fountain/lib/fountain/sandbox_files.ex#L455).
- The earlier task-start and caller-death findings are addressed: the task checks the absolute cutoff before invoking the provider callback and starts an independent watchdog. A late task refuses execution; caller death no longer removes the task's cutoff. Park and destroy drain admitted reads under lease renewal. [Cutoff implementation](https://github.com/managoat/fountain/blob/72819e9625724c1bd2a10702fc67c5d9f77ade66/apps/fountain/lib/fountain/machines/reads.ex#L294).

Validation: an isolated probe using the exact head's window helper implementations passed **3 tests, 0 failures** on the pinned toolchain: delayed supervisor startup, caller death during a blocked read, and prompt watchdog cleanup after normal completion. Inspected the database, lifecycle, and controller regression assertions; did not run the database-backed suite locally or call real providers. The documented remote-provider continuation limit and destroy caller's shorter timeout remain explicit; the owner continues its durable destroy after a timeout response.

PR #2444, opened 2026-09-18
chore(release): v0.21.0

jhgaylor:
Version bump for v0.21.0: both `mix.exs` files, the six self-host image pins, and the changelog roll (`[Unreleased]` plus every `changelog.d/` fragment into `## [0.21.0]`).

Merging this PR tags the squash commit `v0.21.0` and dispatches `release.yml`, which builds the CLI binaries, publishes `ghcr.io/managoat/fountain:v0.21.0`, boots it through the documented quick start and creates the GitHub Release.


BinaryBourbon (approved):
Checked the diff. The version moves from 0.20.1 to 0.21.0 in both mix.exs files and in the compose, k8s, fly and render image tags. The CHANGELOG entry is exactly #2425's fragment, which is removed. It's a minor bump because the release drops columns and needs rolling upgrades to start from v0.20.0 or later.

PR #2199, opened 2026-09-14
refactor(chatgpt): remove the dormant tenant-owner half of ChatGPTAccounts; ADR 0052 stays Proposed

jhgaylor:
The last of the #2176 decision 1 deletions (stage 8, on top of the workers PR).

## What changed

- Deleted `ChatGPTAccounts.status_for_user/1`, `credential_for_user/4` and `refresh_for_user/3` with their private helpers (`user_credential`, `user_account_state`, `pinned_user_grant`, `user_grant_query`, `with_eligible_owner`, `safe_reason`), and every owner branch of the shared refresh path: `validate_refresh_identity`, the `{:user_revoked, ...}` audit row (`chatgpt.reconnect_required`), the user refresh-failure telemetry event, and the owner-scoped fence queries. `current_query`, `current_result` and `swap_in` are now platform-only.
- Deleted `Fountain.ChatGPTAccounts.Grant`, `Fountain.Broker.Native.ProtectedCompiler`, its test, the `test/fixtures/codex_protected/` capture fixture and `scripts/probe-codex-protected.py` (the fixture's generator; nothing else referenced it). `Reserved.conflict?/1` loses its typed-grant clause.
- Deleted the `user_grant!` fixture. Its only users were `chatgpt_user_refresh_test.exs` and `chatgpt_user_refresh_isolation_test.exs`, which #2188 lists here but which were deleted in the workers PR because they reference the coordinator and `refresh_serialized_for_user/3`.
- Rewrote `chatgpt_accounts_test.exs` for one live owner with the column kept: the `Cipher` owner-dispatch tests stay, and so does the proof that no platform read or admin mutation selects an owned row. Tests that read through the deleted functions are gone.
- Rewrote the `ChatGPTAccounts` and `PlatformChatGPT.Account` moduledocs for one owner with the `user_id` column kept.
- ADR 0052's status block records the deletion and names `105fb6d1` (the commit before #2198 merged) as the last commit of `main` that carried the whole tenant-owner half; the ADR stays Proposed and its frontmatter is untouched. That paragraph is the one conflict with the merged #2191 wording, resolved on rebase by keeping #2191's text and folding the pointer in. ADR 0053's stage table no longer lists the compiler as a standing prerequisite (one sentence). `okf validate decisions` passes and the index is unchanged.

## What stayed

[...]

PR #2198, opened 2026-09-14
refactor(chatgpt): delete the user-grant refresh workers, sweep, coordinator and supervisor

jhgaylor:
Stage 8 of #2176 (decision 1: delete the dormant tenant-owner half of ADR 0052) lands as two PRs. This is the root; #2188 planned a three-PR stack with a platform lock contention test first, and that first PR turned out to be already on `main` (see below).

## What changed

- Deleted `Fountain.Workers.ChatGPTGrantKeepalive` (the per-grant Oban worker), `Fountain.Workers.ChatGPTKeepaliveSweep` (the paginated daily sweep), `Fountain.ChatGPTAccounts.RefreshCoordinator` (node-local coalescing of same-grant callers) and `Fountain.ChatGPTAccounts.RefreshSupervisor`, plus its child entry in `application.ex`.
- Deleted the `chatgpt_refresh` Oban queue and the `37 4 * * *` cron entry for the sweep in `config/config.exs`. That cron entry was the only place the sweep was ever enqueued.
- Deleted `ChatGPTAccounts._unsafe_due_user_grants/2` and `refresh_serialized_for_user/3` with the private `refresh_user_locked/1`. `refresh_for_user/3` keeps its pinned-grant and eligibility checks and now answers `{:error, :refresh_unavailable}`; it goes with the other `*_for_user` reads in 8c.
- Deleted five test files: `chatgpt_refresh_coordinator_test.exs`, `workers/chatgpt_grant_keepalive_test.exs`, `workers/chatgpt_keepalive_sweep_test.exs` (the three #2188 lists here), plus `chatgpt_user_refresh_test.exs` and `chatgpt_user_refresh_isolation_test.exs`, which #2188 lists under 8c but which reference the coordinator and `refresh_serialized_for_user/3` and so cannot compile once those are gone.
- Trimmed the `Refresher` and `ChatGPTAccounts` moduledocs that pointed at the deleted modules.

## What stayed

`ChatGPTAccounts.RefreshLock` and every `platform_*` function, `PlatformChatGPT.*` (the `Refresher` still serializes the deployment's grant on one node and the try-lock still excludes other nodes), `Cipher` with its owner dispatch, `Reserved` and its three changeset guards, `Workers.PlatformChatGPTKeepalive` on the `29 4 * * *` cron, `Conversations.CodexChatGPT`, the table and every column.

The audit guardrail's exclusion table listed none of the deleted functions, so there was nothing to remove there.

## The platform lock contention test (#2188's PR 8a)

[...]

Issue #2188, opened 2026-09-14
refactor(chatgpt): remove the dormant tenant-owner half of ChatGPTAccounts; ADR 0052 stays Proposed


jhgaylor:
Stage 8 of #2176, gated on decision 1 there (taken 2026-09-14: **delete**). Independent of stages 1–6; lands as a short stack of three PRs in this order.

## Why

The tenant-owner half of `Fountain.ChatGPTAccounts` (ADR 0052 decisions 1, 3 and 6, built by #2011–#2017) is dormant: nothing in production inserts an owned row (the only writer is the admin page, `fountain_web/live/admin_live/inference.ex:76-133`, and the schema says so, `platform_chatgpt/account.ex:25-26`), and `credential_for_user/4`, `status_for_user/1`, `refresh_for_user/3` have no caller outside the context, its coordinator and two Oban workers. 0052's decisions 2, 4 and 5 (the linking UI and API, selection, per-peer sandbox identity) are the feature, and they are not scheduled. Roughly 700 lib lines and 1,700 test lines are reachable only from their own tests. The ADR stays Proposed; the code comes back from history when the feature is taken up.

## What stays

- The `platform_chatgpt_account` table, its `user_id` owner column, the two partial unique indexes and the `version_chatgpt_grants` columns (generation, `lock_version`). No migration.
- `ChatGPTAccounts.Reserved` (live in three changesets: `environments/secret.ex:30`, `vaults/vault_secret.ex:32`, `secret_bindings/binding.ex:78`).
- `ChatGPTAccounts.Cipher` and `RefreshLock`: both serve the platform row (`chatgpt_accounts.ex:580`).
- Every `platform_*` function, `PlatformChatGPT.{Account, Device, OAuth, Refresher, Tokens}`, `Workers.PlatformChatGPTKeepalive`, `Conversations.CodexChatGPT`.

## PR 8a — the platform lock has a contention test of its own, first

`chatgpt_user_refresh_isolation_test.exs:60` is the only test that exercises the platform refresh lock against a contender, and it does so through two *user* grants. Before anything is deleted, add a platform-only test in `platform_chatgpt_test.exs`: two concurrent `platform_refresh_serialized/1` callers, one upstream refresh, the second waits and reads the rotated row. Same barrier technique as the user test. This PR lands green on its own.

## PR 8b — the workers, sweep, coordinator and supervisor

[...]

PR #2453, opened 2026-09-20
docs(decisions): ADR 0060, several ChatGPT subscriptions per user

jhgaylor:
Adds **ADR 0060 — A user links several ChatGPT subscriptions, and a credential set names one**, Proposed. Decisions only; nothing here is built, and no code, column or endpoint changes in this PR.

## What it decides

Today the only managed ChatGPT subscription is the deployment's own (ADR 0047). Users cannot link theirs: ADR 0052's user surface was never built, and the grant table admits one row per user. 0060 rebuilds that surface with **many grants per user**, and settles three forks:

1. **A grant is a named row.** Drop `unique_index(platform_chatgpt_account, [:user_id]) where user_id IS NOT NULL`, add a `name` and `unique_index([:user_id, :name])`. The platform row's own partial index is untouched. Same table — its create migration already says "the same row with an owner, not a second table" — and a configurable per-account cap.
2. **A credential set names a grant** via a new nullable `chatgpt_grant_id`. This completes ADR 0053 decision 3, which reserved exactly this seam, and means an agent picks a subscription with `inference_credential_id`, the field it already uses to pick an API key. No second selection mechanism, and **no account-wide preference** — 0052 decision 4's user-level flag has nothing to point at once there are several subscriptions.
3. **A named grant is used or the run fails.** Revoked, expired or out of Codex usage produces an actionable error. Nothing switches to another of the user's subscriptions, to the set's own API key, or to platform inference.

Point 3 **amends ADR 0047 decision 6 as amended by #2362**. Skipping an exhausted grant for the platform API key is right for the platform grant, where the deployment pays either way; for a user grant it would silently convert the user's own exhausted subscription into platform-billed inference.

Custody is inherited from 0052 whole — its decisions 1 (ownership and encryption), 3 (one refresh owner per grant) and 6 (managed destinations not overridable by tenant bindings) were already written per grant. None of that is built either, and 0060 reduces none of what 0052 requires before any managed grant is brokered.

## Notes for the reviewer

- **The revert is only half the work.** At `105fb6d1` the lifecycle half is already per grant (`credential_for_user/4`, `refresh_for_user/3` take a `grant_id`), but the lookup half assumes one: `status_for_user/1` is `where: a.user_id == ^user_id` ending in `Repo.one()`, which raises `Ecto.MultipleResultsError` on a user's second grant. Every read-the-user's-grant call site is new work; the ADR says so in Context and stages it accordingly.
- **An open dependency:** ADR 0047's measurement 5 (idle lifetime) was due 2026-09-17 and still reads Pending. It sets the keepalive interval, and this ADR multiplies the grants that interval covers. Decision 5 names it rather than assuming a value.
- The `generation` and `lock_version` fencing columns 0052 decision 1 asked for already exist on the table, so decision 1 here is an index swap plus a `name`.
- Implementation is sequenced into five PR-sized stages, with 0052's adversarial cases required for the platform grant **and** for two grants of one user, which is the new case.

## Checks

Contributor-only Markdown path (`contributing/docs.md`):

- `python3 scripts/conflict-markers.py` — clean
- `python3 scripts/changelog.py check` — 3 fragments, 4 entries
- `git diff --check` — clean
- `scripts/decisions-index.sh` — index refreshed and committed
- `okf validate decisions` — `valid: true`, 0 errors (6 pre-existing WARNs on 0051 and others, none on 0060)

No changelog fragment: nothing a user or operator can observe changes.


jhgaylor:
Pushed `9d9061e` after reading the code more closely. One claim in the first draft was wrong, and three pieces of work were understated. Flagging them since they change the effort estimate, not the design.

**The wrong claim.** Decision 5 originally said several grants per user "need no new coordination model" because 0052's refresh lock was already per grant. Half true. `ChatGPTAccounts.RefreshLock` does key on `:erlang.phash2(grant_id)` (`chatgpt_accounts/refresh_lock.ex:42`), but every grant *write* also takes `InferenceCredentials.lock_platform_source/0` — a single deployment-wide advisory key (`inference_credentials.ex:612`) — and `lock_source/1` takes that same key **shared** before every `resolve/4` (`:604`). With one platform grant that is a non-event. With many user grants it serializes every grant's refresh against every user's credential resolution, deployment-wide. Decision 5 now says to key that lock by owner, and stage 1 does it before anything can contend on it.

**Understated, now named in decision 6** — all three block carrying two grants into one sandbox:

- `CodexChatGPT.auth_path/0` is a fixed `$CODEX_HOME/auth.json` (`conversations/codex_chatgpt.ex:118`).
- `prepare_sandbox/3` calls the global `ChatGPTAccounts.platform_sandbox_auth/0` (`:89`) and its caller (`conversations/provisioning.ex:994`) threads no user, conversation or `Source` — the lookup 0052 decision 5 already said to eliminate.
- The broker derives a placeholder from the env-var name alone (`broker.ex:250`), so **two grants produce the identical `__codex_chatgpt_access_token__`**, and `@inference` holds one entry per name per conversation (`:295-309`).

Also: `Egress` recognises the credential to refresh by comparing token **strings** (`conversations/egress.ex:136`), which 0052 decision 4 forbids and which cannot tell two grants apart. And `Source.platform?/1` (`inference_credentials/source.ex:38`) is the only origin signal, charging the deployment ceiling at `conversations/turn_machine.ex:1099` — so `:own` for a user grant is something decision 6 has to establish, not inherit.

**Decision 1 gains a second index,** unique on `(user_id, account_id)`: the same upstream account linked twice would give one subscription two independent refresh chains and two generations over a single quota, as rows OpenAI cannot tell apart.

**Revert hazard, now in Context.** #2188's deletions were `73fe870d` (#2198) and `4563e3df` (#2199) — ~715 lib and 1,447 test lines, no column or index dropped. But #2362 (`231773be`) later rewrote `do_refresh`, `current_result`, `current_query` and `swap_in`, the same functions whose owner branches #2199 stripped, so those need re-threading by hand. Everything else reverts mechanically; `Cipher` and `RefreshLock` are unchanged since `105fb6d1`.

Every line reference above was read against the tree, not inferred.

jhgaylor:
Merged with `--admin` at the maintainer's explicit instruction (2026-09-21). Why the bypass: every PR in the ADR 0060 stack is authored from the maintainer's own account, so none can receive the approving review the merge queue requires, and no approval was manufactured from a second account. What stands in for the queue: this PR's CI passed on its head commit, `main` has not moved since the stack was based on it, and after the squash the tree of `main` is verified to equal this PR's tested head tree before the next PR is merged. The feature stays dark: the `chatgpt_subscriptions` flag fails closed and is off. #2458 (the platform grant's move onto the protected path) is deliberately NOT part of this pass; it keeps its measurement gate.

PR #2454, opened 2026-09-21
feat(chatgpt): ADR 0060 stage 1, many named grants per user and an owner-keyed source lock

jhgaylor:
Stage 1 of [ADR 0060](https://github.com/managoat/fountain/pull/2453) — "The table and the context". **Stacked on #2453**; review that first. Nothing here is reachable by a user: no route, no UI, no selection, no env var. Stages 2–5 follow as their own PRs.

## What it builds

- **Migration A** (`name_chatgpt_grants`): `name`, check `chatgpt_grant_name_follows_owner` (platform row NULL, owned row non-blank), the one-row-per-user index replaced by unique `(user_id, name)` and `(user_id, account_id)`, plus unique `(id, user_id)` for stage 2's composite FK (unused until then). `platform_chatgpt_account_platform_row` is untouched. `down` raises rather than deletes if a user holds two grants.
- **Migration B** (`key_chatgpt_grant_source_lock_by_owner`): replaces `fountain_lock_inference_source()` so an owned grant row takes `'inference:' || user_id` and only the NULL-owner row takes `'inference:platform'`; also refuses an UPDATE that changes `user_id` (0052 decision 1: no transfer). `down` restores the previous body byte for byte.
- **Migration C** (`tombstone_disconnected_chatgpt_grants`): token ciphertext nullable, check `chatgpt_grant_tokens_follow_status`.
- **Context** (`Fountain.ChatGPTAccounts`): `list_for_user/1`, `get_for_user/2`, `connect_for_user/4`, `reconnect_for_user/4` (optionally fenced on `expected_generation`), `rename_for_user/4`, `disconnect_for_user/3`, `remove_for_user/3`, `grant_ceiling/0`, and the restored `credential_for_user/4` / `refresh_for_user/3` with `Grant`, `RefreshCoordinator`, `RefreshSupervisor` and the `Reserved` clause from `105fb6d1`. Every read and write is scoped by `user_id` **and** grant id; views are metadata only and never carry ciphertext.
- **Cap**: `config :fountain, :chatgpt_grant_ceiling, 5`, counted inside the owner's source lock. The env var and its docs row wait for stage 4, when an operator can first observe it.
- **Audit**: tenant events `chatgpt_grant.{connected,renamed,disconnected,removed,reconnect_required}`, after commit, with name/generation/method/plan only.

## Where the ADR was wrong about the code (corrected in the ADR here)

1. **The lock decision 5 re-keys is also taken by a database trigger**, for every row of the table whatever its owner. Re-keying only the Elixir call sites would have done nothing — hence migration B. The lock test's raw-insert case was run against the old trigger first and blocked.
2. **#2362 did not rewrite `do_refresh`, `current_result`, `current_query` or `swap_in`.** The re-thread is reversing #2199's hunks; #2362's exhaustion code stays platform-only until stage 5.
3. **No application writer for a user grant existed at `105fb6d1`.** Connect, reconnect, rename, disconnect, remove and the cap are new code.
4. A per-user key (`'inference:' || user_id`) already existed and `lock_source/1` already takes it, so the re-key is per user, not per grant. User-grant writers do not take the platform key even shared (a shared request queues behind a waiting exclusive one).

## Design calls made while building — please veto any you disagree with

- **Disconnect is a tombstone, not a delete** (`status: "disconnected"`, tokens NULLed, generation advanced, `lock_version` bumped, name kept). Stage 2's acceptance test is "a set naming a *disconnected* grant resolves with an actionable error and no fallback", so the reference and the name have to outlive the disconnect; a nilifying FK would be exactly the silent fallback decision 4 forbids, and a restricting one would let configuration block revocation. `remove_for_user/3` hard-deletes a disconnected row. Platform disconnect still deletes.
- **The grant id is bound into the user-grant cipher AAD.** The ADR says `Cipher` is inherited unchanged, but its AAD named owner + field only, so with several grants per user one grant's ciphertext would decrypt in another of the same user's rows. No owned row has ever existed, so this is free now and a data migration later. The platform AAD is byte-identical.
- Reconnect onto a different upstream account is allowed unless another of the user's grants holds it (the ADR did not say).
- User rows never set `updated_by_user_id`, so one account's deletion cannot write another account's grant row.

## Known, not fixed here

- **Keepalive gap**: linking ships in stage 4, the user keepalive in stage 5; the console gate stays off until stage 5 lands.
- The refresh-token loss window (OpenAI rotates, then a >20 s holder of the source lock makes `RefreshLock`'s transaction roll back) predates this; the re-key narrows it from the deployment to one user.

## Review

Two independent read-only reviews (tenant isolation/custody; correctness/concurrency/tests) found no critical or high issues. All six mediums and the cheap lows are fixed in the last 11 commits, each with a test seen failing first. One of them turned up a real defect: the refresh coordinator and `RefreshLock` keyed on the caller's spelling of the grant id, so an uppercase id started a second provider exchange for the same grant.

## Checks

Run on the pinned toolchain (Elixir 1.19.2 / OTP 28.3):

- Full `mix test`: fountain 6988 tests + 6 doctests + 2 properties, 0 failures; buzz 136, slack 30, support 33, google 60, microsoft 29, all 0 failures
- `mix precommit compile format credo` — passed; `mix precommit dialyzer` — 0 errors; sobelow — passed
- `scripts/decisions-index.sh` — no diff; `okf validate decisions` (0.3.0) — `valid: true`
- `python3 scripts/changelog.py check`, `python3 scripts/conflict-markers.py`, `git diff --check` — clean

No changelog fragment: nothing a user or operator can observe changes. No OpenAPI, SDK or `docs/` change.


jhgaylor:

PR #2456, opened 2026-09-21
feat(inference): ADR 0060 stage 2, a credential set names a ChatGPT grant with no fallback

jhgaylor:
Stage 2 of ADR 0060 (#2453) — "Selection". **Stacked on #2454** (stage 1). Still nothing a user can reach: no route or form can set the new field yet (that is stage 4), and a temporary guard refuses a grant source at admission until stage 3 builds its transport.

## What it builds

- **Migration** `sets_name_a_chatgpt_grant`: nullable `inference_credentials.chatgpt_grant_id` with a composite FK `(chatgpt_grant_id, user_id) → platform_chatgpt_account(id, user_id)`. Same-owner naming is a database fact, and the platform row (NULL owner) is unnameable. `NO ACTION`, `DEFERRABLE INITIALLY DEFERRED`.
- **`InferenceCredentials.set_grant/3`** — the only writer of the field; owner-scoped read of the grant, audited after commit (`inference_credential_set.chatgpt_grant_changed`), refuses a disconnected grant, re-sending the id a set already names is a no-op. `ChatGPTAccounts.remove_for_user/3` now refuses a grant named by sets with `{:named_by_sets, names}`.
- **`Source`** gains scope `:grant` (identity `chatgpt_grant:<id>`, revision = generation, origin `"own"`). `dump/1` writes the new keys only when set, so every persisted source still compares equal; the platform source's dump is pinned by a golden map.
- **Resolver**: a set that names a grant serves a **codex** run from that grant, outranking an env/vault `OPENAI_API_KEY`. An unusable grant (disconnected, revoked, expired, reconnect-required, exhausted, not found, brokerless) is always `{:error, {:chatgpt_grant_unusable, %{grant_id, name, reason, until}}}` — never the user's other grants, the set's own key, the default set, or platform inference. A set with a grant and no key is `missing` for non-codex OpenAI consumers and says so at selection. `validate_source/2` passes the grant error through instead of flattening it to `:inference_source_changed`.
- HTTP: 409 `chatgpt_grant_unusable` with an actionable message; the turn stream's `admission_refused` and team schedules carry the same sentence.
- **Temporary guard** `CodexChatGPT.transport_ready/1` at launch, reserve, the turn gate and turn creation: today's transport would pair a user's bearer with the platform row's `account_id`. Stage 3 deletes it.

- **NO ACTION, not nilify or restrict** (with stage 1's tombstone disconnect): nilify is the silent fallback decision 4 forbids; restrict lets configuration block revocation.
- **The FK is deferred.** Review found account deletion depended on PostgreSQL's emergent RI-trigger order (a test that deletes grants before sets failed with `23503`). Deferral makes it controlled. Cost: the FK can no longer surface as a changeset error, so the guards are `set_grant/3`'s scoped read and the `named_by_sets` refusal; the FK proofs run `SET CONSTRAINTS … IMMEDIATE`.
- "Resolves `:missing` with an actionable error" in the ADR described two different results. As built, an unusable named grant on a codex run is always the tagged error; `Source.missing/0` is only a grant-only set serving a non-codex OpenAI consumer.
- Reconnecting or repointing does not revive a conversation pinned to the old generation (0052 decision 5), so every such message ends "…and start a new conversation".

## Review

Two independent read-only reviews. **One HIGH, fixed:** a codex agent whose model has no `provider/` prefix (`gpt-5`) bypassed the named grant and ran on the set's own API key with no error, because the grant branch keyed on the model's provider. It now keys on the codex runtime, and `missing_for_model/3` and `SpriteEnv` close the same nil-provider hole. Also fixed: the guard covered binding only (now the turn gate too), a `:grant` source was dropped from the usage stamp (now `"own"`), the FK ordering above, the flattened stream error, and the lows. New tests were shown failing against the pre-fix code (15 failures), then passing.

Owed by stage 4 and recorded in the ADR's deferred list: the 409 body's `grant_id`/`grant`/`until` keys join `Schemas.Error` at the contract regeneration; the `/start` banner; the agent form's handling of a disconnected named grant; a changelog fragment.

## Checks

Pinned toolchain (Elixir 1.19.2 / OTP 28.3):

- Full `mix test`: fountain 7054 tests + 6 doctests + 2 properties, 0 failures; buzz 136, slack 30, support 33, google 60, microsoft 29, all 0 failures
- Test DB dropped and migrated from scratch; the new migration rolls back and re-applies cleanly
- `mix precommit compile format credo dialyzer` — passed (dialyzer 0 errors); sobelow — passed
- `okf validate decisions` — valid; `changelog.py check`, `conflict-markers.py`, `git diff --check` — clean
- Not run locally: `scripts/sdk-contract/generate.py --check` (no schema, operation or route changed; left to CI)

No changelog fragment: nothing observable before stage 4.


jhgaylor:

Issue #1910, opened 2026-09-11
codex's sqlite state runtime is not isolated per conversation on a shared sandbox


jhgaylor:
Split out of #1670, which is closed on its other two halves. This issue tracks the remaining Codex SQLite initialization failure on shared sandboxes.

## Current disposition — 2026-09-22

**Reproduced in production on 2026-09-22 and the mechanism is now evidenced, not inferred: SQLite lock contention (`SQLITE_BUSY`) on a state database shared by many live Codex app-servers.** Two deliverables, in order:

1. **Mitigation (small, Fountain-only, ship first):** retry the peer spawn when `initialize` fails with this specific error, with jittered backoff and a small attempt cap. See "Mitigation" below.
2. **The fix:** per-conversation `CODEX_SQLITE_HOME`, as in "Preferred investigation" below, now for *both* Codex home layouts (see "Two homes since ADR 0060").

The 2026-09-17 assessment (Fountain main `1103fe84`, `managoat_runtimes` 0.4.2, source inspection only) is kept below where it still holds.

## 2026-09-22 production evidence

Read-only: Fountain pod logs, the prod DB, and the sandbox itself (`ls` and a read-only query of Codex's own `logs_2.sqlite`).

- **9 failed turns across 5 conversations, 10:33–10:44 UTC, all on one sandbox** (`7b2d6b59-5a74-400c-941e-6836c77d721c`, sprite `fountain-1633dc35-c82b3439`) and one user ChatGPT grant (`c00d77a5…`, generation `63479ff2…`). Conversations `638dd3a6`, `4c18283b`, `fec50d75`, `b1d684b9`, `68bdb661` (`638dd3a6` once, the other four twice each).
- **Load:** 15 conversations of that grant were started on that sandbox between 09:59 and 10:41 (an app fanning out work on one agent). All share one `CODEX_HOME`, so one `state_5.sqlite`. Codex's log DB in that home records **16 distinct app-server processes**; ~8 were active in the failure windows.
- **Error, verbatim:**
  ```
  {:acp_error, :initialize, %{"code" => 1001, "message" => "Codex process has exited with code 1:\nError: failed to initialize sqlite state runtime under /home/sprite/.codex-grants/<grant>.<generation>: failed to initialize state runtime at /home/sprite/.codex-grants/<grant>.<generation>"}}
  ```
[...]

PR #2457, opened 2026-09-21
feat(broker): ADR 0060 stage 3, a user's ChatGPT grant runs through the protected path and never enters the sandbox

jhgaylor:
Stage 3 of ADR 0060 (#2453) — "Transport and custody". **Stacked on #2456** (stage 2). A conversation that resolves to a user's named ChatGPT grant now runs on it, through the broker's protected path, and the sandbox never holds the bearer. Still nothing a user can reach: no route or form can create a user's grant yet (that is stage 4), so in production no session carries one.

## What it builds

- **No library release was needed.** `managoat_broker` 0.14 already has the protected path 0052 decision 5 asks for (`Session.authorization`, `http_only`, `protected`, a per-request `Store.authorize/2`, `ProtectedRule`/`ProtectedCredential`). Fountain's store implemented `lookup/1` alone; stage 3 is the other half.
- **Migration `20260921021249`**: the pin on `broker_sessions` as plain columns — grant id, generation, owner (NULL for the deployment's grant), ChatGPT account id, `managed_revoked_at` — with two checks (the pin is all or nothing; a user's grant rides only that user's session). Deliberately no FK to the grant: a grant that is gone must deny, not cascade the session away from a conversation whose other egress is good.
- **Issuance**: `Sessions.create/1` with `managed:` locks the grant row `FOR SHARE` and re-reads it in the inserting transaction (`ChatGPTAccounts.lock_active_grant/1`). Every issuance path passes the grant (`Egress.session_opts/1`: provision, reattach, re-mint of an expiring session).
- **Every request**: `authorize/2` admits a request to the Codex backend only while the session is live and unrevoked **and** the grant row still says the same owner, generation, `active` and account. One row, no lock, nothing cached; a failed read is `:unavailable` (503), never a cached allow.
- **Invalidation**: one private seam, `revoke_broker/2`, called inside the transaction of every write that ends what a session was issued for. `update_rules/4` writes no `managed_*` column, so a delayed rewrite cannot restore a grant or move a session to a newer generation.
- **`ProtectedCompiler`**, restored from `105fb6d1` and reshaped to take no bearer: a fixed route (`POST https://chatgpt.com:443/backend-api/codex/responses`) for one account; refuses tenant input that names the managed credential or injects into that route.
- **`CODEX_HOME` per grant and generation** (`/home/sprite/.codex-grants/<grant>.<generation>`, never the grant's name), prepared by a link script because `managoat_runtimes` fixes `~/.codex`. **Migration `20260921023458`** adds `sandboxes.codex_peer_homes` so Codex sources with a home of their own share a machine.
- Stage 2's temporary `transport_ready/1` guard, its call sites, its 409 and its test are gone.

- **A secret's value may now mention the reserved name** (`2fd01c12`). This is the one change here that *loosens* a check. `Reserved.conflict?/1` was a substring match over every brokered value, and the same check runs when a secret is saved (on `main` since #2017), so a setup script or JSON blob mentioning `CODEX_CHATGPT_ACCESS_TOKEN` was refused. A value is now refused only when it is the reserved name by itself, contains either placeholder shape, or contains a `{{ CODEX_CHATGPT_ACCESS_TOKEN }}` reference; keys, binding fields, network patterns and the account id keep the strict rule. Injection into the protected route is stopped by `ProtectedRule.prepare/4` over the effective rules, which values never reach.
- **`Fountain.LogThrottle`** is a new small module and supervised child (ETS, once per key per minute per node, shaped on `FeatureFlags.Cache`). Without it a sandbox controls log volume when an owner's DEK or token fails to open.
- **The link script refuses a symlinked home or homes directory** (exit 3). A deployment that deliberately symlinks `/home/sprite/.codex-grants` would fail provisioning.
- **`CodexChatGPT.env/3` exports nothing** for a `:grant` source that pins no grant (it cannot return an error); `prepare_sandbox/5` refuses the same source with `:invalid_codex_home`.
- **A grant the broker will mint no session for is a permanent, named refusal** (`chatgpt_grant_unusable`, `retryable: false`), not the transient arm it fell into.

## Review

Two independent read-only reviews (security, correctness). **No critical or high finding.** Custody, cross-tenant isolation, generation fencing, `revoke_broker/2` coverage and the protected route were traced end to end and hold. Two mediums, fixed: the per-request path had an unbounded tenant-key read and the ADR understated its cost (now bounded at 5s; the ADR states the real read/AES counts); and "everything else under `~/.codex` stays shared" is order-dependent — the ADR and moduledoc now say so and name #1910. Lows fixed: the misreported `retryable` refusal, two error shapes for one condition, the platform-path fallthrough for an unpinned `:grant` source, the link script's EEXIST race and planted-symlink cases, `revoke_grant/2` touching expired rows, the probe's cleanup masking its failure, a missing `lock_timeout`.

**Not built, recorded in the ADR as gates before the console surface opens** — each needs a `managoat_broker` release, not Fountain code:

- **Gate A**: the grant bearer is the one brokered credential absent from the redaction registry. If `chatgpt.com` or a CDN error page ever reflects `Authorization`, it lands unredacted in `log_events`. Only the proxy holds the bearer, so the scrub belongs there.
- **Gate B**: `ProtectedRule` checks only the path and forwards any query string on the protected route. The pinned client sends none.

Also recorded as not built: a per-session request rate cap; removal of grant homes after disconnect; the setup script running before its `CODEX_HOME` exists; two rare PostgreSQL-detected deadlock shapes (account deletion vs `create_managed`; `update_rules/4` vs `revoke_grant/2`). **Nothing has been measured against the real codex client** — that is stage 5's controlled run, and it gates the stage 3b follow-up that moves the platform grant.

## Checks

- The 17 test files this stage touches: 290 tests, 0 failures
- `mix precommit compile deps format credo dialyzer sobelow` — passed (dialyzer 0 errors)
- `conflict-markers.py`, `changelog.py check` — clean
- A wider local run (conversations, broker, chatgpt, machines, crypto, environments, vaults, secret bindings): 2707 tests, 3 failures, none in a file this PR touches — two `provisioning_test.exs` CA tests (`fatal: not a git repository`: the worktree's `.git` pointer is not mounted in the container) and `termination_cluster_test.exs` (no peer-node connection in the pod). Not re-checked on the pre-fix tip; CI is the authority.
- Not run locally: the full `mix test`, the `release` precommit stage, `okf validate decisions` (not installed here), the SDK contract check (no schema, operation or route changed)


jhgaylor:

PR #2458, opened 2026-09-21
feat(chatgpt): the platform ChatGPT account rides the protected broker path; legacy sessions drain

jhgaylor:
A follow-up to stage 3 of ADR 0060 (#2453). **Stacked on #2457.** Stage 3 put a *user's* ChatGPT grant on the broker's protected path and left the deployment's own account (`/admin/inference`, ADR 0047) on the legacy path, where its bearer sits in each session's `rules_ciphertext`. This moves the platform grant onto the protected path too and drains the legacy sessions.

> [!WARNING]
> **Draft on purpose: this PR carries a merge gate that has not been passed.** It is the only piece of the 0060 stack that changes what a running deployment does: the platform grant has been live in production since 2026-09-08 (ADR 0047), and nothing here has been measured against the real codex client. **Do not merge or deploy until `scripts/probe-codex-protected.py` has been run against the image's client and one real hosted turn has succeeded, with the result recorded in ADR 0047.** If the installed codex needs a second `chatgpt.com` route, or ignores `CODEX_HOME`, every platform-codex turn fails with a 403 or finds no auth file — and the #2362 API-key fallback stays silent, because the grant is still `active` and not exhausted. ADR 0047's measurement 2 recorded 40 requests to `chatgpt.com` in one turn, one answered 204, without recording their targets; that is a reason not to assume the single `POST /backend-api/codex/responses` route is enough, not proof that it is not. The gate is written up in ADR 0060 ("The platform move is gated on a measurement"), and the changelog carries a "release gate, not yet passed" entry to delete once it passes.

## What it builds

- The platform grant is pinned on its broker session like a user's (owner NULL), authorized per request against the live row, and injected by the proxy. `PlatformInference.credential_for/2` hands out a placeholder; `platform_selection/0` returns no token. The egress token-string compare, `refresh_platform_chatgpt` and `prepare_shared` are gone; `ChatGPTAccounts.platform_credential/1` is deleted.
- **Migration `20260921025746`** deletes legacy platform-codex sessions (transactional, `lock_timeout` 5s, idempotent, `down` a no-op). A turn in flight gets a 407; the next server start re-mints.
- **The drain is continuous, not one-shot** (review fix): `lookup/1` answers 407 and the ordinary half of `authorize/2` denies for any session whose rules name the reserved credential, and deletes the row. No DB read per request; the log line is throttled. Without this, an old replica re-minting after the migration would leave a static, exportable bearer served by new replicas for up to the session TTL.
- Billing origin, the #2362 exhaustion fallback, the no-key case, revoked/expired handling and token rotation (no generation bump) are unchanged and tested.

## Upgrade notes (in the changelog fragment)

Both mixed-fleet directions: an old replica minting a legacy session after the drain (now refused by new replicas); and a new replica's managed session served by an old replica's proxy, where `chatgpt.com` gets the placeholder as bearer and the turn fails with a provider 401. Nothing leaks, but **codex turns on the deployment's account may fail until every replica is rolled — roll quickly or stop-the-world.** Conversations on old replicas whose sessions were drained do not recover until their node restarts. Also: a tenant's own `CODEX_HOME` in env vars or a vault is now dropped on platform-grant conversations.

- **Gates A and B** (#2457: a reflected bearer is not scrubbed; a query string is forwarded on the protected route; both need a `managoat_broker` release) **are open for the platform grant from the day this deploys.** The ADR says so and deliberately does not decide whether to deploy ahead of the library release.
- The continuous drain also deletes a *managed* session whose rules name the reserved credential — what an old node's rotation rewrite would produce. That goes past what review asked for.
- An `OPENAI_API_KEY` beside the platform grant is now `{:error, :codex_grant_key_conflict}` rather than "the key wins". Believed unreachable: the resolver reads the environment's and vault's keys before it reaches the platform.
- `platform_access_token/0` stays public and still returns the token; the refresh path depends on it. Noted in the ADR.
- Not built, documented: recovery after a 407 on new code (a session-existence read per turn would do it).

## Review

One independent read-only review (security and correctness). No bearer leak and no cross-owner authorization hole found; NULL-owner semantics, fallbacks, rotation and billing origin traced and hold. Findings fixed: the one-shot drain (medium), the missing mixed-fleet upgrade notes (medium), a vacuous `refresh_before_turn` test now driven through a real rotation, an inconsistent `:skip` branch, the dead token-returning API, and stale sentences in ADRs 0047/0052/0060 and the stage-3 fragment. Rebasing onto the reviewed stage 3 needed platform clauses in `gone/1` and `fenced/2`; without the latter an admin disconnect between resolve and mint would have raised.

## Checks

- The 25 test files stages 3 and 3b touch: 434 tests, 0 failures; every test naming the reserved credential plus `test/fountain/broker`: 355 tests, 0 failures
- `mix precommit compile deps format credo dialyzer sobelow` — passed
- `conflict-markers.py`, `changelog.py check`, `git diff --check` — clean; `decisions/index.md` unchanged after regenerating
- Not run locally: the full `mix test`, the `release` stage, `okf validate decisions` (not installed here). **Not run at all: anything against a real codex client — see the gate above.**


jhgaylor:
Merged with `--admin` at the maintainer's explicit instruction (2026-09-21): asked how to handle this PR, the maintainer chose "forward-port and merge now" over keeping it a draft behind its measurement gate. Why the bypass of the queue: the PR is authored from the maintainer's own account, so it cannot receive the approving review the merge queue requires, and no approval was manufactured from a second account. What stands in for the queue: CI passed on this head (26 checks), the branch sits directly on `main`, and after the squash the tree of `main` is verified equal to this tested tree. The real-client measurement is still owed (#2475, #2465). Merging deploys this: the first production codex turn on the deployment's account and `/admin/broker`'s denied table need watching right after the roll; rollback is a revert of the squash commit.

PR #2459, opened 2026-09-21
feat(api): ADR 0060 stage 4a, link a ChatGPT subscription by device code and name it on a credential set

jhgaylor:
Stage 4a of ADR 0060 (#2453) — the account surface's API half. **Stacked on #2457** (stage 3), deliberately *not* on the draft #2458, so it does not queue behind that PR's real-client measurement. This is the first stage a user can reach: link a ChatGPT subscription by OAuth device code, list, rename, reconnect, disconnect and remove it, and name one on a credential set. It merges dark: the flag `chatgpt_subscriptions` fails closed (in `@flags`, not in `@on_without_posthog`), and no SDK claims the new operations. The console is stage 4b.

## What it builds

- **`chatgpt_link_attempts`** (migration `20260921060000`, explicit `up`/`down`) and `ChatGPTAccounts.LinkAttempt`: a durable, owner-bound row. Device-auth id and user code are encrypted under the tenant DEK with an AAD of user, attempt and field; a check constraint clears them at every terminal state. `pending → completed | cancelled | expired | failed`, all terminal. A reconnect pins `grant_id` and `expected_generation` server-side; one open reconnect per grant (partial unique index).
- **Nothing was restorable from `105fb6d1`** — no attempts, console or API code ever existed. The admin's device flow is a task bound to the LiveView pid and dies with the page, so it could not be reused; this is shaped on `Fountain.Exports` (row + Oban + PubSub + `/api/account`) and on Connections' gating.
- **Worker** `Fountain.Workers.ChatGPTLinkAttempt` on a new `chatgpt` queue: args are two ids; one `device_poll` per run, snoozed at the server's interval (clamped 1..60 s); `device_exchange` outside any lock; then **`complete_attempt_for_user/4`**, which takes the tenant key, then the attempt row `FOR UPDATE`, then writes the grant in the same transaction (a `:within` option on `connect_for_user`/`reconnect_for_user`). Cancel takes the same locks in the same order, so no tokens commit after a cancel succeeded. A replay writes nothing; a late completion against a newer grant ends `failed`/`stale_grant` with the newer credential untouched.
- **Eight routes under `/api/account/chatgpt-subscriptions`**, behind full scope, `cache-control: no-store` wherever a user code appears. Bodies never carry tokens, the device-auth id, claims, `account_id`, `generation` or `lock_version`; `user_code` is null once terminal. Cross-user ids are 404, same query and shape as a missing id.
- **Credential-set API**: nullable `chatgpt_grant_id` and read-only `chatgpt_grant {id, name, status}`; `PATCH` validates the grant before any write.
- `CHATGPT_GRANT_CEILING`; a 7-day purge of ended attempts in `RetentionPruner`; `chatgpt_subscriptions_enabled` on `/api/auth/me`; regenerated contract (TypeScript, Swift), `docs/api.md`, feature status ("In development", a status defined there: gated and not yet working end to end).

## Gating

Only a **new link** asks the flag and a configured broker. List, rename, disconnect, remove, attempt read and cancel ask nothing — a deployment that loses its broker still holds refresh tokens and must be able to disconnect. A reconnect needs a broker, not the flag. The ADR now says plainly that reconnecting a disconnected tombstone can sign in a *different* upstream account, so it works as a new link with linking off.

- **A job polls ChatGPT, not the reader** ("jobs carry ids"; an API client may never poll).
- **The start limit is durable, per user and in the context**: ten starts an hour counted from the table under the tenant key, 429 with `retry-after`. The per-API-key plug was dropped rather than re-keyed (it also counted refused requests). **No deployment-wide ceiling on device codes** — a lock-free count is approximate, and any ceiling small enough to matter lets a few accounts deny everyone a sign-in. Recorded as not built.
- **Disconnect and remove end the grant's open reconnect attempt** in the same transaction, as `failed`/`stale_grant` rather than `cancelled`.
- **Names**: bounded in codepoints (matching the column and `maxLength`), and `\p{C}` is refused — which also refuses zero-width-joiner emoji.
- **The linking recheck at completion runs before the transaction and before the token exchange**, not inside the fence: the flag read may be an HTTP call, and no network I/O in a transaction wins.
- `account_email` is shown to the owner and never audited; `generation` appears in no body and was **dropped from three stage-1 audit events' metadata** (readable by a sprite-scoped key via `GET /api/audit`; nothing consumed it; never released). `attempt_id` joins `chatgpt_grant.connected`.
- `auth_unreachable` is exposed on the attempt; recently ended attempts are readable for 30 minutes (for the console's "why did my sign-in fail").
- A pending link attempt reserves its name, not a slot under the ceiling.
- **`FEATURE_FLAGS_ON=chatgpt_subscriptions` can force linking on at a self-host today**, while #2457's Gates A and B are unpaid. The docs warn against it; nothing enforces it.

## Review

Two independent read-only reviews. **No critical or high finding.** Tenancy, lock order, replay, generation pinning, secret handling (bodies, Oban args, logs, PubSub, `inspect`, ciphertext cleared at terminal states), authorization and audit-after-commit were traced and hold. Mediums, fixed: the start limit was per API key per node and bypassable by minting keys or looping start→cancel — unbounded `device_start` calls to the host every grant's refresh uses; and the worker had no `backoff/1`, so after ~9 snoozes one raised exception scheduled the retry past the 15-minute deadline and an approved code never linked (now a constant 10 s). Lows fixed: the shared queue sized by a per-account number (3 → 10, with the refresh's tight Finch options on all three device legs); grapheme-vs-codepoint name length and NUL bytes as 500s after a device code was spent; an unmapped tagged atom as a 500; the stale clock across `device_start`; in-flight new links completing after the flag goes off; a non-atomic set `PATCH`; pending rows with no job now re-enqueue on read.

**Known limitations, in the ADR**: a job orphaned in `executing` by a killed node is rescued after the attempt expires (Lifeline 30 min > TTL 15 min); the TTL is a hard 15 minutes (the device-start answer carries no `expires_in`); `Retry-After` is not read; a cancel after exchange drops tokens without upstream revocation; a purged `pending` row leaves a `started` event with no ending; over HTTP a suspended account gets 401 on every route and so cannot clear its stored refresh tokens; OAuth-client tokens are full-scope keys, so a self-registered app reaches this surface, user code included (the existing trust model, stated); device-code phishing is inherent — `verification_url` is built from config, never from a response, and `docs/api.md` tells integrators to show a code only to the user who started it.

## Checks

- All chatgpt and platform_chatgpt tests, `test/fountain_web/controllers`, workers, guardrail, config, schema and docs tests: 1864 tests, 0 failures. Docs suites: 0 failures. The migration was run down and up twice.
- New tests were shown biting: dropping the owner scope fails the ownership test; removing the completion fence fails 9 of 17 completion tests, both real-contention race tests included.
- `mix precommit compile deps format credo dialyzer sobelow` — passed
- Contract: `SDK wire contract: ok (181 operations, 203 schemas)`; TypeScript `generate.py --check`; `verify-contract` for TypeScript and Python; `sdk/elixir` contract test; `conformance/lint.py` — all pass
- `conflict-markers.py`, `changelog.py check`, `git diff --check` — clean
- **Not run locally**: Swift (`generate.py --check --target swift`, `swift test --filter ContractTests`) — no toolchain here; the Swift model edits were applied by hand to match `generate-swift.py`'s output, so **CI is the first real check of them**. Also not run: the full `mix test` after the last fixes, the `release` stage, `okf validate decisions`, `check-propagation.py`, Go checks.


jhgaylor:
Labelled `release:skip-sdk`: the regenerated contract adds fields to generated TypeScript and Swift types (`AuthMe`, the error payload) for a surface that is gated off and that no SDK claims (`sdk/contract/omissions.json` omits `/api/account/*`). Per CONTRIBUTING.md ("Do not bump an SDK's version because the contract moved"), no package version moves here. The Swift contract check itself passed in CI, which confirms the hand-applied Swift model edits match the generator.

jhgaylor:

PR #2460, opened 2026-09-21
feat(console): ADR 0060 stage 4b, the ChatGPT subscriptions card and a credential set's subscription picker

jhgaylor:
Stage 4b of ADR 0060 (#2453) — the account surface's console half. **Stacked on #2459** (stage 4a). Everything a user does with a ChatGPT subscription over the API, they can now do at `/account/inference-credentials`. Still dark: the flag `chatgpt_subscriptions` fails closed, and stage 5 leaves it off until the rollout checklist is met.

## What it builds

- **The ChatGPT subscriptions card** (`InferenceCredentialsLive.SubscriptionsCard`, a LiveComponent): connect (name → user code, verification link, expiry in UTC), cancel, rename, reconnect, disconnect, remove, count and limit, empty state. The page subscribes to 4a's per-user topic and re-reads on `{:chatgpt_grants_changed, user_id}`; **no process holds sign-in state**, so a reload or a second tab shows the same pending code from the database, and a completion flips every open page without a reload.
- **Why a sign-in ended survives a remount**: notices are computed from 4a's `list_recent_attempts_for_user/1` (30 minutes), not from component state — a user who switches to the browser to approve and comes back to a dropped socket still sees "already linked — reconnect *Work* instead" or "superseded by a newer sign-in".
- **Reconnect keeps the old credential serving** until the new one commits; disconnecting discards an open sign-in and says so.
- **A credential set picks a subscription**: a provider-style row about the selected set (the set panel is hidden for accounts with at most one set, so the default set could otherwise never name a grant), with the confirm on Save: naming a grant ends the set's running codex conversations. An account with no set gets its Default set when it names a grant.
- **The `/start` banner and the agent form report an unusable named grant** (stage 2's deferred items), via a new `InferenceCredentials.named_grant_problem/4` that delegates to `resolve/4`, recomputed only when the model, runtime, set or environment changes.
- One grant-state function for the card and the resolver: `Resolver.grant_state/1` is public (body unchanged) behind `InferenceCredentials.grant_state/1`, so the card cannot say Connected where a run would be refused — a new **"Cannot serve here"** chip covers a deployment with no broker, and an ineligible owner is said on the card, the picker, the form and `/start`.
- Guide `docs/guides/chatgpt-subscriptions.md` (in development; **only approve a code you started yourself**; Fountain cannot revoke at OpenAI; **a spent plan is not yet detected** — stage 5), feature status, ADR "Stage 4b as built".

## Gating

The card renders when linking is enabled or the account holds a grant. Flag off: the Connect form and the picker's options for a *different* grant are hidden **and their handlers refuse** (the picker re-asks `linking_enabled_for?/1` at the event, not a value cached at mount); a set that already names a grant still shows it and can clear it; rename, reconnect (with a broker), disconnect and remove stay operable.

- **`missing_for_model/3` returns nil for a codex agent on a named grant even with no broker**, flipping two stage-2 assertions. Review reconstructed why stage 2 asserted the opposite (nothing else told the form the run would fail; stage 2's notes flag it as owed) and confirmed the flip: the resolver takes the grant or errors before any key is considered, so collecting a key was misleading; the form now warns through `named_grant_problem/4` (`:broker_required`); the nil-provider hole stays closed; launch is still refused with no fallback.
- `has_any_credential?/1` and the admin funnel count a set naming an `active` grant, whatever its finer state — the dashboard checklist may skip the credential step for a grant that cannot serve; the banner and the form then say why. Recorded in the ADR.
- The console gates naming a *different* grant on the flag; 4a's API does not.
- **The user code is not in the dead-render HTML** (the API sets `no-store` on the same data; a page GET cannot): the disconnected mount reads no attempts and shows a placeholder. The code still lives in the assigns while an attempt is pending, so a LiveView crash report would carry it — recorded, with the known crash triggers closed (catch-alls on every handler; 4a refuses NUL and bidi names).
- Ended-sign-in notices reappear on every mount for 30 minutes (newest three) until that page starts another sign-in.
- No new JavaScript: the expiry is a clock time, not a countdown; an expired code leaves an open page within 60 s, when the job's next run broadcasts.
- A page that has shown the card keeps it until reload, so "Removed *Work*." is seen after the last grant goes with the flag off.
- The guide sits in the Catalog beside the codex runtime page, not under "Connect a service".
- The verification link renders only for `https://auth.openai.com`, port 443, no userinfo; otherwise the constant URL is printed as text. (It is built from config server-side; the predicate is defence in depth.)

## Review

Two independent read-only reviews. **No critical or high finding.** Event trust (every handler passes the socket's user and the page's attribution; foreign, malformed and huge ids are a generic not-found), key-by-key projection (no token, device-auth id, claim, `account_id` or `generation` in assigns, the DOM or `phx-value-*`), escaping, the PubSub topic, and no fallback at launch were traced and hold. Mediums, fixed: console starts were unbounded because a LiveView bypasses a controller plug (4a's limit now lives in the context and the card says how long to wait); the picker dead-ended for an account with no set after one failed save; a failed sign-in left no trace after a socket reconnect; the guide presented "Usage spent" as working. Lows fixed: no catch-all on the picker's handler; the user code in the GET body; resolution on every form keystroke; the restated grant state; success paths relying on a broadcast that a no-op write never sends; "Name" prefixed to every changeset error; labels and long-name overflow.

The review's tests of the tests: the page-reload test drives completion through the real worker and the real broadcast with no sleeps; the redaction test searches three renders and `inspect(load/1)` for the fixture's actual token, refresh token, device id, auth code, verifier, account ids and generations.

## Checks

- Live, controllers, chatgpt, platform_chatgpt and inference suites, guardrail, config, docs, schema, api_spec, audit, cross-tenant and conversations: 4099 tests + 2 properties, 3 failures — exactly the known container-only ones (two `provisioning_test.exs` CA tests: the worktree's `.git` pointer is not mounted; `termination_cluster_test.exs`: no peer node). CI is the authority.
- `mix precommit compile deps format credo dialyzer sobelow` — passed; `scripts/test-docs.sh` — 0 failures
- `conflict-markers.py`, `changelog.py check` — clean; `scripts/decisions-index.sh` — no diff
- Not run locally: the full `mix test`, the extension suites, the `release` stage, `okf validate decisions`. **The card has not been opened in a real browser**: the copy hook and `data-confirm` are covered by attribute assertions only.


jhgaylor:

PR #2461, opened 2026-09-21
feat(chatgpt): ADR 0060 stage 5a, a spent subscription is confirmed with OpenAI and refused by name, and a turn says what served it

jhgaylor:
Stage 5a of ADR 0060 (#2453) — accounting, first half. **Stacked on #2460** (stage 4b). A user's ChatGPT subscription that has spent its plan is now detected and refused by name with its reset time, and **nothing is ever substituted for it** (decision 4). A turn says which inference source served it. Still dark: the flag stays off. Stage 5b (keepalive fan-out, export and deletion, the ADR's "Stage 5 as built", the rollout checklist and the maintainer runbook) follows.

## What it builds

- **Exhaustion for a user's grant, confirmed by OpenAI.** A failed codex prompt whose error passes `UsageLimit.hint?/1` on a `:grant` source starts a background check, scoped by the conversation's owner. The hint comes from an adapter inside a tenant-controlled sandbox, so **it is only a hint**: the row is read through the eligible-owner query by owner, grant id, generation and `active`; `usage_checked_at` is claimed (one conditional UPDATE, 300 s cooldown, under the owner's tenant source lock — never the platform key); the bearer is decrypted from the row *the claim returned* and sent to `/wham/usage` with no lock or transaction held; only a confirmed limit writes `usage_exhausted_at`/`until`, fenced the same way. A forged hint writes nothing. Audit `chatgpt_grant.exhausted` and the owner's broadcast happen after commit. No migration: the columns already existed.
- **The platform path shares that code with an owner argument** (nil = the deployment's row). `platform_chatgpt_exhaustion_test.exs` passes unedited, including the #2362 fallback to the platform API key — which remains reachable only from a platform source.
- **No fallback.** Every later launch, turn, wake and direct server start on that grant is refused `chatgpt_grant_unusable` / `exhausted` with `until` — on the 409 and on both stream stages — including a conversation already pinned to it. Reset is lazy: a passed `until` reads as nil and nothing is written.
- **A turn says which source served it**: read-only `inference: {origin, scope, chatgpt_grant_id}` on the turn JSON and the export, from `turns.inference_source` (already written once per turn; no schema change). Never `generation`, `identity` or `revision`; null for a row with no source or an unknown scope; the id keeps naming a grant after it is removed. A grant turn is stamped `own`, is not priced and never reaches the platform ceiling (`ee/test/fountain/credits_chatgpt_inference_test.exs`, with `PlatformInference` enabled).
- The console: an open page flips to **"Usage spent"** with the reset time without a reload, and re-reads by a timer at the earliest reset it shows (the reset broadcasts nothing). The guide and feature status no longer say a spent plan goes undetected.

- **Which grant served a turn lives in `turns.inference_source`, not in the `usage` stamp**, though the ADR says it "joins the usage stamp": the pricer and `Turn.inference_stamp_only?/1` key off `usage`. 5b records the divergence in the ADR.
- **A failed or inconclusive check still spends the 5-minute cooldown** (the claim precedes the HTTP call, or a sandbox could loop it). A genuinely spent plan can therefore fail turns with codex's own message for up to 5 minutes before it is refused by name. Same as the platform; now documented.
- **A direct server start on a spent grant still ends the conversation `failed`**, like a disconnected grant — for a temporary state that may be too harsh (a deploy while a turn runs on a spent grant; exhaustion landing between the wake check and server init). The prompt's wake door refuses *before* any server starts and leaves the conversation alone. Only the message shape was changed here; the behaviour is pinned in a test with a comment.
- The broker stays unaware of exhaustion: an exhausted grant is `active` with a good token, so the proxy still authorizes it; selection refuses it; a turn in flight ends on OpenAI's own 429.
- Detection only after a failed turn: no proactive usage polling, no email or webhook. An exhausted `active` grant still counts for `has_any_credential?/1`.
- `inference.scope` includes `missing` and `none`; the object is inline on `Turn`; the team summary (a smaller turn shape) carries none.
- Stale fence at write time is `:ignored` (was `:unchanged`); a failed write transaction is `{:error, :usage_write_failed}` and logged with the grant id and step only; `UsageLimit.loggable_reason/1` is public because dialyzer proves a private catch-all dead against today's callers.
- The audit actor is the existing `system:chatgpt_accounts`.

## Review

Two independent read-only reviews. **No critical, high or medium security finding; no platform regression.** Traced and holding: a hint cannot send another owner's bearer or the platform's anywhere (no query compares `== nil`; a binary owner can never take the platform clause or key); the usage host comes from config only; the claim is race-free across nodes; no path gets platform inference stamped `own`; the refusal is returned before overrides, the set's key, the default set or the platform are considered. Findings, fixed: the no-fallback acceptance proved only one of its six alternatives usable, so it could have passed vacuously — **each alternative is now asserted by the value it hands the runtime, before the limit is recorded and again after the refusals**; driving the real reattach path exposed a server start that published an inspected tuple instead of the structured refusal; a throttled hint no longer opens a transaction or takes the tenant lock (telemetry count of lock queries: 0); `start_child` failing can no longer crash a ConversationServer finishing a failed turn; the usage fetch is wrapped so a dependency's stacktrace can never carry the header list, and the platform's log line goes through the same sanitiser; the bearer could be stale by a concurrent refresh; `chatgpt_grant_id` renders only as a valid UUID; an unboxed test now races two checkers on independent connections for exactly one upstream call; the restart case the docs claimed is tested.

## Checks

- ee suite, `test/fountain/conversations`, controllers, live, all chatgpt tests, broker, platform exhaustion (unedited): 4085 tests + 2 properties, 3 failures — exactly the known container-only ones (`provisioning_test.exs:590`, `:681`: the worktree's `.git` pointer is not mounted; `termination_cluster_test.exs:6`: no peer node). Guardrail, schema and docs tests: 280 tests, 0 failures. CI is the authority.
- `mix precommit compile deps format credo dialyzer sobelow` — passed; `scripts/test-docs.sh` — 0 failures
- Contract: `SDK wire contract: ok (181 operations, 203 schemas)`; TypeScript `generate.py --check` clean; TypeScript and Python `verify-contract`, the `sdk/elixir` contract test and `conformance/lint.py` passed on the pre-rebase tree (no schema changed since)
- `conflict-markers.py`, `changelog.py check` — clean
- **Not run locally**: Swift — no toolchain here; the `TurnInference` model was applied by hand to match `generate-swift.py`'s output (the same approach passed CI's Swift contract check on #2459). Also not run: the full `mix test`, the `release` stage, `okf validate decisions`.
- Labelled `release:skip-sdk`: generated TypeScript and Swift types gain the turn's `inference` object, and per CONTRIBUTING.md no SDK version moves because the contract moved.


jhgaylor:

PR #2462, opened 2026-09-21
feat(workers): ADR 0060 stage 5b, idle subscriptions stay signed in, the export and deletion know them, and the rollout is written down

jhgaylor:
Stage 5b of ADR 0060 (#2453) — accounting and rollout, second half, and the last stage. **Stacked on #2461** (stage 5a). Idle subscriptions are kept signed in, the account export and deletion know about subscriptions, and the ADR gains "Stage 5 as built" with a rollout checklist and the maintainer's controlled-run runbook. **The flag `chatgpt_subscriptions` stays off**: opening it is gated on things no PR in this stack can do (below).

## What it builds

- **Keepalive fan-out** (closes stage 1's recorded gap, "an idle user grant still lapses"). A daily sweep (`Workers.ChatGPTKeepaliveSweep`, `maintenance` queue, `37 4 * * *`, leader-only) pages an ids-only `_unsafe_due_user_grants/2` — active, ChatGPT, holds a refresh token, eligible owner, idle past the 6-day cutoff; never the platform row, a tombstone, a revoked/expired grant or a suspended owner's — and queues one `Workers.ChatGPTGrantKeepalive` per grant on a new queue `chatgpt_refresh: 2` (below the refresh coordinator's 4 per node, so a turn's own renewal always has room), jittered over `min(6 h, max(300 s, due × 5 s))`: about one request per 5 s toward `auth.openai.com` however many grants exist. The sweep's answer is only a hint: each job re-checks through the owner-scoped `refresh_for_user/3`. Forged args (another owner's id, the platform row's id) cancel with no HTTP call. `refresh_token_reused` on A marks only A reconnect-required; B of the same user is still renewed. The workers are restored from `105fb6d1` and adapted.
- **A per-node circuit breaker toward the auth server** (`ChatGPTAccounts.RefreshBreaker`), because upstream throttling of Fountain's address is a cross-tenant outage. Evidence is a 429, or a 403 whose body is not a JSON object (a 403 that names anything is that grant's own `:refresh_failed`). It opens only when **two different owners** are refused within 10 minutes; each grant is heard once per window. Open, it is **half-open**: one probe per node per 15 minutes; a refused probe is a snooze that spends no attempt; **any successful renewal closes it for everyone**. A held job's wait is capped at the two-hour probe deadline and at its grant's seventh idle day; a job stops for good 72 hours after its first run. A turn's own renewal never waits on it.
- **Export**: `chatgpt_subscriptions` and `chatgpt_link_attempts` sections, key-by-key projections from owner-scoped reads that decrypt nothing — never a token, ciphertext, device-auth id, user code, provider account id, generation or lock version (the test searches the rendered JSON for the fixture's real values). **Deletion**: proved cascade-only under `SET CONSTRAINTS ALL IMMEDIATE` (grants, attempts, broker sessions; another user's untouched); `chatgpt_grants_removed` joins the `account.deleted` metadata; queued keepalives and pollers for a deleted account end quietly with no HTTP call; the account-deleted email says plainly that **Fountain cannot revoke a sign-in at OpenAI** (`PlatformChatGPT.OAuth` has no revoke leg) and to sign the device out there.
- Five metrics: `fountain_chatgpt_keepalive_sweep_due`, `fountain_chatgpt_keepalive_grant_count{result}`, `fountain_chatgpt_refresh_rate_limited_count`, `fountain_chatgpt_refresh_breaker_opened_count`, `fountain_chatgpt_refresh_breaker_closed_count`. Log lines carry a grant id and an atom only; the `:error` lines are throttled. No new environment variable.
- **Docs and decisions**: the guide (an idle subscription stays signed in; export and deletion), feature status (still "In development"; the two broker protections and upstream revocation stay not built), the architecture cron table, ADR 0060 **"Stage 5 as built"** for both halves — including the recorded divergence that the serving grant lives in `turns.inference_source`, not the `usage` stamp — ADR 0047's amendment note, and the new unpublished **`contributing/chatgpt-subscriptions-rollout.md`**. `adr_status` is left as it is: **accepting the ADR is the maintainer's decision.**

## What this stack leaves to the maintainer (the rollout checklist, in the new file)

1. **#2458** (draft): the platform grant's move onto the protected path, gated on `scripts/probe-codex-protected.py` against the image's client and one real hosted turn.
2. A **`managoat_broker` release** with Gate A (scrub or refuse a protected response that reflects the request's bearer) and Gate B (refuse a query string on the protected route), then the pin bump (#2457).
3. **The controlled run** — the ADR's own last acceptance step, written up as a runbook: pin and record versions; two subscriptions, two sets, two agents, one shared sandbox; a turn on each with `broker_requests` evidence; a forced refresh of A (staging database only) with B untouched; reattach through the symlinked home; restart; disconnect A mid-tunnel; optionally exhaustion. **Also to observe there, because it is unmeasured and the breaker rests on it: whether `auth.openai.com` throttles per address or per account, and what a per-account 403 looks like.** Results go in ADR 0060's new "Measured" table.
4. ADR 0047 measurement 5's day-9 and day-30 readings (the keepalive interval is provisional on them); seven days of keepalive observation; a staged PostHog release before `@on_without_posthog`. Rollback is the flag: existing grants stay operable.

- **The breaker** is per node, not cluster-wide (each node learns from its own refusals); the platform grant counts as an owner for evidence; success clears the evidence heard so far; "once per grant per window" applies on every path, since the context cannot tell which path asked. Under abuse, a victim's job is delayed by at most the capped wait, and its probe succeeds; under genuine throttling a node sends one probe per 15 minutes. **A victim's job is no longer guaranteed its own request after two hours** — it competes for the node's single probe.
- A `:rate_limited` answer is a snooze and spends no attempt; `max_attempts: 3`; the absolute stop is 72 hours; **no damping for a grant that keeps failing without a terminal code** (it is retried daily; needs durable per-grant state — recorded as not built).
- A disconnect bumps the generation, so a job queued before it cancels as `:stale_grant`.
- The export includes `account_email` and pending attempts (without the user code); `version` stays 1. Deletion counts tombstones. The email's OpenAI paragraph appears only when the account held a grant row; the count rides in the email job's args.
- Not built, all in the ADR: upstream revocation; a cluster-wide breaker; reading `Retry-After`; an index for the due scan (the table is bounded by users × ceiling); credential sets in the export beyond `named_by_sets` names; tombstone pruning; a grant linked behind a running sweep cursor waits a day; a run or sweep page killed at Oban's timeout is not logged.

## Review

Two independent read-only review rounds, security and correctness together. **Round one** (the code half): no critical finding; one **high, marked plausible** because it rests on an unmeasured assumption — the first breaker tripped on any non-terminal 403 or 429 and discarded the code, so one tenant (a grant that reliably answers a bare 403, or a scripted turn loop) could hold a node's breaker open and starve everyone else's keepalives against a thin margin (worst-case first attempt about 7 d 6 h after the last renewal, against an assumed 8-day idle window). Fixed by the evidence rule, the two-owner bar, the deferral cap and once-per-grant hearing. Also fixed: a thundering herd when the breaker cleared; a grant skipped for a day behind an old incomplete job; a silent stop when a sweep page exhausted its attempts. **Round two** (the rewrite): nothing to revert; three mediums, fixed — the two-hour cap did not bound a snooze of up to 6 h 15 m (and three sentences overstated it); under *real* throttling refused probes retried in 30–900 s and sent more requests, not fewer (now half-open, one probe per node per pause); a successful renewal never closed the breaker. Lows fixed: a deferral stamp never cleared; no termination argument for "never gives up while due"; a JSON-object 403 under an unrecognised key counted as throttling; the breaker's ETS calls could raise on the turn path if the table owner died. **The final fix pass has been tested but not independently re-reviewed.** The first round also looked for a cause, in this diff, of one full-suite run with 120 `DBConnection` `queue_timeout` failures and found none; it did not reproduce in three later full runs (a second worktree was running suites against the same Postgres container at the time).

## Checks

- **Full `apps/fountain` suite: 7464 tests + 6 doctests + 2 properties, 3 failures** — exactly the known container-only ones (two `provisioning_test.exs` CA tests: the worktree's `.git` pointer is not mounted; `termination_cluster_test.exs`: no peer node). Targeted suites: 1170 tests, 0 failures. CI is the authority.
- `mix precommit compile deps format credo dialyzer sobelow` — passed; `scripts/test-docs.sh` — 0 failures across its five suites
- `conflict-markers.py`, `changelog.py check` — clean; `scripts/decisions-index.sh` — no diff
- Not run locally: the ee and extension suites outside the docs script, the `release` stage, `okf validate decisions` (not installed here), the SDK contract checks (no schema changed). **Nothing in this stack has been run against the real auth server or a real codex client** — that is the runbook.


jhgaylor:

